PR #425, opened 2020-02-12
[Release 1.1] Revert Async changes in SNIPacket to fix deadlock issues + Update SNI Package Reference

cheenamalhotra:
Addresses internal request to publish Deadlock issue fix in a Stable Release by patching **v1.1.0**.

This PR brings below changes for the new driver **v1.1.1**:
- Ports PR #349 (Reverts Async changes (https://github.com/dotnet/corefx/pull/34184) in SNIPacket class to fix deadlock issues)
- Ports PR #420 (Update SNI package reference to allow auto-update of patch releases)

✔️ Deadlock issue tested and validated with repro from [alex-zyl/EFCore3OnLinuxIssue](https://github.com/alex-zyl/EFCore3OnLinuxIssue)


tools/specs/Microsoft.Data.SqlClient.nuspec
```diff
@@ -21,17 +21,14 @@ Microsoft.Data.SqlClient.SqlTransaction
...
-
 When using NuGet 3.x this package requires at least version 3.4.</description>
     <releaseNotes>https://go.microsoft.com/fwlink/?linkid=2090501</releaseNotes>
     <copyright>© Microsoft Corporation. All rights reserved.</copyright>
     <tags>sqlclient microsoft.data.sqlclient</tags>
     <dependencies>
       <group targetFramework="net46">
         <dependency id="System.Data.Common" version="4.3.0" />
-        <dependency id="Microsoft.Data.SqlClient.sni" version="1.1.0" />
+        <dependency id="Microsoft.Data.SqlClient.sni" version="[1.1.0, 1.2.0)" />
```
cheenamalhotra: @David-Engel 

_Highlighting this additional change:_
I'll also port this to 'master' branch, once you confirm!

karinazhou: Why we need the upper bound here?

cheenamalhotra: Thanks! I got it confused with auto-fetch of SNI package dependency, but no upper bound is needed.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft.Data.SqlClient.csproj
```diff
@@ -341,7 +341,7 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.Data.SqlClient.SNI">
-      <Version>1.1.0</Version>
+      <Version>1.1.*</Version>
```
cheenamalhotra: Although this may be not a big concern, but I'm thinking of using a range here. 
```suggestion
      <Version>[1.1.0,)</Version>
```
Open for suggestions, which one is better and safer.

cheenamalhotra: Comment [here](https://docs.microsoft.com/en-us/nuget/concepts/package-versioning#examples) just got my attention:

>Always specify a version or version range for package dependencies in project files, packages.config files, and .nuspec files. Without a version or version range, NuGet 2.8.x and earlier chooses the latest available package version when resolving a dependency, **whereas NuGet 3.x and later chooses the lowest package version.** Specifying a version or version range avoids this uncertainty.

But testing it locally with NuGet.exe v5+ does not pick smaller version, it does pick largest package version. 🤔

cheenamalhotra: Picked up range with upper bound, after internal discussion.

tools/specs/Microsoft.Data.SqlClient.nuspec
cheenamalhotra: Confirmed range with upper bound, after internal discussion.

PR #427, opened 2020-02-12
Add Enclave Simulator for Internal Testing

karinazhou:
- Added support for enclave simulator bypassing the actual attestation.
- Added `BuildSimulator` property to enable the simulator support.

How to use the enclave simulator: 

Given `Attestation Protocol = SIM` and `Enclave Attestation Url= dummyURL` in the connection string and run the AE enclave-enabled tests.

NOTE:  This change is for internal testing only. The simulator support should be disabled in the official nuget package releases.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.NetCoreApp.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.NetCoreApp.cs
@@ -113,6 +113,11 @@ namespace Microsoft.Data.SqlClient
         /// <include file='..\..\..\..\doc\snippets\Microsoft.Data.SqlClient\SqlConnectionAttestationProtocol.xml' path='docs/members[@name="SqlConnectionAttestationProtocol"]/AAS/*' />
         AAS = 1,
 
+#if ENCLAVE_SIMULATOR
+        /// <include file='..\..\..\..\doc\snippets\Microsoft.Data.SqlClient\SqlConnectionAttestationProtocol.xml' path='docs/members[@name="SqlConnectionAttestationProtocol"]/SIM/*' />
+        SIM = 2,
+#endif
+
         /// <include file='..\..\..\..\doc\snippets\Microsoft.Data.SqlClient\SqlConnectionAttestationProtocol.xml' path='docs/members[@name="SqlConnectionAttestationProtocol"]/HGS/*' />
         HGS = 3
     }
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
@@ -3017,23 +3017,41 @@ namespace Microsoft.Data.SqlClient
 
         private bool IsValidAttestationProtocol(SqlConnectionAttestationProtocol attestationProtocol, string enclaveType)
         {
-            switch (enclaveType)
+            switch (enclaveType.ToUpper())
             {
                 case TdsEnums.ENCLAVE_TYPE_VBS:
                     if (attestationProtocol != SqlConnectionAttestationProtocol.AAS
+#if ENCLAVE_SIMULATOR
+                        && attestationProtocol != SqlConnectionAttestationProtocol.HGS
+                        && attestationProtocol != SqlConnectionAttestationProtocol.SIM)
+#else
                         && attestationProtocol != SqlConnectionAttestationProtocol.HGS)
+#endif
                     {
                         return false;
                     }
                     break;
 
                 case TdsEnums.ENCLAVE_TYPE_SGX:
+#if ENCLAVE_SIMULATOR
+                    if (attestationProtocol != SqlConnectionAttestationProtocol.AAS
+                        && attestationProtocol != SqlConnectionAttestationProtocol.SIM)
+#else
                     if (attestationProtocol != SqlConnectionAttestationProtocol.AAS)
+#endif
                     {
                         return false;
                     }
                     break;
 
+#if ENCLAVE_SIMULATOR
+                case TdsEnums.ENCLAVE_TYPE_SIMULATOR:
+                    if (attestationProtocol != SqlConnectionAttestationProtocol.SIM)
+                    {
+                        return false;
+                    }
+                    break;
+#endif
                 default:
                     // if we reach here, the enclave type is not supported
                     throw SQL.EnclaveTypeNotSupported(enclaveType);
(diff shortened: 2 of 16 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -3017,23 +3017,41 @@ private bool TryProcessFeatureExtAck(TdsParserStateObject stateObj)
...
         private bool IsValidAttestationProtocol(SqlConnectionAttestationProtocol attestationProtocol, string enclaveType)
         {
-            switch (enclaveType)
+            switch (enclaveType.ToUpper())
             {
                 case TdsEnums.ENCLAVE_TYPE_VBS:
                     if (attestationProtocol != SqlConnectionAttestationProtocol.AAS
+#if ENCLAVE_SIMULATOR
+                        && attestationProtocol != SqlConnectionAttestationProtocol.HGS
+                        && attestationProtocol != SqlConnectionAttestationProtocol.SIM)
```
David-Engel: Nit: If you put SIM above HGS, you can eliminate a line of code.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -3485,23 +3485,41 @@ private bool TryProcessFeatureExtAck(TdsParserStateObject stateObj)
...
         private bool IsValidAttestationProtocol(SqlConnectionAttestationProtocol attestationProtocol, string enclaveType)
         {
-            switch (enclaveType)
+            switch (enclaveType.ToUpper())
             {
                 case TdsEnums.ENCLAVE_TYPE_VBS:
                     if (attestationProtocol != SqlConnectionAttestationProtocol.AAS
+#if ENCLAVE_SIMULATOR
+                        && attestationProtocol != SqlConnectionAttestationProtocol.HGS
+                        && attestationProtocol != SqlConnectionAttestationProtocol.SIM)
```
David-Engel: Nit: If you put SIM above HGS, you can eliminate a line of code.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
karinazhou: Good point. I will move SIM a line up.

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.NetCoreApp.cs
```diff
@@ -113,6 +113,11 @@ public enum SqlConnectionAttestationProtocol
         /// <include file='..\..\..\..\doc\snippets\Microsoft.Data.SqlClient\SqlConnectionAttestationProtocol.xml' path='docs/members[@name="SqlConnectionAttestationProtocol"]/AAS/*' />
         AAS = 1,
 
+#if ENCLAVE_SIMULATOR
+        /// <include file='..\..\..\..\doc\snippets\Microsoft.Data.SqlClient\SqlConnectionAttestationProtocol.xml' path='docs/members[@name="SqlConnectionAttestationProtocol"]/SIM/*' />
+        SIM = 2,
```
cheenamalhotra: Is it okay to reserve enum value 2 for SIM?
Does this enum correspond to SQL Server's response as well?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -3017,23 +3017,41 @@ private bool TryProcessFeatureExtAck(TdsParserStateObject stateObj)
...
+#endif
                     {
                         return false;
                     }
                     break;
 
                 case TdsEnums.ENCLAVE_TYPE_SGX:
+#if ENCLAVE_SIMULATOR
+                    if (attestationProtocol != SqlConnectionAttestationProtocol.AAS
+                        && attestationProtocol != SqlConnectionAttestationProtocol.SIM)
```
cheenamalhotra: I guess same can be done here, as David commented above.

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.NetCoreApp.cs
karinazhou: This protocol number is defined in SimulatorEnclaveProviders.cs. It is used to create the SqlEnclaveAttestationParameters. I think we need this number to communicate with SQL Server to get a response.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
karinazhou: This one is a bit different. If I remove the `#else` branch, I will need to put `)` in a separate line which is a bit weird.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
karinazhou: Changed.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
cheenamalhotra: Fair enough!

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.NetCoreApp.cs
cheenamalhotra: My doubt was is it us who own 1 and 3 enum values for supported types (would also answer why 1 and 3 are for AAS/VBS), and if assigning 2 for SIM is also driver specific only?

Does server not care if we assign it 2 or 5 or any other value?

karinazhou: I don't think we can randomly pick other numbers as these protocolIDs (AAS / HGS). The protocolID is part of the attestationParameters which the driver needs to be sent to the server through sp_describe_parameter_encryption RPC call. It should be the same numbers across all the drivers targeting the same SQL Server.

cheenamalhotra: Resolving based on Offline discussion.

PR #443, opened 2020-02-24
Fix access token behavior in connection pool

### LINKED ISSUE #438: Connection pool doesn't work well with access tokens
### Describe the bug
The connection pool doesn't work well with explicit access tokens because SqlConnectionPoolKey uses reference equality to compare access tokens. It should use string.Equals instead.

The error is in line https://github.com/dotnet/SqlClient/blob/654933a0811ab30b843a20aff051b8d585238505/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnectionPoolKey.cs#L113

### To reproduce
Create two strings at1 and at2 which contain the same valid access token.
Open a connection with pooling enabled (which is the default); set SqlConnection.AccessToken to at1 and run a query. Close the connection. Open a second connection with pooling enabled; set SqlConnection.AccessToken to at2 and run a second query. Check the number of open connections on the server; it will be 2.

Repeat the experiment except using at1 both times. There will only be a single open connection on the server.

### Expected behavior
Pooled connections using an explicit access token should be shared even if different string objects containing the same value were used to initialize the connection.

### Further technical details
N/A

cheenamalhotra:
- Fixes #438 + tests added
- Modifies tests to dynamically generate Access Token for testing instead of specifying it in `config.json` file.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnectionPoolKey.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnectionPoolKey.cs
@@ -65,7 +65,7 @@ namespace Microsoft.Data.SqlClient
             SqlConnectionPoolKey key = obj as SqlConnectionPoolKey;
             return (key != null
                 && _credential == key._credential
-                && string.CompareOrdinal(ConnectionString, key.ConnectionString) == 0
+                && ConnectionString == key.ConnectionString
                 && string.CompareOrdinal(_accessToken, key._accessToken) == 0);
         }
 
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnectionPoolKey.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnectionPoolKey.cs
@@ -63,7 +63,7 @@ namespace Microsoft.Data.SqlClient
         public override bool Equals(object obj)
         {
             SqlConnectionPoolKey key = obj as SqlConnectionPoolKey;
-            return (key != null && _credential == key._credential && ConnectionString == key.ConnectionString && Object.ReferenceEquals(_accessToken, key._accessToken));
+            return (key != null && _credential == key._credential && ConnectionString == key.ConnectionString && _accessToken == key._accessToken);
         }
 
         public override int GetHashCode()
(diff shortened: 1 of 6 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnectionPoolKey.cs
```diff
@@ -63,7 +63,7 @@ internal string AccessToken
         public override bool Equals(object obj)
         {
             SqlConnectionPoolKey key = obj as SqlConnectionPoolKey;
-            return (key != null && _credential == key._credential && ConnectionString == key.ConnectionString && Object.ReferenceEquals(_accessToken, key._accessToken));
+            return (key != null && _credential == key._credential && ConnectionString == key.ConnectionString && _accessToken == key._accessToken);
```
saurabh500: Can you do a ordinal comparison here for AccessToken ?

cheenamalhotra: Sure, I just kept it similar to other comparisons.
Should I change that for Access Token only or for other properties too?

saurabh500: Access Token only.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnectionPoolKey.cs
```diff
@@ -63,7 +63,10 @@ internal string AccessToken
         public override bool Equals(object obj)
         {
             SqlConnectionPoolKey key = obj as SqlConnectionPoolKey;
-            return (key != null && _credential == key._credential && ConnectionString == key.ConnectionString && _accessToken == key._accessToken);
+            return (key != null
+                && _credential == key._credential
+                && string.CompareOrdinal(ConnectionString, key.ConnectionString) == 0
```
saurabh500: Dont change the check for ConnectionString. Only Access Token

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnectionPoolKey.cs
```diff
@@ -63,7 +63,10 @@ internal string AccessToken
         public override bool Equals(object obj)
         {
             SqlConnectionPoolKey key = obj as SqlConnectionPoolKey;
-            return (key != null && _credential == key._credential && ConnectionString == key.ConnectionString && Object.ReferenceEquals(_accessToken, key._accessToken));
+            return (key != null
+                && _credential == key._credential
+                && ConnectionString == key.ConnectionString
+                && string.CompareOrdinal(_accessToken, key._accessToken) == 0);
```
saurabh500: I forgot to ask, what about NetFx ?

saurabh500: @cheenamalhotra

saurabh500: Never mind. I see it

PR #453, opened 2020-03-03
Fixed the ConnectionString's password leak in NetCore

### LINKED ISSUE #448: Persist Security Info=false isn't respected
With Microsoft.Data.SqlClient on netcoreapp3.1, the password is present in SqlConnection.ConnectionString after Open has been called. [As per the docs](https://docs.microsoft.com/en-us/dotnet/api/system.data.sqlclient.sqlconnection.connectionstring?view=netframework-4.8#remarks), once Open is called, all security info is supposed to be stripped from the user-exposed connection string. System.Data.SqlClient behaves correctly. This can be seen with MS.Data.SqlClient versions 1.1.1 and 2.0.0-preview1.20021.1.

```c#
using System;
//using System.Data.SqlClient;

namespace Test
{
    class Program
    {
        {
            var conn = new SqlConnection(@"Server=localhost;Database=test;User=SA;Password=Abcd5678;Persist Security Info=false");
            conn.Open();
            Console.WriteLine("Connection string: " + conn.ConnectionString);
            // With Microsoft.Data.SqlClient, outputs Server=localhost;Database=test;User=SA;Password=Abcd5678;Persist Security Info=false
            // With System.Data.SqlClient, outputs Server=localhost;Database=test;User=SA;Persist Security Info=false
        }
    }
}
```

DavoudEshtehari:
Fixes #448

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Common/src/Microsoft/Data/Common/DbConnectionOptions.Common.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Common/src/Microsoft/Data/Common/DbConnectionOptions.Common.cs
@@ -101,7 +101,6 @@ namespace Microsoft.Data.Common
         private readonly string _usersConnectionString;
         private readonly Dictionary<string, string> _parsetable;
         internal readonly NameValuePair _keyChain;
-        internal readonly bool _hasPasswordKeyword;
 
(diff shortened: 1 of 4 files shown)


DavoudEshtehari:
If PersistSecurityInfo = false be set in a ConnectionString, we should not get the password in an open sqlConnection's ConnectionString property.

PR #454, opened 2020-03-04
Multiple source code improvements

cheenamalhotra:
This PR:
- Update NetFx Project to latest .NET SDK
- Moves common files to shared folder and projects reference from common props. (Useful in longer run to avoid double effort on common file edits)


Wraith2:
Looks good to me.

Issue #29, opened 2016-12-20
SqlClient's connection resiliency adds 10 seconds to database existence checks. Consider exposing a way to override it at runtime


divega:
This is to follow up on a specific issue a customer created today on EF Core at https://github.com/aspnet/EntityFramework/issues/7283, but the problem has been brought to our attention before.  

Both EF6 and EF Core provide APIs that allow applications to create and initialize databases, check for their existence and modify their schemas. These operations can happen at design time (e.g. when using the migrations feature) or at runtime. **Attempting to connect to a database that doesn't exist is part of the normal flow of these operations.** 

In .NET Framework 4.5.1 SqlClient introduced a connection resiliency feature that performs automatic retries if failures occur during `SqlConnection.Open()`. Two new settings `ConnectRetryCount` and `ConnectRetryInterval` were added to the list of settings recognized in connection strings and a default behavior was adopted so if the first attempt to connect to a database fails a retry will occur after 10 seconds if these settings are not specified. The same feature is now included in the .NET Core version of SqlClient.

With only the default behavior **this feature introduces a lag of 10 second for any calling code that attempts to connect to a database that doesn't exist.** It also prevents any calling code from implementing its own (potentially more efficient) retry logic correctly. This severely affects customer experience and can potentially affect runtime performance. 

Two main approaches have been proposed to mitigate this issue on EF code, but they have severe disadvantages:

1. Recommend customers to disable SqlClient's connection resiliency altogether by setting `ConnectRetryCount=0` when working EF and make sure the feature is disabled any time EF runtime or tooling creates a connection string, or even throw if an attempt is made to use a connection that has the feature enabled. The main disadvantages of this approach are:
 * It is not discoverable for customers and so users will still experience the 10 second lags (or exceptions if we decide to throw) for the default case.
 * It prevents the connection resiliency feature in SqlClient from being available in scenarios in which it could actually have been helpful.
 * EF6 supports older versions of .NET Framework than 4.5.1 and for those versions these settings are not valid. The logic would need to take the executing version of .NET Framework into account.

2. Store aside a copy of the original connection string so that we can modify it with `ConnectRetryCount = 0` and create a separate `SqlConnection` object to perform existence checks. 
 * The main problem with this approach is that both EF Core and EF6 support passing a `SqlConnection` object to be used in the EF context. It is possible that the password would have already been removed from the `ConnectionString` in that connection object if it was open before, so in that case the connection string would not contain enough credentials to be able to create a separate functional connection object.

[...]

PR #463, opened 2020-03-09
Connection Open with fast fail option

David-Engel:
Putting this out there for consideration related to #29 

Summary:
Implementing a programmatic way of overriding the behavior of SqlConnection.Open() to disable automatic connection retries in the event of transient errors. The transient error handling introduces a 10 second delay when used with default values. If an application is passing a connection object around and just wants to do a quick Open() availability check, they can pass SqlConnectionOverrides to Open() and disable transient fault handling logic.

I tried to apply the overrides to OpenAsync(), too. However OpenAsync() already seems to not apply transient fault handling and fails quickly and I was unable to change its behavior in a reasonable amount of time. Since the request was really only for Open(), I opted to not add the override option to OpenAsync(), too.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
@@ -591,12 +591,22 @@ namespace Microsoft.Data.SqlClient
         public override System.Data.DataTable GetSchema(string collectionName, string[] restrictionValues) { throw null; }
         /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnection.xml' path='docs/members[@name="SqlConnection"]/Open/*'/>
         public override void Open() { }
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnection.xml' path='docs/members[@name="SqlConnection"]/OpenWithOverrides/*'/>
+        public void Open(SqlConnectionOverrides overrides) { }
         /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnection.xml' path='docs/members[@name="SqlConnection"]/OpenAsync/*'/>
         public override System.Threading.Tasks.Task OpenAsync(System.Threading.CancellationToken cancellationToken) { throw null; }
         /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnection.xml' path='docs/members[@name="SqlConnection"]/ResetStatistics/*'/>
         public void ResetStatistics() { }
         /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnection.xml' path='docs/members[@name="SqlConnection"]/RetrieveStatistics/*'/>
         public System.Collections.IDictionary RetrieveStatistics() { throw null; }
+    }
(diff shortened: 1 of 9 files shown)


David-Engel:
@saurabh500 - Thoughts on creating a new Open method for this?

For background, EF would like a programmatic way to fast-fail connection attempts (without modifying the connection string) in cases when they are just checking for the existence of a database (before creating one, for example). There does not seem to be a way to check for database existence without a connection. When the driver applies transient fault handling to connection attempts, it considers error 4060 (Cannot open database "%.*ls" requested by the login. The login failed.) a transient error since it could be [an Azure connection reconfiguring the back end](https://docs.microsoft.com/en-us/azure/sql-database/troubleshoot-connectivity-issues-microsoft-azure-sql-database#transient-faults-connection-loss-and-other-temporary-errors). Applying transient error handling adds a 10 second delay while the driver retries the connection when using the default connection properties. This results in a sub-par user experience while the user waits for SqlClient to fail.

cheenamalhotra:
1 option to consider is connection class APIs, keeping `SqlConnection.Open()` 'bool' parameter free for future standardization if needed at all.. since we inherit `DbConnection`.

We can design the new API such that it is understood as a temp flag, which when set only applies within it's scope. and does not reflect change in connection properties. Likewise we won't have to worry about pooling and connection property overriding.

e.g. `SqlConnection.PauseConnectionResiliency()` and `SqlConnection.ResumeConnectionResiliency()`

saurabh500:
@David-Engel @cheenamalhotra  I went through the issue and it looks like we definitely need a programmatic way of doing this, since the connection string may not be available if the SqlConnection Object is being passed around. I hadn't realized from our discussions on the call that SqlConnection instance is being passed around. In that case reconstructing the connection string from the connection object may not be possible. Let's discuss in the call tomorrow.

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -589,13 +589,25 @@ public sealed partial class SqlConnection : System.Data.Common.DbConnection, Sys
...
         public override System.Threading.Tasks.Task OpenAsync(System.Threading.CancellationToken cancellationToken) { throw null; }
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnection.xml' path='docs/members[@name="SqlConnection"]/OpenAsyncWithOverrides/*'/>
+        public System.Threading.Tasks.Task OpenAsync(System.Threading.CancellationToken cancellationToken, SqlConnectionOverrides overrides) { throw null; }
         /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnection.xml' path='docs/members[@name="SqlConnection"]/ResetStatistics/*'/>
         public void ResetStatistics() { }
         /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnection.xml' path='docs/members[@name="SqlConnection"]/RetrieveStatistics/*'/>
         public System.Collections.IDictionary RetrieveStatistics() { throw null; }
     }
+    /// <include file='..\..\..\..\doc\snippets\Microsoft.Data.SqlClient\SqlConnectionOverrides.xml' path='docs/members[@name="SqlConnectionOverrides"]/SqlConnectionOverrides/*' />
```
cheenamalhotra, line 603: Please change to Forward slashes `/` universally.

Issue #408, opened 2020-02-04
Support parallel Physical Connection creation for multi-threaded scenarios


joshmouch:
I'm trying to spin up 50 threads to immediately open 50 `SqlConnection`s to a Microsoft SQL Server connection pool.  However, no matter what I try, the `SqlConnection`s seem to only open one at a time.  In other words, if each connection takes 2 seconds to open, then the 50'th thread's `Open` method will take 100 seconds to complete.  Further to this point, if the Connection Timeout isn't set to > 100 seconds then the connection attempts will eventually start timing out.  With the default Connection Timeout = 30, the 16'th thread's `Open` times out.

To put this another way, it seems like the 1st `SqlConnection.Open` takes 2 seconds, the 2nd `SqlConnection.Open` takes 4 seconds, the 3rd `SqlConnection.Open` takes 6 seconds... the 16th `SqlConnection.Open` takes 32 seconds and causes a timeout exception.

I also tried setting Min Pool Size=50 and just opening a single SqlConnection, and that has the same behavior.  There seems to be a background process that starts opening connections until it gets to 50, but the 16th one times out because it's over the 30 second connection timeout.

It's like there's something blocking internally to only allow this one-at-a-time addition to the connection pool.
I've asked around on stack overflow, and the responses seem to suggest that if you open 50 connections in parallel, they should all open.

So this makes me wonder if this is a .Net Core bug or "feature".

cheenamalhotra:
@joshmouch 

I'm looking into the issue, could you also test and confirm if this issue is reproducible with previous System.Data.SqlClient versions or it's only happening recently; to rule out the factor of being a regression?

Wraith2:
When you say you've tried it with and without threads do you mean that you manually created 100 threads and used them to connect? or that you used the threadpool? With direct threadpool or tasks (indirectly threadpool assuming `ConfigureAwait(false)`) you'll hit the rate limiter on thread creation through the pool of roughly 1 per second which would mean you'd see blocking and slow creation of threads to service the queued requests, even with 100 connection objects if all the threads are busy (assuming you don't connect and go idle) you can't do more than the pool thread count in parallel.

joshmouch:
What I mean by "without threads" is that if I simply set "Min Pool Size=50", then do this:
```
var connection1 = new SqlConnection(sqlConnectionString);
connection1.Open();
connection1.Close();
connection1.Dispose();
System.Threading.Thread.Sleep(60000);
```

[...]

PR #466, opened 2020-03-11
Fix | Fixes concurrent connection speed issues due to 'Authentication Context' cache not maintained in NetCore

David-Engel:
The authentication context cache of the connection pool was not being updated upon successful fedauth authentication in the netcore code. This was resulting in poor connection speed for subsequent connections which used AAD auth since they could not take advantage of the cached token. This code looks to have been missed when AAD authentication was ported from netfx to netcore.

This was found during investigation of #408 

The new test should detect if subsequent connections are not using the cache. Hopefully connection times are consistent enough that this does not produce intermittent failures.


Wraith2:
It might be worth using private reflection to reach into the internals and find the cache to verify that it has been populated rather than relying on a timing artefact that could be flaky. I know private reflection is generally bad but as the owners you're allowed to assume your own internal details in this case I'd say.

David-Engel:
> It might be worth using private reflection to reach into the internals and find the cache to verify that it has been populated rather than relying on a timing artefact that could be flaky. I know private reflection is generally bad but as the owners you're allowed to assume your own internal details in this case I'd say.

@Wraith2 
I expect _dbConnectionPool.AuthenticationContexts.Count > 0 in SqlInternalConnectionTds after the first connection. I spent some time trying to wind my way through to view that via Reflection, but it was not apparent to me how to get to it. I'll have to revisit it later if I find the time. (If you have some time to spare, you're probably quicker than I am. :grin:)

PR #473, opened 2020-03-19
Remove sync over async in named pipe EnableSsl

Wraith2:
This is a simple one line change that brings named pipe handle into line with the tcp version by removing a sync-over-async call in favour of using a sync call directly.

I'm not sure why there are 3 commits in this PR. I've rebased my master with no changes to the upstream master. I can't work out what's happened or why I have extra commits. I don't know if they're harmful but since I can't really see them I can't get rid of them. I don't really want to destroy my entire working directly and download cleanly because I can't be sure it'll work and i'll lose two complex unmerged branches (packet rework and batch api)


cheenamalhotra:
Hi @Wraith2

Could you try cherry-picking your commit to another branch created on dotnet/master (with latest commits fetched from upstream)

PR #477, opened 2020-03-20
Enable Managed SNI on Windows with AppContext switch

cheenamalhotra:
Enabling preview of Managed SNI on Windows for testing and debugging purposes.

With this change, users will be able to specify the below `AppContext` switch to toggle driver's behavior to use Managed SNI in .NET Core 2.1+ and .NET Standard 2.0+ projects on Windows.

**Microsoft.Data.SqlClient.UseManagedNetworkingOnWindows**

```cs
AppContext.SetSwitch("Microsoft.Data.SqlClient.UseManagedNetworkingOnWindows", true);
```
**[NOTE]** Known differences when compared to Native SNI.dll:
- Managed SNI does not support non-domain Windows Authentication.

### DIFF
--- a/BUILDGUIDE.md
+++ b/BUILDGUIDE.md
@@ -175,6 +175,6 @@ Tests can be built and run with custom Target Frameworks. See the below examples
 
 ## Using Managed SNI on Windows
 
-Managed SNI can be enabled on Windows by enabling below AppContext switch:
+Managed SNI can be enabled on Windows by enabling the below AppContext switch:
 
 **"Microsoft.Data.SqlClient.UseManagedNetworkingOnWindows"**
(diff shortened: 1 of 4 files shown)


Wraith2:
The environment variable option is useful both in debugging and some deployment scenarios like azure where recompiling may be difficult but the environment is mutable. Any change of keeping that option around?

cheenamalhotra:
@Wraith2 you can use Environment variable and use that in your app startup to Set App Context switch like I did in DataTestUtility class.

On discussion with Saurabh, we concluded 1 enabling option only with App Context switch as that's .NETs standardized approach to toggle optional/preview behaviours.

We shall be using more of these for future optional/preview features or toggle options for backwards compatibility.

ErikEJ:
If you are going to do that, it would be very useful to have them documented in a central location.

Wraith2:
Some random thoughts. 

- there should probably be a trace event for this so you can see what mode was chosen
- it needs to be noted in the docs that the appcontext switch has to be set as early as possible and certainly before any queries are executed. This is because it's a static property and it's execution will be caused by whatever touches the class first which is dependent on the dependency graph inside the library. Making it Lazy would probably delay it a little but not much.

PR #479, opened 2020-03-22
perf: Reduce tracing related allocations

Wraith2:
Now that tracing has been added to netcore I did a quick profile over the usual DataAccessPerformance benchmark app and investigated a few new hot allocation sources.

- ActivityCorrelator was creating a new copy of the current activity on every call. I've made the class effectively immutable so the current instance can be returned every time without danger and change the Next() implementation to correctly generate the next in sequence as before.
- TraceString() was being called unconditionally to pass into a method which wasn't going to use it. It's expensive so I added the check before the call, this means there are now two check but that should take less time than required for all the information generation and formatting.
- TraceString was passing a lot of bools in a params object array which causes them to be boxed, I simply did a ternary check and passed in the preallocated string version instead.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/Common/ActivityCorrelator.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/Common/ActivityCorrelator.cs
@@ -14,30 +14,15 @@ namespace Microsoft.Data.Common
 
     internal static class ActivityCorrelator
     {
-        internal class ActivityId
+        internal sealed class ActivityId
         {
-            internal Guid Id { get; private set; }
-            internal uint Sequence { get; private set; }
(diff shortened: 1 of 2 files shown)


cheenamalhotra:
Hi @Wraith2 

Could you please update the same in NetFx project? We don't want to break consistency between the two if possible.

Wraith2:
> Could you please update the same in NetFx project? We don't want to break consistency between the two if possible.

I don't usually port my changes to netfx for a few reasons. 
- I don't want to try and do everything twice and there are a lot of differences between the two codebases. 
- netfx is legacy at this point with netcore being net5 this year. 
- my hope/expectation was that the codebases would begin to merge so that any perf improvements were picked up by everyone equally. I expected the netcore codebase to be the better of the two most of the time for this effort.

PR #484, opened 2020-03-23
Standardization of Event Source tracing + Improvements for readability

cheenamalhotra:
Updates in EventSource tracing implementation:
- Re-ordered Keywords
- Code Organization for readability
- Standardization of method names
- Inline documentations
- SNI Trace + Scope segregation
- Event proofing
- Adds support in ManualTests to enable tracing for all tests from pipeline.
- Modifies enabling "UseManagedSNIOnWindows" from config.json instead of Env variable.

Also to mention:
* This change also makes EventSource traces all disabled by default.
End users would have to enable "_ExecutionTrace_" to continue to capture Begin/End execute commands.
* Keeping this trace enabled by default does not seem appropriate when comparing with other EventSource implementations. As we head to 2.0 (major), we are making amendments to set things right.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlClientEventSource.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlClientEventSource.cs
@@ -428,7 +428,7 @@ namespace Microsoft.Data.SqlClient
 
         #region Notification Scope
         [NonEvent]
-        internal long notificationScopeEnterEvent<T0>(string message, T0 args0)
+        internal long NotificationScopeEnterEvent<T0>(string message, T0 args0)
         {
             if (Log.IsNotificationScopeEnabled())
             {
@@ -438,7 +438,7 @@ namespace Microsoft.Data.SqlClient
         }
 
         [NonEvent]
-        internal long notificationScopeEnterEvent<T0, T1>(string message, T0 args0, T1 args1)
+        internal long NotificationScopeEnterEvent<T0, T1>(string message, T0 args0, T1 args1)
         {
             if (Log.IsNotificationScopeEnabled())
             {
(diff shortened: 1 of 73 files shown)


src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
```diff
@@ -558,8 +558,8 @@ internal bool TryInitialize(TdsParserStateObject stateObj, int columnsCount)
                 {
                     return false;
                 }
-                SqlClientEventSource.Log.TraceEvent("<sc.TdsParserStateObject.NullBitmap.Initialize|INFO|ADV> {0}#, NBCROW bitmap received, column count = {1}", stateObj.ObjectID, columnsCount);
```
karinazhou: From the initial code, I assume that TraceEvent is different from AdvanceTrace. Is there any reason we the new version of AdvancedTraceEvent here?

cheenamalhotra: Looking at "ADV" in log, I would expect this is advance trace log.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlClientEventSource.cs
```diff
@@ -27,440 +29,616 @@ internal class SqlClientEventSource : EventSource
         /// Defines EventId for EndExecute (Reader, Scalar, NonQuery, XmlReader).
         /// </summary>
         private const int EndExecuteEventId = 2;
+
+        /// <summary>
```
Wraith2: These comments really don't add any value. I know they're enforced by the build. Is there a way we can configure the rule so that only public methods are required to have doc comments?

JRahnama: I agree with Wraith, since the class itself is internal, none of the keywords and their API documentation would show on IntelliSense. those could only be referenced by numbers only.

cheenamalhotra: It's not for public, it's for developers and maintainers.
Inline documentation adds a lot of value for teams where developers are not always the same!

cheenamalhotra:
Enabling all traces seems to capture issues in Managed SNI tracing implementation from PR #461, to be continued tomorrow!

PR #490, opened 2020-03-25
Address MARS TDS header contained errors

David-Engel:
Trying to address #85 

Here are my findings:

It seems that TryProcessDone in fact was running twice in certain cases. Once in the context of the first ReadAsync call, and again in the context of a continuation task when Cancel was called on the SqlCommand. I'm guessing Cancel happened really fast in these cases because both threads appeared to be executing nearly in parallel, hitting `stateObj.DecrementOpenResultCount();` close enough together that the count incorrectly went to -1 (which gets passed in the MARS header and is invalid).

This underlying problem was exposed by https://github.com/dotnet/corefx/pull/26595.

I think this is a valid fix without reverting the above PR. It fixes the repro case and makes sense from what I've been able to digest from working with the async code (for what that's worth).

Note: The netfx code has the same issue, but it is hidden by the fact that `LocalAppContextSwitches.MakeReadAsyncBlocking` is true by default which sets `stateObj._syncOverAsync = true;`. I did not change the behavior of netfx as part of this change but I did make the same fix to the netfx code as is applied to the netcore code.


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -2938,12 +2938,12 @@ private bool TryProcessDone(SqlCommand cmd, SqlDataReader reader, ref RunBehavio
             // _pendingData set by e.g. 'TdsExecuteSQLBatch'
             // _hasOpenResult always set to true by 'WriteMarsHeader'
             //
-            if (!stateObj.HasPendingData && stateObj.HasOpenResult)
+            if (!stateObj._attentionSent && !stateObj.HasPendingData && stateObj.HasOpenResult)
```
Wraith2: If attentionSent is set within the parser lock this seems reasonable. It might be worth amending the comment above this with a mention of attentionSent.

David-Engel: Done.

David-Engel:
This is a candidate to back port to 1.1 but I'd like to run the EF tests against it and get some people testing it to be more confident of the fix before putting it into a GA hotfix.

ErikEJ:
I can test the 2.0 version against EF Core - where can I get a daily build?

cheenamalhotra:
Hi @ErikEJ 

Since the PR hasn't been merged yet, nightly builds don't contain this change. You may use the Nuget package I attached under to validate changes:

[Microsoft.Data.SqlClient.2.0.0-dev-pr490.nupkg.zip](https://github.com/dotnet/SqlClient/files/4389471/Microsoft.Data.SqlClient.2.0.0-dev-pr490.nupkg.zip)

```xml
<PackageReference Include="Microsoft.Data.SqlClient" Version="2.0.0-dev-pr490" />
```

ErikEJ:
Maybe you should consider automatically running the 18.000 EF Core SQL Server functional tests against any PR?

ErikEJ:
I tested against latest EF Core master, and all tests passed (single run)

cheenamalhotra:
> I tested against latest EF Core master, and all tests passed (single run)

Hi @ErikEJ 

The issue was reproducible with their "release/3.0" tag only and not in master as I mentioned here:
https://github.com/dotnet/SqlClient/issues/85#issuecomment-570703000 Could you also try that?

> Maybe you should consider automatically running the 18.000 EF Core SQL Server functional tests against any PR?

Yes that is in our backlog of things. Will be done soon!

ErikEJ:
@cheenamalhotra I was unable to build that tag.

cheenamalhotra:
@ErikEJ I too faced issues with running tests with NuGet package, tested with direct project reference and everything looks good!

PR #492, opened 2020-03-26
Allow username with Active Directory Interactive Authentication (NetFx)

cheenamalhotra:
This PR adds support for specifying "User ID" or "UID" (optionally) with "Active Directory Interactive" authentication mode in Connection String.

* It does not break existing behavior of Microsoft.Data.SqlClient.
* Also makes the driver backwards compatible with System.Data.SqlClient.

Dev NuGet Package for testing:
[Microsoft.Data.SqlClient.2.0.0-dev.nupkg.zip](https://github.com/dotnet/SqlClient/files/4389268/Microsoft.Data.SqlClient.2.0.0-dev.nupkg.zip)

Connection Strings acceptable:
"Server={servername}; Authentication=Active Directory Interactive;"
"Server={servername}; Authentication=Active Directory Interactive; User Id={username};"

### DIFF
--- a/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/ActiveDirectoryNativeAuthenticationProvider.cs
+++ b/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/ActiveDirectoryNativeAuthenticationProvider.cs
@@ -52,6 +52,7 @@ namespace Microsoft.Data.SqlClient
             {
                 result = await app.AcquireTokenInteractive(scopes)
                   .WithUseEmbeddedWebView(true)
+                  .WithLoginHint(parameters.UserId)
                   .ExecuteAsync();
             }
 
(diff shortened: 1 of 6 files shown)


src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/ActiveDirectoryNativeAuthenticationProvider.cs
```diff
@@ -52,6 +52,7 @@ internal class ActiveDirectoryNativeAuthenticationProvider : SqlAuthenticationPr
             {
                 result = await app.AcquireTokenInteractive(scopes)
                   .WithUseEmbeddedWebView(true)
+                  .WithLoginHint(parameters.UserId)
```
shueybubbles: WithLoginHint handles an empty userid ok?

cheenamalhotra: Yes, it handles null, empty string values, prompts for login in those cases.

karinazhou: Do we need this change in netcore as well?

cheenamalhotra: We don't support Interactive in .NET Core yet.

PR #495, opened 2020-03-28
Attempt to fix Promote Transaction issue

### LINKED ISSUE #237: SqlException: The PROMOTE TRANSACTION request failed because there is no local transaction active
Under certain circumstances, when a combination of SqlCommands and transactional WCF calls is being run in a massively parallel manner, the current TransactionScope is occasionally "left over", SqlCommands do not get attached to it and therefore their updates are successfully committed, while the TransactionScope is rolled back. This leads to **severe data corruption** in a Line-of-Business application of one of our corporate customers.

Stable repro is outlined in this test project: https://github.com/scale-tone/InconsistencyDemo. See more detailed instructions on how to run it there. To reveal the issue more quickly, start and stop *LoopHammer* console app several times.

The [LoopHammer console app](https://github.com/scale-tone/InconsistencyDemo/blob/master/src/DemoService/LoopHammer/Program.cs) emulates a message processing, consisting of the following steps:
* Creating a TransactionScope.
* As part of that TransactionScope, executing a couple of SQL statements against a local DB.
* As part of that TransactionScope, making a transactional WCF call, which fails with an exception. The call forces the transaction to be promoted to a distributed one. The exception causes the entire TransactionScope to be rolled back (including the above SQL statements).

This algorithm is executed multiple times, in parallel (just in the way the real message processing service operates). Note that the repro code only creates one TransactionScope and doesn't explicitly mix it with SqlTransactions or any other nested transactions.

### Expected behavior
The distributed transaction is rolled back entirely, nothing is written to DB.

### Actual behavior

After seconds to minutes of normal execution the WCF call fails with the following (not expected!) exception:

> System.ServiceModel.ProtocolException: The transaction has aborted. ---> System.Transactions.TransactionAbortedException: The transaction has aborted. ---> System.Transactions.TransactionPromotionException: Failure while attempting to promote transaction. ---> Microsoft.Data.SqlClient.SqlException: The PROMOTE TRANSACTION request failed because there is no local transaction active.
   at Microsoft.Data.SqlClient.TdsParser.Run(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj)
   at Microsoft.Data.SqlClient.TdsParser.TdsExecuteTransactionManagerRequest(Byte[] buffer, TransactionManagerRequestType request, String transactionName, TransactionManagerIsolationLevel isoLevel, Int32 timeout, SqlInternalTransaction transaction, TdsParserStateObject stateObj, Boolean isDelegateControlRequest)
   at Microsoft.Data.SqlClient.SqlInternalConnectionTds.ExecuteTransactionYukon(TransactionRequest transactionRequest, String transactionName, IsolationLevel iso, SqlInternalTransaction internalTransaction, Boolean isDelegateControlRequest)
   at Microsoft.Data.SqlClient.SqlDelegatedTransaction.Promote()
   at Microsoft.Data.SqlClient.SqlDelegatedTransaction.Promote()
   at System.Transactions.TransactionStatePSPEOperation.PSPEPromote(InternalTransaction tx)
   at System.Transactions.TransactionStateDelegatedBase.EnterState(InternalTransaction tx)
   at System.Transactions.TransactionStateAborted.CheckForFinishedTransaction(InternalTransaction tx)
   at System.Transactions.Transaction.Promote()
   at System.Transactions.TransactionInterop.ConvertToOletxTransaction(Transaction transaction)
   at System.Transactions.TransactionInterop.GetTransmitterPropagationToken(Transaction transaction)
   at System.ServiceModel.Transactions.WsatTransactionFormatter.WriteTransaction(Transaction transaction, Message message)
   at System.ServiceModel.Channels.TransactionChannel`1.WriteTransactionToMessage(Message message, TransactionFlowOption txFlowOption)
Server stack trace:
   at System.ServiceModel.Channels.TransactionChannel`1.WriteTransactionToMessage(Message message, TransactionFlowOption txFlowOption)
   at System.ServiceModel.Channels.TransactionChannel`1.WriteTransactionDataToMessage(Message message, MessageDirection direction)
   at System.ServiceModel.Channels.TransactionRequestChannelGeneric`1.Request(Message message, TimeSpan timeout)
   at System.ServiceModel.Dispatcher.RequestChannelBinder.Request(Message message, TimeSpan timeout)
   at System.ServiceModel.Channels.ServiceChannel.Call(String action, Boolean oneway, ProxyOperationRuntime operation, Object[] ins, Object[] outs, TimeSpan timeout)
   at System.ServiceModel.Channels.ServiceChannelProxy.InvokeService(IMethodCallMessage methodCall, ProxyOperationRuntime operation)
   at System.ServiceModel.Channels.ServiceChannelProxy.Invoke(IMessage message)
Exception rethrown at [0]:
   at System.Runtime.Remoting.Proxies.RealProxy.HandleReturnMessage(IMessage reqMsg, IMessage retMsg)
   at System.Runtime.Remoting.Proxies.RealProxy.PrivateInvoke(MessageData& msgData, Int32 type)
   at DemoService.WcfAgents.IService1.DoWork(String correlationId)
   at DemoService.WcfAgents.Service1Client.DoWork(String correlationId)
   at LoopHammer.DostuffHandler.MakeTransactionalWcfCallThatThrows()
   at LoopHammer.DostuffHandler.Handle()

This happens presumably because at the moment of promotion attempt, the TransactionScope (not itself, but some underlying internal transaction object) appears to be in an aborted state already. Because it appears to be in that state, the above SQL statements do not get attached to it and got committed immediately. So the TransactionPromotionException is just a visible part of the problem, while the actual issue occurs somewhere in between TransactionScope creation and execution of SqlCommands. We're able to detect that faulty state of TransactionScope beforehand, by executing "select @@trancount" against the DB - in a failure case it appears to be 0 (normally should be 1).

WORKAROUND: **if we disable SqlConnection pooling (add "Pooling=false" to connection string), the problem disappears**. This makes us thinking, that the bug is somewhere in connection sharing mechanism. Probably, a connection is sometimes not fully detached from one thread, and occasionally allows it to influence a TransactionScope belonging to another thread. If that is the case, then it can not only lead to data corruption but also introduce **serious security issues**. 

The problem is reproducible with the latest Microsoft.Data.SqlClient v1.0.19269.1. Originally it was detected in System.Data.SqlClient coming from .Net Framework - see this variant of the same repro project: https://github.com/samegutt/InconsistencyDemo.
.Net Framework version doesn't matter (tried with all from 4.5.1 to 4.8).

When compiled against sources from this repo, the problem is *also reproducible*. 
If we compile and start LoopHammer.exe in Debug mode, we immediately hit [this debug assert](https://github.com/dotnet/SqlClient/blob/master/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs#L9239). If we disable it, the further execution either leads to the same TransactionPromotionException OR to some others:

> System.ObjectDisposedException: Cannot access a disposed object.
Object name: 'SqlDelegatedTransaction'.
   at Microsoft.Data.SqlClient.SqlDelegatedTransaction.GetValidConnection()
   at Microsoft.Data.SqlClient.SqlDelegatedTransaction.Rollback(SinglePhaseEnlistment enlistment)
   at System.Transactions.TransactionStateDelegatedAborting.EnterState(InternalTransaction tx)
   at System.Transactions.Transaction.Rollback()
   at System.Transactions.TransactionScope.InternalDispose()
   at System.Transactions.TransactionScope.Dispose()
   at LoopHammer.DostuffHandler.Handle()

or even to:

> Microsoft.Data.SqlClient.SqlException (0x80131904): The Microsoft Distributed Transaction Coordinator (MS DTC) has cancelled the distributed transaction.
   at Microsoft.Data.SqlClient.SqlConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction) in C:\projects\CSA\temp\SqlClient\src\Microsoft.Data.SqlClient\netfx\src\Microsoft\Data\SqlClient\SqlConnection.cs:line 2103
   at Microsoft.Data.SqlClient.SqlInternalConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction) in C:\projects\CSA\temp\SqlClient\src\Microsoft.Data.SqlClient\netfx\src\Microsoft\Data\SqlClient\SqlInternalConnection.cs:line 814
   at Microsoft.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObject stateObj, Boolean callerHasConnectionLock, Boolean asyncClose) in C:\projects\CSA\temp\SqlClient\src\Microsoft.Data.SqlClient\netfx\src\Microsoft\Data\SqlClient\TdsParser.cs:line 1572
   at Microsoft.Data.SqlClient.TdsParser.TryRun(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj, Boolean& dataReady) in C:\projects\CSA\temp\SqlClient\src\Microsoft.Data.SqlClient\netfx\src\Microsoft\Data\SqlClient\TdsParser.cs:line 2792
   at Microsoft.Data.SqlClient.SqlDataReader.TryConsumeMetaData() in C:\projects\CSA\temp\SqlClient\src\Microsoft.Data.SqlClient\netfx\src\Microsoft\Data\SqlClient\SqlDataReader.cs:line 1310
   at Microsoft.Data.SqlClient.SqlDataReader.get_MetaData() in C:\projects\CSA\temp\SqlClient\src\Microsoft.Data.SqlClient\netfx\src\Microsoft\Data\SqlClient\SqlDataReader.cs:line 275
   at Microsoft.Data.SqlClient.SqlCommand.FinishExecuteReader(SqlDataReader ds, RunBehavior runBehavior, String resetOptionsString, Boolean isInternal, Boolean forDescribeParameterEncryption, Boolean shouldCacheForAlwaysEncrypted) in C:\projects\CSA\temp\SqlClient\src\Microsoft.Data.SqlClient\netfx\src\Microsoft\Data\SqlClient\SqlCommand.cs:line 5550
   at Microsoft.Data.SqlClient.SqlCommand.RunExecuteReaderTds(CommandBehavior cmdBehavior, RunBehavior runBehavior, Boolean returnStream, Boolean async, Int32 timeout, Task& task, Boolean asyncWrite, Boolean inRetry, SqlDataReader ds, Boolean describeParameterEncryptionRequest) in C:\projects\CSA\temp\SqlClient\src\Microsoft.Data.SqlClient\netfx\src\Microsoft\Data\SqlClient\SqlCommand.cs:line 5312
   at Microsoft.Data.SqlClient.SqlCommand.RunExecuteReader(CommandBehavior cmdBehavior, RunBehavior runBehavior, Boolean returnStream, String method, TaskCompletionSource`1 completion, Int32 timeout, Task& task, Boolean& usedCache, Boolean asyncWrite, Boolean inRetry) in C:\projects\CSA\temp\SqlClient\src\Microsoft.Data.SqlClient\netfx\src\Microsoft\Data\SqlClient\SqlCommand.cs:line 4907
   at Microsoft.Data.SqlClient.SqlCommand.RunExecuteReader(CommandBehavior cmdBehavior, RunBehavior runBehavior, Boolean returnStream, String method) in C:\projects\CSA\temp\SqlClient\src\Microsoft.Data.SqlClient\netfx\src\Microsoft\Data\SqlClient\SqlCommand.cs:line 4762
   at Microsoft.Data.SqlClient.SqlCommand.ExecuteReader(CommandBehavior behavior, String method) in C:\projects\CSA\temp\SqlClient\src\Microsoft.Data.SqlClient\netfx\src\Microsoft\Data\SqlClient\SqlCommand.cs:line 2401
   at Microsoft.Data.SqlClient.SqlCommand.ExecuteReader() in C:\projects\CSA\temp\SqlClient\src\Microsoft.Data.SqlClient\netfx\src\Microsoft\Data\SqlClient\SqlCommand.cs:line 2322
   at LoopHammer.DostuffHandler.GetTransactionDetails()
   at LoopHammer.DostuffHandler.Handle()
ClientConnectionId:48c113c4-6fcd-4082-85bf-0f39023375bf
Error Number:1206,State:137,Class:18

A transactional WCF call is the essential part of the scenario. Surprisingly, forcing a transaction promotion by other means (e.g. by accessing another DB) **does not** lead to the same problem

For some reason, the MaxDegreeOfParallelism set to **4** maximizes chances for the issue to appear. Experiments with connection pool size didn't show any differences.

As mentioned above, the issue causes a serious damage to the customer's large LoB system. Furthermore, concerns are that it might indicate a severe security issue (multiple threads being able to influence each other's transaction flow).

cheenamalhotra:
Fixes #237 

After more investigations we found that this issue occurs due to SqlClient's inability to handle "Transaction Aborted" callback in **SqlDelegatedTransaction**. When transaction promotion is triggered by WCF Service after first set of commands, the transaction is also requested to be aborted due to exception scenarios from `SysTx.Transactions`.

The driver receives callback when transaction gets aborted, but it sends this connection back to the pool, without updating its "Current Transaction" state, which gets picked up later by other new connection requests. In the second round of transaction scope, it does not respect new transaction scope created for the new connection request because there was no "currentTransaction" expected.
Hence the [Debug.Assert](https://github.com/dotnet/SqlClient/blob/98834fc04003771cd596b11b8bd1b4c6fb5cf7dd/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs#L9305) failed all the time.

Moving on with this "old" transaction, which is aborted externally, we also notice warning from SQL Server as it sends "TransactionEnded" token followed by "Begin" to initiate new transaction request, Everytime! as explained here: https://github.com/dotnet/SqlClient/issues/237#issuecomment-569965788

Most of the times we get lucky, that the transaction was aborted by SQL Server, before we tried to use it again and it always gave us new transaction to work with. But in 1% scenarios, where the error happened, server got a little late to end this transaction and we end up enlisting our connection for 2nd Transaction Scope in the 1st transaction. But an abort is pending on server side, so as soon as it aborts the transaction, all further requests to server are free of the 1st "TransactionScope", 2nd TransactionScope had no links to this round of queries, and queries get executed with "Implicit" transactions, hence the "Commits".

Everything is linked to the fact that this connection was sent back to "GeneralPool" without cleaning it's transaction state where only connections WITHOUT any "Active" transaction reference must reside.

Since this connection has enlisted transaction that got promoted to DTC but was recently aborted externally, safest approach in my opinion is to Doom this connection as soon as we capture the aborted state of current transaction to block sending back to the pool.

I couldn't find any other clean way to reset this connection properly in order to send it back to General Pool. Would welcome suggestions if that can achieved!

Please give the below package a try, that contains fix:
[Microsoft.Data.SqlClient.2.0.0-dev-437.nupkg.zip](https://github.com/dotnet/SqlClient/files/4396779/Microsoft.Data.SqlClient.2.0.0-dev-437.nupkg.zip)

cc @saurabh500 @David-Engel @scale-tone

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDelegatedTransaction.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDelegatedTransaction.cs
@@ -137,79 +137,92 @@ namespace Microsoft.Data.SqlClient
             //  from an operational standpoint (i.e. logging connection's ObjectID should be fine,
             //  but the PromotedDTCToken can change over calls. so that must be protected).
             SqlInternalConnection connection = GetValidConnection();
-
             Exception promoteException;
             byte[] returnValue = null;
-            SqlConnection usersConnection = connection.Connection;
-            SqlClientEventSource.Log.TraceEvent("<sc.SqlDelegatedTransaction.Promote|RES|CPOOL> {0}, Connection {1}, promoting transaction.", ObjectID, connection.ObjectID);
-            RuntimeHelpers.PrepareConstrainedRegions();
-            try
+
+            if (null != connection)
             {
-                lock (connection)
+                SqlConnection usersConnection = connection.Connection;
+                SqlClientEventSource.Log.TraceEvent("<sc.SqlDelegatedTransaction.Promote|RES|CPOOL> {0}, Connection {1}, promoting transaction.", ObjectID, connection.ObjectID);
+                RuntimeHelpers.PrepareConstrainedRegions();
+                try
                 {
-                    try
+                    lock (connection)
                     {
-                        // Now that we've acquired the lock, make sure we still have valid state for this operation.
-                        ValidateActiveOnConnection(connection);
-
-                        connection.ExecuteTransaction(SqlInternalConnection.TransactionRequest.Promote, null, System.Data.IsolationLevel.Unspecified, _internalTransaction, true);
-                        returnValue = _connection.PromotedDTCToken;
-
-                        // For Global Transactions, we need to set the Transaction Id since we use a Non-MSDTC Promoter type.
-                        if (_connection.IsGlobalTransaction)
+                        try
                         {
-                            if (SysTxForGlobalTransactions.SetDistributedTransactionIdentifier == null)
+                            // Now that we've acquired the lock, make sure we still have valid state for this operation.
+                            ValidateActiveOnConnection(connection);
+
+                            connection.ExecuteTransaction(SqlInternalConnection.TransactionRequest.Promote, null, System.Data.IsolationLevel.Unspecified, _internalTransaction, true);
+                            returnValue = _connection.PromotedDTCToken;
+
+                            // For Global Transactions, we need to set the Transaction Id since we use a Non-MSDTC Promoter type.
+                            if (_connection.IsGlobalTransaction)
                             {
-                                throw SQL.UnsupportedSysTxForGlobalTransactions();
+                                if (SysTxForGlobalTransactions.SetDistributedTransactionIdentifier == null)
+                                {
+                                    throw SQL.UnsupportedSysTxForGlobalTransactions();
+                                }
+
+                                if (!_connection.IsGlobalTransactionsEnabledForServer)
+                                {
+                                    throw SQL.GlobalTransactionsNotEnabled();
+                                }
+
+                                SysTxForGlobalTransactions.SetDistributedTransactionIdentifier.Invoke(_atomicTransaction, new object[] { this, GetGlobalTxnIdentifierFromToken() });
                             }
 
-                            if (!_connection.IsGlobalTransactionsEnabledForServer)
-                            {
-                                throw SQL.GlobalTransactionsNotEnabled();
-                            }
-
-                            SysTxForGlobalTransactions.SetDistributedTransactionIdentifier.Invoke(_atomicTransaction, new object[] { this, GetGlobalTxnIdentifierFromToken() });
+                            promoteException = null;
                         }
+                        catch (SqlException e)
+                        {
+                            promoteException = e;
 
-                        promoteException = null;
-                    }
-                    catch (SqlException e)
-                    {
-                        promoteException = e;
-
-                        // Doom the connection, to make sure that the transaction is
-                        // eventually rolled back.
-                        // VSTS 144562: doom the connection while having the lock on it to prevent race condition with "Transaction Ended" Event
-                        connection.DoomThisConnection();
-                    }
-                    catch (InvalidOperationException e)
-                    {
-                        promoteException = e;
-                        connection.DoomThisConnection();
+                            // Doom the connection, to make sure that the transaction is
+                            // eventually rolled back.
+                            // VSTS 144562: doom the connection while having the lock on it to prevent race condition with "Transaction Ended" Event
+                            connection.DoomThisConnection();
+                        }
+                        catch (InvalidOperationException e)
+                        {
+                            promoteException = e;
+                            connection.DoomThisConnection();
+                        }
                     }
                 }
-            }
-            catch (System.OutOfMemoryException e)
-            {
-                usersConnection.Abort(e);
-                throw;
-            }
-            catch (System.StackOverflowException e)
-            {
-                usersConnection.Abort(e);
-                throw;
-            }
-            catch (System.Threading.ThreadAbortException e)
-            {
-                usersConnection.Abort(e);
-                throw;
-            }
+                catch (System.OutOfMemoryException e)
+                {
+                    usersConnection.Abort(e);
+                    throw;
+                }
+                catch (System.StackOverflowException e)
+                {
+                    usersConnection.Abort(e);
+                    throw;
+                }
+                catch (System.Threading.ThreadAbortException e)
+                {
+                    usersConnection.Abort(e);
+                    throw;
+                }
 
-            if (promoteException != null)
-            {
-                throw SQL.PromotionFailed(promoteException);
+                //Throw exception only if Transaction is till active and not yet aborted.
+                if (promoteException != null && Transaction.TransactionInformation.Status != TransactionStatus.Aborted)
+                {
+                    throw SQL.PromotionFailed(promoteException);
+                }
+                else
+                {
+                    // The transaction was aborted externally, since it's already doomed above, we only log the same.
+                    SqlClientEventSource.Log.TraceEvent("<sc.SqlDelegatedTransaction.Promote|RES|CPOOL> {0}, Connection {1}, aborted during promotion.", ObjectID, connection.ObjectID);
+                }
             }
-
+            else
+            {
+                // The transaction was aborted externally, doom the connection to make sure it's eventually rolled back and log the same.
+                SqlClientEventSource.Log.TraceEvent("<sc.SqlDelegatedTransaction.Promote|RES|CPOOL> {0}, Connection {1}, aborted before promoting.", ObjectID, connection.ObjectID);
+            }
             return returnValue;
         }
 
@@ -217,13 +230,12 @@ namespace Microsoft.Data.SqlClient
         public void Rollback(SinglePhaseEnlistment enlistment)
         {
             Debug.Assert(null != enlistment, "null enlistment?");
-            SqlInternalConnection connection = null;
+            SqlInternalConnection connection = GetValidConnection();
 
-            try
+            if (null != connection)
             {
-                connection = GetValidConnection();
                 SqlConnection usersConnection = connection.Connection;
-                SqlClientEventSource.Log.TraceEvent("<sc.SqlDelegatedTransaction.Rollback|RES|CPOOL> {0}, Connection {1}, aborting transaction.", ObjectID, connection.ObjectID);
+                SqlClientEventSource.Log.TraceEvent("<sc.SqlDelegatedTransaction.Rollback|RES|CPOOL> {0}, Connection {1}, rolling back transaction.", ObjectID, connection.ObjectID);
                 RuntimeHelpers.PrepareConstrainedRegions();
                 try
                 {
@@ -288,14 +300,11 @@ namespace Microsoft.Data.SqlClient
                     throw;
                 }
             }
-            catch (ObjectDisposedException)
+            else
             {
-                if (_atomicTransaction.TransactionInformation.Status == TransactionStatus.Active)
-                {
-                    throw;
-                }
-                // Do not throw exception if connection is already aborted/ended from another thread,
-                // replicate as done in TransactionEnded event handler.
+                // The transaction was aborted, report that to SysTx and log the same.
+                enlistment.Aborted();
+                SqlClientEventSource.Log.TraceEvent("<sc.SqlDelegatedTransaction.Rollback|RES|CPOOL> {0}, Connection {1}, aborted before rollback.", ObjectID, connection.ObjectID);
             }
         }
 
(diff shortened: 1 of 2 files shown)


Wraith2:
This must have been a nightmare to trace.

When you say that "The driver receives callback when transaction gets aborted" do you mean that `TransactionEnded` is called? If so isn't it possible that from there you could call `connection.CleanupConnectionOnTransactionCompletion` here as is done in the `Rollback` implementation?

It might also be worth adding some checks in the pool return function for various pieces of dangling state and dooming the connection at the return to pool stage. If done this way any other errors from the inappropriate release could be forced safe without finding all possible caller combinations and we could have have it assert in tests.

scale-tone:
Nice to see progress on this (think it's safe to say) critical issue, @cheenamalhotra . Still we don't seem to have full understanding of the mechanics. 

If you go up in the stack, you'll see that [another attempt to properly cleanup that connection before returning to the pool](https://github.com/dotnet/SqlClient/blob/29b65c04450f1ab8cc175a33b25415b53e106680/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/ProviderBase/DbConnectionInternal.cs#L939) is being made:

```
            CleanupConnectionOnTransactionCompletion(transaction);
```
Do you have an idea, why that attempt isn't successful? If that line effectively does nothing, then why is it in place after all?

I can confirm that with the proposed **DoomThisConnection()**  line added the issue stops happening. But **ObjectDisposedException** still occurs. Why is that?

@Wraith2 , **CleanupConnectionOnTransactionCompletion()** doesn't help. With that line added the issue still occurs (can't explain, why).

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDelegatedTransaction.cs
```diff
@@ -413,6 +413,8 @@ internal void TransactionEnded(Transaction transaction)
                         _active = false;
                         _connection = null;
                     }
+                    // Safest approach is to doom this connection, whose transaction has been aborted externally.
+                    connection.DoomThisConnection();
```
saurabh500: Does this mean that we doom the connection in all cases of external transaction abort and not just in doubtful cases like the one we were chasing?

cheenamalhotra: Yes, this is required to prevent driver trying to reuse the pooled connection and attempting to reuse Open Transaction at client side. The problem is even if this transaction was aborted, the driver got notified, it still didn't mark it "Aborted" and continues to re-use next time.
[This is current behavior applies to all cases, which should be fixed].

If you can suggest an alternative way to mark "Current Transaction" as "Aborted" before sending connection to "GeneralPool", also ignoring that next time, that would also be a viable solution. I couldn't get this to work due to lot of mixed cases that conflict with each other.

Wraith2: Is it possible that we could store whether we expect the transaction to be aborted or not and then if we find that an abort event is received without us expecting one then we know we need to cleanup? This was what I thought calling `connection.CleanupConnectionOnTransactionCompletion` would do but apparently it does not which seems strange.

saurabh500: I am trying to understand what the user will see with this fix. If the connection is reused from the pool, and it gets an abort notification later while it is being used, then we are going to doom the connection in the middle of its usage right ?

saurabh500: Before a connection with existing transaction is being reused, I think we should expect the server to drive this transaction to completion either with Abort or Complete or maybe another state. The problem is that we see the abort *after* we have started using the connection, and perhaps the driver can conclude that for this connection's existing transaction, the driver never received a response from the server about the state of the older transaction, and that could be used as a mechanism to *not* use the connection.

cheenamalhotra: > I am trying to understand what the user will see with this fix. If the connection is reused from the pool, and it gets an abort notification later while it is being used, then we are going to doom the connection in the middle of its usage right ?

No, this connection is picked up from "Transaction Pool" and moved to "General Pool" when "Abort" happens, it's not in-use in that time. User will not notice anything, except performance lag due to a new physical connection being made in this case.

> Before a connection with existing transaction is being reused, I think we should expect the server to drive this transaction to completion either with Abort or Complete or maybe another state. The problem is that we see the abort after we have started using the connection...

Since we discussed this later, I'll just put comments to clarify for readers: The service that notifies driver about Abort also notifies server for this Abort action. It just happens that since this is asynchronous callback, and if server is already doing something on that thread, it gets a a little late to perform abort. The problem actually is we see Abort first, but we don't clean the connection's transaction state, which leads to transaction re-use.

cheenamalhotra:
@scale-tone 

I looked up more into the `CleanupConnectionOnTransactionCompletion` method and it does not cleanup completely the "CurrentTransaction" associated with "SqlInternalConnectionTds" object instance. I was able to modify the code to do so, and could clean first connection's transaction reference to be transferred back to "General Pool" safely.

However, doing so does not work well in multi-threaded scenarios, and seems like there's more to this problem than we're looking at from face-value if we try to set things right here.

Alternatively, I'm also going to investigate making this PR more robust to avoid any possible thread-safety issues so we can have one solution to address this problem, while the other can still be investigated, where we can allow re-use of connection.

>  If that line effectively does nothing, then why is it in place after all?

This is effective to remove the connection from Transaction Pool, but we definitely need to groom it to do only necessary steps in this case.

> But ObjectDisposedException still occurs.

I will come back on this, after I sanitize the code to do only required tasks on "Abort". It seems to be related in initial glance. Could you also share your stacktrace to confirm I'm also looking at the same error.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlDelegatedTransaction.cs
```diff
@@ -331,27 +328,36 @@ public void Rollback(SysTx.SinglePhaseEnlistment enlistment)
...
+                }
 #if DEBUG
                 finally
                 {
                     tdsReliabilitySection.Stop();
                 }
 #endif //DEBUG
             }
-            catch (System.OutOfMemoryException e)
+            catch (ObjectDisposedException)
```
saurabh500, line 353: I remember you talking about ObjectDisposedException. However an ObjectDisposedException being found is an indicator that we are re-using a disposed object and it shouldn't be handled in a try/catch

https://docs.microsoft.com/en-us/dotnet/api/system.objectdisposedexception?view=netframework-4.8

`In most cases, this exception results from developer error. Instead of handling the error in a try/catch block, you should correct the error, typically by reinstantiating the object.
`

cheenamalhotra: I could change the logic to NOT throw object disposed exception, but that would also apply to other places where we DO want to throw this exception, that's why I went this way.

This is very specific to Rollback/Abort, since when there's a race condition, we see different types of transaction terminations at server side:
1. Internal Abort
2. Rollback - Triggered by driver when we receive notifications to Rollback the transaction. (Calls are intermittent and may/may not happen due to application design. DTC might have already parallelly aborted the tx.)

Also commented in this method above and can be seen in the code of `TransactionEnded`, we don't really care if this `_connection` is already aborted or null.

Rollback on a non-active connection should not fail.

saurabh500: i want to know why are we getting the ObjectDispisedException in the first place. What is the object and type in question.
We shouldn't be getting this exception at all.
I agree that rollback shouldn't throw an exception. My point is that we shouldn't be catching the ObjectDisposedException but removing the reason this exception occurred. Occurrence of this exception is more often than not due to a bug that should be chased down.

cheenamalhotra: We get this exception from [here](https://github.com/dotnet/SqlClient/pull/495/files/2f59a6f4b00fd53b2d2322271e6b1f3f9681bfd7#diff-2a7d4b53e8840c0897e356a092430483R525) because TransactionEnded got executed before Rollback, in race condition.

image

The callbacks are triggered externally, so we have no control over that.

cheenamalhotra: We got this exception from [here](https://github.com/dotnet/SqlClient/pull/495/files/2f59a6f4b00fd53b2d2322271e6b1f3f9681bfd7#diff-2a7d4b53e8840c0897e356a092430483R525) because TransactionEnded (triggerd by DTC > SysTx) got executed before Rollback (triggered by SysTx), in race condition and it set `_connection` to null before it was acquired for Rollback to execute.

image

Since these callbacks are external, we cannot control their order of execution.

cheenamalhotra:
@scale-tone 

I did manage to fix the problem with `ObjectDisposedException` as it was thrown due to race condition scenarios between `Rollback` and `Abort`. I did not find the need to modify code any further, and I also verified with DB Performance Pool Counters we don't have any leaks, the connections are getting cleaned up as expected.

However, there may be cases where connection takes time to connect and may lead to timeout if running too many parallel executions, increasing connection timeout is a viable solution in those scenarios.

The alternate solution of cleaning the connection state and sending it back to the General pool needs deep investigations as that leads to big design changes in the driver. I will get back with progress on that one, but also would request you verify this option as solution.

Uploading updated NuGet package here for testing:
[Microsoft.Data.SqlClient.2.0.0-dev-pr495-1.zip](https://github.com/dotnet/SqlClient/files/4424521/Microsoft.Data.SqlClient.2.0.0-dev-pr495-1.zip)

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlDelegatedTransaction.cs
saurabh500: OK, looks like that the GetValidConnection() was only called to check if the user is using a Disposed connection to call this API.
We should be able to track that the transaction ended call was triggered due to the callback, even though we cannot control the timing of the callback trigger since it is Sys.Transaction triggered, which is DTC triggered. 
This affirms that we shouldn't be swallowing the ObjectDisposedException because that was probably added for Api correctness.

saurabh500:
@cheenamalhotra Question about the approach: Is this approach of dooming the connection on ending transaction final? 

There has to be a different approach considered, where if the transaction is not ended, then we wait for it to end, and possibly not return this connection to the pool till its related transaction has ended. 
The DTC and potentially not DTC customers using `SqlTransaction` will hit a performance problem while retrieving connections from the pool. The change effectively disables pooling for all transacted connections.

cheenamalhotra:
Hi @saurabh500 

I am investigating that option too, but wanted to make 1 solution a complete fix that can be considered as a solution if we don't find a working solution with alternate solution as it's more important to fix the problem than to worry about performance which is a very edge case scenario. Also this is not going to completely stop Pooling on DTC Transactions. We will only doom the connection if it has been Aborted externally.

In normal Commit/Rollback cases, we will continue to pool the connection as always.

The alternate solution is a can of worms which is more tricky due to the design of "Transaction Pool" that differs from "General Pool". We may have to do a lot of changes to achieve the right solution we're looking for. It's in progress, but is more time consuming too. If it gets too complex, we may have to take a call if it's worth the effort.

saurabh500:
I see, I agree that the perf problem will hit in case of a DTC transaction abort. Sure. 

Here's how I am looking at this situation, which makes performance very important in failure cases. 
An application is using DTC to coordinate transactions between SQL Server and another system i.e. a WCF service in this case.

Let's say that the transactions start aborting due to WCF service throwing errors due to some issues (networking / bad configuration deployment / something else not working, failover testing). All the connection participating in those transactions will be doomed and new connections will need to be created the next time a sql interaction is needed. This will overload the connection pool causing timeouts in the connection pool in scenarios where some other part of the service is experiencing problems. So the problem will shift from WCF throwing exceptions to application receiving SqlException during transient errors in some other part of the application.

The connection pool is single threaded for each pool and it will have a hard time providing connections in case of a WCF service or another component participating in DTC, experiencing short durations of outages. 
Couple this with the connection pool blocking period where the applications will start seeing a lot of timeouts while trying to get connections from the pool. 
These edge case perf issues will cause outages for the applications and investigations about why a call to WCF failed will then become why is SqlClient timing out in case of failures in some other parts of the system even though the root cause of WCF outage was fixed.

saurabh500:
@cheenamalhotra my comment above was not to dissuade fixing this issue in steps. It was something I wanted to list down before I lost my train of thought. 

I do agree that this fix will at guarantee correctness which is high priority item in the light of this bug you have been working on.

By the end of completing this investigation, if you could open an issue to list a performance impact and possible solutions to address the edge case perf case, that would be nice. 
At least folks hitting any issues can search the internet and come to github to find out what is the problem and the related discussion. And if this perf problem becomes a real issue later down the line, a developer looking at this issue will have context since investigating this problem as a perf issue later is going to be harder than debugging this issue itself. 

Since you have rich context around this issue, I think we should definitely list down your findings in another issue as perf issue, and the potential approaches you have tried and why they failed.

scale-tone:
> Could you also share your stacktrace to confirm I'm also looking at the same error.

It is clearly stated in the [bug](https://github.com/dotnet/SqlClient/issues/237):
image

cheenamalhotra:
Opening a new PR with same branch as CI doesn't seem to trigger here.

PR #497, opened 2020-03-30
Add SslOverTdsStream tests

Wraith2:
I intend to change the internal class `SslOverTdsStream` to remove sync-over-async and remove some allocations. To do that I need to add tests to prevent adding of new bugs. 

These use each pair of `Read` and `Write` overloads on `Stream` to write a known pattern of 4 encapsulated and 4 unencapsulated packets to a memory stream and then read them back and verify the data is the same. Reflection is used to find, instantiate and call `FinishHandshake` on the stream. I chose to do this rather than include the source files because it's easier in this case where there are only two methods required and it doesn't introduce a code share between dev and test.


src/Microsoft.Data.SqlClient/tests/FunctionalTests/Microsoft.Data.SqlClient.Tests.csproj
```diff
@@ -60,6 +61,7 @@
   </ItemGroup>
   <ItemGroup Condition="'$(TargetFramework)' != 'netcoreapp2.1' AND '$(TargetGroup)' == 'netcoreapp'">
     <PackageReference Include="System.Data.Odbc" Version="$(SystemDataOdbcVersion)" />
+    <Compile Include="SslOverTdsStreamTest.NetCoreApp.cs" />
```
cheenamalhotra, line 64: This ItemGroup applies to only netcoreapp3.1.
You can skip testing on NetFx by specifying class level tags as done in other test classes.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SslOverTdsStreamTest.cs
```diff
@@ -0,0 +1,278 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Microsoft.Data.SqlClient.Tests
+{
+    public static partial class SslOverTdsStreamTest
+    {
+        [Theory]
+        [SkipOnTargetFramework(TargetFrameworkMonikers.NetFramework)]
```
cheenamalhotra, line 14: I suggest moving all tests in same class and tagging as done here.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SslOverTdsStreamTest.cs
```diff
@@ -0,0 +1,278 @@
...
+        [InlineData(0),InlineData(3),InlineData(128), InlineData(2048), InlineData(8192)]
+        public static void ReadWrite(int readLimit)
+        {
+            const int EncapulatedPacketCount = 4;
```
cheenamalhotra, line 18: Did you mean 'Encapsulated' here?

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SslOverTdsStreamTest.cs
```diff
@@ -0,0 +1,278 @@
...
+
+        private static void InvokeFinishHandshake(Stream stream)
+        {
+            MethodInfo method = stream.GetType().GetMethod("FinishHandshake", BindingFlags.Public | BindingFlags.Instance);
+            method.Invoke(stream, null);
+        }
+
+        private static Stream CreateSslOverTdsStream(Stream stream)
+        {
+            Type type = typeof(SqlClientFactory).Assembly.GetType("Microsoft.Data.SqlClient.SNI.SslOverTdsStream");
```
cheenamalhotra, line 111: Seems like your "Core" reflects Managed SNI, so I guess method naming needs to be done better.

Wraith2: SslOverTdsStream currently only exists in core. I built the tests so that if core was built for versions of core without with async stream extensions (which would be 2.1 or lower for core or all netfx) that it could still be used. Say for example you wanted to replace parts of netfx build with parts of the netcore build it would make testing easier for you.

I can rename if you want but I thought common things would be the normal case and that things on netcore only could have the core suffix. Any suggestions on what to change it to?

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SslOverTdsStreamTest.cs
Wraith2: There can be no equivalent test on netfx because the class doesn't exist there. If the project were built against netfx the same test could be enabled just by removing this Skip annotation, and in that case only the core parts separated out would need to be compiled out, as they are currently.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/Microsoft.Data.SqlClient.Tests.csproj
Wraith2: The core methods are available on core 2.1 and later so yes they should probably be in another compile group but i was having a lot of trouble getting the test to compile in all configurations locally so I had to rely on the CI to show me the problems. This would only cause a problem if you ran tests against 2.1 and not 3.1 which seems an unlikely scenario. I'll see if there's a define I can change it to though.

Wraith2: Oh, and this and every other project will need revisiting once we know the TFM for net 5 since it shouldn't be netcoreapp anymore

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SslOverTdsStreamTest.cs
cheenamalhotra: Ok, I see.
I think we should in that case consider adding additional Moniker in Xunit Extensions that can be applied to specific tests. without the need to splitting classes.

@JRahnama can we extend our `TargetFrameworkMonikers` to include "Netcoreapp21" that when set would skip on "NetCoreApp2.1" target framework?

Wraith2: It should work on netcore 2.1 according to the api tables so I don't think we need to add more testing capability. I just found it useful to split the functionality into core and netfx files with common parts rather than use #defines so it can be made available on netfx if/when we converge into a single codebase that contains a managed sni implementation.

I'm open to changing the names if you have suggestions.

JRahnama: @cheenamalhotra are we still going to proceed?

cheenamalhotra: If your tests are compatible with netcoreapp2.1, do we still need to split the test file? Keeping it in 1 file would be better.

Wraith2: I explained my thinking above. I think it's ok this way. If you really want I can change it but I don't see it as a problem to have it split apart the way it is, the complexity isn't huge.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SslOverTdsStreamTest.cs
```diff
@@ -0,0 +1,345 @@
...
+                InvokeFinishHandshake(tdsStream);
+                for (int index = 0; index < passthroughPacketCount; index++)
+                {
+                    Array.Clear(bytes, 0, bytes.Length);
+                    int packetBytes = ReadPacket(tdsStream, action, bytes);
+                    Array.Copy(bytes, 0, output, offset, packetBytes);
+                    offset += packetBytes;
+                }
+            }
+
```
cheenamalhotra, line 184: Nit: Remove extra line

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SslOverTdsStreamTest.cs
```diff
@@ -0,0 +1,345 @@
...
+            }
+            else
+            {
+                return base.Read(destination);
+            }
+        }
+
+        public override ValueTask<int> ReadAsync(Memory<byte> destination, CancellationToken cancellationToken = default)
+        {
+
```
cheenamalhotra, line 334: Nit: Remove extra line

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SslOverTdsStreamTest.cs
```diff
@@ -0,0 +1,345 @@
...
+using System.Diagnostics;
+using System.IO;
+using System.Reflection;
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Microsoft.Data.SqlClient.Tests
+{
+    public static partial class SslOverTdsStreamTest
```
cheenamalhotra, line 11: Doesn't have to partial class anymore.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SslOverTdsStreamTest.cs
Wraith2: Done.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SslOverTdsStreamTest.cs
Wraith2: Done.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SslOverTdsStreamTest.cs
Wraith2: Done.

cheenamalhotra:
The PR is currently in review with team @Wraith2. We are hoping to get this merged soon! 😊

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SslOverTdsStreamTest.cs
```diff
@@ -0,0 +1,343 @@
+﻿using System;
```
cheenamalhotra, line 1: Please add license headers

Wraith2: Done.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SslOverTdsStreamTest.cs
```diff
@@ -0,0 +1,347 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Microsoft.Data.SqlClient.Tests
+{
+    public static class SslOverTdsStreamTest
+    {
+        [Theory]
+        [InlineData(0),InlineData(3),InlineData(128), InlineData(2048), InlineData(8192)]
```
JRahnama, line 18: Is there any specific reason we used all inline datas in the same run instead of having 5 different lines of inlinedata?
[InlineData(0)]
[InlineData(3)] and etc...?

JRahnama: I just checked the duration your design was better by 2ms

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SslOverTdsStreamTest.cs
```diff
@@ -0,0 +1,347 @@
...
+
+            SyncTest(EncaspulatedPacketCount, PassThroughPacketCount, readLimit);
+            SyncCoreTest(EncaspulatedPacketCount, PassThroughPacketCount, readLimit);
+            AsyncTest(EncaspulatedPacketCount, PassThroughPacketCount, readLimit);
+            AsyncCoreTest(EncaspulatedPacketCount, PassThroughPacketCount, readLimit);
+        }
+
+        private static void SyncTest(int encapsulatedPacketCount, int passthroughPacketCount, int maxPacketReadLength)
+        {
+            byte[] input;
```
JRahnama, line 32: this variable could be simplified by taking them inside  `SetupArrays(encapsulatedPacketCount + passthroughPacketCount, out input, out output);` with a little bit of change as 
`SetupArrays(encapsulatedPacketCount + passthroughPacketCount, out byte[] input, out  byte[]output);`

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SslOverTdsStreamTest.cs
```diff
@@ -0,0 +1,347 @@
...
+        }
+
+        private static void Validate(byte[] input, byte[] output)
+        {
+            Assert.True(input.AsSpan().SequenceEqual(output.AsSpan()));
+        }
+
+        internal static class TdsEnums
+        {
+            public const int DEFAULT_LOGIN_PACKET_SIZE = 4096;
```
JRahnama, line 273: Adding the comment from TdsEnum could make it easier to follow.
```
// Login packet can be no greater than 4k until server sends us env-change
        // increasing packet size.
```
Question: what happens when server sends the env-change and packet size is changed? the max packet size is 32768. willt he test fail in that situation?

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SslOverTdsStreamTest.cs
Wraith2: It's a test so performance wasn't really a huge concern. It's just personal preference because as far as I know there's no functional difference between mutltiple attribute sections or one. If There's a specific way that is preferred (and it isn't this one) I can change it.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SslOverTdsStreamTest.cs
Wraith2: Yes but doing so have little benefit and changes the location of the setup of the test from the decorated entry point that you see in the test explorer to a callee method that you'd have to dig into the test to find. 

For tests I prefer verbosity and clarity over minimalization. Especially in this case where the reason for the test might not be entirely obvious. I really want the test to explain itself if you look at it because the primary reason for looking is that it has failed and in that case the extra information can help you find your mistake.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SslOverTdsStreamTest.cs
Wraith2: I haven't tested it but I believe that it would. In practice I've only ever seen (and based on the behaviour I expect there to be) only one pre-handshake packet. The test uses more than one packet so that if it ever did manage to happen we have coverage. I don't see a way to change the packetsize mid login.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SslOverTdsStreamTest.cs
```diff
@@ -0,0 +1,345 @@
...
+            Array.Copy(bytes, 0, output, 0, offset);
+            return offset;
+        }
+
+        private static byte[] WritePackets(int encapsulatedPacketCount, int passthroughPacketCount, Action<Stream, int> action)
+        {
+            byte[] buffer = null;
+            using (LimitedMemoryStream stream = new LimitedMemoryStream())
+            {
+                //using (SslOverTdsStream tdsStream = new SslOverTdsStream(stream))
```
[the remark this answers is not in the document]
Wraith2: Possibly, but leaving it in tells you what the following line is equivalent to if it were possible directly instead of needing reflection. I can take it out if you prefer but I prefer verbosity in tests just to make working out why they might have failed easier.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SslOverTdsStreamTest.cs
```diff
@@ -0,0 +1,347 @@
...
+namespace Microsoft.Data.SqlClient.Tests
+{
+    public static class SslOverTdsStreamTest
+    {
+        [Theory]
+        [InlineData(0),InlineData(3),InlineData(128), InlineData(2048), InlineData(8192)]
+        public static void ReadWrite(int readLimit)
+        {
+            const int EncaspulatedPacketCount = 4;
+            const int PassThroughPacketCount = 5;
```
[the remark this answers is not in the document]
Wraith2: Clarity. I think it shows that rather than passing 9 packets through a stream you're doing 4 and then 5 indicating that there's a difference.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SslOverTdsStreamTest.cs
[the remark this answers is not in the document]
DavoudEshtehari: Greate, I agree with verbosity in tests too. In my opinion, it is maybe a better choice to describe the reason for creating `CreateSslOverTdsStream` instead of leaving this code as a comment here. In addition, your function name also illustrates its duty.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SslOverTdsStreamTest.cs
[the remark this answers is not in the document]
DavoudEshtehari: So, maybe it is not bad to add some delay between them and increase the section numbers.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SslOverTdsStreamTest.cs
```diff
@@ -0,0 +1,345 @@
...
+
+namespace Microsoft.Data.SqlClient.Tests
+{
+    public static partial class SslOverTdsStreamTest
+    {
+        [Theory]
+        [InlineData(0),InlineData(3),InlineData(128), InlineData(2048), InlineData(8192)]
+        public static void ReadWrite(int readLimit)
+        {
+            const int EncaspulatedPacketCount = 4;
```
DavoudEshtehari, line 21: Mistyping: `EncapsulatedPacketCount`

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SslOverTdsStreamTest.cs
JRahnama: Fair enough. Thanks for the explanation.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SslOverTdsStreamTest.cs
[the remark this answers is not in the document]
Wraith2: I'm not sure what you mean, can you give me an example of how you'd like it changed?

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SslOverTdsStreamTest.cs
```diff
@@ -0,0 +1,347 @@
...
+            const int PassThroughPacketCount = 5;
+
+            SyncTest(EncaspulatedPacketCount, PassThroughPacketCount, readLimit);
+            SyncCoreTest(EncaspulatedPacketCount, PassThroughPacketCount, readLimit);
```
JRahnama, line 25: Another thing, can we have these as separate 4 tests instead of one single test? If one fails we do not want the rest fails.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SslOverTdsStreamTest.cs
```diff
@@ -0,0 +1,347 @@
...
+
+namespace Microsoft.Data.SqlClient.Tests
+{
+    public static class SslOverTdsStreamTest
+    {
+        [Theory]
+        [InlineData(0),InlineData(3),InlineData(128), InlineData(2048), InlineData(8192)]
+        public static void ReadWrite(int readLimit)
+        {
+            const int EncaspulatedPacketCount = 4;
```
JRahnama, line 21: Can you move these hardcoded items to InlineData as well?

Wraith2: I could but there is no benefit to doing so which is why I chose to leave them hardcoded. The choice of multiple packets is already doing more than the real-world usage will require. The pattern of data is chosen so that it can't accidentally collide (well, it can but it's numerically impractical to do so and well outside the scope of these tests). 

Do you want me to add the additional complexity?

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SslOverTdsStreamTest.cs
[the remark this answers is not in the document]
Wraith2: Comment removed.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SslOverTdsStreamTest.cs
Wraith2: fixed.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SslOverTdsStreamTest.cs
Wraith2: Split, which caused the arg format to change so i made a TheoryData to pass them.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SslOverTdsStreamTest.cs
Wraith2: Split out the items into the TheoryData generator. If for some reason anyone every wants more all the have to do is add more cases to the initialization property function.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SslOverTdsStreamTest.cs
```diff
@@ -0,0 +1,345 @@
...
+                {
+                    return await stream.ReadAsync(
+                        new Memory<byte>(bytes, offset, count)
+                    );
+                }
+            );
+
+            Validate(input, output);
+        }
+
```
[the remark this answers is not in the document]
Wraith2: Removed.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SslOverTdsStreamTest.cs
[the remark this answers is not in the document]
DavoudEshtehari: This test is perfect. 
I have been thinking about a comprehensive TDS test to examine every aspect of its functionality like time-consuming streams by adding a delay between packets and increasing the number of them.
You have started a new thread on testing. I like your way of developing and the thought behind it.
Good job @Wraith2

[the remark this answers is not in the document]
Wraith2: The wrapper memory stream did have a delay option for async methods in the original but it added a lot of complexity and all it was really testing was whether the runtime was working the way it should. I chose to trust the runtime team with async and deal with the bits that I can change.

PR #504, opened 2020-04-02
Common File Sharing between NetFx and NetCore

cheenamalhotra:

PR #506, opened 2020-04-05
perf: remove ActivityCorrelator.Current.ToString() calls in tracing

Wraith2:
Another small one. While profiling I noticed some calls to `ActivityCorrelator.Current.ToString()` coming from calls to trace methods, the trace methods call ToString() on the args if logging of that method is enabled so these are just wasted string allocations for most people.


cheenamalhotra:
Please also update on NetFx side.

Wraith2:
I've said repeatedly that maintaining two separate codebases for this library is not helpful. Can you please come up with a policy decision on this instead of ignoring it and repeatedly asking me to backport things to netfx that aren't going to make a difference to it.

cheenamalhotra:
Ok, I guess we will then tag such PRs with the tag "Port to NetFx" and create a parallel PR to be merged along with yours.

David-Engel:
> I've said repeatedly that maintaining two separate codebases for this library is not helpful. Can you please come up with a policy decision on this instead of ignoring it and repeatedly asking me to backport things to netfx that aren't going to make a difference to it.

@Wraith2 We've said in the past that we would like to keep the code bases equivalent, when it is reasonable to do so. Thus we have consistently asked to make changes to both in previous PRs. We have been flexible about it when it does not make sense to change the netfx code or is too burdensome to do so. Long-term, we would like to have one code base and we will try to work incrementally towards that (see #504). Diverging code now makes that more difficult when someone in the future tries to reconcile differences.

Issue #382, opened 2020-01-15
Implement DBConnectionPoolCounters for .NET Core


cheenamalhotra:
Currently SqlClient supports [Performance Counters](https://docs.microsoft.com/en-us/dotnet/framework/data/adonet/performance-counters) in .NET Framework implementation, but since it was not supported in .NET Core 2.1 when the driver was ported over, the support was removed in CoreFx.

.NET Core 3.0 added support for [PerformanceCounters](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.performancecounter?view=netcore-3.0) and that brings an opportunity to introduce [DbConnectionPoolCounters](https://github.com/dotnet/SqlClient/blob/master/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/ProviderBase/DbConnectionPoolCounters.cs) for .NET Core applications as well.

This would require to build a separate DLL for .NET Core 3.1+ with similar API support.

roji:
Following conversation on diagnostics today, it would be good to have a good database perf counters story, that's consistent across database providers and well-integrated in VS.

Some resources:

* Blog post announcing the .NET Core 3.0 diagnostics improvements (including dotnet counters): https://devblogs.microsoft.com/dotnet/introducing-diagnostics-improvements-in-net-core-3-0/
* Quick demo of Npgsql's current perf counters being displayed by dotnet counters: https://asciinema.org/a/256182. Note that the list of counters and their names isn't final and can change based on our discussions.
* Npgsql's EventSource producing the counters: https://github.com/npgsql/npgsql/blob/dev/src/Npgsql/NpgsqlEventSource.cs.

Note that the new EventSource-based API is completely different from the old .NET Framework Performance Counters API. Since consumption from outside the process is also completely different, we really can decide which counters to expose, rather than just carry over what existed in .NET Framework.

@karpinsn @davidmikh what logic is planned to be built into VS around perf counters? Is the idea for you guys to know about certain specific event source names (e.g. System.Runtime), and even specific counters (e.g. cpu-usage within System.Runtime), to display them in a special way? Or would this be a totally generic feature, which ideally allows the user to choose any event source name exposed by the process, and any counter within those?

/cc @ajcvickers

daniel-munch-cko:
Very interested in this as well. Note that there's been a similar discussion for the MySqlConnector around here: https://github.com/mysql-net/MySqlConnector/issues/471

roji:
[...]

PR #509, opened 2020-04-07
Introducing support for DBConnectionPoolCounters in .NET Core

cheenamalhotra:
Addresses #382 

Sharing implementation of `DBConnectionPoolCounters` between .NET Framework and .NET Core 3.1.


cheenamalhotra:
Closing this as we'd be moving focus to Event Counters instead in .NET Core, see original issue for discussion.

PR #512, opened 2020-04-07
remove redundant DbConnectionPoolKey gettype

SimonCropp:
i think the GetType is redundant since an as is done below. also converted to a pattern match instead

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Common/src/Microsoft/Data/Common/DbConnectionPoolKey.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Common/src/Microsoft/Data/Common/DbConnectionPoolKey.cs
@@ -42,14 +42,12 @@ namespace Microsoft.Data.Common
 
         public override bool Equals(object obj)
         {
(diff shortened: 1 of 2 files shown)


cheenamalhotra:
LGTM

Could you also update "netcore" class ([here](https://github.com/dotnet/SqlClient/blob/22c9cc49f15f93e11205f1b886dcdb68ad5b611b/src/Microsoft.Data.SqlClient/netcore/src/Common/src/Microsoft/Data/Common/DbConnectionPoolKey.cs#L43)) to reflect the same change?

SimonCropp:
@cheenamalhotra yep. nice attention to detail, done

SimonCropp:
do u want all the other double ups of "as and null checks" converted to pattern matching?

cheenamalhotra:
> do u want all the other double ups of "as and null checks" converted to pattern matching?

We'd have to be extra careful for those changes to verify it does not change behavior, improvements are always welcome!

SimonCropp:
well i will look at that in a diff PR.

PR #519, opened 2020-04-09
Combine shared elements

Wraith2:
Add more shared elements from netcore and netfx to the common folder. I believe the lack of increment in the ActivityCorrelator for netfx was an error so I took the netcore version. The others are identical.


src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
```diff
@@ -36,6 +36,12 @@
     <Compile Include="..\..\src\Microsoft\Data\SqlClient\SqlClientLogger.cs" >
       <Link>Microsoft\Data\SqlClient\SqlClientLogger.cs</Link>
     </Compile>
+    <Compile Include="..\..\src\Microsoft\Data\Sql\SqlNotificationRequest.cs">
```
JRahnama, line 39: NIT: is it possible to add a path to Directory.Build.props to use the variable name instead of` ..\..\`.
We do have an example of it in the code in that directory as below:
`<CommonPath>$(NetCoreSource)src\Common\src</CommonPath>` and that $(CommonPath) could be used all over the project. Can we do the same for this newly added shared folder?

cheenamalhotra: The reason why I didn't go for this is due to conflicting terms. We already have $(CommonPath) that can be misleading. There's some cleanup needed in that section of common code, after that we can address this.

PR #521, opened 2020-04-12
Change _SqlMetaData to lazy initialize hidden column map

Wraith2:
This PR started as a small perf optimization to eliminate the visible column map in the `_SqlMetaDataSet` type but in doing so I found that it wasn't going to work as expected anyway so I've fixed it.

Some background. SQL Server can return additional hidden columns that the user didn't request and if it does this it will mark the metadata entry as hidden. I've never seen one of these and can't find any way to cause them to happen (so no tests) but the behaviour is in the TDS spec. There are only a small number of places that need to care if a column is visible and those places are all acting on entire rows rather than getting a column requested by the user, `GetValues` and `GetSqlValues` are the primary consumers with one additional use in Smi (SQL Metadata Interface?) extended metadata generation. 

Currently every time the metadata is read from the TDS stream an int array is allocated and populated with a map of actual metadata to visible metadata indices. This is almost never used and is thrown away. It can't be cached usefully because it is query dependent. This PR changes the generation of that map to be lazy initialized inside the `_SqlMetaDataSet` type and adds some abstraction so that callers only need to know about whether they want to know about real columns or visible columns.

The way that the visible column map in the current implementation is used causes incorrect output. Consider this gist: https://gist.github.com/Wraith2/06c9391808a43f9bae729a5b52a6bd10
If I have 4 columns (0,1,2,3) and set 1 to be hidden when I call `GetSqlValues` I would expect to get (0,2,3), in the current implementation I do not, I get (0,2,default) and other hidden column locations give similar wrong results.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
@@ -410,7 +410,7 @@ namespace Microsoft.Data.SqlClient
                 {
                     return 0;
                 }
-                return (md.VisibleColumnCount);
+                return md.VisibleColumnCount;
             }
         }
 
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
@@ -278,8 +278,8 @@ namespace Microsoft.Data.SqlClient
 
             if (null != metaData && 0 < metaData.Length)
             {
-                metaDataReturn = new SmiExtendedMetaData[metaData.visibleColumns];
-
+                metaDataReturn = new SmiExtendedMetaData[metaData.VisibleColumnCount];
+                int returnIndex = 0;
                 for (int index = 0; index < metaData.Length; index++)
                 {
                     _SqlMetaData colMetaData = metaData[index];
@@ -318,7 +318,7 @@ namespace Microsoft.Data.SqlClient
                             length /= ADP.CharSize;
                         }
 
-                        metaDataReturn[index] =
+                        metaDataReturn[returnIndex] =
                             new SmiQueryMetaData(
                                 colMetaData.type,
                                 length,
@@ -346,6 +346,8 @@ namespace Microsoft.Data.SqlClient
                                 colMetaData.IsExpression,
                                 colMetaData.IsDifferentName,
                                 colMetaData.IsHidden);
+
+                        returnIndex += 1;
                     }
                 }
             }
@@ -408,7 +410,7 @@ namespace Microsoft.Data.SqlClient
                 {
                     return 0;
                 }
-                return (md.visibleColumns);
+                return (md.VisibleColumnCount);
             }
         }
 
@@ -1141,31 +1143,6 @@ namespace Microsoft.Data.SqlClient
                     return false;
                 }
                 Debug.Assert(!ignored, "Parser read a row token while trying to read metadata");
-            }
-
-            // we hide hidden columns from the user so build an internal map
-            // that compacts all hidden columns from the array
-            if (null != _metaData)
-            {
-                if (_snapshot != null && object.ReferenceEquals(_snapshot._metadata, _metaData))
-                {
-                    _metaData = (_SqlMetaDataSet)_metaData.Clone();
-                }
-
-                _metaData.visibleColumns = 0;
-
-                Debug.Assert(null == _metaData.indexMap, "non-null metaData indexmap");
-                int[] indexMap = new int[_metaData.Length];
-                for (int i = 0; i < indexMap.Length; ++i)
-                {
-                    indexMap[i] = _metaData.visibleColumns;
-
-                    if (!(_metaData[i].IsHidden))
-                    {
-                        _metaData.visibleColumns++;
-                    }
-                }
-                _metaData.indexMap = indexMap;
             }
 
             return true;
@@ -2623,11 +2600,11 @@ namespace Microsoft.Data.SqlClient
 
                 SetTimeout(_defaultTimeoutMilliseconds);
 
-                int copyLen = (values.Length < _metaData.visibleColumns) ? values.Length : _metaData.visibleColumns;
+                int copyLen = (values.Length < _metaData.VisibleColumnCount) ? values.Length : _metaData.VisibleColumnCount;
 
                 for (int i = 0; i < copyLen; i++)
                 {
-                    values[_metaData.indexMap[i]] = GetSqlValueInternal(i);
+                    values[i] = GetSqlValueInternal(_metaData.GetVisibleColumnIndex(i));
                 }
                 return copyLen;
             }
@@ -2966,7 +2943,7 @@ namespace Microsoft.Data.SqlClient
 
                 CheckMetaDataIsReady();
 
-                int copyLen = (values.Length < _metaData.visibleColumns) ? values.Length : _metaData.visibleColumns;
+                int copyLen = (values.Length < _metaData.VisibleColumnCount) ? values.Length : _metaData.VisibleColumnCount;
                 int maximumColumn = copyLen - 1;
 
                 SetTimeout(_defaultTimeoutMilliseconds);
@@ -2984,12 +2961,19 @@ namespace Microsoft.Data.SqlClient
                 for (int i = 0; i < copyLen; i++)
                 {
                     // Get the usable, TypeSystem-compatible value from the internal buffer
-                    values[_metaData.indexMap[i]] = GetValueFromSqlBufferInternal(_data[i], _metaData[i]);
+                    int fieldIndex = _metaData.GetVisibleColumnIndex(i);
+                    values[i] = GetValueFromSqlBufferInternal(_data[fieldIndex], _metaData[fieldIndex]);
 
                     // If this is sequential access, then we need to wipe the internal buffer
                     if ((sequentialAccess) && (i < maximumColumn))
                     {
-                        _data[i].Clear();
+                        _data[fieldIndex].Clear();
+                        if (fieldIndex > i && fieldIndex>0)
+                        {
+                            // if we jumped an index forward because of a hidden column see if the buffer before the 
+                            // current one was populated by the seek forward and clear it if it was
+                            _data[fieldIndex - 1].Clear();
+                        }
                     }
                 }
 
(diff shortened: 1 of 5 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
```diff
@@ -278,8 +278,8 @@ internal virtual SmiExtendedMetaData[] GetInternalSmiMetaData()
 
             if (null != metaData && 0 < metaData.Length)
             {
-                metaDataReturn = new SmiExtendedMetaData[metaData.visibleColumns];
-
+                metaDataReturn = new SmiExtendedMetaData[metaData.VisibleColumnCount];
+                int returnIndex = 0;
```
Wraith2, line 282: This code was creating an array of size visibleColumns and then iterating through total column count and setting at index of total column count. If a hidden column was ever present it would fall over with an out of range exception.

ErikEJ:
Does hidden columns refer to temporal data? If so, it should be possible to test?

```
CREATE SCHEMA History;
GO

ALTER TABLE InsurancePolicy
    ADD
        SysStartTime DATETIME2 GENERATED ALWAYS AS ROW START HIDDEN
            CONSTRAINT DF_SysStart DEFAULT SYSUTCDATETIME()
      , SysEndTime DATETIME2 GENERATED ALWAYS AS ROW END HIDDEN
            CONSTRAINT DF_SysEnd DEFAULT CONVERT(DATETIME2, '9999-12-31 23:59:59'),
        PERIOD FOR SYSTEM_TIME (SysStartTime, SysEndTime);
GO

ALTER TABLE InsurancePolicy
    SET (SYSTEM_VERSIONING = ON (HISTORY_TABLE = History.InsurancePolicy));

```

Wraith2:
I did find mention of them and history tables. I'm pretty sure they've existed for a while. Given that the code is pretty clearly wrong everywhere that deals with them I doubt anyone has ever really used this library around them.

Wraith2:
I tried a temporal table and the metadata for the hidden columns isn't passed back to the client at all, so it isn't that.

David-Engel:
@Wraith2 I figured out how to produce hidden columns. According to the [TDS spec](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-tds/58880b9f-381c-43b2-bf8b-0727a98c4f4c), it's for the [FOR BROWSE](https://docs.microsoft.com/en-us/sql/t-sql/queries/select-for-clause-transact-sql?view=sql-server-ver15#for-browse) option:

```
create table users (
	user_id int not null identity(1,1),
	first_name varchar(100) null,
	last_name varchar(100) null);
go

alter table users add constraint pk_users_user_id primary key (user_id);
go

insert into users (first_name,last_name) values ('Joe','Smith')
go
```
And then use the query:
```
select first_name, last_name, birth_date from users for browse
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserHelperClasses.cs
```diff
@@ -560,8 +563,8 @@ private _SqlMetaDataSet(_SqlMetaDataSet original)
         {
             this.id = original.id;
             // although indexMap is not immutable, in practice it is initialized once and then passed around
```
David-Engel, line 565: This comment is now out-of-date.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserHelperClasses.cs
```diff
@@ -598,10 +613,55 @@ internal int Length
...
+            {
+                if (_metaDataArray[index].IsHidden)
+                {
+                    hiddenColumnCount += 1;
+                }
+            }
+
+            if (hiddenColumnCount > 0)
+            {
+                int[] hiddenColumnMap = new int[Length - hiddenColumnCount];
```
David-Engel, line 650: Should this variable instead be named visibleColumnMap?

Wraith2: yes, change.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserHelperClasses.cs
Wraith2: removed, and removed the unneeded this prefixes around it too.

Wraith2:
> I think you are working with TVPs or XML somehow

I found this while using TVPs. they're one of the few things that escapes the standard metadata and end up in the extended metadata types. Unfortunately everyone is paying the price for this capability when they don't need to. 

I'll see if I can work up a test which exercises the hidden column feature in some way using FOR BROWSE or KeyInfo now I know that they're the way to see it in action. Well done on finding that.

cheenamalhotra:
@Wraith2 
Please add test case here when you find time so we can pick up this PR in next releases :)

Wraith2:
Test added and a minor bug fixed.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataReaderTest/DataReaderTest.cs
```diff
@@ -225,5 +226,81 @@ private static bool IsColumnBitSet(SqlConnection con, string selectQuery, int in
...
+	first_name varchar(100) null,
+	last_name varchar(100) null);
+
+alter table [{tempTableName}] add constraint pk_{tempTableName}_user_id primary key (user_id);
+
+insert into [{tempTableName}] (first_name,last_name) values ('Joe','Smith')
+";
+
+            string dataQuery = $@"select first_name, last_name from [{tempTableName}]";
+
```
johnnypham, line 253: Any value in adding a test for a result set with one PK column? Both fieldCount and visibleFieldCount would be 1.

Wraith2: If you select a single field that isn't the key you'll get a count of total:visible of 2:1, there isn't a way to select no fields and get the hidden key. The case of 3:2 and 2:1 are the same really so I didn't see any value in adding it. 

You could also say that we should test to see if we can get the hidden field at the start or the middle of the list but I've no way to know how to do that so I didn't try it.

johnnypham: Sorry, I meant if you executed `select user_id from [{tempTableName}]`.

Wraith2: it'll have the same basic result but the number will be 1 smaller. If you want I can change this one or add a second but it doesn't cover any more functionality to do so. Want me to add or change?

johnnypham: I've just started going through the changes but if it doesn't cover any more functionality, then never mind what I said.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
johnnypham: According to the CommandBehavior docs, hidden columns are appended to the result set so it seems like the exception would never have occurred in the original code.

Wraith2: Lucky. The code looked wrong and even if it was intentional that it worked it wasn't clear what it was doing was right. I can revert if you like but I think this version is better.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataReaderTest/DataReaderTest.cs
```diff
@@ -225,5 +226,81 @@ private static bool IsColumnBitSet(SqlConnection con, string selectQuery, int in
             }
             return columnSetPresent;
         }
+
+        [ConditionalFact(typeof(DataTestUtility), nameof(DataTestUtility.AreConnStringsSetup))]
+        public static void CheckHiddenColumns()
+        {
+            // hidden columns can be found by using CommandBehavious.KeyInfo or at the sql level
```
johnnypham, line 233:
```suggestion
            // hidden columns can be found by using CommandBehavior.KeyInfo or at the sql level
```

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataReaderTest/DataReaderTest.cs
```diff
@@ -225,5 +226,81 @@ private static bool IsColumnBitSet(SqlConnection con, string selectQuery, int in
...
+            // by using the FOR BROWSE option. These features return the column information requested and
+            // also include any key information required to be able to find the row containing the data
+            // requested. The additional key infomration is provided as hidden columns and can be seen using
```
johnnypham, line 236:
```suggestion
            // requested. The additional key information is provided as hidden columns and can be seen using
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
```diff
@@ -408,7 +410,7 @@ override public int VisibleFieldCount
                 {
                     return 0;
                 }
-                return (md.visibleColumns);
+                return (md.VisibleColumnCount);
```
johnnypham, line 413:
```suggestion
                return md.VisibleColumnCount;
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
johnnypham: Agreed, your changes look good to me.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
Wraith2: I didn't change this file...

Wraith2: Done.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataReaderTest/DataReaderTest.cs

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataReaderTest/DataReaderTest.cs
Wraith2: Done, Though I disagree on principle with the spelling of behaviour without a u it's a reference to the type so I did it with gritted teeth.

Wraith2:
anything else needed?

PR #528, opened 2020-04-16
perf: Rework ExecuteReaderAsync to minimize allocations

Wraith2:
Benchmarking against the usual DataAccessPerformance project I've got all the way down to the once per call allocations now. ExecuteDataReader uses callbacks delegates and closures which can be cached eliminated or minimized in many cases.

The AAsyncCallContext class is very similar to the one already used internally in SqlDataReader because the pattern of allocation and caching is similar. Once this PR is reviewed and merged I intend to go back and rebase the SqlDataReader internal classes on this new base class.

The changes here come in a number of patterns:
1) replace an instance bound delegate which must be allocated on each call with a static cached caller which has a new first parameter of the `this` type and bounces the call through a translation to an instance call again.
2) move continuations into state functions and pass state as parameters wherever possible to allow caching and avoid closures.
3) replace a compiler generated closures (which are always in method scope) with an explicitly and limited scope state object.

I also changed a void returning function which was altering a ref parameter to one which returns the new value because it's easier to understand what it's doing using the more familiar pattern.

some of the allocations eliminated are:
Capture2

there are some more off the bottom of the screenshot. 

The same work can be done for ExecuteScalar eventually.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/ProviderBase/DbConnectionFactory.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/ProviderBase/DbConnectionFactory.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System;
 using System.Data;
 using System.Data.Common;
 using System.Diagnostics;
@@ -15,6 +16,7 @@ namespace Microsoft.Data.ProviderBase
 {
     internal abstract partial class DbConnectionFactory
     {
+        private static readonly Action<Task<DbConnectionInternal>, object> s_tryGetConnectionCompletedContinuation = TryGetConnectionCompletedContinuation;
 
         internal bool TryGetConnection(DbConnection owningConnection, TaskCompletionSource<DbConnectionInternal> retry, DbConnectionOptions userOptions, DbConnectionInternal oldConnection, out DbConnectionInternal connection)
         {
@@ -82,25 +84,7 @@ namespace Microsoft.Data.ProviderBase
 
                             // now that we have an antecedent task, schedule our work when it is completed.
                             // If it is a new slot or a completed task, this continuation will start right away.
-                            newTask = s_pendingOpenNonPooled[idx].ContinueWith((_) =>
-                            {
-                                Transaction originalTransaction = ADP.GetCurrentTransaction();
-                                try
-                                {
-                                    ADP.SetCurrentTransaction(retry.Task.AsyncState as Transaction);
-                                    var newConnection = CreateNonPooledConnection(owningConnection, poolGroup, userOptions);
-                                    if ((oldConnection != null) && (oldConnection.State == ConnectionState.Open))
-                                    {
-                                        oldConnection.PrepareForReplaceConnection();
-                                        oldConnection.Dispose();
-                                    }
-                                    return newConnection;
-                                }
-                                finally
-                                {
-                                    ADP.SetCurrentTransaction(originalTransaction);
-                                }
-                            }, cancellationTokenSource.Token, TaskContinuationOptions.LongRunning, TaskScheduler.Default);
+                            newTask = CreateReplaceConnectionContinuation(s_pendingOpenNonPooled[idx], owningConnection, retry, userOptions, oldConnection, poolGroup, cancellationTokenSource);
 
                             // Place this new task in the slot so any future work will be queued behind it
                             s_pendingOpenNonPooled[idx] = newTask;
@@ -114,29 +98,11 @@ namespace Microsoft.Data.ProviderBase
                         }
 
                         // once the task is done, propagate the final results to the original caller
-                        newTask.ContinueWith((task) =>
-                        {
-                            cancellationTokenSource.Dispose();
-                            if (task.IsCanceled)
-                            {
-                                retry.TrySetException(ADP.ExceptionWithStackTrace(ADP.NonPooledOpenTimeout()));
-                            }
-                            else if (task.IsFaulted)
-                            {
-                                retry.TrySetException(task.Exception.InnerException);
-                            }
-                            else
-                            {
-                                if (!retry.TrySetResult(task.Result))
-                                {
-                                    // The outer TaskCompletionSource was already completed
-                                    // Which means that we don't know if someone has messed with the outer connection in the middle of creation
-                                    // So the best thing to do now is to destroy the newly created connection
-                                    task.Result.DoomThisConnection();
-                                    task.Result.Dispose();
-                                }
-                            }
-                        }, TaskScheduler.Default);
+                        newTask.ContinueWith(
+                            continuationAction: s_tryGetConnectionCompletedContinuation, 
+                            state: Tuple.Create(cancellationTokenSource, retry),
+                            scheduler: TaskScheduler.Default
+                        );
 
                         return false;
                     }
@@ -187,6 +153,63 @@ namespace Microsoft.Data.ProviderBase
             }
 
             return true;
+        }
+
+        private Task<DbConnectionInternal> CreateReplaceConnectionContinuation(Task<DbConnectionInternal> task, DbConnection owningConnection, TaskCompletionSource<DbConnectionInternal> retry, DbConnectionOptions userOptions, DbConnectionInternal oldConnection, DbConnectionPoolGroup poolGroup, CancellationTokenSource cancellationTokenSource)
+        {
+            return task.ContinueWith(
+                (_) =>
+                {
+                    Transaction originalTransaction = ADP.GetCurrentTransaction();
+                    try
+                    {
+                        ADP.SetCurrentTransaction(retry.Task.AsyncState as Transaction);
+                        var newConnection = CreateNonPooledConnection(owningConnection, poolGroup, userOptions);
+                        if ((oldConnection != null) && (oldConnection.State == ConnectionState.Open))
+                        {
+                            oldConnection.PrepareForReplaceConnection();
+                            oldConnection.Dispose();
+                        }
+                        return newConnection;
+                    }
+                    finally
+                    {
+                        ADP.SetCurrentTransaction(originalTransaction);
+                    }
+                },
+                cancellationTokenSource.Token,
+                TaskContinuationOptions.LongRunning,
+                TaskScheduler.Default
+            );
+        }
+
+        private static void TryGetConnectionCompletedContinuation(Task<DbConnectionInternal> task, object state)
+        {
+            Tuple<CancellationTokenSource, TaskCompletionSource<DbConnectionInternal>> parameters = (Tuple<CancellationTokenSource, TaskCompletionSource<DbConnectionInternal>>)state;
+            CancellationTokenSource source = parameters.Item1;
+            source.Dispose();
+
+            TaskCompletionSource<DbConnectionInternal> retryCompletionSrouce = parameters.Item2;
+
+            if (task.IsCanceled)
+            {
+                retryCompletionSrouce.TrySetException(ADP.ExceptionWithStackTrace(ADP.NonPooledOpenTimeout()));
+            }
+            else if (task.IsFaulted)
+            {
+                retryCompletionSrouce.TrySetException(task.Exception.InnerException);
+            }
+            else
+            {
+                if (!retryCompletionSrouce.TrySetResult(task.Result))
+                {
+                    // The outer TaskCompletionSource was already completed
+                    // Which means that we don't know if someone has messed with the outer connection in the middle of creation
+                    // So the best thing to do now is to destroy the newly created connection
+                    task.Result.DoomThisConnection();
+                    task.Result.Dispose();
+                }
+            }
         }
     }
 }
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/ProviderBase/DbConnectionFactory.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/ProviderBase/DbConnectionFactory.cs
@@ -189,19 +189,19 @@ namespace Microsoft.Data.ProviderBase
             CancellationTokenSource source = parameters.Item1;
             source.Dispose();
 
-            TaskCompletionSource<DbConnectionInternal> retryCompletionSrouce = parameters.Item2;
+            TaskCompletionSource<DbConnectionInternal> retryCompletionSource = parameters.Item2;
 
             if (task.IsCanceled)
             {
-                retryCompletionSrouce.TrySetException(ADP.ExceptionWithStackTrace(ADP.NonPooledOpenTimeout()));
+                retryCompletionSource.TrySetException(ADP.ExceptionWithStackTrace(ADP.NonPooledOpenTimeout()));
             }
             else if (task.IsFaulted)
             {
-                retryCompletionSrouce.TrySetException(task.Exception.InnerException);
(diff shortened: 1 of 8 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/AAsyncCallContext.cs
```diff
@@ -0,0 +1,81 @@
...
+{
+    // this class is a base class for creating derived objects that will store state for async operations
+    // avoiding the use of closures and allowing caching/reuse of the instances for frequently used async
+    // calls
+    //
+    // DO derive from this and seal and your class
+    // DO add additional fields or properties needed for the async operation and then override Clear to zero them
+    // DO override AfterClear and use the owner parameter to return the object to a cache location if you have one, this is the purpose of the method
+    // CONSIDER creating your own Set method that calls the base Set rather than providing a parameterized ctor, it is friendlier to caching
+    // DO NOT use this class after Dispose has been called. It will not throw ObjectDisposedException but it will be a cleared object
```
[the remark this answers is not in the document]
Wraith2: It's a cache class so the idea is to keep it around and re-use it. Dispose clears the state it contains and puts it into a blank state ready for re-use so you shouldn't use it expecting it to still contain async state. The only thing that should be done with it after dispose is to store it for another use or discard it to the GC.

Wraith2:
The [DataAccessPerformance](https://github.com/aspnet/DataAccessPerformance) project emulates the [TechEmpower fortunes benchmark](https://www.techempower.com/benchmarks/#section=data-r17&hw=ph&test=fortune) and I was pointed to it as a good performance based benchmarking solution. If you look at my history of pull requests here and in corefx you'll see I use it a lot and have been gradually improving performance and memory usage using it.

The screenshot is from dotMemory which is a memory profiler. I highlighted the items that I've affected in red using paint

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/ProviderBase/DbConnectionFactory.cs
```diff
@@ -188,5 +154,62 @@ internal bool TryGetConnection(DbConnection owningConnection, TaskCompletionSour
...
+                        ADP.SetCurrentTransaction(originalTransaction);
+                    }
+                },
+                cancellationTokenSource.Token,
+                TaskContinuationOptions.LongRunning,
+                TaskScheduler.Default
+            );
+        }
+
+        private static void TryGetConnectionCompletedContinuation(Task<DbConnectionInternal> task, object state)
```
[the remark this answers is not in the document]
Wraith2: Adding try-catch can introduce control flow changes and i've learned not to make changes to the async code in this library without being totally sure i understand what the implications might be. So yes it might be a good idea to try-catch but the original code didn't have it and i want to preserve the control flow at the moment.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/ProviderBase/DbConnectionFactory.cs
```diff
@@ -188,5 +154,62 @@ internal bool TryGetConnection(DbConnection owningConnection, TaskCompletionSour
...
+        {
+            Tuple<CancellationTokenSource, TaskCompletionSource<DbConnectionInternal>> parameters = (Tuple<CancellationTokenSource, TaskCompletionSource<DbConnectionInternal>>)state;
+            CancellationTokenSource source = parameters.Item1;
+            source.Dispose();
+
+            TaskCompletionSource<DbConnectionInternal> retryCompletionSrouce = parameters.Item2;
+
+            if (task.IsCanceled)
+            {
+                retryCompletionSrouce.TrySetException(ADP.ExceptionWithStackTrace(ADP.NonPooledOpenTimeout()));
```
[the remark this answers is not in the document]
Wraith2: The way I read it. If it is in RanToCompletion then there is no need to set the exception, if it's Faulted the exception already contains the error that cause the fault so we don't want to override it, there is no other way to cause cancelled than to set the cancellation token on the task like we're doing because is another cancelation token is used it'll throw if checked properly and that'll end up here as a Faulted state.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/AAsyncCallContext.cs
```diff
@@ -0,0 +1,81 @@
...
+        protected virtual void Clear()
+        {
+        }
+
+        /// <summary>
+        /// override this method to do work after the instance has been totally blanked, intended for cache return etc
+        /// </summary>
+        protected virtual void AfterCleared(TOwner owner)
+        {
+
```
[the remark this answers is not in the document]
Wraith2: fixed.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/AAsyncCallContext.cs
```diff
@@ -0,0 +1,81 @@
...
+        {
+            _source = null;
+            _owner = default;
+            IDisposable copyDisposable = _disposable;
+            _disposable = null;
+            copyDisposable?.Dispose();
+        }
+
+        /// <summary>
+        /// overrride this method to cleanup instance data before ClearCore is called which will blank the base data
```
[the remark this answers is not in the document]
Wraith2: fixed.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/AAsyncCallContext.cs
```diff
@@ -0,0 +1,81 @@
...
+using System;
+using System.Threading.Tasks;
+
+namespace Microsoft.Data.SqlClient
+{
+    // this class is a base class for creating derived objects that will store state for async operations
+    // avoiding the use of closures and allowing caching/reuse of the instances for frequently used async
+    // calls
+    //
+    // DO derive from this and seal and your class
```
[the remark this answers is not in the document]
Wraith2: fixed

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/AAsyncCallContext.cs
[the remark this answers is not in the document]
Wraith2: I've clarified the comment.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/ProviderBase/DbConnectionFactory.cs
```diff
@@ -188,5 +154,62 @@ internal bool TryGetConnection(DbConnection owningConnection, TaskCompletionSour
 
             return true;
         }
+
+        private Task<DbConnectionInternal> CreateReplaceConnectionContinuation(Task<DbConnectionInternal> task, DbConnection owningConnection, TaskCompletionSource<DbConnectionInternal> retry, DbConnectionOptions userOptions, DbConnectionInternal oldConnection, DbConnectionPoolGroup poolGroup, CancellationTokenSource cancellationTokenSource)
+        {
+            return task.ContinueWith(
```
roji, line 160: Small comment @Wraith2: ContinueWith generally performs much worse than a simple async method with await (not to mention that's it's more brittle wrt exception handling etc.). I can post some benchmarking if you want.

roji: Though this may be totally justified if TaskContinuationOptions.LongRunning is important here - just making the general remark.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnection.cs
```diff
@@ -1864,14 +1873,22 @@ private static void ChangePassword(string connectionString, SqlConnectionString
         // this only happens once per connection
         // SxS: using named file mapping APIs
 
-        internal void RegisterForConnectionCloseNotification<T>(ref Task<T> outerTask, object value, int tag)
+        internal Task<T> RegisterForConnectionCloseNotification<T>(Task<T> outerTask, object value, int tag)
```
roji, line 1876: This could be a candidate for rewriting with simple async/await

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/ProviderBase/DbConnectionFactory.cs
Wraith2: Language supported async await would be easier to understand as well. 
Whoever wrote the original code had some reason to use imperative constructs instead of language support but I've no idea what it was.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnection.cs
Wraith2: @David-Engel @cheenamalhotra I can give this a try in the PR if you like or we can leave it for a later date, what do you think?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/ProviderBase/DbConnectionFactory.cs
ErikEJ: I vote for prioritizing 1 - it is currently a quite confusing codebase, which makes community contributions harder than the should be.

Wraith2: working on it bit by bit. https://github.com/dotnet/SqlClient/pull/625 The easy part is the identical files. The complex part is the files which have opposing or complex changes.

roji: Sure thing, thing kind of cleanup can be done separately of course. I'd suggest that if ContinueWith continuations are being rewritten for unrelated reasons, it may be a good opportunity to just do that with async/await, but of course that's up to the team and you.

I also totally agree that merging the netfx/netcore codebases is a high-priority task.

DavoudEshtehari:
Can I ask you to produce some information to illustrate how do these changes affect the performance?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnection.cs
cheenamalhotra: Are there more similar cases that we would like to address?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/ProviderBase/DbConnectionFactory.cs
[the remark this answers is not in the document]
johnnypham: typo: retryCompletion**Srouce**

Wraith2:
Using Northwind and the benchmark code:

```csharp

		[Benchmark]
		public async Task<int> AsyncReadOrders()
		{
			int max = 0;

			using (SqlDataReader reader = await command.ExecuteReaderAsync())
			{
				while (await reader.ReadAsync())
				{
					int value = reader.GetInt32(0);
					if (value > max)
					{
						max = value;
					}
				}
			}
			
			return max;
		}
```

|          Method |     Mean |   Error |  StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|---------------- |---------:|--------:|--------:|------:|------:|------:|----------:|
| AsyncReadOrders branch | 353.9 μs | 6.93 μs | 9.71 μs |     - |     - |     - |   2.15 KB |
| AsyncReadOrders master | 349.1 μs | 6.92 μs | 8.24 μs |     - |     - |     - |   2.59 KB |

It's a really tiny benchmark as you can see from the times. It simply gets some ints using a data reader and reads them. The main push is the allocated memory. Very little of those allocations are user data they're task machinery and library internals. The reduction you see is delegates which are removed by caching static versions of them and some async closures which are removed by using direct state objects to avoid capture.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnection.cs
Wraith2: The original code tooks a ref Task<T> and added a continuation to it then returned the new wrapper task in the same variable. It doesn't do any awaiting and if the caller was rewritten to use language async then this function would be totally rewritten or removed as well. It'd prefer not to touch it further in this PR and work on it at a stage when a larger scale transition from contiunation to langauge async is being done.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/ProviderBase/DbConnectionFactory.cs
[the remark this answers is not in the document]
Wraith2: Fixed.

johnnypham:
I was able to verify the difference in allocations but am I correct in thinking that the benefits would mostly be lost when returning larger result sets for each call to ExecuteReaderAsync?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
```diff
@@ -1961,9 +1994,55 @@ private SqlDataReader EndExecuteReaderInternal(IAsyncResult asyncResult)
...
+        }
+
+        private static void CleanupExecuteReaderAsyncCallback(Task<SqlDataReader> task)
+        {
+            ExecuteReaderAsyncCallContext context = (ExecuteReaderAsyncCallContext)task.AsyncState;
+            SqlCommand command = context.Command;
+            Guid operationId = context.OperationID;
+            TaskCompletionSource<SqlDataReader> source = context.TaskCompletionSource;
+            context.Dispose();
+
```
johnnypham, line 2044: Why not call Dispose inside CleanupExecuteReaderAsync?

Wraith2:
The benefits wouldn't be lost they'd just be less of the total memory allocated if you do a lot of reading. If people are making an effort to do sensible things, chunky not chatty, then this is just a nice to have, if they're being bad at sql and doing chatty calls then this is nicer to have. It's never fundamental. 

In an ideal situation I'd like the library to only allocate memory for the results that a user requested, that isn't practical but if there is benefit to users for relatively small increase in complexity of the library it's possibly worth taking. The reason I started working on this library is that I had written some code using the new pipelines and span apis in netcore and had a really lean fast project and then when I started trying to get that data in and out of a database using this library obliterated all the careful performance work I'd done.  So yes these are smaller gains but they're still of value if you want to support high performance scenarios.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
Wraith2: Cleaning it up here returns it to the cache earlier. It also releases any resources being held by the context object (buffers etc).

cheenamalhotra:
@Wraith2 seems there are conflicts with recent merge of PR #499 - please resolve when you have time.

Wraith2:
merged and fixed up.

PR #534, opened 2020-04-20
Standardize Connection String properties for enhanced user experience

cheenamalhotra:
Some connection string properties are breaking the theme of connection string properties for SqlClient and end users need to track which connection string properties are spelled differently. This change standardizes the connection properties for better user experience.

The below connection properties are now updated with synonyms created for backwards compatibility:

- Application Intent _(previously `ApplicationIntent`)_
- Connect Retry Count _(previously `ConnectRetryCount`)_
- Connect Retry Interval _(previously `ConnectRetryInterval`)_
- Pool Blocking Period _(previously `PoolBlockingPeriod`)_
- Multiple Active Result Sets _(previously `MultipleActiveResultSets`)_
- Multiple Subnet Failover _(previously `MultiSubnetFailover`)_
- Transparent Network IP Resolution _(previously `TransparentNetworkIPResolution`)_
- Trust Server Certificate _(previously `TrustServerCertificate`)_

Since this is not a breaking change, client applications are not expected to be impacted with this change.

P.S. Also made some improvements on documentation.


src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -682,11 +682,11 @@ public sealed partial class SqlConnectionStringBuilder : System.Data.Common.DbCo
         [System.ComponentModel.RefreshPropertiesAttribute(System.ComponentModel.RefreshProperties.All)]
         public int MinPoolSize { get { throw null; } set { } }
         /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnectionStringBuilder.xml' path='docs/members[@name="SqlConnectionStringBuilder"]/MultipleActiveResultSets/*'/>
-        [System.ComponentModel.DisplayNameAttribute("MultipleActiveResultSets")]
+        [System.ComponentModel.DisplayNameAttribute("Multiple Active Result Sets")]
         [System.ComponentModel.RefreshPropertiesAttribute(System.ComponentModel.RefreshProperties.All)]
         public bool MultipleActiveResultSets { get { throw null; } set { } }
         /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnectionStringBuilder.xml' path='docs/members[@name="SqlConnectionStringBuilder"]/MultiSubnetFailover/*'/>
-        [System.ComponentModel.DisplayNameAttribute("MultiSubnetFailover")]
+        [System.ComponentModel.DisplayNameAttribute("Multi Subnet Failover")]
```
JRahnama, line 689: Just from curiosity, is there any specific reason we dont add the System.ComponentModel as a using statement on the top of the file and simplify these attributes?

cheenamalhotra: I think this has always been there, we can simplify in a separate PR if you have a proposition.

jeremy-waguet:
@cheenamalhotra 
This PR supposed to be backward compatible with the non spaced names but on my project ApplicationIntent=ReadWrite produces and 'unknown keyword exception applicationintent' whereas Application Intent=ReadWrite work !?? The other properties appears to work (I am using only TrustServerCertificate +MultiSubnetFailover)

cheenamalhotra:
Hi @jeremy-waguet 

Thanks for trying out and capturing the issue, our tests covered this property but was limited to connection string builder, hence this got missed. https://github.com/dotnet/SqlClient/pull/585 will fix this issue.

Thanks!

VladimirKhvostov:
This change introduced huge compat issues. System.Data.Client does not understand connection strings generated by Microsoft.Data.Client.

PR #535, opened 2020-04-21
Remove Diagnostic Source + Configuration Manager from .NET Standard DLL as it's not supported

cheenamalhotra:
For enhanced experience with .NET Standard 2.0, below changes are needed:

1. Remove dependency on [DiagnosticSource](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.diagnosticsource?view=netcore-3.1) and [DiagnosticListener](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.diagnosticlistener?view=netcore-3.1) classes as they are not available for .NET Standard 2.0.

2. Remove dependency on [ConfigurationSection](https://docs.microsoft.com/en-us/dotnet/api/system.configuration.configurationsection?view=netframework-4.8) as this is also not available in .NET Standard 2.0.

P.S. Tests will be added in a separate PR (post-merge) as that would require rework of existing tests + adding new .NET Standard tests.

Dev NuGet Package for testing the changes:
[Microsoft.Data.SqlClient.2.0.0-dev-pr535.nupkg.zip](https://github.com/dotnet/SqlClient/files/4542423/Microsoft.Data.SqlClient.2.0.0-dev-pr535.nupkg.zip)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
```diff
@@ -460,8 +462,12 @@ internal SqlStatistics Statistics
...
                 if (null != _activeConnection)
                 {
-                    if (_activeConnection.StatisticsEnabled ||
-                        _diagnosticListener.IsEnabled(SqlClientDiagnosticListenerExtensions.SqlAfterExecuteCommand))
+                    if (_activeConnection.StatisticsEnabled
+#if netcoreapp
+                        || _diagnosticListener.IsEnabled(SqlClientDiagnosticListenerExtensions.SqlAfterExecuteCommand))
+#else
+                        )
+#endif
```
Wraith2, line 470: The compiler directives feel bad when there's a netcoreapp version of this file already split out in the PR. Just checking that this is because it's prohibitively difficult to do this any other way, if you've deliberately chosen to use both then it's ok.

cheenamalhotra: I just updated the code to work with platform specific implementations instead.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlAuthenticationProviderManager.NetCoreApp.cs
```diff
@@ -0,0 +1,114 @@
...
+                    {
+                        var providerType = Type.GetType(providerSettings.Type, true);
+                        provider = (SqlAuthenticationProvider)Activator.CreateInstance(providerType);
+                    }
+                    catch (Exception e)
+                    {
+                        throw SQL.CannotCreateAuthProvider(authentication.ToString(), providerSettings.Type, e);
+                    }
+                    if (!provider.IsSupported(authentication))
+                        throw SQL.UnsupportedAuthenticationByProvider(authentication.ToString(), providerSettings.Type);
```
karinazhou, line 70: Maybe add curly braces for readability?

cheenamalhotra: Sure, I just picked it raw from another class.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDiagnosticListener.NetCoreApp.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.Diagnostics;
```
karinazhou, line 1: Please add license header

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDiagnosticListener.NetStandard.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System;
```
karinazhou, line 1: Please add license header

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -367,8 +367,7 @@ internal SqlConnectionTimeoutErrorInternal TimeoutErrorInternal
                 SessionData reconnectSessionData = null,
                 bool applyTransientFaultHandling = false,
                 string accessToken = null,
-                DbConnectionPool pool = null,
-                SqlAuthenticationProviderManager sqlAuthProviderManager = null
+                DbConnectionPool pool = null
```
David-Engel, line 370: I'm not sure about these and the changes in TdsParser.cs. Why was `SqlAuthenticationProviderManager` previously allowed to be passed in instead of just referencing the Instance? I don't see how this change was necessary. It almost looks like just removing unneeded code. But I feel like I'm missing something.

cheenamalhotra: This is just unused parameter, no calling method sets it and it's always passed as null.

PR #543, opened 2020-04-27
Fix | Avoid enlistment of pooled connection in aborted transaction

### LINKED ISSUE #237: SqlException: The PROMOTE TRANSACTION request failed because there is no local transaction active

cheenamalhotra:
Fixes #237 by dooming the connection when transaction ends externally (re-created PR #495) . 

As of now, this fix is promising and does not send connection back to connection pool, hence avoiding enlistment in aborted transaction.

Alternatively, if we want to send connection back to pool, it needs to be in same state as any other connection in pool with no aborted transaction reference, but that doesn't seem likely to be achieved without complex changes in the driver that would also be a risky attempt. Investigations are in progress that will bring back same performance, but in order to address the issue on priority, I would propose this fix in the driver for end user application consistency.


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDelegatedTransaction.cs
```diff
@@ -137,260 +137,291 @@ public byte[] Promote()
...
-                        // eventually rolled back.
-                        // VSTS 144562: doom the connection while having the lock on it to prevent race condition with "Transaction Ended" Event
-                        connection.DoomThisConnection();
-                    }
-                    catch (InvalidOperationException e)
-                    {
-                        promoteException = e;
-                        connection.DoomThisConnection();
-                    }
+                //Throw exception only if Transaction is till active and not yet aborted.
```
David-Engel, line 210:
```suggestion
                //Throw exception only if Transaction is still active and not yet aborted.
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDelegatedTransaction.cs
```diff
@@ -137,260 +137,291 @@ public byte[] Promote()
...
+            if (null != connection)
+            {
+                SqlConnection usersConnection = connection.Connection;
+                SqlClientEventSource.Log.TraceEvent("<sc.SqlDelegatedTransaction.SinglePhaseCommit|RES|CPOOL> {0}, Connection {1}, committing transaction.", ObjectID, connection.ObjectID);
+                RuntimeHelpers.PrepareConstrainedRegions();
+                try
                 {
-                    Exception commitException;
-                    lock (connection)
+                    // If the connection is dooomed, we can be certain that the
```
David-Engel, line 324:
```suggestion
                    // If the connection is doomed, we can be certain that the
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDelegatedTransaction.cs
```diff
@@ -413,6 +444,8 @@ internal void TransactionEnded(Transaction transaction)
                         _active = false;
                         _connection = null;
                     }
+                    // Safest approach is to doom this connection, whose transaction has been aborted externally.
```
David-Engel, line 447:
```suggestion
                    // Safest approach is to doom this connection, whose transaction has been aborted externally.
                    // If we want to avoid dooming the connection for performance, state needs to be properly restored. (future TODO)
```

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlDelegatedTransaction.cs
```diff
@@ -167,122 +166,130 @@ public Byte[] Promote()
...
-                throw;
-            }
-            catch (System.Threading.ThreadAbortException e)
-            {
-                usersConnection.Abort(e);
-                throw;
-            }
 
-            if (promoteException != null)
+                //Throw exception only if Transaction is till active and not yet aborted.
```
David-Engel, line 257:
```suggestion
                //Throw exception only if Transaction is still active and not yet aborted.
```

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlDelegatedTransaction.cs
```diff
@@ -504,6 +525,8 @@ internal void TransactionEnded(SysTx.Transaction transaction)
                         _active = false;
                         _connection = null;
                     }
+                    // Safest approach is to doom this connection, whose transaction has been aborted externally.
```
David-Engel, line 528:
```suggestion
                    // Safest approach is to doom this connection, whose transaction has been aborted externally.
                    // If we want to avoid dooming the connection for performance, state needs to be properly restored. (future TODO)
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDelegatedTransaction.cs
```diff
@@ -421,7 +454,7 @@ internal void TransactionEnded(Transaction transaction)
         private SqlInternalConnection GetValidConnection()
         {
             SqlInternalConnection connection = _connection;
-            if (null == connection)
+            if (null == connection && Transaction.TransactionInformation.Status != TransactionStatus.Aborted)
```
David-Engel, line 457: This new logic seems odd to me. So if the connection is null but we have an active transaction, we should now return a null connection instead of throwing?

cheenamalhotra: If the connection is null but we have an "Aborted" transaction, then we return a null connection.
If the connection is null but we have an "Active" transaction, then we throw exception.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlDelegatedTransaction.cs
```diff
@@ -331,156 +338,170 @@ public void Rollback(SysTx.SinglePhaseEnlistment enlistment)
...
-                {
-#endif //DEBUG
-                    // If the connection is dooomed, we can be certain that the
-                    // transaction will eventually be rolled back, and we shouldn't
-                    // attempt to commit it.
-                    if (connection.IsConnectionDoomed)
                     {
-                        lock (connection)
+#endif //DEBUG
+                        // If the connection is dooomed, we can be certain that the
```
David-Engel, line 394: I missed one:
```suggestion
                        // If the connection is doomed, we can be certain that the
```

ramonsmits:
@cheenamalhotra will this fix also be backported?

cheenamalhotra:
We are going to backport the fix to System.Data.SqlClient in .NET Core, while for .NET Framework, @saurabh500 can confirm if this could be considered?

ramonsmits:
The original issue was found in .net framework. It seems to me this needs backporting to .net framework too.

PR #548, opened 2020-05-01
Consistency updates

cheenamalhotra:
- Remove ConfigurationManager dependency from nuspec
- Standardize App Context switch name for "UseManagedNetworkingOnWindows"

### DIFF
--- a/BUILDGUIDE.md
+++ b/BUILDGUIDE.md
@@ -177,7 +177,7 @@ Tests can be built and run with custom Target Frameworks. See the below examples
 
 Managed SNI can be enabled on Windows by enabling the below AppContext switch:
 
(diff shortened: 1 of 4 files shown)


David-Engel:
> If you are updating the switch name, you'll also need to update the BUILDGUIDE and DataTestUtility.cs. Then I assume the pipeline will need to be updated, too.

Looks like you just did that. :)

cheenamalhotra:
> Be sure to think about what the pipelines should look like for building slightly older code, too. Since this switch just came out in 2.0 previews, maybe we don't need to support the previous name.

Pipelines are run based on `config.json` property `UseManagedSNIOnWindows` which is untouched, and AppContext name is taken care in DataTestUtility, so we are good :)

PR #547, opened 2020-04-30
Revert change that made `SqlDataReader.ReadAsync()` non-blocking

cheenamalhotra:
This PR reverts https://github.com/dotnet/corefx/pull/26595 and #490 (attempted to fix MARS TDS header issues introduced with https://github.com/dotnet/corefx/pull/26595) with additional tests.

This is an alternate solution to fix #542 but due to #544 and #545 failing intermittently on Unix w. Managed SNI, it doesn't seem an appropriate fix either. 

This revert is likely to hit performance by re-introducing old behavior of SqlDataReader.ReadAsync() method to run synchronously, and will block the calling thread until data is fed from SQL Server. But given the side-effects of the original change, reverting seems most appropriate at this point and a proper investigation would be needed to be done for original performance issue.

PR #550, opened 2020-05-01
[Release 1.1] Revert ReadAsync changes that caused issues

cheenamalhotra:
Ports #547 to 1.1-servicing

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
@@ -2861,8 +2861,8 @@ namespace Microsoft.Data.SqlClient
                 }
             }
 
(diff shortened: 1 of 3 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -2855,11 +2861,10 @@ private bool TryProcessDone(SqlCommand cmd, SqlDataReader reader, ref RunBehavio
                 }
             }
 
-            // _attentionSent set by 'SendAttention'
             // _pendingData set by e.g. 'TdsExecuteSQLBatch'
```
DavoudEshtehari, line 2864: I do not see `_pendingData `in this context.

cheenamalhotra: I guess the comments were not updated to reflect variable rename

Issue #553, opened 2020-05-06
Build failure on Linux via dotnet build


roji:
When trying to do a simple dotnet build in src/Microsoft.Data.SqlClient/netcore/src, I get the error messages below (this also happens when doing dotnet msbuild at the root). This seems to be because [TargetsUnix gets set to true in Directory.Build.props](https://github.com/dotnet/SqlClient/blob/master/src/Directory.Build.props#L7) (since I'm running on Linux), but in [Microsoft.Data.SqlClient.csproj](https://github.com/dotnet/SqlClient/blob/master/src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj#L6) OSGroup defaults to Windows_NT, which leads to setting TargetsWindows to true as well. This leads to both Windows and Unix file versions getting compiled.

As an aside, I'm wondering about the need to build a Unix vs. Windows version of the project - at the end of the day only one of them goes into the nupkg (presumably the Windows one). Is there some reason not to just compile with the Windows support, and have it throw at runtime on Unix (as it already does)? In other words, removing the conditional compilation based on platform would likely simplify various things here.

```
CSC : error CS2001: Source file '/home/roji/projects/sqlclient/src/Microsoft.Data.SqlClient/netcore/src/Common/src/CoreLib/Interop/Windows/kernel32/Interop.FileTypes.cs' could not be found. [/home/roji/projects/sqlclient/src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj]
CSC : error CS2001: Source file '/home/roji/projects/sqlclient/src/Microsoft.Data.SqlClient/netcore/src/Common/src/CoreLib/Interop/Windows/kernel32/Interop.GetFileType_SafeHandle.cs' could not be found. [/home/roji/projects/sqlclient/src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj]
CSC : error CS2001: Source file '/home/roji/projects/sqlclient/src/Microsoft.Data.SqlClient/netcore/src/Common/src/CoreLib/Interop/Windows/kernel32/Interop.SetThreadErrorMode.cs' could not be found. [/home/roji/projects/sqlclient/src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj]
[...]

PR #554, opened 2020-05-06
Fix filename case in csproj

roji:
Fails the build under Linux


cheenamalhotra:
I don't mind the change here, but I was able to compile in Linux docker images.
Does it come across as a warning or error and which OS/Version of Linux did you try on?

roji:
Yes, this was happening while I was fooling around with #553, it may not be an actual problem. Up to you.

PR #557, opened 2020-05-08
perf: Reduce tracing related allocations from TVP write

Wraith2:
There is a case where extended metadata trace information is formatted before being passed into an event tracing method which then may not use the string. 
I chose not to try and make the tracing code more efficient because it will almost never be used. I've added precheck to see if advanced logging is enabled and passed that as a parameter to the function logging the information because it avoids a per-loop invocation of the event trace flag check.

|              Method |     Mean |   Error |  StdDev |     Gen 0 | Gen 1 | Gen 2 | Allocated |
|-------------------- |---------:|--------:|--------:|----------:|------:|------:|----------:|
| SyncWriteTableParam master | 191.0 ms | 1.69 ms | 1.41 ms | 1000.0000 |     - |     - |  12.05 MB |
| SyncWriteTableParam branch | 185.2 ms | 0.42 ms | 0.33 ms |     - |     - |     - | 313.33 KB |

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
@@ -9656,7 +9656,7 @@ namespace Microsoft.Data.SqlClient
             if (advancedTraceIsOn)
             {
                 var sendDefaultValue = sendDefault ? 1 : 0;
-                SqlClientEventSource.Log.AdvancedTraceEvent("<sc.TdsParser.WriteSmiParameter|ADV> {0}, Sending parameter '{1}', default flag={2}, metadata:{3}", ObjectID, param.ParameterName, sendDefaultValue, metaData);
+                SqlClientEventSource.Log.AdvancedTraceEvent("<sc.TdsParser.WriteSmiParameter|ADV> {0}, Sending parameter '{1}', default flag={2}, metadata:{3}", ObjectID, param.ParameterName, sendDefaultValue, metaData.TraceString(3));
             }
 
             //
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
@@ -8872,6 +8872,8 @@ namespace Microsoft.Data.SqlClient
                         // Stream out parameters
                         int parametersLength = rpcext.userParamCount + rpcext.systemParamCount;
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -9651,8 +9653,11 @@ private void WriteSmiParameter(SqlParameter param, int paramIndex, bool sendDefa
                 typeCode = MetaDataUtilsSmi.DetermineExtendedTypeCodeForUseWithSqlDbType(metaData.SqlDbType, metaData.IsMultiValued, value, null);
             }
 
-            var sendDefaultValue = sendDefault ? 1 : 0;
-            SqlClientEventSource.Log.AdvancedTraceEvent("<sc.TdsParser.WriteSmiParameter|ADV> {0}, Sending parameter '{1}', default flag={2}, metadata:{3}", ObjectID, param.ParameterName, sendDefaultValue, metaData.TraceString(3));
+            if (advancedTraceIsOn)
```
cheenamalhotra, line 9656: Why not `if (SqlClientEventSource.Log.IsAdvancedTraceOn())` directly?

Wraith2: This call is inside a loop, once for each parameter. I didn't see the point repeatedly checking since it's highly unlikely to have the trace flag turned on in the middle. It can be cheeked directly if you prefer.

cheenamalhotra: If there's no perf impact, I'd rather prefer keeping it straightforward.

Wraith2: The perf impact is whatever it takes to call `SqlClientEventSource.Log.IsAdvancedTraceOn()` multiplied by the number of parameters. I'm not sure how the trace flags are actually checked or what overhead they have. There's some perf impact to going the same thing N times instead of once and storing the result but I can't say I know what it is or whether it's important. Based on that, do you want me to make the change?

David-Engel: I lean a bit towards keeping it as-is for the perf benefit in this case... even though it's probably pretty minor.

JRahnama: I approved this PR. I agree with David. The effect on the perf is pretty minor.

PR #570, opened 2020-05-20
Native SNI Improvements

cheenamalhotra:
This PR brings below changes:

- **For .NET Core and .NET Standard targets** (on Windows)
**We now introduce `Microsoft.Data.SqlClient.SNI.runtime` as new dependency to replace the old `runtime.native.System.Data.SqlClient.SNI`.** Below platforms are supported by Microsoft.Data.SqlClient.SNI.runtime:
  - ARM
  - ARM64
  - x64
  - x86

- **For .NET Framework targets**
The `SNI.dll` files previously downloaded to _x64_ and _x86_ folders are now renamed to `Microsoft.Data.SqlClient.SNI.x64.dll` and `Microsoft.Data.SqlClient.SNI.x86.dll` to support respective platforms.

- **Test Expansions**
The SqlClient test suite now allows executing same tests for below reference types:
  - Project (default)
    - Build and run tests with _Microsoft.Data.SqlClient_ as Project Reference
  - Package
    - Build and run tests with _Microsoft.Data.SqlClient_ as Package Reference with configured "TestMicrosoftDataSqlClientVersion" in "Versions.props" file.
  - NetStandard
    - Build and run tests with _Microsoft.Data.SqlClient_ as Project Reference via .NET Standard Library to validate .NET Standard support.
  - NetStandardPackage
    - Build and run tests with _Microsoft.Data.SqlClient_ as Package Reference via .NET Standard Library to validate .NET Standard support with configured "TestMicrosoftDataSqlClientVersion" in "Versions.props" file.


Wraith2:
> Below platforms are supported by Microsoft.Data.SqlClient.SNI.runtime 

It might be worth noting at release that this is just for windows os, all non-windows will use managed.

tools/specs/Microsoft.Data.SqlClient.nuspec
```diff
@@ -99,75 +99,75 @@ When using NuGet 3.x this package requires at least version 3.4.</description>
     <file src="..\icons\dotnet.png" target="" />
 
     <!-- ref NetFx -->
-    <file src="..\..\bin\Windows_NT\$Configuration$\Microsoft.Data.SqlClient\ref\net46\Microsoft.Data.SqlClient.dll" target="ref\net46\" exclude="" />
-    <file src="..\..\bin\Windows_NT\$Configuration$\Microsoft.Data.SqlClient\ref\net46\Microsoft.Data.SqlClient.pdb" target="ref\net46\" exclude="" />
-    <file src="..\..\bin\Windows_NT\$Configuration$\Microsoft.Data.SqlClient\ref\net46\Microsoft.Data.SqlClient.xml" target="ref\net46\" exclude="" />
+    <file src="..\..\artifacts\Project\bin\Windows_NT\$Configuration$\Microsoft.Data.SqlClient\ref\net46\Microsoft.Data.SqlClient.dll" target="ref\net46\" exclude="" />
```
DavoudEshtehari, line 102: Can I ask you what does the reason of adding the `Project` subfolder? After the build, I saw just this folder inside the `artifacts`.

cheenamalhotra: For consistency in architecture.
All binaries are moved inside it.

IanKemp:
Will this be available in both the legacy `System.Data.SqlClient` and `Microsoft.Data.SqlClient` packages, or only the latter?

cheenamalhotra:
Hi @IanKemp

These changes are applicable only to Microsoft.Data.SqlClient latest preview (2.0.0-preview4) and future versions.

irontoby:
@cheenamalhotra sorry for the late response on this but can you confirm that with the switch to this package, that using Microsoft.Data.SqlClient 2.0.0 or later will no longer require the Microsoft Visual C++ Redistributable to be installed on Windows? It looks like that's the case but I can't find any documentation on that specifically.

cheenamalhotra:
Hi @irontoby 

We removed the dependency notes from Microsoft.Data.SqlClient NuGet [v1.1.1](https://www.nuget.org/packages/Microsoft.Data.SqlClient/1.1.1) onwards. But since the SNI is closed source and is made available to public with driver releases (as it's not meant for direct reference), we probably didn't publish SNI related changes.
I will discuss internally to see how can we make it more visible.

irontoby:
@cheenamalhotra thanks! Understand your reasoning, but this has definitely been a bit of a frustrating experience for us. The need for having to install the redistributable was not obvious at first, since we weren't even using `Microsoft.Data.SqlClient` directly; we upgraded to EF Core 3.0 which depends on it.

So it worked fine on devs' machines but broke on the server & took a good while to figure out what was going wrong. So then we had to update our provisioning tools to install that runtime on all servers.

Anyway thanks for the feedback; based on this it sounds like we can remove the runtime from our update process as long as we directly reference `Microsoft.Data.SqlClient` 2.0 or higher.

Issue #165, opened 2019-08-27
hung at connection.open


omhappy:
I have an app in Azure AKS container. When the app call connection.open to connect to Azure Sql, it hung at ReadInternal method of class SslOverTdsStream occasionally. 
We create a dump and find that the ReadInternal method doesn't check the return value of _stream.Read.
if the socket is closed, the return value is 0, then the method will be an infinite loop.

omhappy:
I use .Net Core 2.1 the code is 
https://github.com/dotnet/corefx/blob/release/2.1/src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SslOverTdsStream.cs

oldzhu:
This issue occurred frequently in one of our customer environment. when the issue occurs, we could see the high CPU utilization of the thread which call stack is like the below:

(lldb) clrstack
OS Thread Id: 0x73f8 (39)
        Child SP               IP Call Site
00007F918B1A53D0 00007f9516af53c7 [InlinedCallFrame: 00007f918b1a53d0] Interop+Sys.ReceiveMessage(System.Runtime.InteropServices.SafeHandle, MessageHeader*, System.Net.Sockets.SocketFlags, Int64*)
00007F918B1A53D0 00007f949f6df19b [InlinedCallFrame: 00007f918b1a53d0] Interop+Sys.ReceiveMessage(System.Runtime.InteropServices.SafeHandle, MessageHeader*, System.Net.Sockets.SocketFlags, Int64*)
00007F918B1A53C0 00007F949F6DF19B ILStubClass.IL_STUB_PInvoke(System.Runtime.InteropServices.SafeHandle, MessageHeader*, System.Net.Sockets.SocketFlags, Int64*)
00007F918B1A5460 00007F94A0782568 System.Net.Sockets.SocketPal.Receive(System.Net.Sockets.SafeSocketHandle, System.Net.Sockets.SocketFlags, System.Span`1<Byte>, Byte[], Int32 ByRef, System.Net.Sockets.SocketFlags ByRef, Error ByRef)
[...]

Issue #567, opened 2020-05-17
SqlConnection.Open doesn't seem to work on Ubuntu 20.04


zanfranceschi:
I've recently upgraded to Ubuntu 20.04 (clean install) and found out the SqlConnection.Open doesn't seem to work using dotnet core sdk 3.1 and Microsoft SQL Server 2012 (SP4) versions.

It does not produce an error ─ it simply gets stuck and blocks execution. No timeout, not apparent error. I've even sniffed tcp traffic with wireshark to see if I could get something. Nothing.

It was working normally while using Ubuntu 19.10.

The steps to reproduce is to simply instantiate an SqlConnection object and try invoke "Open" on it.

I've also installed dotnet core 2.1 and tried System.Data.SqlClient. Same behaviour.

Any ideas or more information needed?

ErikEJ:
Are you using the 3.1.4 SDK, that is required for Ubuntu 20.04 support?

zanfranceschi:
Yes. I'm using it.

karinazhou:
Hi @zanfranceschi , 

Thank you for reporting this. I will take a look at it. 

Just to confirm the following behaviors with you again:
* neither of net core 2.1 and 3.1 works on Ubuntu 20.04
* both of net core 2.1 and 3.1 works on Ubuntu 19.10

Is it correct?

zanfranceschi:
> Hi @zanfranceschi ,
> 
> Thank you for reporting this. I will take a look at it.
> 
> Just to confirm the following behaviors with you again:
> 
> * neither of net core 2.1 and 3.1 works on Ubuntu 20.04
> * both of net core 2.1 and 3.1 works on Ubuntu 19.10
> 
> Is it correct?

Correct. That's it.

And just to add that the CPU consumption increases and takes 100% of one of the cores in a few seconds. But again, no dotnet error.

cheenamalhotra:
It looks related to #538

karinazhou:
@zanfranceschi 
I tried these docker images which are for Ubuntu 20.04 with the `docker-compose` project in SqlClient\ repository: 
mcr.microsoft.com/dotnet/core/runtime:3.1-focal
mcr.microsoft.com/dotnet/core/runtime:2.1-focal

When I connect to a SQL Server 2015, which doesn't support TLS 1.2, it gives me the following error instead of hanging:
`OpenSslCryptographicException: error:1425F102:SSL routines:ssl_choose_client_version:unsupported protocol`

[...]

PR #577, opened 2020-05-23
Fix infinite loop when end of stream is reached

alexdougie:
This is to fix issue #165 

We're seeing this issue in our production and development environments. We're using .NET Core 3.1 on CentOS 7. A reliable way to reproduce it is calling `SqlConnection.Open()` every half a second while Windows Server is restarting. This causes the process to go up to 100% CPU usage, which as you can imagine can cause problems. 😄 

We've tested against our environment and this fixes the problem, as the exception is caught further up and another is thrown out of `SqlConnection.Open()`.

Of course I am open to different approaches if this isn't the preferred way to handle it. 

- I'm not entirely happy with the variable name `readBytesForHeader`, any suggestions are welcome. 
- The exception message seems a bit redundant given the name of the exception. Any preference to having the message or not?
- This is also a problem in System.Data.SqlClient, so it needs backporting. I'm not sure how this is done. Of course I'm happy to submit a PR to fix it there too if that's preferred.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SslOverTdsStream.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SslOverTdsStream.cs
@@ -104,9 +104,16 @@ namespace Microsoft.Data.SqlClient.SNI
                     // Account for split packets
                     while (readBytes < TdsEnums.HEADER_LEN)
                     {
-                        readBytes += async ?
+                        var readBytesForHeader = async ?
                             await _stream.ReadAsync(packetData, readBytes, TdsEnums.HEADER_LEN - readBytes, token).ConfigureAwait(false) :
                             _stream.Read(packetData, readBytes, TdsEnums.HEADER_LEN - readBytes);
+
+                        if (readBytesForHeader == 0)
+                        {
+                            throw new EndOfStreamException("End of stream reached");
+                        }
+
+                        readBytes += readBytesForHeader;
                     }
 
                     _packetBytes = (packetData[TdsEnums.HEADER_LEN_FIELD_OFFSET] << 8) | packetData[TdsEnums.HEADER_LEN_FIELD_OFFSET + 1];


Wraith2:
Nice find. I was aware of this as a theoretical problem and have a fix in https://github.com/dotnet/SqlClient/pull/541 but you've actually verified that if can happen in production on 3.1 which I didn't think was possible.

cheenamalhotra:
Thanks for catching and confirming the fix @alexdougie !

We've not been able to reproduce the issue yet (also efforts ongoing in #567), without which we weren't sure if proposed fix (https://github.com/dotnet/SqlClient/issues/165#issuecomment-627880385) was accurate. But we can take your confirmation and also the fix isn't risky to implement so we can consider safely backporting it to v1.1.

We generally backport critical issues only to System.Data.SqlClient, and since we know the issue has greater impact, I think we can take it forward for backporting.

karinazhou:
Hi @alexdougie, I am trying to reproduce the hanging behavior on my side but no luck for now. From your comments, is the trick to restart the windows server where the SQL Server lives on when executing the SqlConnection.Open() ?

Wraith2:
The repro is tricky. You have to end up with a freshly opened (because encapsulation only applies on the first packet) managed sni connection which doesn't throw an exception when you call Read or ReadAsync on it which isn't supposed to be possible in 3.1 anymore see https://github.com/dotnet/runtime/issues/36293#event-3333460230 for related discussion. This combination is hard to engineer for obvious reasons which is why I thought it was theoretical until recently.

alexdougie:
It is very tricky to reproduce. It took us a long time to narrow it down. I've stuck the repro code below that we're running while restarting Windows Server (just restarting the SQL server service doesn't seem to cause the issue). Even then, it actually only happens about 50% of the time so it can take a few tries to reproduce.

If #541 fixes it then I don't mind this PR being left. So long as a fix is in the next version of SqlClient we'll be happy! 😄 

````
class Program
    {
        static async Task Main(string[] args)
        {
            while (true)
            {
                try
                {
                    var connectionString = args[0];

                    await using var connection = new SqlConnection(connectionString);
                    connection.Open();
                    Console.WriteLine("Opened connection.");
                }
                catch (Exception e)
                {
                    Console.WriteLine(e);
                }
                finally
                {
                    await Task.Delay(500);
                }
            }
        }
    }

karinazhou:
I tried this fix with [Issue#567](https://github.com/dotnet/SqlClient/issues/567)  which also triggers the hanging behavior. From the test, I can see the application goes into an infinite loop when `readBytes` remains unchanged inside the while loop in `ReadInternal` task.

The hanging can be reproduced both for .NET Core 2.1 and 3.1. One thing I am not sure about is what causes this `End of Stream` scenario.

Wraith2:
So even though I'd already got an open PR to fix this you merged this one? I really hope you're not going to complain that mine is out of date and request I update it to master because that would just be taking the piss at this point.

cheenamalhotra:
@Wraith2 

We will continue to review your PR in the next preview release cycle as it does bring lot of enhancements but also requires a thorough review and we would like to bring it in a preview version for end users to try on. Also we are currently wrapping up for GA 2.0 release, and are only picking up PRs that offer minimal changes and lowest risk for fixing important driver issues.

Issue #459, opened 2020-03-07
NotSupportedException & InvalidOperationException for unknown reasons


joshlang:
I'm lumping these two bugs into one, because I suspect they have the same root cause.

.NET Standard 2.1
Running in Azure AKS (linux)
System.Data.SqlClient version 4.8.1
Connecting to a tiny little sql azure database instance

We've run into two issues:

## Issue 1

InvalidOperationException: `Invalid operation. The connection is closed.`

^-- FOREVER!

Our code is exceedingly simple (flattened):
```
using var connection = new SqlConnection(ConnectionString);
await connection.OpenAsync(cancellationToken).ConfigureAwait(false);
using var cmd = connection.CreateCommand();
cmd.CommandText = command;
using var r = await cmd.ExecuteReaderAsync(cancellationToken).ConfigureAwait(false);
// consume reader results
// end of method
```

Stack trace:
```
System.InvalidOperationException:
   at System.Data.SqlClient.TdsParserStateObject.WritePacket (System.Data.SqlClient, Version=4.6.1.1, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a)
   at System.Data.SqlClient.TdsParserStateObject.WriteBytes (System.Data.SqlClient, Version=4.6.1.1, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a)
   at System.Data.SqlClient.TdsParser.WriteString (System.Data.SqlClient, Version=4.6.1.1, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a)
   at System.Data.SqlClient.TdsParser.TdsExecuteSQLBatch (System.Data.SqlClient, Version=4.6.1.1, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a)
   at System.Data.SqlClient.SqlCommand.RunExecuteReaderTds (System.Data.SqlClient, Version=4.6.1.1, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a)
   at System.Data.SqlClient.SqlCommand.BeginExecuteReader (System.Data.SqlClient, Version=4.6.1.1, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a)
   at System.Threading.Tasks.TaskFactory`1.FromAsyncImpl (System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e)
[...]

Issue #54, opened 2018-07-04
"The connection is closed" errors in System.Data.SqlClient on Linux without MARS. 


JulianRooze:
We've recently run into an issue where executing SQL commands results in `"Invalid operation. The connection is closed."` exceptions being thrown. 

- This seems to happen exclusively on Async operations.
- This seems to have started happening since we upgrade to .NET Core 2.1. We've seen occurrences in our logs before the upgrade, but that was a big burst of them (probably real, but transient, connectivity errors), but they've started showing up consistently literally the hour after we migrated to 2.1. 
- It looks like this happens only on Linux. We can't reproduce it on our local Windows machines. To rule out this is purely Azure related we reproduced it on Linux by running our application in a Linux Docker container on our local machines. 
- Turning on `MultipleActiveResultSets` (MARS) in the connection string seems to eliminate this error entirely. 
- The fact that it doesn't happen on Windows and that MARS solves it, rules out transient connectivity errors.
- Our usage of `SqlConnection` is pretty standard: they come from a connection pool, and we don't use them concurrently.
- I've attempted to reproduce the issue in an isolated project, but I've been unable to, so I'm not sure where the difference between that project and our full application lies.

The exception that gets thrown is this one:

https://github.com/dotnet/corefx/blob/a74bf4200926c47c11ba0383e2bddeaa59227266/src/System.Data.SqlClient/src/System/Data/Common/AdapterUtil.SqlClient.cs#L395

The exception gets thrown (or rethrown) from these three points in the code:

1. https://github.com/dotnet/corefx/blob/e0ba7aa8026280ee3571179cc06431baf1dfaaac/src/System.Data.SqlClient/src/System/Data/SqlClient/TdsParserStateObject.cs#L3098

Example stacktrace:

```csharp
at Task System.Data.SqlClient.TdsParserStateObject.WritePacket(byte flushMode, bool canAccumulate)
   at Task System.Data.SqlClient.TdsParserStateObject.WriteByteArray(byte[] b, int len, int offsetBuffer, bool canAccumulate, TaskCompletionSource<object> completion)
   at void System.Data.SqlClient.TdsValueSetter.SetString(string value, int offset, int length)
   at void Microsoft.SqlServer.Server.ValueUtilsSmi.SetString_LengthChecked(SmiEventSink_Default sink, ITypedSettersV3 setters, int ordinal, SmiMetaData metaData, string value, int offset)
[...]

PR #579, opened 2020-05-25
WIP | Prevent WriteAsync collision

DavoudEshtehari:
Related issue #459 and #54 
According to Microsoft documentation, [WriteAsync](https://docs.microsoft.com/en-us/dotnet/api/system.net.security.sslstream.writeasync?view=netcore-3.1) a new method call just is able if there is not another one in progress.
This issue happened in a race condition, cause of asynchronous programming.

_**Update: Proposal extended and reworked in #796**_


Trontin:
Any news on the progress of this PR ? It seems unactive since one month.

DavoudEshtehari:
> Any news on the progress of this PR ? It seems unactive since one month.

This PR is blocked by an issue in dotnet/runtime. You can follow it at [#38185](https://github.com/dotnet/runtime/issues/38185#issuecomment-666308513).

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNICommon.cs
```diff
@@ -99,6 +102,44 @@ internal enum SNISMUXFlags
...
+        }
+
+        public SslStreamProxy(Stream innerStream, bool leaveInnerStreamOpen, RemoteCertificateValidationCallback userCertificateValidationCallback, LocalCertificateSelectionCallback userCertificateSelectionCallback) : base(innerStream, leaveInnerStreamOpen, userCertificateValidationCallback, userCertificateSelectionCallback)
+        {
+        }
+
+        public SslStreamProxy(Stream innerStream, bool leaveInnerStreamOpen, RemoteCertificateValidationCallback userCertificateValidationCallback, LocalCertificateSelectionCallback userCertificateSelectionCallback, EncryptionPolicy encryptionPolicy) : base(innerStream, leaveInnerStreamOpen, userCertificateValidationCallback, userCertificateSelectionCallback, encryptionPolicy)
+        {
+        }
+        #endregion
```
cheenamalhotra, line 129: You don't need all the constructors actually..

```suggestion
        public SslStreamProxy(Stream innerStream, bool leaveInnerStreamOpen, RemoteCertificateValidationCallback userCertificateValidationCallback) 
            : base(innerStream, leaveInnerStreamOpen, userCertificateValidationCallback)
        { }
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNINpHandle.cs
```diff
@@ -93,7 +93,7 @@ public SNINpHandle(string serverName, string pipeName, long timerExpire, object
                 }
 
                 _sslOverTdsStream = new SslOverTdsStream(_pipeStream);
-                _sslStream = new SslStream(_sslOverTdsStream, true, new RemoteCertificateValidationCallback(ValidateServerCertificate), null);
+                _sslStream = new SslStreamProxy(_sslOverTdsStream, true, new RemoteCertificateValidationCallback(ValidateServerCertificate), null);
```
cheenamalhotra, line 96:
```suggestion
                _sslStream = new SslStreamProxy(_sslOverTdsStream, true, new RemoteCertificateValidationCallback(ValidateServerCertificate));
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
```diff
@@ -226,7 +226,7 @@ public SNITCPHandle(string serverName, int port, long timerExpire, object callba
                 _tcpStream = new NetworkStream(_socket, true);
 
                 _sslOverTdsStream = new SslOverTdsStream(_tcpStream);
-                _sslStream = new SslStream(_sslOverTdsStream, true, new RemoteCertificateValidationCallback(ValidateServerCertificate), null);
+                _sslStream = new SslStreamProxy(_sslOverTdsStream, true, new RemoteCertificateValidationCallback(ValidateServerCertificate), null);
```
cheenamalhotra, line 229:
```suggestion
                _sslStream = new SslStreamProxy(_sslOverTdsStream, true, new RemoteCertificateValidationCallback(ValidateServerCertificate));
```

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/SqlCommand/SqlCommandExecuteTest.cs
```diff
@@ -0,0 +1,151 @@
...
+
+        private static async Task ExecuteScalarAsync(string connectionString, CancellationToken token)
+        {
+            using (var connection = new SqlConnection(connectionString))
+            {
+                using (var cmd = GetCommand(connection))
+                {
+                    await cmd.ExecuteScalarAsync(token);
+                }
+            }
```
cheenamalhotra, line 85:
```suggestion
            using (var connection = new SqlConnection(connectionString))
            using (var cmd = GetCommand(connection))
            {
                await cmd.ExecuteScalarAsync(token);
            }
```

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/SqlCommand/SqlCommandExecuteTest.cs
```diff
@@ -0,0 +1,151 @@
...
+                {
+                    var r = await cmd.ExecuteReaderAsync(token);
+                    while (await r.ReadAsync(token))
+                    {
+                        await r.GetFieldValueAsync<string>(0);
+                        await r.GetFieldValueAsync<string>(1);
+                        await r.GetFieldValueAsync<string>(2);
+                    }
+                }
+            }
```
cheenamalhotra, line 74:
```suggestion
            using (var connection = new SqlConnection(connectionString))
            using (var cmd = GetCommand(connection))
            using (var r = await cmd.ExecuteReaderAsync(token))
            {
                while (await r.ReadAsync(token))
                {
                    await r.GetFieldValueAsync<string>(0);
                    await r.GetFieldValueAsync<string>(1);
                    await r.GetFieldValueAsync<string>(2);
                }
            }
```

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/SqlCommand/SqlCommandExecuteTest.cs
```diff
@@ -0,0 +1,151 @@
...
+
+        private static async Task ExecuteNonQueryAsync(string connectionString, CancellationToken token)
+        {
+            using (var connection = new SqlConnection(connectionString))
+            {
+                using (var cmd = GetCommand(connection))
+                {
+                    await cmd.ExecuteNonQueryAsync(token);
+                }
+            }
```
cheenamalhotra, line 96:
```suggestion
            using (var connection = new SqlConnection(connectionString))
            using (var cmd = GetCommand(connection))
            {
                await cmd.ExecuteNonQueryAsync(token);
            }
```

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/SqlCommand/SqlCommandExecuteTest.cs
```diff
@@ -0,0 +1,151 @@
...
+        private static async Task ExecuteXmlReaderAsync(string connectionString, CancellationToken token)
+        {
+            using (var connection = new SqlConnection(connectionString))
+            {
+                using (var cmd = GetCommand(connection))
+                {
+                    cmd.CommandText += " FOR XML AUTO, XMLDATA";
+                    var r = await cmd.ExecuteXmlReaderAsync(token);
+                }
+            }
```
cheenamalhotra, line 108:
```suggestion
            using (var connection = new SqlConnection(connectionString))
            using (var cmd = GetCommand(connection))
            {
                cmd.CommandText += " FOR XML AUTO, XMLDATA";
                var r = await cmd.ExecuteXmlReaderAsync(token);
            }
```

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/SqlCommand/SqlCommandExecuteTest.cs
```diff
@@ -0,0 +1,151 @@
...
+            string aRecord = "('2455cf1b-ebcf-418d-8cce-88e21e1683e3', 'something', 'updated'),";
+            string query = "SELECT * FROM (VALUES"
+                + string.Concat(Enumerable.Repeat(aRecord, 200)).Substring(0, (aRecord.Length * 200) - 1)
+                + ") tbl_A ([Id], [Name], [State])";
+            cnn.Open();
+            using (var cmd = cnn.CreateCommand())
+            {
+                cmd.CommandText = query;
+                return cmd;
+            }
```
cheenamalhotra, line 122:
```suggestion
            var cmd = cnn.CreateCommand()
            cmd.CommandText = query;
            return cmd;
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
```diff
@@ -706,12 +706,23 @@ public override void SetAsyncCallbacks(SNIAsyncCallback receiveCallback, SNIAsyn
...
+            long scopeID = SqlClientEventSource.Log.TrySNIScopeEnterEvent("<sc.SNI.SNITcpHandle.SendAsync |SNI|SCOPE>");
+            SNIPacket errorPacket;
             SNIAsyncCallback cb = callback ?? _sendCallback;
-            lock (this)
+            try
             {
                 packet.WriteToStreamAsync(_stream, cb, SNIProviders.TCP_PROV);
+                return TdsEnums.SNI_SUCCESS_IO_PENDING;
+            }
+            catch (Exception e) when (e is ObjectDisposedException || e is InvalidOperationException || e is IOException)
```
cheenamalhotra, line 717: Also capture these for SNINpHandle class?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIPacket.cs
```diff
@@ -304,16 +304,21 @@ public void WriteToStream(Stream stream)
...
-                SNILoadHandle.SingletonInstance.LastError = new SNIError(provider, SNICommon.InternalExceptionError, e);
-                status = TdsEnums.SNI_ERROR;
-            }
-            callback(this, status);
+                Exception e = t.Exception?.InnerException;
+                if (e != null)
+                {
+                    SNILoadHandle.SingletonInstance.LastError = new SNIError(provider, SNICommon.InternalExceptionError, e);
+                    status = TdsEnums.SNI_ERROR;
+                    Release();
```
cheenamalhotra, line 315: Just wanted to callout "Release()" added here, I guess this is made intentionally?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNICommon.cs
```diff
@@ -99,6 +102,44 @@ internal enum SNISMUXFlags
...
+
+        // Prevent the WriteAsync's collision
+        public override Task WriteAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
+        {
+            if (_currentTask != null && _currentTask.Status != TaskStatus.RanToCompletion)
+            {
+                _currentTask.Wait(cancellationToken);
```
roji, line 136: Unless I'm mistaken, this is a synchronous call which blocks the current thread - that's a pretty bad idea in an async method (mixing sync and async can lead to all kinds of pain). Since _currentTask represents an asynchronous task (if I'm getting the code right), this is also a case of sync-over-async, which can cause deadlocks.

roji: The driver definitely needs to avoid starting a write if another one is in progress (the same is true for reads); this isn't just true for SslStream, but also for NetworkStream - and pretty much anything that does networking. FWIW I don't think this restriction is against async API design - async is about not blocking the thread, rather than about allowing multiple concurrent operations to occur on the same resource (i.e. connection).

More concretely... I don't doubt the importance of this fix (I have no context here) - all I'm saying is that any waiting for previous writes to complete should be done asynchronously, otherwise you're likely creating a new category of future bugs. One easy way to do this would be to have a SemaphoreSlim on the connection representing writing, and to simply call [await WaitAsync](https://docs.microsoft.com/en-us/dotnet/api/system.threading.semaphoreslim.waitasync?view=netcore-3.1#System_Threading_SemaphoreSlim_WaitAsync) on it. This would ensure only one writing operation occurs at any given moment, while making all waiting happens asynchronously. There may be some perf implication here, but I really don't have enough context to know how hot this code path is etc.

roji: BTW here's an example scenario on where sync-over-async could go wrong.

Say you have a lot of writers all of a sudden (e.g. because of a thundering herd effect or whatever). One of the writers wins and starts to do their thing; meanwhile, all the other enter the Wait call, blocking the thread. If your writers happen to be executed on the thread pool (which they are in ASP.NET, for example), then you've just exhausted the thread pool completely. The really bad part, is that when your very first writer completes - the one which actually started writing - it also needs a thread pool thread to execute the continuation (the code that runs after the async operation completed), but the pool is exhausted. This is a pseudo-deadlock, and your application will hang until the thread pool allocates new threads, which can take quite a while.

Since I'm lacking context, I don't know if the above can occur here specifically. But with all the bad scenarios around mixing sync and async code, it really best be avoided...

DavoudEshtehari: hi @roji 
Thank you for the valuable point.

Since the sequence of calling the `WriteAsync` method is important in this scenario, and `SemaphoreSlim` doesn't respect the order, Do you have any solution/guidance for this case?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIPacket.cs
DavoudEshtehari: I couldn't see a reason except resue the packet's allocated memory after encountering by an exception.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNICommon.cs
roji: @DavoudEshtehari apologies for disappearing for so long, I had some personal issues.

Looking at the current method as it is in the PR, I don't see any enforcing of ordering - if _currentTask is not-null (i.e. someone else is writing), we wait until that Task is over. I'm basically proposing substituting _currentTask for SemaphoreSlim as the synchronization check, that's all.

If you need to have multiple writers wait in line based on FIFO, then a more elaborate solution with ConcurrentQueue can be built - but I'd definitely need more context to understand needs here.

DavoudEshtehari: @roji thank you for your kind reply.

The issue might occur when more than one write request waits behind the semaphore to get permission to run the WriteAsync function. I think this is a possible condition.

Also, I've tried queue patterns but touching the WriteAsync throws the exception even though I wrap it up with a cold task.

roji: Note that with the current code, if more than one write request Waits on the Task, both will get unblocked the moment the task completes, and will concurrently call into base.WriteAsync - which is what this PR is trying to prevent. SemaphoreSlim would easily take care of that for you - when its WaitAsync method returns, it's guaranteed that only one request has acquired the semaphore. However, there is no guarantee of ordering - if more than one attempt did WaitAsync, they would get unblocked in "random" order.

martyt:
Following this PR in hopes of seeing resolution soon. We ended up forking the repo and applying a fix based on the suggestions above and it's working wonderfully. Any chance we can get this PR updated with the code we put in and move forward with the PR?

`
    internal class SslStreamProxy : SslStream
    {
        private readonly SemaphoreSlim _semaphore;
        public SslStreamProxy(Stream innerStream, bool leaveInnerStreamOpen, RemoteCertificateValidationCallback userCertificateValidationCallback) 
            : base(innerStream, leaveInnerStreamOpen, userCertificateValidationCallback)
        {
            _semaphore = new SemaphoreSlim(1); /* granting one concurrent write on sslStream */
        }
        // Prevent the WriteAsync's collision
        public override async Task WriteAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
        {
            await _semaphore.WaitAsync(cancellationToken);
            try
            {
                await base.WriteAsync(buffer, offset, count, cancellationToken);
            }
            finally
            {
                _semaphore.Release();
            }
            return;
        }
        protected override void Dispose(bool disposing)
        {
            _semaphore?.Dispose();
            base.Dispose(disposing);
        }
    }
`

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNICommon.cs
cheenamalhotra: Hi @roji 

Thanks for your tips, I got an implementation of Concurrent Queue with Semaphore Slim working (PR #796), would request you to take a look too.

However, this makes me think couldn't it be done by SslStream class itself? So that libraries don't land into surprises... It seems a basic case of ReadAsync/WriteAsync + parallel tasks is not supported by them and it can cause dangerous problems. Maybe a thread-safe design can be proposed to System.Net.Security?

DavoudEshtehari:
It got fixed on PR #796.

PR #590, opened 2020-06-03
Add support for the new native SNI

karinazhou:
This is for the coming new version of native SNI.


karinazhou:
Hold this for now since we need to update the depending SNI nuget package version.

karinazhou:
Close this one since we have another with full implementation : [PR594](https://github.com/dotnet/SqlClient/pull/594)

PR #594, opened 2020-06-06
Add DNS Caching Phase 1

karinazhou:
This PR adds support for the upcoming DNS Caching feature on the server-side. There are two phases for this feature and this is the first phase support.

When using TCP protocol, the SqlClient driver will record the DNS information such as IP address and port number in an in-memory cache if the server sends back the corresponding TDS token for this caching. The DNS information will stay in the cache during the application lifetime unless the server tells the driver to clean the cache.

In phase 1, the IP and port number are from the DNS resolution during pre-login handshake. If the DNS server fails to resolve the server name, the driver will connect with its cached IP and port if they existed. Otherwise, the connection will fail. 

- Add support for both .NET Framework and .NET Core.
- This change requires the corresponding change in native SNI.
- The caching behavior is only for TCP connections.
- Need to fail the DNS Server to test this behavior.

image


src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DNSCachingTest/DNSCachingTest.cs
```diff
@@ -0,0 +1,81 @@
...
+                    {
+                        Assert.Null(parameters[1]);
+                    }
+                }
+            }
+        }
+
+    }
+
+}
```
cheenamalhotra, line 81: Adjust empty lines/formatting and add EOF line.

src/Microsoft.Data.SqlClient/tests/ManualTests/config.default.json
```diff
@@ -14,5 +14,10 @@
     "SupportsIntegratedSecurity": true,
     "SupportsLocalDb": false,
     "SupportsFileStream": false,
-    "UseManagedSNIOnWindows": false
+    "UseManagedSNIOnWindows": false,
+    "DNSCachingConnString": "",
+    "DNSCachingServerCR": "",
+    "DNSCachingServerTR": "",
+    "IsDNSCachingSupportedCR": false,
+    "IsDNSCachingSupportedTR": false
```
cheenamalhotra, line 22: Update BUILDGUIDE

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SQLDNSCache.cs
```diff
@@ -0,0 +1,87 @@
...
+                {
+                    return (newItem.AddrIPv4 == oldItem.AddrIPv4 &&
+                            newItem.AddrIPv6 == oldItem.AddrIPv6 &&
+                            newItem.Port == oldItem.Port);
+                }
+            }
+
+            return false;
+        }
+
```
cheenamalhotra, line 69: Nit: Format + adjust line spacings.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SQLDNSCache.cs
```diff
@@ -0,0 +1,87 @@
...
+using System;
+using System.Collections.Concurrent;
+
+namespace Microsoft.Data.SqlClient
+{
+    internal class SQLDNSCache
+    {
+        private static readonly SQLDNSCache _SQLDNSCache = new SQLDNSCache();
+        private static readonly int initialCapacity = 100;
```
cheenamalhotra, line 13: What happens if capacity is exhausted?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SQLDNSCache.cs
```diff
@@ -0,0 +1,87 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Concurrent;
+
+namespace Microsoft.Data.SqlClient
+{
+    internal class SQLDNSCache
+    {
+        private static readonly SQLDNSCache _SQLDNSCache = new SQLDNSCache();
```
cheenamalhotra, line 12: When will this cache clear at all?
Can we provide a way to clear DNS Cache like we have "ClearAllPools" public API on `SqlConnection`?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
```diff
@@ -120,33 +125,68 @@ public SNITCPHandle(string serverName, int port, long timerExpire, object callba
...
-                {
-                    Task<IPAddress[]> serverAddrTask = Dns.GetHostAddressesAsync(serverName);
-                    serverAddrTask.Wait(ts);
-                    IPAddress[] serverAddresses = serverAddrTask.Result;
+                bool reportError = true;
 
-                    if (serverAddresses.Length > MaxParallelIpAddresses)
+                try
+                {
+                    if (parallel)
```
JRahnama, line 132: We need to add more comments here. parallel I think is true when the MultiSubnetFailover is set to true. While you are there can we add some comments please?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
```diff
@@ -181,25 +225,87 @@ public SNITCPHandle(string serverName, int port, long timerExpire, object callba
             _status = TdsEnums.SNI_SUCCESS;
         }
 
-        private static Socket Connect(string serverName, int port, TimeSpan timeout, bool isInfiniteTimeout)
+        private Socket TryConnectParallel(string hostName, int port, TimeSpan ts, bool isInfiniteTimeOut, ref bool callerReportError, string cachedFQDN, ref SQLDNSInfo pendingDNSInfo)
```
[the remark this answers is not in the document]
karinazhou: I need to keep this function non-static because `ReportTcpSNIError` is non-static.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SQLDNSCache.cs
karinazhou: Good question. According to the [document](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.-ctor?view=netcore-3.1):

>The default capacity (DEFAULT_CAPACITY), which represents the initial number of buckets, is a trade-off between the size of a very small dictionary and the number of resizes when constructing a large dictionary. Also, the capacity should not be divisible by a small prime number. The default capacity is 31.

When the capacity is reached, it will resize the dictionary. I will change it to a prime number as the initial capacity, e.g. 101

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SQLDNSCache.cs
karinazhou: Currently, there is no control of clearing all items in the cache. Neither the client or the server has control of that. We will remove specific items from cache when the server tells us to do so but it should not impact other items in the cache. Only when the application is stopped, the cache will be recycled.

src/Microsoft.Data.SqlClient/tests/ManualTests/config.default.json
karinazhou: They will be removed as per offline discussion.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SQLDNSCache.cs
```diff
@@ -0,0 +1,87 @@
...
+        {
+            if (null != item)
+            {
+                if (DNSInfoCache.ContainsKey(item.FQDN))
+                {
+
+                    DeleteDNSInfo(item.FQDN);
+                }
+
+                return DNSInfoCache.TryAdd(item.FQDN, item);
```
cheenamalhotra, line 37: Could you use `AddorUpdate` instead of deleting first?

src/Microsoft.Data.SqlClient/netcore/src/Interop/SNINativeMethodWrapper.Windows.cs
```diff
@@ -236,6 +253,17 @@ internal struct SNI_Error
         [DllImport(SNI, CallingConvention = CallingConvention.Cdecl)]
         private static extern uint SNIGetInfoWrapper([In] SNIHandle pConn, SNINativeMethodWrapper.QTypes QType, out Guid pbQInfo);
 
+        // kz start
```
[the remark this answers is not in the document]
karinazhou: I will remove all such comments.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SQLDNSCache.cs
cheenamalhotra: I don't know if there would be users who would like to clear this up and control memory allocations.
@David-Engel any thoughts?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SQLDNSCache.cs
karinazhou: I was trying to use the AddorUpate method. But from the [documents](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.addorupdate?view=netcore-3.1), it mentions 
>If you call AddOrUpdate simultaneously on different threads, addValueFactory may be called multiple times, but its key/value pair might not be added to the dictionary for every call.
>
>For modifications and write operations to the dictionary, ConcurrentDictionary<TKey,TValue> uses fine-grained locking to ensure thread safety. (Read operations on the dictionary are performed in a lock-free manner.) However, the addValueFactory and updateValueFactory delegates are called outside the locks to avoid the problems that can arise from executing unknown code under a lock. Therefore, AddOrUpdate is not atomic with regards to all other operations on the ConcurrentDictionary<TKey,TValue> class.

And it requires addition function for updating the records so I think remove and add the value again will be safer and straight-forward.

Wraith2:
I can't currently reply inline, you said.

https://github.com/dotnet/SqlClient/pull/594#discussion_r437656345
> Currently, there is no control of clearing all items in the cache.

That's a leak then. A cache is normally bounded, is this one?  And more importantly if users can't clear it they will have to fall back to the nuclear option of killing the process to get rid of it and that's a fairly annoying thing to need to do.

As @cheenamalhotra said I think there should be an api for users to control this feature. 

Can you point me at the part of the spec that describes this?

karinazhou:
@Wraith2  There is no spec for this for now since the server-side hasn't released this feature yet. From the current design, this cache is hidden from the client and used by the driver internally. From the behavior on the server-side, it will send back the IsSupported flag in the feature ext tds to tell the driver whether to remove the item from the cache or not. If the driver receives this flag, it will remove its corresponding item from the cache. But the cache itself will not be cleared up because it is shared by all the connections when the application is running.

Wraith2:
If you can't connect to the server because of an invalid cache entry you can't get an eviction notification so the cache is permanently poisoned for the process. There needs to be some eviction mechanism even if it isn't available to users.

karinazhou:
@Wraith2  I understand your concern. Basically, we will always first use the information from the connection string to connect even we have items in the cache. The server name will be resolved by DNS server as before. Only when the DNS resolution fails, we will try again with the IP in the cache. 

Currently, the SQL Servers without DNS Caching feature won't have IsSupported flag in their feature ext token. This means that we won't put anything in the DNS cache for these servers. For other servers on Azure which may have this DNS Caching feature supported, we will cache the IPs for them as required or clean the IP in the cache as they want. 

This PR is not the complete version for the entire DNS caching support implementation. There will be a continuing phase 2 for both server and driver side. We will cache the actual information sent from the server every time we reach the server endpoint. Of course, we will take your advice and try to see how we can clean up the whole cache if needed.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsEnums.cs
```diff
@@ -214,6 +214,7 @@ public enum EnvChangeType : byte
         public const byte FEATUREEXT_AZURESQLSUPPORT = 0x08;
         public const byte FEATUREEXT_DATACLASSIFICATION = 0x09;
         public const byte FEATUREEXT_UTF8SUPPORT = 0x0A;
+        public const byte FEATUREEXT_AZURESQLDNSCACHING = 0x0B;
```
David-Engel, line 217: I think we should just name this feature SQLDNSCaching everywhere instead of using Azure in the name.

karinazhou: Yeah, I know it's a bit inconsistent here. In the specs from the server team, they use `AZURESQLDNSCACHING` for this new feature ext so I use it here.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SQLDNSCache.cs
David-Engel: If we do a little math, estimate 60 bytes of data per entry (ex: 12345678901234567890.database.windows.net 111.111.111.111), 3(?) entries per DB (GW, CR, TR?), an application that makes 1000 connections to different databases, and we get 176kB. Maybe double that for overhead of the ConcurrentDictionary. It really should never grow beyond the number of different databases the application connects to. It doesn't seem like applications should need to control it.

karinazhou: Based on the previous discussion with server team, this feature is boxed inside the driver and is not exposed to the client application. I think the cache should hold the records during the application's lifetime. For now, if we connect to the server endpoint and receive the flag from server saying that we need to remove its entry, the driver will remove that particular one from the cache. But for other records for other servers / FQDNs, they will still in the cache. They can still be used by other connections while the application is running. It's a bit tricky to find the proper timing to clean up the whole DNS cache while the application is running if it is not controlled by the client.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
karinazhou: I will add the comments for this part. Basically, the connection logic is the same except for giving more tries when the DNS resolution fails for the first time. I wrapped the parts for parallel connection into the `TryConnectParallel` function which will make the retry logic clearer.

Wraith2:
I think it might be a good idea to call this a fallback dns cache. Typically caches are used to prevent the need for an operation by memorization of a result which in the case of this feature could cause reliability problems. What this feature is really doing is being a backup in case the operation fails, so we fall back to it to increase reliability. I understand that the name of the defines from the sql server team can't be changed but in this library we have free reign to use descriptive naming.

doc/snippets/Microsoft.Data.SqlClient/SqlConnection.xml
```diff
@@ -1055,6 +1055,20 @@ GO
...
+        <RetrieveInternalInfo>
+            <summary>Returns a name value pair collection of internal properties at the point in time the method is called.</summary>
+            <returns>Returns a reference of type <see cref="T:System.Collections.Generic.IDictionary" /> of (string, object) items.</returns>
+            <remarks>
+            <format type="text/markdown"><![CDATA[  
+  
+## Remarks  
+ When this method is called, the values retrieved are those at the current point in time. If you continue using the connection, the values are incorrect. You need to re-execute the method to obtain the most current values.
+ 
+ Supported internal properties: `SQLDNSCachingSupportedState` and `SQLDNSCachingSupportedStateBeforeRedirect`.
```
cheenamalhotra, line 1067: Can you also describe about these properties what they mean and what info they provide. Maybe create a table below?

karinazhou: Updated

PR #597, opened 2020-06-10
Introduce support for AAD Device Code Flow authentication

cheenamalhotra:
This PR adds support for "Device Code Flow" authentication mode with MSAL.NET.
Reference documentation: [OAuth2.0 Device Authorization Grant flow](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-device-code)

Example of Connection string:

`Server=<server>.database.windows.net; Authentication=Active Directory Device Code Flow; Database=Northwind;`

Also enables customization of Device Code Flow callback with below API:
`ActiveDirectoryAuthenticationProvider.SetDeviceCodeFlowCallback(Func<DeviceCodeResult, Task> deviceCodeFlowCallbackMethod)`

**Edit:**

Added new APIs as under to enable customizations for "Active Directory Interactive" authentication experience:
- `ActiveDirectoryAuthenticationProvider.SetIWin32WindowFunc(Func<IWin32Window> iWin32WindowFunc)`
   -- _Applies to .NET Framework only_
- `ActiveDirectoryAuthenticationProvider.SetParentActivityOrWindowFunc(Func<object> parentActivityOrWindowFunc)`
   -- _Applies to .NET Standard only_
- `ActiveDirectoryAuthenticationProvider.SetCustomWebUi(ICustomWebUi customWebUi)`
   -- _Applies to .NET Framework, .NET Core and .NET Standard_


doc/samples/CustomDeviceCodeFlowAzureAuthenticationProvider.cs
```diff
@@ -0,0 +1,56 @@
...
+{
+    /// <summary>
+    /// Example demonstrating creating a custom device code flow authentication provider and attaching it to the driver.
+    /// This is helpful for applications that wish to override the Callback for the Device Code Result implemented by the SqlClient driver.
+    /// </summary>
+    public class CustomDeviceCodeFlowAzureAuthenticationProvider : SqlAuthenticationProvider
+    {
+        public override async Task<SqlAuthenticationToken> AcquireTokenAsync(SqlAuthenticationParameters parameters)
+        {
+            string clientId = "my-client-id";
```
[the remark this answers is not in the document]
cheenamalhotra: I added the public API to `SetDeviceCodeFlowCallback` and made the class public along with another sample, please give it a try 🙏 Thanks!

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ActiveDirectoryAuthenticationProvider.cs
```diff
@@ -12,21 +11,24 @@
...
         private static readonly string s_defaultScopeSuffix = "/.default";
         private readonly string _type = typeof(ActiveDirectoryAuthenticationProvider).Name;
         private readonly SqlClientLogger _logger = new SqlClientLogger();
+        private Func<DeviceCodeResult, Task> _deviceCodeFlowCallback;
+
+        /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/ActiveDirectoryAuthenticationProvider.xml' path='docs/members[@name="ActiveDirectoryAuthenticationProvider"]/ctor/*'/>
+        public ActiveDirectoryAuthenticationProvider() => new ActiveDirectoryAuthenticationProvider(DefaultDeviceFlowCallback);
+
+        /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/ActiveDirectoryAuthenticationProvider.xml' path='docs/members[@name="ActiveDirectoryAuthenticationProvider"]/ctor2/*'/>
+        public ActiveDirectoryAuthenticationProvider(Func<DeviceCodeResult, Task> deviceCodeFlowCallbackMethod)
```
David-Engel, line 26: Opening the up as a public API is very interesting. It would allow us to do some other things, too:

1. For the Interactive flow, we could allow applications to pass in an IWin32Window which we could send to MSAL (.NET Framework only) via .WithParentActivityOrWindow (someone just asked me for this privately) for a more controlled application experience. We could even take in a Microsoft.Identity.Client.Extensibility.ICustomWebUi and pass that via .WithCustomWebUi (netfx/netcore) for even more flexibility.

I'm not sure about adding new public APIs this close to the release without time to discuss. We can always add them later, but it's hard to take them away.

@saurabh500 What do you think about these?

src/Microsoft.Data.SqlClient/tests/FunctionalTests/AADTests.cs
```diff
@@ -4,11 +4,13 @@
...
 using System;
 using System.Security;
+using System.Threading.Tasks;
+using Microsoft.Identity.Client;
 using Xunit;
 
 namespace Microsoft.Data.SqlClient.Tests
 {
-    public class AADAccessTokenTest
+    public class AADTests
```
saurabh500, line 13: Might be better to call this AADAuthenticationTests

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ActiveDirectoryAuthenticationProvider.cs
saurabh500: I like the idea. I always wondered what would happen for interactive flow different systems. This seems reasonable. 
And +1 to adding a task to backlog and not handling in this release.

saurabh500: @cheenamalhotra  do we need to open the whole class up? Wouldn't an API on SqlAuthenticationProvider itself, suffice?

cheenamalhotra: `SqlAuthenticationProvider` contains no implementation so there's nothing to override there. Making this class public is the only way to allow users to take advantage of driver's implementation and customize it as needed.

saurabh500: This class is meant for inheritance as such. I would recommend only adding APIs which allow us to override the behavior which we want users to override. Adding an API on SqlAuthenticationProvider to set the DeviceAuthFlowCallback should be propagated all the way to `ActiveDirectoryAuthenticationProvider` and that would help achieve what we want. 
In case we open up the whole class then we have to provide a lot of guidance on what it means to override a particular method and this class is based on a lot of internal workings of AAD Auth and whether we are using one kind of Auth over another etc.

cheenamalhotra: `SqlAuthenticationProvider` is abstract class, it can't be instantiated by itself.

Users need to be able to instantiate our implementation of `ActiveDirectoryAuthenticationProvider` to override our callback method and for creating instance, it needs to be public.

saurabh500: You can always expose a static method in `SqlAuthenticationProvider`. I can imaging different callback per connection, but it doesn't seem like something needed per AuthProvider instance. 
Adding a static to `SqlAuthenticationProvider` is an easier adoption path when it comes to using it.

saurabh500: I don't imagine an application instance providing different ways of surfacing the URL for device flow on different connections. This is going to be a console app in most cases. I think a static global callback should be sufficient in this case.

David-Engel: @saurabh500 Cheena's point about opening up `ActiveDirectoryAuthenticationProvider` versus `SqlAuthenticationProvider` is that if we don't open up `ActiveDirectoryAuthenticationProvider`, users have to implement all of `ActiveDirectoryAuthenticationProvider` themselves just to customize a few MSAL options. Letting them pass something to handle the device code flow output or interactive window would be much simpler for them.

cheenamalhotra: Do you mean we should push this callback to `SqlAuthenticationProviderManager` from static APIs to control custom callback method? But putting callback in `SqlAuthenticationProvider` would mean MSAL.NET using will be part of `SqlAuthenticationProvider` as well as `SqlAuthenticationProviderManager`.

Because this callback is very specific to MSAL.NET as it contains DeviceCodeResult in params, , I don't think MSAL.NET should be used outside the scope of `ActiveDirectoryAuthenticationProvider` that does all the authentication work. So that users are not be forced with using MSAL.NET when implementing `SqlAuthenticationProvider`.

saurabh500: I see. I understand better now. In that case it makes sense to open up this type and I guess we will have to add more constructors or properties to accept the customizations. Can we make this class sealed though? I am still of the opinion that this class is not.meant to be inherited

cheenamalhotra: Yes sealing it would be ideal 👍

doc/snippets/Microsoft.Data.SqlClient/ActiveDirectoryAuthenticationProvider.xml
```diff
@@ -0,0 +1,82 @@
...
+      <returns>Represents an asynchronous operation that returns the AD authentication token.</returns>
+    </AcquireTokenAsync>
+    <SetDeviceCodeFlowCallback>
+      <param name="deviceCodeFlowCallbackMethod">The callback method to be used by driver when performing 'Active Directory Device Code Flow' authentication.</param>
+      <summary>Sets callback method that overrides driver's default implementation to process result when performing 'Active Directory Device Code Flow' authentication.</summary>
+    </SetDeviceCodeFlowCallback>
+    <SetParentActivityOrWindowFunc>
+      <param name="parentActivityOrWindowFunc">The parent as an object, in order to be used from shared NetStandard assemblies.</param>
+      <summary>Sets a reference to the ViewController (if using Xamarin.iOS), Activity (if using Xamarin.Android) IWin32Window or IntPtr (if using .Net Framework). Used for invoking the browser for Active Directory Interactive authentication.</summary>
+      <remarks>Mandatory only on Android to be set. See https://aka.ms/msal-net-android-activity for further documentation and details.</remarks>
```
David-Engel, line 31:
```suggestion
      <remarks>Mandatory to be set only on Android. See https://aka.ms/msal-net-android-activity for further documentation and details.</remarks>
```

doc/snippets/Microsoft.Data.SqlClient/ActiveDirectoryAuthenticationProvider.xml
```diff
@@ -0,0 +1,82 @@
...
+    </SetParentActivityOrWindowFunc>
+    <SetIWin32WindowFunc>
+      <param name="iWin32WindowFunc">A function to return the current window.</param>
+      <summary>Sets a reference to the current <see cref="T:System.Windows.Forms.IWin32Window" /> that triggers the browser to be shown. Used to center the browser that pop-up onto this window."</summary>
```
David-Engel, line 35:
```suggestion
      <summary>Sets a reference to the current <see cref="T:System.Windows.Forms.IWin32Window" /> that triggers the browser to be shown. Used to center the browser pop-up onto this window."</summary>
```

doc/snippets/Microsoft.Data.SqlClient/ActiveDirectoryAuthenticationProvider.xml
```diff
@@ -0,0 +1,82 @@
...
+    </SetIWin32WindowFunc>
+    <SetCustomWebUI>
+      <param name="customWebUi">Customer implementation for the Web UI</param>
+      <summary>Sets a custom Web UI that will let the user sign-in with Azure AD, present consent if needed, and get back the authorization code. Applicable when working with Active Directory Interactive authentication.</summary>
```
David-Engel, line 39:
```suggestion
      <summary>Sets a custom Web UI that will let the user sign-in with Azure Active Directory, present consent if needed, and get back the authorization code. Applicable when working with Active Directory Interactive authentication.</summary>
```

doc/snippets/Microsoft.Data.SqlClient/ActiveDirectoryAuthenticationProvider.xml
```diff
@@ -0,0 +1,82 @@
...
+|Authentication Method|
+|----------------|
+|<xref:Microsoft.Data.SqlClient.SqlAuthenticationMethod.ActiveDirectoryPassword%2A>|
+|<xref:Microsoft.Data.SqlClient.SqlAuthenticationMethod.ActiveDirectoryIntegrated%2A>|
+|<xref:Microsoft.Data.SqlClient.SqlAuthenticationMethod.ActiveDirectoryInteractive%2A>|
+|<xref:Microsoft.Data.SqlClient.SqlAuthenticationMethod.ActiveDirectoryServicePrincipal%2A>|
+|<xref:Microsoft.Data.SqlClient.SqlAuthenticationMethod.ActiveDirectoryDeviceCodeFlow%2A>|
+
+## Examples  
+ The following example demonstrates providing a custom device flow callback to SqlClient driver for Device Code Flow authentication method:
```
David-Engel, line 73:
```suggestion
 The following example demonstrates providing a custom device flow callback to SqlClient for the Device Code Flow authentication method:
```

doc/snippets/Microsoft.Data.SqlClient/ActiveDirectoryAuthenticationProvider.xml
```diff
@@ -0,0 +1,82 @@
...
+      <summary>Sets a custom Web UI that will let the user sign-in with Azure AD, present consent if needed, and get back the authorization code. Applicable when working with Active Directory Interactive authentication.</summary>
+    </SetCustomWebUI>
+    <BeforeLoad>
+      <param name="authentication">The authentication method.</param>
+      <summary>This method is called immediately before the provider is added to SQL drivers registry. </summary>
+      <remarks>Avoid performing long-waiting tasks in this method, since it can block other threads from accessing the provider registry.</remarks>
+    </BeforeLoad>
+    <BeforeUnload>
+      <param name="authentication">The authentication method.</param>
+      <summary>This method is called immediately before the provider is removed from the SQL drivers registry. </summary>
```
David-Engel, line 48:
```suggestion
      <summary>This method is called immediately before the provider is removed from the authentication provider registry. </summary>
```

doc/snippets/Microsoft.Data.SqlClient/ActiveDirectoryAuthenticationProvider.xml
```diff
@@ -0,0 +1,82 @@
...
+      <param name="iWin32WindowFunc">A function to return the current window.</param>
+      <summary>Sets a reference to the current <see cref="T:System.Windows.Forms.IWin32Window" /> that triggers the browser to be shown. Used to center the browser that pop-up onto this window."</summary>
+    </SetIWin32WindowFunc>
+    <SetCustomWebUI>
+      <param name="customWebUi">Customer implementation for the Web UI</param>
+      <summary>Sets a custom Web UI that will let the user sign-in with Azure AD, present consent if needed, and get back the authorization code. Applicable when working with Active Directory Interactive authentication.</summary>
+    </SetCustomWebUI>
+    <BeforeLoad>
+      <param name="authentication">The authentication method.</param>
+      <summary>This method is called immediately before the provider is added to SQL drivers registry. </summary>
```
David-Engel, line 43:
```suggestion
      <summary>This method is called immediately before the provider is added to authentication provider registry. </summary>
```

doc/snippets/Microsoft.Data.SqlClient/ActiveDirectoryAuthenticationProvider.xml
```diff
@@ -0,0 +1,82 @@
...
+      <remarks>Avoid performing long-waiting tasks in this method, since it can block other threads from accessing the provider registry.</remarks>
+    </BeforeLoad>
+    <BeforeUnload>
+      <param name="authentication">The authentication method.</param>
+      <summary>This method is called immediately before the provider is removed from the SQL drivers registry. </summary>
+      <remarks>For example, this method is called when a different provider with the same authentication method overrides this provider in the SQL drivers registry. Avoid performing long-waiting task in this method, since it can block other threads from accessing the provider registry.</remarks>
```
David-Engel, line 49:
```suggestion
      <remarks>For example, this method is called when a different provider with the same authentication method overrides this provider in the authentication provider registry. Avoid performing long-waiting task in this method, since it can block other threads from accessing the provider registry.</remarks>
```

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -30,6 +30,26 @@ public sealed partial class SqlNotificationRequest
...
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/ActiveDirectoryAuthenticationProvider.xml' path='docs/members[@name="ActiveDirectoryAuthenticationProvider"]/AcquireTokenAsync/*'/>
+        public override System.Threading.Tasks.Task<SqlAuthenticationToken> AcquireTokenAsync(SqlAuthenticationParameters parameters) { throw null; }
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/ActiveDirectoryAuthenticationProvider.xml' path='docs/members[@name="ActiveDirectoryAuthenticationProvider"]/SetDeviceCodeFlowCallback/*'/>
+        public void SetDeviceCodeFlowCallback(System.Func<Microsoft.Identity.Client.DeviceCodeResult, System.Threading.Tasks.Task> deviceCodeFlowCallbackMethod) { }
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/ActiveDirectoryAuthenticationProvider.xml' path='docs/members[@name="ActiveDirectoryAuthenticationProvider"]/SetCustomWebUI/*'/>
+        public void SetCustomWebUI(Microsoft.Identity.Client.Extensibility.ICustomWebUi customWebUi) { }
```
David-Engel, line 45: Having Microsoft.Identity.Client.Extensibility.ICustomWebUi as part of our public API feels like the future may hold the same problem that we have in the AKV provider API, which we are now breaking. However, I'm not sure creating our own interface will save us from much future pain if MSAL decides to change theirs. Ideas? Thought?

cheenamalhotra: We can do that by accepting delegate/callback for [AcquireAuthorizationCodeAsync](https://docs.microsoft.com/en-us/dotnet/api/microsoft.identity.client.extensibility.icustomwebui.acquireauthorizationcodeasync?view=azure-dotnet) and pass it internally with CustomWebUi instance?

cheenamalhotra: It would save us from API changes, but if their whole design changes some day, we'll have to drop this API based on MSAL version upgrade, but so far it doesn't seem to go that far.

cheenamalhotra: I updated the PR accordingly, it does open more possibilities for a more generic implementation.

doc/snippets/Microsoft.Data.SqlClient/ActiveDirectoryAuthenticationProvider.xml
```diff
@@ -0,0 +1,83 @@
...
+      <param name="parentActivityOrWindowFunc">The parent as an object, in order to be used from shared NetStandard assemblies.</param>
+      <summary>Sets a reference to the ViewController (if using Xamarin.iOS), Activity (if using Xamarin.Android) IWin32Window or IntPtr (if using .NET Framework). Used for invoking the browser for Active Directory Interactive authentication.</summary>
+      <remarks>Mandatory to be set only on Android. See https://aka.ms/msal-net-android-activity for further documentation and details.</remarks>
+    </SetParentActivityOrWindowFunc>
+    <SetIWin32WindowFunc>
+      <param name="iWin32WindowFunc">A function to return the current window.</param>
+      <summary>Sets a reference to the current <see cref="T:System.Windows.Forms.IWin32Window" /> that triggers the browser to be shown. Used to center the browser pop-up onto this window."</summary>
+    </SetIWin32WindowFunc>
+    <SetAcquireAuthorizationCodeAsyncCallback>
+      <param name="acquireAuthorizationCodeAsyncCallback">The callback method to be called by MSAL.NET to delegate the authentication code Web with the Secure Token Service (STS).</param>
```
David-Engel, line 38:
```suggestion
      <param name="acquireAuthorizationCodeAsyncCallback">The callback method to be called by MSAL.NET to delegate the Web user interface with the Secure Token Service (STS).</param>
```

doc/snippets/Microsoft.Data.SqlClient/ActiveDirectoryAuthenticationProvider.xml
```diff
@@ -0,0 +1,83 @@
...
+      <param name="parameters">The Active Directory authentication parameters passed to authentication providers.</param>
+      <summary>Acquires a security token from the authority.</summary>
+      <returns>Represents an asynchronous operation that returns the authentication token.</returns>
+    </AcquireTokenAsync>
+    <SetDeviceCodeFlowCallback>
+      <param name="deviceCodeFlowCallbackMethod">The callback method to be used when performing 'Active Directory Device Code Flow' authentication.</param>
+      <summary>Sets the callback method, overriding the default implementation that processes the result when performing 'Active Directory Device Code Flow' authentication.</summary>
+    </SetDeviceCodeFlowCallback>
+    <SetParentActivityOrWindowFunc>
+      <param name="parentActivityOrWindowFunc">The parent as an object, in order to be used from shared NetStandard assemblies.</param>
```
David-Engel, line 29:
```suggestion
      <param name="parentActivityOrWindowFunc">The parent as an object, in order to be used from shared .NET Standard assemblies.</param>
```

doc/samples/AADAuthenticationCustomDeviceFlowCallback.cs
```diff
@@ -0,0 +1,30 @@
...
+            SqlAuthenticationProvider authProvider = new ActiveDirectoryAuthenticationProvider(CustomDeviceFlowCallback);
+            SqlAuthenticationProvider.SetProvider(SqlAuthenticationMethod.ActiveDirectoryDeviceCodeFlow, authProvider);
+            using (SqlConnection sqlConnection = new SqlConnection("Server=<myserver>.database.windows.net;Authentication=Active Directory Device Code Flow;Database=<db>;"))
+            {
+                sqlConnection.Open();
+                Console.WriteLine("Connected successfully!");
+            }
+        }
+
+        private Task CustomDeviceFlowCallback(DeviceCodeResult result)
```
johnnypham, line 22:
```suggestion
        private static Task CustomDeviceFlowCallback(DeviceCodeResult result)
```

Must be static since it's being accessed in a static method above.

doc/samples/AADAuthenticationCustomDeviceFlowCallback.cs
```diff
@@ -0,0 +1,30 @@
...
+        {
+            // Provide custon logic to process result information and read device code.
```
johnnypham, line 24:
```suggestion
            // Provide custom logic to process result information and read device code.
```

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ActiveDirectoryAuthenticationProvider.cs
```diff
@@ -125,7 +209,9 @@ internal class ActiveDirectoryAuthenticationProvider : SqlAuthenticationProvider
             return new SqlAuthenticationToken(result.AccessToken, result.ExpiresOn);
         });
 
-        private async Task<AuthenticationResult> AcquireTokenInteractive(IPublicClientApplication app, string[] scopes, Guid connectionId, string userId)
+
+        private async Task<AuthenticationResult> AcquireTokenInteractiveDeviceFlow(IPublicClientApplication app, string[] scopes, Guid connectionId, string userId,
```
johnnypham, line 213: Do we use the `Async `suffix for private methods?

cheenamalhotra: I'll add it for clarity!

Issue #600, opened 2020-06-11
Bug: Async SequentialStream reads do not complete


Wraith2:
Based on the binary read benchmark in https://github.com/dotnet/SqlClient/issues/593#issuecomment-642833062 I tried the code below expecting the behaviour to be fast and found that it simply blocks indefinitely.

```csharp
            using var conn = new SqlConnection(ConnectionString);
            using var cmd = new SqlCommand("SELECT foo FROM data", conn);
            await conn.OpenAsync();
            using var reader = await cmd.ExecuteReaderAsync(
                System.Data.CommandBehavior.SequentialAccess
            );
            await reader.ReadAsync();
            using var stream =  reader.GetStream(0);
            using var memory = new MemoryStream(16 * 1024);

            await stream.CopyToAsync(memory);
            return (int)memory.Length;
```

This should work but freezes in a task wait after a single read cycle, the second read never completed. If you change it to standard mode it'll work but it does so by fetching the entire field and giving you a reader over the byte[]. So no workaround, this needs fixing if the team agree that it's a bug.

Edit: for clarity the problem is when multiple packets are needed. The current tests read from a single packet and thus do not do a network read.

cheenamalhotra:
Thanks @Wraith2 

Ack. Yes that does seem like an issue, if you have a proposed fix, let us know if you'd like to create PR, or we will start working on it soon.

cheenamalhotra:
Adding note: This issue is not reproducible in NetFx driver (for .NET Framework) so it should be investigated since when this issue has started to occur for .NET Core applications.

PR #603, opened 2020-06-14
Fix SqlSequentialStream multipacket reads stalling and add covering test

Wraith2:
Fixes https://github.com/dotnet/SqlClient/issues/600

When I reworked async infrastructure a couple of subtle errors crept in and because there was no covering test weren't visible. This corrects the problem and adds a test.

@cheenamalhotra @David-Engel I'd suggest that this go into 2.0.0 release.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
@@ -4517,14 +4517,23 @@ namespace Microsoft.Data.SqlClient
             SetTimeout(_defaultTimeoutMilliseconds);
 
             // Try to read without any continuations (all the data may already be in the stateObj's buffer)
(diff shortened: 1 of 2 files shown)


cheenamalhotra:
Hi @Wraith2 

This looks like a workaround but not the real issue fix.
The same code exists in NetFx project but does not cause any problems.

We should investigate when this issue was introduced in NetCore and fixing the real problem when it was introduced.

Wraith2:
The async changes I've made aren't in the fx codebase and they are what will have introduced the issue so I'm not surprised that it isn't reproducible on fx. 

It is intended to be the correct fix and not a workaround though. I debugged through for a quite a while and I was satisfied that there wasn't a huge underlying problem but that it was due to the changes from async captures to using a state object.

Wraith2:
It'll have been introduced in https://github.com/dotnet/SqlClient/pull/328

cheenamalhotra:
We'll take it in for GA.

Wraith2:
Excellent thanks.

PR #606, opened 2020-06-16
WIP | Suggested fix for Multiple Active Result sets in Managed SNI

JRahnama:
This might be the solution for Multiple Active Result Sets (MARS)  [Issue 422](https://github.com/dotnet/SqlClient/issues/422).

According to [Microsoft Documentation](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.taskcontinuationoptions?view=netcore-3.1) The `DenyChildAttach` will make each child task (that is, any nested inner task created by this continuation) to not be attached to parent task and to execute as a detach child task. Changing this to ExecuteSynchronously, Specifies that the continuation task should be executed synchronously. With this option specified, the continuation runs on the same thread that causes the antecedent task to transition into its final state. If the antecedent is already complete when the continuation is created, the continuation will run on the thread that creates the continuation.
@Wraith2

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIPacket.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIPacket.cs
@@ -257,7 +257,7 @@ namespace Microsoft.Data.SqlClient.SNI
                     continuationAction: _readCallback,
                     state: callback,
                     CancellationToken.None,
-                    TaskContinuationOptions.DenyChildAttach,
+                    TaskContinuationOptions.ExecuteSynchronously,
                     TaskScheduler.Default
                 );
         }


Wraith2:
It's worth a try but I'm sceptical.

from https://blog.stephencleary.com/2015/01/a-tour-of-task-part-7-continuations.html

"The “parenting options” control how the continuation task is attached to the antecedent task. Attached child tasks change the behavior of their parent task in ways that are convenient in some dynamic task parallelism scenarios, but are unexpected and awkward anywhere outside that (extremely small) use case. AttachedToParent will attach the continuation as a child task of the antecedent task. In modern code, you almost never want this option; more importantly, you almost never want other code to attach child tasks to your tasks. For this reason, the DenyChildAttach option was introduced in .NET 4.5, which prevents any continuations from using AttachedToParent.
"

So denying child attach is probably still the right thing to do. 
Allowing synchronous execution is probably ok in situations where you know you're not going to have to wait (so the data is already at the os level waiting to be picked up, same sort of situation you'd want to use a ValueTask I think) but if it isn't already there we may see perf problems. 

Lets see what happens.

JRahnama:
@Wraith2 

> Attached child tasks change the behavior of their parent task in ways that are convenient in some dynamic task parallelism scenarios, but are unexpected and awkward anywhere outside that (extremely small) use case.

This could be solved, in case other attempts failed (it just came to my mind right now and I dont know the side effects or any problem with it yet) by changing the design to have two different implementations inside  ReadFromStreamAsync with _fMars bool vaue from TdsParser. If Mars is activated then it attaches the child otherwise it will Deny that.


> Lets see what happens.
Totally agree. Time will show how the fix will go.

PR #608, opened 2020-06-17
WIP | Managed SNI, Add single threaded task scheduler 

Wraith2:
To avoid thread pool starvation handle receipts very specifically on a dedicated thread by using a custome task scheduler. No waits must be performed on this thread and execution must be pushed back to the standard thread pool for user callbacks and data handling.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITaskScheduler.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITaskScheduler.cs
@@ -1,0 +1,59 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Threading;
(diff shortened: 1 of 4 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITaskScheduler.cs
```diff
@@ -0,0 +1,59 @@
...
+            _thread.Start();
+        }
+
+        private void ExecuteThreadStart()
+        {
+            foreach (Task task in _tasks.GetConsumingEnumerable())
+            {
+                TryExecuteTask(task);
+            }
+        }
```
David-Engel, line 45: If anything throws an exception in here, does the thread die?

Wraith2: Probably. The code that can run on the thread is heavily constrained to a small part of the SNI receive loop and it's all inside this library so the opportunity for crashing should be quite low but I acknowledge the point that if this approach proves useful it would need protecting.

Wraith2:
Closing because unix testing has indicated that windows and unix SNI don't work the same way with this fix. It fixed the problem on windows only. /cc @JRahnama

PR #618, opened 2020-06-19
Merge more shared files

Wraith2:
This is another chunk of code mostly shared between the netfx and netcore codebases.
The majority of files needed tiny changes like Bool to bool, whitespace or other harmless changes. Two files I've put a [Serializable] attribute in a 'NETFRAMEWORK define after checking that the symbol is only defined (by the msbuild not the csproj files) on netfx builds.

The only file changed not merged and moved is SqlBuffer. I synced most harmless changes from core to netfx but left out functional changes.


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ApplicationIntent.cs
```diff
@@ -0,0 +1,19 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.Data.SqlClient
+{
+    /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/ApplicationIntent.xml' path='docs/members[@name="ApplicationIntent"]/ApplicationIntent/*'/>
+#if NETFRAMEWORK
```
cheenamalhotra, line 8: `NETFRAMEWORK` isn't recognized for .NET Framework implicitly.
You'll need to add `DefineConstants` in csproj and register this constant (similar to [here](https://github.com/dotnet/SqlClient/pull/597/files#diff-eb4be71042e9e5d81f7b29c13ab6d27cR23)).
Would recommend registering as `netfx` (uniform with folder arch and existing constants).

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/PoolBlockingPeriod.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.Data.SqlClient
+{
+#if NETFRAMEWORK
```
cheenamalhotra, line 7: Same as before:
NETFRAMEWORK isn't recognized for .NET Framework implicitly.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlBuffer.cs
```diff
@@ -42,55 +41,67 @@ internal enum StorageType
         internal struct DateTimeInfo
         {
             // This is used to store DateTime
-            internal Int32 daypart;
-            internal Int32 timepart;
+            internal int daypart;
```
cheenamalhotra, line 44: Could you exclude this file changes as part of this PR?
I don't mind the changes, but not necessarily here.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ApplicationIntent.cs
cheenamalhotra: @Wraith2 we need the constants applied in solution, which doesn't seem to happen when working in Visual Studio.

image

Including it in csproj will guarantee that.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlBuffer.cs
cheenamalhotra: I also wanted to mention.. there's a more changes than this required to make types uniform. I've seen this before but I believe a separate PR to make all primitive type changes would be better.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ApplicationIntent.cs
Wraith2: If you use /bl to get the binlog you can look at the full set of defines that are produced and you'll find NETFRAMEWORK is in there only if you're building for a framework target. There's a similar one for netcore but I didn't need that. So it doesn't need to be defined because it's already an integral part of the build system multi framework targeting.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlBuffer.cs
Wraith2: There are many other files that need primitive changes like this but I chose to try and keep this PR small so I stopped once I reached 30 changes. I'll revert this file an put it in another batch.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ApplicationIntent.cs
cheenamalhotra: I believe that's true, but my concern is for developer experience. If we continue to use this for future developments, it would never load source code that applies to .NET Framework in IDE when testing/debugging applications. So maybe IDE is to blame, but that constant isn't available by default in Visual Studio.

Wraith2: Ok, define added.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/PoolBlockingPeriod.cs
```diff
@@ -0,0 +1,22 @@
...
+    [System.Serializable]
+#endif
+    /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/PoolBlockingPeriod.xml' path='docs/members[@name="PoolBlockingPeriod"]/PoolBlockingPeriod/*'/>
```
cheenamalhotra, line 10:
```suggestion
    /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/PoolBlockingPeriod.xml' path='docs/members[@name="PoolBlockingPeriod"]/PoolBlockingPeriod/*'/>
#if NETFRAMEWORK
    [System.Serializable]
#endif
```

Wraith2: Done.

PR #626, opened 2020-06-24
Feature | Add support for Data Classification v2

cheenamalhotra:
Introduces Sensitivity Rank with Sensitivity Classification information.

_Additional Notes:_
The driver supports Max version of v2. So if server only supports v1 , based on received version in FEATUREEXTACK packages, we do not read Sensitivity Rank. Only if server supports v2, we read this new Rank info. The Rank value for queries made on v1 server is "NOT DEFINED". This way the driver supports both v1 and v2 of Data Classification support from Server.

Testing has been done with both v1 and v2 servers.

### DIFF
--- a/doc/snippets/Microsoft.Data.SqlClient.DataClassification/SensitivityClassification.xml
+++ b/doc/snippets/Microsoft.Data.SqlClient.DataClassification/SensitivityClassification.xml
@@ -3,26 +3,47 @@
     <members name="SensitivityClassification">
         <SensitivityClassification>
             <summary>Provides the functionlity to retrieve Sensitivity Classification data as received from SQL Server for the active 'SqlDataReader'</summary>
-            <remarks>To be added.</remarks>
         </SensitivityClassification>
         <ColumnSensitivities>
             <summary>Returns the column sensitivity for this 'SensitivityClassification' Object</summary>
             <value>List of column sensitivities.</value>
-            <remarks>To be added.</remarks>
         </ColumnSensitivities>
         <InformationTypes>
             <summary>Returns the information types collection for this 'SensitivityClassification' Object</summary>
             <value>List of information types.</value>
-            <remarks>To be added.</remarks>
         </InformationTypes>
         <Labels>
             <summary>Returns the labels collection for this 'SensitivityClassification' Object</summary>
             <value>List of labels.</value>
-            <remarks>To be added.</remarks>
         </Labels>
+        <SensitivityRank>
+          <summary>Returns the relative sensitivity rank for the query associated with the active 'SqlDataReader'.</summary>
+          <value>Relative sensitivity ranking for this query.</value>
+          <remarks>
+            <format type="text/markdown">
+              <![CDATA[
+
+## Remarks
+A relative ranking of the sensitivity of a query or of a column that is part of percolumn data. Available values are as below:
+
+| Sensitivity Rank | Description |
+| --- | --- |
+| -1 | Not Defined (default) |
+| 0 | None |
+| 10 | Low |
+| 20 | Medium |
+| 30 | High |
+| 40 | Critical |
+]]>
+            </format>
+          </remarks>
+        </SensitivityRank>
         <ctor>
-            <summary>Initializes a new instance of the <see cref="T:Microsoft.Data.SqlClient.DataClassification.SensitivityClassification" />
- class.</summary>
+            <summary>Initializes a new instance of the <see cref="T:Microsoft.Data.SqlClient.DataClassification.SensitivityClassification" /> class.</summary>
+            <param name="labels">List of labels.</param>
+            <param name="informationTypes">List of information types.</param>
+            <param name="columnSensitivity">List of column sensitivities.</param>
+            <param name="sensitivityRank">Relative sensitivity rank for the query associated with the active 'SqlDataReader'.</param>
         </ctor>
     </members>
 </docs>
--- a/doc/snippets/Microsoft.Data.SqlClient.DataClassification/SensitivityClassification.xml
+++ b/doc/snippets/Microsoft.Data.SqlClient.DataClassification/SensitivityClassification.xml
@@ -24,7 +24,7 @@
               <![CDATA[
 
 ## Remarks
-A relative ranking of the sensitivity of a query or of a column that is part of percolumn data. Available values are as below:
+A relative sensitivity ranking of the query. Available values are as below:
 
 | Sensitivity Rank | Description |
 | --- | --- |
--- a/src/Microsoft.Data.SqlClient/netfx/ref/Microsoft.Data.SqlClient.cs
+++ b/src/Microsoft.Data.SqlClient/netfx/ref/Microsoft.Data.SqlClient.cs
@@ -2194,7 +2194,7 @@ namespace Microsoft.Data.SqlClient.DataClassification
     public partial class SensitivityClassification
     {
         /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient.DataClassification/SensitivityClassification.xml' path='docs/members[@name="SensitivityClassification"]/ctor/*' />
-        public SensitivityClassification(System.Collections.Generic.IList<Microsoft.Data.SqlClient.DataClassification.Label> labels, System.Collections.Generic.IList<Microsoft.Data.SqlClient.DataClassification.InformationType> informationTypes, System.Collections.Generic.IList<Microsoft.Data.SqlClient.DataClassification.ColumnSensitivity> columnSensitivity, int sensitivityRank) { }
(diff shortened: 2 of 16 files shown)


src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -3075,7 +3076,7 @@ internal void OnFeatureExtAck(int featureId, byte[] data)
                             throw SQL.ParsingError(ParsingErrorState.CorruptedTdsStream);
                         }
                         byte enabled = data[1];
-                        _parser.DataClassificationVersion = (enabled == 0) ? TdsEnums.DATA_CLASSIFICATION_NOT_ENABLED : supportedDataClassificationVersion;
+                        _parser.DataClassificationVersion = (enabled == 0) ? TdsEnums.DATA_CLASSIFICATION_NOT_ENABLED : _supportedDataClassificationVersion;
```
Wraith2, line 3079: Is this copy of the same information needed? The parser and internal connection are very solidly linked together so I think it might be possible to just keep one copy and refer to it using a getter property if needed.

While working with the current library I find that the copying of data around makes auditing usage of variables very time consuming and the existing habit of using direct access to internal members of other classes does the same.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsEnums.cs
```diff
@@ -938,7 +938,8 @@ internal enum FedAuthInfoId : byte
 
         // Data Classification constants
         internal const byte DATA_CLASSIFICATION_NOT_ENABLED = 0x00;
-        internal const byte MAX_SUPPORTED_DATA_CLASSIFICATION_VERSION = 0x01;
+        internal const byte DATA_CLASSIFICATION_VERSION_WITHOUT_RANK_SUPPORT = 0x01;
+        internal const byte MAX_SUPPORTED_DATA_CLASSIFICATION_VERSION = 0x02;
```
Wraith2, line 942: If this isn't a translation of a macro from the Sql native headers it might be useful to rename it to something like 
`DATA_CLASSIFICATION_VERSION_MAX_SUPPORTED` so that the words data classification always come first.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -3777,6 +3777,20 @@ private bool TryProcessDataClassification(TdsParserStateObject stateObj, out Sen
                 informationTypes.Add(new InformationType(informationType, id));
             }
 
+            // get sensitivity rank
+            Int32 sensitivityRank = (int)SensitivityRank.NOT_DEFINED;
```
Wraith2, line 3781: nit: Int32->int

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsEnums.cs
cheenamalhotra: It isn't new, but I think it's no norm either. If you see underneath in next lines.. there's MAX/MIN TCE related enums. Probably driving factor for this one.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -3827,14 +3827,23 @@ private bool TryProcessDataClassification(TdsParserStateObject stateObj, out Sen
...
+                    int sensitivityRankProperty = (int)SensitivityRank.NOT_DEFINED;
+                    if (DataClassificationVersion > TdsEnums.DATA_CLASSIFICATION_VERSION_WITHOUT_RANK_SUPPORT)
+                    {
+                        if (!stateObj.TryReadInt32(out sensitivityRankProperty) || !Enum.IsDefined(typeof(SensitivityRank), sensitivityRankProperty))
+                        {
+                            return false;
+                        }
+                    }
+
                     // add sentivity properties for the source
```
lilgreenbird, line 3840: typo - sensitivity

doc/snippets/Microsoft.Data.SqlClient.DataClassification/SensitivityClassification.xml
```diff
@@ -3,26 +3,47 @@
...
         </Labels>
+        <SensitivityRank>
+          <summary>Returns the relative sensitivity rank for the query associated with the active 'SqlDataReader'.</summary>
+          <value>Relative sensitivity ranking for this query.</value>
+          <remarks>
+            <format type="text/markdown">
+              <![CDATA[
+
+## Remarks
+A relative ranking of the sensitivity of a query or of a column that is part of percolumn data. Available values are as below:
```
johnnypham, line 27: Since this is the rank for the reader, it should only mention the query? Same thing with column in SensitivityProperty.
```suggestion
A relative ranking of the sensitivity of a query. Available values are as below:
```

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/DataClassification/SensitivityClassification.cs
```diff
@@ -41,6 +41,23 @@ public InformationType(string name, string id)
...
+        /// <include file='../../../../../../../doc/snippets/Microsoft.Data.SqlClient.DataClassification/SensitivityRank.xml' path='docs/members[@name="SensitivityRank"]/Medium/*' />
+        MEDIUM = 20,
+        /// <include file='../../../../../../../doc/snippets/Microsoft.Data.SqlClient.DataClassification/SensitivityRank.xml' path='docs/members[@name="SensitivityRank"]/High/*' />
+        HIGH = 30,
+        /// <include file='../../../../../../../doc/snippets/Microsoft.Data.SqlClient.DataClassification/SensitivityRank.xml' path='docs/members[@name="SensitivityRank"]/Critical/*' />
+        CRITICAL = 40
+    }
+
     /// <include file='../../../../../../../doc/snippets/Microsoft.Data.SqlClient.DataClassification/SensitivityProperty.xml' path='docs/members[@name="SensitivityProperty"]/SensitivityProperty/*' />
     public class SensitivityProperty
```
DavoudEshtehari, line 62: From the security point of view, we seal a public concrete class. I found a bunch of `public classes` without the `sealed` keyword. Can I ask you to explain If you did it by purpose?

DavoudEshtehari: Also, I suggest the `struct` instead of `class` for the simple data type like this. If you don't want to support inheritance.

cheenamalhotra: > From the security point of view, we seal a public concrete class. I found a bunch of public classes without the sealed keyword. Can I ask you to explain If you did it by purpose?

I didn't write these classes, they were added while ago before open-soucing. But sure I can update things.

cheenamalhotra: > Also, I suggest the `struct` instead of `class` for the simple data type like this. If you don't want to support inheritance.

Struct is stored in stack, I don't think this metadata should be stored in stack as can lead to stack overflow for 1000s of fields of data.

src/Microsoft.Data.SqlClient/netfx/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -1921,332 +1921,352 @@ public partial class SqlDataRecord : System.Data.IDataRecord
...
-        public SensitivityClassification(System.Collections.Generic.IList<Microsoft.Data.SqlClient.DataClassification.Label> labels, System.Collections.Generic.IList<Microsoft.Data.SqlClient.DataClassification.InformationType> informationTypes, System.Collections.Generic.IList<Microsoft.Data.SqlClient.DataClassification.ColumnSensitivity> columnSensitivity) { }
-        /// <include file='..\..\..\..\doc\snippets\Microsoft.Data.SqlClient.DataClassification\SensitivityClassification.xml' path='docs/members[@name="SensitivityClassification"]/ColumnSensitivities/*' />
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient.DataClassification/SensitivityClassification.xml' path='docs/members[@name="SensitivityClassification"]/ctor/*' />
+        public SensitivityClassification(System.Collections.Generic.IList<Microsoft.Data.SqlClient.DataClassification.Label> labels, System.Collections.Generic.IList<Microsoft.Data.SqlClient.DataClassification.InformationType> informationTypes, System.Collections.Generic.IList<Microsoft.Data.SqlClient.DataClassification.ColumnSensitivity> columnSensitivity, int sensitivityRank) { }
```
karinazhou, line 2197: `SensitivityRank sensitivityRank` instead of `int sensitivityRank`?

PR #628, opened 2020-06-26
[Release/2.0] Fix | Fix Enclave Session Cache Issue with Azure Database

karinazhou:
This PR is to fix the issue when connecting to Azure SQL databases with secure enclaves. 

The initial enclave session cache key only contains the enclave attestation URL and the server name. It works well with on-prem SQL Servers if we switch between various databases under the same server. However, this will fail for Azure SQL. Sharing the same enclave session among different databases under the same logic Azure server will cause the Azure attestation to fail.  

To fix this issue, the database name is also added to the enclave session cache key.

### DIFF
--- a/doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionEnclaveProvider.xml
+++ b/doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionEnclaveProvider.xml
@@ -21,7 +21,7 @@ the enclave attestation protocol as well as the logic for creating and caching e
             <param name="clientDiffieHellmanKey">A Diffie-Hellman algorithm object that encapsulates a client-side key pair.</param>
             <param name="attestationUrl">The endpoint of an attestation service for attesting the enclave.</param>
             <param name="servername">The name of the SQL Server instance containing the enclave.</param>
-            <param name="database">The database that SqlClient contacts to.</param>
+            <param name="database">The database that SqlClient contacts to request an enclave session.</param>
             <param name="customData">The set of extra data needed for attestating the enclave.</param>
             <param name="customDataLength">The length of the extra data needed for attestating the enclave.</param>
             <param name="sqlEnclaveSession">The requested enclave session or <see langword="null" /> if the provider doesn't implement session caching.</param>
--- a/doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionEnclaveProvider.xml
+++ b/doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionEnclaveProvider.xml
@@ -40,7 +40,7 @@ the enclave attestation protocol as well as the logic for creating and caching e
         <GetEnclaveSession>
             <param name="serverName">The name of the SQL Server instance containing the enclave.</param>
             <param name="attestationUrl">The endpoint of an attestation service, SqlClient contacts to attest the enclave.</param>
-            <param name="database">The database that SqlClient contacts to.</param>
+            <param name="database">The database that SqlClient contacts to request an enclave session.</param>
             <param name="generateCustomData"><see langword="true" /> to indicate that a set of extra data needs to be generated for attestation; otherwise, <see langword="false" />.</param>
             <param name="sqlEnclaveSession">When this method returns, the requested enclave session or <see langword="null" /> if the provider doesn't implement session caching. This parameter is treated as uninitialized.</param>
             <param name="counter">A counter that the enclave provider is expected to increment each time SqlClient retrieves the session from the cache. The purpose of this field is to prevent replay attacks.</param>
--- a/doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionEnclaveProvider.xml
+++ b/doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionEnclaveProvider.xml
@@ -53,7 +53,7 @@ the enclave attestation protocol as well as the logic for creating and caching e
         <InvalidateEnclaveSession>
             <param name="serverName">The name of the SQL Server instance containing the enclave.</param>
             <param name="enclaveAttestationUrl">The endpoint of an attestation service, SqlClient contacts to attest the enclave.</param>
-            <param name="database">The database that SqlClient contacts to.</param>
+            <param name="database">The database that SqlClient contacts to request an enclave session.</param>
             <param name="enclaveSession">The session to be invalidated.</param>
             <summary>When overridden in a derived class, looks up and evicts an enclave session from the enclave session cache, if the provider implements session caching.</summary>
             <remarks>To be added.</remarks>
--- a/doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionEnclaveProvider.xml
+++ b/doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionEnclaveProvider.xml
@@ -21,6 +21,7 @@ the enclave attestation protocol as well as the logic for creating and caching e
             <param name="clientDiffieHellmanKey">A Diffie-Hellman algorithm object that encapsulates a client-side key pair.</param>
             <param name="attestationUrl">The endpoint of an attestation service for attesting the enclave.</param>
             <param name="servername">The name of the SQL Server instance containing the enclave.</param>
+            <param name="database">The database that SqlClient contacts to.</param>
             <param name="customData">The set of extra data needed for attestating the enclave.</param>
             <param name="customDataLength">The length of the extra data needed for attestating the enclave.</param>
             <param name="sqlEnclaveSession">The requested enclave session or <see langword="null" /> if the provider doesn't implement session caching.</param>
@@ -39,6 +40,7 @@ the enclave attestation protocol as well as the logic for creating and caching e
         <GetEnclaveSession>
             <param name="serverName">The name of the SQL Server instance containing the enclave.</param>
             <param name="attestationUrl">The endpoint of an attestation service, SqlClient contacts to attest the enclave.</param>
+            <param name="database">The database that SqlClient contacts to.</param>
             <param name="generateCustomData"><see langword="true" /> to indicate that a set of extra data needs to be generated for attestation; otherwise, <see langword="false" />.</param>
             <param name="sqlEnclaveSession">When this method returns, the requested enclave session or <see langword="null" /> if the provider doesn't implement session caching. This parameter is treated as uninitialized.</param>
             <param name="counter">A counter that the enclave provider is expected to increment each time SqlClient retrieves the session from the cache. The purpose of this field is to prevent replay attacks.</param>
@@ -51,6 +53,7 @@ the enclave attestation protocol as well as the logic for creating and caching e
         <InvalidateEnclaveSession>
             <param name="serverName">The name of the SQL Server instance containing the enclave.</param>
             <param name="enclaveAttestationUrl">The endpoint of an attestation service, SqlClient contacts to attest the enclave.</param>
+            <param name="database">The database that SqlClient contacts to.</param>
             <param name="enclaveSession">The session to be invalidated.</param>
             <summary>When overridden in a derived class, looks up and evicts an enclave session from the enclave session cache, if the provider implements session caching.</summary>
             <remarks>To be added.</remarks>
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/EnclaveDelegate.NetCoreApp.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/EnclaveDelegate.NetCoreApp.cs
@@ -64,11 +64,12 @@ namespace Microsoft.Data.SqlClient
         /// <param name="enclaveType">enclave type</param>
         /// <param name="serverName">servername</param>
         /// <param name="attestationUrl">attestation url for attestation service endpoint</param>
+        /// <param name="database">The database that SqlClient contacts to.</param>
         /// <param name="attestationInfo">attestation info from SQL Server</param>
         /// <param name="attestationParameters">attestation parameters</param>
         /// <param name="customData">A set of extra data needed for attestating the enclave.</param>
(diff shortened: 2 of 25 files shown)


src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/EnclaveSessionCache.cs
```diff
@@ -67,9 +67,9 @@ internal SqlEnclaveSession CreateSession(string attestationUrl, string serverNam
         }
 
         // Generates the cache key for the enclave session cache
-        private string GenerateCacheKey(string serverName, string attestationUrl)
+        private string GenerateCacheKey(string serverName, string attestationUrl, string database)
```
ErikEJ, line 70: Maybe use a parameters class instead of passing three strings?

karinazhou: The strings we used are passed from multiple layers of function calls before. From my point of view, it's straightforward to keep using them here instead of wrapping them in an extra class before generating the cache key. Is there any benefit of using a class for this specific function?

ErikEJ: I think a good reason could be to avoid/make it less likely  passing parameters in the wrong order (since all parameters are strings)

karinazhou: Thank you for your explanation. I understand the concerns. If we are going to wrap these strings, I think we need to do it before we pass them to the enclave delegate. It requires the corresponding changes in all enclave providers too.

ErikEJ: It was just a suggestion.

karinazhou: We value your suggestion. The current APIs for enclave providers are inherited from System.Data.SqlClient.SqlColumnEncryptionEnclaveProvider interface. To avoid passing multiple strings for server name, attestation URL, and database name, the APIs need to be updated. It is doable for Microsoft.Data.SqlClient since we have made these APIs internal.

This PR is a hotfix for the GA-2.0.0 which introduces minor impacts on the released driver. I will bring up your suggestions within the team and will probably create a separate PR for the API changes if we are good to go with it.

cheenamalhotra: We will consider this design change in PR for 'master' branch.

cheenamalhotra:
cc @Xtrimmer

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/EnclaveAzureDatabaseTests.cs
```diff
@@ -0,0 +1,182 @@
...
+                    using (SqlConnection sqlConnection = new SqlConnection(connStr))
+                    {
+                        sqlConnection.Open();
+                        databaseObjects.ForEach(o => o.Drop(sqlConnection));
+                    }
+                }
+            }
+        }
+    }    
+}
```
cheenamalhotra, line 182: New Line EOF

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/EnclaveAzureDatabaseTests.cs
```diff
@@ -0,0 +1,182 @@
...
+                akvColumnMasterKey = new AkvColumnMasterKey(DatabaseHelper.GenerateUniqueName("AKVCMK"), akvUrl: DataTestUtility.AKVUrl, sqlColumnEncryptionAzureKeyVaultProvider, DataTestUtility.EnclaveEnabled);
+                databaseObjects.Add(akvColumnMasterKey);
+
+                akvColumnEncryptionKey= new ColumnEncryptionKey(DatabaseHelper.GenerateUniqueName("AKVCEK"),
+                                                              akvColumnMasterKey,
+                                                              sqlColumnEncryptionAzureKeyVaultProvider);
+                databaseObjects.Add(akvColumnEncryptionKey);
+
+                SqlConnectionStringBuilder connString1 = new SqlConnectionStringBuilder(DataTestUtility.EnclaveAzureDatabaseConnString);
+                connString1.InitialCatalog = "testdb001";
```
cheenamalhotra, line 46: This test class is very specific to our DB setup, so also add that in comments above as "For Internal Use only"

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
```diff
@@ -284,6 +289,8 @@ public static bool IsSupportedDataClassification()
 
         public static bool IsDNSCachingSetup() => !string.IsNullOrEmpty(DNSCachingConnString);
 
+        public static bool IsEnclaveAzureDatabaseSetup() => !string.IsNullOrEmpty(EnclaveAzureDatabaseConnString);
```
cheenamalhotra, line 292: The "EnclaveEnabled" keyword in config.json would mislead this case, it would be better to also include that in the condition.

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/EnclaveAzureDatabaseTests.cs
```diff
@@ -0,0 +1,182 @@
...
+             
+        public EnclaveAzureDatabaseTests()
+        {
+            if (DataTestUtility.IsEnclaveAzureDatabaseSetup() && DataTestUtility.EnclaveEnabled)
+            {
+                // Initialize AKV provider
+                sqlColumnEncryptionAzureKeyVaultProvider = new SqlColumnEncryptionAzureKeyVaultProvider(AADUtility.AzureActiveDirectoryAuthenticationCallback);
+
+                // Register AKV provider
+                SqlConnection.RegisterColumnEncryptionKeyStoreProviders(customProviders: new Dictionary<string, SqlColumnEncryptionKeyStoreProvider>(capacity: 1, comparer: StringComparer.OrdinalIgnoreCase)
```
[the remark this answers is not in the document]
karinazhou: I have seen our `SQLSetupStrategyAzureKeyVault` class will register the AKVProvider. There is a boolean `isAKVProviderRegistered` to indicate the registration status. I can do a check in the newly added tests to see whether we need to register or not.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
karinazhou: Regarding this `EnclaveEnabled` keyword, we don't actually need it for this test because the Azure DB in the connection string is always enclave-enabled. The connection string check should be enough. I will remove it in the test.

karinazhou: Well, after testing locally, I have seen failures running this test since our current ColumnMasterKey generation will check the DataTestUtility.EnclaveEnabled to determine the key type (allow rich computation or not). I will add this to the test condition as you suggested.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/EnclaveSessionCache.cs
```diff
@@ -67,9 +67,9 @@ internal SqlEnclaveSession CreateSession(string attestationUrl, string serverNam
...
         {
-            return (serverName + attestationUrl).ToLowerInvariant();
+            return (serverName + database + attestationUrl).ToLowerInvariant();
```
Xtrimmer, line 72: It would be good to add a delimiter character that is not possible to be in the database/server identifier. For example:
If you have a server named `AAB` and a database named `BCC` then together you get `AABBCC`.
Likewise, if you have a server named `AABB` and a database named CC then you still get `AABBCC`
But with a delimiter like '+' then the concatenations would be different: `AAB+BCC` vs `AABB+CC`

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/EnclaveSessionCache.NetCoreApp.cs
```diff
@@ -67,9 +67,9 @@ internal SqlEnclaveSession CreateSession(string attestationUrl, string serverNam
         }
 
         // Generates the cache key for the enclave session cache
-        private string GenerateCacheKey(string serverName, string attestationUrl)
+        private string GenerateCacheKey(string serverName, string attestationUrl, string database)
         {
-            return (serverName + attestationUrl).ToLowerInvariant();
+            return (serverName + database + attestationUrl).ToLowerInvariant();
```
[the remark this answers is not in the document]
karinazhou: @Xtrimmer Good point. I will add the delimiter in between.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/EnclaveSessionCache.NetCoreApp.cs
```diff
@@ -69,7 +69,7 @@ internal SqlEnclaveSession CreateSession(string attestationUrl, string serverNam
         // Generates the cache key for the enclave session cache
         private string GenerateCacheKey(string serverName, string attestationUrl, string database)
         {
-            return (serverName + database + attestationUrl).ToLowerInvariant();
+            return (serverName + '+' + database + attestationUrl).ToLowerInvariant();
```
ErikEJ, line 72: Can a database name be case sensitive?

karinazhou: No, it is not. You can test it from SSMS.

doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionEnclaveProvider.xml
```diff
@@ -21,6 +21,7 @@ the enclave attestation protocol as well as the logic for creating and caching e
             <param name="clientDiffieHellmanKey">A Diffie-Hellman algorithm object that encapsulates a client-side key pair.</param>
             <param name="attestationUrl">The endpoint of an attestation service for attesting the enclave.</param>
             <param name="servername">The name of the SQL Server instance containing the enclave.</param>
+            <param name="database">The database that SqlClient contacts to.</param>
```
cheenamalhotra, line 24: I see some inconsistencies in description.. could you fix them plz?
```suggestion
            <param name="database">The database that SqlClient contacts to request an enclave session.</param>
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/EnclaveDelegate.NetCoreApp.cs
```diff
@@ -64,11 +64,12 @@ internal byte[] GetSerializedAttestationParameters(SqlEnclaveAttestationParamete
         /// <param name="enclaveType">enclave type</param>
         /// <param name="serverName">servername</param>
         /// <param name="attestationUrl">attestation url for attestation service endpoint</param>
+        /// <param name="database">The database that SqlClient contacts to.</param>
```
cheenamalhotra, line 67:
```suggestion
        /// <param name="database">The database that SqlClient contacts to request an enclave session.</param>
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SimulatorEnclaveProvider.NetCoreApp.cs
```diff
@@ -107,10 +107,11 @@ public override void CreateEnclaveSession(byte[] attestationInfo, ECDiffieHellma
         /// </summary>
         /// <param name="serverName">The name of the SQL Server instance containing the enclave.</param>
         /// <param name="enclaveAttestationUrl">The endpoint of an attestation service, SqlClient contacts to attest the enclave.</param>
+        /// <param name="database">The database that SqlClient contacts to.</param>
```
cheenamalhotra, line 110:
```suggestion
        /// <param name="database">The database that SqlClient contacts to request an enclave session.</param>
```

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/EnclaveDelegate.cs
```diff
@@ -42,8 +42,9 @@ internal class EnclaveDelegate
         /// <param name="enclaveType">enclave type</param>
         /// <param name="serverName">server name</param>
         /// <param name="enclaveAttestationUrl">url for attestation endpoint</param>
+        /// <param name="database">The database that SqlClient contacts to.</param>
```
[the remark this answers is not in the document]
cheenamalhotra: [Pending]

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/EnclaveDelegate.cs
```diff
@@ -159,11 +160,12 @@ private byte[] GetUintBytes(string enclaveType, int intValue, string variableNam
         /// <param name="enclaveType">enclave type</param>
         /// <param name="serverName">servername</param>
         /// <param name="attestationUrl">attestation url for attestation service endpoint</param>
+        /// <param name="database">The database that SqlClient contacts to.</param>
```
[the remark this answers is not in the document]
cheenamalhotra: [Pending]

Issue #294, opened 2019-10-31
AzureKeyVaultProvider should move to the latest Azure KeyVault Libraries


Petermarcu:
### Describe the bug
AzureKeyVaultProvider should move to the latest Azure KeyVault Libraries.

https://www.nuget.org/packages/Azure.Security.KeyVault.Keys
https://www.nuget.org/packages/Azure.Security.KeyVault.Secrets

Petermarcu:
@AlexGhiondea

Petermarcu:
@David-Engel @cheenamalhotra

m1nkeh:
but for why.. is there actually something which will be fixed, or just for the fun? :)

heaths:
Azure.Security.KeyVault.* packages will be our focus going forward, including any new features that the Key Vault service provides. It's also built around Azure.Core which has unified tracing, logging, retries, and more built into the pipeline. We also have a cross-language focus to provide consistency across .NET, Python, JavaScript (both Node and in-browser), and Java, with more languages on the horizon so that, if you need to interact with other languages or frameworks, you have a similar developer experience.

kmscode:
The [Microsoft.Azure.KeyVault package](https://www.nuget.org/packages/Microsoft.Azure.KeyVault/) has been officially deprecated for a number of months now (coming up on a year if it was deprecated when the last build was posted).

Would be nice to have fewer dependencies and everything on current versions. Thanks!

David-Engel:
Azure.Core (the main dependency under the new AKV packages) only supports .NET Framework 4.6.1 and up. Microsoft.Data.SqlClient 2.1 supports .NET Framework 4.6 and up. In order to move to the new libraries, SqlClient has to make a breaking change, dropping support for 4.6.0. So it will only happen on a major version bump.

PR #630, opened 2020-06-30
Feature | AKV Provider Upgrade to use new Azure key Vault libraries

cheenamalhotra:
Addresses #294 

Changes full design and introduces new constructors to accept implementation of `TokenCredential` to support new libraries.

**Breaking Changes:**
- Drops support for .NET Framework 4.6, supports .NET Framework 4.6.1+
- Drops support for old constructors that accepted "AuthenticationCallback".

cc @xtrimmer

**[6 Aug, 2020] Update**:
The old constructors are now removed totally and driver does not implement/support authentication callback in new design now. For user applications willing to support multi-user credentials for acquiring tokens, can implement a custom TokenCredential class with their own logic in "GetToken" and "GetTokenAsync" APIs. An example has been added (doc/samples/AzureKeyVaultProviderLegacyExample_2_0.cs) to the repository to demonstrate the same, will be published to Microsoft Docs.


src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/AzureSqlKeyCryptographer.cs
```diff
@@ -0,0 +1,232 @@
...
+        /// Validates and parses the Azure Key Vault URI and key name.
+        /// </summary>
+        /// <param name="masterKeyPath">The Azure Key Vault key identifier</param>
+        /// <param name="vaultUri">The Azure Key Vault URI</param>
+        /// <param name="masterKeyName">The name of the key</param>
+        private void ParseAKVPath(string masterKeyPath, out Uri vaultUri, out string masterKeyName)
+        {
+            Uri masterKeyPathUri = new Uri(masterKeyPath);
+            vaultUri = new Uri(masterKeyPathUri.GetLeftPart(UriPartial.Authority));
+            masterKeyName = masterKeyPathUri.Segments[2];
```
Xtrimmer, line 229: Is the master key path getting validated that it is the right format somewhere? We are not sure that Segment[2] exists by the time we are here, right? I think we were validating these things when the key URI was entering via the public provider interface methods but I do not see where the validation is happening.

Update: I later found that this was in the `ValidateNonEmptyAKVPath`

src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/SqlColumnEncryptionAzureKeyVaultProvider.cs
```diff
@@ -230,45 +233,36 @@ public override byte[] DecryptColumnEncryptionKey(string masterKeyPath, string e
...
+            byte[] signature = encryptedColumnEncryptionKey.Skip(currentIndex).Take(signatureLength).ToArray();
 
             // Compute the hash to validate the signature
-            byte[] hash;
-            using (SHA256 sha256 = SHA256.Create())
-            {
-                sha256.TransformFinalBlock(encryptedColumnEncryptionKey, 0, encryptedColumnEncryptionKey.Length - signature.Length);
-                hash = sha256.Hash;
-            }
+            byte[] hash = encryptedColumnEncryptionKey.Take(encryptedColumnEncryptionKey.Length - signatureLength).ToArray();
```
Xtrimmer, line 251: This is still the preimage at this point and not a hash yet since keyvault does the hashing now. The new library makes this variable name misleading now.

src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/SqlColumnEncryptionAzureKeyVaultProvider.cs
```diff
@@ -282,173 +276,69 @@ public override byte[] DecryptColumnEncryptionKey(string masterKeyPath, string e
...
+            KeyCryptographer.AddKey(masterKeyPath);
+            int keySizeInBytes = KeyCryptographer.GetKeySize(masterKeyPath);
 
             // Construct the encryptedColumnEncryptionKey
             // Format is 
             //          version + keyPathLength + ciphertextLength + ciphertext + keyPath + signature
             //
             // We currently only support one version
-            byte[] version = new byte[] { firstVersion[0] };
+            byte[] version = new byte[] { s_firstVersion[0] };
```
Xtrimmer, line 294: `s_firstVersion` is alredy a byte array. No need to unwrap it and then put it back into a new byte array.

From the member variables above:
`private readonly static byte[] s_firstVersion = new byte[] { 0x01 };`

src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/SqlColumnEncryptionAzureKeyVaultProvider.cs
```diff
@@ -282,173 +276,69 @@ public override byte[] DecryptColumnEncryptionKey(string masterKeyPath, string e
...
-            using (SHA256 sha256 = SHA256.Create())
-            {
-                sha256.TransformBlock(version, 0, version.Length, version, 0);
-                sha256.TransformBlock(keyPathLength, 0, keyPathLength.Length, keyPathLength, 0);
-                sha256.TransformBlock(cipherTextLength, 0, cipherTextLength.Length, cipherTextLength, 0);
-                sha256.TransformBlock(masterKeyPathBytes, 0, masterKeyPathBytes.Length, masterKeyPathBytes, 0);
-                sha256.TransformFinalBlock(cipherText, 0, cipherText.Length);
-                hash = sha256.Hash;
-            }
+            byte[] hash = version.Concat(keyPathLength).Concat(cipherTextLength).Concat(masterKeyPathBytes).Concat(cipherText).ToArray();
```
Xtrimmer, line 311: This is still the preimage at this point and not a hash yet since keyvault does the hashing now. The new library makes this variable name misleading now.

src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/SqlColumnEncryptionAzureKeyVaultProvider.cs
```diff
@@ -282,173 +276,69 @@ public override byte[] DecryptColumnEncryptionKey(string masterKeyPath, string e
...
-            using (SHA256 sha256 = SHA256.Create())
-            {
-                sha256.TransformFinalBlock(masterkeyMetadataBytes, 0, masterkeyMetadataBytes.Length);
-                hash = sha256.Hash;
-            }
-            return hash;
-        }
 
         /// <summary>
         /// Checks if the Azure Key Vault key path is Empty or Null (and raises exception if they are).
```
Xtrimmer, line 332: Nitpick: It looks like this does a bit more than simply check for null/empty now. Maybe update the summary and or method name to reflect this.

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/AKVUnitTests.cs
```diff
@@ -0,0 +1,67 @@
...
+            SqlColumnEncryptionAzureKeyVaultProvider oldAkvProvider = new SqlColumnEncryptionAzureKeyVaultProvider(AzureActiveDirectoryAuthenticationCallback);
+
+            byte[] encryptedCekWithNewProvider = newAkvProvider.EncryptColumnEncryptionKey(DataTestUtility.AKVUrl, EncryptionAlgorithm, ColumnEncryptionKey);
+            byte[] decryptedCekWithOldProvider = oldAkvProvider.DecryptColumnEncryptionKey(DataTestUtility.AKVUrl, EncryptionAlgorithm, encryptedCekWithNewProvider);
+            Assert.Equal(ColumnEncryptionKey, decryptedCekWithOldProvider);
+
+            byte[] encryptedCekWithOldProvider = oldAkvProvider.EncryptColumnEncryptionKey(DataTestUtility.AKVUrl, EncryptionAlgorithm, ColumnEncryptionKey);
+            byte[] decryptedCekWithNewProvider = newAkvProvider.DecryptColumnEncryptionKey(DataTestUtility.AKVUrl, EncryptionAlgorithm, encryptedCekWithOldProvider);
+            Assert.Equal(ColumnEncryptionKey, decryptedCekWithNewProvider);
+        }
```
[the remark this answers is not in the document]
cheenamalhotra: Yes, it's different now, and our goal now is to make sure constructors work interchangeably.
Also because we cannot have both implementations simultaneously referenced.

If you think there's no value in it, I can remove it.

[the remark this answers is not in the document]
Xtrimmer: Yes there is definitely value in this test. It is a shame that we cannot test that older version of the AKV provider works interchangeably with the new version using the new AKV client too. This seems important.

src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/AzureKeyVaultProviderTokenCredential.cs
```diff
@@ -0,0 +1,64 @@
...
+            string[] pairs = trimmedChallenge.Split(new string[] { "," }, StringSplitOptions.RemoveEmptyEntries);
+
+            Authority = pairs[0].Split('=')[1].Trim().Trim(new char[] { '\"' });
+            Resource = pairs[1].Split('=')[1].Trim().Trim(new char[] { '\"' });
+        }
+
+        public override AccessToken GetToken(TokenRequestContext requestContext, CancellationToken cancellationToken)
+        {
+            string token = Callback.Invoke(Authority, Resource, string.Empty).GetAwaiter().GetResult();
+            return new AccessToken(token, DateTimeOffset.Now.AddHours(1));
```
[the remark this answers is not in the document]
cheenamalhotra: I wanted to discuss about this actually, good you brought it up.
Do we know how AKV consumes these tokens and what would be expected lifetime of this token?
Also should we cache tokens here in our implementation?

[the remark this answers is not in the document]
cheenamalhotra: Also, we have "ColumnEncryptionKeyCacheTtl" connection property, that sets expiration time on cache of encryption key entries. Just some thoughts to run by..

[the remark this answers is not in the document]
Xtrimmer: Good questions. I do not know how AKV consumes these tokens under the hood. It may take some digging into their open source code to fully understand. As for the caching, I will need to ask around. I was not aware of that connection property before you mentioned it.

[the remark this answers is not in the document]
cheenamalhotra: @heaths @Petermarcu 

We'd require some assistance from Azure key Vault side to know if there's a better way to handle Expiration Time? Could you explain us how AKV (old design) set Expiration Time on Access tokens requested by calling "[AuthenticationContext.AcquireTokenAsync](https://docs.microsoft.com/en-us/dotnet/api/microsoft.identitymodel.clients.activedirectory.authenticationcontext.acquiretokenasync)" so we can keep the experience same for Authentication Callback support in the new design?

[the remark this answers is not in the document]
heaths: Key Vault uses challenge-based authentication using the initial tokens, which we cache and invalidate after they expire based on the challenge returned. See https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/keyvault/Azure.Security.KeyVault.Shared/src/ChallengeBasedAuthenticationPolicy.cs for details. I'll take a look through this PR as well for more context. In general, if you use the new Azure.Security.KeyVault.* packages, you shouldn't need to implement anything on your own. The Azure.Identity package - notably `DefaultAzureCredential` - efficiently and effectively supports both developer and production workflows.

src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/AzureKeyVaultProviderTokenCredential.cs
```diff
@@ -0,0 +1,90 @@
...
+        private AuthenticationCallback Callback { get; set; }
+
+        private string Authority { get; set; }
+
+        private string Resource { get; set; }
+
+        internal AzureKeyVaultProviderTokenCredential(AuthenticationCallback authenticationCallback, string masterKeyPath)
+        {
+            Callback = authenticationCallback;
+            using (HttpClient httpClient = new HttpClient())
```
heaths, line 29: Instantiating `HttpClient` can be expensive. I recommend declaring this once and using it both here and in the `KeyClient`, which you can do using the `KeyClientOptions` to set HTTP transport options, including which `HttpClient` to use.

src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/AzureKeyVaultProviderTokenCredential.cs
```diff
@@ -0,0 +1,90 @@
...
+                        string[] pair = pairs[i]?.Split('=');
+
+                        if (pair.Length == 2)
+                        {
+                            string key = pair[0]?.Trim().Trim(new char[] { '\"' });
+                            string value = pair[1]?.Trim().Trim(new char[] { '\"' });
+
+                            if (!string.IsNullOrEmpty(key))
+                            {
+                                if (key.Equals("authorization", StringComparison.InvariantCultureIgnoreCase))
```
heaths, line 49: There are a couple of other possibilities from older systems. I'm not sure if it's really important to handle those in practical cases, but see our ChallengeBasedAuthenticationPolicy for what we did.

src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/AzureKeyVaultProviderTokenCredential.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        public override AccessToken GetToken(TokenRequestContext requestContext, CancellationToken cancellationToken)
+        {
+            string token = Callback.Invoke(Authority, Resource, string.Empty).GetAwaiter().GetResult();
+            return new AccessToken(token, DateTimeOffset.Now.AddHours(1));
+        }
+
+        public override ValueTask<AccessToken> GetTokenAsync(TokenRequestContext requestContext, CancellationToken cancellationToken)
+        {
+            string token = Callback.Invoke(Authority, Resource, string.Empty).GetAwaiter().GetResult();
```
heaths, line 72: Why not just declare an `async` method and `await` the results here? In async library code, you most often should call `ConfigureAwait(false)` to avoid potential hangs in calling applications.

src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/AzureKeyVaultProviderTokenCredential.cs
```diff
@@ -0,0 +1,90 @@
...
+                        }
+                    }
+                }
+            }
+        }
+
+        public override AccessToken GetToken(TokenRequestContext requestContext, CancellationToken cancellationToken)
+        {
+            string token = Callback.Invoke(Authority, Resource, string.Empty).GetAwaiter().GetResult();
+            return new AccessToken(token, DateTimeOffset.Now.AddHours(1));
```
heaths, line 67: See https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/keyvault/Azure.Security.KeyVault.Shared/src/ChallengeBasedAuthenticationPolicy.cs#L112 for how we handle expiration.

src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/AzureSqlKeyCryptographer.cs
```diff
@@ -0,0 +1,233 @@
...
+        internal KeyVaultKey GetKey(string keyIdentifierUri)
+        {
+            if (_keyDictionary.ContainsKey(keyIdentifierUri))
+            {
+                return _keyDictionary[keyIdentifierUri];
+            }
+
+            if (_keyFetchTaskDictionary.ContainsKey(keyIdentifierUri))
+            {
+                return Task.Run(() => _keyFetchTaskDictionary[keyIdentifierUri]).Result;
```
heaths, line 107: `Result` can result in deadlocks. As much as possible, use async code. Where you can't, use `GetAwaiter().GetResult()`.

src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/AzureSqlKeyCryptographer.cs
```diff
@@ -0,0 +1,233 @@
...
+            }
+
+            // Not a public exception - not likely to occur.
+            throw new KeyNotFoundException($"The key with identifier {keyIdentifierUri} was not found.");
```
heaths, line 111: 404 might be rare (certainly possible; an admin could've deleted it), but 403 may be if the application isn't configured to use a service principal with "get" access to the key. What will you do in those cases?

src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/AzureSqlKeyCryptographer.cs
```diff
@@ -0,0 +1,233 @@
...
+
+        private CryptographyClient GetCryptographyClient(string keyIdentifierUri)
+        {
+            if (cryptoClientDictionary.ContainsKey(keyIdentifierUri))
+            {
+                return cryptoClientDictionary[keyIdentifierUri];
+            }
+
+            CryptographyClient cryptographyClient = new CryptographyClient(GetKey(keyIdentifierUri).Id, TokenCredential);
+            cryptoClientDictionary[keyIdentifierUri] = cryptographyClient;
```
heaths, line 162: There's an atomic `AddOrUpdate` method that may yield better performance and less potential overhead (you may allocate more here than you need to with concurrent calls).

src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/AzureKeyVaultProviderTokenCredential.cs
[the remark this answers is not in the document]
cheenamalhotra: Hi @heaths 

We are implementing our own Credential to support old "AuthenticationCallback" in our AKVProvider constructors as otherwise accepting "TokenCredential" only brings breaking changes to customers. We have added new methods to also accept TokenCredential that aligns with Azure.Security.KeyVault APIs.

This class "AzureKeyVaultProviderTokenCredential.cs" is our implementation of TokenCredential which only receives AccessToken from the user through authentication callback, and i**t's expiration time is not known to us**.

What we want to know is what would be expiration time of access tokens generated by calling "[AuthenticationContext.AcquireTokenAsync](https://docs.microsoft.com/en-us/dotnet/api/microsoft.identitymodel.clients.activedirectory.authenticationcontext.acquiretokenasync)" APIs, is that configurable?

And what is the risk of setting expiration time to 1 hour or any other static value in our implementation above?
What would you recommend?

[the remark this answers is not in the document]
heaths: The server sends those back. If the expiresOn attribute is configured by the user, you should probably leave it. In the sample I set, we merely expire/invalidate cached tokens 2 minutes before expiration, to allow for low-bandwidth scenarios.

/cc @schaabs

src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/AzureKeyVaultProviderTokenCredential.cs
cheenamalhotra: In our case token is just the "string" access token, outcome of below, we don't have "expirationTime" from callback.

```
public static async Task<string> AzureActiveDirectoryAuthenticationCallback(string authority, string resource, string scope)
{
    var authContext = new AuthenticationContext(authority);
    ClientCredential clientCred = new ClientCredential(s_clientId, s_clientSecret);
    AuthenticationResult result = await authContext.AcquireTokenAsync(resource, clientCred);
    if (result == null)
    {
        throw new InvalidOperationException($"Failed to retrieve an access token for {resource}");
    }
    return result.AccessToken;
}
```

src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/AzureKeyVaultProviderTokenCredential.cs
cheenamalhotra: This is the callback received from `SqlColumnEncryptionAzureKeyVaultProvider` [constructor](https://github.com/dotnet/SqlClient/pull/630/files#diff-b402b979fad98d38b35b0522ab2fabb6R105) - to support old KeyVaultClient.AuthenticationCallback, that we invoke here to acquire accesstoken (string).

src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/AzureKeyVaultProviderTokenCredential.cs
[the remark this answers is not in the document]
cheenamalhotra: Hi @heaths 

As I posted [here](https://github.com/dotnet/SqlClient/pull/630#discussion_r452541642), we don't have expirationTime from Callback method, it returns us only string "access token" that could be directly used by KeyVaultClient in old design. I did go through your design and looked at more examples but no where I found anything like we are doing here.

ChallengeBased Auth Policy receives "AccessToken" instance [here](https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/keyvault/Azure.Security.KeyVault.Shared/src/ChallengeBasedAuthenticationPolicy.cs#L107-L109), whereas we only get "string" access token [here](https://github.com/dotnet/SqlClient/pull/630/files/6d7eb45ffd7da077370d119128a3961c38a17916#diff-b2c8cd0d3bbeed67854fca5124d813d9R66) that is the response of old KeyVaultClient.AuthenticationCallback.

We have to construct "AccessToken" instance using only "string" access token for which expiry may be different than what we are trying to set.

src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/AzureSqlKeyCryptographer.cs
cheenamalhotra: Is there a strategy you can advice in this case?

src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/AzureKeyVaultProviderTokenCredential.cs
heaths: Per internal discussion, you may be able to write a pipeline policy to get the authority from the challenge response (rather, the request back to authenticate the challenge.) and pass that to the callback. You'd have to associate that with the request - perhaps stored as a field in your token credential, but in a way where if the token credential is used for multiple clients, is associated with the right  resource (vault URI).

src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/AzureSqlKeyCryptographer.cs
```diff
@@ -0,0 +1,233 @@
...
+        internal void AddKey(string keyIdentifierUri)
+        {
+            if (TheKeyHasNotBeenCached(keyIdentifierUri))
+            {
+                if (isUsingLegacyAuthentication)
+                {
+                    TokenCredential = new AzureKeyVaultProviderTokenCredential(AuthenticationCallback, keyIdentifierUri);
+                }
+
+                ParseAKVPath(keyIdentifierUri, out Uri vaultUri, out string keyName);
```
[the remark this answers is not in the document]
cheenamalhotra: I'm not sure I follow, how do we know the version of Key?

cheenamalhotra:
I believe the reason of synchronous methods only is the class `SqlColumnEncryptionKeyStoreProvider`.
I think it would be possible to extend this to also support counter Async APIs, @Xtrimmer any thoughts?

src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/AzureSqlKeyCryptographer.cs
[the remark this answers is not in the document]
Xtrimmer: The Key Identifier URI should contain the version information. The URI is formated like the following:
`https://{keyvault-name}.vault.azure.net/{object-type}/{object-name}/{object-version}`
where `{object-version}` is the version of the key.
I believe that it is possible to leave off the version, but in this case, the latest version of the key would be retrieved from the vault.

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/AKVUnitTests.cs
```diff
@@ -0,0 +1,52 @@
+﻿using Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider;
```
David-Engel, line 1:
```suggestion
// Licensed to the .NET Foundation under one or more agreements.
// The .NET Foundation licenses this file to you under the MIT license.
// See the LICENSE file in the project root for more information.

﻿﻿using Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider;
```

johnnypham (approved):

src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/SqlColumnEncryptionAzureKeyVaultProvider.cs
```diff
@@ -37,7 +32,7 @@ namespace Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider
 	///	    <format type="text/markdown"><![CDATA[
     /// ## Remarks
     /// 
-    /// **SqlColumnEncryptionAzureKeyVaultProvider** is implemented for Microsoft.Data.SqlClient driver and supports .NET Framework 4.6+ and .NET Core 2.1+.
+    /// **SqlColumnEncryptionAzureKeyVaultProvider** is implemented for Microsoft.Data.SqlClient driver and supports .NET Framework 4.6.1+ and .NET Core 2.1+.
```
David-Engel, line 35:
```suggestion
    /// **SqlColumnEncryptionAzureKeyVaultProvider** is implemented for Microsoft.Data.SqlClient and supports .NET Framework 4.6.1+ and .NET Core 2.1+.
```

src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/SqlColumnEncryptionAzureKeyVaultProvider.cs
```diff
@@ -37,7 +32,7 @@ namespace Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider
...
     /// The provider name identifier for this implementation is "AZURE_KEY_VAULT" and it is not registered in driver by default.
     /// Client applications must call <xref=Microsoft.Data.SqlClient.SqlConnection.RegisterColumnEncryptionKeyStoreProviders> API only once in the lifetime of driver to register this custom provider by implementing a custom Authentication Callback mechanism.
```
David-Engel, line 37:
```suggestion
    /// The provider name identifier for this implementation is "AZURE_KEY_VAULT" and it is not registered in driver by default.
    /// Client applications must call the <xref=Microsoft.Data.SqlClient.SqlConnection.RegisterColumnEncryptionKeyStoreProviders> API only once in the lifetime of the driver to register this custom provider by implementing a custom Authentication Callback mechanism.
```

src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/Strings.resx
```diff
@@ -117,23 +117,20 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
-  <data name="InvalidTrustedEndpointsList" xml:space="preserve">
-    <value>trustedEndPoints cannot be null or empty.</value>
+  <data name="NullOrWhitespaceForEach" xml:space="preserve">
+    <value>One or more of the elements in {0} are null or empty or consist of only whitespace.</value>
```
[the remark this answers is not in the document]
David-Engel: @JRahnama "whitespace" is correct here as it's already plural. Whitespace is any character **or series of characters** that represent horizontal or vertical space in typography.

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/AKVUnitTests.cs
```diff
@@ -0,0 +1,107 @@
...
+            byte[] encryptedCek = akvProvider.EncryptColumnEncryptionKey(DataTestUtility.AKVUrl, EncryptionAlgorithm, s_columnEncryptionKey);
+            byte[] decryptedCek = akvProvider.DecryptColumnEncryptionKey(DataTestUtility.AKVUrl, EncryptionAlgorithm, encryptedCek);
+
+            Assert.Equal(s_columnEncryptionKey, decryptedCek);
+        }
+
+        [ConditionalFact(typeof(DataTestUtility), nameof(DataTestUtility.IsAKVSetupAvailable))]
+        public static void TokenCredentialRotationTest()
+        {
+            // SqlClientCustomTokenCredential implements legacy authentication callback to request access token at client-side.
```
David-Engel, line 45:
```suggestion
            // SqlClientCustomTokenCredential implements a legacy authentication callback to request the access token from the client-side.
```

src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/AzureSqlKeyCryptographer.cs
```diff
@@ -0,0 +1,220 @@
...
+        }
+
+        /// <summary>
+        /// Returns the key specified by the Key Identifier URI
+        /// </summary>
+        /// <param name="keyIdentifierUri"></param>
+        /// <returns></returns>
+        internal KeyVaultKey GetKey(string keyIdentifierUri)
+        {
+            if (_keyDictionary.ContainsKey(keyIdentifierUri))
```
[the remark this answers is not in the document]
cheenamalhotra: Thanks @JRahnama 

I'm doing it manually, but just a tip:
You can suggest multi-line changes too. That way we can commit all line changes directly :)

JRahnama (approved):

PR #637, opened 2020-07-08
Fix transient fault handling for Pooled connections

cheenamalhotra:
All versions of SqlClient (including System.Data.SqlClient) that support Transient Fault Handling do not reset the connection before retrying for transient fault handling cases and if Azure connections fail to connect first time **due to server timeout**, the next round of connect fails with below error:

```bash
Type=System.InvalidOperationException.Message=Internal .Net Framework Data Provider error 6.
```

When pooling is disabled, apps are able to connect with Transient Fault Handling.
The PR fixes this issue for pooled connections.

NuGet package for testing: [Published Artifacts](https://dev.azure.com/sqlclientdrivers-ci/sqlclient/_build/results?buildId=13583&view=artifacts&type=publishedArtifacts)


JRahnama:
Logic wise LGM, but how can we test transient errors? Is it possible to have tests to do so?

cheenamalhotra:
Transient errors occur on Database side, so I don't think we can trigger faults on server side.

PR #635, opened 2020-07-06
Fix pooled connection re-use on access token expiry

cheenamalhotra:
Fixes #617 

When federated authentication mode is used, the driver continues to re-use pooled connection even if associated access token is expired. This leads to failures on client environments when Azure DB re-validates credentials after 10 hours (as of now), and physical connection ends up no longer usable.

This PR fixes the behavior and does not let the pooled connection be re-used again if associated access token is due expire soon.

NuGet package for testing is available here:
https://dev.azure.com/sqlclientdrivers-ci/sqlclient/_build/results?buildId=13352&view=artifacts&type=publishedArtifacts

PR #639, opened 2020-07-08
[Release/2.0] Fix pooled connection re-use on access token expiry

cheenamalhotra:
Ports #635 to v2.0

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/ProviderBase/DbConnectionInternal.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/ProviderBase/DbConnectionInternal.cs
@@ -203,7 +203,8 @@ namespace Microsoft.Data.ProviderBase
             }
         }
 
-        virtual internal bool IsAccessTokenExpired => false;
+        internal virtual bool IsAccessTokenExpired => false;
+
         abstract protected void Activate(Transaction transaction);
 
         internal void ActivateConnection(Transaction transaction)
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/ProviderBase/DbConnectionInternal.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/ProviderBase/DbConnectionInternal.cs
@@ -203,14 +203,7 @@ namespace Microsoft.Data.ProviderBase
             }
         }
 
-        virtual internal bool IsAccessTokenExpired
-        {
-            get
-            {
-                return false;
-            }
-        }
-
+        virtual internal bool IsAccessTokenExpired => false;
         abstract protected void Activate(Transaction transaction);
 
         internal void ActivateConnection(Transaction transaction)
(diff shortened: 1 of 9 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/ProviderBase/DbConnectionInternal.cs
```diff
@@ -203,6 +203,14 @@ virtual protected bool ReadyToPrepareTransaction
             }
         }
 
+        virtual internal bool IsAccessTokenExpired
```
JRahnama, line 206: can we fix it to` internal virtual` please?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/ProviderBase/DbConnectionInternal.cs
```diff
@@ -203,6 +203,14 @@ virtual protected bool ReadyToPrepareTransaction
...
+        {
+            get
+            {
+                return false;
```
JRahnama, line 210: another question, why are we returning false always? Isn't it safer to consider this expired by default unless it is not?

cheenamalhotra: I would see it other way.. token expiry is only specific to federated authentication, otherwise there's no need to doom pooled connection, hence this defaults to false.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/ProviderBase/DbConnectionInternal.cs
cheenamalhotra: I also see there's a lot of such inconsistent patterns.. will need a separate PR for all this. But I'll fix this one.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -684,6 +699,10 @@ protected override bool UnbindOnTransactionCompletion
             }
         }
 
+        /// <summary>
+        /// Validates if federated authentication is used, Access Token used by this connection is active for the value of 'accessTokenExpirationBufferTime'.
+        /// </summary>
+        internal override bool IsAccessTokenExpired => _federatedAuthenticationInfoRequested && DateTime.FromFileTimeUtc(_fedAuthToken.expirationFileTime) < DateTime.UtcNow.AddSeconds(accessTokenExpirationBufferTime);
```
JRahnama, line 705: So, here if the connection time out is greater than 600 or infinite time out is requested, it will expire the access token 10 minutes before its actual time. otherwise, it will expire for the connection time out. Am I right? If yes, there is a big gap between like 30 or 60 seconds and 10 minutes. Isn't  it going to make a problem?

JRahnama: besides this question the PR looks good to me.

cheenamalhotra: No the intent is to expire access tokens only when they should expire.

PR #643, opened 2020-07-11
Merge exceptions

Wraith2:
This combines 3 exceptions from both codebases into a single definition. Now without sln changes.
It feels like generally the simplification work I'm looking at has already been done in the netcore codebase, are the netfx and netcore codebases from different times?


David-Engel:
@Wraith2 The netcore code base was forked from .NET Framework in 2015 and had a lot of clean up for .NET Core use. New SQL features were mostly added to .NET Framework for a few years. M.D.S was formed from those diverged code bases in the second half of 2018.

Wraith2:
It seems like the split is that features are in netfx but cleaner code is in netcore. Neither is overall superior as far as I can see but in general I find that code converges on something close to netcore. 

For example in this PR the netcore version of one class was missing a designer serialization attribute even though that attribute is present in netcore and could have been included. Each decision like this takes time to consider and I still might be wrong because there could have been some good reason for the removal.

There are also other questions of intent for this library. Is the netfx version going to need to be able to be hosted in-proc sql server which would require keeping all the CER support code and a couple of Task related workarounds?

David-Engel:
> It seems like the split is that features are in netfx but cleaner code is in netcore. Neither is overall superior as far as I can see but in general I find that code converges on something close to netcore.

That makes sense to me. The netfx code is very old. The fork to netcore and cleanup was an opportunity to bring the code up-to-date in easy places.

> For example in this PR the netcore version of one class was missing a designer serialization attribute even though that attribute is present in netcore and could have been included. Each decision like this takes time to consider and I still might be wrong because there could have been some good reason for the removal.

A large part of the original netcore fork/cleanup was about removing code which could not be supported under netcore. Maybe those attributes were not available in netcore at the time of the fork. (I haven't investigated.)

> There are also other questions of intent for this library. Is the netfx version going to need to be able to be hosted in-proc sql server which would require keeping all the CER support code and a couple of Task related workarounds?

I've thought about the same thing. For now, I wouldn't let that hold us back. We aren't currently hosted in-proc and it isn't a current requirement. @saurabh500 Do you have any insight into this area?

Wraith2:
Ok. My intention is to unify everything I can without losing anything so if an attribute is supported in all the build targets but not in the netcore codebase I'd probably decide to include it. I'm doing small chunks at to keep things review friendly but eventually it's going to reach a stage where some serious time needs to be put in from the MS side to finish up.

saurabh500:
SQL server supports hosting CLR assemblies as UDTs and we should continue to be able to keep the capabilities that are needed for in-proc hosting. E.g. a user can deploy .Net CLR typed based on .Net framework which uses MDS and the expectation would be for sqlclient to work.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/DataException.cs
```diff
@@ -1,1533 +0,0 @@
...
-        }
-        static internal void TraceExceptionWithoutRethrow(Exception e)
-        {
-            TraceException("<comm.ADP.TraceException|ERR|CATCH> Message = '{0}'", e);
-        }
-
-        //
-        // COM+ exceptions
-        //
-        static internal ArgumentException _Argument(string error)
```
cheenamalhotra, line 59: Ack: Dead code in rest of the file.

cheenamalhotra:
@Wraith2 

Please resolve conflicts so we can proceed with this PR, thanks!

Wraith2:
I'll close this in favour of https://github.com/dotnet/SqlClient/pull/676 which changes at least one of these types and is likely to get merged first, i'll be easier to deal with that large PR without merge conflicts and I can easily revisit this afterwards. /cc @johnnypham

cheenamalhotra:
Fair enough, appreciate your thought!

PR #667, opened 2020-07-24
Small spanifications

Wraith2:
Two pretty small changes.

The use of a shared temp array in some places was bothering me so I've replaced it with a const sized stack allocated span that the jit will simply allocate as part of the stack frame and then work entirely on locals. This should allow better codegen and referential locality.

The presence of multiple ways of writing an int to an array was wasteful. The most efficient way to do this is to use the netcore BitConverter.TryWriteByte method which uses MemoryMarshal to do safe il level tricks to write all 4 bytes in one. If that isn't present (netstandard, netfx) then do it using const offsets so the codegen can reorder or pipeline them efficiently.

Measuring a real-world impact of either of these in the context of the entire library is impossible. These are opinionated changes based on good performance practice picked up from work in corefx and I believe they improve the code quality overall.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
@@ -1397,36 +1397,28 @@ namespace Microsoft.Data.SqlClient
         {
             Debug.Assert(_syncOverAsync || !_asyncReadWithoutSnapshot, "This method is not safe to call when doing sync over async");
 
-            byte[] buffer;
-            int offset;
+            Span<byte> buffer = stackalloc byte[2];
             if (((_inBytesUsed + 2) > _inBytesRead) || (_inBytesPacket < 2))
             {
                 // If the char isn't fully in the buffer, or if it isn't fully in the packet,
                 // then use ReadByteArray since the logic is there to take care of that.
-
-                if (!TryReadByteArray(_bTmp, 2))
+                if (!TryReadByteArray(buffer, 2))
                 {
                     value = '\0';
                     return false;
                 }
-
-                buffer = _bTmp;
-                offset = 0;
             }
             else
             {
                 // The entire char is in the packet and in the buffer, so just return it
                 // and take care of the counters.
-
-                buffer = _inBuff;
-                offset = _inBytesUsed;
-
+                buffer = _inBuff.AsSpan(_inBytesUsed, 2);
                 _inBytesUsed += 2;
                 _inBytesPacket -= 2;
             }
 
             AssertValidState();
-            value = (char)((buffer[offset + 1] << 8) + buffer[offset]);
+            value = (char)((buffer[1] << 8) + buffer[0]);
             return true;
         }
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
```diff
@@ -1397,107 +1397,87 @@ internal bool TryReadChar(out char value)
...
             AssertValidState();
-            value = (short)((buffer[offset + 1] << 8) + buffer[offset]);
+            value = (short)((buffer[1] << 8) + buffer[0]);
             return true;
         }
 
         internal bool TryReadInt32(out int value)
         {
             Debug.Assert(_syncOverAsync || !_asyncReadWithoutSnapshot, "This method is not safe to call when doing sync over async");
+            Span<byte> buffer = stackalloc byte[4];
```
cheenamalhotra, line 1457: Wondering if this stack allocation is an overhead in attempt to merge code underneath?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
```diff
@@ -1397,107 +1397,87 @@ internal bool TryReadChar(out char value)
...
             if (((_inBytesUsed + 4) > _inBytesRead) || (_inBytesPacket < 4))
             {
                 // If the int isn't fully in the buffer, or if it isn't fully in the packet,
                 // then use ReadByteArray since the logic is there to take care of that.
-
-                if (!TryReadByteArray(_bTmp, 4))
```
cheenamalhotra, line 1478: The old design reused _bTmp to reduce allocations, coming from Perf perspective, wouldn't that be more desired?

Wraith2: It depends on the context. If we're attempting to use a heap allocated array each time then reusing a single one is a good idea. So the original design was better than allocating each time the method was called and probably had less overhead than trying to rent such a trivially small buffer. Without being able to use spans this is good.

The span as added is allocated as part of the stack frame (and is constant sized so the jit can pre-allocate and zero as part of frame setup) so there's no heap alloc, which means no GC. it doesn't require covariant checks on assignment and it's always exactly the size defined so offset calculations change from calculations to constants which means they can be parallelized. 

have a look at the IL and ASM produced in this example:
https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBLANgHwAEAmARgFgAoEgBipIGYACAbyqY6cOZKYFl2nNpU6imAByjYAbgEMMMJsACeCgNoBdJgH1gAFQC24gNyCxTM2O5cUTAOpSFASQB2GBsQAUK9VoyyoAHMYDDQmbDcmCAAzaIBnELCIjCZpAEpLUWFzc38gkLUY+JCtAF4mb1UYNM9pJgAyJhoEWLTjCxEc0TzgjELYhJSAaiZSMoqfas9apgA+WaYADjSGppboto6u7oDe/uLhpmJxyoUamfnRpBXG5tb2zJyegqLBphGGE8nzusviFBua3ujw4AF8QVsctZCLYHNhnG4PJ4AMriWQuAA8kwWz1C4Ui6Qh2W2HFxak+THKpymFwW/0Bdw2pk6JLJNC+VRqdVu602EJ2+T6Y0pE0501+C2Wq0ZfJZ2zJxxF1J+cwWpGu0t5DzlonBOsh5mhtgAqgk9DAjABBKBQWTKTzJVIZfUcYkk7DRCq6QziUqlFwAV1wuGdJNd/Jy3qMIpcMAA7koqmoUBpmWGmHr0xGOJNNImFHERVGTNn7I4YK53F5iwA6S1xVHozw0NAAtDpNM5TM5CFGpimmDI/xgADWluDEDADoJoZybu2jcx2PzMEL5Tiw5HsgnYBXydTpbhCKr1Li7baEO75ksetBQA=

the span versions are shorter and contain fewer branching instructions in general so they execute faster (less instructions in general) and have fewer dependencies between instructions so multiple parts can be execute in parallel. 

I think the code is easier to read as well. :grin:
We already have a dependency on span for the netfx builds so this doesn't add anything new there.

cheenamalhotra: I agree, was just wondering Perf (I'm sure you have considered 🙌)
Thanks for the details!

Wraith2: Perf will be very slightly better. Or perhaps it's better to say that there's opportunity for the JIT to produce better code with this version.  In the day-to-day running of the library it will make an unmeasurable positive change but I feel the code is clearer and better separated out this way because you can see everything the functions are using in the parameter list. As I mentioned it's an opinionated change so it if doesn't feel easier to read to everyone working on the library it's fine to close.

PR #668, opened 2020-07-27
Only check inexact match when no exact match is found.

yyjdelete:
Fix dotnet/runtime#39620

The same as dotnet/runtime#39961

Avoid the exact match result be overwrite by later inexact ones.
This may not happen due to limited callsites, just make the code more clear.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Common/src/Microsoft/Data/ProviderBase/DbMetaDataFactory.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Common/src/Microsoft/Data/ProviderBase/DbMetaDataFactory.cs
@@ -283,7 +283,7 @@ namespace Microsoft.Data.ProviderBase
                             exactCollectionName = candidateCollectionName;
                             haveExactMatch = true;
                         }
-                        else
+                        else if (haveExactMatch == false)
                         {
                             // have an inexact match - ok only if it is the only one
                             if (exactCollectionName != null)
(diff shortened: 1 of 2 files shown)


cheenamalhotra:
Is there a test case to verify the fix?

yyjdelete:
Infact, I don't know how to create an test for this internal api. And maybe it's not a problem due to it's internal and there is no dupe tables with different case actually used by public apis.

https://github.com/dotnet/sqlclient/search?q=filename%3AMetaData.xml&unscoped_q=filename%3AMetaData.xml

roji:
@yyjdelete it may be worth looking into this a little further and finding an actual user impact - since the current implementation does seem buggy.

Wraith2:
If it isn't testable but is clearly a fix I'm fine with merging it but testing does always need to be considered een though writing tests sucks.

David-Engel:
This bug would only surface if the embedded resource, Microsoft.Data.SqlClient.SqlMetaData.xml, contained multiple MetaDataCollections with a CollectionName differing only by case and the desired entry (case sensitive match) appeared before the undesired entry (case insensitive match). Testing the bug would be awkward. I'm fine with simply fixing the bug. Bonus points if a couple lines are added to TestGetSchema to test `conn.GetSchema("Databases")` in addition to `conn.GetSchema("DATABASES")` to cover both blocks of the if statement in question.

yyjdelete:
> This bug would only surface if the embedded resource, Microsoft.Data.SqlClient.SqlMetaData.xml, contained multiple MetaDataCollections with a CollectionName differing only by case and the desired entry (case sensitive match) appeared before the undesired entry (case insensitive match). Testing the bug would be awkward. I'm fine with simply fixing the bug. Bonus points if a couple lines are added to TestGetSchema to test `conn.GetSchema("Databases")` in addition to `conn.GetSchema("DATABASES")` to cover both blocks of the if statement in question.

@David-Engel
There is already another test with `conn.GetSchema("Databases")`(case sensitive match) in `ConnectionSchemaTest`.
https://github.com/dotnet/SqlClient/blob/4258e4b44220c5f96997717a9b3441968108abd4/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataBaseSchemaTest/ConnectionSchemaTest.cs#L32-L36

PR #671, opened 2020-07-29
Rename netcore string resource file and helpers

Wraith2:
Brings the resources and helper file into sync between netcore and netfx by using the netfx naming scheme. This may allow localization in netcore in future.

/cc @cheenamalhotra @David-Engel


Wraith2:
Is there any chance of getting this merged to unblock further merge pr's?

Issue #680, opened 2020-08-09
CacheConnectionStringProperties is called twice when initializing SqlConnection.


TrayanZapryanov:
When calling any constructor of **SqlConnection** - there is an explicit call of **CacheConnectionStringProperties**.
This looks redundant as both ctor's are setting property **ConnectionString** and there is a call in it's set method.
I hope that I've read correctly the code, if not please excuse me and close the issue.

JRahnama:
@TrayanZapryanov by saying any constructor you mean I can see the issue on ` new SqlConnection(sourceString.ConnectionString)` for example?
If the answer is no, can you kindly provide the repro/cpde sample that you were looking at while facing the issue?

And second is it happening at netcore or dotnet framework?

TrayanZapryanov:
I've mean both public constructors in netcore:
[https://github.com/dotnet/SqlClient/blob/dbbcde899c646df362ac8369f2d1edd82437759f/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnection.cs#L117](url)

and

[https://github.com/dotnet/SqlClient/blob/dbbcde899c646df362ac8369f2d1edd82437759f/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnection.cs#L160](url)

Although same is valid and for dotnet framework.

karinazhou:
Hi @TrayanZapryanov ,

Thank you for pointing this out. I agree that there are additional calls of CacheConnectionStringProperties in some SqlConnection constructors after setting ConnectionString property.

PR #683, opened 2020-08-10
Remove duplicate CacheConnectionStringProperties calls

karinazhou:
According to #680, we have some unnecessary CacheConnectionStringProperties calls after setting ConnectionString properties.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnection.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnection.cs
@@ -114,7 +114,6 @@ namespace Microsoft.Data.SqlClient
         public SqlConnection(string connectionString) : this()
         {
             ConnectionString = connectionString;    // setting connection string first so that ConnectionOption is available
-            CacheConnectionStringProperties();
         }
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnection.cs
```diff
@@ -157,7 +156,6 @@ public SqlConnection(string connectionString, SqlCredential credential) : this()
             // else
```
DavoudEshtehari, line 156: Also, the code from line 128 to 152 do not require, cause of they already have checked inside the `ConnectionString` property.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnection.cs
```diff
@@ -336,7 +336,6 @@ public SqlConnection(string connectionString, SqlCredential credential) : this()
             // else
```
DavoudEshtehari, line 336: Like `.net core`'s comment.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnection.cs
karinazhou: The validation from line 128 to 152 is to check the connection options from the pool group while ConnectionString property setter checks the connectionString in use. Same logic in netfx.

PR #686, opened 2020-08-12
Port Fix for EnclaveSessionCache Issue

karinazhou:
This PR ports the changes from https://github.com/dotnet/SqlClient/pull/628 .

Update:
Wrap the information such as server name, attestationURL, and database required by Enclave Session into a parameter class.

### DIFF
--- a/doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionEnclaveProvider.xml
+++ b/doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionEnclaveProvider.xml
@@ -19,7 +19,7 @@ the enclave attestation protocol as well as the logic for creating and caching e
         <CreateEnclaveSession>
             <param name="enclaveAttestationInfo">The information the provider uses to attest the enclave and generate a symmetric key for the session. The format of this information is specific to the enclave attestation protocol.</param>
             <param name="clientDiffieHellmanKey">A Diffie-Hellman algorithm object that encapsulates a client-side key pair.</param>
-            <param name="enclaveSessionParameters">The set of parameters required for enclave session.</param>
+            <param name="enclaveSessionParameters">The set of parameters required for an enclave session.</param>
             <param name="customData">The set of extra data needed for attestating the enclave.</param>
             <param name="customDataLength">The length of the extra data needed for attestating the enclave.</param>
             <param name="sqlEnclaveSession">The requested enclave session or <see langword="null" /> if the provider doesn't implement session caching.</param>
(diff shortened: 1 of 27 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlEnclaveSession.cs
```diff
@@ -56,4 +56,18 @@ internal SqlEnclaveSession(byte[] sessionKey, long sessionId/*, long counter*/)
             SessionId = sessionId;
         }
     }
+
+    internal class EnclaveSessionParameters
```
TrayanZapryanov, line 60: Isn't is good this class to be immutable?

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlEnclaveSession.cs
```diff
@@ -53,4 +53,18 @@ internal SqlEnclaveSession(byte[] sessionKey, long sessionId/*, long counter*/)
             SessionId = sessionId;
         }
     }
+
+    internal class EnclaveSessionParameters
```
JRahnama, line 57: I think we could use a struct instead of a class here.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlEnclaveSession.cs
JRahnama: We could use struct instead of a class here.

karinazhou: This class wraps several strings that are required by the enclave session. From the changes in this PR, we only assign the values when we create the EnclaveSessionParameters object.

doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionEnclaveProvider.xml
```diff
@@ -19,8 +19,7 @@ the enclave attestation protocol as well as the logic for creating and caching e
         <CreateEnclaveSession>
             <param name="enclaveAttestationInfo">The information the provider uses to attest the enclave and generate a symmetric key for the session. The format of this information is specific to the enclave attestation protocol.</param>
             <param name="clientDiffieHellmanKey">A Diffie-Hellman algorithm object that encapsulates a client-side key pair.</param>
-            <param name="attestationUrl">The endpoint of an attestation service for attesting the enclave.</param>
-            <param name="servername">The name of the SQL Server instance containing the enclave.</param>
+            <param name="enclaveSessionParameters">The set of parameters required for enclave session.</param>
```
johnnypham, line 22:
```suggestion
            <param name="enclaveSessionParameters">The set of parameters required for an enclave session.</param>
```

Issue #688, opened 2020-08-13
AsyncHelper.WaitForCompletion leaks unobserved exceptions


jm771:
### To reproduce
I have a reproduction of the issue along with a possible fix here:
https://github.com/dotnet/SqlClient/compare/master...jm771:ReproUnobservedException

### Bug description
WaitForCompletion waits for a task to complete with a timeout, and allows you to specify a handler for the timeout case. However, once the timeout has occurred it ditches the task, which means if that task later completes with whatever error caused it to timeout, the exception will be unobserved, and the UnobservedTaskException event will fire.

As AsyncHelper is internal, the bug wasn't experienced directly, but certain timeout conditions in Sql connections would lead to unobserved exceptions. My former employer had a policy of no unobserved exceptions, and so had a handler that killed the application when they occurred. We worked around this at the time by changing the handler to ignore this type of exception, and it's only now I'm finding time to file a proper bug report.

karinazhou:
Hi @jm771 ,

This is a good catch. I will test your branch locally to validate your change. Meanwhile, please feel free to create the PR so we can review it properly.

jm771:
I have raised a PR here:
https://github.com/dotnet/SqlClient/pull/692

karinazhou:
Hi @jm771 ,

We are closing this issue because your PR has been merged. 


Thanks,

PR #692, opened 2020-08-14
Issue 688 AsyncHelper.WaitForCompletion leaks unobserved exceptions

jm771:
Fix for https://github.com/dotnet/SqlClient/issues/688

I have some comments which I'll leave inline on the code review


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/SqlClient?pullRequest=692) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/SqlClient?pullRequest=692) before we can accept your contribution.<br/><br/>:x: jm771 [sign now](https://cla.dotnetfoundation.org/dotnet/SqlClient?pullRequest=692)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/SqlClient?pullRequest=692) it.</sub>

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlHelperTest.cs
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        [Fact]
+        public void WaitForCompletion_DoesNotCreateUnobservedException()
+        {
+            var unobservedExceptionHappenedEvent = new AutoResetEvent(false);
+            EventHandler<UnobservedTaskExceptionEventArgs> handleUnobservedException = 
+                (o, a) => { unobservedExceptionHappenedEvent.Set(); };
+
+            TaskScheduler.UnobservedTaskException += handleUnobservedException;
```
jm771, line 28: I don't much like messing with global state from a unit test, but I'm not sure of what other way to test this. I had a look at TaskExceptionHolder (which is what is responsible for raising the event when finalized) but it looks anything along those lines would be messing with the internals of the Task system too much.

jm771:
So here's something exciting

```2020-08-14T15:43:06.0285145Z   X Microsoft.Data.SqlClient.Tests.SqlHelperTest.WaitForCompletion_DoesNotCreateUnobservedException [1s 35ms]
2020-08-14T15:43:06.0285944Z   Error Message:
2020-08-14T15:43:06.0286678Z    Did not expect an unobserved exception, but found a System.InvalidOperationException with message "Fail requested"```

This is not the unobserved exception that I'm creating, so some other unit test is leaking unobserved exceptions which my unit test is picking up on - would welcome advice on how to proceed.

File: src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlUtil.cs [hunk 2]
```diff
@@ -204,6 +206,7 @@ internal static void WaitForCompletion(Task task, int timeout, Action onTimeout
             }
             if (!task.IsCompleted)
             {
+                task.ContinueWith(_ => { }); //Ensure the task does not leave an unobserved exception
```
Wraith2, line 209: Can you verify that this doesn't allocate a delegate and closure/state object on each invocation? I've done a lot of work to try and reduce allocations for async calls and I'd like to make sure things aren't regressed without good reason.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlUtil.cs
```diff
@@ -17,6 +17,8 @@
 using System.Transactions;
 using Microsoft.Data.Common;
 
+[assembly: InternalsVisibleTo("FunctionalTests")]
```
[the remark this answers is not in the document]
Wraith2: Avoiding it altogether would be best. Can the file be included in the test project and used that way?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlUtil.cs
jm771: Shouldn't do - as there is nothing to capture in the closure - did this as an experiment to confirm:

`	private static Action<Task> GetLambdaNoClosure(int context)
	{
		return _ => { };
	}
	
	private static Action<Task> GetLambdaWithClosure(int context)
	{
		return _ => { context++; };
	}
	
	public static void Main()
	{
		var a = GetLambdaNoClosure(1);
		var b = GetLambdaNoClosure(2);
		Console.WriteLine(object.ReferenceEquals(a, b));
		
		var c = GetLambdaWithClosure(1);
		var d = GetLambdaWithClosure(2);
		Console.WriteLine(object.ReferenceEquals(c, d));
	}`

Output:
`True
False`

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlUtil.cs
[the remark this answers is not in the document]
jm771: Agree not polluting the DLL with attributes which aren't needed by end users would be better.

The problem with the file inclusion approach is a copy of SqlUtil.cs appears in each of netfx Microsoft.Sql.DataClient and netcore Microsoft.Sql.DataClient - but there is only a single Microsoft.Sql.DataClient.Tests which tests both builds - so it's impossible to include both files. (I also think there's some philosophical considerations about whether you want to test the source file, or the built binary)

I could do some trickery with reflection or IgnoresAccessChecksToAttribute to get at the internals without them being exposed, but that sounds pretty nasty

src/Microsoft.Data.SqlClient/tests/FunctionalTests/BaseProviderAsyncTest/BaseProviderAsyncTest.cs
```diff
@@ -90,14 +95,14 @@ public static void TestDbCommand()
             {
                 Fail = true
             };
-            commandFail.ExecuteNonQueryAsync().ContinueWith((t) => { }, TaskContinuationOptions.OnlyOnFaulted).Wait();
```
jm771, line 93: Setting TaskContinuationOptions.OnlyOnFaulted and not checking the exception leads to the exception being unobserved:

https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.taskcontinuationoptions?view=netcore-3.1

"If you do not access the Exception property, the exception is unhandled."

I also was not a fan of the fact that if the command did not fail, the test would hand indefinitely, so I swapped to the current approach

jm771: Think I have found the source of these exceptions, and should be fixed in latest commit

jm771:
Trying to go through the test failures now - there's a bunch of manual test failures which it's hard to see how I'd have caused, some failures relating to encryption which I have no idea about, and then the one failure that is definitely me is the test I've added failing on build configuration Ubuntu 16 AnyCPU,Release,netcoreapp,true - the same test does not fail on build configuration Ubuntu 16 AnyCPU,Release,netcoreapp,false - but I don't have any idea as to what that bool parameter is referring to - I don't have permission to view the pipeline, and couldn't glean any difference from viewing the log files - can someone help me out?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlUtil.cs
Wraith2: Ok. Correctness has priority before performance anyway so probably not is good enough for me.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlUtil.cs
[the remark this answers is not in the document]
Wraith2: If the required wirjng to use reflection is small enough (in your opinion) to be managable then it's a valid approach that is used in some tests, SslOverTdsStream iirc if you want to take a look. If it's too complicated to use reflection then is there a way to provide a reflectable interface class in the main binary?

I have been doing work to merge the implementations of netfx and netcore but I'm not sure how complex that work might be for sqlutil.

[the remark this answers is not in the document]
jm771: Here is what the reflection solution might look like: https://github.com/dotnet/SqlClient/compare/master...jm771:ReflectiveSoln

If I were doing this at my work, I'd be choosing between using the attribute, or omitting the test entirely (partly due to its fiddling with static state), I don't like the reflective way of doing it as it's very easy to break the test with simple refactoring, and doesn't play nicely with standard "find usages" IDE tools. However the requirements at my work are different to the requirements here (in particular we aren't publishing a library for other people to consume). You are much more familiar with the requirements and style here, and I'm happy to go with whichever option you prefer.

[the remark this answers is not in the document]
Wraith2: Fair enough. I think the reflective solution is too cumbersome in this case but I really don't want to add internals attributes if possible. We'll see what the team think.

Wraith2:
Most the AE test runs fail for configuration reasons. I ignore them, i really hope the SQL team aren't relying on them.

The only one i can find that's relevant to your PR is this: https://dev.azure.com/sqlclientdrivers-ci/sqlclient/_build/results?buildId=15188&view=logs&j=9d906afe-3eb5-5d79-316d-86f3f3fa360c&t=52375b9b-c5db-514a-5e98-413583db2186

```
2020-08-15T11:22:49.7596685Z [xUnit.net 00:00:17.23]     Microsoft.Data.SqlClient.Tests.SqlHelperTest.WaitForCompletion_DoesNotCreateUnobservedException [FAIL]
2020-08-15T11:22:49.7602056Z [xUnit.net 00:00:17.23]       Did not expect an unobserved exception, but found a System.TimeoutException with message "Dummy timeout exception"
2020-08-15T11:22:49.7611531Z [xUnit.net 00:00:17.23]       Expected: False
2020-08-15T11:22:49.7612839Z [xUnit.net 00:00:17.23]       Actual:   True
2020-08-15T11:22:49.7613874Z [xUnit.net 00:00:17.24]       Stack Trace:
2020-08-15T11:22:49.7614567Z [xUnit.net 00:00:17.24]         SqlHelperTest.cs(46,0): at Microsoft.Data.SqlClient.Tests.SqlHelperTest.WaitForCompletion_DoesNotCreateUnobservedException()
2020-08-15T11:22:49.7615169Z   âˆš Microsoft.Data.SqlClient.Tests.SqlConnectionBasicTests.RetrieveWorkstationId(workstation: null, withDispose: False, shouldMatchSetWorkstationId: False) [1ms]
2020-08-15T11:22:49.7615733Z   âˆš Microsoft.Data.SqlClient.Tests.SqlConnectionBasicTests.RetrieveWorkstationId(workstation: "RandomStringForTargetServer", withDispose: False, shouldMatchSetWorkstationId: True) [1ms]
2020-08-15T11:22:49.7616815Z   âˆš Microsoft.Data.SqlClient.Tests.SqlConnectionBasicTests.RetrieveWorkstationId(workstation: "RandomStringForTargetServer", withDispose: True, shouldMatchSetWorkstationId: False) [1ms]
2020-08-15T11:22:49.7617389Z   âˆš Microsoft.Data.SqlClient.Tests.SqlConnectionBasicTests.RetrieveWorkstationId(workstation: "", withDispose: False, shouldMatchSetWorkstationId: False) [1ms]
2020-08-15T11:22:49.7617837Z   âˆš Microsoft.Data.SqlClient.Tests.SqlConnectionBasicTests.ConnectionTest [652ms]
2020-08-15T11:22:49.7618230Z   âˆš Microsoft.Data.SqlClient.Tests.SqlConnectionBasicTests.ConnectionTestValidCredentialCombination [1ms]
2020-08-15T11:22:49.7618637Z   âˆš Microsoft.Data.SqlClient.Tests.SqlConnectionBasicTests.SqlConnectionEmptyParameters [1ms]
2020-08-15T11:22:49.7619256Z   âˆš Microsoft.Data.SqlClient.Tests.SqlConnectionBasicTests.ClosedConnectionSchemaRetrieval [1ms]
2020-08-15T11:22:49.7620040Z   âˆš Microsoft.Data.SqlClient.Tests.SqlConnectionBasicTests.IntegratedAuthConnectionTest [161ms]
2020-08-15T11:22:49.7620562Z   X Microsoft.Data.SqlClient.Tests.SqlHelperTest.WaitForCompletion_DoesNotCreateUnobservedException [1s 30ms]
2020-08-15T11:22:49.7621135Z   Error Message:
2020-08-15T11:22:49.7621966Z    Did not expect an unobserved exception, but found a System.TimeoutException with message "Dummy timeout exception"
2020-08-15T11:22:49.7622793Z Expected: False
2020-08-15T11:22:49.7623560Z Actual:   True
2020-08-15T11:22:49.7624319Z   Stack Trace:
2020-08-15T11:22:49.7625178Z      at Microsoft.Data.SqlClient.Tests.SqlHelperTest.WaitForCompletion_DoesNotCreateUnobservedException() in E:\ci-agent1\_work\1\s\src\Microsoft.Data.SqlClient\tests\FunctionalTests\SqlHelperTest.cs:line 46
```

src/Microsoft.Data.SqlClient/tests/FunctionalTests/BaseProviderAsyncTest/BaseProviderAsyncTest.cs
```diff
@@ -116,17 +121,17 @@ public static void TestDbCommand()
             source = new CancellationTokenSource();
             Task.Factory.StartNew(() => { command.WaitForWaitingForCancel(); source.Cancel(); });
             Task result = command.ExecuteNonQueryAsync(source.Token);
-            Assert.True(result.IsFaulted, "Task result should be faulted");
+            Assert.True(result.Exception != null, "Task result should be faulted");
```
jm771, line 124: Checking IsFaulted does not count as observing an exception, accessing the Exception property does

Wraith2:
The test suite got updated recently to make all the AE runs function. Can you merge or rebase to master and see if you get clean test runs?

src/Microsoft.Data.SqlClient/tests/FunctionalTests/BaseProviderAsyncTest/BaseProviderAsyncTest.cs
```diff
@@ -13,6 +13,11 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests
 {
     public static class BaseProviderAsyncTest
     {
+        private static void AssertTaskFaults(Task t)
+        {
+            Assert.ThrowsAny<Exception>(() => t.Wait(TimeSpan.FromMilliseconds(1)));
```
cheenamalhotra, line 18: Nit: Could you confirm if if providing timeout here is intentional?
Although it's a test, but changes existing behavior.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlHelperTest.cs
```diff
@@ -0,0 +1,55 @@
...
+            return e?.InnerException != null ? UnwrapException(e.InnerException) : e;
+        }
+
+        [Fact]
+        public void WaitForCompletion_DoesNotCreateUnobservedException()
+        {
+            var unobservedExceptionHappenedEvent = new AutoResetEvent(false);
+            Exception unhandledException = null;
+            EventHandler<UnobservedTaskExceptionEventArgs> handleUnobservedException = 
+                (o, a) => { unhandledException = a.Exception; unobservedExceptionHappenedEvent.Set(); };
```
cheenamalhotra, line 32:
```suggestion
            void handleUnobservedException(object o, UnobservedTaskExceptionEventArgs a)
            { unhandledException = a.Exception; unobservedExceptionHappenedEvent.Set(); }
```

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlHelperTest.cs
```diff
@@ -0,0 +1,55 @@
...
+            tcs.SetException(new TimeoutException("Dummy timeout exception")); //Our task now completes with an error
+        }
+
+        private Exception UnwrapException(Exception e)
+        {
+            return e?.InnerException != null ? UnwrapException(e.InnerException) : e;
+        }
+
+        [Fact]
+        public void WaitForCompletion_DoesNotCreateUnobservedException()
```
cheenamalhotra, line 27: This test does need a bit of rework to handle randomness. It fails randomly in pipelines.

You'll notice if you run this test in loop, the second round does not pass ever. Any thoughts why can't we run this test in second round? It maybe a hint towards random errors.

jm771: Was just looking into this - it seems the fix I had written to WaitForCompletion wasn't actually a fix. I needed to make the continuation actually observe the exception, the behaviour I had lead to a race condition, where if the continuation hadn't been executed at the point we called GC.Collect, then there was still a reference to the original task and it wasn't Collected. Creating a false positive for me having fixed it.

I proved this by modifying WaitForCompletion to return the continuation, and waited for it to complete, at which point the error happened deterministically. Making the continuation observe the exception fixed the error. My bad, I misinterpreted the text here: https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.taskcontinuationoptions?view=netcore-3.1 saying "If you do not access the Exception property, the exception is unhandled." To only apply to OnlyOnFaulted, when it applies to all TaskContinuationOptions.

I can't see a way to make the failure deterministic without passing some sort of signal back from the continuation, which doesn't fit the signature and isn't desirable.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlUtil.cs
```diff
@@ -204,6 +206,7 @@ internal static void WaitForCompletion(Task task, int timeout, Action onTimeout
             }
             if (!task.IsCompleted)
             {
+                task.ContinueWith(t => { var ignored = t.Exception; }); //Ensure the task does not leave an unobserved exception
```
jm771, line 209: N.B. I repeated the experiment for the new lambda to confirm it's still not creating lots of instances of the closure

src/Microsoft.Data.SqlClient/tests/FunctionalTests/BaseProviderAsyncTest/BaseProviderAsyncTest.cs
jm771: Yes it's intentional - note that t.Wait returns false in the case of timeout (doesn't throw a TimeoutException) so if the task under test times out the test will fail, as no exception will be thrown. I thought this was preferable to the previous behaviour where if the Task hung the test would hang indefinitely (or until killed by the framework)

In this particular case all the Tasks under test are already completed ones generated by Task.FromResult, so we can get away with a 1ms timeout and know nothing will timeout.

However this is somewhat a matter of taste, and I don't know the details of whether your test suite is set up to handle timing out tests well. So if you'd rather have a longer timeout or no timeout I'm happy to change it.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlHelperTest.cs
Wraith2: This part is just a test so make it as messy as it needs to be to get a deterministic answer.

jm771: It does pass deterministically now with the fix in place - what I couldn't get deterministic was the failure without the fix in place.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlUtil.cs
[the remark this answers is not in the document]
karinazhou: Though it's not encouraged to include this assembly attribute in the main code, it is acceptable to expose SqlUtil to the tests only if we cannot find another better way to do it.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/BaseProviderAsyncTest/BaseProviderAsyncTest.cs
karinazhou: The original design for this is to verify that the unhandled exception can be propagated through tasks. We don't want the tests to fail due to this 1 ms timeout although it should not happen according to your explanation. I personally would prefer not setting timeout here.

karinazhou:
Hi @jm771 ,

Could you please merge master into your branch again so we can trigger the failed AzureSQLDB pipelines? The failure indicates some missing changes from the new test framework. Just to make sure your fixes working well on these pipelines before we can merge it. Sorry for the inconvenience. 

Thanks,

jm771:
On holiday at the moment, I'll take a look when I get back next week.

karinazhou:
Hi @jm771  , do you have any chance to update your branch so we can merge your changes soon?

Thanks,

jm771:
> 
> 
> Hi @jm771 , do you have any chance to update your branch so we can merge your changes soon?
> 
> Thanks,

Sorry about the delay, I've merged just now. I'll have a look over all the results of the tests tomorrow.

karinazhou:
@jm771   Thank you for the updates. I have checked the failed tests on the pipelines and they are known issues that are not introduced by your PR.

karinazhou (approved):

jm771:
> @jm771 Thank you for the updates. I have checked the failed tests on the pipelines and they are known issues that are not introduced by your PR.

Excellent, thanks Karina!

cheenamalhotra (approved):

PR #419, opened 2020-02-10
Return empty DataTable from GetSchemaTable.

ErikEJ:
Remove unused code from DataReaderTest

fixes #417

PR #697, opened 2020-08-19
[Release/2.0] Revert "Return empty DataTable from GetSchemaTable. (#419)"

cheenamalhotra:
Reverts #419 - As confirmed by @roji offline.

cc @ajcvickers @David-Engel


roji:
Thank you @cheenamalhotra, and sorry once again for the mess this created.

PR #699, opened 2020-08-19
[Cleanup] Remove unused algorithm classes.

cheenamalhotra:
cc @XTrimmer


cheenamalhotra:
@Wraith2 this was legacy code, exists only in NetFx, copied over as-is, but it's unused code.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlSecurityUtility.cs
```diff
@@ -150,10 +150,6 @@ private static string ValidateAndGetEncryptionAlgorithmName(byte cipherAlgorithm
             {
                 return SqlAeadAes256CbcHmac256Algorithm.AlgorithmName;
             }
-            else if (TdsEnums.AES_256_CBC == cipherAlgorithmId)
```
Xtrimmer, line 153: Is safe to purge this constant from netfx and netcoreTdsEnums.cs too?

Issue #523, opened 2020-04-14
Add support for producing Event Source based counters in .NET Core


cheenamalhotra:
Based on discussions in #382 , it would be a good idea to introduce support for `EventSource` based `EventCounters` in **Microsoft.Data.SqlClient** for .NET Core.

`EventCounters` are supported in all OS platforms - Windows, Linux, and macOS. It can be thought of as a cross-platform equivalent for the `PerformanceCounters` that is only supported in Windows systems.

Articles to get started:
- [Event Counters - Tutorial](https://github.com/dotnet/runtime/blob/e9c8f6a778d7aade9f9a13b9c13a77fb986bbf07/src/libraries/System.Diagnostics.Tracing/documentation/EventCounterTutorial.md)

PR #719, opened 2020-08-31
Supporting Event counter in .Net core

DavoudEshtehari:
The following counters are added to `.Net core 3.x`, `.Net standard 2.1` and above:

1.	**HardConnects**:  The number of actual connections that are being made to servers
2.	**HardDisconnects**: The number of actual disconnects that are being made to servers
3.	**ActiveHardConnections**: The number of actual active connections that are being made to servers 
4.	**SoftConnects**: The number of connections we get from the pool 
5.	**SoftDisconnects**: The number of connections we return to the pool 
6.	**ActiveSoftConnections**: The number of active pool connections 
7.	**NumberOfNonPooledConnections**: The number of connections that are not using connection pooling
8.	**NumberOfPooledConnections**: The number of connections that are managed by the connection pooler 
9.	**NumberOfActiveConnectionPoolGroups**: The number of unique connection strings
10.	**NumberOfInactiveConnectionPoolGroups**: The number of unique connection strings waiting for pruning 
11.	**NumberOfActiveConnectionPools**: The number of connection pools
12.	**NumberOfInactiveConnectionPools**: The number of connection pools 
13.	**NumberOfActiveConnections**: The number of connections currently in-use
14.	**NumberOfFreeConnections** : The number of connections currently available for use
15.	**NumberOfStasisConnections**: The number of connections currently waiting to be made ready for use
16.	**NumberOfReclaimedConnections**: The number of connections we reclaim from GC'd external connections

For `Microsoft.Data.SqlClient.EventSource`, you can use [.Net core global CLI tools](https://docs.microsoft.com/en-us/dotnet/core/diagnostics/#dotnet-counters): `dotnet-counters` and `dotnet-trace` in Windows or Linux and [PerfView](https://github.com/microsoft/perfview) in Windows.

- **dotnet-counters** monitor Microsoft.Data.SqlClient.EventSource -p <process-id>
- **PerfView** /onlyProviders=*Microsoft.Data.SqlClient.EventSource:EventCounterIntervalSec=1 collect

related issue #523


src/Microsoft.Data.SqlClient/netfx/src/Microsoft.Data.SqlClient.csproj
```diff
@@ -318,18 +318,26 @@
...
     <Compile Include="Microsoft\Data\SqlClient\SqlColumnEncryptionCspProvider.cs" />
     <Compile Include="Microsoft\Data\SqlClient\SqlColumnEncryptionEnclaveProvider.cs" />
-    <Compile Include="Microsoft\Data\SqlClient\SqlCommand.cs" />
-    <Compile Include="Microsoft\Data\SqlClient\SqlCommandBuilder.cs" />
+    <Compile Include="Microsoft\Data\SqlClient\SqlCommand.cs">
+      <SubType>Component</SubType>
+    </Compile>
+    <Compile Include="Microsoft\Data\SqlClient\SqlCommandBuilder.cs">
+      <SubType>Component</SubType>
+    </Compile>
```
cheenamalhotra, line 326: Quick note: Please revert unintentional changes from this file.

Wraith2:
When trying to use pre-processor definitions to include functionality it's hard to make sure that you include code only where it is supported and make sure it is included in later builds using definitions that don't exist yet. For example using `#if NETCOREAPP3 || NETSTANDARD21` is ok but when a NET50 build target or later is added the code won't be included without a time consuming audit of the codebase to add it to these lists. 

Annoying as it is, the best way to avoid this is to is to use exclusions. so something like `#if !NETCOREAPP20 && !NETCOREAPP21 && !NETSTANDARD20` because the code is then included in any new TFM's that are added and it should be the case that any new TFM contains the required api.

roji (commented):
In addition to the below implementation comment, the de-facto standard seems to be to just use the package name/namespace - so I'd use Microsoft.Data.SqlClient (and omit the trailing .EventSource)

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlClientEventSource.NetCoreApp.cs
```diff
@@ -0,0 +1,384 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Diagnostics.Tracing;
+using System.Threading;
+
+namespace Microsoft.Data.SqlClient
+{
+    internal partial class SqlClientEventSource : EventSource
```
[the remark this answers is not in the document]
DavoudEshtehari: Hi Shay,
Among the event counter types, just `EventCounter` supports `.NetCore 2.0` and above. This is the only reason for choosing it for now.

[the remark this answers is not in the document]
roji: I see... I remember having a similar discussion back when first implementing this in Npgsql. As this is a new feature and the new polling counters are already available in .NET Core 3.0, I'd advise going with the latter... Otherwise you're penalizing perf of modern applications using perf counters for the sake of having the new perf counters in pre-3.0 .NET...

Another thought is that if you have actual demand for this in pre-.NET 3.0, it's always possible to use conditional complication, using the polling counters in >3.0 and the current code when targeting older TFMs (though I'd really wait to see that this is required).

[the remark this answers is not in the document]
cheenamalhotra: Hi Shay,

We support NET Core 2.1 applications and aim to provide full feature support in all supported environments if possible. However, it's not clearly defined whether it is "required" for .NET Core 2.1 or not. It does keep us little behind from using latest features, but that's only for best feature compatibility and user experience.

But if there're new features that brings considerable performance difference/user experience, we're definitely in favor of splitting implementation and keeping it TFM specific, like we've been doing always. Our only ask would be does it make that big of a difference that we maintain double code for .NET Core 2.1 and 3.1+? Or we can continue it as it is for code simplicity, and when .NET Core 2.1 goes out of support, we can update this to new Counter classes for future.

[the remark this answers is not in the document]
cheenamalhotra: On the contrary,
I see value in splitting classes as it is in this case as there's a lot to offer in .NET Core 3.1 and .NET Standard 2.1 DLLs.

cc @DavoudEshtehari

DavoudEshtehari:
> When trying to use pre-processor definitions to include functionality it's hard to make sure that you include code only where it is supported and make sure it is included in later builds using definitions that don't exist yet. For example using `#if NETCOREAPP3 || NETSTANDARD21` is ok but when a NET50 build target or later is added the code won't be included without a time consuming audit of the codebase to add it to these lists.
> 
> Annoying as it is, the best way to avoid this is to is to use exclusions. so something like `#if !NETCOREAPP20 && !NETCOREAPP21 && !NETSTANDARD20` because the code is then included in any new TFM's that are added and it should be the case that any new TFM contains the required api.

Great point @Wraith2,

Our approach is to keep update the directive definitions in the `csproj` file to avoid redundant directive conditions in addition to support next .Net target frameworks. 
So, according to the current supported frameworks `netcoreapp2.1;netcoreapp3.1;netstandard2.0;netstandard2.1` , the following definition supports `.Netcore 3.1` and above:
```
  <PropertyGroup Condition="'$(TargetGroup)' == 'netcoreapp' AND '$(TargetFramework)' != 'netcoreapp2.1'">
    <DefineConstants>$(DefineConstants);NETCORE3</DefineConstants>
  </PropertyGroup>
```

DavoudEshtehari:
> In addition to the below implementation comment, the de-facto standard seems to be to just use the package name/namespace - so I'd use Microsoft.Data.SqlClient (and omit the trailing .EventSource)

This would be a good idea. The only concern is this is a breaking change.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlClientEventSource.NetCoreApp.cs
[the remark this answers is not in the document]
roji: @cheenamalhotra I understand, but the .NET Core performance counters feature itself is new - the dotnet counters tool wasn't even introduced until .NET Core 3.0, if memory serves. The feature called "performance counters" in .NET Framework is similar in goals, but entirely different in the way it works and incompatible with .NET Core, so perf counters in .NET Core can be considered essentially a new feature in .NET Core 3.0, IMHO.

In other words, I'd be quite surprised if anyone using versions of .NET Core lower than 3.0 would be interested in this, and it seems like a pity to (significantly) compromise perf for users of .NET Core 3.0 and above (or IMHO even to maintain two implementations).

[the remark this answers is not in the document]
cheenamalhotra: Thanks for context @roji 👍 
Linking documentation for readers/reviewers: [Event Counters](https://docs.microsoft.com/en-us/dotnet/core/diagnostics/event-counters)

Even though the APIs exist in .NET Core 2.1, from official Microsoft documentation it sounds like we can consider this for .NET Core 3.1 and .NET Standard 2.1 onwards.

@David-Engel your thoughts?

roji:
>> In addition to the below implementation comment, the de-facto standard seems to be to just use the package name/namespace - so I'd use Microsoft.Data.SqlClient (and omit the trailing .EventSource)
>
> This would be a good idea. The only concern is this is a breaking change.

Did you already have an EventSource in .NET Core before this PR? I'm not sure what the backwards compat bar is on EventSource names.

In any case, the perf counters (which are new) could go into a new EventSource as well with the new naming, even if that could be slightly confusing.

src/Microsoft.Data.SqlClient/netcore/src/Common/src/Microsoft/Data/ProviderBase/DbConnectionFactory.cs
```diff
@@ -373,6 +378,7 @@ private void PruneConnectionPoolGroups(object state)
                         // otherwise process entry which may move it from active to idle
                         if (entry.Value.Prune())
                         { // may add entries to _poolsToRelease
+                            SqlClientEventSource.Log.ActiveConnectionPoolGroupRequest(false);
```
karinazhou, line 381: Should we call `ActiveConnectionPoolGroupRequest()` in the else condition? There is another `newConnectionPoolGroups.Add()` called there.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
```diff
@@ -247,6 +247,10 @@
     <Compile Include="Microsoft\Data\SqlClient\EnclaveProviderBase.cs" />
     <Compile Include="Microsoft\Data\SqlClient\EnclaveSessionCache.cs" />
 	</ItemGroup>
+  <!-- netcoreapp 3.1 & netstandard 2.1 and above -->
+  <ItemGroup Condition="'$(OSGroup)' != 'AnyOS' AND '$(TargetFramework)' != 'netcoreapp2.1' AND '$(TargetFramework)' != 'netstandard2.0'">
+    <Compile Include="Microsoft\Data\SqlClient\SqlClientEventSource.NetCoreApp3.cs" />
```
karinazhou, line 252: Since you have added condition for this file, there is no need to add suffix number '3' in its name. Starting from netcoreapp 3.1 and netstandard 2.1, this can be supported and extended in the future.

DavoudEshtehari: I believe the file name will distinct the aim of this type for a developer just by seeing the file name beside the other files before getting into it.

karinazhou: For now, netcoreapp 3.1 and netstandard 2.1 are the versions we support. It can be extended to support .net core 5.0 or other .net standard versions. By that time, the file name could be a bit misleading. This is just a suggestion anyway.

src/Microsoft.Data.SqlClient/netcore/src/Common/src/Microsoft/Data/ProviderBase/DbConnectionFactory.cs
```diff
@@ -138,6 +139,7 @@ internal DbConnectionInternal CreatePooledConnection(DbConnectionPool pool, DbCo
             DbConnectionInternal newConnection = CreateConnection(options, poolKey, poolGroupProviderInfo, pool, owningObject, userOptions);
             if (null != newConnection)
             {
+                SqlClientEventSource.Log.HardConnectRequest();
```
karinazhou, line 142: Should we also call `PooledConnectionRequest()` here since we generated a new pooled connection?

src/Microsoft.Data.SqlClient/netcore/src/Common/src/Microsoft/Data/ProviderBase/DbConnectionFactory.cs
```diff
@@ -124,6 +124,7 @@ internal DbConnectionInternal CreateNonPooledConnection(DbConnection owningConne
             DbConnectionInternal newConnection = CreateConnection(connectionOptions, poolKey, poolGroupProviderInfo, null, owningConnection, userOptions);
             if (null != newConnection)
             {
+                SqlClientEventSource.Log.HardConnectRequest();
```
karinazhou, line 127: Should we also call `NonPooledConnectionRequest()` here since we generate a non-pooled connection?

src/Microsoft.Data.SqlClient/netcore/src/Common/src/Microsoft/Data/ProviderBase/DbConnectionFactory.cs
DavoudEshtehari: The `newConnectionPoolGroups` in the else clause keeps the current connection that its `Prune()` was not successful. So, it means there is no change in the number of active connection pool groups.

TaffarelJr:
Is there a timeline for when this code will be released? We're trying to optimize our connection pools in our .NET Core 3.1 app (running in Azure), and could really use this feature.

I found this [announcement](https://erikej.github.io/sqlclient/2020/05/04/mds-version2-preview3.html), but I can't seem to get it to work - even in the latest previews.

JRahnama:
> Is there a timeline for when this code will be released? We're trying to optimize our connection pools in our .NET Core 3.1 app (running in Azure), and could really use this feature.
> 
> I found this [announcement](https://erikej.github.io/sqlclient/2020/05/04/mds-version2-preview3.html), but I can't seem to get it to work - even in the latest previews.

Regarding `EventCounter`, our next preview will be released toward end of October and we are trying to make Eventcounter  a part of that release.

`EventSource` is already included in our driver. The easiest way to do so is using Perfview with run or collect command with
`perfview /onlyProviders=*Microsoft.Data.SqlClient.EventSource run <Your_Application.exe> ` and set the location of your file to the folder where your ` .exe ` file resides in.
There is another way by adding a listener to your application implementing `EventListener`  and set that to write the traces to a file our any desired output console.

src/Microsoft.Data.SqlClient/netcore/src/Common/src/Microsoft/Data/ProviderBase/DbConnectionFactory.cs
karinazhou: OK. I see. The `newConnectionPoolGroups` is the subset of the old `_connectionPoolGroups` after we removing the ones to be pruned.

src/Microsoft.Data.SqlClient/netcore/src/Common/src/Microsoft/Data/ProviderBase/DbConnectionFactory.cs
karinazhou: Refer to the .net framework implementation.

src/Microsoft.Data.SqlClient/netcore/src/Common/src/Microsoft/Data/ProviderBase/DbConnectionFactory.cs
karinazhou: Refer to the .net framework implementation

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
```diff
@@ -15,6 +15,7 @@
 using Microsoft.Identity.Client;
 using Microsoft.Data.SqlClient.TestUtilities;
 using Xunit;
+using System.Linq;
```
karinazhou, line 18: This is unnecessary for this file.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
```diff
@@ -694,7 +695,7 @@ protected override void OnEventSourceCreated(EventSource eventSource)
             {
                 if (eventSource.Name.Equals("Microsoft.Data.SqlClient.EventSource"))
                 {
-                    // Collect all traces for better code coverage
+                    //// Collect all traces for better code coverage
```
karinazhou, line 698: Unnecessary change.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/ProviderBase/DbConnectionPool.cs
```diff
@@ -293,6 +294,7 @@ internal void TransactionEnded(Transaction transaction, DbConnectionInternal tra
                 // connections, we'll put it back...
                 if (0 <= entry)
                 {
+                    SqlClientEventSource.Log.FreeConnectionRequest();
```
karinazhou, line 297: We should decrease the counter here according to here:
https://github.com/dotnet/SqlClient/blob/master/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/ProviderBase/DbConnectionPool.cs#L307

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlClientEventSource.NetCoreApp.cs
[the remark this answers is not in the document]
cheenamalhotra: Resolving as implementation targeted for netcoreapp3.1+ and netstandard2.1+

src/Microsoft.Data.SqlClient/tests/ManualTests/Microsoft.Data.SqlClient.ManualTesting.Tests.csproj
```diff
@@ -233,6 +233,10 @@
     <Compile Include="TracingTests\DiagnosticTest.cs" />
     <Compile Include="TracingTests\FakeDiagnosticListenerObserver.cs" />
   </ItemGroup>
+  <!--.Net core 3.0 & .Net standard 2.1 and above-->
+  <ItemGroup Condition="'$(TargetGroup)'=='netcoreapp' AND !$(TargetFramework.StartsWith('netcoreapp2')) AND !$(ReferenceType.Contains('netcoreapp2')) AND !$(ReferenceType.Contains('NetStandard2.0'))">
```
johnnypham, line 237: The ReferenceType conditions are not needed. When running the tests against .NET Standard 2.1, the target framework is set to `netcoreapp3.1`.

pi3k14:
We are now looking at December and this was ready in September ...
What is not going on?

DavoudEshtehari:
Hi @pi3k14,
According to the higher priority on the other tasks and the recent GA release, working on it was postponed to later. The project and milestone will be updated after planning.

winseros:
@DavoudEshtehari, maybe any help wanted on this one?

DavoudEshtehari:
We always welcome other contributors' help @winseros.
To contribute to this work item you can create a new fork from [this](https://github.com/DavoudEshtehari/SqlClient/tree/EventCounter) source code first. The remaining task is the assess and validate the counters through the comprehensive unit tests.

winseros:
Hi @DavoudEshtehari, I inspected the PR and see there are some tests in the [EventCounterTest.cs](https://github.com/DavoudEshtehari/SqlClient/blob/fceefd9346022076b18fe5744d9433e5e0785e01/src/Microsoft.Data.SqlClient/tests/ManualTests/TracingTests/EventCounterTest.cs#L34-L35).

So there have to be more sophisticated tests or assertions?

DavoudEshtehari:
Hi @winseros,

Happy holidays. 🙂 
Those tests assess the implemented event counters. Investigating their accuracy according to where these counters are called needs a profound study of the driver to discover the new possible spots and adding related tests if it's required.

roji:
@DavoudEshtehari you may want to check out the event counter testing approach being done by @ajcvickers for EF in https://github.com/dotnet/efcore/pull/23826. Actually setting up an EventListener may be quite flaky (and not to mention slow) - the idea is to simply use reflection to access the counters directly.

DavoudEshtehari:
> the idea is to simply use reflection to access the counters directly

Hi @roji,

After checking the source code, we found EF core increment/decrement the inner variables without checking if `EventSource` is enabled by the `IsEnabled()` method and the only use of this method is before creating the `PollingCounter` objects inside the `EntityFrameworkEventSource .OnEventCommand` event.
The inner variables always manipulate; even, when there is no listener. May I ask you to clarify its reason if it was intentional?
Generally, we skip manipulating the inner variables when there is no listener by checking the `IsEnabled()`. So, we couldn't use reflection in these tests.

roji:
@DavoudEshtehari the logic is that the increment/decrement is too cheap to justify the IsEnabled check ([ASP.NET do the same with the event counters](https://github.com/dotnet/aspnetcore/blob/master/src/Servers/Kestrel/Core/src/Internal/Infrastructure/KestrelEventSource.cs)). In fact, logic branches (conditions) frequently turn out to be more expensive at the CPU level than the code they encompass.

Note the difference with normal, non-polling event counters, where calling the method actually triggers the event being sent, as opposed to a simple increment/decrement. In that case the condition is indeed justified since sending the event is an expensive operation.

omatrot:
Is there something we could do to allow this merge request to make progress? I have a weird problem with connection pooling and can't diagnose it without Event counters.

src/Microsoft.Data.SqlClient/tests/ManualTests/Microsoft.Data.SqlClient.ManualTesting.Tests.csproj
```diff
@@ -231,6 +231,10 @@
...
```
cheenamalhotra, line 235: When reference type is .NET Standard, target framework for tests will still be .NET Core.

```suggestion
  <ItemGroup Condition="'$(TargetGroup)'=='netcoreapp' AND !$(TargetFramework.StartsWith('netcoreapp2'))">
```

cheenamalhotra (approved):
Just 1 pending suggestion, rest approved.

karinazhou (approved):

PR #720, opened 2020-09-02
Update support.md

cheenamalhotra:


karinazhou:
Are we going to include supported SQL Server versions in this file in the future?

cheenamalhotra:
> Are we going to include supported SQL Server versions in this file in the future?

That info will be updated in Microsoft Docs, hence the redirection.

karinazhou (approved):

David-Engel (approved):

PR #730, opened 2020-09-18
Feature | Azure Active Directory Managed Identity authentication support

### LINKED ISSUE #616: Azure Managed Identity Support
Is there a plan to add support for Azure Managed Identity or is that something the Azure SDK team should support?

cheenamalhotra:
Fixes #616 

Introduces AAD Managed Identity support with below authentication mode keywords:
- Active Directory Managed Identity
- Active Directory MSI _(for cross MS SQL drivers compatibility)_

Connection strings applicable:
```cs
// For System Assigned Managed Identity
"Server:{serverURL}; Authentication=Active Directory MSI; Initial Catalog={db};"

// For System Assigned Managed Identity
"Server:{serverURL}; Authentication=Active Directory Managed Identity; Initial Catalog={db};"

// For User Assigned Managed Identity
"Server:{serverURL}; Authentication=Active Directory MSI; User Id={ObjectIdOfManagedIdentity}; Initial Catalog={db};"

// For User Assigned Managed Identity
"Server:{serverURL}; Authentication=Active Directory Managed Identity; User Id={ObjectIdOfManagedIdentity}; Initial Catalog={db};"
```

_Dev Notes:_
No new dependency is added to driver.
This feature is supported with direct Rest API calls to IMDS and Identity endpoints (as MSSQL JDBC driver)

Reference: https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/how-to-use-vm-token#get-a-token-using-http


David-Engel (commented):
We also need to provide an synonym "ActiveDirectoryMSI" for the authentication keyword value to align the connection string with what all our other drivers use. The other drivers will eventually add a synonym for "ActiveDirectoryManagedIdentity".

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/Common/DbConnectionStringCommon.cs
```diff
@@ -672,40 +682,41 @@ internal static partial class DbConnectionStringDefaults
...
         internal const bool Replication = false;
-        internal const string WorkstationID = "";
+        internal const string WorkstationID = _emptyString;
         internal const string TransactionBinding = "Implicit Unbind";
         internal const int ConnectRetryCount = 1;
         internal const int ConnectRetryInterval = 10;
         internal static readonly SqlAuthenticationMethod Authentication = SqlAuthenticationMethod.NotSpecified;
         internal const SqlConnectionColumnEncryptionSetting ColumnEncryptionSetting = SqlConnectionColumnEncryptionSetting.Disabled;
-        internal const string EnclaveAttestationUrl = "";
+        internal const string EnclaveAttestationUrl = _emptyString;
```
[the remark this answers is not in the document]
cheenamalhotra: I tried that, but `string.Empty` is not a constant, raises error [CS0133](https://docs.microsoft.com/en-us/dotnet/csharp/misc/cs0133)

src/Microsoft.Data.SqlClient/netfx/src/Microsoft.Data.SqlClient.csproj
```diff
@@ -442,10 +445,10 @@
   </ItemGroup>
   <!-- Resources  -->
   <ItemGroup>
-    <Compile Include="Resources\$(ResxFileName).Designer.cs">
+    <Compile Include="Resources\Strings.Designer.cs">
       <AutoGen>True</AutoGen>
       <DesignTime>True</DesignTime>
-      <DependentUpon>$(ResxFileName).resx</DependentUpon>
+      <DependentUpon>Strings.resx</DependentUpon>
```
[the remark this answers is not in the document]
cheenamalhotra: I donno why we do that tbh, it's auto-generated everytime. I'm sure same is being built into the NuGet. It never stays as variable.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ActiveDirectoryAuthenticationProvider.cs
```diff
@@ -3,12 +3,16 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
+using System.Collections;
```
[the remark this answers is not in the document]
cheenamalhotra: Done

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/AADUtility.cs
```diff
@@ -17,10 +21,186 @@ public static async Task<string> AzureActiveDirectoryAuthenticationCallback(stri
...
+#if NETFRAMEWORK
+        private static readonly HttpClient s_defaultHttpClient = new HttpClient();
+#else
+        private static readonly HttpClient s_defaultHttpClient = new HttpClient(new HttpClientHandler() { CheckCertificateRevocationList = true });
+#endif
+
+        private const string AzureVmImdsApiVersion = "&api-version=2018-02-01";
+        private const string AccessToken = "access_token";
+        private const string Resource = "https://database.windows.net/";
+
```
David-Engel, line 48: Nit: double blank lines

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AzureManagedIdentityAuthenticationProvider.cs
```diff
@@ -0,0 +1,251 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
```
David-Engel, line 1: In general for this file, I would add some logging lines so that we can see things like entering and exiting the main routines, success and failure, perhaps unexpected return values from the endpoints, etc.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AzureManagedIdentityAuthenticationProvider.cs
```diff
@@ -0,0 +1,251 @@
...
+
+                // If user assigned managed identity is specified, include object ID parameter in request
+                string objectIdParameter = parameters.UserId != default
+                    ? $"&object_id={parameters.UserId}"
+                    : string.Empty;
+
+                // Craft request as per the MSI protocol
+                var requestUrl = isAppServicesMsiAvailable
+                    ? $"{msiEndpoint}?resource={parameters.Resource}{objectIdParameter}{AzureSystemApiVersion}"
+                    : $"{AzureVmImdsEndpoint}?resource={parameters.Resource}{objectIdParameter}{AzureVmImdsApiVersion}";
```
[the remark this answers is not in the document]
cheenamalhotra: Users can provide only "UserId" as resource is internally maintained.
If "UserId" contains invalid characters, we get below errors **instantly**:

When it contains space:

`
System.AggregateException: One or more errors occurred. ---> System.AggregateException: One or more errors occurred. ---> Microsoft.Data.SqlClient.SqlException: Received a non-retryable error. Identity Response Code: BadRequest, Response: {"error":"invalid_request","error_description":"Identity not found"}
   at Microsoft.Data.SqlClient.AzureManagedIdentityAuthenticationProvider.<AcquireTokenAsync>d__15.MoveNext()
`

When it contains invalid characters:

`
System.AggregateException: One or more errors occurred. ---> System.AggregateException: One or more errors occurred. ---> Microsoft.Data.SqlClient.SqlException: Received a non-retryable error. Identity Response Code: BadRequest, Response: Bad Request
   at Microsoft.Data.SqlClient.AzureManagedIdentityAuthenticationProvider.<AcquireTokenAsync>d__15.MoveNext()
`

Whereas if I escape UserId and then send request, it goes into retry mode and continues to retry as per settings.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -1587,6 +1587,19 @@ private void Login(ServerInfo server, TimeoutTimer timeout, string newPassword,
...
+                requestedFeatures |= TdsEnums.FeatureExtension.FedAuth;
+                _federatedAuthenticationInfoRequested = true;
+                _fedAuthFeatureExtensionData =
+                    new FederatedAuthenticationFeatureExtensionData
+                    {
+                        libraryType = TdsEnums.FedAuthLibrary.MSAL,
+                        authentication = ConnectionOptions.Authentication,
+                        fedAuthRequiredPreLoginResponse = _fedAuthRequired
+                    };
+            }
```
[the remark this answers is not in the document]
cheenamalhotra: I initially configured it differently but had to set same settings later. Fixed 👍

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AzureManagedIdentityAuthenticationProvider.cs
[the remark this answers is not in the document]
cheenamalhotra: What if we check if UserId is valid for URL use and throw error if not before sending it forward to IMDS?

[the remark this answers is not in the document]
David-Engel: In this case, UserId should be a GUID. But I'm not sure that's for us to enforce. What if Azure decides that object_id can be something other than a GUID? I'm inclined to send what the user provides and rely on the IMDS parameter validation.

> Whereas if I escape UserId and then send request, it goes into retry mode and continues to retry as per settings.

If you escape and send an invalid UserId, it goes into retry mode? I would expect to get a non-retryable error from the service if we did that. Like invalid object_id or identity not found kind of error. What error is being returned that we are retrying on?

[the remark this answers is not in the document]
cheenamalhotra: Sorry my bad, it does throw error instantly. I was debugging on machine where retry was looping due to service not available.

`
One or more errors occurred. ---> System.AggregateException: One or more errors occurred. ---> Microsoft.Data.SqlClient.SqlException: Received a non-retryable error. Identity Response Code: BadRequest, Response: {"error":"invalid_request","error_description":"Identity not found"}
   at Microsoft.Data.SqlClient.AzureManagedIdentityAuthenticationProvider.<AcquireTokenAsync>d__15.MoveNext()
`

I will add tests to validate the same and we can also escape User Id in that case.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AzureManagedIdentityAuthenticationProvider.cs
cheenamalhotra: Will do that in both providers, as both seem to lack EventSource tracing logs.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AzureManagedIdentityAuthenticationProvider.cs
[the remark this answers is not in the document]
cheenamalhotra: Done

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AzureManagedIdentityAuthenticationProvider.cs
cheenamalhotra: Added

David-Engel (approved):

sopelt:
Hi @cheenamalhotra ,

this looks very nice for production. For local development and non-MSI AAD scenarios I was hoping for some support via Azure.Identity. What are your thoughts on that?

Thanks!

cheenamalhotra:
Hi @sopelt 

We've discussed that internally and we'd have loved to do that if that was feasible. As of now we could not take dependency on Azure.Identity as it's partially compatible with our driver and does not support .NET Framework 4.6 that SqlClient does. We'd be able to support that in a future version hopefully by taking a leap on the lowest Full Framework version SqlClient supports.

As of this feature and upcoming stable v2.1, Visual Studio (for local development) and Azure CLI (for non-MSI implicit) based token support shall remain unsupported.

ericsampson:
@cheenamalhotra I'd also love Azure.Identity support for seamless localdev.

Once this PR is in, what is the recommended way to connect in localdev using the VS/CLI token stores?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -1302,6 +1302,8 @@ private void Login(ServerInfo server, TimeoutTimer timeout, string newPassword,
                 || ConnectionOptions.Authentication == SqlAuthenticationMethod.ActiveDirectoryInteractive
                 || ConnectionOptions.Authentication == SqlAuthenticationMethod.ActiveDirectoryDeviceCodeFlow
                 || ConnectionOptions.Authentication == SqlAuthenticationMethod.ActiveDirectoryServicePrincipal
+                || ConnectionOptions.Authentication == SqlAuthenticationMethod.ActiveDirectoryManagedIdentity
```
karinazhou, line 1305: Better to update the comments with new required AD Authentication methods.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsEnums.cs
```diff
@@ -5,6 +5,7 @@
 using System;
 using System.Data;
 using System.Diagnostics;
+using System.Net;
```
karinazhou, line 8: Unnecessary library.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AzureManagedIdentityAuthenticationProvider.cs
```diff
@@ -0,0 +1,254 @@
...
+#endif
+
+        private const string AzureSystemApiVersion = "&api-version=2019-08-01";
+        private const string AzureVmImdsApiVersion = "&api-version=2018-02-01";
+        private const string AccessToken = "access_token";
+        private const string Expiry = "expires_on";
+        private const int FileTimeLength = 10;
+
+        private const int DefaultRetryTimeout = 0;
+        private const int DefaultMaxRetryCount = 5;
```
karinazhou, line 29: Any reason we pick this max retry count?

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/AADUtility.cs
```diff
@@ -17,10 +20,185 @@ public static async Task<string> AzureActiveDirectoryAuthenticationCallback(stri
             AuthenticationResult result = await authContext.AcquireTokenAsync(resource, clientCred);
             if (result == null)
             {
-                throw new InvalidOperationException($"Failed to retrieve an access token for {resource}");
+                throw new Exception($"Failed to retrieve an access token for {resource}");
```
karinazhou, line 23: Why we change the exception type here?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ActiveDirectoryAuthenticationProvider.cs
```diff
@@ -198,19 +201,23 @@ public override void BeforeUnload(SqlAuthenticationMethod authentication)
                     try
                     {
                         result = await app.AcquireTokenSilent(scopes, account).ExecuteAsync();
+                        SqlClientEventSource.Log.TryTraceEvent("AcquireTokenAsync | Acquired access token for {0} auth mode. Expiry Time: {1}", parameters.AuthenticationMethod, result.ExpiresOn);
```
karinazhou, line 204: Can we add extra information to distinguish these three identical event trace messages (Line: 204, 209, and 215)?

JRahnama (approved):

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/AADUtility.cs
cheenamalhotra: I didn't think "InvalidOperationException" is an appropriate exception type here.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AzureManagedIdentityAuthenticationProvider.cs
cheenamalhotra: Based on Retry guidance:
https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/how-to-use-vm-token#retry-guidance

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AzureManagedIdentityAuthenticationProvider.cs
```diff
@@ -0,0 +1,254 @@
...
+                        catch (HttpRequestException e)
+                        {
+                            if (attempts == maxRetryCount)
+                            {
+                                throw;
+                            }
+                            SqlClientEventSource.Log.TryTraceEvent("SendAsyncWithRetry | Exception occurred {0} | Attempting retry: {1} of {2}", e.Message, attempts, maxRetryCount);
+                        }
+
+                        if (s_waitBeforeRetry)
```
karinazhou, line 230: Though this is for the unit test purpose, the following strategy is always executed. Is this necessary when we do retry? If it is only for testing, maybe we can enable this when in debug mode.

cheenamalhotra: I think you're talking about a different file (AADUtility), but I can answer it here.
I implemented retry in tests to avoid random errors since this retry policy is going to come into action when the token cannot be acquired in the first attempt. If we successfully acquire token in first attempt, no retry occurs, so this remains no-op, there are no side-effects.

cheenamalhotra:
@ericsampson 

You would be able to override our implementation of "Active Directory Managed Identity" authentication provider with your own class that implements "[SqlAuthenticationProvider](https://docs.microsoft.com/en-us/dotnet/api/microsoft.data.sqlclient.sqlauthenticationprovider?view=sqlclient-dotnet-core-2.0)" and register it in your application by calling "[SetProvider](https://docs.microsoft.com/en-us/dotnet/api/microsoft.data.sqlclient.sqlauthenticationprovider.setprovider?view=sqlclient-dotnet-core-2.0)" API.

There you can simply acquire token from Azure.Identity library when implementing "AcquireTokenAsync", using "[DefaultAzureCredential](https://docs.microsoft.com/en-us/dotnet/api/azure.identity.defaultazurecredential?view=azure-dotnet)" as they support .NET Core and .NET Framework 4.6.1+ and their "GetToken" API does the following, as you can find documented:

_The following credential types if enabled will be tried, in order:_
- _EnvironmentCredential_
- _ManagedIdentityCredential_
- _SharedTokenCacheCredential_
- _VisualStudioCredential_
- _VisualStudioCodeCredential_
- _AzureCliCredential_
- _InteractiveBrowserCredential_

Once you call SetProvider, all future connection requests for "Active Directory Managed Identity" authentication mode from driver will go through your implementation using same connection strings.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AzureManagedIdentityAuthenticationProvider.cs
karinazhou: The code is duplicated in both files (AzureManagedIdentityAuthenticationProvider.cs and AADUtility.cs) which don't happen only in tests. The set of actions under the `s_waitBeforeRetry` condition will always be executed if the first attempt fails for some reason. It is acceptable if this retry logic is required all the time.

karinazhou (approved):

ericsampson:
@cheenamalhotra thanks for the notes, that seems doable. 

Is there an existing "Extensions" NuGet package that it would be possible to put something like that code in, where it would be OK to have the Identity library dependency? 
I was just thinking that if so, I could make a PR so that everyone could take advantage of the functionally.

chamikasandamal:
Is this available now with 3.x or 5.x?

ErikEJ:
There is no version 5 of this provider.

This feature is available in 3.0 preview, which will RTM quite soon, and this version work with both .nEt Core 3 and .Net 5

PR #734, opened 2020-09-20
Merge shared files

Wraith2:
Merge of more identical or similar files. 

The only notable difference is in SignatureVerificationCache.cs
netcore: `_cache.Set(cacheLookupKey, result, DateTimeOffset.UtcNow.AddDays(10));`
netfx: `_cache.Set(cacheLookupKey, result, DateTimeOffset.UtcNow.AddHours(8));`

I chose the 10 days but I think this needs consideration before merging.


karinazhou:
I don't see any troubles using 10 days for both .net framework and .net core. I have re-run the failing pipelines and the previously failing tests look like random ones.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/OperationAbortedException.cs
```diff
@@ -7,7 +7,7 @@
 
 namespace Microsoft.Data
 {
```
karinazhou, line 9: For function `internal static OperationAbortedException Aborted(Exception inner)`, can we keep the ` ADP.TraceExceptionAsReturnValue(e);` to have event source trace here?

Wraith2: Done.

karinazhou (approved):

cheenamalhotra (approved):

PR #765, opened 2020-10-21
Upgrade MSAL library version

cheenamalhotra:
Also removed Asserts that compared error messages not owned by SqlClient.


David-Engel (approved):

DavoudEshtehari (approved):
Can I ask why did you just remove these five expected messages among all tests?

cheenamalhotra:
> Can I ask why did you just remove these five expected messages among all tests?

As mentioned above, these error messages are "not owned by SqlClient", so they are subject to change from MSAL. It's reasonable to validate that an exception occurred but comparing error message texts to be of match didn't seems a very good idea for our library tests.

johnnypham (approved):

karinazhou (approved):

PR #770, opened 2020-10-22
Implement public client application global cache

cheenamalhotra:
Fixes an issue where driver continues to prompt for credentials when requesting access token for AAD authentication, mainly visible when using Interactive authentication.

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ActiveDirectoryAuthenticationProvider.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ActiveDirectoryAuthenticationProvider.cs
@@ -312,13 +312,7 @@ namespace Microsoft.Data.SqlClient
 
         private IPublicClientApplication GetPublicClientAppInstance(PublicClientAppKey publicClientAppKey)
         {
-            IPublicClientApplication clientApplicationInstance;
-
-            if (s_pcaMap.ContainsKey(publicClientAppKey))
-            {
-                s_pcaMap.TryGetValue(publicClientAppKey, out clientApplicationInstance);
-            }
-            else
+            if (!s_pcaMap.TryGetValue(publicClientAppKey, out IPublicClientApplication clientApplicationInstance))
             {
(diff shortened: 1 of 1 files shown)


karinazhou:
Though the testing for this requires human interaction, can we have some comments or sample code somewhere about how to validate it manually?

DavoudEshtehari:
Assume that there are two active tokens with different access to two databases; The first token with index 0 in the accounts has access to `database_1` and the second token has access to `database_2`.
In the next connection with the initial catalog = database_2 always we return the first token if we don't specify the user.
Maybe we need more data to pick the correct token.

David-Engel:
> Assume that there are two active tokens with different access to two databases; The first token with index 0 in the accounts has access to `database_1` and the second token has access to `database_2`.
> In the next connection with the initial catalog = database_2 always we return the first token if we don't specify the user.
> Maybe we need more data to pick the correct token.

The scope of the token is database.windows.net, not the specific database. So using the same token for two databases will be fine. The token identifies the user to the database and database ACL settings determine whether that user is authorized to access the database. There really should never be two tokens in the situation you described. The first token would have been used when the second was requested.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ActiveDirectoryAuthenticationProvider.cs
```diff
@@ -320,5 +297,168 @@ private class CustomWebUi : ICustomWebUi
...
+            }
+
+            public override bool Equals(object obj)
+            {
+                if (obj == null)
+                {
+                    return false;
+                }
+
+                PublicClientAppKey pcaKey = obj as PublicClientAppKey;
```
johnnypham, line 400: If this conversion fails, a null ref exception will be thrown below. This check should be sufficient.

```
if (!(obj is PublicClientAppKey pcaKey))
{
    return false;
}
```

karinazhou (approved):

JRahnama (approved):

DavoudEshtehari:
For the first named-pipe connection the exception is:

DavoudEshtehari (approved):

PR #665, opened 2020-07-23
Interop Changes Part 1 - Cleanup Interop source

cheenamalhotra:
Cleanups only.

PR #777, opened 2020-10-28
Fix | Fixes issue in Integrated Security auth on Windows (Managed SNI)

cheenamalhotra:
Fixed a regression in latest preview release from this change in SNIProxy:
https://github.com/dotnet/SqlClient/pull/665/files#diff-a9b4126995728b5ff4d99d7c10d8e6af5259ee81a790274529b028e7213b2707L101


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIProxy.cs
```diff
@@ -96,7 +96,8 @@ internal void GenSspiClientContext(SspiClientContextStatus sspiClientContextStat
                 inSecurityBufferArray = Array.Empty<SecurityBuffer>();
             }
 
-            int tokenSize = MaxTokenSize;
```
David-Engel, line 99: I think this is the only reference to MaxTokenSize, so we can remove the declaration at the top.

David-Engel (approved):

Issue #755, opened 2020-10-09
Deadlock in WriteToServerAsync (possible reentrancy issue)


coderb:
(version nuget 4.7.0)

I'm not sure sure if you'll consider this a bug or by design, but there seems to be an inconsistency between the sync and async versions of bulk copy- `WriteToServer` and `WriteToServerAsync`.

When executing inside a transaction, the sync version `WriteToServer` allows for commands to be executed during the call to `WriteToServer` but prior to the first bulk inserted row.

When using the async version `WriteToServerAsync` if you try to execute a command even prior to the first row being sent it results in a deadlock (waiting on a semphore).

Unfortunately I'm porting a large amount of code to use async and this is going to be a big job to address.  Is this something that you would consider addresing in SqlClient?

karinazhou:
Hi @coderb , 
Do you have MARS enabled in your connection string when doing bulk copy? If MARS is enabled for your case, could you try to disable it to see how things are going?

We got a relevant issue https://github.com/dotnet/SqlClient/issues/716 about WriteToServerAsync. And we are currently investigating it.

I notice that you are using nuget 4.7.0 which I think is the System.Data.SqlClient driver. Could you also try your application with Microsoft.Data.SqlClient v2.0.1 (latest stable version)? The source code for both drivers is slightly different and more fixes are applied to MDS. 

And if possible, could you provide us a standalone repro application?

Thanks,

coderb:
@karinazhou hi karina- nope, no MARS, in fact i didn't even know it existed.

so the actual version i'm using the is latest stable nuget 2.01.   i misspoke in the original post.

do you consider what i am describing to be a bug or (assuming what i'm seeing is accurate) is it by design?

thanks

karinazhou:
@coderb  It sounds like a bug in the driver if the same code works for WriteToServer but fails for WriteToServerAsync due to deadlock. In order to understand what your application is doing and see what triggers the deadlock in the driver, it would be helpful with the reproducible code if you can provide. 

Thanks,

coderb:
@karinazhou please see the reproduction at https://github.com/coderb/SqlClientBulkDeadlock

karinazhou:
@coderb  Thank you for the repro. We will test it locally.

karinazhou:
Hi @coderb ,

I tried your repro app. Actually, the deadlock is not inside WriteToServerAsync(). If you comment out this piece of code:

```
            var sql = "select count(*) from " + TableName;
[...]

PR #779, opened 2020-10-30
Fix Bulk Copy Async deadlock with custom IDataReader using SqlDataReader internally.

cheenamalhotra:
Fix for issue #755

User-scenario:
- Using 1 SqlConnection, perform Bulk Copy Async with datasource an instance of an implementation of IDataReader (internally calls SqlDataReader.ExecuteScalarAsync).

Currently, entire driver's async operations are single threaded when it comes to "parser" activity, and every Async operation is done with lock acquired from "SqlInternalConnectionTds._parserLock".

Pseudo scenario:
```
SqlConnection (I allow only 1 async operation to use parser) {
    SqlBulkCopy.WriteToServerAsync (parser lock acquired) {
        ReadFromRowSourceAsync (
            SimpleDataReader.enumerator => SqlCommand.ExecuteScalarAsync (I also need parser lock - Waits for lock forever!!!!!)
        )
        WriteToServerInternalRestAsync ( << parser lock is needed here)
    }
}
```

This PR fixes this behavior as under:
```
SqlConnection (I allow only 1 async operation to use parser) {
    SqlBulkCopy.WriteToServerAsync (parser lock acquired) {
        ReadFromRowSourceAsync (**release lock for reader) (
            SimpleDataReader.enumerator => SqlCommand.ExecuteScalarAsync (parser lock acquired - data read successfully)
        (**acquire lock again) )
        WriteToServerInternalRestAsync ( << parser lock is needed here)
    }
}
```
cc @karinazhou

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
@@ -1159,6 +1159,11 @@ namespace Microsoft.Data.SqlClient
             }
             else
             { // This will call Read for DataRows, DataTable and IDataReader (this includes all IDataReader except DbDataReader)
+              // Release lock to prevent possible deadlocks
+                SqlInternalConnectionTds internalConnection = _connection.GetOpenTdsConnection();
+                bool semaphoreLock = internalConnection._parserLock.CanBeReleasedFromAnyThread;
+                internalConnection._parserLock.Release();
+
                 _hasMoreRowToCopy = false;
                 try
                 {
@@ -1174,6 +1179,11 @@ namespace Microsoft.Data.SqlClient
                     {
                         throw;
                     }
+                }
+                finally
+                {
+                    _insideRowsCopiedEvent = false;
+                    internalConnection._parserLock.Wait(canReleaseFromAnyThread: semaphoreLock);
                 }
                 return null;
             }
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
@@ -1182,7 +1182,6 @@ namespace Microsoft.Data.SqlClient
                 }
                 finally
                 {
-                    _insideRowsCopiedEvent = false;
                     internalConnection._parserLock.Wait(canReleaseFromAnyThread: semaphoreLock);
                 }
                 return null;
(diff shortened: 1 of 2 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
```diff
@@ -1175,6 +1180,11 @@ private Task ReadFromRowSourceAsync(CancellationToken cts)
                         throw;
                     }
                 }
+                finally
+                {
+                    _insideRowsCopiedEvent = false;
```
karinazhou, line 1185: Why do we set the `_insideRowsCopiedEvent `?

cheenamalhotra: I've reverted that, wasn't intended.

Wraith2:
The diff is a bit noisy because the IDE has fixed a lot of comments. It looks like the only change is in the bulk copy and that it saves a property releases the parser lock and then in a finally block waits, should it not re-acquire per your description?

karinazhou:
After debugging this more, here is what I find for general scenarios where we use connection A for SqlBulkCopy instance and connection B for the data source (associated with SqlCommand). The _parserLock is associated with each connection. The lock behavior before the change is like this:
 
```
SqlBulkCopy.WriteToServerAsync() --> WriteRowSourceToServerAsync() -->  require lock on _parserLockA of connectionA
    SqlBulkCopy.ReadFromRowSourceAsync
    (
            SimpleDataReader.enumerator => SqlCommand.ExecuteScalarAsync -->  **require lock on _parserLockB of connectionB**
    )
      SqlBulkCopy.WriteToServerInternalRestAsync()  -->  need the lock on _parserLockA of connectionA
```

The lock behavior after the change is like this:

```
SqlBulkCopy.WriteToServerAsync() -->  WriteRowSourceToServerAsync()   -->  require lock on _parserLockA of connectionA
    SqlBulkCopy.ReadFromRowSourceAsync
    (
         ** release lock on _parserLockA of connectionA
        SimpleDataReader.enumerator => SqlCommand.ExecuteScalarAsync -->  **require lock on _parserLockB of connectionB**
         ** require lock on _parserLockA of connectionA
    )
      SqlBulkCopy.WriteToServerInternalRestAsync()  -->  need the lock on _parserLockA of connectionA
```    

It looks like the original design of SqlBulkCopy doesn't expect the same connection to be used for both data source and SqlBulkCopy instance. So far, it looks OK to release the semaphore if we don't actually need it at this point to avoid the deadlock in the single connection edge case.

cheenamalhotra:
> The diff is a bit noisy because the IDE has fixed a lot of comments. It looks like the only change is in the bulk copy and that it saves a property releases the parser lock and then in a finally block waits, should it not re-acquire per your description?

It's a weird design but that's how it is.
ParserLock semaphore is of size 1, when a thread calls 'Wait' it decrements the parserlock size to 0, blocking all other 'Wait' calls till the same thread calls 'Release'. That's how we reacquire lock by calling 'Wait'.

karinazhou (approved):

David-Engel (approved):

Issue #659, opened 2020-07-21
Async reader may return wrong results on MacOS, Linux, WSL, Docker


Nercury:
### Describe the bug

Sometimes the query may return wrong results under high load, on Non-Windows clients.

### To reproduce

- .NET Core
- MSSQL instance
- Client is not running on Windows (reproduced on MacOS or Linux)
- Query uses async reader methods
- High load that saturates the connection pool and causes timeouts.

Bug is seen more often when `MultipleActiveResultSets=True` (MARS) is in the connection string. It might be prevented by disabling MARS (unclear at this point).

The issue is hard to reproduce and rare, therefore [this project was created with the code that attempts to simulate the situation](https://github.com/Nercury/mars-wrong-results).

This program starts 2000 concurrent connections that runs `select @Id as Id`
statement. Each `Id` is different, and the query result should always return 
the id that was queried, yet under certain circumstances that's not the case. It looks like
sometimes the reader returns the stale data from random previous connection that experienced the timeout.

There are more details in the linked project.

Network latency can be more easily simulated using this method here: https://github.com/trainline/SqlClient659

### Expected behavior

Under no circumstances a select statement should return a different result. 

### Further technical details

Microsoft.Data.SqlClient version: 4.8.1 or earlier
.NET target: netcoreapp2.0 - netcoreapp3.1
SQL Server version: SQL Server Web Edition 13.00.5426.0.v1, SQL Server Express Edition 14.00.3281.6.v1
Operating system: macOS 10.15.5, Docker container

### Workaround

Do not use this SQL client on linux/docker.

Always check the id of the returned result and retry or crash if it is wrong.

To decrease the likelihood of this or maybe prevent it (unclear at this point), do not use MARS feature, if you are using .NET Core with MSSQL database, ensure that the connection string does not contain `MultipleActiveResultSets=True`.

JRahnama:
@Nercury thanks for reaching out. Issue #422 is similar to the problem you have and is under investigation.

DavoudEshtehari:
Hi @Nercury,

Thank you for producing the sample app and describing the issue well.
I ran your sample app multiple times at a virtual machine running `Ubuntu 18.0`, `.Net core 3.1`, and `SQL Server 19` with `MARS` enabled and `connection pool` equals **100**. I didn't get any `Wrong results` as you expected.

Do you have any comments?

Nercury:
> I didn't get any `Wrong results` as you expected.
> 
> Do you have any comments?

[...]

Issue #792, opened 2020-11-12
Queries with many Guids in Linux: System.InvalidOperationException: Invalid operation. The connection is closed.


simader:
I have a query which runs under windows without problems, but if we get around 4000 guids there the query fails on a Linux App Service 
`var issues = await context.Issues.Where(i => guids.Contains(i.Id)).ToListAsync();`

```
Exception message: System.InvalidOperationException: Invalid operation. The connection is closed.
Stack trace:
at Microsoft.Data.SqlClient.SqlCommand.<>c.b__164_0(Task1 result) at System.Threading.Tasks.ContinuationResultTaskFromResultTask2.InnerInvoke()
at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location where exception was thrown ---
at System.Threading.Tasks.Task.ExecuteWithThreadLocal(Task& currentTaskSlot, Thread threadPoolThread)
--- End of stack trace from previous location where exception was thrown ---
at Microsoft.EntityFrameworkCore.Storage.RelationalCommand.ExecuteReaderAsync(RelationalCommandParameterObject parameterObject, CancellationToken cancellationToken)
at Microsoft.EntityFrameworkCore.Storage.RelationalCommand.ExecuteReaderAsync(RelationalCommandParameterObject parameterObject, CancellationToken cancellationToken)
at Microsoft.EntityFrameworkCore.Storage.RelationalCommand.ExecuteReaderAsync(RelationalCommandParameterObject parameterObject, CancellationToken cancellationToken)
at Microsoft.EntityFrameworkCore.Query.Internal.QueryingEnumerable1.AsyncEnumerator.InitializeReaderAsync(DbContext _, Boolean result, CancellationToken cancellationToken) at Microsoft.EntityFrameworkCore.SqlServer.Storage.Internal.SqlServerExecutionStrategy.ExecuteAsync[TState,TResult](TState state, Func4 operation, Func4 verifySucceeded, CancellationToken cancellationToken) at Microsoft.EntityFrameworkCore.Query.Internal.QueryingEnumerable1.AsyncEnumerator.MoveNextAsync()
[...]

Issue #786, opened 2020-11-08
InvalidOperationException when writing a lot of packets asynchronously and using managed SNI implementation


Berreek:
### Describe the bug
Hey guys. 
We are executing stored procedure that sometimes is taking as parameter a data table with a little bit of data (in the prepared repro case it is arround 0.5mb). When we try to do that we sometimes get following exception:

```
System.InvalidOperationException: Invalid operation. The connection is closed.
   at Microsoft.Data.SqlClient.TdsParserStateObject.WritePacket(Byte flushMode, Boolean canAccumulate)
   at Microsoft.Data.SqlClient.TdsParserStateObject.WriteBytes(ReadOnlySpan`1 b, Int32 len, Int32 offsetBuffer, Boolean canAccumulate, TaskCompletionSource`1 completion, Byte[] array)
   at Microsoft.Data.SqlClient.TdsParserStateObject.WriteByteArray(Byte[] b, Int32 len, Int32 offsetBuffer, Boolean canAccumulate, TaskCompletionSource`1 completion)
   at Microsoft.Data.SqlClient.TdsParser.WriteString(String s, Int32 length, Int32 offset, TdsParserStateObject stateObj, Boolean canAccumulate)
   at Microsoft.Data.SqlClient.TdsValueSetter.SetString(String value, Int32 offset, Int32 length)
   at Microsoft.Data.SqlClient.TdsRecordBufferSetter.SetString(SmiEventSink sink, Int32 ordinal, String value, Int32 offset, Int32 length)
   at Microsoft.Data.SqlClient.Server.ValueUtilsSmi.SetString_Unchecked(SmiEventSink_Default sink, ITypedSettersV3 setters, Int32 ordinal, String value, Int32 offset, Int32 length)
   at Microsoft.Data.SqlClient.Server.ValueUtilsSmi.SetString_LengthChecked(SmiEventSink_Default sink, ITypedSettersV3 setters, Int32 ordinal, SmiMetaData metaData, String value, Int32 offset)
   at Microsoft.Data.SqlClient.Server.ValueUtilsSmi.SetCompatibleValue(SmiEventSink_Default sink, ITypedSettersV3 setters, Int32 ordinal, SmiMetaData metaData, Object value, ExtendedClrTypeCode typeCode, Int32 offset)
[...]

PR #796, opened 2020-11-13
Fixes "InvalidOperationException" errors by performing async operations in SemaphoreSlim

### LINKED ISSUE #459: NotSupportedException & InvalidOperationException for unknown reasons

cheenamalhotra:
Fixes #459 #54 #786 #792 
Rework over PR #579 with helpful ideas from @roji and @DavoudEshtehari 

This PR fixes the issue where below exception would occur randomly due to [SslStream.WriteAsync](https://docs.microsoft.com/en-us/dotnet/api/system.net.security.sslstream.writeasync?view=netcore-3.1) collisions.

```
System.InvalidOperationException: Invalid operation. The connection is closed.
   at Microsoft.Data.SqlClient.TdsParserStateObject.WritePacket(Byte flushMode, Boolean canAccumulate)
```

The design change introduces use of `SemaphoreSlim` with `SNISslStream` and `SNINetworkStream` overrides to ensure only 1 task is allowed to read/write to stream asynchronously at a time.

#### Not fixed in this issue
- _#422 remains unfixed, investigations in progress._
- _#659 remains unfixed, investigations in progress._

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNINpHandle.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNINpHandle.cs
@@ -189,6 +189,8 @@ namespace Microsoft.Data.SqlClient.SNI
             try
             {
                 SNIPacket errorPacket;
+                lock (this)
+                {
                     packet = null;
                     try
                     {
@@ -220,6 +222,7 @@ namespace Microsoft.Data.SqlClient.SNI
                     }
                     return TdsEnums.SNI_SUCCESS;
                 }
+            }
             finally
             {
                 SqlClientEventSource.Log.TrySNIScopeLeaveEvent(scopeID);
@@ -268,43 +271,40 @@ namespace Microsoft.Data.SqlClient.SNI
                 bool releaseLock = false;
                 try
                 {
-                    lock (this)
+                    // is the packet is marked out out-of-band (attention packets only) it must be
+                    // sent immediately even if a send of recieve operation is already in progress
+                    // because out of band packets are used to cancel ongoing operations
+                    // so try to take the lock if possible but continue even if it can't be taken
+                    if (packet.IsOutOfBand)
                     {
-                        // is the packet is marked out out-of-band (attention packets only) it must be
-                        // sent immediately even if a send of recieve operation is already in progress
-                        // because out of band packets are used to cancel ongoing operations
-                        // so try to take the lock if possible but continue even if it can't be taken
-                        if (packet.IsOutOfBand)
-                        {
-                            Monitor.TryEnter(this, ref releaseLock);
-                        }
-                        else
-                        {
-                            Monitor.Enter(this);
-                            releaseLock = true;
-                        }
+                        Monitor.TryEnter(this, ref releaseLock);
+                    }
+                    else
+                    {
+                        Monitor.Enter(this);
+                        releaseLock = true;
+                    }
 
-                        // this lock ensures that two packets are not being written to the transport at the same time
-                        // so that sending a standard and an out-of-band packet are both written atomically no data is
-                        // interleaved
-                        lock (_sendSync)
+                    // this lock ensures that two packets are not being written to the transport at the same time
+                    // so that sending a standard and an out-of-band packet are both written atomically no data is
+                    // interleaved
+                    lock (_sendSync)
+                    {
+                        try
                         {
-                            try
-                            {
-                                packet.WriteToStream(_stream);
-                                return TdsEnums.SNI_SUCCESS;
-                            }
-                            catch (ObjectDisposedException ode)
-                            {
-                                SqlClientEventSource.Log.TrySNITraceEvent("<sc.SNI.SNINpHandle.Send |SNI|ERR> ObjectDisposedException message = {0}.", ode.Message);
-                                return ReportErrorAndReleasePacket(packet, ode);
-                            }
-                            catch (IOException ioe)
-                            {
-                                SqlClientEventSource.Log.TrySNITraceEvent("<sc.SNI.SNINpHandle.Send |SNI|ERR> IOException message = {0}.", ioe.Message);
+                            packet.WriteToStream(_stream);
+                            return TdsEnums.SNI_SUCCESS;
+                        }
+                        catch (ObjectDisposedException ode)
+                        {
+                            SqlClientEventSource.Log.TrySNITraceEvent("<sc.SNI.SNINpHandle.Send |SNI|ERR> ObjectDisposedException message = {0}.", ode.Message);
+                            return ReportErrorAndReleasePacket(packet, ode);
+                        }
+                        catch (IOException ioe)
+                        {
+                            SqlClientEventSource.Log.TrySNITraceEvent("<sc.SNI.SNINpHandle.Send |SNI|ERR> IOException message = {0}.", ioe.Message);
 
-                                return ReportErrorAndReleasePacket(packet, ioe);
-                            }
+                            return ReportErrorAndReleasePacket(packet, ioe);
                         }
                     }
                 }
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNINpHandle.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNINpHandle.cs
@@ -189,39 +189,36 @@ namespace Microsoft.Data.SqlClient.SNI
             try
             {
                 SNIPacket errorPacket;
-                lock (this)
+                packet = null;
+                try
                 {
-                    packet = null;
-                    try
-                    {
-                        packet = RentPacket(headerSize: 0, dataSize: _bufferSize);
-                        packet.ReadFromStream(_stream);
+                    packet = RentPacket(headerSize: 0, dataSize: _bufferSize);
+                    packet.ReadFromStream(_stream);
 
-                        if (packet.Length == 0)
-                        {
-                            errorPacket = packet;
-                            packet = null;
-                            var e = new Win32Exception();
-                            SqlClientEventSource.Log.TrySNITraceEvent("<sc.SNI.SNINpHandle.Receive |SNI|ERR> packet length is 0.");
-                            return ReportErrorAndReleasePacket(errorPacket, (uint)e.NativeErrorCode, 0, e.Message);
-                        }
-                    }
-                    catch (ObjectDisposedException ode)
+                    if (packet.Length == 0)
                     {
                         errorPacket = packet;
                         packet = null;
-                        SqlClientEventSource.Log.TrySNITraceEvent("<sc.SNI.SNINpHandle.Receive |SNI|ERR> ObjectDisposedException message = {0}.", ode.Message);
-                        return ReportErrorAndReleasePacket(errorPacket, ode);
+                        var e = new Win32Exception();
+                        SqlClientEventSource.Log.TrySNITraceEvent("<sc.SNI.SNINpHandle.Receive |SNI|ERR> packet length is 0.");
+                        return ReportErrorAndReleasePacket(errorPacket, (uint)e.NativeErrorCode, 0, e.Message);
                     }
-                    catch (IOException ioe)
-                    {
-                        errorPacket = packet;
-                        packet = null;
-                        SqlClientEventSource.Log.TrySNITraceEvent("<sc.SNI.SNINpHandle.Receive |SNI|ERR> IOException message = {0}.", ioe.Message);
-                        return ReportErrorAndReleasePacket(errorPacket, ioe);
-                    }
-                    return TdsEnums.SNI_SUCCESS;
                 }
+                catch (ObjectDisposedException ode)
+                {
+                    errorPacket = packet;
+                    packet = null;
+                    SqlClientEventSource.Log.TrySNITraceEvent("<sc.SNI.SNINpHandle.Receive |SNI|ERR> ObjectDisposedException message = {0}.", ode.Message);
+                    return ReportErrorAndReleasePacket(errorPacket, ode);
+                }
+                catch (IOException ioe)
+                {
+                    errorPacket = packet;
+                    packet = null;
+                    SqlClientEventSource.Log.TrySNITraceEvent("<sc.SNI.SNINpHandle.Receive |SNI|ERR> IOException message = {0}.", ioe.Message);
+                    return ReportErrorAndReleasePacket(errorPacket, ioe);
+                }
+                return TdsEnums.SNI_SUCCESS;
             }
             finally
             {
@@ -286,7 +283,7 @@ namespace Microsoft.Data.SqlClient.SNI
                     }
 
                     // this lock ensures that two packets are not being written to the transport at the same time
-                    // so that sending a standard and an out-of-band packet are both written atomically no data is 
+                    // so that sending a standard and an out-of-band packet are both written atomically no data is
                     // interleaved
                     lock (_sendSync)
                     {
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNINpHandle.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNINpHandle.cs
@@ -93,7 +93,7 @@ namespace Microsoft.Data.SqlClient.SNI
                 }
 
                 _sslOverTdsStream = new SslOverTdsStream(_pipeStream);
-                _sslStream = new SslStream(_sslOverTdsStream, true, new RemoteCertificateValidationCallback(ValidateServerCertificate), null);
+                _sslStream = new SNISslStream(_sslOverTdsStream, true, new RemoteCertificateValidationCallback(ValidateServerCertificate));
 
                 _stream = _pipeStream;
                 _status = TdsEnums.SNI_SUCCESS;
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNINpHandle.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNINpHandle.cs
@@ -189,37 +189,37 @@ namespace Microsoft.Data.SqlClient.SNI
             try
             {
                 SNIPacket errorPacket;
-                packet = null;
-                try
-                {
-                    packet = RentPacket(headerSize: 0, dataSize: _bufferSize);
-                    packet.ReadFromStream(_stream);
+                    packet = null;
+                    try
+                    {
+                        packet = RentPacket(headerSize: 0, dataSize: _bufferSize);
+                        packet.ReadFromStream(_stream);
 
-                    if (packet.Length == 0)
+                        if (packet.Length == 0)
+                        {
+                            errorPacket = packet;
+                            packet = null;
+                            var e = new Win32Exception();
+                            SqlClientEventSource.Log.TrySNITraceEvent("<sc.SNI.SNINpHandle.Receive |SNI|ERR> packet length is 0.");
+                            return ReportErrorAndReleasePacket(errorPacket, (uint)e.NativeErrorCode, 0, e.Message);
+                        }
+                    }
+                    catch (ObjectDisposedException ode)
                     {
                         errorPacket = packet;
                         packet = null;
-                        var e = new Win32Exception();
-                        SqlClientEventSource.Log.TrySNITraceEvent("<sc.SNI.SNINpHandle.Receive |SNI|ERR> packet length is 0.");
-                        return ReportErrorAndReleasePacket(errorPacket, (uint)e.NativeErrorCode, 0, e.Message);
+                        SqlClientEventSource.Log.TrySNITraceEvent("<sc.SNI.SNINpHandle.Receive |SNI|ERR> ObjectDisposedException message = {0}.", ode.Message);
+                        return ReportErrorAndReleasePacket(errorPacket, ode);
                     }
+                    catch (IOException ioe)
+                    {
+                        errorPacket = packet;
+                        packet = null;
+                        SqlClientEventSource.Log.TrySNITraceEvent("<sc.SNI.SNINpHandle.Receive |SNI|ERR> IOException message = {0}.", ioe.Message);
+                        return ReportErrorAndReleasePacket(errorPacket, ioe);
+                    }
+                    return TdsEnums.SNI_SUCCESS;
                 }
-                catch (ObjectDisposedException ode)
-                {
-                    errorPacket = packet;
-                    packet = null;
-                    SqlClientEventSource.Log.TrySNITraceEvent("<sc.SNI.SNINpHandle.Receive |SNI|ERR> ObjectDisposedException message = {0}.", ode.Message);
-                    return ReportErrorAndReleasePacket(errorPacket, ode);
-                }
-                catch (IOException ioe)
-                {
-                    errorPacket = packet;
-                    packet = null;
-                    SqlClientEventSource.Log.TrySNITraceEvent("<sc.SNI.SNINpHandle.Receive |SNI|ERR> IOException message = {0}.", ioe.Message);
-                    return ReportErrorAndReleasePacket(errorPacket, ioe);
-                }
-                return TdsEnums.SNI_SUCCESS;
-            }
             finally
             {
                 SqlClientEventSource.Log.TrySNIScopeLeaveEvent(scopeID);
@@ -268,40 +268,43 @@ namespace Microsoft.Data.SqlClient.SNI
                 bool releaseLock = false;
                 try
                 {
-                    // is the packet is marked out out-of-band (attention packets only) it must be
-                    // sent immediately even if a send of recieve operation is already in progress
-                    // because out of band packets are used to cancel ongoing operations
-                    // so try to take the lock if possible but continue even if it can't be taken
-                    if (packet.IsOutOfBand)
+                    lock (this)
                     {
-                        Monitor.TryEnter(this, ref releaseLock);
-                    }
-                    else
-                    {
-                        Monitor.Enter(this);
-                        releaseLock = true;
-                    }
-
-                    // this lock ensures that two packets are not being written to the transport at the same time
-                    // so that sending a standard and an out-of-band packet are both written atomically no data is
-                    // interleaved
-                    lock (_sendSync)
-                    {
-                        try
+                        // is the packet is marked out out-of-band (attention packets only) it must be
+                        // sent immediately even if a send of recieve operation is already in progress
+                        // because out of band packets are used to cancel ongoing operations
+                        // so try to take the lock if possible but continue even if it can't be taken
+                        if (packet.IsOutOfBand)
                         {
-                            packet.WriteToStream(_stream);
-                            return TdsEnums.SNI_SUCCESS;
+                            Monitor.TryEnter(this, ref releaseLock);
                         }
-                        catch (ObjectDisposedException ode)
+                        else
                         {
-                            SqlClientEventSource.Log.TrySNITraceEvent("<sc.SNI.SNINpHandle.Send |SNI|ERR> ObjectDisposedException message = {0}.", ode.Message);
-                            return ReportErrorAndReleasePacket(packet, ode);
+                            Monitor.Enter(this);
+                            releaseLock = true;
                         }
-                        catch (IOException ioe)
-                        {
-                            SqlClientEventSource.Log.TrySNITraceEvent("<sc.SNI.SNINpHandle.Send |SNI|ERR> IOException message = {0}.", ioe.Message);
 
-                            return ReportErrorAndReleasePacket(packet, ioe);
+                        // this lock ensures that two packets are not being written to the transport at the same time
+                        // so that sending a standard and an out-of-band packet are both written atomically no data is
+                        // interleaved
+                        lock (_sendSync)
+                        {
+                            try
+                            {
+                                packet.WriteToStream(_stream);
+                                return TdsEnums.SNI_SUCCESS;
+                            }
+                            catch (ObjectDisposedException ode)
+                            {
+                                SqlClientEventSource.Log.TrySNITraceEvent("<sc.SNI.SNINpHandle.Send |SNI|ERR> ObjectDisposedException message = {0}.", ode.Message);
+                                return ReportErrorAndReleasePacket(packet, ode);
+                            }
+                            catch (IOException ioe)
+                            {
+                                SqlClientEventSource.Log.TrySNITraceEvent("<sc.SNI.SNINpHandle.Send |SNI|ERR> IOException message = {0}.", ioe.Message);
+
+                                return ReportErrorAndReleasePacket(packet, ioe);
+                            }
                         }
                     }
                 }
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNISslStream.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNISslStream.cs
@@ -14,30 +14,28 @@ namespace Microsoft.Data.SqlClient.SNI
     /// </summary>
     internal class SNISslStream : SslStream
     {
-        private readonly ConcurrentQueueSemaphore _writeQueueSemaphore;
-        private readonly ConcurrentQueueSemaphore _readQueueSemaphore;
+        private readonly ConcurrentQueueSemaphore _writeAsyncQueueSemaphore;
+        private readonly ConcurrentQueueSemaphore _readAsyncQueueSemaphore;
 
         public SNISslStream(Stream innerStream, bool leaveInnerStreamOpen, RemoteCertificateValidationCallback userCertificateValidationCallback)
             : base(innerStream, leaveInnerStreamOpen, userCertificateValidationCallback)
         {
-            _writeQueueSemaphore = new ConcurrentQueueSemaphore(1);
-            _readQueueSemaphore = new ConcurrentQueueSemaphore(1);
+            _writeAsyncQueueSemaphore = new ConcurrentQueueSemaphore(1);
+            _readAsyncQueueSemaphore = new ConcurrentQueueSemaphore(1);
         }
 
         // Prevent the ReadAsync's collision by running task in Semaphore Slim
         public override Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
         {
-            _readQueueSemaphore.Wait();
-            return base.ReadAsync(buffer, offset, count, cancellationToken)
-                .ContinueWith(t => _readQueueSemaphore.Release(t.Result));
+            return _readAsyncQueueSemaphore.WaitAsync().ContinueWith<int>(t => base.ReadAsync(buffer, offset, count, cancellationToken).Result)
+                .ContinueWith(t => _readAsyncQueueSemaphore.Release(t.Result));
         }
 
         // Prevent the WriteAsync's collision by running task in Semaphore Slim
         public override Task WriteAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
         {
-            _writeQueueSemaphore.Wait();
-            return base.WriteAsync(buffer, offset, count, cancellationToken)
-                .ContinueWith(t => _writeQueueSemaphore.Release());
+            return _writeAsyncQueueSemaphore.WaitAsync().ContinueWith(t => base.WriteAsync(buffer, offset, count, cancellationToken))
+                .ContinueWith(t => _writeAsyncQueueSemaphore.Release());
         }
     }
 }
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNISslStream.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNISslStream.cs
@@ -28,9 +28,8 @@ namespace Microsoft.Data.SqlClient.SNI
         public override Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
         {
             _readQueueSemaphore.Wait();
-            Task<int> t = base.ReadAsync(buffer, offset, count, cancellationToken);
-            _readQueueSemaphore.Release();
-            return t;
+            return base.ReadAsync(buffer, offset, count, cancellationToken)
+                .ContinueWith(t => _readQueueSemaphore.Release(t.Result));
         }
 
         // Prevent the WriteAsync's collision by running task in Semaphore Slim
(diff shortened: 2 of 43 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
```diff
@@ -706,12 +706,17 @@ public override void SetAsyncCallbacks(SNIAsyncCallback receiveCallback, SNIAsyn
         /// <returns>SNI error code</returns>
         public override uint SendAsync(SNIPacket packet, SNIAsyncCallback callback = null)
         {
-            SNIAsyncCallback cb = callback ?? _sendCallback;
-            lock (this)
```
cheenamalhotra, line 710: Note: This lock seemed unnecessary to me, hence removed.
Just wanted to call out, it's removed cautiously.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObjectManaged.cs
```diff
@@ -72,13 +72,13 @@ internal override void AssignPendingDNSInfo(string userProtocol, string DNSCache
...
             ReadAsyncCallback(IntPtr.Zero, PacketHandle.FromManagedPacket(packet), error);
-            _sessionHandle.ReturnPacket(packet);
+            _sessionHandle?.ReturnPacket(packet);
         }
 
         internal void WriteAsyncCallback(SNIPacket packet, uint sniError)
         {
             WriteAsyncCallback(IntPtr.Zero, PacketHandle.FromManagedPacket(packet), sniError);
-            _sessionHandle.ReturnPacket(packet);
+            _sessionHandle?.ReturnPacket(packet);
```
cheenamalhotra, line 81: There's another issue in driver where this sessionHandle is null here, performing null check also fixes "Object reference not found" errors. But these sessionHandles are leaking data, which also need to be fixed. TBD.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNISslStream.cs
```diff
@@ -0,0 +1,44 @@
...
+            _writeQueueSemaphore = new ConcurrentQueueSemaphore(1);
+            _readQueueSemaphore = new ConcurrentQueueSemaphore(1);
+        }
+
+        // Prevent the ReadAsync's collision by running task in Semaphore Slim
+        public override Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
+        {
+            _readQueueSemaphore.Wait();
+            Task<int> t = base.ReadAsync(buffer, offset, count, cancellationToken);
+            _readQueueSemaphore.Release();
```
martyt, line 32: Shouldn't this be in a `.ContinueWith()` as used in `WriteAsync()`?

cheenamalhotra: I was trying something like this, but it made ReadAsync go synchronous:
```
public override Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
{
    _readQueueSemaphore.Wait();
    return base.ReadAsync(buffer, offset, count, cancellationToken)
        .ContinueWith(_ => { _readQueueSemaphore.Release(); return _.GetAwaiter().GetResult(); });
}
```

`Release()` is sync call, and returns `void`. Any ideas?

roji: Calling the sync Release in ReadAsync is async-over-sync, meaning that the thread will block instead being truly async, which is what ReadAsync advertises itself to be. If this is a long wait, this can cause thread pool starvation, as TP threads get synchronously blocked.

Is there any reason not to make this method async, and await on _readQueueSemaphore.WaitAsync?

martyt: @cheenamalhotra I was suggesting that the pattern used in `WriteAsync()` should be used here as well by moving the `_readQueueSemaphore.Release()` call to ` .ContinueWith()` (minus all the `.GetAwaiter().GetResult()` stuff) - but in light of @roji 's comments, I guess neither pattern is correct.

roji: Looking at this again, I accidentally wrote "Release" where I meant to write "Wait" above. Release doesn't actually block, but Wait certainly does

martyt: In that case, my original suggestion seems valid -- certainly the `.Wait()` will block, but that's exactly what has to happen to ensure that packets are read correctly.

```
            _readQueueSemaphore.Wait();
            return base.ReadAsync(buffer, offset, count, cancellationToken)
                .ContinueWith(t => _readQueueSemaphore.Release());
```

cheenamalhotra: @roji 

Also this is call to ConcurrentQueueSemaphore > `Wait()`, that delegates to Semaphore's `WaitAsync()` internally and waits on it. We can't call it async way as we need to block here.

cheenamalhotra: Thanks for the tip @roji !
I tested and hit perf issue with Blocking call to wait().. using WaitAsync() solves that issue, alongwith issue #422 is now getting fixed when specifying "Encrypt=true" ❤️ 

Thanks so much! :)

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNISslStream.cs
```diff
@@ -0,0 +1,41 @@
...
+            : base(innerStream, leaveInnerStreamOpen, userCertificateValidationCallback)
+        {
+            _writeAsyncQueueSemaphore = new ConcurrentQueueSemaphore(1);
+            _readAsyncQueueSemaphore = new ConcurrentQueueSemaphore(1);
+        }
+
+        // Prevent the ReadAsync's collision by running task in Semaphore Slim
+        public override Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
+        {
+            return _readAsyncQueueSemaphore.WaitAsync().ContinueWith<int>(t => base.ReadAsync(buffer, offset, count, cancellationToken).Result)
```
roji, line 30: Although this no longer blocks synchronously on the semaphore, calling Task.Result still blocks synchronously on ReadAsync (sync-over-async, with the associated TP starvation issues).

Is there any particular reason to avoid async/await here, i.e. something like the following:

```c#
public override async Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
{
    await _readAsyncQueueSemaphore.WaitAsync().ConfigureAwait(false);
    try {
        await base.ReadAsync(buffer, offset, count, cancellationToken).ConfigureAwait(false);
    } finally {
        _readAsyncQueueSemaphore.Release(t.Result);
    }
}
```

_readAsyncQueueSemaphore.WaitAsync could even be changed to return an IDisposable which calls Release, in which case the above could be simplified with `using`.

cheenamalhotra: I'm learning :) sure, will update it!

cheenamalhotra: We await on these async methods (ReadAsync/WriteAsync) in SslOverTdsStream class where `Stream` is actually in use. Since this class is mainly an extension to these APIs to ensure we don't execute simultaneous read/writes, we don't await here. We also don't want this class to be blocking (correction to my last statement) and only want calls to be sent in queue 1 at a time.

I also mentioned in another thread/issue, can repost it here. I think this could have been handled by `SslStream` class itself. This design of Semaphore Slim + Concurrent Queue on SslStream creates an extended class `SNISslStream` that we will now base our Stream on. It's very much an independent design, and maybe something for System.Net.Security to consider ? Since like us, many other clients can land into concurrency issues with this class (as per their documentation of limited support).

Have you come across similar case in Ngpsql?

roji: > We await on these async methods (ReadAsync/WriteAsync) in SslOverTdsStream class where Stream is actually in use. Since this class is mainly an extension to these APIs to ensure we don't execute simultaneous read/writes, we don't await here.

OK. But note that implementing the above ReadAsync wrapper with async/await or with ContinueWith will in the end produce exactly the same method - async/await is precisely about breaking the method up and wiring the different parts with callbacks. In the end, both produce a method that doesn't block and returns a Task.

In other words, async/await is just a simpler and safer way to express the same thing you're doing with ContinueWith (since the compiler is helping you out).

Re concurrent usage of streams... I know that NetworkStream does internally serialize concurrent requests via a queue, and that it's a think that is very much regretted nowadays, as the queue adds some significant (mandatory!) overhead for all network operations. At the end of the day, sockets cannot be used concurrently, and like any other non-thread-safe resource, it should IMHO be up to the application to perform any necessary synchronization to avoid concurrent usage. So I would be surprised if the BCL folk would agree to introducing a serializing queue mechanism inside SslStream...

I'm not sure what the exact requirements of SqlClient are, but in Npgsql locking/synchronization happen at a higher level than the stream. For example, I think SqlClient already employs high-level synchronization to prevent multiple commands to be executed concurrently on the same connection - that's the kind of stuff that Npgsql does as well. I'm not sure what other cases you have to deal with, when multiple writes are attempted concurrently...

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIStreams.cs
```diff
@@ -0,0 +1,97 @@
...
+        private readonly ConcurrentQueueSemaphore _readAsyncQueueSemaphore;
+
+        public SNISslStream(Stream innerStream, bool leaveInnerStreamOpen, RemoteCertificateValidationCallback userCertificateValidationCallback)
+            : base(innerStream, leaveInnerStreamOpen, userCertificateValidationCallback)
+        {
+            _writeAsyncQueueSemaphore = new ConcurrentQueueSemaphore(1);
+            _readAsyncQueueSemaphore = new ConcurrentQueueSemaphore(1);
+        }
+
+        // Prevent the ReadAsync's collision by running task in Semaphore Slim
```
David-Engel, line 28:
```suggestion
        // Prevent ReadAsync collisions by running the task in a Semaphore Slim
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIStreams.cs
```diff
@@ -0,0 +1,97 @@
...
+                return _readAsyncQueueSemaphore.WaitAsync()
+                    .ContinueWith(_ => base.ReadAsync(buffer, offset, count, cancellationToken).GetAwaiter().GetResult());
+            }
+            finally
+            {
+                _readAsyncQueueSemaphore.Release();
+            }
+        }
+
+        // Prevent the WriteAsync's collision by running task in Semaphore Slim
```
David-Engel, line 42:
```suggestion
        // Prevent the WriteAsync collisions by running the task in a Semaphore Slim
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIStreams.cs
```diff
@@ -0,0 +1,97 @@
...
+        private readonly ConcurrentQueueSemaphore _writeAsyncQueueSemaphore;
+        private readonly ConcurrentQueueSemaphore _readAsyncQueueSemaphore;
+
+        public SNINetworkStream(Socket socket, bool ownsSocket) : base(socket, ownsSocket)
+        {
+            _writeAsyncQueueSemaphore = new ConcurrentQueueSemaphore(1);
+            _readAsyncQueueSemaphore = new ConcurrentQueueSemaphore(1);
+        }
+
+        // Prevent the ReadAsync's collision by running task in Semaphore Slim
```
David-Engel, line 70:
```suggestion
        // Prevent the ReadAsync collisions by running the task in a Semaphore Slim
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIStreams.cs
```diff
@@ -0,0 +1,97 @@
...
+                return _readAsyncQueueSemaphore.WaitAsync()
+                    .ContinueWith(_ => base.ReadAsync(buffer, offset, count, cancellationToken).GetAwaiter().GetResult());
+            }
+            finally
+            {
+                _readAsyncQueueSemaphore.Release();
+            }
+        }
+
+        // Prevent the WriteAsync's collision by running task in Semaphore Slim
```
David-Engel, line 84:
```suggestion
        // Prevent the WriteAsync collisions by running the task in a Semaphore Slim
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNISslStream.cs
cheenamalhotra: > In other words, async/await is just a simpler and safer way to express the same thing you're doing with ContinueWith (since the compiler is helping you out).

I couldn't get async/await to fix things as ContinueWith does.. Maybe there're more areas in Managed SNI we need to scan and fix. I'll make a note of revisiting that when we do improvements in future. For now, we're in crunch for release so we'll let it stay ContinueWith since that makes a whole lot of difference.

> I'm not sure what the exact requirements of SqlClient are, but in Npgsql locking/synchronization happen at a higher level than the stream. For example, I think SqlClient already employs high-level synchronization to prevent multiple commands to be executed concurrently on the same connection - that's the kind of stuff that Npgsql does as well. I'm not sure what other cases you have to deal with, when multiple writes are attempted concurrently...

We have a whole lot of issues due to concurrency and I suspect the higher level of synchronization is broken somewhere. We definitely need to redesign and make things consistent in Managed SNI as it's not a clean design right now. This time we're going to fix issues for customers who are blocked with huge problems that are getting fixed here. We'll revisit this and try to set things straight! Thanks again for your thoughts and ideas, highly appreciated! :)

roji: Sure thing @cheenamalhotra. I'd still highly recommend not blocking the thread - the part where Task.Result is called - but putting a ContinueWith on it instead to make it truly async. Beyond that, async/await would improve code quality (and likely also perf), but both patterns are correct bug-wise.

cheenamalhotra: I can see why async/await was needed as we tested more and more cases (changed it).
But implementing async/awaits brings back #422 and Readers go to blocking. I'm continuing to look further.

cheenamalhotra: We're keeping the implementation as it is (using async/await) for now, as we need to dig more into other 2 issues.

I'll put some notes what we're facing issue:

Basically in #422 repro, we have parallel execution of sync APIs: `Connection.Open()` followed by `SqlCommand.ExecuteReader()` and when MARS is enabled, the driver only reads MARS packet asynchronously during `Connection.Open()`, rest data is read synchronously. We are suspecting these `ReadAsync()` calls are blocking the second sync API `ExecuteReader()` to read data as every other thread is simply waiting for packet to arrive that shall contain metadata information.

When I used the ContinueWith pattern above, that unblocked everything and I think that happened because these calls became synchronous. It also fixed #659 that also is on the same path. Let us know if you have ideas/tips to what we might be dealing with here, whether it's threadpool starvation or sync-async collision.. there're also so many locks in the code that are suspicious.

cheenamalhotra: @Wraith2 

Hope you're doing well! 
As you have much more experience in Managed SNI design, please feel free to chime in and share your thoughts 🙂

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIStreams.cs
```diff
@@ -26,12 +26,12 @@ public SNISslStream(Stream innerStream, bool leaveInnerStreamOpen, RemoteCertifi
...
         // Prevent the ReadAsync's collision by running task in Semaphore Slim
-        public override Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
+        public override async Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
         {
+            await _readAsyncQueueSemaphore.WaitAsync().ConfigureAwait(false);
             try
             {
-                return _readAsyncQueueSemaphore.WaitAsync()
-                    .ContinueWith(_ => base.ReadAsync(buffer, offset, count, cancellationToken).GetAwaiter().GetResult());
+                return await base.ReadAsync(buffer, offset, count, cancellationToken).ConfigureAwait(false);
```
roji, line 34: :heart:

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNISslStream.cs
Wraith2: I don't know about "design", I just fiddle with the existing implementation more than is healthy.

We should use `ConfigureAwait(false)` everywhere and then root cause any bugs that it exposes. The intention is that we do not care about which thread continuations are called on and those bugs are evidence that something somewhere relies on something we don't want.

If the semaphore slim is analogous to a multithread/task aware lock then it's a reasonable thing to do to replace the `lock` keyword usage with it. Otherwise I'd have to do some digging. 

So far I've only seen one reason for multiple send operations to be in flight at once and that's cancellation of a running query. What sort of queue depth are you expecting to see here and is it possible to use less expensive primitive to optimize for the 90% case of single packets?

On the idea of returning a disposable instead of using `try{}finally{}` I agree with the code purity but it has a perf penalty because the disposable will now be heap allocated as it becomes part of the closure, so I'd probably decide against it given how low level and frequently called this path is.

roji: > We should use ConfigureAwait(false) everywhere and then root cause any bugs that it exposes.

Absolutely.

Note that the latest compiler versions come with a warning for missing ConfigureAwait ([CA2007](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2007)), which can make your build fail if you forget one. This can be enabled as an error in an .editorconfig [as here](https://github.com/dotnet/efcore/blob/main/src/.editorconfig#L48), and you'd probably also need to set `<AnalysisLevel>latest</AnalysisLevel>` in your csproj.

> On the idea of returning a disposable instead of using try{}finally{} I agree with the code purity but it has a perf penalty because the disposable will now be heap allocated as it becomes part of the closure, so I'd probably decide against it given how low level and frequently called this path is.

Where is there a closure here?

Note also that if the disposable is a readonly struct, there's no heap allocation (as long as the struct type is being passed around and never cast to IDisposable). So this is a zero-allocation mechanism used frequently in high-perf scenarios ([see this example in EF Core](https://github.com/dotnet/efcore/blob/main/src/EFCore/Infrastructure/ConcurrencyDetectorCriticalSectionDisposer.cs)).

Wraith2: The specific case we're talking about is the async method, because it's async the compiler generated object for params and locals is the closure I'm refering to, perhaps I'm using incorrect terminology. Even if you use a struct it'll end up in the state object and if it goes async it'll get boxed. You can't use ref structs in an async method so that method isn't possible in this case.

cheenamalhotra: Thanks for the details.
I will continue on these issues and keep eye on async/await to happen everywhere and see if we can remove locks and come to SemaphoreSlim to control entry if needed.

I just realized [SemaphoreSlim ](https://docs.microsoft.com/en-us/dotnet/api/system.threading.semaphoreslim) is concurrent, so Queue can be removed and design can be simplified (learning 🙃)

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlUtil.cs
```diff
@@ -2131,4 +2132,48 @@ public static MethodInfo GetPromotedToken
...
+        {
+            var tcs = new TaskCompletionSource<bool>();
+            _queue.Enqueue(tcs);
+            _semaphore.WaitAsync().ContinueWith(t =>
+            {
+                if (_queue.TryDequeue(out TaskCompletionSource<bool> popped))
+                    popped.SetResult(true);
+            });
+            return tcs.Task;
+        }
```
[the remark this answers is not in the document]
cheenamalhotra: Removed queue as it's not needed and SemaphoreSlim is concurrent.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIStreams.cs
```diff
@@ -0,0 +1,99 @@
...
+            }
+        }
+
+        // Prevent the WriteAsync collisions by running the task in a Semaphore Slim
+        public override async Task WriteAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
+        {
+            await _writeAsyncQueueSemaphore.WaitAsync().ConfigureAwait(false);
+            try
+            {
+                await base.WriteAsync(buffer, offset, count, cancellationToken).ConfigureAwait(false);
```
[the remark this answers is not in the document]
Wraith2: I'm not sure that's the case. By my understanding any point of entry/exit from the state machine caused by an await should instruct the state machine how to deal with the awaiter and as such all calls to await should be `ConfigureAwait(false)` or you will get mixed behaviour. Take a look at the codegen and two calls to `Task.Delay()` in [this sharplab example](https://sharplab.io/#v2:D4AQTAjAsAUCAMACEECsBuWsQGZlkQGFEBvWRC5PEADmQDZEBZACgEpTzLuUBOFliF6IAIgHsAKgHcxAQSkBDAJYAXAM4sI8NmwB0EsQGUVAJyUA7AObs2mGNwC+We5VzIIjWgwA8FlQD5RSRl5ZXUWP0QANwUAGwBXAFM2LgoyF24KSJNElRjYxABeaLiku0zXYRB6XRFE2IUAT014bV1CMXMAMyVLeJzQ1RYuuLVk8orEHLy4ooBqEoTEiYqhBlr6ppbtFcyQAHYp3PzdpxgHIA===)

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
Wraith2: The lock isn't unnecessary. I dug into it at some point in the past and it's there to prevent overlapping send and receive  operations if you take it out you can observe strange behaviour under load. However, since locks don't work in async calls it isn't of benefit in those situations. In reality you should never see contention on those locks because nothing above this layer should ever try multiple operations (apart from sending out of band packets) and those have locks higher up to guard them. Ultimately they were a way to fix the sync version of the problem you're working on for async i think.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlUtil.cs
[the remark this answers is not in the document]
Wraith2: If you pass in `_queue` as state you avoid the async closure allocation.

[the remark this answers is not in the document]
DavoudEshtehari: I'm curious about the correctness of the order here.
Assume a condition that there are 5 items in the queue that are waiting until the semaphore receives a signal; if the 5th `semaphore` receives the signal before others, then it will dequeue the first `tcs` and set its value and will return the 5th `tcs.Task` while the first item is still waiting to receive a signal to unblock.
Any thoughts?

@roji , @Wraith2

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNINpHandle.cs
```diff
@@ -271,40 +268,43 @@ public override uint Send(SNIPacket packet)
...
-                    // is the packet is marked out out-of-band (attention packets only) it must be
-                    // sent immediately even if a send of recieve operation is already in progress
-                    // because out of band packets are used to cancel ongoing operations
-                    // so try to take the lock if possible but continue even if it can't be taken
-                    if (packet.IsOutOfBand)
-                    {
-                        Monitor.TryEnter(this, ref releaseLock);
-                    }
-                    else
+                    lock (this)
```
Wraith2, line 271: This is going to break the code below which selectively takes the lock on `this` and regress the cancelation behaviour that it fixed.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
```diff
@@ -223,10 +223,10 @@ public SNITCPHandle(string serverName, int port, long timerExpire, object callba
                 }
 
                 _socket.NoDelay = true;
-                _tcpStream = new NetworkStream(_socket, true);
+                _tcpStream = new SNINetworkStream(_socket, true);
```
Wraith2, line 226: If we're going to have our own versions of NetworkStream and SslStream can we combine some of them into a single class along with SslOverTdsStream?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
cheenamalhotra: This one is on SendAsync method, sync method locks are intact.

Wraith2: Good to go then.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlUtil.cs
```diff
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
+using System.Collections.Concurrent;
```
karinazhou, line 6: This is unnecessary after removing the ConcurrentQueue

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNINpHandle.cs
cheenamalhotra: That change was unintentional, reverted.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlUtil.cs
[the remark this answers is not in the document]
DavoudEshtehari: I'm not sure if the conversation is going forward on the same page. If I'm wrong please skip this comment.

Semaphore doesn't support the order and in this case, we need to guarantee the order besides supporting the concurrency: Please, follow the last line in the remark section in [MS documents SemaphoreSlim](https://docs.microsoft.com/en-us/dotnet/api/system.threading.semaphoreslim?view=net-5.0#remarks).

Please, look at the conversation with Shay [here](https://github.com/dotnet/SqlClient/pull/579#discussion_r493018970).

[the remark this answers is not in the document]
cheenamalhotra: I think the initial design I proposed when calling ContinueWith on WaitAsync was going to preserve the order, as it would continue the read/write operation on the same task from queue without losing handle. With async/await it would look something like this:

```cs
public override async Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
{
    try
    {
        return await await _readAsyncSemaphore.WaitAsync().ContinueWith(t => base.ReadAsync(buffer, offset, count, cancellationToken));
    }
    finally
    {
        _readAsyncSemaphore.Release();
    }
}
```

Let me know what you think.
[Tests are passing]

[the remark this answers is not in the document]
Wraith2: It is correct as far as I can see. the use of queue popping ensures that the waits are released in the same order that they arrived in the queue. The semaphore just prevents concurrency.

I'm not aware of any reason that multiple reads would ever be issued at the same time legitimately though. Sending you can get our of band packets but you wouldn't see that as out of band on the receive side it's just another packet. The only thing that should be causing a read is asking for another packet and asking for a read while one is in progress had to be an error in that case doesn't it? So is there a repro where this queue is ever used?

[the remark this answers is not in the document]
cheenamalhotra: @roji Right, I agree the continue is not necessary.. it does the same thing.

> I'm not aware of any reason that multiple reads would ever be issued at the same time legitimately though. 

@Wraith2 yes, since we acquire only 1 semaphore handle - we're good!

> The only thing that should be causing a read is asking for another packet and asking for a read while one is in progress had to be an error in that case doesn't it? So is there a repro where this queue is ever used?

I have a handy code to study how queue executes tasks, out of driver.. in case you're interested: https://gist.github.com/cheenamalhotra/d6d4140b35536064db14d744d4d700c1

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlUtil.cs
cheenamalhotra: We've added it back.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlUtil.cs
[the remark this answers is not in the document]
Wraith2: Remember to sprinkle in those `ConfigureAwaits()` 😁 

I wasn't questioning the functionality of this class, it looks ok to me. I just think it shouldn't be needed because the situation it's guarding against is an error in the first place. Only the parser should cause reads and it's entire purpose is state maintenance so any case where multiple concurrent reads are issued is an error.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
cheenamalhotra: Yes, there's room for improvement.
We can work on modifying design in future PRs, not trivial for this fix to go in for coming release.

Wraith2: Sounds sensible.

DavoudEshtehari (approved):

PR #800, opened 2020-11-17
Introduce new API to clear global Public Client Application instances holding User Token Cache

cheenamalhotra:
New API added as under to clear globally held Public Client Application instances in order to reduce prompts for Access token acquisitions from MSAL.NET.

```cs
public sealed class ActiveDirectoryAuthenticationProvider : SqlAuthenticationProvider
{
    public static void ClearUserTokenCache();
}
```

### DIFF
--- a/doc/snippets/Microsoft.Data.SqlClient/ActiveDirectoryAuthenticationProvider.xml
+++ b/doc/snippets/Microsoft.Data.SqlClient/ActiveDirectoryAuthenticationProvider.xml
@@ -40,6 +40,9 @@
       <summary>Acquires a security token from the authority.</summary>
       <returns>Represents an asynchronous operation that returns the authentication token.</returns>
     </AcquireTokenAsync>
+    <ClearUserTokenCache>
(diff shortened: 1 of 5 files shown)


doc/snippets/Microsoft.Data.SqlClient/ActiveDirectoryAuthenticationProvider.xml
```diff
@@ -40,6 +40,9 @@
       <summary>Acquires a security token from the authority.</summary>
       <returns>Represents an asynchronous operation that returns the authentication token.</returns>
     </AcquireTokenAsync>
+    <ClearUserTokenCache>
+      <summary>Clears cached instances of Public Client Application maintained by driver to access cached tokens from underlying token provider library.</summary>
```
David-Engel, line 44: I think we can keep the description simple:

```suggestion
      <summary>Clears cached user tokens from the token provider.</summary>
      <remarks>This will cause interactive authentication prompts to appear again if tokens were previously being obtained from the cache.</remarks>
```

karinazhou (approved):

PR #802, opened 2020-11-18
Some API Review Changes

Xtrimmer:
After doing an 1:1 API review with Saurabh, I have made the following changes:

1. Change the `IColumnarDataReader` and `IColumnarDataWriter`'s `Read` and `Write` methods to return and consume `IEnumerable<IColumn>` rather than `IList<IColumn>`
2. Remove the `RegisterKeyStoreProviders` method form the file reader and provide the set of providers via the constructor.
3. Write metadata to parquet files only once. It was writing the same info with every iteration.
4. Change the `Demo` project to get credentials from the `CommonTestUtiliy` project's config and not from environment variables.
5. Upgrade to a stable `Microsoft.NET.Test.Sdk` to Version 16.8.0


johnnypham (approved):
Conflicts need resolving. 4 is not necessary as the demo .cs files have been converted to standalone sample programs (in doc/sample/MicrosoftDataEncryption).

PR #827, opened 2020-11-29
Share more files and add MultiPartIdentifier tests

Wraith2:
Merged a couple of files into the shared location, MultipartIdentifier and DBConnectionPoolKey. Datastorage in netfx was not included in the project and was not used so i removed it. I've also added some tests for the MultipartIdentifier by including the file in the functional tests project and giving it some local context (ADP methods it calls) in the test class.


ErikEJ:
Wonder if it would be good to add `StringComparison.Ordinal`  with IndexOf (but maybe that is a general issue in the entire codebase) - see https://github.com/dotnet/runtime/issues/43736

Wraith2:
Perhaps. It's above my level to make that decision. It'll only affect .Net5 at the moment but moving to 5 is a good idea for sql perf not least because of https://github.com/dotnet/runtime/pull/1637
I've got a complete replacement for MultipartIdentifier which works in terms of spans which is why I have the test coverage, I'll see if there's anywhere in there that IndexOf is used.

Wraith2:
@David-Engel @cheenamalhotra any chance of getting this and the other couple of recent merge PR's reviewed and merged? There's more to do but having too many PR's open makes it confusing to work on.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft.Data.SqlClient.csproj
```diff
@@ -355,18 +361,26 @@
...
     <Compile Include="Microsoft\Data\SqlClient\SqlConnectionPoolGroupProviderInfo.cs" />
     <Compile Include="Microsoft\Data\SqlClient\SqlConnectionPoolKey.cs" />
     <Compile Include="Microsoft\Data\SqlClient\SqlConnectionString.cs" />
     <Compile Include="Microsoft\Data\SqlClient\SqlConnectionStringBuilder.cs" />
     <Compile Include="Microsoft\Data\SqlClient\SqlConnectionTimeoutErrorInternal.cs" />
     <Compile Include="Microsoft\Data\SqlClient\SqlCredential.cs" />
-    <Compile Include="Microsoft\Data\SqlClient\SqlDataAdapter.cs" />
+    <Compile Include="Microsoft\Data\SqlClient\SqlDataAdapter.cs">
+      <SubType>Component</SubType>
+    </Compile>
```
David-Engel, line 383: Please revert these lines. Not that it hurts the build process. It's just annoying that VS wants to treat these classes as components with the resulting VS behavior of trying to load them in the designer (broken) instead of opening straight to the code.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft.Data.SqlClient.csproj
```diff
@@ -432,7 +444,9 @@
     <Compile Include="Microsoft\Data\ProviderBase\DbBuffer.cs" />
     <Compile Include="Microsoft\Data\ProviderBase\DbConnectionClosed.cs" />
     <Compile Include="Microsoft\Data\ProviderBase\DbConnectionFactory.cs" />
-    <Compile Include="Microsoft\Data\ProviderBase\SqlConnectionHelper.cs" />
+    <Compile Include="Microsoft\Data\ProviderBase\SqlConnectionHelper.cs">
+      <SubType>Component</SubType>
+    </Compile>
```
David-Engel, line 449: Ditto as above.

David-Engel (commented):
Looks good to me apart from the csproj issues I noted.

David-Engel:
> any chance of getting this and the other couple of recent merge PR's reviewed and merged? There's more to do but having too many PR's open makes it confusing to work on.

@Wraith2 Sorry for the delay. Holidays + vacations has made for less time to get things done. Thanks for the contributions! I'll try and look at the others ASAP.

Wraith2:
Netfx subtype project changes reverted.

David-Engel (approved):
LGTM. Thanks!

src/Microsoft.Data.SqlClient/tests/FunctionalTests/Microsoft.Data.SqlClient.Tests.csproj
```diff
@@ -55,6 +56,7 @@
     <Compile Include="SerializeSqlTypesTest.cs" />
     <Compile Include="TestTdsServer.cs" />
     <Compile Include="SqlHelperTest.cs" />
+    <Compile Include="..\..\src\Microsoft\Data\Common\MultipartIdentifier.cs" />
```
cheenamalhotra, line 59: I understand why you did this to be able to access this source code here, but this approach will not give code coverage to netcore\M.D.S project csproj. This class will be executed from within this project.

You could instead provide access to test namespace by adding assembly tag:
`[assembly:InternalsVisibleTo("")]`

You'd not need ADB overrides too.

Wraith2: In general I avoid InternalsVisibleTo because it's deprecated in netcore and wasn't a good idea to begin with. While it doesn't give code coverage it doesn't decrease code coverage either and it does increase test coverage so that the behaviour, which is most important, is now tested and deviations can be caught.

cheenamalhotra: Is there an official doc on it's deprecated status or why it's not a good idea? I mean there should be a better way of referencing internals if this is going to be deprecated! I don't think we need to limit ourselves.

If not, I'd rather increase code coverage than use this approach.

Wraith2: Good question. I know there are many placed in corefx/runtime where it's been removed and in general they don't seem to want to use it but i don't have concrete information about why, some mention of build perf and trimming problems. I've asked on the discord if there's an official opinion on it's use.

There is the obvious negative that it adds the name of the test assemblies irrevocably into the released binary which is unneeded and since there is no strong name capability on core anything that wants to can see those names and pretend to be that assembly allowing them to screw with internals. 

I can understand why you'd want to increase coverage but contributors still can't see any of those coverage numbers and as I said I've increased test coverage., Requiring me to hit a target I can't evaluate seems unreasonable.

cheenamalhotra: It's actually a work in progress to be able to report coverage on PRs as they come along, there's just so much activity on higher priority. But we regularly run code coverage internally ([link here](https://dev.azure.com/sqlclientdrivers-ci/sqlclient/_build?definitionScope=%5CCoco)). I was saying in reference to new tests that we are adding, they should contribute to code coverage, and doesn't sound like we have large concerns over using the assembly tag.

> it adds the name of the test assemblies irrevocably into the released binary 
We already have a few assembly tags in other classes for internal testing, so it's not going to make additional difference.

Wraith2: I've had a look and there are no test related InternalsVisibleTo attributes in the library. I also tried to look at the code coverage from your link but all I can find is an overall percentage and a .coverage file which I can't open.

Wraith2:
Any chance this can get merged and then when code coverage support is ready the tests can be changed around to use internals visibility if needed? I'd really like to get my open PR's in before the holiday so I've got a clean start on some of the more complex files.

cheenamalhotra (approved):
Everything else looks good.
Todo: Improve internal class exposure without duplicating reference.

Wraith2:
Excellent, thanks 😁

src/Microsoft.Data.SqlClient/tests/FunctionalTests/Microsoft.Data.SqlClient.Tests.csproj
cheenamalhotra: Sorry my bad. I think the alternate approach is using Reflection to access internal classes if not this way.
You're right, we should not use InternalsVisibleTo, as it causes assembly signing issues too.

PR #835, opened 2020-12-03
Update similar files, Share more files

Wraith2:
SqlMetaData has been reformatted to conform to the naming rules for the repo and moved to shared. It's a lot of diff but functionally it's the same code it just looks better and smaller now.

SqlSer and ValueUtilsSmi have had their trivial differences synced but the important differences in functionality have been left in each location. This should make it easier to make an informed decision about whether those differences that remain are important.

The other files are identical and have been moved to shared.


David-Engel (commented):
I got through everything except SqlMetadata.cs today. I'll try to spend some time on this PR tomorrow, too.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/Server/sqlser.cs
```diff
@@ -273,54 +258,21 @@ private void DontDoIt()
...
-            get
-            {
-                return this.m_size;
-            }
-            set
-            {
-                this.m_size = value;
-            }
+            get => _size;
+            set =>_size = value;
```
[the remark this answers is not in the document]
Wraith2: Fixed.

David-Engel (commented):
Much cleaner than the old code. Thanks!

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SqlMetaData.cs
```diff
@@ -472,27 +621,28 @@ internal string UdtTypeName
...
+            {
                 throw SQL.InvalidSqlDbTypeForConstructor(dbType);
+            }
 
             SetDefaultsForType(dbType);
 
             if (SqlDbType.NText == dbType || SqlDbType.Text == dbType)
             {
-                _lLocale = CultureInfo.CurrentCulture.LCID;
+                _locale = System.Globalization.CultureInfo.CurrentCulture.LCID;
```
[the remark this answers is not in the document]
Wraith2: Fixed. Oddly visual studio didn't flag that as a possible issue even though it was greyed out as unnecessary.

David-Engel (approved):
Looks good to me. Thanks!

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/Server/SqlNorm.cs
```diff
@@ -16,32 +16,34 @@ namespace Microsoft.Data.SqlClient.Server
...
         {
-            FieldInfo = fi;
-            Offset = offset;
             Debug.Assert(normalizer != null, "normalizer argument should not be null!");
             Normalizer = normalizer;
+            FieldInfo = fi;
+            _offset = offset;
         }
+        public FieldInfo FieldInfo { get; private set; }
+        public Normalizer Normalizer { get; private set; }
```
cheenamalhotra, line 29: Any special reason to make these props public?
They were internal in both projects.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/Server/sqlser.cs
```diff
@@ -340,7 +292,7 @@ public override int Read(byte[] buffer, int offset, int count)
 
         public override void Write(byte[] buffer, int offset, int count)
         {
-            this.m_size += count;
+            _size += count;
```
cheenamalhotra, line 295: Could use lambda here too.

Wraith2: Changed. There are other places where it can be used but eventually looking through these files you get snowblind for some things. When the files are merged it'll get another pass so more can be cleaned up.

It will also depend on the style policy you want to apply for this repo. There are many versions of language constructs we can possibly use like switch expressions, expression bodies, static lambdas, type switching. And then the level of clarity we feel each provides. I'm being conservative at the moment.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/Server/SqlNorm.cs
Wraith2: Personal preference. I've changed it. 

Those members are accessed by users of the class and in those cases I prefer to use public indicating that they're part of the public contract for the type. In general my preference is to mark types as internal and members of those types as public private protected as you would a publicly visible class. I'll follow the convention for the project as always though.

cheenamalhotra (approved):
LGTM

PR #838, opened 2020-12-03
Update similar files

Wraith2:
Small changes on both sides to sync trivial changes between them and apply code formatting equally. These minor changes allow visualisation of real differences and do small cleanups.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlBuffer.cs
+++ b/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlBuffer.cs
@@ -11,10 +11,8 @@ using Microsoft.Data.SqlTypes;
 
 namespace Microsoft.Data.SqlClient
 {
-
-    internal sealed class SqlBuffer
+    internal sealed partial class SqlBuffer
     {
-
         internal enum StorageType
         {
             Empty = 0,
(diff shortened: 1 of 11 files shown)


src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlBuffer.cs
```diff
@@ -856,7 +855,7 @@ internal object Value
                     case StorageType.Empty:
                         return DBNull.Value;
                     case StorageType.Boolean:
-                        return Boolean;
+                        return Boolean ? s_cachedTrueObject : s_cachedFalseObject;
```
David-Engel, line 858: Comment for other reviewers: This change had me wondering if there might be hidden consequences here. But I see this is what the code is in netcore. So I assume it's been hashed out and tested.

David-Engel (approved):
LGTM. Thanks!

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlBuffer.cs
Wraith2: It's a change in added in corefx. Returning a literal boolean each time caused the value to be boxed into the object return variable location so if you were reading bool fields you got a lot of gen0 garbage. This removed the garbage by returning a pre-boxed boolean, essentially caching the value.

Wraith2: A thought, Is it worth putting a comment in the code explaining this? Corefx tended to be very sceptical of comments but I think it's helpful to explain non-obvious things like this.

cheenamalhotra: Would appreciate an inline comment over the 2 private cached instances. Thanks for explanation!

cheenamalhotra (approved):

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlBuffer.cs
Wraith2: Done.

PR #867, opened 2021-01-07
[.NET Core] Event source traces revision - Part 1 & 2

cheenamalhotra:
The PR standardizes event source traces with additional info added where applicable like Client Connection Id, SPID, etc. as they're extremely helpful while debugging. I'll be working on all other classes (most importantly .NET Core) including Managed SNI later once this gets merged to avoid conflicts.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
@@ -103,7 +103,7 @@ namespace Microsoft.Data.SqlClient
         /// Internal flag for testing purposes that forces all queries to internally end async calls.
         /// </summary>
         private static bool _forceInternalEndQuery = false;
-#endif 
+#endif
 
         private static readonly SqlDiagnosticListener _diagnosticListener = new SqlDiagnosticListener(SqlClientDiagnosticListenerExtensions.DiagnosticListenerName);
         private bool _parentOperationStarted = false;
@@ -172,7 +172,7 @@ namespace Microsoft.Data.SqlClient
 
         /// <summary>
         /// Return if column encryption setting is enabled.
-        /// The order in the below if is important since _activeConnection.Parser can throw if the 
+        /// The order in the below if is important since _activeConnection.Parser can throw if the
         /// underlying tds connection is closed and we don't want to change the behavior for folks
         /// not trying to use transparent parameter encryption i.e. who don't use (SqlCommandColumnEncryptionSetting.Enabled or _activeConnection.IsColumnEncryptionSettingEnabled) here.
         /// </summary>
@@ -312,7 +312,7 @@ namespace Microsoft.Data.SqlClient
 
         // Volatile bool used to synchronize with cancel thread the state change of an executing
         // command going from pre-processing to obtaining a stateObject.  The cancel synchronization
-        // we require in the command is only from entering an Execute* API to obtaining a 
+        // we require in the command is only from entering an Execute* API to obtaining a
         // stateObj.  Once a stateObj is successfully obtained, cancel synchronization is handled
         // by the stateObject.
         private volatile bool _pendingCancel;
@@ -438,7 +438,7 @@ namespace Microsoft.Data.SqlClient
                         }
                         catch (Exception)
                         {
-                            // we do not really care about errors in unprepare (may be the old connection went bad)                                        
+                            // we do not really care about errors in unprepare (may be the old connection went bad)
                         }
                         finally
                         {
@@ -449,7 +449,7 @@ namespace Microsoft.Data.SqlClient
                     }
                 }
                 _activeConnection = value;
-                SqlClientEventSource.Log.TryTraceEvent("<sc.SqlCommand.set_Connection|API> {0}, {1}", ObjectID, value?.ObjectID);
+                SqlClientEventSource.Log.TryTraceEvent("SqlCommand.Set_Connection | API | ObjectId {0}, Client Connection Id {1}, SPID {2}", ObjectID, value?.ClientConnectionId, value?.ServerProcessId);
             }
         }
 
@@ -483,9 +483,9 @@ namespace Microsoft.Data.SqlClient
             }
             set
             {
-                SqlClientEventSource.Log.TryTraceEvent("<sc.SqlCommand.set_Notification|API> {0}", ObjectID);
                 _sqlDep = null;
                 _notification = value;
+                SqlClientEventSource.Log.TryTraceEvent("SqlCommand.Set_Notification | API | Object Id {0}", ObjectID);
             }
         }
 
@@ -527,8 +527,8 @@ namespace Microsoft.Data.SqlClient
                         throw SQL.CannotModifyPropertyAsyncOperationInProgress();
                     }
                 }
-                SqlClientEventSource.Log.TryTraceEvent("<sc.SqlCommand.set_Transaction|API> {0}", ObjectID);
                 _transaction = value;
+                SqlClientEventSource.Log.TryTraceEvent("SqlCommand.Set_Transaction | API | Object Id {0}, Internal Transaction Id {1}, Client Connection Id {2}, SPID {3}", ObjectID, value?.InternalTransaction?.TransactionId, Connection?.ClientConnectionId, Connection?.ServerProcessId);
             }
         }
 
@@ -542,6 +542,7 @@ namespace Microsoft.Data.SqlClient
             set
             {
                 Transaction = (SqlTransaction)value;
+                SqlClientEventSource.Log.TryTraceEvent("SqlCommand.Set_DbTransaction | API | Object Id {0}, Client Connection Id {1}, SPID {2}", ObjectID, Connection?.ClientConnectionId, Connection?.ServerProcessId);
             }
         }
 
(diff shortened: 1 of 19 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnection.cs
```diff
@@ -464,31 +463,31 @@ private void EnlistNonNull(Transaction tx)
                 PropagateTransactionCookie(cookie);
 
                 _isEnlistedInTransaction = true;
-                SqlClientEventSource.Log.TryAdvancedTraceEvent("<sc.SqlInternalConnection.EnlistNonNull|ADV> {0}, enlisted with transaction {1} with transactionId=0x{2}", ObjectID, null != CurrentTransaction ? CurrentTransaction.ObjectID : 0, null != CurrentTransaction ? CurrentTransaction.TransactionId : SqlInternalTransaction.NullTransactionId);
+                SqlClientEventSource.Log.TryAdvancedTraceEvent("SqlInternalConnection.EnlistNonNull | ADV | Object Id {0}, Client Connection Id {1}, Enlisted in transaction with transactionId=0x{2}", ObjectID, Connection?.ClientConnectionId, tx?.TransactionInformation?.LocalIdentifier);
```
karinazhou, line 466: You can remove the 0x to keep consistency.

cheenamalhotra: Sure.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlTransaction.cs
```diff
@@ -141,8 +141,8 @@ override public void Commit()
             ZombieCheck();
 
             SqlStatistics statistics = null;
-            long scopeID = SqlClientEventSource.Log.TryScopeEnterEvent("<sc.SqlTransaction.Commit|API> {0}", ObjectID);
-            SqlClientEventSource.Log.TryCorrelationTraceEvent("<sc.SqlTransaction.Commit|API|Correlation> ObjectID {0}, ActivityID {1}", ObjectID, ActivityCorrelator.Current);
+            long scopeID = SqlClientEventSource.Log.TryScopeEnterEvent("SqlTransaction.Commit | API | Object Id {0}", ObjectID);
+            SqlClientEventSource.Log.TryCorrelationTraceEvent("SqlTransaction.Commit | API | Correlation | Object Id {0}, Activity Id {1}, Client Connection Id {2}, SPID {3}", ObjectID, ActivityCorrelator.Current, Connection?.ClientConnectionId, Connection?.ServerProcessId);
```
karinazhou, line 145: For the null Connection case, it will be empty in the log message. Is it expected or can we put some "null" string instead? Same to all the other log messages which may contain null.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlTransaction.cs
```diff
@@ -255,7 +255,7 @@ public void Rollback(string transactionName)
             Guid operationId = s_diagnosticListener.WriteTransactionRollbackBefore(_isolationLevel, _connection, InternalTransaction, transactionName);
 
             ZombieCheck();
-            long scopeID = SqlClientEventSource.Log.TryScopeEnterEvent("<sc.SqlTransaction.Rollback|API> {0} transactionName='{1}'", ObjectID, transactionName);
+            long scopeID = SqlClientEventSource.Log.TryScopeEnterEvent("SqlTransaction.Rollback | API | Object Id {0}, Transaction Name='{1}', ActivityID {2}, Client Connection Id {3}", ObjectID, transactionName, ActivityCorrelator.Current, Connection.ClientConnectionId);
```
karinazhou, line 258: Missing null check for Connection?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlTransaction.cs
karinazhou: NVM. After a 2nd look, we have already handle null object when calling TryXXXEvent methods.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
```diff
@@ -4597,7 +4601,7 @@ private SqlDataReader RunExecuteReaderTds(CommandBehavior cmdBehavior, RunBehavi
                     optionSettings = GetSetOptionsString(cmdBehavior);
                     if (returnStream)
                     {
-                        SqlClientEventSource.Log.TryTraceEvent("<sc.SqlCommand.ExecuteReader|INFO> {0}, Command executed as RPC.", ObjectID);
+                        SqlClientEventSource.Log.TryTraceEvent("SqlCommand.RunExecuteReaderTds | Info | Object Id {0}, Activity Id {1}, Client Connection Id {2}, SPID {3}, Command executed as RPC, RPC Name '{4}' ", ObjectID, ActivityCorrelator.Current, Connection?.ClientConnectionId, Connection?.ServerProcessId, rpc.rpcName);
```
karinazhou, line 4604: Will rpc be null here?

Wraith2: Accessing `ActivityCorrelator.Current` can cause an allocation and pinvoke, can it be avoided unless the trace message is actually going to be printed?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnection.cs
```diff
@@ -942,7 +942,7 @@ public SqlTransaction BeginTransaction(string transactionName)
         [SuppressMessage("Microsoft.Reliability", "CA2004:RemoveCallsToGCKeepAlive")]
         override protected DbTransaction BeginDbTransaction(System.Data.IsolationLevel isolationLevel)
         {
-            long scopeID = SqlClientEventSource.Log.TryScopeEnterEvent("<prov.SqlConnection.BeginDbTransaction|API> {0}, isolationLevel={1}", ObjectID, (int)isolationLevel);
+            long scopeID = SqlClientEventSource.Log.TryScopeEnterEvent("SqlConnection.BeginDbTransaction | API | Object Id {0}, Isolation Level {1}", ObjectID, (int)isolationLevel);
```
JRahnama, line 945: "<prov" is removed. is this intentional?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
cheenamalhotra: It's value is static, but the `ToString()` method getting called indeed concatenates string. I'll try to fetch Id and sequence instead of formatted string, but we need current activity Id in logs.

cheenamalhotra: > Will rpc be null here?

I'll add a check never-the-less.

Wraith2: > It's value is static, 

It's thread static and lazy allocated which means the first touch on a thread will create it which is the allocation and pinvoke for guid that I referred to. It's not a huge cost but if it's just going to be ignored I'd rather not have the allocation happen.

In this case I can't see a good way to keep the capture semantic and avoid the allocation unless it's needed. I'll make a mental note to see if there's something I can do in future but don't worry about it here.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnection.cs
cheenamalhotra: Yes

karinazhou (approved):

JRahnama (approved):

JRahnama:
LGTM, it looks better now.

Wraith2:
@cheenamalhotra is there a part 3? i'm working through a diff of some files and it's struck me that this PR has changed netcore but not netfx

cheenamalhotra:
Yes, I'm working on Managed SNI and TDS Parser layers (PR on way..)
NetFx is low priority for these improvements, as we need these changes to debug and traces issues in Managed SNI first.

Wraith2:
Ok. I'll put an item on my ever-expanding list of things to do.

I'm working to try and get as much shared or synced between netfx and netcore so these new changes stand out. Doing a single file at a time would leave the netfx tracing in a bad state so it'll need to be done in a single PR at a later date.

cheenamalhotra:
Ok, thanks for keeping up with that! Please also keep your branches up-to-date and resolve conflicts in case you encounter (hopefully not many, since they're just trace changes).

At this point, our primary focus is on solving #659 and #422, and these improvements are helpful in investigations.
We will definitely make them uniform with NetFx in future 👍

Wraith2:
I was going to have a look at #659 over the holiday but was diverted. My thought was to start in the mars connection demultiplexing code and make sure that the lookup was locked/concurrent safe because that was the only way I could think it would be possible to get a packet to the wrong virtual connection.

As always, I'll try and avoid conflicts.

cheenamalhotra:
That would be nice!

Please check my last comment on #659 maybe it rings a bell to you. it doesn't seem a concurrency issue, it's more due to uncleared information. To cut short, data from first connection is arriving in the "receivedPacketQueue" even after we closed the first connection (sent it to pool) since the physical connection is still active which is dequeued in second connection's query. There should be a cleanup/disposal happening before putting connection to pool (so we don't listen to more packets) or after connection is reactivated (cleanup everything) [I'm still trying to establish this part]. We do perform fResetConnection, so that's not related, because when RESET happens the queue already contains data on the client side.

We should be cleaning up MarsHandle or packets which are still connected to StateObj on the connection in pooling.
I also always hit the Debug assert in SNIPacket's destructor, and we have a lot of unreleased packets, maybe it's related?

For #422, I have some ideas to try, but we'll come to that after this more serious issue.

PR #870, opened 2021-01-07
Fix | Fix issue connecting with instance name from Unix environment

### LINKED ISSUE #868: The instance name is ignored when running on Linux
### Describe the bug
When connecting to a non-default instance from a Linux client, Microsoft.Data.SqlClient 2.1 and later ignores the instance name and instead connects to the default instance. Obviously this is a problem as it raises the possibility of executing code on an instance that you didn't know you were connected to.

### To reproduce
Setup a Windows server with two SQL Server instances. Let's call them `server\defaultinstance` and `server\nondefaultinstance`. I have SQL Server Profiler open with two traces running - one on each instance.

Using `dotnet new` create a console application using the following code and reference `Microsoft.Data.SqlClient` 2.1.1. 

```c#
using System;
using System.IO;

namespace test
{
    class Program
    {
        {
            var commandText = @"SELECT 1";

            var connectionString = new SqlConnectionStringBuilder() {
                InitialCatalog = "master",
                DataSource = @"server\nondefaultinstance",
                UserID = "sa",
                Password = "redacted",
                IntegratedSecurity = false
            };

            try {
                Console.WriteLine($"Executing SQL - {commandText}");

                using (var connection = new SqlConnection(connectionString.ToString()))
                {
                    connection.Open();
                    var command = new SqlCommand(commandText, connection);
                    command.ExecuteNonQuery();
                }
            }
            catch (Exception e)
            {
                Console.WriteLine("Caught an exception");
                Console.WriteLine(e.Message);
            }
            finally 
            {
                Console.WriteLine("End of program");
            }
        }
    }
}
```

I'm using an Ubuntu terminal in WSL2. In the terminal, change to the project directory and run `dotnet run`.

### Expected behavior
To see a batch with `SELECT 1` appear in the trace for `server\nondefaultinstance`

### Actual behavior
A batch with `SELECT 1` appears in the trace for `server\defaultinstance`

### Further technical details
Microsoft.Data.SqlClient version: 2.1.1 (but 2.1.0 is also affected)
.NET target: 5.0
SQL Server version: Have tested on 2017 and 2019
Operating system: Server 2019 for running the two SQL Server instances, Ubuntu 20.04 for running the application

cheenamalhotra:
Fixes #868

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIProxy.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIProxy.cs
@@ -685,7 +685,7 @@ namespace Microsoft.Data.SqlClient.SNI
 
             int commaIndex = _dataSourceAfterTrimmingProtocol.IndexOf(CommaSeparator);
 
-            int backSlashIndex = _dataSourceAfterTrimmingProtocol.IndexOf(PipeBeginning);
+            int backSlashIndex = _dataSourceAfterTrimmingProtocol.IndexOf(BackSlashCharacter);
 
             // Check the parameters. The parameters are Comma separated in the Data Source. The parameter we really care about is the port
             // If Comma exists, the try to get the port number


karinazhou (approved):

DavoudEshtehari (approved):

JRahnama (approved):

Crono1981:
Just to add on this: if the server only has named instances, then any attempt at opening a connection will throw an exception.

My understanding is that this happens because in resolving "address\instance", the targeted data source becomes in fact "address", which would then point to a non-existing Sql Server. The fix should therefore cover this case too.

Is my assumption correct?

cheenamalhotra:
Yes, that's correct!
We're also backporting this to v2.1 (PR #874) for next patch release.

Crono1981:
Thanks for the blazing fast reply, @cheenamalhotra ! I am happy to hear I will not have to wait until v3 for this fix. 😁

Keep up the good work!

PR #871, opened 2021-01-09
Merge some netcore and netfx files

Wraith2:
SqlCollection and SqlParameter were defined over two files in both solutions and some of the files were in different places. They are now defined in a single file each. SqlParameterCollection is now shared between both implementations. SqlParameter is not shared because there are differences; is sql 2005 supported in this library? Some other files have been changed to support the major files being merged.

The majority of changes are SqlParameter and looking at those diffs is very difficult. I suggest pulling the changed files locally and reviewing the files themselves.


src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlCommand.cs
```diff
@@ -730,7 +730,7 @@ override public string CommandText
             {
                 SqlClientEventSource.Log.TryTraceEvent("<sc.SqlCommand.set_CommandText|API> {0}, String Value = '{1}'", ObjectID, value);
 
-                if (0 != ADP.SrcCompare(_commandText, value))
+                if (_commandText != value)
```
ErikEJ, line 733: This seems unrelated?

Wraith2: Sort of. The intent was to clean up ADP as well but the PR was getting too large and complicated so I stopped part way. This change on it's own isn't harmful though so I left it in. 

Functions like SrcCompare and DestCompare which just do string comparisons don't really have any benefit to being functions instead of inline, they're trivial and not worth the extra effort required to investigate what they're doing.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlParameterCollection.cs
```diff
@@ -1,140 +0,0 @@
...
-using System.Data;
-using System.Data.Common;
-
-namespace Microsoft.Data.SqlClient
-{
-    /// <include file='../../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlParameterCollection.xml' path='docs/members[@name="SqlParameterCollection"]/SqlParameterCollection/*' />
-    public sealed partial class SqlParameterCollection : DbParameterCollection
-    {
-        private bool _isDirty;
-        private static Type s_itemType = typeof(SqlParameter);
```
johnnypham, line 15: Why remove this and instead call `typeof(SqlParameter` every time?

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/Common/NameValuePair.cs
```diff
@@ -1,71 +0,0 @@
...
-        }
-
-        internal int Length
-        {
-            get
-            {
-                // this property won't exist when deserialized from Everett to Whidbey
-                // it shouldn't matter for DbConnectionString/DbDataPermission
-                // which should only use Length during construction
-                // not deserialization or post-ctor runtime
```
johnnypham, line 36: Why not bring this comment block over to the new file?

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/ProviderBase/FieldNameLookup.cs
```diff
@@ -1,159 +0,0 @@
...
-
-        public FieldNameLookup(System.Collections.ObjectModel.ReadOnlyCollection<string> columnNames, int defaultLocaleID)
-        {
-
-            int length = columnNames.Count;
-            string[] fieldNames = new string[length];
-            for (int i = 0; i < length; ++i)
-            {
-                fieldNames[i] = columnNames[i];
-                Debug.Assert(null != fieldNames[i], "MDAC 66681");
```
johnnypham, line 45: I saw these `Debug.Assert` lines were removed from the constructors. Any idea what they were for?

johnnypham (approved):

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlParameterCollection.cs
Wraith2: The JIT can optimize away the `typeof(SqlParameter)` because at the time the method is compiled it knows the exactly value of the type handle and can insert it. If we put that handle into a variable then the jit only sees it as a variable access and will generate all the code needed to load the variable. It seems counter intuitive from a c# perspective to repeatedly use typeof rather than refer to a constant but it's actually faster.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/Common/NameValuePair.cs
Wraith2: Everett to Whidbey are .net 1.1 and 2.0, neither is supported so the comment is no longer relevant.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/ProviderBase/FieldNameLookup.cs
Wraith2: They make sure that there aren't column names missing. They're there because of the way that the hidden column mapping used to work and since i changed that there isn't a way to get null column named in the map, it uses an integer crossmap which is lazy initialized.

Wraith2:
@cheenamalhotra any chance of getting this PR reviewed next? I've got a few things I'd like to be able to change in SqlParameter that it's blocking.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
```diff
@@ -339,7 +357,8 @@
       <Link>Microsoft\Data\SqlClient\VirtualSecureModeEnclaveProvider.cs</Link>
     </Compile>
     <Compile Include="Microsoft\Data\SqlClient\VirtualSecureModeEnclaveProviderBase.cs" />
-	</ItemGroup>
+    <Compile Include="Microsoft\Data\SqlClient\AlwaysEncryptedEnclaveProviderUtils.cs" />
```
cheenamalhotra, line 360:
```suggestion
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlParameter.cs
```diff
@@ -849,332 +707,265 @@ public override object Value
...
-                    }
+            }
+        }
 
-                    throw ADP.ParameterConversionFailed(value, destinationType.ClassType, e);
+
+        /// <include file='../../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlParameter.xml' path='docs/members[@name="SqlParameter"]/System.ICloneable.Clone/*' />
+        object ICloneable.Clone() => new SqlParameter(this);
+
+
```
David-Engel, line 855:
```suggestion

```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlParameter.cs
```diff
@@ -1537,86 +1361,386 @@ private void GetActualFieldsAndProperties(out List<SmiExtendedMetaData> fields,
...
+            else if (
+                (maxLen > SmiMetaData.MaxBinaryLength && (SqlDbType.Binary == mt.SqlDbType || SqlDbType.VarBinary == mt.SqlDbType)) ||
+                (maxLen > SmiMetaData.MaxANSICharacters && (SqlDbType.Char == mt.SqlDbType || SqlDbType.VarChar == mt.SqlDbType)) ||
+                (maxLen > SmiMetaData.MaxUnicodeCharacters && (SqlDbType.NChar == mt.SqlDbType || SqlDbType.NVarChar == mt.SqlDbType))
+            )
+            {
+                maxLen = -1;
+            }
+
+
```
David-Engel, line 1626:
```suggestion

```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlParameter.cs
```diff
@@ -1823,65 +1958,272 @@ internal MetaType ValidateTypeLengths()
...
+                        return cvalue.Length;
+                    case byte _:
+                    case char _:
+                        return 1;
+                }
+            }
+            return 0;
+        }
+
+
```
David-Engel, line 2082:
```suggestion

```

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlParameter.cs
```diff
@@ -938,339 +694,279 @@ override public object Value
...
-                {
-                    // UNDONE - should not be catching all exceptions!!!
-                    if (!ADP.IsCatchableExceptionType(e))
-                    {
-                        throw;
-                    }
+            }
+        }
+
+
```
David-Engel, line 838:
```suggestion

```

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlParameter.cs
```diff
@@ -1891,108 +1959,319 @@ internal MetaType ValidateTypeLengths(bool yukonOrNewer)
...
+                        return cvalue.Length;
+                    case byte _:
+                    case char _:
+                        return 1;
+                }
+            }
+            return 0;
+        }
+
+
```
David-Engel, line 2130:
```suggestion

```

David-Engel (approved):
Apart from the minor nit-pick issues, LGTM.

I know we have a lot of these types of refactors ahead of us. As a suggestion for future ones, it might be helpful to reviewers to break the refactor into multiple steps/commits. For example, keep any method/property re-ordering to a single commit without other changes. That way someone can review each commit to see a "no change" re-order commit, followed by a commit that has cleanup of code within the methods. This will make it much easier to identify potential differences that may introduce bugs and give us more confidence in approving them.

I do appreciate the work!

David-Engel:
> is sql 2005 supported in this library?

To answer your question, no. We only have to support versions that are within extended support (10 years). So SQL 2012 and up, at this point.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlParameter.cs
Wraith2: Done

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlParameter.cs
Wraith2: Done.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlParameter.cs
Wraith2: Done.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlParameter.cs
Wraith2: Done.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlParameter.cs
Wraith2: Done.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft.Data.SqlClient.csproj
```diff
@@ -417,18 +433,29 @@
     <Compile Include="Microsoft\Data\SqlClient\SqlColumnEncryptionCngProvider.cs" />
     <Compile Include="Microsoft\Data\SqlClient\SqlColumnEncryptionCspProvider.cs" />
     <Compile Include="Microsoft\Data\SqlClient\SqlColumnEncryptionEnclaveProvider.cs" />
-    <Compile Include="Microsoft\Data\SqlClient\SqlCommand.cs" />
-    <Compile Include="Microsoft\Data\SqlClient\SqlCommandBuilder.cs" />
+    <Compile Include="Microsoft\Data\SqlClient\SqlCommand.cs">
+      <SubType>Component</SubType>
+    </Compile>
```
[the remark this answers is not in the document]
Wraith2: Me too. The only say i've found to deal with them is to keep the project unloaded but that isn't helpful when trying to merge thing. They're cleared.

Wraith2:
> > is sql 2005 supported in this library?
> 
> To answer your question, no. We only have to support versions that are within extended support (10 years). So SQL 2012 and up, at this point.

In a future merge the 2005 code can be dropped then. 

> As a suggestion for future ones, it might be helpful to reviewers to break the refactor into multiple steps/commits. 

This one did end up being too large. I've tried to keep subsequent ones smaller. I'll bear in mind the phased approach in future.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft.Data.SqlClient.csproj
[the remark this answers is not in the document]
fowl2: untested drive-by... but: adding `[System.ComponentModel.DesignerCategory("")]` to the class seems to be the magic incantation :)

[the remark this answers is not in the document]
Wraith2: I'll give that a try, thanks.

PR #880, opened 2021-01-15
Test | Addressing Net5 runtime issues 

JRahnama:
While trying to test the driver against Net5, runtime caught some unobserved issues/warnings

1. Classes implementing IDisposable should dispose all related fields.
2. [Re-Throwing](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2200) caught exception from catch block will change the stack, hence should not be done.
3. [Calling overridable methods](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2214) from constructions are prohibited.
4. [Raising exception from unexpected places](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1065#:~:text=Dispose%20is%20often%20called%20as%20part%20of%20the,Dispose%20is%20almost%20always%20called%20from%20a%20finalizer.) such as dispose is not recommended.
5. [BinaryFormatter is not safe](https://docs.microsoft.com/en-us/dotnet/standard/serialization/binaryformatter-security-guide) and should not be used in all versions of .NETFramework and   netcoreapp 2.1 and upper versions.
6. [Warning CA1416: Platform compatibility](https://docs.microsoft.com/en-us/dotnet/core/compatibility/code-analysis/5.0/ca1416-platform-compatibility-analyzer) this issue came up for tests running on Windows for [CspParameters](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.cspparameters?view=net-5.0) class. In net5.0 and later versions [OperatindSystem.IsWindows
](https://docs.microsoft.com/en-us/dotnet/api/system.operatingsystem.iswindows?view=net-5.0) is introduced to get the operating system but the API is not known in lower versions.

In some cases such as overridable methods usage I have moved the method inside a concrete  method and called it from there and that made the driver to compile and run.  I am eager to hear all comments about the changes to improve them and make them better.

Thanks

Update:

I noticed some improvement opportunities in our test lab:

SqlNotificationTest was using `ManualResetEvent` I have changed them to use [ManualResetEventSlim](https://docs.microsoft.com/en-us/dotnet/api/system.threading.manualreseteventslim?view=net-5.0#remarks). Reasoning to do so:

1. ManualResetEventSlim is less expensive than ManualResetEvent.
2. ManualResetEventSlim is fully managed version of its predecessor.

there are small differences such as kernel related behavior and blocking the thread which may have better effect on our test lab.


src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/TestFixtures/SQLSetupStrategy.cs
```diff
@@ -277,7 +278,9 @@ public PlatformSpecificTestContext()
 
         public void Dispose()
         {
-            Fixture.Dispose();
+            akvFixture?.Dispose();
```
Wraith2, line 281: If any of these wrap system resources (cert store?) it might be worth putting a try-finally around each invoke so that even if one of them fail the others will still be tried. It doesn't matter for managed resources since they'll be freed when the process dies but for things like pinvokes that can be system-wide.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectivityTests/AADConnectionTest.cs
```diff
@@ -202,9 +202,11 @@ public static void testADPasswordAuthentication()
                     }
                 }
             }
-            catch (SqlException e)
+            catch (SqlException)
```
Wraith2, line 205: Why bother catching it if all we're going to do is rethrow it?

Wraith2 (approved):
Tiny suggestions otherwise LGTM.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectivityTests/AADConnectionTest.cs
JRahnama: Good point I will remove it.

cheenamalhotra (approved):

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlExceptionTest.cs
```diff
@@ -33,28 +33,6 @@ public void SerializationTest()
...
-                    stream.Position = 0;
-                    var e2 = (SqlException)formatter.Deserialize(stream);
-                }
-                catch (Exception ex)
-                {
-                    Assert.False(true, $"Unexpected Exception occurred: {ex.Message}");
-                }
-            }
-        }
-
```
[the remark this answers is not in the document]
JRahnama: This is something that we need to spend more time on. Supported serialization on NetFx is [DataContractSerializer](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.serialization.datacontractserializer?view=net-5.0) or [XMLSerializer](https://docs.microsoft.com/en-us/dotnet/api/system.xml.serialization.xmlserializer?view=net-5.0). Json is not in netfx. XMLSerializer only works on public members and cannot work on ICollections apparently. Our only option is DataContractSerializer and that requires breaking changes on the SqlException class, such as attributes. A class cannot have [Serializable] and [DataContract] at the same time. that is the main issue I hit while I was working on it. However, .NET 5 compalins about it in our test. it could be marked off for netcore and still get tested or since binaryFormatter is not supported removed from our test lab as well. Both will do the job for Net 5 test labs.

DavoudEshtehari (approved):

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlExceptionTest.cs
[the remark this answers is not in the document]
cheenamalhotra: > A class cannot have [Serializable] and [DataContract] at the same time. 

Does it mean existing applications doing as above will break?
Do we have no alternative? That doesn't sound right to me.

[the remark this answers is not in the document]
ErikEJ: >> Json is not in netfx.
Maybe I misunderstand but: 
https://docs.microsoft.com/en-us/dotnet/api/system.runtime.serialization.json.datacontractjsonserializer?view=net-5.0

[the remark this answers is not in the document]
JRahnama: @ErikEJ Thanks for mentioning this. My source for the changes was this [MS document](https://docs.microsoft.com/en-us/dotnet/standard/serialization/binaryformatter-security-guide#preferred-alternatives), However what you mentioned here looks interesting to me. I will spend some time to test how I can fit it to our problem here. For now, we have decided to keep the test, but disable it for .NET 5 and higher versions as the error was thrown by .NET 5 runtime.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlExceptionTest.cs
```diff
@@ -33,6 +33,29 @@ public void SerializationTest()
             Assert.Equal(e.StackTrace, sqlEx.StackTrace);
         }
 
+#if !NET5_0
```
Wraith2, line 36: This should probably either be handled in the csproj with a condition or use `[SkipOnTargetFramework]`, like the other tests

JRahnama: `NET5_0` is a defined constant of the NetCore directives. The preferred is  to not run on `NET 5` only and I believe there is no specific `[SkipOnTargetFramework]` to emphasis on version, please correct me if I am wrong. I have a plan to try and check if it could be added to XUnitExtensions in some free time intervals in future. Adding a constant in csproj is another option, but what would be the difference between adding a new constant to csproj or use predefined preprocessor value? Is there any benefit doing so?

All being said,  I don't see any issue with disabling the test on NetCore as we have other kind of serialization options there.

cheenamalhotra: We also want to ensure we don't end up running this on .NET 6.0 in future, so having something for ".NET 5 and above" is what we should include.

I see a point in using [SkipOnTargetFramework] to categorize .NET 5 and further to a totally new [Moniker](https://github.com/dotnet/SqlClient/blob/6a9e4f104ce77ba631e2ac8ea2e366a4d41ba235/src/Microsoft.Data.SqlClient/tests/tools/Microsoft.DotNet.XUnitExtensions/TargetFrameworkMonikers.cs) (e.g. Net=0x16)?

JRahnama: > I see a point in using [SkipOnTargetFramework] to categorize .NET 5 and further to a totally new Moniker (e.g. Net=0x16)? It does get complex later, since later we will also have new features/tests for new versions, and running / skipping them with moniker / target framework would be something we'll have to toss often.

I totally agree. This is on my radar to be added. Let me check if it could be a quick amendment to our XUnitExtensions . It will be needed as `.NET` is progressing to a different direction than `NetCoreapps` and `NetFramework`.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlExceptionTest.cs
```diff
@@ -33,7 +33,7 @@ public void SerializationTest()
             Assert.Equal(e.StackTrace, sqlEx.StackTrace);
         }
 
-#if !NET5_0
+#if !NET5_AND_ABOVE
```
Wraith2, line 36: If it's not a big job could this be something like NET50_OR_LATER since the versions are continuous in time not number, that way if we somehow end up with a NetNext1.0 it's still later then net5 and the define isn't confusing, OR also makes it exclusive since you can't be both net5 and later than net5.

JRahnama: done

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlExceptionTest.cs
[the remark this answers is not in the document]
JRahnama: We have decided to disable the test for Net 5 and later versions. A constant has been added to address that matter.

src/Microsoft.Data.SqlClient/tests/ManualTests/Microsoft.Data.SqlClient.ManualTesting.Tests.csproj
```diff
@@ -10,6 +10,7 @@
     <Platforms>AnyCPU;x86;x64</Platforms>
     <DefineConstants Condition="'$(TargetGroup)'=='netcoreapp'">$(DefineConstants);NETCOREAPP</DefineConstants>
     <DefineConstants Condition="'$(TargetGroup)'=='netfx'">$(DefineConstants);NETFRAMEWORK</DefineConstants>
+    <DefineConstants Condition="'$(TargetGroup)' == 'netcoreapp' AND (!$(TargetFramework.StartsWith('netcoreapp')) OR (!$(TargetNetCoreVersion.StartsWith('netcoreapp2.1')) AND !$(TargetNetCoreVersion.StartsWith('netcoreapp3.1'))))">$(DefineConstants);NET50_OR_LATER</DefineConstants>
```
cheenamalhotra, line 13:
```suggestion
    <DefineConstants Condition="'$(TargetGroup)' == 'netcoreapp' AND !$(TargetFramework.StartsWith('netcoreapp'))">$(DefineConstants);NET50_OR_LATER</DefineConstants>
```
It should be simplified to above.

Please also ensure "TargetGroup" is set to 'netcoreapp' for .NET 5.0+ in **both Manual Tests and Functional Tests.**
I'm curious to see if pipelines actually test for **TargetGroup = netcoreapp** when targetFramework is **"net5.0"**?

Best solution seems to me:
```
    <TargetGroup Condition="$(TargetFramework.StartsWith('net4'))">netfx</TargetGroup>
    <TargetGroup Condition="$(TargetGroup) == ''">netcoreapp</TargetGroup>
```

src/Microsoft.Data.SqlClient/tests/FunctionalTests/Microsoft.Data.SqlClient.Tests.csproj
```diff
@@ -7,6 +7,7 @@
     <Configurations>Debug;Release;net46-Release;net46-Debug;netcoreapp2.1-Debug;netcoreapp2.1-Release;netcoreapp3.1-Debug;netcoreapp3.1-Release</Configurations>
     <Platforms>AnyCPU;x86;x64</Platforms>
     <DefineConstants Condition="'$(TargetGroup)'=='netcoreapp'">$(DefineConstants);NETCOREAPP</DefineConstants>
+    <DefineConstants Condition="'$(TargetGroup)' == 'netcoreapp' AND (!$(TargetFramework.StartsWith('netcoreapp')) OR (!$(TargetNetCoreVersion.StartsWith('netcoreapp2.1')) AND !$(TargetNetCoreVersion.StartsWith('netcoreapp3.1'))))">$(DefineConstants);NET50_OR_LATER</DefineConstants>
```
cheenamalhotra, line 10: Same as for Manual Tests

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/SqlNotificationTest/SqlNotificationTest.cs
```diff
@@ -11,7 +11,7 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests
     public class SqlNotificationTest : IDisposable
     {
         // Misc constants
-        private const int CALLBACK_TIMEOUT = 5000; // milliseconds
+        private const int CALLBACK_TIMEOUT = 6000; // milliseconds
```
cheenamalhotra, line 14: Any special reason for this change?

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/RandomStressTest/RandomizerPool.cs
```diff
@@ -293,6 +292,14 @@ public void Dispose()
                 }
             }
 
+            /// <summary>
+            /// Disposes the scope. 
+            /// </summary>
+            public void Dispose()
+            {
+                CheckCurrentRandomizerState();
```
cheenamalhotra, line 300: Why do we need delegated call?

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectivityTests/AADConnectionTest.cs
```diff
@@ -202,9 +202,11 @@ public static void testADPasswordAuthentication()
                     }
                 }
             }
-            catch (SqlException e)
+            catch
             {
-                throw e;
+                // re-throwing exception in here will change the stack.
+                // we just throw caught exception.
+                throw;
```
cheenamalhotra, line 209: Might as well not catch it?

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/RandomStressTest/Randomizer.cs
```diff
@@ -237,18 +237,21 @@ protected static int DeserializeInt(byte[] buf, ref int offset)
...
 
         /// <summary>
         /// deserialization constructor
         /// </summary>
         public Randomizer(StreamingContext context)
         {
-            string base64State = GetCurrentState().ToString();
-            int offset;
-            Deserialize(Convert.FromBase64String(base64State), out offset);
+            DeserializeInitializer();
```
cheenamalhotra, line 252: This constructor has no reference, can be removed instead of delegating virtual method call.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/RandomStressTest/RandomizerPool.cs
David-Engel: Also, why are we throwing an exception at all from Dispose()? Dispose() generally should never throw...

cheenamalhotra: Yeah although this is a test - not written well it seems.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/SqlNotificationTest/SqlNotificationTest.cs
JRahnama: these tests were failing on my local machine due to this time out and changing it made them pass. Increasing the wait time made them pass. I think it is up to resources available to the machine that tests are running at. I have reverted the change.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataStreamTest/DataStreamTest.cs
```diff
@@ -2009,11 +2009,12 @@ private static void TimeoutDuringReadAsyncWithClosedReaderTest(string connection
                 }
                 proxy.Stop();
             }
-            catch (SqlException ex)
+            catch (SqlException)
             {
                 // In case of error, stop the proxy and dump its logs (hopefully this will help with debugging
+                // re-throwing exception will change the stack. we just throw the caught exception.
```
johnnypham, line 2015: Since this is the recommended rule, comments like these are unnecessary imo.

src/Microsoft.Data.SqlClient/tests/ManualTests/Microsoft.Data.SqlClient.ManualTesting.Tests.csproj
JRahnama: Done as the suggestion.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/Microsoft.Data.SqlClient.Tests.csproj
JRahnama: done.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/SqlNotificationTest/SqlNotificationTest.cs
JRahnama: reverted.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/RandomStressTest/Randomizer.cs
JRahnama: done.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataStreamTest/DataStreamTest.cs
JRahnama: done.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/RandomStressTest/RandomizerPool.cs
JRahnama: so, here what I did since we are testing, I have replaced the exception with `Trace.Assert` to check for the required the condition. It was really time consuming following the logic and changing the pattern.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectivityTests/AADConnectionTest.cs
JRahnama: removed the try/catch on that part.

cheenamalhotra (approved):

johnnypham (approved):

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/CspProviderExt.cs
```diff
@@ -7,13 +7,19 @@
...
+#endif
 
 namespace Microsoft.Data.SqlClient.ManualTesting.Tests.AlwaysEncrypted
 {
     /// <summary>
     /// Always Encrypted public CspProvider Manual tests.
     /// TODO: These tests are marked as Windows only for now but should be run for all platforms once the Master Key is accessible to this app from Azure Key Vault.
     /// </summary>
+#if NET50_OR_LATER
+    [SupportedOSPlatform("windows")]
```
karinazhou, line 21: Won't the following `PlatformSpecific(TestPlatforms.Windows)` be sufficient?

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/UdtTest/UdtTest.cs
```diff
@@ -294,8 +294,6 @@ public void NullTest()
                         {
                             if(e.Message.Contains("Rerun the transaction"))
                                 rerun = true;
-                            else
```
karinazhou, line 297: Why are we swallowing the exception silently?

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/UdtTest/UdtTest.cs
```diff
@@ -294,8 +294,6 @@ public void NullTest()
                         {
                             if(e.Message.Contains("Rerun the transaction"))
                                 rerun = true;
```
karinazhou, line 296: Can you add `{}` for this single statement?

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/TestFixtures/SQLSetupStrategy.cs
```diff
@@ -272,17 +272,22 @@ public PlatformSpecificTestContext()
...
             {
                 akvFixture = new SQLSetupStrategyAzureKeyVault();
             }
-#endif
         }
 
         public void Dispose()
         {
-            Fixture.Dispose();
+            try
```
karinazhou, line 283: Why don't we use the initial Fixture which will either be `certStoreFixture` or `akvFixture`?

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/Common/AsyncDebugScope.cs
```diff
@@ -101,6 +101,8 @@ public void Dispose()
                     _command.CompletePendingReadWithSuccess(true);
                 }
             }
+            _reader.Dispose();
```
karinazhou, line 104: We need null check before calling Dispose(). Same to _command

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/CspProviderExt.cs
cheenamalhotra: No. That's our own Attribute from XUnit extensions, .NET 5 doesn't recognize it.

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/TestFixtures/SQLSetupStrategy.cs
cheenamalhotra: .NET 5 compiler raises error that these fixtures are not disposed.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/Common/AsyncDebugScope.cs
cheenamalhotra: Moved them in blocks above

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/UdtTest/UdtTest.cs
cheenamalhotra: Done

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/UdtTest/UdtTest.cs
cheenamalhotra: Fixed, was due to .NET error.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/SqlNotificationTest/SqlNotificationTest.cs
```diff
@@ -174,7 +174,8 @@ public void Test_SingleDependency_AllDefaults_SqlAuth()
 
                 updateCompleted.Set();
 
-                Assert.True(notificationReceived.WaitOne(CALLBACK_TIMEOUT, false), "Notification not received within the timeout period");
+                Assert.True(notificationReceived.Wait(CALLBACK_TIMEOUT), "Notification not received within the timeout period");
+                notificationReceived.Reset();
```
karinazhou, line 178: Why we reset the `notificationReceived`  to nonsignaled?

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/TestFixtures/SQLSetupStrategy.cs
karinazhou: If that is the case, maybe simply use this?
```
akvFixture?.Dispose();
certStoreFixture?.Dispose();
```

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/SqlNotificationTest/SqlNotificationTest.cs
cheenamalhotra: Interesting find.
I don't see any reason to make this change, taking out.

karinazhou (approved):

PR #881, opened 2021-01-15
Sync similar elements of SqlBulkCopy

Wraith2:
Another single file change to sync up unimportant changes and do some cleanup so that real differences can be seen.

I included two minor changes.
1) changed the SqlRowsCopied event from using a mateialized backing field and custom add/remove handlers to an auto generated event. I did this because no extra functionality was added by the manual versions and the codegen by the compiler is now safer than the .net 1.0 era manual version, compare [manual](https://sharplab.io/#v2:D4AQTAjAsAUCDMACciDCiDetE8d3+OADgE4CWAbgIYAuApogKIV0B2NAElawCYA2dEogD6dFuwDchPDFzIkYtjSbjO3foJVLpWWXNxUePaXIwz9FkYvYBqALzU+AVzpS9FgL4ncJOgFsAexZvHF1LfVFVAFoHKmdXEMQvdyTYDyA) [auto](https://sharplab.io/#v2:D4AQTAjAsAUCDMACciDCiDetE+UgpgG74B2ALogKLHkASAhiQCYA2+ATlTWQNywC+QA=)
2) removed a this(,,,) ctor call in netfx (Was already removed in netcore) because it was duplicating work done in the ctor body meaning that the only extra work was to create a connection object which was immediately overwritten.

One question has arisen from this set of changes. Why does netfx use a connectionToDoom parameter when setting up async continuations? Netcore doesn't do it but I don't know if this is because netcore had it removed or netfx had it added after the branch for netcore.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
+++ b/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
@@ -16,20 +16,14 @@ using System.Threading.Tasks;
 using System.Xml;
 using Microsoft.Data.Common;
 
-// todo list:
-// * An ID column need to be ignored - even if there is an association
-// * Spec: ID columns will be ignored - even if there is an association
-// * Spec: How do we publish CommandTimeout on the bcpoperation?
-//
-
 namespace Microsoft.Data.SqlClient
 {
     // This internal class helps us to associate the metadata from the target.
     // with ColumnOrdinals from the source.
     internal sealed class _ColumnMapping
     {
-        internal int _sourceColumnOrdinal;
-        internal _SqlMetaData _metadata;
+        internal readonly int _sourceColumnOrdinal;
+        internal readonly _SqlMetaData _metadata;
 
         internal _ColumnMapping(int columnId, _SqlMetaData metadata)
         {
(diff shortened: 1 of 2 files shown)


DavoudEshtehari (approved):

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
```diff
@@ -474,41 +384,15 @@ public int NotifyAfter
...
-        internal int ObjectID
-        {
-            get
-            {
-                return _objectID;
-            }
-        }
-
-        /// <include file='../../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlBulkCopy.xml' path='docs/members[@name="SqlBulkCopy"]/SqlRowsCopied/*'/>
-        public event SqlRowsCopiedEventHandler SqlRowsCopied
```
cheenamalhotra, line 491: It's a public API! Was it removed accidentally?

Wraith2: It was moved to the top of the file as an autogenerated event instead of being down here, the xml comment was moved with it.

cheenamalhotra:
> One question has arisen from this set of changes. Why does netfx use a connectionToDoom parameter when setting up async continuations? Netcore doesn't do it but I don't know if this is because netcore had it removed or netfx had it added after the branch for netcore.

Thta's a good observation. The `connectionToDoom` flow does seem peculiar, and it seems the connection reference is passed over to async tasks to handle events of `Out Of Memory`, `Stack Overflow`, `Thread Aborted Exceptions` in [SqlUtil.cs](https://github.com/dotnet/SqlClient/blob/635f38b598af1efa1cf8b491d87290cc8db33654/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlUtil.cs#L115). I think the goal would have been to cleanup server connections in such cases, and prevent them from pooling.

On the other hand in .NET Core, we simply set Exception on `TaskCompletionSource` (maybe somewhere else it's handled?).
I think we should study this difference carefully to understand if we are cleaning them properly in NetCore too.

cc @saurabh500 who might be able to answer thought behind this difference.

cheenamalhotra (approved):

PR #884, opened 2021-01-15
Security Fix | Prohibit DtdProcessing on XmlTextReader instance in .NET Core

cheenamalhotra:
The setting is already set in NetFx.
Fixes vulnerability.


karinazhou (approved):

DavoudEshtehari (approved):
This is a good practice to set the critical parameters even they're default behaviours.

PR #887, opened 2021-01-18
Sync similar elements Enclaves

Wraith2:
A set of changes to being netcore and netfx closer to parity without changing functionality. Mostly reformatting to the same ruleset and then moving shared core to the shared location.

Observations:
1) netfx uses Cng classes directly but since Cng is the underlying implementation we could use the cross platform version on all targets without any change in functionality. I haven't done this but I think it's strongly worth considering.
2) I changed the CombineByteArrays function to two separate functions because the allocation of a byte[][] each time it was called bothered my optimization sensibilities. 
3) a lot of the code dealing with hashes and keys could be made simpler using Memory<byte> refering to the the original blob if escape analysis proved the lifetimes were closed (and copies made if needed on a late bound basis) and if the code is used often enough for that to be worthwhile.


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AlwaysEncryptedKeyConverter.cs
```diff
@@ -82,10 +35,16 @@ internal static RSAParameters RSAPublicKeyBlobToParams(byte[] keyBlob)
         {
             Debug.Assert(keyBlob.Length == RSAPublicKeyBlob.Size,
                 $"RSA public key blob was not the expected length. Actual: {keyBlob.Length}. Expected: {RSAPublicKeyBlob.Size}");
+
+            byte[] exponent = new byte[RSAPublicKeyBlob.ExponentSize];
+            byte[] modulus = new byte[RSAPublicKeyBlob.ModulusSize];
+            Buffer.BlockCopy(keyBlob, RSAPublicKeyBlob.ExponentOffset, exponent, RSAPublicKeyBlob.ExponentOffset, RSAPublicKeyBlob.ExponentSize);
+            Buffer.BlockCopy(keyBlob, RSAPublicKeyBlob.ModulusSize, modulus, RSAPublicKeyBlob.ModulusOffset, RSAPublicKeyBlob.ModulusSize);
```
DavoudEshtehari, line 42: There is a `System.ArgumentException` related to these lines.

> Offset and length were out of bounds for the array or count is greater than the number of elements from index to the end of the source collection.

Probable fix:
> Buffer.BlockCopy(keyBlob, RSAPublicKeyBlob.ExponentOffset, exponent, 0, RSAPublicKeyBlob.ExponentSize);
> Buffer.BlockCopy(keyBlob, RSAPublicKeyBlob.ModulusOffset, modulus, 0, RSAPublicKeyBlob.ModulusSize);

Wraith2: Good catch, fixed.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AlwaysEncryptedKeyConverter.cs
```diff
@@ -82,10 +35,16 @@ internal static RSAParameters RSAPublicKeyBlobToParams(byte[] keyBlob)
         {
             Debug.Assert(keyBlob.Length == RSAPublicKeyBlob.Size,
                 $"RSA public key blob was not the expected length. Actual: {keyBlob.Length}. Expected: {RSAPublicKeyBlob.Size}");
+
+            byte[] exponent = new byte[RSAPublicKeyBlob.ExponentSize];
+            byte[] modulus = new byte[RSAPublicKeyBlob.ModulusSize];
+            Buffer.BlockCopy(keyBlob, RSAPublicKeyBlob.ExponentOffset, exponent, 0, RSAPublicKeyBlob.ExponentSize);
+            Buffer.BlockCopy(keyBlob, RSAPublicKeyBlob.ModulusSize, modulus, 0, RSAPublicKeyBlob.ModulusSize);
```
DavoudEshtehari, line 42: Shouldn't use `RSAPublicKeyBlob.ModulusOffset` for the second parameter?

Wraith2: Yup. Good catch. I think this must have been the point where I'd run out of attention and decided it was time to finish.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/EnclaveDelegate.CrossPlatformCrypto.cs
```diff
@@ -95,6 +59,68 @@ internal byte[] GetSerializedAttestationParameters(SqlEnclaveAttestationParamete
...
+                    case SqlConnectionAttestationProtocol.HGS:
+                        HostGuardianServiceEnclaveProvider hostGuardianServiceEnclaveProvider = new HostGuardianServiceEnclaveProvider();
+                        s_enclaveProviders[attestationProtocol] = hostGuardianServiceEnclaveProvider;
+                        sqlColumnEncryptionEnclaveProvider = s_enclaveProviders[attestationProtocol];
+                        break;
+
+#if ENCLAVE_SIMULATOR
+                    case SqlConnectionAttestationProtocol.SIM:
+                        SimulatorEnclaveProvider simulatorEnclaveProvider = new SimulatorEnclaveProvider();
+                        EnclaveProviders[attestationProtocol] = (SqlColumnEncryptionEnclaveProvider)simulatorEnclaveProvider;
```
karinazhou, line 106: This SIM case block needs to be changed accordingly as well.

Wraith2: Done!

src/Microsoft.Data.SqlClient/netfx/src/Microsoft.Data.SqlClient.csproj
```diff
@@ -375,18 +387,16 @@
     <Compile Include="Microsoft\Data\Interop\SNINativeManagedWrapperX64.cs" />
     <Compile Include="Microsoft\Data\Interop\SNINativeManagedWrapperX86.cs" />
     <Compile Include="Microsoft\Data\Interop\SNINativeMethodWrapper.cs" />
-    <Compile Include="Microsoft\Data\SqlClient\AlwaysEncryptedEnclaveProviderUtils.cs" />
     <Compile Include="Microsoft\Data\SqlClient\assemblycache.cs" />
     <Compile Include="Microsoft\Data\SqlClient\AzureAttestationBasedEnclaveProvider.cs" />
-    <Compile Include="Microsoft\Data\SqlClient\EnclaveDelegate.cs" />
-    <Compile Include="Microsoft\Data\SqlClient\LocalAppContextSwitches.netfx.cs" />
-    <Compile Include="Microsoft\Data\SqlClient\LocalDBAPI.cs" />
+	  <Compile Include="Microsoft\Data\SqlClient\EnclaveDelegate.CngCryto.cs" />
```
karinazhou, line 392: Please indent Line 392 - 394 properly.

Wraith2: Done.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/EnclaveDelegate.CngCryto.cs
```diff
@@ -95,6 +60,68 @@ internal byte[] GetSerializedAttestationParameters(SqlEnclaveAttestationParamete
...
+                    case SqlConnectionAttestationProtocol.HGS:
+                        HostGuardianServiceEnclaveProvider hostGuardianServiceEnclaveProvider = new HostGuardianServiceEnclaveProvider();
+                        s_enclaveProviders[attestationProtocol] = hostGuardianServiceEnclaveProvider;
+                        sqlColumnEncryptionEnclaveProvider = s_enclaveProviders[attestationProtocol];
+                        break;
+
+#if ENCLAVE_SIMULATOR
+                    case SqlConnectionAttestationProtocol.SIM:
+                        SimulatorEnclaveProvider simulatorEnclaveProvider = new SimulatorEnclaveProvider();
+                        EnclaveProviders[attestationProtocol] = (SqlColumnEncryptionEnclaveProvider)simulatorEnclaveProvider;
```
karinazhou, line 107: Please change this to `s_enclaveProviders`.

Wraith2: Done.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/EnclaveDelegate.CngCryto.cs
```diff
@@ -0,0 +1,230 @@
...
+                    throwIfNull: true
+                );
+            }
+            catch (Exception e)
+            {
+                throw new RetryableEnclaveQueryExecutionException(e.Message, e);
+            }
+
+            List<ColumnEncryptionKeyInfo> decryptedKeysToBeSentToEnclave = GetDecryptedKeysToBeSentToEnclave(keysToBeSentToEnclave, enclaveSessionParameters.ServerName);
+            byte[] queryStringHashBytes = ComputeQueryStringHash(queryText);
```
karinazhou, line 159: Since you are still keeping the CngCrypto version, can we still use the Sha256cng here? Currently, this ComputeQueryStringHash uses Sha256 across .net core and .net framework.

Wraith2: I kept the Cng version because I don't think I should make decisions about standardizing on the cross platform version. I don't own the code so that isn't my decisions to make. I believe there are no reasons not to use only the cross platform version and that it would simplify the codebase. What do you think?

I can break this function out into CrossPlatform and CngCrypto versions if you like but there should be no difference other than the letters Cng at the end of the name as far as I know, is it worth it?

karinazhou: Yea, I am just wondering about the benefit to use the Cng version in the first place. For this specific Sha256 case, I am OK with using the CrossPlatform version if they return the same hash value.

I can see the advantage to keep things standardized instead of maintaining two versions of code. The .net framework version is originally from the System.Data.SqlClient to keep the consistency. We will need to discuss this in the team before we entirely get rid of the whole Cng version stuff. For this PR, I think the current changes are good for now.

karinazhou (approved):

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/EnclaveDelegate.CngCryto.cs
Wraith2: My understanding is that on netfx the use of base classes will actually produce an implementaiton using the Cng classes. So it will be functionally identical. On non-windows it'll use ssl or mac crypto libraries through the netcore PAL layer. Contractually the runtime behaviour should be the same and the algorithms used are well defined and tested.

My assumption is that when the version using Cng was written it was a windows only codebase and there was no need to considering using the base classes.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AlwaysEncryptedKeyConverter.cs
```diff
@@ -137,16 +102,16 @@ internal static byte[] ECDHPublicKeyToECCKeyBlob(ECDiffieHellmanPublicKey public
...
             // Set key size
             keyBlob[4] = (byte)ECCPublicKeyBlob.KeySize;
 
             ECPoint ecPoint = publicKey.ExportParameters().Q;
             Debug.Assert(ecPoint.X.Length == ECCPublicKeyBlob.KeySize && ecPoint.Y.Length == ECCPublicKeyBlob.KeySize,
                $"ECDH public key was not the expected length. Actual (X): {ecPoint.X.Length}. Actual (Y): {ecPoint.Y.Length} Expected: {ECCPublicKeyBlob.Size}");
             // Copy x and y coordinates to key blob
-            Array.Copy(ecPoint.X, 0, keyBlob, ECCPublicKeyBlob.HeaderSize, ECCPublicKeyBlob.KeySize);
-            Array.Copy(ecPoint.Y, 0, keyBlob, ECCPublicKeyBlob.HeaderSize + ECCPublicKeyBlob.KeySize, ECCPublicKeyBlob.KeySize);
+            Buffer.BlockCopy(ecPoint.X, 0, keyBlob, ECCPublicKeyBlob.HeaderSize, ECCPublicKeyBlob.KeySize);
```
johnnypham, line 113: From what I've read, using either one would make little difference in this situation. Any particular reason for changing it?

Wraith2: Consistency with the rest of the library. We use blockcopy 41 times and array copy 24 times. As you say there is no performance advantage either way. If you prefer the original I'll change it back, let me know.

johnnypham: Doesn't matter to me, just curious :grin:

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/EnclaveDelegate.CngCryto.cs
johnnypham: I'm confused here. We've decided to use the cross platform SHA256 class, so do we still need this CngCrypto file (typo in the name btw)?

Wraith2: Has it been decided? If so I agree with the decision but I didn't know it had been made. 
I think it would be cleanest to merge this PR as it is and I will follow up with another to make only the changes required to move to managed for all builds. If I try to do it here you'll have to re-review everything.

johnnypham: Sorry, yeah we discussed it internally. You're right, let's leave it for another pr.

johnnypham (approved):

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/EnclaveDelegate.CngCryto.cs
Wraith2: Well I suppose it's nice to be considered part of the team but I feel I'm missing a lot of meetings that I should be attending if that's the case 😁

PR #889, opened 2021-01-26
small perf changes

Wraith2:
I did a memory profile of the test case in https://github.com/dotnet/SqlClient/issues/659 and noted some easy to fix issues:

Untitled

The ConnectionState change was using `.Equals(object)` which caused boxing, it's just changed to state value capture and a direct comparison. The `ConcurrentQueueSemaphore` simply changes a delegate instantiation and context capture to a static delegate and state object parameter invocation as we've used elsewhere to reduce allocations.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnection.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnection.cs
@@ -722,8 +722,7 @@ namespace Microsoft.Data.SqlClient
         {
             get
             {
-                ConnectionState state = State;
-                if (state == ConnectionState.Open || state == ConnectionState.Executing || state == ConnectionState.Fetching)
+                if ((State & (ConnectionState.Open | ConnectionState.Executing | ConnectionState.Fetching)) > 0)
                 {
                     return GetOpenTdsConnection().ServerProcessId;
                 }
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnection.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnection.cs
@@ -720,8 +720,15 @@ namespace Microsoft.Data.SqlClient
         /// <include file='../../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnection.xml' path='docs/members[@name="SqlConnection"]/ServerProcessId/*' />
         public int ServerProcessId
         {
-            get => State.Equals(ConnectionState.Open) | State.Equals(ConnectionState.Executing) | State.Equals(ConnectionState.Fetching) ?
-                GetOpenTdsConnection().ServerProcessId : 0;
+            get
+            {
+                ConnectionState state = State;
+                if (state == ConnectionState.Open || state == ConnectionState.Executing || state == ConnectionState.Fetching)
+                {
+                    return GetOpenTdsConnection().ServerProcessId;
+                }
+                return 0;
+            }
         }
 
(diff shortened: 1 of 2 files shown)


cheenamalhotra (approved):

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnection.cs
```diff
@@ -720,8 +720,15 @@ public override string ServerVersion
         /// <include file='../../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnection.xml' path='docs/members[@name="SqlConnection"]/ServerProcessId/*' />
         public int ServerProcessId
         {
-            get => State.Equals(ConnectionState.Open) | State.Equals(ConnectionState.Executing) | State.Equals(ConnectionState.Fetching) ?
-                GetOpenTdsConnection().ServerProcessId : 0;
+            get
+            {
+                ConnectionState state = State;
```
DavoudEshtehari, line 725: I believe in using the `State` directly to access the latest connection status. Unless there would be some concerns if it changed in between.

NIT: `if (State & (ConnectionState.Open | ConnectionState.Executing | ConnectionState.Fetching) > 0)`

Wraith2: I considered that. ConnectionState isn't `[Flags]` and while the values are discrete and would work I prefer not to set bad precedents by taking shortcuts like that. The [IL is shorter](https://sharplab.io/#v2:C4LglgNgPgAgTARgLACgYAYAEMEDoAiAhsIQNyowDM2cmAwpgN6qauYttXYAsmAggAoAlOxRsmHcazAAzTAIEBlEsACmmAGTy6AewB2e1QGNgYfcuKrcAeQAOqvZij19hk2b0W1uAKIAPYwBXUz0AcycXA2NTcxUrADFVYCMACzAwoREAPkx0IUkpZjEpEpwATgEAIj5KoXJikoBfAuaG0XEuGF4AIWF2tiKS1l0o91jLTABnOMwAXkwvVXqhzFl5KZnZ+ZG3GM84m3tHKGdpia3I3Y9F3wCjYPTwk43z7ddo64PE5LSMgvFBitOAgKpVurVlkNWuJoWwClwdh9xmoFjNGKEkqRJpjWo0gA=) in your version but since I was only trying to remove the repeated box allocations. Do you feel strongly that it should be changed?

DavoudEshtehari: My thoughts are about a possible concurrent program flow that a connection state is changed after assigning it to a local variable that means the connection won't be valid anymore while comparing it at the next level. Its minor possibility will decrease by the suggested change.
[ConnectionState Enum](https://docs.microsoft.com/en-us/dotnet/api/system.data.connectionstate?view=netcore-3.1)

Wraith2: If we don't take a local copy of State and then do 3 checks we're at risk of getting a misleading answer as you say. Both our approaches are avoiding that, mine by copy yours by single access. You use masking which I chose not to because of type purity but you get shorter code, as i said i wasn't trying to minimize code size just memory.
So both approaches are valid I'd just slightly prefer mine while acknowledging the benefits of yours. I'm open to change if you feel strongly. Do you want me to?

DavoudEshtehari: Thank you for the quick responses and clarifications. I didn't want to persist on my proposal first, but I believe there are some benefits. So, I'd rather it be applied. 😉

Wraith2: Ok, done.

DavoudEshtehari (approved):

PR #890, opened 2021-01-26
Fix | Fix possible server connection leak if an exception occurs in pooling layer

cheenamalhotra:
This PR is additional fix for consequences of error `Internal .NET Framework Data Provider error 6.` that is thrown from `DbConnectionPool.CreateObject` method when something goes wrong in Pooling layer.

Continued investigations over original issue that was fixed by #637 led to figuring out that the catch blocks in this method simply set `newObj` to `null`, leaking resources and TDS connections on server, that can only be garbage collected on client when server drops them.

This causes expensive connection leaks on server, and issue exists in System.Data.SqlClient and Microsoft.Data.SqlClient (<2.0.1).
In Microsoft.Data.SqlClient we fixed this issue with PR #638 and #637, this PR is also a fix that ensures resources are cleaned properly in any similar future events.


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/ProviderBase/DbConnectionPool.cs
```diff
@@ -772,7 +772,14 @@ private DbConnectionInternal CreateObject(DbConnection owningObject, DbConnectio
 
                 CheckPoolBlockingPeriod(e);
 
+                // Close associated Parser if connection already established.
+                if (newObj != null && newObj.IsConnectionAlive())
```
saurabh500, line 776: Nit, this can be `newObj?.IsConnectionAlive()`

saurabh500:
This change can be tested using the TestTdeServer and injecting transient failures 
e.g. https://github.com/dotnet/SqlClient/blob/9e3f0d625eb5ec15e069a1d8f118d3bac41b61f5/src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlConnectionBasicTests.cs#L18

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/ProviderBase/DbConnectionPool.cs
cheenamalhotra: I tried that, got an error "Cannot cast `bool?` to `bool`"

JRahnama:
LGTM.

JRahnama (approved):

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/ProviderBase/DbConnectionPool.cs
flts: `if (newObj?.IsConnectionAlive() == true)` could solve the cast error message.

David-Engel (approved):

DavoudEshtehari (approved):

cheenamalhotra:
> This change can be tested using the TestTdeServer and injecting transient failures

Hi @saurabh500 I've added the test case and I can also reproduce `Internal .NET Framework Data Provider error 6.` with connection leak with the new test "**TransientFaultTest**". You'd need to revert changes from this PR and https://github.com/dotnet/SqlClient/pull/637 (just comment out changes) to reproduce it locally.

In order to capture the leaked connection, you'll need to add breakpoint in exception handling logic in the test and capture heap snapshot. `SqlConnection` instance no longer contains reference of this "Open" internal connection so it's reference cannot be captured via reflection and hence state is always _Closed_.

Navigating through Heap Snashot you will come across 1 instance of `TdsParserStateObjectNative` which has `TdsParser` reference and it's State is **OpenLoggedIn**. Since only internal connection reference is dropped in the code where I fixed in this PR, the TCP connection to server is intact and is leaked.

image

Issue #262, opened 2019-10-13
System.Data.SqlClient 4.7.0 tends to deadlock on async


marcwittke:
I am afraid I cannot provide a clean reproduction, but I want to share my experience, so that you could decide if this is an issue or not.

Turns out, that with the bump to 4.7.0 (from 4.6.1) our CI build deadlocks almost every run during unit tests. There is neither a problem on development environments nor on production.

The setup: a xunit test suite starts a .net core 2.1 service host as shared fixture to run some integration tests. part of the application stack is asp.net identity core. The whole environment is being started via `docker-compose` so that the dependent SQL Server is also available. A rather complete test database seeding is done during test initialization. Build agents are cheap Azure VMs of B2s size (2vCPUs, 4GB RAM) running Ubuntu 18.04 with docker. 

Since I was unable to dump/debug the locked process on the VM inside docker, I spent some time to reproduce the issue locally (i7-8850H, Archlinux, Jetbrains Rider). I finally made it by running `stress --cpu 12 --io 10 --vm 5 --hdd 1`, starting the `sqlserver` process  inside docker and limiting it via `taskset -p 0x01 nnnn` to one CPU, and starting the integration test restricted to the same CPU directly via `taskset 0x01 dotnet test`.

Attaching a debugger on the locked process reveals the following stack trace (besides various others waiting on reset events, like the xunit runner, test diagnostics and Application insights):

![image](https://user-images.githubusercontent.com/4075651/66707544-60c48280-ed18-11e9-8085-8a969ade641e.png)

the last frame on the stack resulting from our application code is this line:
```csharp
await userManager.GetAuthenticationTokenAsync(user, TokenManager.TokenProviderName, "refreshToken"); 
``` 

Note that there is not a single `Task.Wait()` or `Task.Result` in the whole code base, everything is "async all the way down". After suspecting the advanced parallelization features of xUnit as part of the problem, I disabled all test parallelization and lifted all restrictions on allowed threads, but without any effect. The SQL Server is only hit by one open connection, state "await command". 

Reverting the solution to System.Data.SqlClient 4.6.1 made all problems go away, there was not a single locking build since then, so I am pretty sure it's related.

Note: The solution is on a private GitHub Repo, but I'd be able to share an export or more details, when needed.

Wraith2:
Unless you're using mars can you try turning it off? It significantly increases internal complexity on the managed implementation which is what's used on Unix.

marcwittke:
[...]

PR #902, opened 2021-02-07
Add ValueTask stream overloads on SNI streams and Waits

Wraith2:
Some more perf improvements from investigating https://github.com/dotnet/SqlClient/issues/659

This adds overloads on netcore builds which contain ValueTask so waits which can succeed synchronously can avoid allocating tasks and completion sources. This eliminates or reduces the highlighted items below. In my benching run it increases throughput by ~30%.

Untitled


cheenamalhotra:
You might want to run some tests locally that are not completing and hanging in the pipelines, they don't seem usual errors from test lab.

Wraith2:
Test run fine locally, can someone with access re-run them?

cheenamalhotra:
Could you run with Managed SNI enabled on Windows? Just enable that property in config.json.

Wraith2:
I ran the tests with managed mode hardcoded in the stateobject factory. Any specific things you see of concern? i checked a few errors and didn't see anything unusual.

cheenamalhotra:
Try with netcoreapp2.1

Wraith2:
I'm not sure what's going on with the tests. I rebased on master and force pushed which caused the CI to re-run and has greened a lot of the failing legs. There are still some which stop executing but give no indication why. Importantly there are managed sni tests that complete successfully.

On the legs that didn't complete the next test should have been: Passed Microsoft.Data.SqlClient.ManualTesting.Tests.AsyncTimeoutTest.TestDelayedAsyncTimeout

cheenamalhotra:
If you check here, this is actually using Managed SNI for all tests (UseManagedSNIOnWindows = true) : https://dev.azure.com/sqlclientdrivers-ci/sqlclient/_build/results?buildId=25670&view=results

`netcoreapp2.1` jobs are "hung up" after running few tests:

image

Same is with other pipelines.
I suspect this is due to some changes here.

cheenamalhotra:
Hang is captured by recently added `AsyncTimeoutTests` on .NET Core 2.1.
More tests were possibly also capturing it before.

You can use below scripts to reproduce:
```bash
#build
msbuild /p:configuration=Release

#hangs forever
msbuild /p:configuration=Release /t:BuildTestsNetCore
dotnet test "src\Microsoft.Data.SqlClient\tests\ManualTests\Microsoft.Data.SqlClient.ManualTesting.Tests.csproj" /p:Configuration="Release" /p:TestTargetOS="Windowsnetcoreapp" --no-build -v n --filter "AsyncTimeoutTest"

#works
msbuild /p:configuration=Release /t:BuildTestsNetCore /p:TargetNetCoreVersion=netcoreapp3.1
dotnet test "src\Microsoft.Data.SqlClient\tests\ManualTests\Microsoft.Data.SqlClient.ManualTesting.Tests.csproj" /p:Configuration="Release" /p:TestTargetOS="Windowsnetcoreapp" --no-build -v n --filter "AsyncTimeoutTest" /p:TargetNetCoreVersion=netcoreapp3.1
```

Wraith2:
Yup, saw that was where it hung but I can't work out why and locally it doesn't. There are legs which use the managed sni that are succeeding
```
CI-Win10-SQL19W-ManagedSNI (Win10-SQL19 AnyCPU,Release,netcoreapp,false,netcoreapp3.1) Successful in 18m
```
```
CI-Ub20-SQL17L-Core31 (Ubuntu 16 AnyCPU,Release,netcoreapp,true) Successful in 20m
```

The changes I've made are working fine for those. The netcore files are included in 2.1 and 3.1 builds so if there's a difference it's in the runtime and not this code. Given that 2.1 is out of support I could make these changes 3.1> only but I'd really like to know what that problem is rather than just avoiding it. 

Is there a way to get some sort of diagnostic information out of the CI machines so I can see what they were hung on?

Wraith2:
I've found what I think is the problem but I'm going to need some help. The manual tests project can be debugged in visual studio if you change the build configuration to netcoreapp2.1-Debug and by doing that I see what's happening. What i need to be able to do is see the same situation with netcoreapp3.1-Debug but that configuration won't build and i can't work out what's wrong with it. Is there any chance you could take a look at the project config for me?

Wraith2:
I've found it and it's annoying. Netcore 2.1 implemented ValueTask overrides by calling the Task versions so you can end up with a cycle. I've changed 2.1 builds to use the netcore strategy and put a note at the top of the file explaining why. Lets see what the CI thinks.

Wraith2:
@cheenamalhotra @David-Engel can this be reviewed now please? It's been open 4 months now and if it was too risky for 3.0 for some reason that has now shipped.

JRahnama:
@Wraith2 can you address the conflict here please?

Wraith2:
Done.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIStreams.Task.cs
```diff
@@ -0,0 +1,75 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Threading;
+using System.Threading.Tasks;
+
+
```
JRahnama, line 8: NIT: remove extra line here.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIStreams.Task.cs
```diff
@@ -0,0 +1,75 @@
...
+            finally
+            {
+                _writeAsyncSemaphore.Release();
+            }
+        }
+    }
+
+    internal sealed partial class SNINetworkStream
+    {
+
```
JRahnama, line 47: NIT: remove extra line here please.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIStreams.ValueTask.cs
```diff
@@ -0,0 +1,122 @@
...
+        public override Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
+        {
+            ValueTask<int> valueTask = ReadAsync(new Memory<byte>(buffer, offset, count), cancellationToken);
+            if (valueTask.IsCompletedSuccessfully)
+            {
+                return Task.FromResult(valueTask.Result);
+            }
+            else
+            {
+                return valueTask.AsTask();
```
JRahnama, line 22: What are the benefits of starting with a value task and returning a task? can we just start with a task and return a task?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIStreams.cs
```diff
@@ -13,7 +13,7 @@ namespace Microsoft.Data.SqlClient.SNI
     /// <summary>
```
JRahnama, line 13: There are two  unused libraries at the top System.Threading and System.Threading.Tasks can you remove them please?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIStreams.ValueTask.cs
Wraith2: The only version that does the real work is the ValueTask version so everything must call that one and then convert the result to whatever upstream type it needs. I agree that it looks wasteful but it's more complex to have two complete implementations and this is the pattern used in /runtime 

In an ideal world all callers will use ValueTask directly and avoid the Task. In cases where somehow the caller can't do that (isn't aware of ValueTask) they'll incur the same task allocation penalty they were already going to have to and the only change is a couple of additional stack copy instructions. Overall it's a net win. 

Since we (sqlclient) are the only caller of these functions this PR completes the vertical call chain that uses ValueTask instead of Task and we'll see the benefits, this method might only get called through sequential reader streams.

JRahnama: In several cases like this, the consumer is not waiting them directly they are waiting for a Task and that makes me wonder if value task is a good choice.


> In an ideal world all callers will use ValueTask directly and avoid the Task.

This is a bit controversial. Tasks are still preferred over value tasks. Micro benchmarking shows a bit of perf cost when using value tasks. They are easier to use correctly. 

According to [this article](https://devblogs.microsoft.com/dotnet/understanding-the-whys-whats-and-whens-of-valuetask/#should-every-new-asynchronous-api-return-valuetask-valuetasktresult) a Value task should only be used if:
a) you expect consumers of your API to only await them directly, 
b) allocation-related overhead is important to avoid for your API, and 
c) either you expect synchronous completion to be a very common case, or you’re able to effectively pool objects for use with asynchronous completion. 

I would ask @roji's help/comment on this as well.

Wraith2: This fits into cases b and c. The fact that we can't change the external api surface easily to be ValueTask instead of Task based prevents it being a as well.

> This is a bit controversial. Tasks are still preferred over value tasks. Micro benchmarking shows a bit of perf cost when using value tasks. They are easier to use correctly.

One of the things I do for fun is listen to the framework design council meetings when they're streamed on YouTube and I've heard the long conversations about when ValueTask is suitable and when it is not. I investigated this because I had good reason to believe that using ValueTasks is appropriate and would sometimes allow us to have allocation free sync calls. The items I highlighted in the original post show the reduction in allocations from these changes. There will be speed increases as well because we spend less time in GC but I didn't think it necessary to quantify them because they weren't the reason for making the changes.

Note that this works on netcore3 best because the runtime did the work to make SSLStream more efficient by adding ValueTask overloads, we're just following the same pattern to allow the most efficient reads we can.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIStreams.cs
Wraith2: Done.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIStreams.Task.cs
Wraith2: Done.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIStreams.Task.cs
Wraith2: Done.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIStreams.ValueTask.cs
```diff
@@ -0,0 +1,122 @@
...
+using System;
+
+namespace Microsoft.Data.SqlClient.SNI
+{
+    internal sealed partial class SNISslStream
+    {
+        public override Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
+        {
+            ValueTask<int> valueTask = ReadAsync(new Memory<byte>(buffer, offset, count), cancellationToken);
+            if (valueTask.IsCompletedSuccessfully)
```
roji, line 16: This seems redundant - ValueTask.AsTask already knows to efficiently return a completed Task.

Though I'd personally just call the corresponding Task-returning method on the base class directly, and apply the semaphore check here. At least theoretically, the Task-returning method could be doing something slightly different since it knows it's returning a Task (e.g. no need to deal with IValueTaskSource etc.).

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIStreams.ValueTask.cs
roji: I do agree it's worth introducing ValueTask implementations here - this seems to be a Stream implementation that all driver traffic goes through, and it could indeed reduce lots of allocations when the call completes synchronously (and possibly even asynchronously, depending on how optimized the rewritten SslStream is). There's indeed a lot of discussions around whether to **expose** ValueTask vs. Task when designing APIs, but when **consuming** an API that exposes ValueTask it's generally good practice to use that (unless you're sure the invocation will return asynchronously, and that no asynchronous-completion optimizations are in place in the implementation).

> In my benching run it increases throughput by ~30%.

That seems like quite a high number for this change... FWIW I always find it helpful to post the actual BenchmarkDotNet results and code, to show what exactly is being tested etc.

roji: However... @Wraith2 I'm not seeing changes to actually call the ValueTask methods, e.g. SNIPacket.ReadFromStreamAsync still calls the Task-returning overload of Stream.ReadAsync, and chains a ContinueWith call on it. I'm not sure what value that would have - it gets a `ValueTask<int>`, only to convert it to a `Task<int>` (so this should in fact increase overhead).

What am I missing here?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIStreams.ValueTask.cs
Wraith2: So you think it would be better to just do `return ReadAsync(new Memory<byte>(buffer, offset, count), cancellationToken).AsTask();` ?

roji: Yeah, it's basically the same. Take a look at the source for AsTask.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIStreams.ValueTask.cs
Wraith2: I was surprised by the numbers but remember it's throughput not speed. The number comes from looking at the number of Fortunes objects allocated in before and after profiles. Since the profile is limited to a specific time it can be extrapolated. I didn't measure increase more carefully because it wasn't speed I was aiming to improve, it was allocations.

Wraith2: Using ValueTask in SNIPacket is on my backlog, i'll get there eventually.

roji: I'd be interested in understanding how allocations were reduced if the codebase is still calling the Task-returning overloads though... I'm not doubting what you say - just trying to understand the reduction is working (is there another path I'm missing which calls the ValueTask-returning overloads?). 

> Using ValueTask in SNIPacket is on my backlog, i'll get there eventually.

I'm not sure how you'd do that, considering that ContinueWith works with Task only. And since that is used all across the codebase, I think that moving towards ValueTask would require a more general overhaul towards async/await - which is something we've discussed in the past.

Re this PR specifically... Assuming my understanding is correct and the codebase still calls the Task-returning overloads, then I don't see how there's any value in this change on its own...

Wraith2: The change to `ConcurrentQueueSemaphore` is a big part of it, and I think I already split that out into another PR because it was so valuable. It prevents the allocation of a TCS, Task, and two continue objects each time we want to check a lock (which is usually uncontested) for every single packet receipt.

The other changes allow value task paths to be used further down. Without this PR I can't use them in SNIPacket so I've got to start somewhere and I started here. If I'd tried to use them in SNIPacket first you'd complain that it was no use doing that because SslStream didn't have it in. If I try to enable ValueTask everywhere in one PR it'll be a very large PR would be treated with even more suspicion than this one already is.

If you like just ignore the throughput increase and focus on what I was really trying to do here which was enable ValueTask incrementally. At the moment we still have to use AsTask() which clearly isn't a good thing but if the ValueTask paths work and are just converted to Task later then it lets me cleanly start to use ValueTask later for gains.

I used some of the existing runtime stream implementations to inform me on implementing these methods. I can't remember which one exactly because it's been 4 months but I didn't just make all of this up.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/ConcurrentQueueSemaphore.NetCoreApp.cs
```diff
@@ -0,0 +1,32 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Data.SqlClient.SNI
+{
+    internal sealed partial class ConcurrentQueueSemaphore
+    {
+        public ValueTask WaitAsync(CancellationToken cancellationToken)
```
roji, line 12: I'm not sure what the point of returning ValueTask is - is this better than returning a Task? For the first path below, you can simply return Task.CompletedTask instead of `new ValueTask()`, and for the second one, just return `tcs.Task`...

Non-generic ValueTask is generally useful only in quite rare circumstances. It only really makes sense if it's going to be backed with IValueTaskSource (i.e. optimizing for asynchronous completion). It doesn't help for optimizing for synchronous completion, since you already have the option of returning Task.CompletedTask; that option doesn't exist in generic contexts, which is why `ValueTask<T>` is far more useful than non-generic `ValueTask`.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIStreams.ValueTask.cs
roji: > The change to ConcurrentQueueSemaphore is a big part of it, and I think I already split that out into another PR because it was so valuable. It prevents the allocation of a TCS, Task, and two continue objects each time we want to check a lock (which is usually uncontested) for every single packet receipt.

It's great that you're making progress. I haven't review this PR thoroughly, though specifically for making ConcurrentQueueSemaphore not allocate for the uncontended case, this seems to [already be the case as far as I can tell](https://github.com/dotnet/SqlClient/pull/902/files#diff-992e2003f7c476cd1e0a3de1867f48ecddbb82ad2a57951db415a8aead4eb92aL2165) seems to have already done that for the uncontested case. Also see [my comment](https://github.com/dotnet/SqlClient/pull/902/files#r653146582) on the non-utility of non-generic ValueTask there.

> The other changes allow value task paths to be used further down. Without this PR I can't use them in SNIPacket so I've got to start somewhere and I started here. If I'd tried to use them in SNIPacket first you'd complain that it was no use doing that because SslStream didn't have it in.

@Wraith2 I'm not complaining about anything. My opinion was asked for, and I'm pointing out that this change, as it currently is, doesn't improve performance in any way, and indeed may slightly regress it. I very much support improving how async is implemented in SqlClient, though I think the real problems really are elsewhere, and not in Task vs. ValueTask.

Wraith2: > specifically for making ConcurrentQueueSemaphore not allocate for the uncontended case, this seems to already be the case as far as I can tell

I split that one out after talking with cheena a few months back. It has a big impact and I wanted to make sure it got in as soon as possible. I don't keep rebasing my open PR's until there's some reason to do so, there's no point me taking the time if no-one is ready to review them, I used to do that but it took ages and the number of branches I'd have to review got out hand.

Outisde a microbenchmark designed to show the difference between ValueTask and Task implementations do you think there is likely to be a noticeable performance difference between the versions? I accept and agree with your statements but is it either harmful or wrong to have a ValueTask version?

If you can point me at the other problems I can look at them. Until then I'll keep investigating and addressing the problems I find. I still hope that one day we'll be able to get Sql Server included in the TE benchmarks. The TE Fortunes benchmark is also a simple and easily understood optimization scenario to work on so I'll come back to it often to see if there are new optimization opportunities.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/ConcurrentQueueSemaphore.NetStandard.cs
```diff
@@ -0,0 +1,32 @@
...
+using System.Threading.Tasks;
+
+namespace Microsoft.Data.SqlClient.SNI
+{
+    internal sealed partial class ConcurrentQueueSemaphore
+    {
+        public Task WaitAsync(CancellationToken cancellationToken)
+        {
+            // try sync wait with 0 which will not block to see if we need to do an async wait
+            if (_semaphore.Wait(0, cancellationToken))
```
vonzshik, line 15: Pretty sure I’m missing something, but what’s the purpose of this wrapper, isn’t it enough to just call SemaphoreSlim.WaitAsync directly?

Wraith2: Calling WaitAsync causes 4 objects to be allocated. Avoiding those allocations and the GC that they become part of increases throughput. In practice this semaphore is very unlikely to be contended because it's quite hard to run the client faster than the server can send data.

vonzshik: I've just run this example through dotMemory on .net5 and I see almost zero allocations:
```csharp
Console.WriteLine("Ready");
Console.ReadKey();

var cts = new CancellationTokenSource();
var sem = new SemaphoreSlim(5);

for (var i = 0; i < 10000000; i++)
{
	await sem.WaitAsync(cts.Token);
	sem.Release();
}

Console.WriteLine("Done");
Console.ReadKey();
```

image

vonzshik: Same thing for .net core 3.1.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/ConcurrentQueueSemaphore.NetStandard.cs
```diff
@@ -0,0 +1,32 @@
...
+                return Task.CompletedTask;
+            }
+            else
+            {
+                var tcs = new TaskCompletionSource<bool>();
+                _queue.Enqueue(tcs);
+                _semaphore.WaitAsync().ContinueWith(
+                    continuationAction: s_continuePop,
+                    state: _queue,
+                    cancellationToken: cancellationToken
```
vonzshik, line 26: Doesn't it mean that it's going to cancel the continuation on `_semaphore.WaitAsync()`, so there will be no one to dequeue the tcs, and the tcs in question is never going to be cancelled?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/ConcurrentQueueSemaphore.NetStandard.cs
Wraith2: Sure, looks right, now try it without being able to use language based async and add in a TCS. Allocations then become 1 Task, 1 TCS, 2 continuation objects.
```
var tcs = new TaskCompletionSource<bool>();
_queue.Enqueue(tcs);
_semaphore.WaitAsync().ContinueWith(
    continuationAction: s_continuePop,
    state: _queue,
    cancellationToken: cancellationToken
);
return tcs.Task;
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/ConcurrentQueueSemaphore.NetStandard.cs
Wraith2: Unknown, I'm not adding this functionality just making it faster, you probably need @cheenamalhotra

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/ConcurrentQueueSemaphore.NetStandard.cs
vonzshik: Yeah, but you can replace the whole `ConcurrentQueueSemaphore.WaitAsync` with just a call to `_semaphore.WaitAsync()`, it essentially does the same thing. The only difference, `SemaphoreSlim` doesn't guarantee FIFO (even though the references do suggest otherwise, at the very least for async path), but I can't say whether it's actually even required here.

Wraith2: The fifi part is important, ordering of requests should be preserved or packets could effectively be seen our of order. But again, I didn't write this part I'm just making it faster.

roji: I do get the need for FIFO here. My main view on this code, as you probably already know, is that it should be rewritten with async/await. That would both make it more efficient and considerably simplify the code.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIStreams.ValueTask.cs
roji: > I accept and agree with your statements but is it either harmful or wrong to have a ValueTask version?

No - I don't think it's either harmful or wrong. As I wrote above, I'd implement the Task-returning versions by calling the base SslStream counterparts (rather than the ValueTask), but that's a really small suggestion.

> The TE Fortunes benchmark is also a simple and easily understood optimization scenario to work on so I'll come back to it often to see if there are new optimization opportunities.

I agree that TE Fortunes is a great way to find optimization opportunities (regardless of whether results are officially published on techempower.com or not). But I'll go out on a limb here, and say that IMHO a few micro-optimizations here and there simply aren't going to have a substantial impact given the larger problems in the codebase. At some point these need to be addressed.

PS I'm still unclear on where the aforementioned allocation reductions and throughput improvements come from in this PR, since the new Stream ValueTask methods aren't called, and I can't see how the ConcurrentQueueSemaphore changes improve anything.

Wraith2: Ignore the semaphore stuff, that got merged separately. It was the perf improvement.
The stream valuetasks just enable me to use them in a future PR.

roji: OK, though consider https://github.com/dotnet/SqlClient/pull/902#discussion_r653146582.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/ConcurrentQueueSemaphore.NetStandard.cs
cheenamalhotra: The reason why we needed to wrap calls to Stream was to ensure Stream's `ReadAsync` and `WriteAsync` don't get triggered concurrently, as they don't support concurrency. The FIFO queue is mainly to set the order right as there's no order guarantee in SemaphoreSlim. https://github.com/dotnet/SqlClient/pull/796

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/ConcurrentQueueSemaphore.NetStandard.cs
cheenamalhotra: It could be addressed, as it's important to end the wait on cancellation than to cancel the deque. The deque should ideally happen even on cancellation.

Wraith2:
Issues addressed. Concurrent queue simplified to task only version for all builds.

Wraith2:
What is blocking this?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
```diff
@@ -393,6 +394,8 @@
     <Compile Include="Microsoft\Data\SqlClient\SNI\SslOverTdsStream.NetCoreApp.cs" />
     <Compile Include="Microsoft\Data\SqlClient\SqlConnectionFactory.AssemblyLoadContext.cs" />    
     <Compile Include="Microsoft\Data\SqlClient\SqlDependencyUtils.AssemblyLoadContext.cs" />
+    <Compile Condition="$(TargetFramework.StartsWith('netcoreapp2.'))" Include="Microsoft\Data\SqlClient\SNI\SNIStreams.Task.cs" />
+    <Compile Condition="!$(TargetFramework.StartsWith('netcoreapp2.'))" Include="Microsoft\Data\SqlClient\SNI\SNIStreams.ValueTask.cs" />
```
cheenamalhotra, line 398: This should also be handled for **.NET Standard 2.0 (netstandard2.0)**, since .NET Core 2.1 can be targeted via .NET Standard 2.0 DLL too.

Also for context, we will be removing .NET Core 2.1 DLL from NuGet at some point in future as it's going out of mainstream support soon, but .NET Standard 2.0 DLL will continue to work for .NET Core 2.1.

cheenamalhotra (approved):
Overall LGTM.
PR is also tested with repro from #262 as deadlocks occurred in the past from similar changes.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
Wraith2: The netstandard inclusion is done with the task version on line 347 above.

cheenamalhotra: Ok, so then we need to address it for .NET Standard 2.1. Since we have 2 different DLLs:
- **netstandard2.0** (also applies indirectly to .NET Core 2.1+)
- **netstandard2.1** (also applies indirectly to .NET Core 3.1+)
 
And [ValueTask](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.valuetask-1?view=net-5.0) is available in netstandard2.1, so we should provide it there IMO.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
```diff
@@ -344,6 +344,7 @@
     <Compile Include="Microsoft\Data\SqlClient\SqlDiagnosticListener.NetStandard.cs" />
     <Compile Include="Microsoft\Data\SqlClient\SqlDelegatedTransaction.NetStandard.cs" />
     <Compile Include="Microsoft\Data\SqlClient\TdsParser.NetStandard.cs" />
+    <Compile Include="Microsoft\Data\SqlClient\SNI\SNIStreams.Task.cs" />
```
cheenamalhotra, line 347: Applies to `'$(TargetGroup)' == 'netstandard'` :

```suggestion
    <Compile Condition="'$(TargetFramework)' == 'netstandard2.0'" Include="Microsoft\Data\SqlClient\SNI\SNIStreams.Task.cs" />
    <Compile Condition="'$(TargetFramework)' != 'netstandard2.0'" Include="Microsoft\Data\SqlClient\SNI\SNIStreams.ValueTask.cs" />
```

Anything above netstandard2.0 will take in ValueTask code.

Wraith2: anything above netstandard 2.0 is netstandard 2.1 or netcore 3.1 or later, so valuetask is fine.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
Wraith2: It depends on the exact implementation. If you tried to use the ValueTask version on netcore2.1, which is possible to do because it's netstandard 2.0 compliant, it would end up in the infinite loop that I fixed earlier on because while ValueTask is supported the _implementation_ on netcore2.1 forwards valuetask to task. It's not safe to use ValueTask unless you know the runtime supports it directly. 

The current setup is safe. Anything else you'll need to check independantly.

cheenamalhotra: We can leave it too, there's no requirement or blocking concern - it just creates separate paths for .NET Standard 2.1 and .NET Core 3.1, diverging their code further. But I agree unknown cases can occur with other targets.

JRahnama (approved):

PR #910, opened 2021-02-13
Fix | MARS header contains errors on .NET Framework 4.8

David-Engel:
Fixing #85 MARS header contains errors for .NET Framework 4.8.

The underlying issue of why this was failing for .NET Framework 4.8 was that there was conditional logic on the default app context switch, MakeReadAsyncBlocking, setting it to false when the .NET runtime version was > 40702 (inherited from S.D.S netfx code). We don't need that complexity, so the app context code was cleaned up and made mostly common between netfx and netcore code.

This also brings parity to netcore for the existing MakeReadAsyncBlocking switch in netfx. It's already documented. We'll just need to update the doc  to reflect that it is now available for .NET Core and .NET Standard, too:
https://docs.microsoft.com/en-us/sql/connect/ado-net/appcontext-switches?view=sql-server-ver15


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -2911,11 +2911,11 @@ private bool TryProcessDone(SqlCommand cmd, SqlDataReader reader, ref RunBehavio
             ushort status;
             int count;
 
-            // This is added back since removing it from here introduces regressions in Managed SNI.
-            // It forces SqlDataReader.ReadAsync() method to run synchronously,
-            // and will block the calling thread until data is fed from SQL Server.
-            // TODO Investigate better solution to support non-blocking ReadAsync().
-            stateObj._syncOverAsync = true;
+            if (LocalAppContextSwitches.MakeReadAsyncBlocking)
```
cheenamalhotra, line 2914: Could you update description to enlist new AppContext switches applying to .NET Core with their default values too? We'll need that for documentation in release notes.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnectionFactory.cs
```diff
@@ -206,44 +206,18 @@ override protected DbConnectionPoolGroupOptions CreateConnectionPoolGroupOptions
...
-            }
-
-            NameValueCollection settings = (NameValueCollection)PrivilegedConfigurationManager.GetSection("Microsoft.Data.SqlClient");
-            Stream XMLStream = null;
-            if (settings != null)
-            {
-                string[] values = settings.GetValues(_metaDataXml);
-                if (values != null)
-                {
-                    XMLStream = ADP.GetXmlStreamFromValues(values, _metaDataXml);
```
cheenamalhotra, line 231: Seems like we're dropping this way of populating xmlStream.
Could you elaborate more if it could have any impact?

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/LocalDBAPI.cs
```diff
@@ -300,7 +300,7 @@ internal static void CreateLocalDBInstance(string instance)
                     if (s_configurableInstances == null)
                     {
                         Dictionary<string, InstanceInfo> tempConfigurableInstances = new Dictionary<string, InstanceInfo>(StringComparer.OrdinalIgnoreCase);
-                        object section = PrivilegedConfigurationManager.GetSection("system.data.localdb");
+                        object section = ConfigurationManager.GetSection("system.data.localdb");
```
cheenamalhotra, line 303: I'm just not sure why a privileged config manager was needed in .NET Framework, and if removing it has some consequences. Could you provide more details here?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
David-Engel: Updated.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/LocalDBAPI.cs
David-Engel: PrivilegedConfigurationManager.GetSection was simply calling ConfigurationManager.GetSection. I just cut out the middle man. I couldn't see anything in PrivilegedConfigurationManager that would make is behave differently.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnectionFactory.cs
David-Engel: The way I read the previous code was that this was attempting to read the Microsoft.Data.SqlClient.SqlMetaData.xml stuff first from the system's machine.config. Then when that failed, it would read from Microsoft.Data.SqlClient.SqlMetaData.xml. Since SqlClient is no longer shipped within the .NET Framework itself and that content will never be there, there is no need to try and read it from there.

cheenamalhotra (approved):

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlException.cs
```diff
@@ -19,14 +19,15 @@ public sealed class SqlException : System.Data.Common.DbException
     {
         private const string OriginalClientConnectionIdKey = "OriginalClientConnectionId";
         private const string RoutingDestinationKey = "RoutingDestination";
+        private const int SqlExceptionHResult = unchecked((int)0x80131904);
```
David-Engel, line 22: This was moved from HResult.cs so that HRseult.cs could be deleted. This matches the netcore code, too. No functional change.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/LocalAppContextSwitches.netfx.cs
```diff
@@ -7,38 +7,27 @@
...
         internal const string UseMinimumLoginTimeoutString = @"Switch.Microsoft.Data.SqlClient.UseOneSecFloorInTimeoutCalculationDuringLogin";
-        private static int _useMinimumLoginTimeout;
+        private static bool _useMinimumLoginTimeout;
         public static bool UseMinimumLoginTimeout
         {
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
             get
             {
-                return LocalAppContext.GetCachedSwitchValue(UseMinimumLoginTimeoutString, ref _useMinimumLoginTimeout);
+                return AppContext.TryGetSwitch(UseMinimumLoginTimeoutString, out _useMinimumLoginTimeout) ? _useMinimumLoginTimeout : false;
```
JRahnama, line 19: return AppContext.TryGetSwitch(UseMinimumLoginTimeoutString, out _useMinimumLoginTimeout) ? _useMinimumLoginTimeout : false;

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/LocalAppContextSwitches.netfx.cs
```diff
@@ -7,38 +7,27 @@
...
         internal const string DisableTNIRByDefaultString = @"Switch.Microsoft.Data.SqlClient.DisableTNIRByDefaultInConnectionString";
-        private static int _disableTNIRByDefault;
+        private static bool _disableTNIRByDefault;
         public static bool DisableTNIRByDefault
         {
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
             get
             {
-                return LocalAppContext.GetCachedSwitchValue(DisableTNIRByDefaultString, ref _disableTNIRByDefault);
+                return AppContext.TryGetSwitch(DisableTNIRByDefaultString, out _disableTNIRByDefault) ? _disableTNIRByDefault : false;
```
JRahnama, line 30: Can we change it to : return AppContext.TryGetSwitch(DisableTNIRByDefaultString, out _disableTNIRByDefault) && _disableTNIRByDefault;

JRahnama:
LGTM. Couple of minor changes requested. Thank you @David-Engel.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/LocalAppContextSwitches.netfx.cs
David-Engel: That would change the logic. If TryGetSwitch returns true, it has found the switch and set _useMinimumLoginTimeout to its value and we want to return the value of _useMinimumLoginTimeout. If TryGetSwitch returns false, no setting for the switch was found. In that case, false is the default value we want to use for that switch.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/LocalAppContextSwitches.netfx.cs
David-Engel: Same reply as above.

JRahnama (approved):

PR #853, opened 2020-12-17
Implementation of DateTime2 + accompanying GetSqlDateTime2 method (#846)

bjornbouetsmith:
Initial implementation - let me know what you think so far.
I have tried to do everything I could think of and have tried to follow the "style"
I don't know why VS insists on changing the .csproj for SqlCommand etc - I have done nothing, but will try to revert the changes out side of VS

And I need some help understanding why VS can build my stuff when I use the ADP class, but I cannot build it outside of VS - possibly some "tweaks" needed - which is why I am throwing exceptions directly and not going through the ADP class like the other classes do.

PR #912, opened 2021-02-14
Perf: Improve perf of SqlDateTimeToDateTime

Wraith2:
While reviewing https://github.com/dotnet/SqlClient/pull/853 I did some digging into the related code and profiling. The function used to convert sql date times to .net datetimes came up as hot so I tinkered with it a bit. The most important part is to avoid creating a datetime just to get a constant value from it, hardcode it instead.

Gist of the benchmark code is here: https://gist.github.com/Wraith2/d0d220f07e640c17daeb9db2d446aaa4

results: 
```
BenchmarkDotNet=v0.12.1, OS=Windows 10.0.19042
Intel Core i9-9900K CPU 3.60GHz (Coffee Lake), 1 CPU, 16 logical and 8 physical cores
.NET Core SDK=5.0.200-preview.21079.7
  [Host]     : .NET Core 3.1.12 (CoreCLR 4.700.21.6504, CoreFX 4.700.21.6905), X64 RyuJIT
  DefaultJob : .NET Core 3.1.12 (CoreCLR 4.700.21.6504, CoreFX 4.700.21.6905), X64 RyuJIT
```

|   Method |      Mean |    Error |   StdDev | Ratio |
|--------- |----------:|---------:|---------:|------:|
| Original | 271.50 ms | 4.619 ms | 4.321 ms |  1.00 |
|      New |  82.45 ms | 0.232 ms | 0.206 ms |  0.30 |

/cc @bjornbouetsmith who this may interest

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlTypes/SqlTypeWorkarounds.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlTypes/SqlTypeWorkarounds.cs
@@ -61,23 +61,25 @@ namespace Microsoft.Data.SqlTypes
             const int SQLTicksPerMinute = SQLTicksPerSecond * 60;
             const int SQLTicksPerHour = SQLTicksPerMinute * 60;
             const int SQLTicksPerDay = SQLTicksPerHour * 24;
-            const int MinDay = -53690;                // Jan 1 1753
+            //const int MinDay = -53690;                // Jan 1 1753
+            const uint MinDayOffset = 53690;            // postive value of MinDay used to pull negative values up to 0 so a single check can be used
             const uint MaxDay = 2958463;               // Dec 31 9999 is this many days from Jan 1 1900
             const uint MaxTime = SQLTicksPerDay - 1; // = 25919999,  11:59:59:997PM
-            const int MinTime = 0;                    // 00:00:0:000PM
             const long BaseDateTicks = 599266080000000000L;//new DateTime(1900, 1, 1).Ticks;
 
-            if ((uint)daypart > MaxDay || (uint)timepart > MaxTime || daypart < MinDay || timepart < MinTime)
+            // casting to uint wraps negative values to large positive ones above the valid 
+            // ranges so the lower bound doesn't need to be checked
+            if ((uint)(daypart + MinDayOffset) > (MaxDay + MinDayOffset) || (uint)timepart > MaxTime)
             {
                 ThrowOverflowException();
             }
 
             long dayticks = daypart * TimeSpan.TicksPerDay;
             double timePartPerMs = timepart / SQLTicksPerMillisecond;
-            double int1 = timePartPerMs + 0.5;
-            long timeticks1 = ((long)int1) * TimeSpan.TicksPerMillisecond;
-            long ticks1 = BaseDateTicks + dayticks + timeticks1;
-            return new DateTime(ticks1);
+            timePartPerMs += 0.5;
+            long timeTicks = ((long)timePartPerMs) * TimeSpan.TicksPerMillisecond;
+            long totalTicks = BaseDateTicks + dayticks + timeTicks;
+            return new DateTime(totalTicks);
         }
 
         private static void ThrowOverflowException()
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlTypes/SqlTypeWorkarounds.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlTypes/SqlTypeWorkarounds.cs
@@ -8,6 +8,7 @@ using System.Diagnostics;
 using System.IO;
 using System.Runtime.InteropServices;
 using System.Xml;
+using Microsoft.Data.SqlClient;
 
 namespace Microsoft.Data.SqlTypes
 {
@@ -82,10 +83,8 @@ namespace Microsoft.Data.SqlTypes
             return new DateTime(totalTicks);
         }
 
-        private static void ThrowOverflowException()
-        {
-            throw new OverflowException(SQLResource.DateTimeOverflowMessage);
-        }
+        private static Exception ThrowOverflowException() => throw SQL.DateTimeOverflow();
+
         #endregion
 
         #region Work around inability to access SqlMoney.ctor(long, int) and SqlMoney.ToSqlInternalRepresentation
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlTypes/SqlTypeWorkarounds.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlTypes/SqlTypeWorkarounds.cs
@@ -62,20 +62,27 @@ namespace Microsoft.Data.SqlTypes
             const int SQLTicksPerHour = SQLTicksPerMinute * 60;
             const int SQLTicksPerDay = SQLTicksPerHour * 24;
             const int MinDay = -53690;                // Jan 1 1753
-            const int MaxDay = 2958463;               // Dec 31 9999 is this many days from Jan 1 1900
+            const uint MaxDay = 2958463;               // Dec 31 9999 is this many days from Jan 1 1900
+            const uint MaxTime = SQLTicksPerDay - 1; // = 25919999,  11:59:59:997PM
             const int MinTime = 0;                    // 00:00:0:000PM
-            const int MaxTime = SQLTicksPerDay - 1; // = 25919999,  11:59:59:997PM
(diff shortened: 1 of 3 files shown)


cheenamalhotra:
Seems thr's a test failure you may want to take a look:

```
System.OverflowException : SqlDateTime overflow. Must be between 1/1/1753 12:00:00 AM and 12/31/9999 11:59:59 PM.
...
   at Microsoft.Data.SqlClient.ManualTesting.Tests.DateTimeTest.ReaderParameterTest() 
... ManualTests\SQL\DateTimeTest\DateTimeTest.cs:line 188

```

Wraith2:
Nice catch. Didn't see it among the other CI failures.
Reworked slightly to deal with the bounds and I've added a uint range check trick that wasn't worth using before (regressed it by 4%) but now does seem worthwhile. The range checks work by knowing that negative integers cast to extremely large positive uints so if we do zero based checks we only have to check 0 to max and not a min check.

|   Method |      Mean |    Error |   StdDev | Ratio |
|--------- |----------:|---------:|---------:|------:|
| Original | 258.05 ms | 2.267 ms | 2.010 ms |  1.00 |
|      New |  85.68 ms | 0.589 ms | 0.551 ms |  0.33 |

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlTypes/SqlTypeWorkarounds.cs
```diff
@@ -61,21 +61,30 @@ internal static DateTime SqlDateTimeToDateTime(int daypart, int timepart)
...
-            const int MaxDay = 2958463;               // Dec 31 9999 is this many days from Jan 1 1900
-            const int MinTime = 0;                    // 00:00:0:000PM
-            const int MaxTime = SQLTicksPerDay - 1; // = 25919999,  11:59:59:997PM
-
-            if (daypart < MinDay || daypart > MaxDay || timepart < MinTime || timepart > MaxTime)
+            //const int MinDay = -53690;                // Jan 1 1753
+            const uint MinDayOffset = 53690;            // postive value of MinDay used to pull negative values up to 0 so a single check can be used
+            const uint MaxDay = 2958463;               // Dec 31 9999 is this many days from Jan 1 1900
+            const uint MaxTime = SQLTicksPerDay - 1; // = 25919999,  11:59:59:997PM
+            const long BaseDateTicks = 599266080000000000L;//new DateTime(1900, 1, 1).Ticks;
```
cheenamalhotra, line 68: We should consider moving these magic numbers to a common location that both NetFx and NetCore can take benefit from, maybe a static constant class?

cheenamalhotra (approved):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlTypes/SqlTypeWorkarounds.cs
Wraith2: The overall merge process for the two projects should do that over time.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlTypes/SqlTypeWorkarounds.cs
```diff
@@ -61,21 +61,30 @@ internal static DateTime SqlDateTimeToDateTime(int daypart, int timepart)
...
+            long timeTicks = ((long)timePartPerMs) * TimeSpan.TicksPerMillisecond;
+            long totalTicks = BaseDateTicks + dayticks + timeTicks;
+            return new DateTime(totalTicks);
+        }
 
-            return new DateTime(baseDateTicks + dayticks + timeticks);
+        private static void ThrowOverflowException()
+        {
+            throw new OverflowException(SQLResource.DateTimeOverflowMessage);
         }
```
DavoudEshtehari, line 88: Could you use utility class to gather the exception message, please?

Wraith2: Any particular reason? and any particular place you want it to be?

DavoudEshtehari: To prevent redundancy and have better management over them. You can use the `SqlUtil.cs` file in the common source folder.
FYI, we already have three `SqlUtil.cs` files, one for each project plus what I mentioned.

Wraith2: Done. I'm not convinced that it's a great idea because it seems to be blurring the distinction between a couple of namespaces. It's also important for the codegen that the exception be thrown from a helper method so it just adds another level of indirection that doesn't seem useful.

DavoudEshtehari: It seems you included it in the wrong file. As I mentioned, there are three **SqlUtil.cs** files that one of them is in the common folder to use for both projects that it's a better choice for your case. _You can find it if you pull the latest source code._

Also, we usually create a specific new exception object and use it where it's required to throw (don't throw an exception in **SqlUtil**). I don't think you need the private local static method (**ThrowOverflowException**), you can throw `SQL.DateTimeOverflow()` where you need it.

I found `SqlUtil.UDTInvalidSize()` method has the same problem; could you modify it in the same way, please?

About the points, you mentioned we can discuss more in a separate session to find a better way. Until then, keeping the source unify prevents ambiguity.

Wraith2: > I don't think you need the private local static method (ThrowOverflowException), you can throw SQL.DateTimeOverflow() where you need it.

No, I do need it because throwing inline has a significant performance penalty. Use sharplab or disasmo and check the tier1 assembly that's generated. A call to a throw helper function is faster than doing it inline. Or spin up a bdn benchmark like I did and see what changes make a difference. This one is the best I can get it, even tiny changes have big differences.

There really shouldn't be multiple SqlUtil files and whatever they contain should be split out into correctly named files. I can move around the things for this PR but other changes will have to be done in the larger merge process or in individual PR's.

DavoudEshtehari: Since we still don't have a perf lab, clarify your reasons by adding some notes to avoid a performance regression hereafter, please.

Wraith2: NoInlining hint added (though it's currently not necessary) and a comment about why the method is extracted from it's caller added for future users.

Wraith2: I've also added the DateTimeOverflow method in netfx. In this source there are only two SqlUtil files there aren't conflicts. I have seen that in the main branch there is a shared one but it also comes with a compiler warning that indicates one of them is being ignored, so I don't think that's doing whatever it was intended to do.

DavoudEshtehari: Consider the difference between my machine configuration and Warith in result values.
Don't think I'm jealous of you @Wraith2 at all. 😉 

``` ini

BenchmarkDotNet=v0.12.1, OS=Windows 10.0.19042
Intel Core i7-8665U CPU 1.90GHz (Coffee Lake), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=5.0.201
  [Host]     : .NET Core 3.1.13 (CoreCLR 4.700.21.11102, CoreFX 4.700.21.11602), X64 RyuJIT
  DefaultJob : .NET Core 3.1.13 (CoreCLR 4.700.21.11102, CoreFX 4.700.21.11602), X64 RyuJIT


```
|                  Method |     Mean |    Error |   StdDev | Ratio |
|------------------------ |---------:|---------:|---------:|------:|
|                Original | 481.6 ms | 12.74 ms | 36.76 ms |  1.00 |
|                     New | 130.2 ms |  1.76 ms |  1.81 ms |  0.27 |
| NewUtilPrivateException | 136.0 ms |  2.95 ms |  8.22 ms |  0.28 |
|  NewUtilInlineException | 145.1 ms |  3.28 ms |  9.47 ms |  0.30 |

- The benchmark code based on Wraith's code:  https://gist.github.com/DavoudEshtehari/267fb6cc6dd09475f51a41ed48364f17

DavoudEshtehari (approved):

Issue #414, opened 2019-11-25
Cannot unload a collectible AssemblyLoadContext if System.Data.SqlClient is used


stasgaonkar:
The Assembly we want to load (and unload) needs the ability to connect to a database for which we use Dapper. Dapper internally makes use of System.Data.SqlClient. We find that when we create such an assembly, then we are unable to Unload the **collectible AssemblyLoadContext**.

The sample source code is available at: https://github.com/stasgaonkar/DynamicLoading

Having followed the steps mentioned in https://docs.microsoft.com/en-us/dotnet/standard/assembly/unloadability#, the following is the output of `gcroot` command,

```
0:000> !dumpheap -type LoaderAllocator
         Address               MT     Size
000002022c46e1f0 00007fff74999410       48     
000002022c46e260 00007fff74999538       24     

Statistics:
              MT    Count    TotalSize Class Name
00007fff74999538        1           24 System.Reflection.LoaderAllocatorScout
00007fff74999410        1           48 System.Reflection.LoaderAllocator
Total 2 objects
0:000> !gcroot -all 0x000002022c46e1f0
HandleTable:
    000002022A9D12A0 (strong handle)
    -> 000002022C4E8C48 System.Data.SqlClient.TdsParserStateObjectNative
    -> 000002022C46E1F0 System.Reflection.LoaderAllocator

    000002022A9D12A8 (strong handle)
    -> 000002022C4E00F0 System.Data.SqlClient.TdsParserStateObjectNative
    -> 000002022C46E1F0 System.Reflection.LoaderAllocator

    000002022A9D12E0 (strong handle)
    -> 000002022C4B38C0 System.Data.SqlClient.TdsParserStateObjectNative
    -> 000002022C46E1F0 System.Reflection.LoaderAllocator

    000002022A9D15F8 (pinned handle)
    -> 000002023C461038 System.Object[]
    -> 000002022C475768 System.Threading.TimerQueue[]
    -> 000002022C475840 System.Threading.TimerQueue
    -> 000002022C4ABE88 System.Threading.TimerQueueTimer
    -> 000002022C475708 System.Threading.TimerQueueTimer
    -> 000002022C475580 System.Threading.TimerCallback
    -> 000002022C474D60 System.Data.SqlClient.SqlConnectionFactory
    -> 000002022C46E1F0 System.Reflection.LoaderAllocator

Found 4 roots.
```

[...]

PR #913, opened 2021-02-15
[.NET Core] Support Assembly Context Unloading

Wraith2:
Initial changes required to get same app from https://github.com/dotnet/SqlClient/issues/414 unloading successfully.
Then the .net 5 build is ready it will need to take account of the ALC files to make sure they are included in 5 and later builds which use ALC.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnectionFactory.AssemblyLoadContext.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnectionFactory.AssemblyLoadContext.cs
@@ -1,4 +1,8 @@
-﻿using System;
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
 using System.Reflection;
 using System.Runtime.Loader;
 
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnectionFactory.AssemblyLoadContext.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnectionFactory.AssemblyLoadContext.cs
@@ -1,0 +1,19 @@
+﻿using System;
+using System.Reflection;
+using System.Runtime.Loader;
+
+namespace Microsoft.Data.SqlClient
+{
+    sealed internal partial class SqlConnectionFactory
+    { 
+        partial void SubscribeToAssemblyLoadContextUnload()
+        {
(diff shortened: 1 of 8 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDiagnosticListener.NetCoreApp.cs
```diff
@@ -3,13 +3,21 @@
...
     internal sealed class SqlDiagnosticListener : DiagnosticListener
     {
         public SqlDiagnosticListener(string name) : base(name)
         {
+            AssemblyLoadContext.GetLoadContext(Assembly.GetExecutingAssembly()).Unloading += SqlDiagnosticListener_Unloading;
+        }
+
+        private void SqlDiagnosticListener_Unloading(AssemblyLoadContext obj)
+        {
+            Dispose();
```
JRahnama, line 20: Just a question here to share some thoughts, it might not be even related to this change but I think it is good to have this in mind when implementing `IDisposable` interface : By [Microsoft Best practices documentation](https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose#the-dispose-method) should we consider having a better control over Dispose method? the `bool` value passed to the overridden method will determine about disposing `Managed` and `Unmanaged` resource. I have seen the pattern has been used in some class in our driver, but is not consistent. It definitely needs more thought and testing.

cheenamalhotra (commented):
LGTM

cheenamalhotra (approved):

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDiagnosticListener.NetCoreApp.cs
Wraith2: These classes have no finalizers and hold no unmanaged resources so the distinction between disposing from the finalizer or dispose methods isn't relevant. If we were talking about one of the SafeHandle types then it would be sensible to use the bool parameter pattern. In this case there is no need for the method to even be called dispose or implement IDisposable because we're doing cleanup directly and not with using blocks or cascaded dispose calls. I followed the convention that was there because the method name Dispose has become synonymous with cleanup/finalize semantics for most .net developers and it's a reasonable name to use.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnectionFactory.AssemblyLoadContext.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
```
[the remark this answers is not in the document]
Wraith2: Oops, done.

johnnypham (approved):

Issue #672, opened 2020-07-29
SqlDataReader.DisposeAsync throw SqlException Execution: Timeout Expired.  The timeout period elapsed prior to completion of the operation or the server is not responding


LIFEfreedom:
### Describe the bug
I got an exception when trying to call DisposeAsync on a SqlDataReader

```
Unhandled exception. Microsoft.Data.SqlClient.SqlException (0x80131904): Execution Timeout Expired.  The timeout period elapsed prior to completion of the operation or the server is not responding.
 ---> System.ComponentModel.Win32Exception (258): Unknown error 258
   at Microsoft.Data.SqlClient.SqlInternalConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   at Microsoft.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObject stateObj, Boolean callerHasConnectionLock, Boolean asyncClose)
   at Microsoft.Data.SqlClient.TdsParserStateObject.ThrowExceptionAndWarning(Boolean callerHasConnectionLock, Boolean asyncClose)
   at Microsoft.Data.SqlClient.TdsParserStateObject.ReadSniError(TdsParserStateObject stateObj, UInt32 error)
   at Microsoft.Data.SqlClient.TdsParserStateObject.ReadSniSyncOverAsync()
   at Microsoft.Data.SqlClient.TdsParserStateObject.TryReadNetworkPacket()
   at Microsoft.Data.SqlClient.TdsParserStateObject.TryPrepareBuffer()
   at Microsoft.Data.SqlClient.TdsParserStateObject.TryReadByteArray(Span`1 buff, Int32 len, Int32& totalRead)
   at Microsoft.Data.SqlClient.TdsParser.TrySkipValue(SqlMetaDataPriv md, Int32 columnOrdinal, TdsParserStateObject stateObj)
   at Microsoft.Data.SqlClient.TdsParser.TrySkipRow(_SqlMetaDataSet columns, Int32 startCol, TdsParserStateObject stateObj)
   at Microsoft.Data.SqlClient.TdsParser.TryRun(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj, Boolean& dataReady)
   at Microsoft.Data.SqlClient.SqlDataReader.TryCloseInternal(Boolean closeReader)
[...]

PR #920, opened 2021-02-18
FIX | Throwing exception from Dispose should not happen

JRahnama:
Addressing Issue #672, throwing exception from Dispose should not happen. 
`TryCloseInternal` passes an unhandled exception to `Close` function and the exception comes up from `Dispose` method. Adding the `try/catch` block  will fix the issue.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
@@ -844,7 +844,7 @@ namespace Microsoft.Data.SqlClient
             }
             catch(SqlException ex)
             {
-                SqlClientEventSource.Log.TryTraceEvent("SqlDataReader.Dispose | ERR | Error: {0}", ex);
+                SqlClientEventSource.Log.TryTraceEvent("SqlDataReader.Dispose | ERR | Error Message: {0}, Stack Trace: {1}", ex.Message, ex.StackTrace);
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
```diff
@@ -834,11 +834,18 @@ private void CleanPartialReadReliable()
...
-                Close();
+                if (disposing)
+                {
+                    Close();
+                }
+                base.Dispose(disposing);
+            }
+            catch(SqlException ex)
+            {
+                SqlClientEventSource.Log.TryTraceEvent("SqlDataReader.Dispose | ERR | Error: {0}", ex);
```
cheenamalhotra, line 847: Do we want to this catching in Close instead? That way it will be in parity with .NET Framework, and that's the only place any driver errors will be caught.

Also make changes in .NET Framework (SqlDataReader.Close()) method.

cheenamalhotra, line 847: Log ex.Message and (optionally) stacktrace.

David-Engel: That's what his original change did. But we don't want to change the behavior or Close(), just Dispose(). Dispose() is what carries the best practice that it should not throw. Also, if we want parity with netfx, we would bring the Dispose() method with its try/catch to netfx and leave Close() as-is.

JRahnama: I have added Dispose() with try/catch block to netfx for parity as @David-Engel suggested.

JRahnama: done.

cheenamalhotra (approved):

David-Engel (approved):

PR #925, opened 2021-02-21
Clean up AAsyncCallContext and SqlDataReader

Wraith2:
Looking though https://github.com/dotnet/SqlClient/pull/921 I realized that I hadn't got around to merging the SqlDataReader.AAsyncCallContext class with the top level one that I later derived from it. So I did that. I've cleaned up the inheritance and made the relationships between the classes and the functions that use them a little clearer with some comments so hopefully others will find it less confusing.

There's a small change in GetFieldValueAsync to use a static delegate that already existed rather than allocating a lambda. This will have a minor memory improvement.

I'm still looking for a good way of caching the async call context for GetFieldValueAsync but that will wait for a future PR.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
@@ -4298,16 +4298,16 @@ namespace Microsoft.Data.SqlClient
             if (task != null)
             {
                 SqlClientEventSource.Log.TryTraceEvent("SqlDataReader.NextResultAsyncExecute | attempt retry {0}", ObjectID);
-                context._reader.PrepareForAsyncContinuation();
+                context.Reader.PrepareForAsyncContinuation();
             }
 
-            if (context._reader.TryNextResult(out bool more))
+            if (context.Reader.TryNextResult(out bool more))
             {
                 // completed
                 return more ? ADP.TrueTask : ADP.FalseTask;
             }
 
-            return context._reader.ExecuteAsyncCall(context);
+            return context.Reader.ExecuteAsyncCall(context);
         }
 
         // NOTE: This will return null if it completed sequentially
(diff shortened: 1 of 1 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
```diff
@@ -5009,19 +5009,22 @@ override public Task<T> GetFieldValueAsync<T>(int i, CancellationToken cancellat
             IDisposable registration = null;
             if (cancellationToken.CanBeCanceled)
             {
-                registration = cancellationToken.Register(s => ((SqlCommand)s).CancelIgnoreFailure(), _command);
+                registration = cancellationToken.Register(SqlCommand.s_cancelIgnoreFailure, _command);
```
JRahnama, line 5012: Can we summarize it in a single line? something like:
`IDisposable registration = cancellationToken.CanBeCanceled? cancellationToken.Register(SqlCommand.s_cancelIgnoreFailure, _command) : (IDisposable) null;`

Wraith2: We could but there is no reason to do so. It adds a useless null cast and hides the optionality of the statement by making it look like a single assignment when it's optional not a choice between two values.

DavoudEshtehari (approved):
The class definition is in contrast with [document](https://docs.microsoft.com/en-us/dotnet/api/microsoft.data.sqlclient.sqldatareader?view=sqlclient-dotnet-standard-3.0). According to the design, my guess is the document should update.

CC @cheenamalhotra

Wraith2:
About inheritance? Yes, it does seem to say that it can't be inherited. In this case since we don't intend to provide any kind of extensibility and users could serious screw up their data or programs by tinkering with things that they shouldn't I think sealing it might be a good idea. It is a breaking change though.

johnnypham (approved):

cheenamalhotra:
> The class definition is in contrast with document. According to the design, my guess is the document should update.

@DavoudEshtehari We can make a separate PR to address this.

Issue #926, opened 2021-02-23
Kerberos authentication doesn't work with NuGet package 2.1.1


Code-DJ:
Can't authenticate with kerberos 2.1.1. Works fine with 2.0.1

Following is the error I get with Azure Data Studio where they upgraded their reference from older version to 2.1.1 in ADS 1.26.0:

```
Microsoft.Data.SqlClient.SqlException (0x80131904): Cannot authenticate using Kerberos. Ensure Kerberos has been initialized on the client with 'kinit' and a Service Principal Name has been registered for the SQL Server to allow Kerberos authentication.
ErrorCode=InternalError, Exception=Interop+NetSecurityNative+GssApiException: GSSAPI operation failed with error -  An unsupported mechanism was requested (unknown mech-code 0 for mech unknown).
   at System.Net.Security.NegotiateStreamPal.GssInitSecurityContext(SafeGssContextHandle& context, SafeGssCredHandle credential, Boolean isNtlm, SafeGssNameHandle targetName, GssFlags inFlags, Byte[] buffer, Byte[]& outputBuffer, UInt32& outFlags, Int32& isNtlmUsed)
   at System.Net.Security.NegotiateStreamPal.EstablishSecurityContext(SafeFreeNegoCredentials credential, SafeDeleteContext& context, String targetName, ContextFlagsPal inFlags, SecurityBuffer inputBuffer, SecurityBuffer outputBuffer, ContextFlagsPal& outFlags)
   at Microsoft.Data.SqlClient.SNI.SNIProxy.GenSspiClientContext(SspiClientContextStatus sspiClientContextStatus, Byte[] receivedBuff, Byte[]& sendBuff, Byte[] serverName)
   at Microsoft.Data.SqlClient.SNI.TdsParserStateObjectManaged.GenerateSspiClientContext(Byte[] receivedBuff, UInt32 receivedLength, Byte[]& sendBuff, UInt32& sendLength, Byte[] _sniSpnBuffer)
   at Microsoft.Data.SqlClient.TdsParser.SNISSPIData(Byte[] receivedBuff, UInt32 receivedLength, Byte[]& sendBuff, UInt32& sendLength)
   at Microsoft.Data.SqlClient.SqlInternalConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
[...]

PR #930, opened 2021-02-25
Fix | Fixes Kerberos auth when SPN does not contain port

cheenamalhotra:
This change shall allow both with and without port SPNs to be accepted by default when no port specified.
Addresses #926


src/Microsoft.Data.SqlClient/netcore/src/Common/src/System/Net/Security/NegotiateStreamPal.Unix.cs
```diff
@@ -135,15 +135,16 @@ internal static partial class NegotiateStreamPal
             }
             catch (Exception ex)
             {
-                if (NetEventSource.IsEnabled) NetEventSource.Error(null, ex);
+                if (NetEventSource.IsEnabled)
```
karinazhou, line 138: Can we add the `{ }` to make it easier to read?

src/Microsoft.Data.SqlClient/netcore/src/Common/src/System/Net/Security/NegotiateStreamPal.Unix.cs
```diff
@@ -156,20 +157,29 @@ internal static partial class NegotiateStreamPal
...
+                    spn,
+                    requestedContextFlags,
+                    ((inSecurityBufferArray != null && inSecurityBufferArray.Length != 0) ? inSecurityBufferArray[0] : null),
+                    outSecurityBuffer,
+                    ref contextFlags);
+
+                if (status.ErrorCode != SecurityStatusPalErrorCode.InternalError)
+                    break;
+                else 
+                    securityContext = null; // Reset security context to be generated again
```
karinazhou, line 181: This securityContext is passed as reference from caller. Can we save its initial status and assign it back if we need to try another SPN?

cheenamalhotra: It's reset to null when trying another SPN, so it can be generated again for next SPN. Old one failed with error, hence not needed to be saved.

src/Microsoft.Data.SqlClient/netcore/src/Common/src/System/Net/Security/NegotiateStreamPal.Unix.cs
```diff
@@ -156,20 +159,33 @@ internal static partial class NegotiateStreamPal
...
+                    outSecurityBuffer,
+                    ref contextFlags);
+
+                if (status.ErrorCode != SecurityStatusPalErrorCode.InternalError)
+                {
+                    break; // Successful case, exit the loop with current SPN.
+                }
+                else
+                {
+                    securityContext = null; // Reset security context to be generated again for next SPN.
```
karinazhou, line 186: `securityContext` will be set to null even if we are not going to retry with a different spn if `SecurityStatusPalErrorCode.InternalError` is returned from `EstablishSecurityContext()`. That null will be returned back to the caller. This is kind of changing the initial behavior when we have only 1 SPN which fails with InternalError. 
Can we only set `securityContext = null` when we need an additional try in the for loop?

cheenamalhotra: I thought about it - this is an error event - we eitherway don't need securityContext anymore.
We're going to throw exception in that case.

David-Engel (approved):

karinazhou (approved):

PR #933, opened 2021-02-26
Rework MARSConnection state machine

Wraith2:
The MARSConnection class uses a confusing implicit state machine in HandleRecieveComplete to demux incoming packets and dispatch them to the correct MARSHandle instances. This PR reworks that state machine to be explicit allowing easier visualization of exactly what is happening as packets travel through it. It also allows a small optimization to be added which is where a packet header is decoded and the rest of the packet contents is exactly the payload it will be forwarded rather than reconstructed, this cuts down on copies allowing faster running.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNICommon.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNICommon.cs
@@ -90,8 +90,7 @@ namespace Microsoft.Data.SqlClient.SNI
     /// <summary>
     /// SMUX packet flags
     /// </summary>
-    [Flags]
-    internal enum SNISMUXFlags
+    internal enum SNISMUXFlags : uint
     {
         SMUX_SYN = 1,       // Begin SMUX connection
         SMUX_ACK = 2,       // Acknowledge SMUX packets
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNICommon.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNICommon.cs
@@ -35,30 +35,40 @@ namespace Microsoft.Data.SqlClient.SNI
     /// <summary>
     /// SMUX packet header
     /// </summary>
-    internal sealed class SNISMUXHeader
+    internal struct SNISMUXHeader
     {
         public const int HEADER_LENGTH = 16;
 
-        public byte SMID;
-        public byte flags;
-        public ushort sessionId;
-        public uint length;
-        public uint sequenceNumber;
-        public uint highwater;
+        public byte SMID { get; private set; }
+        public byte Flags { get; private set; }
+        public ushort SessionId { get; private set; }
+        public uint Length { get; private set; }
+        public uint SequenceNumber { get; private set; }
+        public uint Highwater { get; private set; }
+
+        public void Set(byte smid, byte flags, ushort sessionID, uint length, uint sequenceNumber, uint highwater)
+        {
+            SMID = smid;
+            Flags = flags;
+            SessionId = sessionID;
+            Length = length;
+            SequenceNumber = sequenceNumber;
+            Highwater = highwater;
+        }
 
         public void Read(byte[] bytes)
         {
             SMID = bytes[0];
-            flags = bytes[1];
-            sessionId = BitConverter.ToUInt16(bytes, 2);
-            length = BitConverter.ToUInt32(bytes, 4) - SNISMUXHeader.HEADER_LENGTH;
-            sequenceNumber = BitConverter.ToUInt32(bytes, 8);
-            highwater = BitConverter.ToUInt32(bytes, 12);
+            Flags = bytes[1];
+            SessionId = BitConverter.ToUInt16(bytes, 2);
+            Length = BitConverter.ToUInt32(bytes, 4) - SNISMUXHeader.HEADER_LENGTH;
+            SequenceNumber = BitConverter.ToUInt32(bytes, 8);
+            Highwater = BitConverter.ToUInt32(bytes, 12);
         }
 
         public void Write(Span<byte> bytes)
         {
-            uint value = highwater;
+            uint value = Highwater;
             // access the highest element first to cause the largest range check in the jit, then fill in the rest of the value and carry on as normal
             bytes[15] = (byte)((value >> 24) & 0xff);
             bytes[12] = (byte)(value & 0xff); // BitConverter.GetBytes(_currentHeader.highwater).CopyTo(headerBytes, 12);
@@ -66,24 +76,46 @@ namespace Microsoft.Data.SqlClient.SNI
             bytes[14] = (byte)((value >> 16) & 0xff);
 
             bytes[0] = SMID; // BitConverter.GetBytes(_currentHeader.SMID).CopyTo(headerBytes, 0);
-            bytes[1] = flags; // BitConverter.GetBytes(_currentHeader.flags).CopyTo(headerBytes, 1);
+            bytes[1] = Flags; // BitConverter.GetBytes(_currentHeader.flags).CopyTo(headerBytes, 1);
 
-            value = sessionId;
+            value = SessionId;
             bytes[2] = (byte)(value & 0xff); // BitConverter.GetBytes(_currentHeader.sessionId).CopyTo(headerBytes, 2);
             bytes[3] = (byte)((value >> 8) & 0xff);
 
-            value = length;
+            value = Length;
             bytes[4] = (byte)(value & 0xff); // BitConverter.GetBytes(_currentHeader.length).CopyTo(headerBytes, 4);
             bytes[5] = (byte)((value >> 8) & 0xff);
             bytes[6] = (byte)((value >> 16) & 0xff);
             bytes[7] = (byte)((value >> 24) & 0xff);
 
-            value = sequenceNumber;
+            value = SequenceNumber;
             bytes[8] = (byte)(value & 0xff); // BitConverter.GetBytes(_currentHeader.sequenceNumber).CopyTo(headerBytes, 8);
             bytes[9] = (byte)((value >> 8) & 0xff);
             bytes[10] = (byte)((value >> 16) & 0xff);
             bytes[11] = (byte)((value >> 24) & 0xff);
 
+        }
+
+        public SNISMUXHeader Clone()
+        {
+            SNISMUXHeader copy = new SNISMUXHeader();
+            copy.SMID = SMID;
+            copy.Flags = Flags;
+            copy.SessionId = SessionId;
+            copy.Length = Length;
+            copy.SequenceNumber = SequenceNumber;
+            copy.Highwater = Highwater;
+            return copy;
+        }
+
+        public void Clear()
+        {
+            SMID = 0;
+            Flags = 0;
+            SessionId = 0;
+            Length = 0;
+            SequenceNumber = 0;
+            Highwater = 0;
         }
     }
 
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIMarsConnection.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIMarsConnection.cs
@@ -245,7 +245,7 @@ namespace Microsoft.Data.SqlClient.SNI
             Demux,
             HandleAck,
             HandleData,
-            Recieve,
+            Receive,
             Finish,
             Error
         }
@@ -312,7 +312,7 @@ namespace Microsoft.Data.SqlClient.SNI
                                         }
                                         else
                                         {
-                                            state = State.Recieve;
+                                            state = State.Receive;
                                         }
                                         break;
 
@@ -346,14 +346,14 @@ namespace Microsoft.Data.SqlClient.SNI
                                             {
                                                 // payload is complete so dispatch the current packet
                                                 _demuxState = DemuxState.Dispatch;
-                                                state = State.Recieve;
+                                                state = State.Receive;
                                                 goto case DemuxState.Dispatch;
                                             }
                                             else if (packet.DataLeft == 0)
                                             {
                                                 // no more data in the delivered packet so wait for a new one
                                                 _demuxState = DemuxState.Payload;
-                                                state = State.Recieve;
+                                                state = State.Receive;
                                             }
                                             else
                                             {
@@ -388,7 +388,7 @@ namespace Microsoft.Data.SqlClient.SNI
                                                     }
                                                     else
                                                     {
-                                                        nextState = State.Recieve;
+                                                        nextState = State.Receive;
                                                     }
                                                     break;
 
@@ -407,7 +407,7 @@ namespace Microsoft.Data.SqlClient.SNI
                                                     }
                                                     else
                                                     {
-                                                        nextState = State.Recieve;
+                                                        nextState = State.Receive;
                                                     }
                                                     break;
 
@@ -438,7 +438,7 @@ namespace Microsoft.Data.SqlClient.SNI
                                                 {
                                                     SqlClientEventSource.Log.TrySNITraceEvent(nameof(SNIMarsConnection), EventType.INFO, "MARS Session Id {0}, run out of data , queuing receive", args0: _lowerHandle?.ConnectionId, args1: _header.SessionId);
                                                 }
-                                                state = State.Recieve;
+                                                state = State.Receive;
                                             }
 
                                         }
@@ -500,7 +500,7 @@ namespace Microsoft.Data.SqlClient.SNI
                             nextState = State.Finish;
                             break;
 
-                        case State.Recieve:
+                        case State.Receive:
                             if (packet != null)
                             {
                                 Debug.Assert(packet.DataLeft == 0, "loop exit with data remaining");
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIMarsConnection.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIMarsConnection.cs
@@ -14,6 +14,7 @@ namespace Microsoft.Data.SqlClient.SNI
     /// </summary>
     internal class SNIMarsConnection
     {
+        private readonly object _sync;
         private readonly Guid _connectionId;
         private readonly Dictionary<int, SNIMarsHandle> _sessions;
         private SNIHandle _lowerHandle;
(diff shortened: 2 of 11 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
```diff
@@ -2519,6 +2519,11 @@ internal void ReadSni(TaskCompletionSource<object> completion)
                     Debug.Assert(IsValidPacket(readPacket), "ReadNetworkPacket should not have been null on this async operation!");
                     // Evaluate this condition for MANAGED_SNI. This may not be needed because the network call is happening Async and only the callback can receive a success.
                     ReadAsyncCallback(IntPtr.Zero, readPacket, 0);
+
+                    if (!IsPacketEmpty(readPacket))
+                    {
+                        ReleasePacket(readPacket);
+                    }
```
Wraith2, line 2526: This looks redundant because of the check after this block of code for ManagedSNI before doing a packet release but it isn't. I tried putting just one release in place but any other combination causes a leak either in managed or native.

In Managed mode we have to release the packet only if there was a synchronous result. 
In Native mode we have to release the packet in if we received one.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNICommon.cs
```diff
@@ -90,8 +90,7 @@ public void Write(Span<byte> bytes)
     /// <summary>
     /// SMUX packet flags
     /// </summary>
-    [Flags]
```
Wraith2, line 93: The MC-SMP standard states that these cannot be combined. They aren't being combined in the code but I chose to prevent possible future confusion and remove the attribute.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIMarsConnection.cs
```diff
@@ -219,228 +192,250 @@ public void HandleSendComplete(SNIPacket packet, uint sniErrorCode)
...
 
-                                if (bytesTaken == 0)
+                            case State.Payload:
+                                if (packet.DataLeft == _payloadLength && _partial == null)
                                 {
-                                    sniErrorCode = ReceiveAsync(ref packet);
-                                    SqlClientEventSource.Log.TrySNITraceEvent(s_className, EventType.INFO, "MARS Session Id {0}, Non-SMUX Header SNI Packet received with code {1}", args0: ConnectionId, args1: sniErrorCode);
+                                    // if the data in the packet being processed is exactly and only the data that is going to sent
+                                    // on to the parser then don't copy it to a new packet just forward the current packet once we've
+                                    // fiddled the data pointer so that it skips the header data when
```
David-Engel, line 309: Don't leave me in suspense. When what?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIMarsConnection.cs
```diff
@@ -219,228 +192,250 @@ public void HandleSendComplete(SNIPacket packet, uint sniErrorCode)
...
+                                            Debug.Fail("unknown smux packet flag");
+                                            break;
                                     }
 
+                                    // partial packet has been fully reconstructed and dispatched so change to header state 
+                                    // to be ready to decode another smux header
+                                    _headerCount = 0;
+                                    _state = State.Header;
+
+                                    if (packet==null || packet.DataLeft == 0)
```
David-Engel, line 391:
```suggestion
                                    if (packet == null || packet.DataLeft == 0)
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIPacket.cs
```diff
@@ -357,5 +309,33 @@ public async void WriteToStreamAsync(Stream stream, SNIAsyncCallback callback, S
...
+            int copyLength = Math.Min(Math.Min(sourceBuffer.Count, targetBuffer.Count), maximumLength);
+
+            Buffer.BlockCopy(sourceBuffer.Array, sourceBuffer.Offset, targetBuffer.Array, targetBuffer.Offset, copyLength);
+
+            source._dataOffset += copyLength;
+            target._dataLength += copyLength;
+
+            return copyLength;
+        }
+
```
David-Engel, line 339: Nit
```suggestion
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIPhysicalHandle.cs
```diff
@@ -42,53 +42,33 @@ public override SNIPacket RentPacket(int headerSize, int dataSize)
...
             Interlocked.Add(ref packet._refCount, -1);
             packet._traceTag = null;
-#if TRACE_HISTORY
-            if (packet._history != null)
-            {
-                packet._history.Add(new SNIPacket.History { Action = SNIPacket.History.Direction.Return, Stack = GetStackParts(), RefCount = packet._refCount });
-            }
-#endif
+            packet.AddHistory(false);
+            //packet.PushPath(new StackTrace().ToString());
```
David-Engel, line 68: Remove?

David-Engel (commented):
I love it. It's definitely an improvement over the already-complex MARS logic. In order to get people comfortable with this change, can you list the tests that already cover the impacted functionality? Or maybe collect code coverage that shows all the changed code is covered? Also, are there any before/after perf changes?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIPhysicalHandle.cs
Wraith2: Yes, I'd left it as an example but I've moved it to the comment in the packet.debug file instead.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIPacket.cs
Wraith2: Done.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIMarsConnection.cs
Wraith2: Done. I thought the editorconfig file used to catch this when auto formatting but it doesn't seem to anymore.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIMarsConnection.cs
Wraith2: When you know, the thing, that thing, happens....
I think originally it was that that partial packet was null but it's just a trailing off now. Removed, The rest of the comment makes sense without that word I think but I'll left you judge if the explanations make sense.

Wraith2:
For coverage I'm relying on the standard tests. I rewrote and then ran all the functional and manual tests with forced managed mode. The first few times I found some problems and resolved them so it's certainly being used. I also had to track individual packet lifetimes as you can tell from the code I needed to add into Packet.Debug.cs and for that i used a version of the repro from 659 (which was where @cheenamalhotra noticed that there were packets being dropped to the GC) so I have run that for a while while exercises the single packet passthrough because they're small packets.

I haven't tried to benchmark this because the main purpose was correctness and code clarity but it's a good point, I should see if anything interesting has changed. We should gain some memory back from recycling and a tiny amount of cpu on single packet reads.

Wraith2:
I tried to get some numbers and it looked like there might be a slight regression but I've found that there's also a very subtle locking issue as well that only seems to show under high load. I'm going to see what I can do to track it down but until I do please consider this too dangerous to merge.

Wraith2:
I've got a cleanly running version now which has changed the state machine to be more complicated. It has to do this to follow the exact and important locking scenario. We have to lock while demuxing but must not lock while dispatching to the lower handle, this was implicit in the code as written before and now I understand it I've put code checks in place as well as a few comments.

The SMUXHeader has been changed from a class into a struct because I needed to be able to take an allocation free clone of it. It's setup in the lock and must be used outside so a copy is required. There have been some renames and scope changes to make it two nested state switches.

I've been unable to find any noticeable performance difference between master and this PR but as I said earlier perf wasn't an objective. The perf improvements are long-term amortization of the packet arrays and in standard operations these happened enough to be noticeable under the debugger but not enough to cause an obvious degradation in throughput. 

Some perf numbers from the DataAccessPerformance project which implements the techempower Fortunes benchmark which is a raw throghput measurement:
```
pr: ado-sqlclient async 16 Threads, tps: 20348.00, stddev(w/o best+worst): 1267.49
main: ado-sqlclient async 16 Threads, tps: 20415.00, stddev(w/o best+worst): 1225.99
```
Minor probably noise differences. I can't pick up anything interesting in microbenchmarks because they won't GC enough.

So my the question is whether the code is easier to understand and maintain than the previous version. I can't say I'm happy about how complicated or how much new code there is. I would have liked to keep it a lot shorter. I do think that it has value for documenting the flow of data through the method and adds some important information about what locks need to be taken. Do the team think it's an improvement or adds enough value to be used?

Wraith2:
Incidentally eiriktsarpalis posted a really good example of why I choose to use inline typeof(T) instead of put it into a variable https://github.com/dotnet/runtime/pull/46414#discussion_r589728897

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIMarsHandle.cs
```diff
@@ -55,6 +55,11 @@ internal sealed class SNIMarsHandle : SNIHandle
         /// </summary>
         public override void Dispose()
         {
+            // SendControlPacket will lock so make sure that it cannot deadlock by failing to enter the DemuxerLock
+            if (_connection != null && Monitor.IsEntered(_connection.DemuxerSync))
+            {
+                throw new InvalidOperationException("SNIMarsHandle.HandleRecieveComplete should be be called while holding the SNIMarsConnection.DemuxerSync because it can cause deadlocks");
```
JRahnama, line 61: Do we need to clear the connection here before throwing an exception?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIMarsHandle.cs
```diff
@@ -449,15 +468,19 @@ public void HandleAck(uint highwater)
...
         public void HandleReceiveComplete(SNIPacket packet, SNISMUXHeader header)
         {
+            if (Monitor.IsEntered(_connection.DemuxerSync))
+            {
+                throw new InvalidOperationException("SNIMarsHandle.HandleRecieveComplete should be be called while holding the SNIMarsConnection.DemuxerSync because it can cause deadlocks");
+            }
             long scopeID = SqlClientEventSource.Log.TrySNIScopeEnterEvent(s_className);
             try
             {
                 lock (this)
```
JRahnama, line 478: not related to your changes, but in general `lock(this)` is not a proper way of locking. When we use `This` unrelated areas of the code/driver are using the same object for locking. Should not we have a lock object in this situation or similar situations?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIMarsHandle.cs
```diff
@@ -26,7 +26,7 @@ internal sealed class SNIMarsHandle : SNIHandle
         private readonly ushort _sessionId;
         private readonly ManualResetEventSlim _packetEvent = new ManualResetEventSlim(false);
         private readonly ManualResetEventSlim _ackEvent = new ManualResetEventSlim(false);
-        private readonly SNISMUXHeader _currentHeader = new SNISMUXHeader();
+        //private readonly SNISMUXHeader _currentHeader = new SNISMUXHeader();
```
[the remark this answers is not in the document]
Wraith2: All the commented text was left to make things easier when reviewing. It's a complex piece of code and i found it helpful to refer to the original. I can remove it now or wait until it's been thoroughly reviewed.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIMarsHandle.cs
Wraith2: I investigated this. The reason that `lock (this)` is advised against is that if the object is public then unexpected users can use the same object to lock on causing unexpected behaviour. The suggested alternative is using dedicated and private lock object. The same reason applies for not locking on types. In the situation we have here we know that the object is never exposed to users outside our library and we know that we are being careful about locking and lifetimes. So the guidance is good but we don't need to follow it because we have control.
However, adding an additional dedicated locking object isn't expensive if you and others prefer, i just thought it wasn't necessary.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIMarsHandle.cs
Wraith2: Unknown, this is a new exception and may not even need to be an exception it could just be an assertion. I needed it in place during development to make sure i didn't accidentally call this while holding the lock. What do you think?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIMarsHandle.cs
JRahnama: I would vote for a dedicated private lock for two reasons:
1. We follow the guideline, although there is no harm done by its current implementation.
2.  We don't know what changes will be forced, by issues fixation or new features coming in.... I am waiting for `Windows 11` to kick in. Things may move around and we may have to give access to some of our APIs, most probably that will never  happen in this case, but we make our own pattern and style, to use a a dedicated private lock everywhere inside the driver.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIMarsHandle.cs
JRahnama: I vote for assertion.

JRahnama:
Also, can you make some changes and push them back to the pipelines to rerun the the tests? I could not do it from our side.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIMarsHandle.cs
Wraith2: Ok. done. Just bear in mind that the tests for this repo are never run in debug mode so assertions won't fire in testing.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIMarsHandle.cs
Wraith2: Windows 11 isn't going to change anything at this level. It's an OS update and if it changes dotnet runtime behaviours it'd be a massive ecosystem breaking change. That said, I'm ok with adding a dedicated sync object so that's done.

cheenamalhotra (approved):

JRahnama (approved):

DavoudEshtehari (commented):
With this change you shows a well knowledge of the SMP. Please, include any additional information like diagram here, especially,  using 'State' and 'DemuxState' enums as a history to help perceiving it in the future if you have something.

Here are my understanding(feel free to verify it or replace it with a better version):
image

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIMarsConnection.cs
```diff
@@ -14,50 +14,45 @@ namespace Microsoft.Data.SqlClient.SNI
...
-            get
-            {
-                return _connectionId;
-            }
-        }
+        public Guid ConnectionId => _connectionId;
 
         public int ProtocolVersion => _lowerHandle.ProtocolVersion;
 
+        public object DemuxerSync => this;
```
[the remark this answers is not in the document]
Wraith2: It could but the similar properties above it that already existed are not. I prefer to use public/private even on internal classes because it indicates the items that I consider to be the public contract and private implementation details cleanly, I find internal things are often private but someone needed them and didn't want to go to the effort of making them a property or something similar. A lot of the weird coupling we have in the library in places like TdsParser and the state object come from things being exposed that possibly shouldn't have been.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIMarsConnection.cs
```diff
@@ -218,229 +192,341 @@ public void HandleSendComplete(SNIPacket packet, uint sniErrorCode)
...
+            Payload = 2,
+            Dispatch = 3
+        }
+
+        private enum State : uint
+        {
+            Demux,
+            HandleAck,
+            HandleData,
+            Recieve,
```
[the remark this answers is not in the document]
Wraith2: Opps, fixed.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIMarsHandle.cs
```diff
@@ -55,6 +55,8 @@ internal sealed class SNIMarsHandle : SNIHandle
         /// </summary>
         public override void Dispose()
         {
+            // SendControlPacket will lock so make sure that it cannot deadlock by failing to enter the DemuxerLock
+            Debug.Assert(_connection != null && Monitor.IsEntered(_connection.DemuxerSync), "SNIMarsHandle.HandleRecieveComplete should be be called while holding the SNIMarsConnection.DemuxerSync because it can cause deadlocks");
```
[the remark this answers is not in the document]
Wraith2: All instances fixed.

Wraith2:
> Here are my understanding(feel free to verify it or replace it with a better version):
> image

I spent some time thinking through and trying to draw out a diagram and I couldn't produce anything better. 
I think that the code has to be the final expression of what is going on. My hope was/is that there are enough comments and assertions to explain that is happening and guard against any harmful changes that might be introduced. I also hope that it's clear enough and fast enough that no-one needs to change or rewrite it any time soon.

DavoudEshtehari (approved):

PR #937, opened 2021-02-26
Change MakeReadAsyncBlocking default to false

David-Engel:
Now that the underlying open result set counter issue is fixed, it should be safe to change this to false and regain the performance we lost when we had to change it to true.


cheenamalhotra (approved):

cheenamalhotra:
azp run "CI-BuildNPack"

karinazhou (approved):

PR #940, opened 2021-03-02
Fix | Fix trace string bug

cheenamalhotra:
Fixes #939

PR #941, opened 2021-03-02
[Release 2.1] Fix Trace String bug

cheenamalhotra:
Minor change.
Port #940 to v2.1


karinazhou (approved):

DavoudEshtehari (approved):
In `netfx` you should rearrange the index order since it skips the 9 index.

PR #943, opened 2021-03-02
Import Dotnet GenAPI source for local management

cheenamalhotra:
Since "GenAPI" is an internally managed tool we depend on, that isn't released and does not seem planned for release anytime soon, we shall manage it locally to remove dependency on "dotnet-eng" feed.

_Tool imported from dotnet/arcade_
(Developers should not notice any difference in build and execution)

I did not import these new projects in MDS Solution since they're not to be actively worked upon, let me know for any suggestions/opinions.

cc @David-Engel


tools/GenAPI/Microsoft.Cci.Extensions/Experimental/TrimAPIMutator.cs
```diff
@@ -0,0 +1,125 @@
+// Licensed to the .NET Foundation under one or more agreements.
```
karinazhou, line 1: Do we need this file?

tools/GenAPI/Microsoft.Cci.Extensions/Extensions/CustomAttributeProvider.cs
```diff
@@ -0,0 +1,100 @@
+using System;
```
karinazhou, line 1: Missing headings

tools/GenAPI/Microsoft.Cci.Extensions/Extensions/FakeCustomAttribute.cs
```diff
@@ -0,0 +1,63 @@
+﻿using System;
```
karinazhou, line 1: Missing headings

tools/GenAPI/Microsoft.Cci.Extensions/SRMetadataPEReaderCache.cs
```diff
@@ -0,0 +1,64 @@
+using System;
```
karinazhou, line 1: License

karinazhou (approved):

JRahnama (approved):

Issue #44, opened 2018-01-28
Canceling SQL Server query with while loop hangs forever


madelson:
See StackOverflow post (https://stackoverflow.com/questions/48461567/canceling-query-with-while-loop-hangs-forever?noredirect=1#comment83955305_48461567) here for a full description of the issue.

Essentially, the issue is that for a certain query I am finding that calling `CancellationTokenSource.Cancel()` hangs indefinitely instead of canceling the query. The same query cancels instantly in SQL Server Management Studio. Here is code the reproduces the issue:

```
void Main()
{   
    var cancellationTokenSource = new CancellationTokenSource();
    var blocked = RunSqlAsync(cancellationTokenSource.Token);
    Console.WriteLine(blocked.Wait(TimeSpan.FromSeconds(1))); // false (blocked in SQL as expected)
    cancellationTokenSource.Cancel(); // hangs forever?!
    Console.WriteLine("Finished calling Cancel()");
    blocked.Wait();
}

public async Task RunSqlAsync(CancellationToken cancellationToken)
{
    var connectionString = new SqlConnectionStringBuilder { DataSource = @".\sqlexpress", IntegratedSecurity = true, Pooling = false }.ConnectionString;
    using (var connection = new SqlConnection(connectionString))
    {
        await connection.OpenAsync().ConfigureAwait(false);

        using (var command = connection.CreateCommand())
        {
            command.CommandText = @"
                WHILE 1 = 1
                BEGIN
                    DECLARE @x INT = 1
                END
            ";
            command.CommandTimeout = 0;
            Console.WriteLine("Running query");
            await command.ExecuteNonQueryAsync(cancellationToken).ConfigureAwait(false);
        }
    }
}
```

madelson:
Thanks for the update @divega . Now that there is renewed focus on SqlClient, can this be labeled as a bug? It's clearly buggy behavior and as more and more people shift to using async queries passing through the RequestAborted cancellation token from ASP.NET Core I anticipate that this will come up more frequently.

divega:
The explanation at https://github.com/dotnet/SqlClient/issues/44#issuecomment-362757646 makes me wonder if async (i.e. CancellationToken-based) cancellation should have a similar implementation to SqlCommand.Cancel(), e.g. using Attention messages.

However I think it in https://github.com/dotnet/SqlClient/issues/109 we were seeing async cancellation work correctly, so I am not sure I understand this issue.   

cc @Wraith2, @roji, @saurabh500

[...]

PR #956, opened 2021-03-03
Fix Async Cancel

Wraith2:
fixes https://github.com/dotnet/SqlClient/issues/44

This removes two contending locks and adds an interlocked state int similar to the way that timeouts were recently changed. This prevents waiting and allows cancel to proceed only if end of execution hasn't started otherwise cancel is ignored. 

There was already an async cancel test which was passing so I added a new one based on the repro from the original thread.


src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/SqlCommand/SqlCommandCancelTest.cs
```diff
@@ -558,5 +573,46 @@ private static async Task AsyncCancelDoesNotWait(string connStr)
...
+            Assert.NotNull(exception);
+            Assert.InRange((ended - started).TotalSeconds, cancelSeconds, delaySeconds - 1);
+        }
+
+        private static async Task ExecuteWaitForAsync(CancellationToken cancellationToken, string connectionString, int delaySeconds)
+        {
+            using (var connection = new SqlConnection(connectionString))
+            {
+                await connection.OpenAsync().ConfigureAwait(false);
+                using (var command = new SqlCommand($"WAITFOR DELAY '00:00:{delaySeconds:D2}'", connection))
```
cheenamalhotra, line 610: When using WAITFOR DELAY, we cannot reproduce this issue, are you sure this test failed without your changes?

Wraith2: No. I did the fix dev using the original repro from the report and then test-ified the repro after so that it used wait which I thought acted the same way. I've changed to the loop from the repro in the test now. Sorry about that.

johnnypham:
Looks good to me. Can you bring the changes to netfx too?

Wraith2:
Netfx added.

johnnypham (approved):

cheenamalhotra:
The failing TVP Test is an actual error and needs further investigation.
Shouldn't be treated as random failure as it's failing in all test pipelines.

Wraith2:
Great, just one tiny test to debug...

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/SqlCommand/SqlCommandCancelTest.cs
```diff
@@ -558,5 +573,50 @@ private static async Task AsyncCancelDoesNotWait(string connStr)
...
+            {
+                await connection.OpenAsync().ConfigureAwait(false);
+                using (var command = new SqlCommand(@"
+WHILE 1 = 1
+BEGIN
+    DECLARE @x INT = 1
+END", connection))
+                {
+                    command.CommandTimeout = delaySeconds + 10;
+                    await command.ExecuteNonQueryAsync(cancellationToken).ConfigureAwait(false);
```
DavoudEshtehari, line 617: To maximize the code coverage I'd suggest running all the `command.ExecuteXXXAsync` functions.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
```diff
@@ -629,68 +637,53 @@ internal void Activate(object owner)
             Debug.Assert(result == 1, "invalid deactivate count");
         }
 
+        internal bool SetCancelStateClosed()
```
DavoudEshtehari, line 640: Did you mean **SetCancelStateUnset**?

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/SqlCommand/SqlCommandCancelTest.cs
```diff
@@ -558,5 +573,50 @@ private static async Task AsyncCancelDoesNotWait(string connStr)
...
+        private static async Task ExecuteWaitForAsync(CancellationToken cancellationToken, string connectionString, int delaySeconds)
+        {
+            using (var connection = new SqlConnection(connectionString))
+            {
+                await connection.OpenAsync().ConfigureAwait(false);
+                using (var command = new SqlCommand(@"
+WHILE 1 = 1
+BEGIN
+    DECLARE @x INT = 1
+END", connection))
```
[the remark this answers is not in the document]
Wraith2: See https://github.com/dotnet/SqlClient/pull/956#discussion_r589835694

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
Wraith2: No. CompareExchange is confusing, it changes the value in the ref for the first arg to the value in the second if the current value is the third, and then returns the original value not the new value. So this changes it from unset to closed and returns a boolean which identifies if it succeeded in changing to closed.

Wraith2:
I've put some time into working through TvpMain but I can't work out what the tests are supposed to be doing so I can't isolate what I might have got wrong. Any chance of some help on this?

DavoudEshtehari:
The TVP failure occurs on [StreamInputParam.ImmediateCancelBin()](https://github.com/dotnet/SqlClient/blob/5f171ae9872bef69f79b65120e728acb33cadee2/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ParameterTest/StreamInputParam.cs#L468).

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
Wraith2: You were wrong about the name but right that this code wasn't correct.

CompareExchange is confusing. Regardless of whether it makes a change or not it returns the old value in the ref location which is the first parameter. So you can't tell whether it's succeeded correctly by checking the return value alone in cases where multiple transitions from a single state are allowed, in this case we have moves from Unset to both Closed and Cancelled. So I've added checks for the value in the ref location after the CompareExchange.

Wraith2:
I think I'm going to give up on this. Tests that work locally don't work in the CI, things that work in CI don't work locally and the bizarre threading nature of cancellation makes trying to track down and juggle the exact steps to get all the errors matching up is baroque and frustrating. I'll leave it to the professionals.

cheenamalhotra:
Hi @Wraith2 

We'll help you with that, I'm reopening PR so we can help investigate test issues, don't worry!

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
```diff
@@ -650,68 +658,57 @@ internal void Activate(object owner)
...
+            return Interlocked.CompareExchange(ref _cancelState, CancelState.Unset, CancelState.Closed) == CancelState.Unset;
+        }
+
         // This method is only called by the command or datareader as a result of a user initiated
         // cancel request.
         internal void Cancel(int objectID)
         {
-            bool hasLock = false;
-            try
+            // Keep looping until we either grabbed the lock (and therefore sent attention) or the connection closes\breaks
```
DavoudEshtehari, line 670:
```suggestion
```

DavoudEshtehari (commented):
Hi @Wraith2, After checking the test failures I noticed  `TdsParserStateObject.Cancel` doesn't get into the `if-block` that will be fixed by applying the recent changes on `netfx` (_be careful about the confusing **CompareExchange** because the current values of the current and new parameters should be swapped_ 😉).

DavoudEshtehari:
The mistaken commits were rebased and merged with the latest source code.

JRahnama (approved):

johnnypham (approved):

PR #437, opened 2020-02-20
Improved Sql bulk copy error message

DavoudEshtehari:
Improved the error message for wrong data type conversion with adding more detail about the source of error in the **sqlBulkCopy**.
In this change, the error message of converting a **string** to a **numerical** column is handled and the complementary information about the **column name** and the **row number** of the data source is added.
We brought this additional information for **DataTable** and **DataRow[]**. For the rest of the datatype supported by SqlBulkCopy, it will add in the next improvements.

PR #960, opened 2021-03-04
Do not truncate given value in exception message for BulkLoadCannotConvertValue

Berreek:
It was introduced in #437 - for some reason the given value is getting truncated when it is longer than 100 chars - I looked trough PR comments and didn't see anything why it is like this. The exception message together with inner exception is really missleading when column is bigger than 100 chars and given value is bigger than column size. It looks like this:

Exception message:
```
The given value 'https://something.com/5e6109fb8ca1de00010b2446/documents/3d47b19d-c469-4c36-a3f4-edda' of type String from the data source cannot be converted to type nvarchar for Column 8 [Url] Row 1.
```

Inner exception message:
```
String or binary data would be truncated in table '[Something].[File]', column 'Url'. Truncated value: 'https://something.com/5e6109fb8ca1de00010b2446/documents/3d47b19d-c469-4c36-a3f4-edda
```

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlUtil.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlUtil.cs
@@ -884,11 +884,7 @@ namespace Microsoft.Data.SqlClient
         }
         internal static Exception BulkLoadCannotConvertValue(Type sourcetype, MetaType metatype, int ordinal, int rowNumber, bool isEncrypted, string columnName, string value, Exception e)
         {
-            string quotedValue = string.Empty;
-            if (!isEncrypted)
-            {
-                quotedValue = string.Format(" '{0}'", (value.Length > 100 ? value.Substring(0, 100) : value));
-            }
+            string quotedValue = isEncrypted ? string.Empty : $" '{value}'";
             if (rowNumber == -1)
             {
                 return ADP.InvalidOperation(System.StringsHelper.GetString(Strings.SQL_BulkLoadCannotConvertValueWithoutRowNo, quotedValue, sourcetype.Name, metatype.TypeName, ordinal, columnName), e);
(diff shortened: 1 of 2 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/SqlClient?pullRequest=960) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/SqlClient?pullRequest=960) before we can accept your contribution.<br/><br/>:x: Berreek [sign now](https://cla.dotnetfoundation.org/dotnet/SqlClient?pullRequest=960)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/SqlClient?pullRequest=960) it.</sub>

Wraith2:
The non-truncated value could be anything up to ~2Gib in size and I don't think many people would appreciate that appearing in their exception logs. Is there something that can be known from the full value that can't be from the first 100 chars or so?

David-Engel:
As Wraith2 said, it was definitely truncated on purpose and we don't want to return the full value. But perhaps we can indicate that by appending "..." to the truncated value?
```
quotedValue = string.Format(" '{0}'", (value.Length > 100 ? value.Substring(0, 100) + "..." : value));
```

Berreek:
> Is there something that can be known from the full value that can't be from the first 100 chars or so?

In my case I had url which had query params that I did not expect to have when I was setting up column size.

> But perhaps we can indicate that by appending "..." to the truncated value?

I like this idea and get your point about about the size of non-truncated value. But except this could we also add atleast the length of given value? In most cases I guess you would like to know by how much limit is exceeded.

cheenamalhotra:
@Berreek 

> I like this idea and get your point about about the size of non-truncated value. But except this could we also add atleast the length of given value? In most cases I guess you would like to know by how much limit is exceeded.

Could you append the actual length to the message instead of full value keeping current message as-is and re-prepare the PR for review?

JRahnama:
Not planed.

PR #964, opened 2021-03-05
AKV provider docs update

cheenamalhotra:
The links for examples do not work anymore (ref: https://github.com/dotnet/SqlClient/pull/581), so I decided to move this entire block of remarks to redirect to official documentation so we can maintain from 1 single point. API ref publishing activity is currently in progress for both v1.2 and v2.0

### DIFF
--- a/src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/SqlColumnEncryptionAzureKeyVaultProvider.cs
+++ b/src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/SqlColumnEncryptionAzureKeyVaultProvider.cs
@@ -312,13 +312,4 @@ namespace Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider
 
         #endregion
     }
-
-    /// <summary>
-    /// The authentication callback delegate which is to be implemented by the client code
-    /// </summary>
-    /// <param name="authority"> Identifier of the authority, a URL. </param>
-    /// <param name="resource"> Identifier of the target resource that is the recipient of the requested token, a URL. </param>
-    /// <param name="scope"> The scope of the authentication request. </param>
-    /// <returns> access token </returns>
-    public delegate Task<string> AuthenticationCallback(string authority, string resource, string scope);
 }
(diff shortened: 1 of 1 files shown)


src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/SqlColumnEncryptionAzureKeyVaultProvider.cs
```diff
@@ -14,36 +14,26 @@ namespace Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider
...
-    /// 
-    /// Once the provider is registered, it can used to perform Always Encrypted operations by creating Column Master Key using Azure Key Vault Key Identifier URL.
-    /// 
-    /// ## Example
-    /// 
-    /// Sample C# applications to demonstrate Always Encrypted use with Azure Key Vault are available at links below:
-    /// 
-    /// - [Example: Using Azure Key Vault with Always Encrypted](~/connect/ado-net/sql/azure-key-vault-example.md)
-    /// - [Example: Using Azure Key Vault with Always Encrypted with enclaves enabled](~/connect/ado-net/sql/azure-key-vault-enclave-example.md)
+    /// For more information, see: [Using the Azure Key Vault Provider](https://docs.microsoft.com/sql/connect/ado-net/sql/sqlclient-support-always-encrypted#using-the-azure-key-vault-provider)
```
cheenamalhotra, line 36: Do we also want to update this for v1.1 and v1.0 API ref documentations?
Current documentation: https://docs.microsoft.com/en-us/dotnet/api/microsoft.data.sqlclient.alwaysencrypted.azurekeyvaultprovider.sqlcolumnencryptionazurekeyvaultprovider

karinazhou (approved):

src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/SqlColumnEncryptionAzureKeyVaultProvider.cs
David-Engel, line 36: I'm fine with removing this block and just linking. Yes, I think we should also update for 1.0 and 1.1.

We should also use a relative URL:
```suggestion
    /// For more information, see: [Using the Azure Key Vault Provider](/sql/connect/ado-net/sql/sqlclient-support-always-encrypted#using-the-azure-key-vault-provider) 
```

David-Engel (approved):

PR #976, opened 2021-03-09
Test: update test build config to copy config.json correctly

Wraith2:
I've tracked down some problems I've been having running tests locally and found that it was due to the config.json file the docs tell you to edit not being copied locally for the tests. For some reason that `localhost` default alias doesn't work for me.

/cc @DavoudEshtehari because I was mentioning this in relation to retry test failures.

### DIFF
--- a/src/Microsoft.Data.SqlClient/tests/tools/Microsoft.Data.SqlClient.TestUtilities/Microsoft.Data.SqlClient.TestUtilities.csproj
+++ b/src/Microsoft.Data.SqlClient/tests/tools/Microsoft.Data.SqlClient.TestUtilities/Microsoft.Data.SqlClient.TestUtilities.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFrameworks>net461;netcoreapp2.1</TargetFrameworks>
     <TargetGroup Condition="$(TargetFramework.StartsWith('netcoreapp'))">netcoreapp</TargetGroup>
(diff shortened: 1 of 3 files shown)


src/Microsoft.Data.SqlClient/tests/tools/Microsoft.Data.SqlClient.TestUtilities/Microsoft.Data.SqlClient.TestUtilities.csproj
```diff
@@ -12,10 +12,9 @@
     <Copy SourceFiles="config.default.json" DestinationFiles="config.json" Condition="!Exists('config.json')" />
   </Target>
   <ItemGroup>
-    <Content Include="config.json">
-      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-      <Link>config.json</Link>
-    </Content>
+    <None Update="config.json">
```
cheenamalhotra, line 15: Config json change can be avoided, pipelines are not liking it.

Wraith2: If we don't do this then the test version ends up being restored which presumably works for pipelines but won't locally (for me). 
There is a conditional copy of config.defaults.json in the project but i couldn't ever see if being used in the binlog. I assume the logic should be that msbuild will copy config.json from manualtests if it exists but if it doesn't then it should use the default version? At the moment the default version always wins and it comes from the TestUtilities project which is confusing.

cheenamalhotra: Manual Tests project does not need to contain config.json anymore.
Config.json belongs to TestUtilities now. Buildguide references were also updated.

Wraith2: In which case I've been trying to solve a problem of my own making. Ooops.

Issue #255, opened 2019-10-10
SqlDataReader incorrectly returns null ROWVERSION as an empty byte[]


dbrownems:
When returned in a resultset a null ROWVERSION is returned as an empty byte[], instead of DbNull.Value.  This problem does not occur when returning using a parameter value, or using a resultset in ODBC (so not a TDS issue).

Here's e a repro:

```c#
using System;
using System.Data.Odbc;
using System.Data.SqlClient;
using System.Data.SqlTypes;
using System.Threading;
using System.Threading.Tasks;
using System.Transactions;

namespace ConsoleApp14
{
    class Program
    {

        static void Main(string[] args)
        {

            using (var con = new SqlConnection("Server=localhost;database=tempdb;Integrated Security=true"))
            {
                con.Open();
                var cmd = con.CreateCommand();
                cmd.CommandText = "select cast(null as rowversion) rv";
                using (var rdr = cmd.ExecuteReader())
                {
                    rdr.Read();
                    var allowDbNull = rdr.GetColumnSchema()[0].AllowDBNull;
                    var isNull = rdr.IsDBNull(0);
                    var val = rdr[0];

                    Console.WriteLine($"SqlClient: AllowDbNull {allowDbNull} IsDbNull: {isNull} {val.GetType().Name} {val}");

                }
            }


            using (var con = new SqlConnection("Server=localhost;database=tempdb;Integrated Security=true"))
            {
                con.Open();
                var cmd = con.CreateCommand();
                cmd.CommandText = "select @val = cast(null as rowversion) ";

                var p = cmd.Parameters.Add(new SqlParameter("@val", System.Data.SqlDbType.Timestamp));
                p.Direction = System.Data.ParameterDirection.Output;

                cmd.ExecuteNonQuery();
                {

                   SqlBinary val = (SqlBinary) p.SqlValue;
                   Console.WriteLine($"SqlClient (parameter): IsDbNull: {val.IsNull} {val.GetType().Name} {val}");

                }
            }

            using (var con = new OdbcConnection("Driver={ODBC Driver 17 for SQL Server};Server=localhost;Trusted_Connection=yes"))
            {
                con.Open();
                var cmd = con.CreateCommand();
                cmd.CommandText = "select cast(null as rowversion) rv";
                using (var rdr = cmd.ExecuteReader())
                {
                    rdr.Read();
                    var allowDbNull = rdr.GetSchemaTable().Rows[0]["AllowDBNull"];
[...]

PR #998, opened 2021-03-19
Fix rowversion null behaviour

Wraith2:
Fixes https://github.com/dotnet/SqlClient/issues/255

Changed RowVersion/Timestamp null behaviour for SqlDataReader so that it returns a DBNull value instead of an empty byte[]. An AppContext switch has been added to allow reverting to the previous (broken but kept that way for compatibility) behaviour, it's called `Switch.Microsoft.Data.SqlClient.LegacyRowVersionNullBehaviour` and is cached on first check for the entire process.

N.B. this is a breaking behavioural change.

/cc @dbrownems @bdebaere @mburbea

### DIFF
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataReaderTest/DataReaderTest.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataReaderTest/DataReaderTest.cs
@@ -5,6 +5,8 @@
 using System;
 using System.Collections.Generic;
 using System.Data;
+using System.Data.SqlTypes;
+using System.Reflection;
 using System.Text;
 using System.Threading;
 using Xunit;
@@ -13,6 +15,8 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests
 {
     public static class DataReaderTest
     {
+        private static object s_rowVersionLock = new object();
+
         [ConditionalFact(typeof(DataTestUtility), nameof(DataTestUtility.AreConnStringsSetup))]
         public static void LoadReaderIntoDataTableToTestGetSchemaTable()
         {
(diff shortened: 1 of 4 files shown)


David-Engel (approved):
LGTM

ErikEJ:
No tests??

Wraith2:
> No tests??

The appcontext switch is process global so I couldn't deterministically make sure that it turned on and off for the right tests. It's also a trivial change tested locally so it didn't seem worth doing lots of private reflection.

roji:
> The appcontext switch is process global so I couldn't deterministically make sure that it turned on and off for the right tests. It's also a trivial change tested locally so it didn't seem worth doing lots of private reflection.

I think it should be possible to turn on an appcontext switch as part of the test (i.e. [AppContext.SetSwitch](https://docs.microsoft.com/en-us/dotnet/api/system.appcontext.setswitch?view=net-5.0)), and back off at the end of it; take care to also disable parallelization for that test to avoid affecting other tests running in parallel.

(no idea on whether testing is important for this particular case etc.)

ErikEJ:
But it should be possible to add a test to confirm the new changed default behaviour? Or do I misunderstand?

Wraith2:
If I can find how to prevent the tests running in parallel as @roji says is possible then yes it can be tested. 
I just hoped I could get away without having to write one for something this trivial, I really hate writing tests.

ErikEJ:
It should be possible to write a test confirming the new behaviour without changing any AppContext switches etc...

Wraith2:
Positive test added.

cheenamalhotra:
> If I can find how to prevent the tests running in parallel as @roji says is possible then yes it can be tested.

@Wraith2 
Our tests don't run in parallel in single process.
We run multiple processes for different configurations only.

Wraith2:
Should I take that to mean that you wan me to add a negative test?

Wraith2:
Negative test added and positive test reworked to use the same way of ensuring that the test has the correct app context value set. I've used private reflection to set the internal field value and then restore it after each test, without doing this the per-process cache will prevent both tests succeeding in the same process. A simple lock prevents them running concurrently which is safe because they're very fast tests.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataReaderTest/DataReaderTest.cs
```diff
@@ -288,5 +292,77 @@ public static void CheckHiddenColumns()
...
+                finally
+                {
+                    SetLegacyRowVersionNullBehaviour(originalValue);
+                }
+            }
+        }
+
+        private static bool? SetLegacyRowVersionNullBehaviour(bool? value)
+        {
+            Type switchesType = typeof(SqlCommand).Assembly.GetType("Microsoft.Data.SqlClient.LocalAppContextSwitches");
```
cheenamalhotra, line 361: Could you not simply call `AppContext.SetSwitch("Switch.Microsoft.Data.SqlClient.LegacyRowVersionNullBehaviour",true);` and change driver behavior ?
Any special reason to use reflection ?

Wraith2: It's cached per process. Whichever test ran first would cause the other to fail. I have to set the value I want explicitly each time.

cheenamalhotra: You need to set/reset before and after running test to ensure guaranteed setting.

Wraith2: No I don't because I'm setting the cache variable and totally bypassing the appcontext switch. Check the code carefully.

cheenamalhotra (approved):

Issue #982, opened 2021-03-11
Allow TVPs to be populated via asynchronous data sources & add a new field to SqlDbType


mburbea:
### Is your feature request related to a problem? Please describe.
Currently when you use IDataReader/DbDataReader as a source for a tvp parameter, it will always call the `Read()` method to consume rows. 

In addition, only `IEnumerable<SqlDataRecord>` can be used as a source, not `IAsyncEnumerable<SqlDataRecord>`.

I think as a solution it would be beneficial to add a new SqlDbType, `StructuredAsync`, which would allow using an asynchronous way of populating a structured table valued parameter.

### Describe the solution you'd like
usage would be like so:
```cs
IAsyncEnumerable<SqlDataRecord> asyncRecords = GetAsyncRecords();
var paramA = new SqlParameter("@tvp", asyncRecords){
                SqlDbType = SqlDbType.StructuredAsync,
                TypeName = "dbo.tvp"
};
IDataReader asyncReader = GetAsyncReader();
var paramB = new SqlParameter("@tvp", asyncReader){
                SqlDbType = SqlDbType.StructuredAsync,
                TypeName = "dbo.tvp"
};
```
### Describe alternatives you've considered
As an alternative, we could always call `ReadAsync(CancellationToken)`, however, this may regress scenarios where the reader is entirely synchronous and no async operations are needed.
Another approach could be add a boolean property in `SqlParameter` like `ReadMode`, but I feel like there are too many constructors and mutable properties on a `SqlParameter` already that is already incredibly difficult to find the options you neccessarily care about.

Wraith2:
The problem with that is that you'd be holding the write operation on an await for the parameters which is complicated. Why would you need to send async enumerable params to a tvp? what is the discovery mechanism for the values?

mburbea:
Presumably, the usual reasons to use a tvp but your source doesn't have all the data immediately. You are sending a datareader from another database, or you are translating some message bus/web call into a sql data record. 

The tvp would query the schema table like it does currently. For the IAsyncEnumerable, the library would read the first row for the sqlmetadata[]. 

The first thing a tvp does is basically insert the rows into the tvp, and rather than streaming the records synchronously, this operation could be done async.

Wraith2:
[...]

PR #999, opened 2021-03-20
Hacky experiment to switch ExecuteNonQuery to modern async/await

roji:
Following discussion in https://github.com/dotnet/SqlClient/issues/982#issuecomment-799647355, here is a super-hacky experiment to convert part of the ExecuteNonQuery code path to use modern async/await. Most of the code implementing ExecuteNonQueryAsync is no longer used, and a unified code path with async/await is shared between sync and async. This is only intended to point at a possible direction of development - I'm sure I've overlooked many important details.

The unified code path is basically the existing sync path, decorated with async/await keywords and adapted to pass down an isAsync flag and eventually call into the correct sync-or-async I/O method. I believe that slowly transforming the codebase from the current callback-based async implementation - which is largely duplicated/separate and very hard to follow - is an important investment to keep the driver modern and bug-free, and successfully evolve it going forward (I'd consider it paying down technical debt).

Notes:

* This refactors the higher levels of the implementation, but doesn't go all the way down; for writing, it stops in TdsExecuteSQLBatch, and for reading, in RunExecuteNonQueryTds. It also takes care of the simple path in ExecuteNonUpdate (where a reader isn't necessary). Methods which have been refactored have the suffix Experimental.
* In many cases, the current code sort-of attempts to handle sync and async within the same code path, up to a point:
    * Many methods return a Task and either receive an async flag as a parameter, or have the flag somewhere as state (e.g. _parser._asyncWrite)
    * These methods return a Task when running in async mode, and null when running in sync mode.
    * Calling methods check if a Task was returned or null. If null, they execute the continuation immediately (sync), otherwise they schedule the continuation after the returned task with ContinueWith (TdsParser.TdsExecuteSQLBatch is a good example).
    * This pattern can be relatively easily converted to modern async code:
        * Callers can be made async, and simply await the task if it's non-null, instead of using ContinueWith.
        * Ideally, callees can be modified to *always* return a Task - for sync mode, just return a completed one.
* There seems to be a lot of guarding against concurrent/overlapping asynchronous operations (e.g. SNINetworkStream wraps async I/O operations in a semaphore). I suspect this could be because a new operation is sometimes (erroneously?) started before the previous operation completed - or is there another, more fundamental reason? If it's just that, then switching to async/await should make it much easier to avoid this error, and could possibly obviate all this guarding.
* Strangly, I see various code that uses the terminology "sync over async", although they aren't in the standard sense. For example, SNIProxy.ReadSyncOverAsync simply calls into sync reading.
* Warning: the above was an experiment. Simply converting sync code to async isn't always the right way performance-wise, as things sometimes have to be refactored further. For example, lots of small async methods have an overhead which does not exist with sync methods. However, I'm seeing various practices in the current codebase that make suspect that even a simple conversion to async/await wouldn't make it slower.

/cc @Wraith2 @cheenamalhotra @David-Engel @ajcvickers

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
@@ -805,7 +805,7 @@ namespace Microsoft.Data.SqlClient
                 else
                 {
                     // Validate the command outside of the try/catch to avoid putting the _stateObj on error
-                    ValidateCommand(isAsync: false);
+                    ValidateCommand();
 
                     bool processFinallyBlock = true;
                     try
@@ -1112,7 +1112,9 @@ namespace Microsoft.Data.SqlClient
         }
 
         /// <include file='../../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlCommand.xml' path='docs/members[@name="SqlCommand"]/ExecuteNonQuery[@name="default"]/*'/>
-        public override int ExecuteNonQuery()
+        public override int ExecuteNonQuery() => ExecuteNonQuery(isAsync: false).GetAwaiter().GetResult();
+
+        private async Task<int> ExecuteNonQuery(bool isAsync, CancellationToken cancellationToken = default)
         {
             // Reset _pendingCancel upon entry into any Execute - used to synchronize state
             // between entry into Execute* API and the thread obtaining the stateObject.
@@ -1131,11 +1133,14 @@ namespace Microsoft.Data.SqlClient
                 WriteBeginExecuteEvent();
                 if (IsRetryEnabled)
                 {
-                    InternalExecuteNonQueryWithRetry(sendToPipe: false, timeout: CommandTimeout, out _, asyncWrite: false, inRetry: false);
+                    throw new NotImplementedException();
+                    // InternalExecuteNonQueryWithRetry(sendToPipe: false, timeout: CommandTimeout, out _, asyncWrite: false, inRetry: false);
                 }
                 else
                 {
-                    InternalExecuteNonQuery(completion: null, sendToPipe: false, timeout: CommandTimeout, out _);
+                    // InternalExecuteNonQuery(completion: null, sendToPipe: false, timeout: CommandTimeout, out _);
+                    await InternalExecuteNonQueryExperimental(sendToPipe: false, timeout: CommandTimeout, usedCache: out _, isAsync)
+                        .ConfigureAwait(false);
                 }
                 return _rowsAffected;
             }
@@ -1549,7 +1554,7 @@ namespace Microsoft.Data.SqlClient
             // returns false for empty command text
             if (!inRetry)
             {
-                ValidateCommand(isAsync, methodName);
+                ValidateCommand(methodName);
             }
 
             CheckNotificationStateAndAutoEnlist(); // Only call after validate - requires non null connection!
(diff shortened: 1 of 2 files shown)


cheenamalhotra:
@roji 

Thanks for diving in! I understand the pain Async APIs are causing everywhere, I hear you. Sad part is this is not just .NET Core driver we have issues in, these issues also exist in .NET Framework since System.Data.SqlClient old versions. So we're dealing with the core design problem for Async APIs.

I've just recently stepped in the puddle (starting with ReadAsync) and there's a lot of mess around, that needs to be cleared to run things properly. I've been meeting with @Wraith2 weekly to discuss painful issues (that's how we fixed wrong data issue 659) and I can see the challenges ahead. I think we're still understanding current Async design in order to pin point exact design issue that's causing all these performance problems.

I would keep this style of change as a backup option if no other solutions can be found, and we just can't get Async design to work the way we expect. But first I would like to attempt fixing the current Async design problems. 

_If you have a top list of issues you want to summarize that you're looking at from EF end, we can prioritize those first._

I'm already started looking into "SqlDataReader.ReadAsync()" API issues that run extremely slow when compared to "SqlDataReader.Read()" and does not respect cancellation token. [Extended use-case of #593 ]

But overall, changing everything to do async/await on Sync methods is not going to be as simple as well, we'll need some tough testing as more products start to use and rely on us. There are some pipeline failures due to these changes too, which may be you can look into and try to fix in the meantime so we can have a prototype ready to what "might" be needed if everything else fails.

cheenamalhotra:
But just to summarize, I do appreciate and support this research idea, but it's definitely very raw at this point and shall need a lot of testing and polishing. We can definitely use it as a reference point to compare current Async design performance to what we can do here to get a better picture of current design flaws and fix it piece by piece.

roji:
Thanks for your comments @cheenamalhotra and for looking at this. FWIW this was only intended as a proof-of-concept - it wasn't meant as working change that passes the tests.

I guess the general points I'm trying to make are the following:

* Local improvements and specific (async-related) bug fixes are important, but in my mind, in the long term transitioning to modern async/await seems necessary in order to keep the driver codebase healthy and maintainable. It's always hard to justify paying technical debt and doing refactoring work, since customers don't see immediately value from it (as they do from bug fixes and new features); but in this case I truly hope you guys can manage to work in this direction.
* I sometimes get the vibe that async/await isn't considered a workable direction for SqlClient, for various reasons... My intention with this PR was to try to understand this, and hopefully indicate that this isn't the case. In the part of the codebase I've worked on, I haven't seen any necessary, major reasons why a conversion to async/await shouldn't be possible, and the sync code seems like it's convertible to async without too much trouble. This might not be the case for the entire codebase, but there's no need for this to be an all-or-nothing thing.
* Given what I've seen in the driver, converting the codebase to use async/await and possibly to unify code paths with sync is something that can be done gradually, rather than all-at-once.

> But overall, changing everything to do async/await on Sync methods is not going to be as simple as well, we'll need some tough testing as more products start to use and rely on us.

This is of course the major point. Without thorough test coverage for the code (and the async path), it's very hard to do any changes with confidence. If you feel that current coverage is insufficient, that definitely sounds like a good direction to improve before attempting any drastic changes.

I am going to go ahead and close this now. I'd be happy to continue the conversation, but of course it's ultimately up to you guys to decide which direction you want to go with this.

cheenamalhotra:
I absolutely agree with you and highly appreciate your ideas here. As you know the teams on this driver have evolved so many times and we're not the original designers of the Async APIs so as much as we're open to adapt and update code to latest methodologies, we're also a bit orthodox towards maintaining existing client trust and not making changes that can go out of our control.

We are focusing primarily towards owning the Async design first so we can be confident towards making big changes as needed and maybe we'll end up doing the same as you proposed with better coverage and understanding. This does help us clarify ideas so again, thank you for taking time on this one 🙂

About test coverage, that's also something we are working to improve, including benchmark testing that has been long delayed.

We're going to reactivate our perf test project that once existed in CoreFx but died in this repository, to use it as a base and continue to expand it with more use cases whenever anyone working towards perf improvements would like to contribute. We have a base prototype ready, just haven't got time and space to open up the PR yet 😊

roji:
Sounds good @cheenamalhotra, I know how big the challenges are and fully appreciate everything you're doing. I unfortunately don't have time to contribute code directly, but regularly keep my eye on the repo, and as always, don't hesitate to ping me on anything.

PR #1006, opened 2021-03-24
Remove API to enable event tracing in SNI.dll + update SNI dependency version

johnnypham:
Event tracing in SNI.dll will no longer be enabled through the client app. Subscribing a session to the `Microsoft.Data.SqlClient.EventSource` provider through tools like xperf or perfview will be all that is needed.


David-Engel (approved):

cheenamalhotra (approved):
Dependency on next Native SNI release.
(Wait for the release to merge)

PR #1010, opened 2021-03-25
Feature | Azure Identity support - Phase 1 changes

cheenamalhotra:
### Phase 1 changes
- Add new dependency on **Azure.Identity v1.3.0**.
- Use _Azure.Identity_ library to acquire token using `ManagedIdentityCredential()` for Managed Identity based authentication modes (`ActiveDirectoryMSI` and `ActiveDirectoryManagedIdentity`)
- Merge "AzureManagedIdentityAuthenticationProvider" into "ActiveDirectoryAuthenticationProvider" to support MSI modes from the same class "**ActiveDirectoryAuthenticationProvider**"
- Transition below authentication modes to start using _Azure.Identity_ to acquire access token:
   - ActiveDirectoryManagedIdentity > `ManagedIdentityCredential()`
   - ActiveDirectoryMSI > `ManagedIdentityCredential()`
   - ActiveDirectoryServicePrincipal > `ClientSecretCredential()`
      - New addition: Tenant id (**required**) - uses default from database info
- Accept "Client Id" for User-Assigned Managed Identity as _Azure.Identity_ doesn't support "Object Id"
   - Use existing "UserId" connection property to accept "Client Id" instead of "Object Id' starting with **v3.0**
- Pass "Connect Timeout" from `SqlConnection` property to `SqlAuthenticationParameters` to provide cancellation token to all Async calls in order to cancel token acquire request after "timeout" milliseconds.
   - _For Interactive Authentication on .NET Core:_ Cancellation token cancels request after 3 mins [as of current design]
	
#### Public Observations:
- 1 Breaking change:
   - "User Id" now accepts "Client Id" instead of "Object Id" for User-Assigned Managed Identity.
      - _It was initially designed to keep the behavior in sync with other SQL drivers, but _Azure.Identity_ doesn't support Object Id_
- 1 new Public API (read only)
   - `SqlAuthenticationParameters.ConnectionTimeout`
- 1 new Dependency on all supported platforms:
   - Azure.Identity v1.3.0


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ActiveDirectoryAuthenticationProvider.cs
```diff
@@ -99,24 +103,37 @@ public override void BeforeUnload(SqlAuthenticationMethod authentication)
 #endif
 
         /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/ActiveDirectoryAuthenticationProvider.xml' path='docs/members[@name="ActiveDirectoryAuthenticationProvider"]/AcquireTokenAsync/*'/>
-        public override Task<SqlAuthenticationToken> AcquireTokenAsync(SqlAuthenticationParameters parameters) => Task.Run(async () =>
+        public override async Task<SqlAuthenticationToken> AcquireTokenAsync(SqlAuthenticationParameters parameters)
```
DavoudEshtehari, line 106: I don't see any benefit in calling this function by `Task.Run()` where it's used.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlAuthenticationProviderManager.cs
```diff
@@ -28,8 +28,26 @@ internal partial class SqlAuthenticationProviderManager
         private readonly SqlClientLogger _sqlAuthLogger = new SqlClientLogger();
         private readonly string _applicationClientId = ActiveDirectoryAuthentication.AdoClientId;
 
-        public static readonly SqlAuthenticationProviderManager Instance;
+        public static readonly SqlAuthenticationProviderManager s_instance;
```
DavoudEshtehari, line 31: NIT: this is a public field and you can keep the uppercase naming convention here.

DavoudEshtehari (approved):

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlAuthenticationProviderManager.cs
cheenamalhotra: The class is internal, so doesn't really affect any public surface.
I just followed recommendations from VS.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -2346,7 +2346,7 @@ internal SqlFedAuthToken GetFedAuthToken(SqlFedAuthInfo fedAuthInfo)
                             }
                             else
                             {
-                                Task.Run(() => _fedAuthToken = authProvider.AcquireTokenAsync(authParamsBuilder).Result.ToSqlFedAuthToken()).Wait();
+                                _fedAuthToken = authProvider.AcquireTokenAsync(authParamsBuilder).Result.ToSqlFedAuthToken();
```
johnnypham, line 2349: Same changes can be made on netfx side?

doc/snippets/Microsoft.Data.SqlClient/SqlAuthenticationParameters.xml
```diff
@@ -46,5 +47,9 @@
             <summary>Gets the database name.</summary>
             <value>The database name.</value>
         </DatabaseName>
+        <TimeoutInMilliseconds>
```
johnnypham, line 50: Why are we using ms here? All other timeouts in this library use seconds, don't they?

johnnypham (approved):

doc/snippets/Microsoft.Data.SqlClient/SqlAuthenticationParameters.xml
cheenamalhotra: Sure, even though this is read only, makes sense to keep it in seconds.

JRahnama (approved):

Issue #1001, opened 2021-03-22
SqlCommandBuilder.DeriveParameters() incorrectly includes the database name in TypeName property values for SqlParameters for table-valued parameters


Metritutus:
### Description
When using `SqlCommandBuilder.DeriveParameters()` with a `SqlCommand` instance to execute a stored procedure with a table-valued parameter, the `SqlParameter` instance it produces (for the aforementioned table-valued parameter) has a `TypeName` property value which incorrectly includes the database name.

Attempting to execute this `SqlCommand` will produce an exception message explaining that it is not valid to include the database name in the type name for a table-valued parameter.

### Exception details
```
Exception message: The incoming tabular data stream (TDS) remote procedure call (RPC) protocol stream is incorrect. Table-valued parameter 1 ("@ids"), row 0, column 0: Data type 0xF3 (user-defined table type) has a non-zero length database name specified.  Database name is not allowed with a table-valued parameter, only schema name and type name are valid.
Stack trace:
at Microsoft.Data.SqlClient.SqlConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   at Microsoft.Data.SqlClient.SqlInternalConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   at Microsoft.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObject stateObj, Boolean callerHasConnectionLock, Boolean asyncClose)
   at Microsoft.Data.SqlClient.TdsParser.TryRun(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj, Boolean& dataReady)
   at Microsoft.Data.SqlClient.SqlDataReader.TryConsumeMetaData()
   at Microsoft.Data.SqlClient.SqlDataReader.get_MetaData()
   at Microsoft.Data.SqlClient.SqlCommand.FinishExecuteReader(SqlDataReader ds, RunBehavior runBehavior, String resetOptionsString, Boolean isInternal, Boolean forDescribeParameterEncryption, Boolean shouldCacheForAlwaysEncrypted)
[...]

PR #1020, opened 2021-04-02
Fix derived parameters containing typename incorrectly

Wraith2:
Fixes https://github.com/dotnet/SqlClient/issues/1001

This ports the fix and tests from corefx. It does not port the `SqlParameter` flags and xml schema changes which are unrelated to this bug, those will be ported in a separate PR.


src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/UdtTest/UdtTest2.cs
```diff
@@ -588,6 +588,128 @@ public void TestSqlUserDefinedAggregateAttributeMaxByteSize()
...
+                    }
+                    finally
+                    {
+                        try
+                        {
+                            transaction.Rollback();
+                        }
+                        catch
+                        {
+                            // ignore rollback failure exceptions to preserve original thrown error in test result
```
cheenamalhotra, line 644: I'm not in favor of "unexpected" test behaviors, that just makes it unreliable code.
If this is an "Expected" exception can we assert the condition it would occur on?
Otherwise if it should NOT occur, we need to fail here.

Secondly, when you say "_to preserve original thrown error in test result_" can you elaborate which exception you're talking about here? I don't see any other exception being throws/asserted.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/UdtTest/UdtTest2.cs
```diff
@@ -588,6 +588,128 @@ public void TestSqlUserDefinedAggregateAttributeMaxByteSize()
...
+                        }
+                    }
+                }
+
```
cheenamalhotra, line 648: Nit: extra line

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/UdtTest/UdtTest2.cs
```diff
@@ -588,6 +588,128 @@ public void TestSqlUserDefinedAggregateAttributeMaxByteSize()
...
+                        SqlParameter p = cmd.Parameters[1];
+
+                        Assert.Equal(customerParameterName, p.ParameterName);
+                        Assert.Equal(SqlDbType.Structured, p.SqlDbType);
+                        Assert.Equal(customerAddressTypeIncorrectName, p.TypeName); // the 3 part name is incorrect but needs to be maintained for compatibility
+                        p.Value = table;
+
+                        p.TypeName = customerAddressTypeIncorrectName; // force using the incorrect name by manually setting it
+
+                        Assert.Throws<SqlException>(
```
cheenamalhotra, line 694: SqlException is a very broad generic type. please also assert what exception you expect here, to ensure the behavior does not change in future and exception is always as "expected".

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/UdtTest/UdtTest2.cs
```diff
@@ -588,6 +588,128 @@ public void TestSqlUserDefinedAggregateAttributeMaxByteSize()
...
+                    }
+                    finally
+                    {
+                        try
+                        {
+                            transaction.Rollback();
+                        }
+                        catch
+                        {
+                            // ignore rollback failure exceptions to preserve original thrown error in test result
```
cheenamalhotra, line 706: Transaction rollback should not throw exception here. The previous `SqlException` is already handled too.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/UdtTest/UdtTest2.cs
Wraith2: Using the exception text is non-portable and means tests would only run under en locales. 
All the other parts of the test code should be covered under their own tests which would fail as well if behaviour changed. A false positive here would be the result of a real negative test elsewhere.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/UdtTest/UdtTest2.cs
Wraith2: We want the error from the com.ExecuteNonQuery to be the result of executing the command but if the transaction throws an exception because there's been an error rolling back because nothing was done then I want to ignore it, otherwise an exception thrown in the finally is the result instead of the exception from the try, it's confusing.

I'm not in favour of unreliable either but it's defensive. If something were expected it would be dealt with specifically. I'm pretty sure the test body was copied from another test and altered to fit this specific case, it was 2 years ago though do I can't be certain.

cheenamalhotra: But it should not throw exception (in ideal scenarios), so it should be removed. It's probably bad old practice to swallow unexpected behaviors..

I was also trying to run this test class manually and was facing some issues, did u manage to run this manually too?
Currently test lab does not run these tests, so relying on checks wouldn't be a good idea for UDT tests. That's one major limitation in our test suite we have in backlog.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/UdtTest/UdtTest2.cs
cheenamalhotra: As general practice we always compare message text in English lang only, and not entire message but the common most part of message should be asserted to be absolutely certain the flow doesn't change.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/UdtTest/UdtTest2.cs
Wraith2: So just take out the additional layer of error handling to do with the transaction you think?

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/UdtTest/UdtTest2.cs
Wraith2: You'll have to drop and re-create your udt test database using the updated script from this PR. otherwise you won't have the required entities in the database, as the comment at the top of the test says.

Yes, I ran both tests under commandline build and in visual studio successfully. Doing so the first time showed I'd made a mistake that needed correcting so I know they were run repeatedly under successful.

cheenamalhotra: Ok. I didn't notice the SQL file changes, will try that. Thanks! :)

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/UdtTest/UdtTest2.cs
Wraith2: I don't approve of testing in English only. I've changed the test to do so anyway.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/UdtTest/UdtTest2.cs
Wraith2: removed.

src/Microsoft.Data.SqlClient/netcore/src/Common/src/Microsoft/Data/Common/AdapterUtil.cs
```diff
@@ -379,24 +379,33 @@ internal static Exception StreamClosed([CallerMemberName] string method = "")
 
         internal static string BuildQuotedString(string quotePrefix, string quoteSuffix, string unQuotedString)
         {
-            var resultString = new StringBuilder();
+            var resultString = new StringBuilder(unQuotedString.Length + quoteSuffix.Length + quoteSuffix.Length);
+            AppendQuotedString(resultString, quotePrefix, quoteSuffix, unQuotedString);
+            return resultString.ToString();
```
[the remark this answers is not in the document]
Wraith2: The constructor argument changes the initial capacity so that a single allocation of at least that length is used for the first segment. If you know how much space you need it's a good idea to use this ctor. Even if you can't presize it's often a good idea to have a guess because the less segments you add the less heap fragmentation you cause. Additional allocations will use the internal sizing rules if they are added so defaults aren't changed.

There are no cons to returning a stringbuilder and having the caller deal with calling ToString on it. I chose to maintain the call pattern and return a string.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
```diff
@@ -3084,6 +3084,12 @@ internal void DeriveParameters()
                         p.TypeName = r[colNames[(int)ProcParamsColIndex.TypeCatalogName]] + "." +
                             r[colNames[(int)ProcParamsColIndex.TypeSchemaName]] + "." +
                             r[colNames[(int)ProcParamsColIndex.TypeName]];
+
+                        // the constructed type name above is incorrectly formatted, it should be a 2 part name not 3
+                        // for compatibility we can't change this because the bug has existed for a long time and been
```
JRahnama, line 3089: In general, I believe we have to address the bug inside the driver instead of bringing the work around inside.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlParameter.cs
```diff
@@ -669,7 +669,11 @@ public string UdtTypeName
         public string TypeName
         {
             get => _typeName ?? ADP.StrEmpty;
```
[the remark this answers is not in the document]
Wraith2: I haven't got there yet.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
Wraith2: If you look up the original fix in corefx you'll find that there was a lot of conversation around it. 

the reason it's fixed this late is that it's a long standing bug and if users have worked around it by assigning the TypeName directly themselves then we'd break their code. So what we do it follow the original pattern and leave it broken in case users are going to fix it and then at the very last chance before we're about to send the query to the server we check if they've fixed it themselves and if they haven't we fix it for them. It'd be simpler to fix it directly but we'd annoy a lot of people, this way everyone stays working and no-one new has the problem.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlParameter.cs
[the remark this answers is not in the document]
Wraith2: Interesting comment on the definition as well:
```csharp
        // security issue, don't rely upon public static readonly values
        internal static readonly string StrEmpty = ""; // String.Empty
```

If you can't rely on string.Empty being an empty string then a lot of the BCL and other libraries aren't going to run correctly. I'd be interested to know if you can find anyone who knows what prompted the comment.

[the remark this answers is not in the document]
ErikEJ: Is it not internal vs public ?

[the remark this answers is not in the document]
Wraith2: Perhaps but even that doesn't make sense. 

The literal "" in the local version is going to be interned by the runtime and will point at the single interned empty string. So if someone grabs hold of string.Empty and mutates it (which action leave you in an unsupported state) they'll also be altering this instance and both will be in incorrectly reached states caused by non permitted operations. 

I thought perhaps it was something to do with runtime guarantees around statics but those are specified by the standards and if they're broken the runtime is broken. At time of access the readonly static WILL but set. So I can't see that this guards anything apart from unsupported runtime alterations and if they're done that then the process/runtime/thread should crash.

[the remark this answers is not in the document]
JRahnama: > I'd be interested to know if you can find anyone who knows what prompted the comment.

I do not think that we have more than the history in visual studio to travel back in time and see who has added that comment, I will be searching for that, but maybe @saurabh500 can guide us to the correct direction.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
JRahnama: I am still hesitant to approve the PR although it looks good to me logic wise. for below reasons:

- It may cause some problems for unseen cases in future.
- It will need more debugging and understanding of this workaround logic, which we do not need to deal with if we fix the issue.
- We are fixing things for client without them knowing.

Will it be too much if ask you to make a draft/regular PR with the actual fix then we can decide on how we are going to proceed?
I will ask @David-Engel opinion on this as well.

Just to make this easy to follow, the issue #1001 was opened in Microsoft.Data.SqlClient Github and that was referring to a [problem in corefx](https://github.com/dotnet/corefx/pull/35549)

**System.Data.SqlClient has accepted the workaround.** and apparently the sample repro provided in issue #1001  works fine using that library.

Wraith2: The unconditional fix version would simply be to change the type name derivation to:
```csharp
p.TypeName = r[colNames[(int)ProcParamsColIndex.TypeSchemaName]] + "." +
                            r[colNames[(int)ProcParamsColIndex.TypeName]];
```
and add the covering test.

You're right that this method of fixing it was chosen with compatibility as a major factor. We have the chance now to make a breaking change. I don't have a strong opinion on that. Lets see what David thinks.

David-Engel: Reading through the old discussion on corefx about the behavior of the implemented fix and in-use workarounds, I'm good with approving this PR as-is.

JRahnama (approved):

cheenamalhotra (approved):

johnnypham (approved):

PR #1023, opened 2021-04-06
Fix wrong data blended with transactions in .NET core

DavoudEshtehari:
**root cause**: The wrong data occurs when a delegated transaction completes (rollback or commit) with open result sets. In this situation, we need a new physical state object to avoid the wrong data issue.

- The other changes came from a comparison with .NET Framework.


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlUtil.cs
```diff
@@ -786,6 +786,13 @@ internal static Exception SqlDependencyNoMatchingServerDatabaseStart()
         //
         // SQL.SqlDelegatedTransaction
         //
+        static internal Exception CannotCompleteDelegatedTransactionWithOpenResults(SqlInternalConnectionTds internalConnection, bool marsOn)
+        {
+            SqlErrorCollection errors = new SqlErrorCollection();
+            errors.Add(new SqlError(TdsEnums.TIMEOUT_EXPIRED, (byte)0x00, TdsEnums.MIN_ERROR_CLASS, null, (StringsHelper.GetString(Strings.ADP_OpenReaderExists, marsOn ? ADP.Command : ADP.Connection)), "", 0, TdsEnums.SNI_WAIT_TIMEOUT));
```
cheenamalhotra, line 792: I'm not entirely convinced why this should be called out as "Timeout" error, I do see the same in NetFx source, but it just doesn't feel accurate 🤔

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
```diff
@@ -3561,6 +3563,36 @@ private bool TryReadInternal(bool setTimeout, out bool more)
...
+            {
+                _isClosed = true;
+                SqlConnection con = _connection;
+                if (con != null)
+                {
+                    con.Abort(e);
+                }
+                throw;
+            }
+            catch (System.Threading.ThreadAbortException e)
```
Wraith2, line 3586: netcore doesn't have ThreadAbortException anymore. The type exists for compatibility but it will never be thrown. If this code will be merged for the two codebases perhaps leave it in and put a note that it's not needed in netcore

cheenamalhotra:
Is it possible to write a test case for this scenario if we can force reproduce it?

Wraith2:
Which part actually makes the difference? It looks like this re-enabled a commented out exception and adds an optional parameter to a function call. How does that resolve the issue, where do those changes track to and what to they cause to happen differently?

DavoudEshtehari:
> Which part actually makes the difference? It looks like this re-enabled a commented out exception and adds an optional parameter to a function call. How does that resolve the issue, where do those changes track to and what to they cause to happen differently?

Yes, that's the change that caused the fix. Raising that exception dooms the using connection while a transaction is aborted.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlUtil.cs
DavoudEshtehari: I agree. We'd better use a more accurate error, even though it's been used internally to add logs. 😊

Wraith2:
Ok. At least I'm not missing something then. Is there any clue why that line was commented out? It's a quite important line to accidentally remove.

DavoudEshtehari:
I couldn't find anything; it was commented out since the very beginning (August 2018 as I found) before publishing on GitHub. 😕

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -1118,7 +1118,7 @@ bool isDelegateControlRequest
                     }
                     else if (internalTransaction.OpenResultsCount != 0)
                     {
-                        //throw SQL.CannotCompleteDelegatedTransactionWithOpenResults(this);
+                        throw SQL.CannotCompleteDelegatedTransactionWithOpenResults(this, _parser.MARSOn);
```
cheenamalhotra, line 1121: How does this change impact if MARS is enabled? Can you try to reproduce issue with MARS on with this fix?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -1116,9 +1116,12 @@ bool isDelegateControlRequest
                         stateObj = _parser.GetSession(this);
                         mustPutSession = true;
                     }
-                    else if (internalTransaction.OpenResultsCount != 0)
+                    if (internalTransaction.OpenResultsCount != 0)
                     {
-                        //throw SQL.CannotCompleteDelegatedTransactionWithOpenResults(this);
+                        SqlClientEventSource.Log.TryTraceEvent("<sc.SqlInternalConnectionTds.ExecuteTransactionYukon|DATA|CATCH> {0}, Connection is going to doom cause of having OpenResultsCount {1} > 0.",
+                                                               ObjectID,
+                                                               internalTransaction.OpenResultsCount);
```
cheenamalhotra, line 1123:
```suggestion
                        SqlClientEventSource.Log.TryTraceEvent("<sc.SqlInternalConnectionTds.ExecuteTransactionYukon|DATA|CATCH> {0}, Connection is marked to be doomed when closed. Transaction ended with OpenResultsCount {1} > 0, MARSOn {2}",
                                                               ObjectID,
                                                               internalTransaction.OpenResultsCount,
                                                               _parser.MARSOn);
```

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -1380,8 +1380,11 @@ override internal void ExecuteTransaction(TransactionRequest transactionRequest,
                         stateObj = _parser.GetSession(this);
                         mustPutSession = true;
                     }
-                    else if (internalTransaction.OpenResultsCount != 0)
+                    if (internalTransaction.OpenResultsCount != 0)
                     {
+                        SqlClientEventSource.Log.TryTraceEvent("<sc.SqlInternalConnectionTds.ExecuteTransactionYukon|DATA|CATCH> {0}, Connection is going to doom cause of having OpenResultsCount {1} > 0.",
+                                                               ObjectID,
+                                                               internalTransaction.OpenResultsCount);
```
cheenamalhotra, line 1387:
```suggestion
                        SqlClientEventSource.Log.TryTraceEvent("<sc.SqlInternalConnectionTds.ExecuteTransactionYukon|DATA|CATCH> {0}, Connection is marked to be doomed when closed. Transaction ended with OpenResultsCount {1} > 0, MARSOn {2}",
                                                               ObjectID,
                                                               internalTransaction.OpenResultsCount,
                                                               _parser.MARSOn);
```

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/TransactionTest/TransactionTest.cs
```diff
@@ -10,6 +10,88 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests
...
+                    MaxPoolSize = 1
+                }.ConnectionString
+                , new SqlConnectionStringBuilder(DataTestUtility.TCPConnectionString)
+                {
+                    MultipleActiveResultSets = true
+                }.ConnectionString
+            };
+
+        // Synapse: Enforced unique constraints are not supported. To create an unenforced unique constraint you must include the NOT ENFORCED syntax as part of your statement.
+        [ConditionalTheory(typeof(DataTestUtility), nameof(DataTestUtility.AreConnStringsSetup), nameof(DataTestUtility.IsNotAzureSynapse))]
```
cheenamalhotra, line 34: I'm not sure if this applies to this test case, please let it run for Synapse. We'll know if it fails.
```suggestion
        [ConditionalTheory(typeof(DataTestUtility), nameof(DataTestUtility.AreConnStringsSetup))]
```

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/TransactionTest/TransactionTest.cs
```diff
@@ -10,6 +10,88 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests
...
+
+                using (SqlConnection sqlConnection = new SqlConnection(connString))
+                using (SqlCommand cmd = new SqlCommand("SELECT TOP(1) 4 Clm0 FROM sysobjects FOR XML AUTO", sqlConnection))
+                {
+                    sqlConnection.Open();
+                    using (System.Xml.XmlReader reader = cmd.ExecuteXmlReader())
+                    {
+                        bool result = reader.Read();
+                        Assert.True(result);
+                        if (result)
```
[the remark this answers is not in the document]
Wraith2: It will in release mode which is how we currently run the tests :(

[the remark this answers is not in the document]
cheenamalhotra: Are you confusing Xunit with Debug Assert?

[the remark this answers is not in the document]
DavoudEshtehari: If there were any confusion, it would raise an error on compile time. I don't see any confusion here since these two types come from different namespaces: `System.Diagnostics.Debug.Assert` and `xUnit.Assert`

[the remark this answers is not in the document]
Wraith2: Yup, Nothing to see here.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlUtil.cs
cheenamalhotra: Non-blocker for PR.

cheenamalhotra (approved):

johnnypham (approved):

JRahnama (approved):

PR #1024, opened 2021-04-08
Sync SqlDataRecord

Wraith2:
This is a small one. Identify similar parts of the two SqlDataRecord classes from netcore and netfx and sync those differences so the functional differences can be seen.

Observations:
1) There are only two constructors for SqlDataRecord and both of them set the _recordBuffer to a non-null value. This means that the frequent calls to `EnsureSubclassOverride` which checks _recordBuffer is non-null or throws are pointless and I have seen them in profile traces in the past when looking at UDT performance. I propose to remove `EnsureSubclassOverride` entirely.
2) All the private fields can be made readonly
3) `ThrowIfInvalidOrdinal(ordinal)` is used only once in `IsDBNull`, either it should be used in all functions that take an ordinal parameter or none of them. If it is not used ordinal will often be used without further checks in called methods which could result in out of range or key not found exceptions.
4) The functional differences between these two implementations seem to fall into two categories, in-proc hosting and old sql versions. The in-proc code could be kept in case it is needed in future and the old sql version checks could be dropped after checking version compatibility for this library.

Any of those changes would be a separate PR.

Oh. and the [docs](https://docs.microsoft.com/en-us/dotnet/api/microsoft.sqlserver.server.sqldatarecord?view=netframework-4.8) state that "Represents a single row of data and its metadata. This class cannot be inherited." when the class is not sealed and can be inherited from. Either the docs are wrong or the class should be sealed.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/Server/SqlDataRecord.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/Server/SqlDataRecord.cs
@@ -796,6 +796,7 @@ namespace Microsoft.Data.SqlClient.Server
         }
 
         /// <include file='../../../../../../../../doc/snippets/Microsoft.Data.SqlClient.Server/SqlDataRecord.xml' path='docs/members[@name="SqlDataRecord"]/System.Data.IDataRecord.GetData/*' />
+        [System.ComponentModel.EditorBrowsableAttribute(System.ComponentModel.EditorBrowsableState.Never)]
         IDataReader System.Data.IDataRecord.GetData(int ordinal)
         {
             throw ADP.NotSupported();
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/Server/SqlDataRecord.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/Server/SqlDataRecord.cs
@@ -21,12 +21,15 @@ namespace Microsoft.Data.SqlClient.Server
         private FieldNameLookup _fieldNameLookup;
         private bool _usesStringStorageForXml;
 
-        private static readonly SmiMetaData s_maxNVarCharForXml = new SmiMetaData(SqlDbType.NVarChar, SmiMetaData.UnlimitedMaxLengthIndicator,
-                                        SmiMetaData.DefaultNVarChar_NoCollation.Precision,
-                                        SmiMetaData.DefaultNVarChar_NoCollation.Scale,
-                                        SmiMetaData.DefaultNVarChar.LocaleId,
-                                        SmiMetaData.DefaultNVarChar.CompareOptions,
-                                        null);
+        private static readonly SmiMetaData s_maxNVarCharForXml = new SmiMetaData(
+            SqlDbType.NVarChar, 
+            SmiMetaData.UnlimitedMaxLengthIndicator,
+            SmiMetaData.DefaultNVarChar_NoCollation.Precision,
+            SmiMetaData.DefaultNVarChar_NoCollation.Scale,
+            SmiMetaData.DefaultNVarChar.LocaleId,
+            SmiMetaData.DefaultNVarChar.CompareOptions,
+            userDefinedType: null
+        );
 
         /// <include file='../../../../../../../../doc/snippets/Microsoft.Data.SqlClient.Server/SqlDataRecord.xml' path='docs/members[@name="SqlDataRecord"]/FieldCount/*' />
         public virtual int FieldCount
@@ -50,7 +53,7 @@ namespace Microsoft.Data.SqlClient.Server
         {
             EnsureSubclassOverride();
             SqlMetaData metaData = GetSqlMetaData(ordinal);
-            if (SqlDbType.Udt == metaData.SqlDbType)
+            if (metaData.SqlDbType == SqlDbType.Udt)
             {
                 return metaData.UdtTypeName;
             }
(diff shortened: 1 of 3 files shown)


cheenamalhotra:
> Oh. and the docs state that "Represents a single row of data and its metadata. This class cannot be inherited." when the class is not sealed and can be inherited from. Either the docs are wrong or the class should be sealed.

The original class in `Microsoft.SqlServer.Server` namespace also is not "sealed" so it does raise confusions. I would leave it as-is for now.

> There are only two constructors for SqlDataRecord and both of them set the _recordBuffer to a non-null value. This means that the frequent calls to EnsureSubclassOverride which checks _recordBuffer is non-null or throws are pointless and I have seen them in profile traces in the past when looking at UDT performance. I propose to remove EnsureSubclassOverride entirely.

Agreed. Also the other constructor (internal) does not seem to be used and is dead code - can be removed.

> All the private fields can be made readonly

Go ahead, except "_fieldNameLookup" I see others are initialized in constructor so it's possible.

> ThrowIfInvalidOrdinal(ordinal) is used only once in IsDBNull, either it should be used in all functions that take an ordinal parameter or none of them. If it is not used ordinal will often be used without further checks in called methods which could result in out of range or key not found exceptions.

I agree it should be checked for in all functions fetching for ordinal.

> The functional differences between these two implementations seem to fall into two categories, in-proc hosting and old sql versions. The in-proc code could be kept in case it is needed in future and the old sql version checks could be dropped after checking version compatibility for this library.

SQL Server 2012 is the minimum official supported version as per [Support Lifecycle](https://docs.microsoft.com/en-us/sql/connect/ado-net/sqlclient-driver-support-lifecycle?view=sql-server-ver15#sql-version-compatibility-with-microsoftdatasqlclient). But we have customers still using SQL 2008 so we try not to break them. But checks to support below SQL 2005 can be considered for removal.

Wraith2:
Ok, as stated I'll make those changes in a follow-up unless you strongly prefer I do them here.

johnnypham (approved):

DavoudEshtehari:
> Oh. and the [docs](https://docs.microsoft.com/en-us/dotnet/api/microsoft.sqlserver.server.sqldatarecord?view=netframework-4.8) state that "Represents a single row of data and its metadata. This class cannot be inherited." when the class is not sealed and can be inherited from. Either the docs are wrong or the class should be sealed.

According to documents, nobody's allowed to inherit this class; but the design doesn't prevent it. So, there is a chance of inheriting from this class by customers!
@David-Engel Could you specify your preference here, please?

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/Server/SqlDataRecord.cs
```diff
@@ -871,13 +862,20 @@ internal void ThrowIfInvalidOrdinal(int ordinal)
...
             {
                 throw SQL.SubclassMustOverride();
             }
         }
+
+        /// <include file='../../../../../../../../doc/snippets/Microsoft.Data.SqlClient.Server/SqlDataRecord.xml' path='docs/members[@name="SqlDataRecord"]/System.Data.IDataRecord.GetData/*' />
+        IDataReader System.Data.IDataRecord.GetData(int ordinal)
+        {
+            throw ADP.NotSupported();
+        }
```
DavoudEshtehari, line 878: Just inserting the removed attribute would be useful; Could you include it in Net core too?
```suggestion
        /// <include file='../../../../../../../../doc/snippets/Microsoft.Data.SqlClient.Server/SqlDataRecord.xml' path='docs/members[@name="SqlDataRecord"]/System.Data.IDataRecord.GetData/*' />
        [System.ComponentModel.EditorBrowsableAttribute(System.ComponentModel.EditorBrowsableState.Never)]
        IDataReader System.Data.IDataRecord.GetData(int ordinal)
        {
            throw ADP.NotSupported();
        }
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/Server/SqlDataRecord.cs
```diff
@@ -21,12 +21,15 @@ public class SqlDataRecord : IDataRecord
         private FieldNameLookup _fieldNameLookup;
         private bool _usesStringStorageForXml;
```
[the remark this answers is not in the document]
Wraith2: https://github.com/dotnet/SqlClient/pull/1024#issuecomment-816323952
> Ok, as stated I'll make those changes in a follow-up unless you strongly prefer I do them here.

David-Engel:
> > Oh. and the [docs](https://docs.microsoft.com/en-us/dotnet/api/microsoft.sqlserver.server.sqldatarecord?view=netframework-4.8) state that "Represents a single row of data and its metadata. This class cannot be inherited." when the class is not sealed and can be inherited from. Either the docs are wrong or the class should be sealed.
> 
> According to documents, nobody's allowed to inherit this class; but the design doesn't prevent it. So, there is a chance of inheriting from this class by customers!
> @David-Engel David Engel (Simba Technologies Inc) Vendor Could you specify your preference here, please?

Docs should be updated to reflect reality. That doesn't need to be a part of this PR, though.

cheenamalhotra:
@Wraith2 

> Docs should be updated to reflect reality.

Could you please update docs as well in this PR, to close on this topic and address one last comment from Davoud?
It'll be better to get all changes addressed together IMO.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/Server/SqlDataRecord.cs
Wraith2: Done in core and netfx.

DavoudEshtehari (approved):

PR #1028, opened 2021-04-11
Share SqlStatistics

Wraith2:
This moves the netcore `SqlStatistics` implementation to shared and applies some minor style changes.

It also adds a new ref struct called ValueSqlStatisticsScope and a factory method on `SqlStatistics.TimedScope` to create them. This will enable the use of statistics to be made less verbose. From

```csharp
override public string GetDataTypeName(int i)
{
    SqlStatistics statistics = null;
    try
    {
        CheckMetaDataIsReady(columnIndex: i);

        return GetDataTypeNameInternal(_metaData[i]);
    }
    finally
    {
        SqlStatistics.StopTimer(statistics);
    }
}
```

to:
```csharp
override public string GetDataTypeName(int i)
{
    using (SqlStatistics.TimedScope(Statistics))
    {
        CheckMetaDataIsReady(columnIndex: i);

        return GetDataTypeNameInternal(_metaData[i]);
    }
}
```

The struct is a ref struct for two reasons. 1) it prevents it being included in async closures which is good because async implementations don't start and end in the same method in this library implementation most of the time. 2) is keeps it on the stack in the same way that the original variable was used again preventing leakage.

The reason for the struct being ref and not implementing IDispose despite being used in a using block are noted at the type definition. ref structs can't implement interfaces because interface dispatch requires the variable to be boxed and ref types can't be. The compiler uses pattern matching on ref structs and the dispose interface method so that it can be used in using blocks idiomatically.

The changes to consumers to use the new scope will be done in a separate PR.


src/Microsoft.Data.SqlClient/netfx/src/Microsoft.Data.SqlClient.csproj
```diff
@@ -540,10 +542,10 @@
   </ItemGroup>
   <!-- Resources  -->
   <ItemGroup>
-    <Compile Include="Resources\$(ResxFileName).Designer.cs">
+    <Compile Include="Resources\Strings.Designer.cs">
       <AutoGen>True</AutoGen>
       <DesignTime>True</DesignTime>
-      <DependentUpon>$(ResxFileName).resx</DependentUpon>
+      <DependentUpon>Strings.resx</DependentUpon>
```
cheenamalhotra, line 548: Please ignore these changes

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlStatistics.cs
```diff
@@ -335,4 +337,16 @@ void ICollection.CopyTo(Array array, int arrayIndex)
...
+    // This is a ref struct to prevent it being included in async closures accidentally. 
+    // Async functions should manage the timer directly using the Start and Stop method
+    // in their invoke and completion functions
+    internal readonly ref struct ValueSqlStatisticsScope // : IDisposable // ref structs cannot implement interfaces but the compiler will use pattern matching to allow use of using on them
+    {
+        private readonly SqlStatistics _statistics;
+
+        public ValueSqlStatisticsScope(SqlStatistics statistics) => _statistics = SqlStatistics.StartTimer(statistics);
+
+        public void Dispose() => SqlStatistics.StopTimer(_statistics);
```
cheenamalhotra, line 350: Could you implement the dispose pattern here, i.e. `public void Dispose() => Dispose(true)`?

Wraith2: Yes but it would have no value. A ref struct cannot ever have a finalizer or be present on the finalization heap so there is no need to distinguish between disposal and finalization. It also holds no native resources that would require free-ing if dropped to GC so it's no a useful pattern in general.

Wraith2: https://docs.microsoft.com/en-us/dotnet/api/system.idisposable?view=net-5.0#implementing-idisposable

src/Microsoft.Data.SqlClient/netfx/src/Microsoft.Data.SqlClient.csproj
Wraith2: Done, I hope this isn't another case where the designer is going to keep making undesired change I have to revert.

cheenamalhotra (approved):

johnnypham (approved):

PR #1015, opened 2021-03-31
Add IP address preference support for TCP connection

karinazhou:
A new connection property `IPAddressPreference` is added to specify the IP preference by the client application when doing TCP connection. There are three valid types:

- IPv4First
  This is the default preference. The driver will traverse IPv4 addresses first. If none of them can be connected successfully, it continues to try IPv6 addresses if there are any.

- IPv6First
  The driver will traverse IPv6 addresses first. If none of them can be connected successfully, it continues to try IPv4 addresses if there are any.

- UsePlatformDefault
  The driver will traverse all the IP addresses in their initial orders from DNS resolution.

To keep the consistency between managed SNI and native SNI, the connection logic is changed in `SNITcpHandle:Connect()`. Instead of keeping the last IPv4 and IPv6 addresses, all the valid IP addresses returned from DNS resolution will be tested in a row. There will be a separate PR containing this change in parallel. 

NOTE:  This PR requires the change in native SNI so it requires a new version of Microsoft.Data.SqlClient.SNI to be released first.

cc @johnnypham  @cheenamalhotra  @David-Engel  @saurabh500

Issue #583, opened 2020-05-29
SqlConnection.Open() is not finished with success to specified Failover Partner on linux.


alexanderinochkin:
### The main problem is:

We use sql connection string with Failover Partner server and Connect Timeout = 30.
The program runs on linux (Ubuntu 19.10).
When Data Source server is not available (server turned off or blocked by firewall), 
SqlConnection.Open() does not connect to Failover Partner server despite it is available and act as SQL principal.  

### The reason description:
As it was discovered, the problem is at SNITCPHandle.Connect(...) procedure: It does not terminate connection at specified timeout.
https://github.com/dotnet/SqlClient/blob/3ca39848735143055d9d7d4864d5f1bfd1976b99/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs

It uses CancellationTokenSource with Cancel() callback, but it does not really cancel socket.Connect() operation.

**Here is the part of code with additional debug logging:**
 
```
private static Socket Connect(string serverName, int port, TimeSpan timeout, bool isInfiniteTimeout)
{
	var sw = Stopwatch.StartNew();

	...

	CancellationTokenSource cts = null;

	void Cancel()
	{
		for (int i = 0; i < sockets.Length; ++i)
		{
			try
			{
				if (sockets[i] != null && !sockets[i].Connected)
				{
					Console.WriteLine($"{sw.Elapsed} Disposing socket {i}...");
					sockets[i].Dispose();
					Console.WriteLine($"{sw.Elapsed} Disposed {i}");
					sockets[i] = null;
				}
			}
			catch { }
		}
	}

	if (!isInfiniteTimeout)
	{
		cts = new CancellationTokenSource(timeout);
		cts.Token.Register(Cancel);
	}

	Socket availableSocket = null;
	try
	{
		for (int i = 0; i < sockets.Length; ++i)
		{
			try
			{
				if (ipAddresses[i] != null)
				{
					sockets[i] = new Socket(ipAddresses[i].AddressFamily, SocketType.Stream, ProtocolType.Tcp);
					Console.WriteLine($"{sw.Elapsed} Connecting socket {i}...");
					sockets[i].Connect(ipAddresses[i], port);
					Console.WriteLine($"{sw.Elapsed} Connect socket finished {i}.");
					
					...
				}
			}
[...]

Issue #449, opened 2020-02-26
Azure SQL:  System.ObjectDisposedException: Cannot access a disposed object on OpenAsync


PashaPash:
### Describe the bug

On Azure SQL, after short server unavailability / failover, subsequent call to OpenAsync fails with  `System.ObjectDisposedException: Cannot access a disposed object.`

Stack trace:

```
Microsoft.Data.SqlClient.SqlException (0x80131904): A network-related or instance-specific error occurred while establishing a connection to SQL Server. The server was not found or was not accessible. Verify that the instance name is correct and that SQL Server is configured to allow remote connections. (provider: TCP Provider, error: 35 - An internal exception was caught)
 ---> System.ObjectDisposedException: Cannot access a disposed object.
Object name: 'System.Net.Sockets.Socket'.
   at System.Net.Sockets.Socket.SetSocketOption(SocketOptionLevel optionLevel, SocketOptionName optionName, Int32 optionValue)
   at Microsoft.Data.SqlClient.SNI.SNITCPHandle..ctor(String serverName, Int32 port, Int64 timerExpire, Object callbackObject, Boolean parallel)
   at Microsoft.Data.ProviderBase.DbConnectionPool.CheckPoolBlockingPeriod(Exception e)
   at Microsoft.Data.ProviderBase.DbConnectionPool.CreateObject(DbConnection owningObject, DbConnectionOptions userOptions, DbConnectionInternal oldConnection)
   at Microsoft.Data.ProviderBase.DbConnectionPool.UserCreateRequest(DbConnection owningObject, DbConnectionOptions userOptions, DbConnectionInternal oldConnection)
   at Microsoft.Data.ProviderBase.DbConnectionPool.TryGetConnection(DbConnection owningObject, UInt32 waitForMultipleObjectsTimeout, Boolean allowCreate, Boolean onlyOneCheckConnection, DbConnectionOptions userOptions, DbConnectionInternal& connection)
   at Microsoft.Data.ProviderBase.DbConnectionPool.WaitForPendingOpen()
```

Exception is re-thrown with `throw e` at `CheckPoolBlockingPeriod` for Azure SQL connections, so only two first lines seems to be related. 
`MultiSubnetFailover` attribute is not set, so it probably follows non-parallel connect path.

### To reproduce
Unable to clearly reproduce it locally - seems to happen on Azure only, and during (or immediately after) Azure SQL downtime.

### Expected behavior
OpenAsync either fails with real network-related error or connects successfully.

[...]

PR #1029, opened 2021-04-13
SqlClient-826 Missed synchronization

### LINKED ISSUE #826: Missed synchronization
While investigating other issues I’ve found probably few hard to reproduce issue.

I haven’t seen them happening, but I’ve seen several issues which are claimed as hard to reproduce/happens under heavy load, so may be it could help there. In original PR (where code has been introduced) that was mentioned also.

This block accessing **sockets** array https://github.com/dotnet/SqlClient/blob/8ad8da6bc1f7314ec37b5fda1b4e991a878a6072/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs#L341
is not synchronized with this one https://github.com/dotnet/SqlClient/blob/8ad8da6bc1f7314ec37b5fda1b4e991a878a6072/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs#L368
The following order seems is possible:
1.	if (sockets[i] != null && !sockets[i].Connected) is true
2.	<Here socket got opened>
3.	if (sockets[i] != null) gives true
4.	if (sockets[i].Connected) gives true
5.	sockets[i].Dispose();
6.	availableSocket = sockets[i];
7.	break; return availableSocket;

In this case socket is first returned and then at random time disposed while being used (user does not receive any exceptions, connection just disposes).
But there are also many hard investigate synchronization/exceptions things there. I think may be that entire file and related should be checked/reviewed again.

jinek:
Fixes #826

Internal method SNITCPHandle.Connect now uses [Socket.Select](https://docs.microsoft.com/en-us/dotnet/api/system.net.sockets.socket.select?view=net-5.0) to connect the socket with a timeout or without

Additionally:

* Exceptions swallowing removed to satisfy CA1031 https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1031
* InternalException refactored to satisfy ca1032 https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1032 (Best practices https://docs.microsoft.com/en-us/dotnet/standard/exceptions/best-practices-for-exceptions#include-three-constructors-in-custom-exception-classes )
* InternalException constructor has been changed to public as the class is not marked as sealed.
* Logic of https://github.com/dotnet/SqlClient/pull/1016 refactored for simplicity and to avoid mentioned issues.
* Check for null of ipAddresses removed as **Dns.GetHostAddresses** does not return null


jinek:
@cheenamalhotra I can't add label "Managed SNI". Could you, please? This comment can be removed then

JRahnama:
@jinek we have PR #1016 in progress. There might be some conflict over these two changes. If that happens, please wait for that PR to merge and pull the changes again.

jinek:
@JRahnama I've took into account that PR and merged it. However I allowed myself to slightly fix that PR while keeping the idea of it.
@karinazhou Could you please check this PR to be sure I adopted your idea correctly.

JRahnama:
@jinek thanks for contributing into this repository, We have a release scheduled for this week and pretty busy with release activities. We will get back to this next week when we have more time.
Thanks again.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
```diff
@@ -322,112 +324,76 @@ private Socket TryConnectParallel(string hostName, int port, TimeSpan ts, bool i
...
-            IPAddress[] ipAddresses = Dns.GetHostAddresses(serverName);
+            // keeping IEnumerable to enumerate entire array only if needed
+            IEnumerable<IPAddress> ipAddresses = Dns.GetHostAddresses(serverName);
+            
+            // First, we try to connect all AddressFamily.InterNetwork
+            // and then all AddressFamily.InterNetworkV6.
+            // Keeping address families inlined in entire method because parameter pendingDNSInfo is bound on them.
+            ipAddresses = ipAddresses
+                .Where (address => address.AddressFamily==AddressFamily.InterNetwork)
+                .Concat(ipAddresses.Where(address => address.AddressFamily==AddressFamily.InterNetworkV6));
```
ilmax, line 335: Can't this be achieved with an `OrderBy`? Also if this is on a hot path maybe linq should be avoided

jinek: It can be achieved with OrderBy, I made so firstly: https://github.com/dotnet/SqlClient/pull/1029/commits/0021191c41c0ce8e5f5b091426961b7351aa05fe
But, that is more code with more enumeration. While this one is shorter and does not enumerate everything upon needed.
This is not hot path, we use exception/async here as well and even thrown exception does not play any difference here.

jinek: Here some information about performance https://github.com/dotnet/SqlClient/issues/826#issuecomment-739660666

tkhodzhaev (approved):

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
jinek: @DavoudEshtehari Helpful for what?
I have updated this PR with https://github.com/dotnet/SqlClient/pull/1016 already.
I don't see anything else in 1015 , which could be in contrast with this PR. That code had same issues which already discussed in 1016

DavoudEshtehari: I understand you're trying to improve the socket establishment here, and our focus on PR [#1015](https://github.com/dotnet/SqlClient/pull/1015) is to add a new configuration not performance improvement. The new IP preference feature uses the resolved IPs from dns in original order that is important especially for the **default preference**. This is a friendly point should be considered here if there is any chance of reordering the IP addresses. For more explanation you can check [this comment](https://github.com/dotnet/SqlClient/pull/1015#issuecomment-812609864).

jinek: @DavoudEshtehari  I see. Well, this PR is not bound on socket selecting method. The point of this PR is to fix issues listed under #826 , for example #449 , or for instance to fix timeout issue described in #1016 . Any IP addresses ordering is fine for this PR.
What I'm saying, is I've already changed the addresses order once to match #1016 and it can be updated again to do any ordering.
However, as I see it, it's easier to merge this PR with any ordering we want and then to update other conflicting PRs, because this one makes code 20% shorter:
image
Anyway, I think, it does not make sense to update this PR each time we change ordering in main branch. Probably it would be more efficient to update this PR once right before the merge.

DavoudEshtehari: I think it would be a good time to prepare your final change if you had time to back to this PR.

jinek:
I've updated the PR, @DavoudEshtehari  .
But, requesting additional attention to two things:
1) Here additional improvement can be done. For infinite timeout case we can avoid exception throwing which would give small performance boost, but code would become more complex. Should I add the code for infinite timeout case (it's about 3 additional if blocks)?
2) I've removed the check of **Dns.GetHostAddresses** for returning null as seems it does not ever return null. However it's not declared by the documentation of the method and I don't understand where it is supposed always return an array or no. Should I brind back the code considering it returning null? (probably just adding one redundent if block)

DavoudEshtehari (commented):
I ran the failed pipelines one more time, and the result remained the same. The reason for these failures as I know is the execution time exceeds the allowed 1 hour. First of all, we need to verify in which situations it happens; Maybe running a benchmark test could help us too!

> 1. Here additional improvement can be done. For infinite timeout case we can avoid exception throwing which would give small performance boost, but code would become more complex. Should I add the code for infinite timeout case (it's about 3 additional if blocks)?

Feel free to add your suggested change for further discussions.

> 2. I've removed the check of **Dns.GetHostAddresses** for returning null as seems it does not ever return null. However it's not declared by the documentation of the method and I don't understand where it is supposed always return an array or no. Should I brind back the code considering it returning null? (probably just adding one redundent if block)

Right, I checked the source code of the **Dns.GetHostAddresses**, it won't return `null` in case of a free-error execution. So, as you mentioned, null-checking is not necessary.

jinek:
> I ran the failed pipelines one more time, and the result remained the same. The reason for these failures as I know is the execution time exceeds the allowed 1 hour. First of all, we need to verify in which situations it happens; Maybe running a benchmark test could help us too!

I have checked the issue: here the connection timeout is set to almost 3 hours: 
https://github.com/dotnet/SqlClient/blob/d65dc864bc2f72321073ad79ed8e13973c0cf8f0/src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/TestFixtures/Setup/CertificateUtility.cs#L298
Same time SqlInternalConnectionTds runs retry for all random errors: https://github.com/dotnet/SqlClient/blob/2ddf3a36b036b60c1c8431875cab277e10f35433/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs#L1618

In my case the error is "Arithmetic operation resulted in an overflow." because 3 hours (10000 seconds) can not be passed to Socket.Select as it requires the value in microseconds: https://github.com/dotnet/runtime/blob/9ce467fe6164e781b67897615ff7dd01820084ff/src/libraries/System.Net.Sockets/src/System/Net/Sockets/Socket.cs#L2199

@DavoudEshtehari 
The question now is if it is fine to limit Connection Timeout to 35 minutes (int.Max microseconds) or must we keep 35000 minutes (int.Max seconds)?
If we have to keep possibility of the timeout to be more than 35 minutes than I can implement "async" alternative as mentioned here https://github.com/dotnet/SqlClient/issues/826#issuecomment-739660666

jinek:
Also fixes this: https://github.com/dotnet/SqlClient/issues/583

cheenamalhotra:
Looking at the [SocketPal.Select](https://github.com/dotnet/runtime/blob/9ce467fe6164e781b67897615ff7dd01820084ff/src/libraries/System.Net.Sockets/src/System/Net/Sockets/SocketPal.Windows.cs#L888) API, it seems they could do well with a `long` type of `microseconds` parameter, but because it's `int`. we have to limit our timeout?

Connection timeout cannot be limited as that would be a major issue for applications (we know of) running high loads of work and setting timeout to as long as 2 hours.

jinek:
> Looking at the [SocketPal.Select](https://github.com/dotnet/runtime/blob/9ce467fe6164e781b67897615ff7dd01820084ff/src/libraries/System.Net.Sockets/src/System/Net/Sockets/SocketPal.Windows.cs#L888) API, it seems they could do well with a `long` type of `microseconds` parameter, but because it's `int`. we have to limit our timeout? _(Timeout in milliseconds should always be `long`/`double` IMHO)_
> 
> There's also an open API Proposal from 6 years ago, you might want to ask dotnet/runtime teams to resurrect: [dotnet/runtime#14336](https://github.com/dotnet/runtime/issues/14336)
> 
> Connection timeout cannot be limited as that would be a major issue for customer applications (we know of) running high loads of work and setting timeout to as long as 2 hours.

Right right, but all this options are time consuming? 
Does socket.Connect really have infinite or any big timeout? Im thinking may be it is already limited, i would just play with the math to leave it as is.
Otherwise,  i can bring Dispose similar as it was before, but with synchronisation.

cheenamalhotra:
> Does socket.Connect really have infinite or any big timeout?

As of now, we use default "Infinite" Socket SendTimeout. It's not changed by driver. We perform cancellation after "timeout" specified as "Connection Timeout" which does allow longer timeouts. If Connection Timeout is 0 - we don't set cancellation token. Looking at this design, it seems current design allows all the timeout limit on socket connect. 

Because [Socket.SendTimeout](https://docs.microsoft.com/en-us/dotnet/api/system.net.sockets.socket.sendtimeout?view=net-5.0) itself is "_int_", there's no option to keep that consistent with "Connection Timeout" and that will be a major behavior change. So let's explore other options.

jinek:
> 
> 
> > Does socket.Connect really have infinite or any big timeout?
> 
> As of now, we use default "Infinite" Socket SendTimeout. It's not changed by driver. We perform cancellation after "timeout" specified as "Connection Timeout" which does allow longer timeouts. If Connection Timeout is 0 - we don't set cancellation token. Looking at this design, it seems current design allows all the timeout limit on socket connect.
> 
> Because [Socket.SendTimeout](https://docs.microsoft.com/en-us/dotnet/api/system.net.sockets.socket.sendtimeout?view=net-5.0) itself is "_int_", there's no option to keep that consistent with "Connection Timeout" and that will be a major behavior change. So let's explore other options.

FYI: I have tried to setup a port which drops packets: 51.15.66.212:30311
You can try to connect to it from linux machine with the command 
```
echo >/dev/tcp/51.15.66.212/30311
```
(note that linux running from docker for windows will probably use windows stack and timeout can be defined by hots OS)
Linux machine gives ```-bash: connect: Connection timed out
-bash: /dev/tcp/51.15.66.212/30311: Connection timed out``` in about 3 minutes.

I've tried to run this application from linux:
```
        {
            var watch = Stopwatch.StartNew();
            using var socket = new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp);
            socket.SetSocketOption(SocketOptionLevel.Socket, SocketOptionName.KeepAlive, true);
            socket.SetSocketOption(SocketOptionLevel.Tcp, SocketOptionName.TcpKeepAliveInterval, 1);
            socket.SetSocketOption(SocketOptionLevel.Tcp, SocketOptionName.TcpKeepAliveTime, 30);
            
            const string host = "51.15.66.212";
            Console.WriteLine($"Connecting to {host}");
            try
            {
                socket.Connect(host, int.Parse(args[0]));
                Console.WriteLine($"{socket.Connected}");
            }
            catch (SocketException socketException)
            {
                Console.WriteLine(socketException);
            }
            watch.Stop();
            Console.WriteLine(watch.Elapsed);

            Console.WriteLine("finished");
            Console.ReadLine();
        }

```
and it also gives timeout in about 3 minutes:
```
Connecting to 51.15.66.212
System.Net.Internals.SocketExceptionFactory+ExtendedSocketException (110): Connection timed out 51.15.66.212:30311
   at System.Net.Sockets.Socket.DoConnect(EndPoint endPointSnapshot, SocketAddress socketAddress)
   at System.Net.Sockets.Socket.Connect(EndPoint remoteEP)
   at System.Net.Sockets.Socket.Connect(IPAddress address, Int32 port)
   at System.Net.Sockets.Socket.Connect(String host, Int32 port)
   at ConsoleApp1.Program.Main(String[] args) in /src/ConsoleApp1/Program.cs:line 24
00:02:11.1637806
finished
```

Which means Connect uses system default timeout. Changes to Socket.SendTimeout and Socket.RecieveTimeout don't have effects. I've found more information here: 
https://stackoverflow.com/questions/1045964/how-to-view-change-socket-connection-timeout-on-linux
--------
Same time I see possibility to use Socket.Select to have bigger timeouts, I just need to call it several times each 35 minutes

jinek:
@cheenamalhotra Could you, please, help with the build. I see this checks fail because of some build error. Same I see on another PR https://github.com/dotnet/SqlClient/pull/1166/checks?check_run_id=3052700899 for example.
Also it shows the error that SqlConnection.RegisterColumnEncryptionKeyStoreProviders can be called once, but it is used in several tests which I don't see any sense about.
What to do with this?

cheenamalhotra:
Hi @jinek 

Yes we're aware of a few random failures coming over in the repository, one of them is related to https://github.com/dotnet/project-system/issues/4781 which is captured randomly when building tests.

Please avoid worrying about those, we'll rerun pipelines to address those errors.

jinek:
There were conflicting changes again: https://github.com/dotnet/SqlClient/commit/c9aa89fc7a5df326855f2ea5bbe6425daf378db5#r53444146

Just resolved the conflicts.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
```diff
@@ -5,7 +5,9 @@
 using System;
 using System.Collections.Generic;
 using System.ComponentModel;
+using System.Diagnostics;
 using System.IO;
+using System.Linq;
```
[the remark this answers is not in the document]
jinek: fixed

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
```diff
@@ -347,149 +349,147 @@ private Socket TryConnectParallel(string hostName, int port, TimeSpan ts, bool i
...
+                        if (!isInfiniteTimeout)
+                            throw SQL.SocketDidNotThrow();
+                        
+                        isConnected = true;
+                    }
+                    catch (SocketException socketException) when (!isInfiniteTimeout &&
+                                                                  socketException.SocketErrorCode ==
+                                                                  SocketError.WouldBlock)
+                    {
+                        // https://github.com/dotnet/SqlClient/issues/826#issuecomment-736224118
```
DavoudEshtehari, line 437: Please, include the outcome of the discussions as the comments in the code. For reference, I'd put links as review notes for reviewers.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
```diff
@@ -347,149 +349,147 @@ private Socket TryConnectParallel(string hostName, int port, TimeSpan ts, bool i
...
+
+                        List<Socket> checkReadLst; List<Socket> checkWriteLst; List<Socket> checkErrorLst;
+
+                        // Repeating Socket.Select several times if our timeout is greater
+                        // than int.MaxValue microseconds because of https://github.com/dotnet/SqlClient/pull/1029#issuecomment-875364044
```
DavoudEshtehari, line 442: Same note here.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
```diff
@@ -347,149 +349,147 @@ private Socket TryConnectParallel(string hostName, int port, TimeSpan ts, bool i
...
+                    if (isConnected)
                     {
-                        break;
+                        socket.Blocking = true;
+                        string iPv4String = null;
+                        string iPv6String = null;
+                        string ipAddressString = ipAddress.ToString();
+                        if (socket.AddressFamily == AddressFamily.InterNetwork)
+                            iPv4String = ipAddressString;
+                        else iPv6String = ipAddressString;
```
DavoudEshtehari, line 473: A minor memory improvement.
```suggestion
                        if (socket.AddressFamily == AddressFamily.InterNetwork)
                        {
                            iPv4String = ipAddress.ToString();
                        }
                        else
                        {
                            iPv6String = ipAddress.ToString();
                        }
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
```diff
@@ -347,149 +349,147 @@ private Socket TryConnectParallel(string hostName, int port, TimeSpan ts, bool i
...
-                        SqlClientEventSource.Log.TrySNITraceEvent(nameof(SNITCPHandle), EventType.ERR, "THIS EXCEPTION IS BEING SWALLOWED: {0}", args0: e?.Message);
-                    }
+                    SqlClientEventSource.Log.TrySNITraceEvent(nameof(SNITCPHandle), EventType.ERR, "THIS EXCEPTION IS BEING SWALLOWED: {0}", args0: e?.Message);
+                    SqlClientEventSource.Log.TryAdvancedTraceEvent(
+                        $"{nameof(SNITCPHandle)}.{nameof(Connect)}{EventType.ERR}THIS EXCEPTION IS BEING SWALLOWED: {e}");
+                }
+                finally
+                {
+                    if (!isSocketSelected)
+                        socket.Dispose();
```
DavoudEshtehari, line 488:
```suggestion
                    if (!isSocketSelected)
                    {
                        socket?.Dispose();
                    }
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
```diff
@@ -347,149 +349,147 @@ private Socket TryConnectParallel(string hostName, int port, TimeSpan ts, bool i
...
-            string IPv6String = null;
+            IEnumerable<IPAddress> ipAddresses = GetHostAddressesSortedByPreference(serverName, ipPreference);
 
-            // Returning null socket is handled by the caller function.
-            if (ipAddresses == null || ipAddresses.Length == 0)
+            foreach (IPAddress ipAddress in ipAddresses)
             {
-                return null;
-            }
+                var socket = new Socket(ipAddress.AddressFamily, SocketType.Stream, ProtocolType.Tcp)
```
DavoudEshtehari, line 404: `Socket`'s constructor can cause a `SocketException` and should be handled.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
jinek: May be by doing this we increase the penalty of failed branch prediction by CPU?
I'm not confident in low level optimizations, but code duplication looks strange to me.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
jinek: > `Socket`'s constructor can cause a `SocketException` and should be handled.

nice one!

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
DavoudEshtehari: I believe an immutable object can't be a threat in this case.

DavoudEshtehari (approved):
Well done 🎉

cheenamalhotra (approved):

PR #1031, opened 2021-04-13
Fix Socket array length to avoid possible out of bounds errors

cheenamalhotra:
Following discussion from https://github.com/dotnet/SqlClient/pull/1016.. 
This change fixes possible out of bounds errors and timeout cancellation issues.
Further improvements to be reviewed in https://github.com/dotnet/SqlClient/pull/1029.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
@@ -327,7 +327,7 @@ namespace Microsoft.Data.SqlClient.SNI
             string IPv4String = null;
             string IPv6String = null;
 
-            Socket[] sockets = new Socket[2];
+            Socket[] sockets = new Socket[ipAddresses.Length];
             AddressFamily[] preferedIPFamilies = new AddressFamily[] { AddressFamily.InterNetwork, AddressFamily.InterNetworkV6 };
 
             CancellationTokenSource cts = null;
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
@@ -327,6 +327,12 @@ namespace Microsoft.Data.SqlClient.SNI
             string IPv4String = null;
             string IPv6String = null;
 
+            // Returning null socket is handled by the caller function.
+            if(ipAddresses == null || ipAddresses.Length == 0)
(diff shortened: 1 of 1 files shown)


Wraith2:
Is `GetHostAddresses` required to return a non-null value even if it can't resolve? If not you need to add a null check and probably early exit.

Also, while you're there... any chance of hoisting that AddressFamily array out as a static readonly so we can avoid allocating and discarding it every single time we make a connection?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
```diff
@@ -327,7 +327,7 @@ private static Socket Connect(string serverName, int port, TimeSpan timeout, boo
             string IPv4String = null;
             string IPv6String = null;
 
-            Socket[] sockets = new Socket[2];
+            Socket[] sockets = new Socket[ipAddresses.Length];
```
jinek, line 330: You probably also need to use another incrementor inside that double loop, because now it is `sockets[i]` where `i` is `for (int i = 0; i < preferedIPFamilies.Length; ++i)`

cheenamalhotra:
@Wraith2 regarding AddressFamily suggestion, there's another PR bringing in preferred family support: https://github.com/dotnet/SqlClient/pull/1015/files#diff-11cfbefb2cd1b54e3dfda7a8eeb5d1b447086f5dbb41204151929886c29ae99fR345 so it won't be a static.

Wraith2:
Ok, LGTM then.

johnnypham (approved):

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
```diff
@@ -327,7 +327,13 @@ private static Socket Connect(string serverName, int port, TimeSpan timeout, boo
...
             string IPv6String = null;
 
-            Socket[] sockets = new Socket[2];
+            // Returning null socket is handled by the caller function.
+            if(ipAddresses == null || ipAddresses.Length == 0)
+            {
+                return null;
+            }
+
+            Socket[] sockets = new Socket[ipAddresses.Length];
```
JRahnama, line 336: I quiet do not understand why we do not use `List` type instead of the array. In that case we won't need the extra incrementor later. Is there any problem using `List`?

cheenamalhotra: This is only as per current design, with changes coming from https://github.com/dotnet/SqlClient/pull/1029, this will change too.

JRahnama (approved):

PR #1033, opened 2021-04-15
Shrink SqlParameter XmlSchema

Wraith2:
In https://github.com/dotnet/corefx/pull/35549 I changed `SqlParameter` to use `SqlMetaDataXmlSchemaCollection` to hold xml schema information if it is present. This reduces the memory size because rather than keeping 3 string references which will mostly be null a single reference is kept for the sub object and it is created before use. This ports that change to netcore and netfx and adds the covering tests, it includes the fixes from dotnet/corefx#41008 which covered a problem in the original.

While working on `SqlParameter` i made a couple of other minor changes:

The direction field was not initialized but had a getter which checked for the 0 value and then returned ParameterDirection.Input. Recent profiling work on parameter passing showed that getting the parameter direction is an unexpectedly hot call. Because it is not possible for users to observe the 0 value it is safe to simply initialize the value in the ctors and then use a simple getter. The only observable change in this is if a user changes the value and is using INotifyPropertyChanged (usually designers use this) they will now not see a change from 0 to 1 if they assign to the Direction property, given that they could never observe the 0 value this seems like it is unlikely to break anyone.

The ctor call chain in netfx and netcore was slightly different so I synchronized them. This should have no practical effect but it will make merging the files later easier.

I removed some uses of `ADP.StrEmpty` and replaced them with `string.Empty`, these two expressions point to the same string instance because of runtime interning of constant strings. Using our own alias for it is no more secure than referring to it as `string.Empty` or `""`, as such just use the runtime supported constant and remove our copy from the library at some future date when all other references have been removed.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlParameter.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlParameter.cs
@@ -332,7 +332,7 @@ namespace Microsoft.Data.SqlClient
             Value = value;
             if (!string.IsNullOrEmpty(xmlSchemaCollectionDatabase) || !string.IsNullOrEmpty(xmlSchemaCollectionOwningSchema) || !string.IsNullOrEmpty(xmlSchemaCollectionName))
             {
-                EnsureXmlSchemaCollectionExists();
+                EnsureXmlSchemaCollection();
                 _xmlSchemaCollection.Database = xmlSchemaCollectionDatabase;
                 _xmlSchemaCollection.OwningSchema = xmlSchemaCollectionOwningSchema;
                 _xmlSchemaCollection.Name = xmlSchemaCollectionName;
@@ -408,11 +408,7 @@ namespace Microsoft.Data.SqlClient
         public string XmlSchemaCollectionDatabase
         {
             get => _xmlSchemaCollection?.Database ?? string.Empty;
-            set
-            {
-                EnsureXmlSchemaCollectionExists();
-                _xmlSchemaCollection.Database = value;
-            }
+            set => EnsureXmlSchemaCollection().Database = value;
         }
 
         /// <include file='../../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlParameter.xml' path='docs/members[@name="SqlParameter"]/XmlSchemaCollectionOwningSchema/*' />
@@ -420,11 +416,7 @@ namespace Microsoft.Data.SqlClient
         public string XmlSchemaCollectionOwningSchema
         {
             get => _xmlSchemaCollection?.OwningSchema ?? string.Empty;
-            set
-            {
-                EnsureXmlSchemaCollectionExists();
-                _xmlSchemaCollection.OwningSchema = value;
-            }
+            set => EnsureXmlSchemaCollection().OwningSchema = value;
         }
 
         /// <include file='../../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlParameter.xml' path='docs/members[@name="SqlParameter"]/XmlSchemaCollectionName/*' />
@@ -432,11 +424,7 @@ namespace Microsoft.Data.SqlClient
         public string XmlSchemaCollectionName
         {
             get => _xmlSchemaCollection?.Name ?? string.Empty;
-            set
-            {
-                EnsureXmlSchemaCollectionExists();
-                _xmlSchemaCollection.Name = value;
-            }
+            set => EnsureXmlSchemaCollection().Name = value;
         }
 
         /// <include file='../../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlParameter.xml' path='docs/members[@name="SqlParameter"]/ForceColumnEncryption/*' />
@@ -997,8 +985,7 @@ namespace Microsoft.Data.SqlClient
             destination._collation = _collation;
             if (_xmlSchemaCollection != null)
             {
-                destination.EnsureXmlSchemaCollectionExists();
-                destination._xmlSchemaCollection.CopyFrom(_xmlSchemaCollection);
+                destination.EnsureXmlSchemaCollection().CopyFrom(_xmlSchemaCollection);
             }
             destination._udtTypeName = _udtTypeName;
             destination._typeName = _typeName;
@@ -1035,12 +1022,13 @@ namespace Microsoft.Data.SqlClient
             return parent;
         }
 
-        private void EnsureXmlSchemaCollectionExists()
+        private SqlMetaDataXmlSchemaCollection EnsureXmlSchemaCollection()
         {
             if (_xmlSchemaCollection is null)
             {
                 _xmlSchemaCollection = new SqlMetaDataXmlSchemaCollection();
             }
+            return _xmlSchemaCollection;
         }
 
         internal void FixStreamDataForNonPLP()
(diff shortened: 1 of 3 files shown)


cheenamalhotra (approved):

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlParameter.cs
```diff
@@ -331,9 +330,13 @@ string xmlSchemaCollectionName
             SourceVersion = sourceVersion;
             SourceColumnNullMapping = sourceColumnNullMapping;
             Value = value;
-            XmlSchemaCollectionDatabase = xmlSchemaCollectionDatabase;
-            XmlSchemaCollectionOwningSchema = xmlSchemaCollectionOwningSchema;
-            XmlSchemaCollectionName = xmlSchemaCollectionName;
+            if (!string.IsNullOrEmpty(xmlSchemaCollectionDatabase) || !string.IsNullOrEmpty(xmlSchemaCollectionOwningSchema) || !string.IsNullOrEmpty(xmlSchemaCollectionName))
+            {
+                EnsureXmlSchemaCollectionExists();
```
DavoudEshtehari, line 335: If you don't have any specific reason for using property instead of a field and calling this function, add a property for `_xmlSchemaCollection`, please.

Wraith2: There is a reason. If you have a property you would have it auto create the `_xmlSchemaCollection` field if it was null and then pass it back. That is harmless. The problem is that people then get used to using property accessor for `XmlSchemaCollection` and you will end up with code like `if (!string.IsNullOrEmpty(XmlSchemaCollection.Database)` which looks harmless but results in the schema container being created just to check if it's empty which isn't good. 

So I chose this pattern even though it's a little more work so that access is always through the field and anyone modifying the code can see that they need to call the create function before they assume that it's non-null.

Wraith2: I could add a property called `EnsureXmlSchemaCollection` and then we end up with the property access pattern used like:
```csharp
public string XmlSchemaCollectionDatabase
        {
            get => _xmlSchemaCollection?.Database ?? string.Empty;
            set
            {
                EnsureXmlSchemaCollection.Database = value;
            }
        }
```

what do you think?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlParameter.cs
```diff
@@ -984,9 +995,11 @@ private void CloneHelper(SqlParameter destination)
 
             destination._metaType = _metaType;
             destination._collation = _collation;
-            destination._xmlSchemaCollectionDatabase = _xmlSchemaCollectionDatabase;
-            destination._xmlSchemaCollectionOwningSchema = _xmlSchemaCollectionOwningSchema;
-            destination._xmlSchemaCollectionName = _xmlSchemaCollectionName;
+            if (_xmlSchemaCollection != null)
```
JRahnama, line 998: If I am not mistaken, with the C# language updated to 9 in the driver recently I think we can use "is not null"

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlParameter.cs
```diff
@@ -439,7 +450,7 @@ public override DbType DbType
         /// <include file='../../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlParameter.xml' path='docs/members[@name="SqlParameter"]/ParameterName/*' />
         public override string ParameterName
         {
-            get => _parameterName ?? ADP.StrEmpty;
+            get => _parameterName ?? string.Empty;
```
JRahnama, line 453: Thanks for changing this.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlParameter.cs
Wraith2: Yes. It's a question of what style we want to use really. I like is not null but I've been trying to synchronise the style throughout the library with these merge PR's so if we want to do a !=null -> is not null change I think I'd want to do it in a separate PR over most of the files at once.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlParameter.cs
Wraith2: I'll get rid of them all everntually. 😁

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlParameter.cs
```diff
@@ -984,9 +995,11 @@ private void CloneHelper(SqlParameter destination)
...
+            {
+                destination.EnsureXmlSchemaCollectionExists();
+                destination._xmlSchemaCollection.CopyFrom(_xmlSchemaCollection);
+            }
```
DavoudEshtehari, line 1002: After defining `EnsureXmlSchemaCollection` you can replace this if-block with the following line:
```
destination.EnsureXmlSchemaCollection.CopyFrom(_xmlSchemaCollection);
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlParameter.cs
DavoudEshtehari: With this approach, I think except in ctor you can use the property `EnsureXmlSchemaCollection` instead of `_xmlSchemaCollection` to assign a value or call a function.

Wraith2: Yup. Done. It looks much neater in general. I've gone with a function over a property because of the api guidance that trivial operations should be properties but anything which has non-obvious side effects should really be a function. See what you think.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlParameter.cs
Wraith2: Done.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlParameter.cs
DavoudEshtehari: NIT: The `CopyFrom()` function internally skips a null parameter and you can get rid of this line.

Wraith2: If we skip the null check we'll create the instance for the current class even though the one we're copying from doesn't have one. We're trying to avoid creating them unless we have to.

DavoudEshtehari: It's reasonable.

Wraith2: For a path that's so little used I'm not too concerned about an extra null check. using xml schema is very rare.

DavoudEshtehari (approved):

JRahnama (approved):

Issue #974, opened 2021-03-09
Poor INSERT performance with ExecuteNonQuery() and many parameters


KristianWedberg:
### Description

When inserting rows with `ExecuteNonQuery()` and placeholder parameters, performance 
(measured as **inserted parameters per second**) with the `Microsoft.Data.SqlClient` v2.1.1  
and `System.Data.SqlClient` providers is progressively impacted from about 50+ parameters 
per statement, both for single rows and for multirow batches.

Inserting with 2048 parameters in the statement is up to 7 times slower than with 
128 parameters.

The issue is _not_ present when using the `System.Data.Odbc` provider (or non-SQL Server
providers & databases), and performance is up to 25 times slower (when using 2048 parameters 
per statement) than with the ODBC provider.

I've tested with Windows 10, .NET5.0 and a local SQL Server 2019 database, varying the 
number of columns from 1 to 1024, and number of rows in each batch from 1 to 1000.

Note: The 'waviness' at 1000 and 2000 parameters is just an effect of the 1000/1024 and 2000/2048 
numbers being separated in the graph - see the write-up for details:

![Average Throughput](https://github.com/KristianWedberg/sql-batch-insert-performance/raw/main/average-throughput.png)

The target table is un-indexed and the inserts uses SQL-92 syntax:

```
INSERT INTO tablename (column-a, [column-b, ...])
VALUES (@r0_c0, [@r0_c1, ...]),
       (@r1_c0, [@r1_c1, ...]),
       ...
```

While table valued parameters and bulk inserts don't have this issue, they won't 
help when inserting a single or just a few very wide rows at a time, so it would be 
really useful to have this addressed (or identify any gremlins on my part!)

Please see the full write-up with charts, tables, BenchmarkDotNet info etc. at:

https://github.com/KristianWedberg/sql-batch-insert-performance

### Reproduce

Fully runnable source (using BenchmarkDotNet) and instructions at:

https://github.com/KristianWedberg/sql-batch-insert-performance

### Expected behavior

Using more parameters (beyond 128) in an insert row or batch should increase throughput, 
measured as the number of parameters inserted per second, just like it does with 
`System.Data.Odbc` and other non-SqlClient providers.

### Technical details

- Microsoft.Data.SqlClient Version="2.1.1"
- System.Data.Odbc Version="5.0.0"
- System.Data.SqlClient Version="4.8.2"
- OS=Windows 10.0.19042
- .NET Core SDK=5.0.103

[...]

PR #1041, opened 2021-04-18
Add SqlCommand.DisableOutputParameters Feature

Wraith2:
fixes https://github.com/dotnet/SqlClient/issues/974

This PR adds a new property to SqlCommand called DisableOutputParameters. When set to true parameters names will not be sent to the sql server when the command is executed. This behaviour has been show to increase performance for commands with very large numbers of parameters which are used in bulk operations.

This is a draft because:
1) This adds public surface area. The external name is open to debate.
2) This needs documenting.
3) This needs tests adding.

@KristianWedberg would you mind running your perf tests with the artifacts from this PR to ensure that this small change still has the performance changes observed in the investigation we did in your original issue?

### DIFF
--- a/doc/snippets/Microsoft.Data.SqlClient/SqlCommand.xml
+++ b/doc/snippets/Microsoft.Data.SqlClient/SqlCommand.xml
@@ -286,8 +286,13 @@ The following console application creates updates data within the **AdventureWor
 			<exception cref="T:System.InvalidOperationException">
 				The
 				<see cref="T:Microsoft.Data.SqlClient.SqlConnection" />
-				closed or dropped during a streaming operation. For more information about streaming, see [SqlClient Streaming Support](/sql/connect/ado-net/sqlclient-streaming-support).
-			</exception>
+        closed or dropped during a streaming operation. For more information about streaming, see [SqlClient Streaming Support](/sql/connect/ado-net/sqlclient-streaming-support).
+
+        - or -
+
+        <see cref="P:Microssoft.Data.SqlClient.SqlCommand.EnableOptimizedParameterBinding" />
+        is set to true and a parameter with direction Output or InputOutput has been added to the <see cref="P:Microsoft.Data.SqlClient.SqlCommand.Parameters" />Parameters collection.
+      </exception>
 			<exception cref="T:System.IO.IOException">
 				An error occurred in a
 				<see cref="T:System.IO.Stream" />
@@ -399,8 +404,13 @@ To set up this example, create a new Windows application. Put a <xref:System.Win
 			<exception cref="T:System.InvalidOperationException">
 				The
 				<see cref="T:Microsoft.Data.SqlClient.SqlConnection" />
-				closed or dropped during a streaming operation. For more information about streaming, see [SqlClient Streaming Support](/sql/connect/ado-net/sqlclient-streaming-support).
-			</exception>
+        closed or dropped during a streaming operation. For more information about streaming, see [SqlClient Streaming Support](/sql/connect/ado-net/sqlclient-streaming-support).
+
+        - or -
+
+        <see cref="P:Microssoft.Data.SqlClient.SqlCommand.EnableOptimizedParameterBinding" />
+        is set to true and a parameter with direction Output or InputOutput has been added to the <see cref="P:Microsoft.Data.SqlClient.SqlCommand.Parameters" />Parameters collection.
+      </exception>
 		</BeginExecuteNonQuery>
 		<BeginExecuteReader name="default">
 			<summary>
@@ -477,8 +487,13 @@ The following console application starts the process of retrieving a data reader
 			<exception cref="T:System.InvalidOperationException">
 				The
 				<see cref="T:Microsoft.Data.SqlClient.SqlConnection" />
-				closed or dropped during a streaming operation. For more information about streaming, see [SqlClient Streaming Support](/sql/connect/ado-net/sqlclient-streaming-support).
-			</exception>
+        closed or dropped during a streaming operation. For more information about streaming, see [SqlClient Streaming Support](/sql/connect/ado-net/sqlclient-streaming-support).
+
+        - or -
+
+        <see cref="P:Microssoft.Data.SqlClient.SqlCommand.EnableOptimizedParameterBinding" />
+        is set to true and a parameter with direction Output or InputOutput has been added to the <see cref="P:Microsoft.Data.SqlClient.SqlCommand.Parameters" />Parameters collection.
+      </exception>
 			<exception cref="T:System.IO.IOException">
 				An error occurred in a
 				<see cref="T:System.IO.Stream" />
@@ -584,8 +599,13 @@ This example also passes the `CommandBehavior.CloseConnection` and `CommandBehav
 			<exception cref="T:System.InvalidOperationException">
 				The
 				<see cref="T:Microsoft.Data.SqlClient.SqlConnection" />
-				closed or dropped during a streaming operation. For more information about streaming, see [SqlClient Streaming Support](/sql/connect/ado-net/sqlclient-streaming-support).
-			</exception>
+        closed or dropped during a streaming operation. For more information about streaming, see [SqlClient Streaming Support](/sql/connect/ado-net/sqlclient-streaming-support).
+
+        - or -
+
+        <see cref="P:Microssoft.Data.SqlClient.SqlCommand.EnableOptimizedParameterBinding" />
+        is set to true and a parameter with direction Output or InputOutput has been added to the <see cref="P:Microsoft.Data.SqlClient.SqlCommand.Parameters" />Parameters collection.
+      </exception>
 			<exception cref="T:System.IO.IOException">
 				An error occurred in a
 				<see cref="T:System.IO.Stream" />
@@ -702,8 +722,13 @@ To set up this example, create a new Windows application. Put a <xref:System.Win
 			<exception cref="T:System.InvalidOperationException">
 				The
 				<see cref="T:Microsoft.Data.SqlClient.SqlConnection" />
-				closed or dropped during a streaming operation. For more information about streaming, see [SqlClient Streaming Support](/sql/connect/ado-net/sqlclient-streaming-support).
-			</exception>
+        closed or dropped during a streaming operation. For more information about streaming, see [SqlClient Streaming Support](/sql/connect/ado-net/sqlclient-streaming-support).
+
+        - or -
+
+        <see cref="P:Microssoft.Data.SqlClient.SqlCommand.EnableOptimizedParameterBinding" />
+        is set to true and a parameter with direction Output or InputOutput has been added to the <see cref="P:Microsoft.Data.SqlClient.SqlCommand.Parameters" />Parameters collection.
+      </exception>
 			<exception cref="T:System.IO.IOException">
 				An error occurred in a
 				<see cref="T:System.IO.Stream" />
@@ -831,8 +856,13 @@ This example passes the `CommandBehavior.CloseConnection` value in the `behavior
 			<exception cref="T:System.InvalidOperationException">
 				The
 				<see cref="T:Microsoft.Data.SqlClient.SqlConnection" />
-				closed or dropped during a streaming operation. For more information about streaming, see [SqlClient Streaming Support](/sql/connect/ado-net/sqlclient-streaming-support).
-			</exception>
+        closed or dropped during a streaming operation. For more information about streaming, see [SqlClient Streaming Support](/sql/connect/ado-net/sqlclient-streaming-support).
+
+        - or -
+
+        <see cref="P:Microssoft.Data.SqlClient.SqlCommand.EnableOptimizedParameterBinding" />
+        is set to true and a parameter with direction Output or InputOutput has been added to the <see cref="P:Microsoft.Data.SqlClient.SqlCommand.Parameters" />Parameters collection.
+      </exception>
 			<exception cref="T:System.IO.IOException">
 				An error occurred in a
 				<see cref="T:System.IO.Stream" />
@@ -940,6 +970,11 @@ The following console application starts the process of retrieving XML data asyn
 				The
 				<see cref="T:Microsoft.Data.SqlClient.SqlConnection" />
 				closed or dropped during a streaming operation. For more information about streaming, see [SqlClient Streaming Support](/sql/connect/ado-net/sqlclient-streaming-support).
+        
+        - or -
+
+        <see cref="P:Microssoft.Data.SqlClient.SqlCommand.EnableOptimizedParameterBinding" />
+        is set to true and a parameter with direction Output or InputOutput has been added to the <see cref="P:Microsoft.Data.SqlClient.SqlCommand.Parameters" />Parameters collection.
 			</exception>
 			<exception cref="T:System.IO.IOException">
 				An error occurred in a
@@ -1067,8 +1102,13 @@ To set up this example, create a new Windows application. Put a <xref:System.Win
 			<exception cref="T:System.InvalidOperationException">
 				The
 				<see cref="T:Microsoft.Data.SqlClient.SqlConnection" />
-				closed or dropped during a streaming operation. For more information about streaming, see [SqlClient Streaming Support](/sql/connect/ado-net/sqlclient-streaming-support).
-			</exception>
+        closed or dropped during a streaming operation. For more information about streaming, see [SqlClient Streaming Support](/sql/connect/ado-net/sqlclient-streaming-support).
+
+        - or -
+
+        <see cref="P:Microssoft.Data.SqlClient.SqlCommand.EnableOptimizedParameterBinding" />
+        is set to true and a parameter with direction Output or InputOutput has been added to the <see cref="P:Microsoft.Data.SqlClient.SqlCommand.Parameters" />Parameters collection.
+      </exception>
 			<exception cref="T:System.IO.IOException">
 				An error occurred in a
 				<see cref="T:System.IO.Stream" />
--- a/doc/snippets/Microsoft.Data.SqlClient/SqlCommand.xml
+++ b/doc/snippets/Microsoft.Data.SqlClient/SqlCommand.xml
@@ -291,7 +291,7 @@ The following console application creates updates data within the **AdventureWor
         - or -
 
         <see cref="P:Microssoft.Data.SqlClient.SqlCommand.EnableOptimizedParameterBinding" />
-        is set to true and a parameter with direction Output or InputOutput has been added to the <see cref="P:Microsoft.Data.SqlClient.SqlCommand.Parameters" />Parameters collection.
+        is set to true and a parameter with direction Output or InputOutput has been added to the <see cref="P:Microsoft.Data.SqlClient.SqlCommand.Parameters" /> collection.
       </exception>
 			<exception cref="T:System.IO.IOException">
 				An error occurred in a
@@ -409,7 +409,7 @@ To set up this example, create a new Windows application. Put a <xref:System.Win
         - or -
 
         <see cref="P:Microssoft.Data.SqlClient.SqlCommand.EnableOptimizedParameterBinding" />
-        is set to true and a parameter with direction Output or InputOutput has been added to the <see cref="P:Microsoft.Data.SqlClient.SqlCommand.Parameters" />Parameters collection.
+        is set to true and a parameter with direction Output or InputOutput has been added to the <see cref="P:Microsoft.Data.SqlClient.SqlCommand.Parameters" /> collection.
       </exception>
 		</BeginExecuteNonQuery>
 		<BeginExecuteReader name="default">
@@ -492,7 +492,7 @@ The following console application starts the process of retrieving a data reader
         - or -
 
         <see cref="P:Microssoft.Data.SqlClient.SqlCommand.EnableOptimizedParameterBinding" />
-        is set to true and a parameter with direction Output or InputOutput has been added to the <see cref="P:Microsoft.Data.SqlClient.SqlCommand.Parameters" />Parameters collection.
+        is set to true and a parameter with direction Output or InputOutput has been added to the <see cref="P:Microsoft.Data.SqlClient.SqlCommand.Parameters" /> collection.
       </exception>
 			<exception cref="T:System.IO.IOException">
 				An error occurred in a
@@ -604,7 +604,7 @@ This example also passes the `CommandBehavior.CloseConnection` and `CommandBehav
         - or -
 
         <see cref="P:Microssoft.Data.SqlClient.SqlCommand.EnableOptimizedParameterBinding" />
-        is set to true and a parameter with direction Output or InputOutput has been added to the <see cref="P:Microsoft.Data.SqlClient.SqlCommand.Parameters" />Parameters collection.
+        is set to true and a parameter with direction Output or InputOutput has been added to the <see cref="P:Microsoft.Data.SqlClient.SqlCommand.Parameters" /> collection.
       </exception>
 			<exception cref="T:System.IO.IOException">
 				An error occurred in a
@@ -727,7 +727,7 @@ To set up this example, create a new Windows application. Put a <xref:System.Win
         - or -
 
         <see cref="P:Microssoft.Data.SqlClient.SqlCommand.EnableOptimizedParameterBinding" />
-        is set to true and a parameter with direction Output or InputOutput has been added to the <see cref="P:Microsoft.Data.SqlClient.SqlCommand.Parameters" />Parameters collection.
+        is set to true and a parameter with direction Output or InputOutput has been added to the <see cref="P:Microsoft.Data.SqlClient.SqlCommand.Parameters" /> collection.
       </exception>
 			<exception cref="T:System.IO.IOException">
 				An error occurred in a
@@ -861,7 +861,7 @@ This example passes the `CommandBehavior.CloseConnection` value in the `behavior
         - or -
 
         <see cref="P:Microssoft.Data.SqlClient.SqlCommand.EnableOptimizedParameterBinding" />
-        is set to true and a parameter with direction Output or InputOutput has been added to the <see cref="P:Microsoft.Data.SqlClient.SqlCommand.Parameters" />Parameters collection.
+        is set to true and a parameter with direction Output or InputOutput has been added to the <see cref="P:Microsoft.Data.SqlClient.SqlCommand.Parameters" /> collection.
       </exception>
 			<exception cref="T:System.IO.IOException">
 				An error occurred in a
@@ -974,7 +974,7 @@ The following console application starts the process of retrieving XML data asyn
         - or -
 
         <see cref="P:Microssoft.Data.SqlClient.SqlCommand.EnableOptimizedParameterBinding" />
-        is set to true and a parameter with direction Output or InputOutput has been added to the <see cref="P:Microsoft.Data.SqlClient.SqlCommand.Parameters" />Parameters collection.
+        is set to true and a parameter with direction Output or InputOutput has been added to the <see cref="P:Microsoft.Data.SqlClient.SqlCommand.Parameters" /> collection.
 			</exception>
 			<exception cref="T:System.IO.IOException">
 				An error occurred in a
@@ -1107,7 +1107,7 @@ To set up this example, create a new Windows application. Put a <xref:System.Win
         - or -
 
         <see cref="P:Microssoft.Data.SqlClient.SqlCommand.EnableOptimizedParameterBinding" />
-        is set to true and a parameter with direction Output or InputOutput has been added to the <see cref="P:Microsoft.Data.SqlClient.SqlCommand.Parameters" />Parameters collection.
+        is set to true and a parameter with direction Output or InputOutput has been added to the <see cref="P:Microsoft.Data.SqlClient.SqlCommand.Parameters" /> collection.
       </exception>
 			<exception cref="T:System.IO.IOException">
 				An error occurred in a
--- a/doc/snippets/Microsoft.Data.SqlClient/SqlCommand.xml
+++ b/doc/snippets/Microsoft.Data.SqlClient/SqlCommand.xml
@@ -1343,6 +1343,33 @@ The <xref:Microsoft.Data.SqlClient.SqlCommand.CreateParameter%2A> method is a st
 				To be added.
 			</remarks>
 		</Dispose>
+    <EnableOptimizedParameterBinding>
+      <summary>
+        Gets or sets a value indicating whether the command object should optimize parameter performance by disabling Output and InputOutput directions when submitting the command to the SQL Server.
+      </summary>
+      <value>
+        A value indicating whether the command object should optimize parameter performance by disabling Output and InputOuput parameter directions when submitting the command to the SQL Server. 
+        The default is <see langword="false" />.
+      </value>
+      <remarks>
+        <format type="text/markdown">
+          <![CDATA[
+## Remarks
+You must set the value for this property before the command is executed for it to take effect.
+
+When a command is submitted to the server with parameters a list of the parameter names is sent as part of the submission. The list is used on the server to match Output and InputOutput parameters to the results of the query execution so that the values can be returned to the caller. This option disables the construction and submission of the parameter name list and as a consequence disables the use of Output and InputOutput parameters. The return parameter is not affected by this option.
+
+A command sent with this option changes the way parameters are handled on the server, because there is no need to maintain an output parameter map. The result of this change is that queries with large numbers of input parameters may execute much faster. 
+
+The fewest number of parameters where this will take effect depends on the individual situation and should be detected by measuring query duration with and without the option enabled. Any query with more than 24 parameters may show lower overall query duration. Queries with parameter counts lower than 24 are unlikely to show a difference.
+
+> [!NOTE]
+If the option is enabled and a parameter with Direction Output or InputOutput is present in the Parameters collection an InvalidOperationException will be thrown when the command is executed.
+
+]]>
+</format>
+      </remarks>
+    </EnableOptimizedParameterBinding>
 		<EndExecuteNonQuery name="IAsyncResult">
 			<param name="asyncResult">
 				The
--- a/src/Microsoft.Data.SqlClient/netfx/ref/Microsoft.Data.SqlClient.cs
+++ b/src/Microsoft.Data.SqlClient/netfx/ref/Microsoft.Data.SqlClient.cs
@@ -556,6 +556,8 @@ namespace Microsoft.Data.SqlClient
         [System.ComponentModel.DesignOnlyAttribute(true)]
         [System.ComponentModel.EditorBrowsableAttribute(System.ComponentModel.EditorBrowsableState.Never)]
         public override bool DesignTimeVisible { get { throw null; } set { } }
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlCommand.xml' path='docs/members[@name="SqlCommand"]/UsePositionalParameters/*'/>
+        public bool DisableOutputParameters { get { throw null; } set { } }
         /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlCommand.xml' path='docs/members[@name="SqlCommand"]/Parameters/*'/>
         public new Microsoft.Data.SqlClient.SqlParameterCollection Parameters { get { throw null; } }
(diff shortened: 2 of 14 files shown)


KristianWedberg:
Hiya @Wraith2, can you make a package available like previously?

ErikEJ:
@KristianWedberg You can download for the azure DevOps CI-BuildNPack pipeline here: https://dev.azure.com/sqlclientdrivers-ci/sqlclient/_build/results?buildId=29463&view=artifacts&pathAsName=false&type=publishedArtifacts

KristianWedberg:
Yup, this new PR (Microsoft.Data.SqlClient.2.1.0-pull-049c895.21108.1.nupkg) also has great performance, scaling to 2048 parameters with about 10% better performance vs. ODBC:

| UsePositionalParameters | provider | RPS | CPR | Parameters | Throughput [params/s] |
|--- | ---: | ---: | ---: | ---: | ---: |
| true | Microsoft | 4 | 8 | 32 | 143,034 |
| true | Microsoft | 8 | 8 | 64 | 257,235 |
| true | Microsoft | 16 | 8 | 128 | 409,405 |
| true | Microsoft | 32 | 8 | 256 | 630,407 |
| true | Microsoft | 32 | 16 | 512 | 846,929 |
| true | Microsoft | 32 | 32 | 1024 | 1,120,485 |
| true | Microsoft | 32 | 64 | 2048 | 1,343,983 |
|       |    Odbc | 4 | 8 | 32 | 133,329 |
|       |    Odbc | 8 | 8 | 64 | 239,711 |
|       |    Odbc | 16 | 8 | 128 | 378,298 |
|       |    Odbc | 32 | 8 | 256 | 549,672 |
|       |    Odbc | 32 | 16 | 512 | 755,040 |
|       |    Odbc | 32 | 32 | 1024 | 984,802 |
|       |    Odbc | 32 | 64 | 2048 | 1,120,003 |

So all good so far!

Wraith2:
Excellent. So, fancy helping write some tests? 😁

Wraith2:
I've added some simple tests which examine the behaviour of parameters with this setting enabled and they surprised me a bit.
1) Input and InputOutput parameters don't work because the names can't be bound, seems sane enough. 
2) All parameters named in the query still have to be specified correctly so you can't just omit parameter names and you can't make them up, they've still got to match in the same way as using named parameters.
3) Parameters can be re-used by name, so you can use an input parameter multiple times

So it seems that on input there is still a name-value map generated but it's populated by position not name match and then the engine proceeds normally requiring names to exist once the query is parsed etc. On output there is no name-value map so you can't use named outputs at all.

roji:
Out of curiosity... Is a new flag on SqlCommand really needed? Is there any reason you can't simply check if the command's parameters have their names set or not (i.e. ParameterName is null)? I'm raising this because I'm planning a similar change in Npgsql.

Wraith2:
Ideally I'd use a new flag in the CommandBehavior enum. That's a breaking change though and it's in /runtime and since I thought this was going to be SqlClient specific it didn't seem to be the right solution to the problem. If you could use the same value in npqsql then perhaps it's worth considering. I don't know if mysql or sqlite (or firebird?) would be able to or want to make use of it. It might be worth a wider discussion.

It does have to be at DbCommand level. Even though the parameter values aren't sent with names the RPC request still contains a list of parameter names and types and those get checked. It's peculiar but I'm finding from the tests that the only thing this changes is that the parameters are matched up to names in the order provided, you can still use a named parameter multiple times and you still have to name each parameter correctly and provide values for each named parameter.

roji:
Thanks @Wraith2, so you're saying users have to specify SqlParameter.ParameterName regardless of this option, right?

That's peculiar indeed... The concept of positional parameter usually means that placeholders themselves are positional: either via numeric placeholders like in PostgreSQL ($1, $2), or with implicit positioning (?). I doubt a CommandBehavior would make sense across multiple providers here.

Wraith2:
If there is a way to suppose true placeholder positional style I don't know what it is. `?` isn't valid syntax and using `@` fails with `Must declare the scalar variable "@"` from the server, so we sent it but it isn't acceptable. I'm working from the way RPC are executed in this driver. It may be that if I didn't send the parameter list string it would change how things work. I'm not aware of a T-SQL spec that tells me whether an unnamed parameter would be acceptable. I do know I've never seen anything other than stored procedure invocations use positional parameters with SQL Server.

If there is a use for the positional option in other drivers the enum flag is a better approach and the limitations of how Sql Server implements it would be something to cover in the documentation or provider.

Wraith2:
We could do this without the formal extension to CommandBehavior by simply oring in a new flag value
```
const int PositionalParameters = 64;
CommandBehavior behavior = (CommandBehavior )((int)CommandBehavior.Default | PositionalParameters);
command.ExecuteNonQuery(behavior);
```

but that looks like a horrible hackjob and if I found that as an official answer to I'd be suspicious about why the people who own the api felt the need to provide such a weird way to access it when they're capable of changing the api surface. 

It would need some internal plumbing to make sure validation of command behavior didn't trip up on it but i can handle that.

roji:
... I agree this looks... not ideal :) There's also a risk of conflict in case CommandBehavior ever does get extended.

Is there any real downside to introducing a flag on SqlCommand as this PR currently does?

Wraith2:
The only danger I can think of it is that people will see and use it without understanding it.

KristianWedberg:
If I understand the new functionality and limitations correctly, from the users' perspective this is _not_ positional parameters. Rather it's an Input only (or InputAndReturn only?) optimization for >64 parameters.

I think a different property name would reduce any incorrect usage, e.g. `UseInputOnlyParameters` (or `UseInputReturnOnlyParameters`), and doc the performance advantage.

Thoughts?

Wraith2:
> If I understand the new functionality and limitations correctly, from the users' perspective this is _not_ positional parameters. Rather it's an Input only (or InputAndReturn only?) optimization for >64 parameters.

Not quite, consider this (passing) test:

```csharp
[Fact]
private static void PositionalParameters_ParametersAreUsedByPosition()
{
    using (var connection = new SqlConnection(DataTestUtility.TCPConnectionString))
    {
        connection.Open();

        using (var command = new SqlCommand("SELECT @Second, @First", connection))
        {
            command.UsePositionalParameters = true;
            command.Parameters.AddWithValue("@First", 1);
            command.Parameters.AddWithValue("@Second", 2);

            using (SqlDataReader reader = command.ExecuteReader())
            {
                reader.Read();

                int firstOutput = reader.GetInt32(0);
                int secondOutput = reader.GetInt32(1);

                Assert.Equal(1, secondOutput);
                Assert.Equal(2, firstOutput);
            }
        }
    }
}
```

On input: Names must all be declared. Names can be reused without taking an additional value. Names and values are associated at first encounter of the name by taking the next value from the list of values.
On output: Names don't exist.

KristianWedberg:
OK. How about `UsePositionalOutputParameters` then?

And are Return parameters affected or not (i.e. should they be referenced in the name or not)?

Wraith2:
> OK. How about `UsePositionalOutputParameters` then?

Capturing all the complexity in the name is virtually impossible unless we use an absurd name. We're going to have to rely on the docs to inform people what it changes and what it doesn't.

> And are Return parameters affected or not (i.e. should they be referenced in the name or not)?

Return parameter names aren't really important because nothing in the driver cares about the name. They work as the did before because their names aren't relevant. There is only one return value it has no name and it has type int32, in the absence of explicitly returned value it has default value 0.

KristianWedberg:
Given what you  said about Return values, I'd certainly vote for `UsePositionalOutputParameters` over `UsePositionalParameters` to reduce incorrect usage...

Wraith2:
It also affects input parameters as I showed in that test so using the word Output with the attendant implication that it only affects output parameters seems misleading. Perhaps `UseAnonymousParameters` or `SendAnonymousParameters` would be closer but as I said you're still going to need people to read the docs.

roji:
@Wraith2 I'm probably missing something, but in [this example](https://github.com/dotnet/SqlClient/pull/1041#issuecomment-822618472), would the code work any different if UsePositionalParameters isn't set? It just looks like regular named parameter code to me...

Wraith2:
> @Wraith2 I'm probably missing something, but in [this example](https://github.com/dotnet/SqlClient/pull/1041#issuecomment-822618472), would the code work any different if UsePositionalParameters isn't set? It just looks like regular named parameter code to me...

Yes, The returned values would be reversed and would match the names.

Note the order of the @First and @Second parameters in the query is reversed from the order they're supplied. @Second occurs first in the query and is bound to the first value in the parameter list, which was added with the name @First. So the overall result is that @First returns the @Second parameter value and @Second returns the @First parameter value.

Wraith2:
Ok. It's fun. The server side seems to be using the rpc parameter order to match values to names so as long as the build of the parameter list parameter isn't screwed up in some way then the order will be preserved.
What happens when you do a query is that it executes an rcp request which can have a variable number of system parameters (depends on batch, sproc, text) and then those parameters are followed by the user parameters. So in the case of that test what you get is a param array like this:

```
rpc.params[0] = ("", "SELECT @Second, @First, @Second", nvarchar(255)) // system param query
rpc.params[1] = ("", "@First int, @Second", nvarchar(255)) // system param params list
rpc.params[2] = ("", 1, int) // user param
rpc.params[3] = ("", 2, int) // user param
```

So the server is matching up the index of the parameter name in the parameterList ( index1 in the rpc) text with the following user params and because they're in order it works. If you found a way to make the order difference then it'd fail but since we build and send in a single operation it'll always work if rpc are built this way.

In which case calling it `DisableOutputParameters` is probably accurate since that's the only effect you'll see. I think more tests might be a good idea to cover the edges of this to make sure we understand it completely.

cheenamalhotra:
`DisableOutputParameters` sounds like you're simply disabling output parameters from the command, whereas that's not the case. Since it's strictly related to "named parameters", I have a few suggestions shared below. I understand the complex behavior this property introduces, we can continue the discussion to pick the best option:
- "SendOutputParameterNames" (default true)
- "DisableOutputParameterNames" (default false)
- "SendRPCParameterNames" (default true)
- "DisableRPCParameterNames" (default false)

Wraith2:
We don't really expose users to the word RPC so I'm not a fan of using that in the name. 
I'd prefer a negative default value because that way any code like designer code will be unchanged by the appearance of a new member. Which leaves `DisableOutputParameterNames`, seems reasonable. 

What would you expect the behaviour to be if the name were `DisableOutputParameters` without the Name suffix? Perhaps it's worth adding in client side code to make it behave that way.

mburbea:
From my understanding this flag is essentially changing what the RPC is sending to the server from:

```sql
exec sp_executeSql N'select @','@ int',@=2
```
to the following:
```sql
exec sp_executeSql N'select @','@ int',2
```
So perhaps `DisableParameterNameBinding`? It's a bit of a mouthful but it's clear as to what it's doing.

Wraith2:
I'll vote for clarity over terseness, seems like a reasonable name to me.

roji:
Don't want to bikeshed too much, but an API name should generally express why a user would use it. If I understand correctly, the point of this flag would be to get better performance, whereas the current naming makes it sound like users would use it in order to DisableOutputParameters, which doesn't really make sense; disabling output parameters is a side-effect of the more efficient transfer technique, no?

So my proposal would be to call this something more like UseOptimizedParameterTransfer or similar. If this is incompatible with output parameters, I'd detect that and throw NotSupportedException with a clear message to the user.

KristianWedberg:
Even documented I think `UseOptimizedParameterTransfer` sounds so good to the user it will get significantly more erroneous uses compared to `DisableParameterNameBinding`.

Wraith2:
> Even documented I think `UseOptimizedParameterTransfer` sounds so good to the user it will get significantly more erroneous uses compared to `DisableParameterNameBinding`.

I agree. If I see that name without context I'll set it to true because of course I want optimized things because they're just better than non-optimized things. So really we want to call it `UseOptimizedParameterTransferByDisablingOutputParameters` but I don't think we will.

roji:
Well, it does sound like we want users to use this in all cases except when they have output parameters :rofl: Hopefully proper documentation and a clear exception would be enough to dispel any confusion. Am just not used to APIs being named by features which they happened to *not* support!

But feel free to ignore me on this if there's a consensus on the other name.

Wraith2:
Ok. So if I add in some code to detect Input and InoutOutput parameters and throw an exception at command execution time (because until that time you can change the parameters) are we at a consensus that we name it something to do with output parameters not working or binding? I'm not opposed to the word binding but it isn't terminology that's already used to describe how the parameters match up, so far there doesn't seem to be any real terminology for it.

ErikEJ:
UseOptimizedParameterBinding ?

KristianWedberg:
> Well, it does sound like we want users to use this in all cases except when they have output parameters 🤣

If this is 100% true, can it instead be automatically enabled for all non-output cases, and have the flag (`DisableOptimizedParameterBinding`?) to allow disabling it? Riskier of course, but maybe OK if it's certain there is no other pesky corner cases?

Wraith2:
> If this is 100% true,

If we can definitively prove that sure, I'm not confident so I'd rather leave users in a known good state and let people with the time to research and change their code with the option to turn it on. In the 90% use case of <32 parameters it makes no performance difference.

ErikEJ:
EnableOptimizedParameterBinding - making the new behaviour default would most likely be too much breaking change.

David-Engel:
It's a tough call. I'll vote for ErikEJ's final suggestion: EnableOptimizedParameterBinding

Be sure put in some good API docs to make it obvious that output parameters are not supported, when enabled.

Wraith2:
I thought of a way to do this without additional surface area. When preparing the RPC we have to count and characterize the parameters so I can count and determine if there are out parameters, if there are >32 and no outs I can enable this mode transparently to the user. If we do that I want to be able to turn it on and off with an appcontext switch to allow/disallow the option. If that's a better choice because it requires no naming then should the it be enabled or disabled by default? I'd say disabled and force knowledgeable users to opt in to the new way.

Or, do you prefer the additional surface area? 
This is a draft because I wanted to prove the performance and talk through the best way to expose the functionality.

KristianWedberg:
> if there are >32 and no outs 

">32" is when the issue impacts run times. It might consume excessive CPU on the server even with fewer parameters (and hitting a one thread bottleneck at >32), so being able to enable the new functionality explicitly would be good (and maybe lower the threshold?). 

> Or, do you prefer the additional surface area?

I'd say either just the new API, or API + AppContext. With only AppContext available, it would be risky to enable on a large application where I can't predict all the type of uses it will get.

Wraith2:
Should I add checks for the feature being enabled on the command and throw if any output or inputoutput parameters are set on the command?

ErikEJ:
I vote for new API/ option, and throwing with a clear error message.

Wraith2:
Renamed to `EnableOptimizedParameterBinding` and direction checks added at the same location other parameter validation is done. Tests changed to cover new behaviour. The exception message is:

> Parameter '{0}' cannot have Direction Output or InputOutput when EnableOptimizedParameterBinding is enabled on the parent command.

I'll think about how to document it. Any help there would be appreciated.

Wraith2:
Proposed docs:

## Remarks
You must set the value for this property before the command is executed for it to take effect.

When a command is submitted to the server with parameters a list of the parameter names is sent as part of the submission. The list is used on the server to match Output and InputOutput parameters to the results of the query execution so that the values can be returned to the caller. This option disables the construction and submission of the parameter name list and as a consequence disables the use of Output and InputOutput parameters. Return parameters are not affected by this option.

When a command with large numbers of parameters is sent with this option changes the way parameters are handled on the server because there is no need to maintain an output parameter map. The result of this change is that queries with large numbers of input parameters may execute much faster. 
The fewest number of parameters where this will take effect depends on the individual situation and should be detected by measuring query duration with and without the option enabled. Any query with more than 24 parameters may show lower overall query duration. Queries with parameter counts lower than 24 are unlikely to show a difference.

> [!NOTE]
If the option is enabled and a parameter with Direction Output or InputOutput is present in the Parameters collection an InvalidOperationException will be thrown when the command is executed.

ErikEJ:
>> Return parameters are not affected by this option.

The return parameter is not affected by this option.

>> When a command with large numbers of parameters is sent with this option changes the way parameters are handled on the server because there is no need to maintain an output parameter map. The result of this change is that queries with large numbers of input parameters may execute much faster.

A command sent with this option changes the way parameters are handled on the server,  because there is no need to maintain an output parameter map. The result of this change is that queries with large numbers of input parameters may execute much faster.

>> The fewest number of parameters where this will take effect depends on the individual situation and should be detected by measuring query duration with and without the option enabled. Any query with more than 24 parameters may show lower overall query duration. Queries with parameter counts lower than 24 are unlikely to show a difference.

The fewest number of parameters where this will take effect depends on the individual situation and should be detected by measuring query duration with and without the option enabled.

Wraith2:
Anyone looking to use this option even if they read the docs should have somewhere to start with the decision. If it were me I'd go googling for a hint about how many parameters it would need before I should consider it. So I do think that giving a number, even a conservative one, is better than not doing so.

KristianWedberg:
As an average user it's hard to follow and understand the implications from the change in the inner workings. I suggest the below start of the doc (and maybe only this doc unless the more detailed description is actually needed).

======

Queries with a large number of parameters can hit a performance bottleneck on the server, making queries progressively slower the more parameters they use. The slow-down has been noticed with 32 or more parameters, and enabling this property will completely avoid this slow-down. You should however perform your own testing to verify the suitability of using this property. 

Typical use cases include inserting many values into very wide rows, and using multi-row batch insert queries. Any return parameter is not affected by this option.

NOTE: This property cannot be used with queries that contain any **Output** or **InputOutput** parameters (and an `InvalidOperationException` would be thrown when the command was executed).

Bosch-Eli-Black:
@KristianWedberg As an average user who is reading through these PRs for fun, I find that description to be clear 🙂

doc/snippets/Microsoft.Data.SqlClient/SqlCommand.xml
```diff
@@ -1346,6 +1346,33 @@ The <xref:Microsoft.Data.SqlClient.SqlCommand.CreateParameter%2A> method is a st
...
+You must set the value for this property before the command is executed for it to take effect.
+
+When a command is submitted to the server with parameters a list of the parameter names is sent as part of the submission. The list is used on the server to match Output and InputOutput parameters to the results of the query execution so that the values can be returned to the caller. This option disables the construction and submission of the parameter name list and as a consequence disables the use of Output and InputOutput parameters. The return parameter is not affected by this option.
+
+A command sent with this option changes the way parameters are handled on the server, because there is no need to maintain an output parameter map. The result of this change is that queries with large numbers of input parameters may execute much faster. 
+
+The fewest number of parameters where this will take effect depends on the individual situation and should be detected by measuring query duration with and without the option enabled. Any query with more than 24 parameters may show lower overall query duration. Queries with parameter counts lower than 24 are unlikely to show a difference.
```
DavoudEshtehari, line 1367: Could you explain in which conditions it would have a negative perf effect (for input parameters)-not neutral- if there was such a condition?

doc/snippets/Microsoft.Data.SqlClient/SqlCommand.xml
```diff
@@ -1346,6 +1346,33 @@ The <xref:Microsoft.Data.SqlClient.SqlCommand.CreateParameter%2A> method is a st
...
+
+> [!NOTE]
+If the option is enabled and a parameter with Direction Output or InputOutput is present in the Parameters collection an InvalidOperationException will be thrown when the command is executed.
```
DavoudEshtehari, line 1370: This property is going to optimize the command execution, I think it'd better to just skip it if it doesn't commit the requirements instead of raising an exception.

doc/snippets/Microsoft.Data.SqlClient/SqlCommand.xml
Wraith2: There isn't a negative case. The only negative consequence is disabling output parameters. For small numbers of parameters the performance will be virtually identical it's only as the numbers rise that the difference becomes meaningful. Check the numbers in https://github.com/dotnet/SqlClient/pull/1041#issuecomment-821995691

doc/snippets/Microsoft.Data.SqlClient/SqlCommand.xml
Wraith2: So just set the value to false? It's compatible but it will mean silent performance regression if someone adds an output parameter to a query that is expected to contain only input parameters. We'd essentially be silently disabling something that the user had to explicitly tell us that they wanted and I'm sure I'd be quite annoyed by that behaviour as a consumer having to track it down as a performance problem.

doc/snippets/Microsoft.Data.SqlClient/SqlCommand.xml
DavoudEshtehari: If there is a match point to decide internally, my suggestion is to have this property with multiple configurations:

#### OptimizedParameterBinding
- Disable (_default_)
- Enable
- Auto

Wraith2: My concern with auto is that we make a change to the way the query is run without user interaction. This leads to the kind of situation where something with 30 parameters works fine but something with 31 works in some unexpected way and since the user can't see a difference in their code it is interpreted as random. The only way to see the problem at that point is probably to use wireshark which is normally a long long way down the list of debugging tools close to last resort because of the difficulty involved.

I view this as an advanced option. I don't believe that this is a casual option that would ever just be defaulted to true. If you opt in then we try to do what you asked and if we can't we throw an exception at you.
The decision making is all down to the user and by default we continue the behaviour that people have been using for the last 20 years so we don't accidentally cause some poor maintenance programmer a really really bad week (or more).

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ParameterTest/ParametersTest.cs
```diff
@@ -520,5 +520,232 @@ private enum MyEnum
...
+            using (SqlCommand cmd = conn.CreateCommand())
+            {
+                conn.Open();
+                cmd.CommandText = cmdText;
+                cmd.ExecuteNonQuery();
+            }
+        }
+
+        [Fact]
+        private static void EnableOptimizedParameterBinding_ParametersAreUsedByName()
```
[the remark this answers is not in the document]
Wraith2: Those scenarios are already covered by the existing parameter tests. I just chose to add these new behaviours as new tests to make it clear what they were testing instead of changing the existing covering tests.

src/Microsoft.Data.SqlClient/netfx/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -550,6 +550,8 @@ public sealed partial class SqlCommand : System.Data.Common.DbCommand, System.IC
         [System.ComponentModel.DesignOnlyAttribute(true)]
         [System.ComponentModel.EditorBrowsableAttribute(System.ComponentModel.EditorBrowsableState.Never)]
         public override bool DesignTimeVisible { get { throw null; } set { } }
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlCommand.xml' path='docs/members[@name="SqlCommand"]/UsePositionalParameters/*'/>
```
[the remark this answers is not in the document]
Wraith2: All renaming cleaned up. I'm not sure how it managed to get the correct name in the tests but not the main project but they're right now, thanks for pointing it out.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -9901,7 +9901,7 @@ internal Task TdsExecuteSQLBatch(string text, int timeout, SqlNotificationReques
 
                             if (mt.IsNewKatmaiType)
                             {
-                                WriteSmiParameter(param, i, 0 != (rpcext.paramoptions[i] & TdsEnums.RPC_PARAM_DEFAULT), stateObj, isAdvancedTraceOn);
+                                WriteSmiParameter(param, i, 0 != (rpcext.paramoptions[i] & TdsEnums.RPC_PARAM_DEFAULT), stateObj, cmd.DisableOutputParameters, isAdvancedTraceOn);
```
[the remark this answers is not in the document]
Wraith2: Done, it also showed that the netfx implementation wasn't identical to netcore (since I was testing only in netcore) but now they are.

DavoudEshtehari:
Could you include the new **InvalidOperationException** conditions in the **Exceptions** section of the `SqlCommand` related functions, please?

Wraith2:
In the docs?

DavoudEshtehari:
Sorry, I didn't point it. Yes, please.

doc/snippets/Microsoft.Data.SqlClient/SqlCommand.xml
```diff
@@ -286,8 +286,13 @@ The following console application creates updates data within the **AdventureWor
...
 				The
 				<see cref="T:Microsoft.Data.SqlClient.SqlConnection" />
-				closed or dropped during a streaming operation. For more information about streaming, see [SqlClient Streaming Support](/sql/connect/ado-net/sqlclient-streaming-support).
-			</exception>
+        closed or dropped during a streaming operation. For more information about streaming, see [SqlClient Streaming Support](/sql/connect/ado-net/sqlclient-streaming-support).
+
+        - or -
+
+        <see cref="P:Microssoft.Data.SqlClient.SqlCommand.EnableOptimizedParameterBinding" />
+        is set to true and a parameter with direction Output or InputOutput has been added to the <see cref="P:Microsoft.Data.SqlClient.SqlCommand.Parameters" />Parameters collection.
```
DavoudEshtehari, line 294: Just remove the extra "**Parameters**" word at the end of the sentence.
- Applies to the other sections too.

`is set to true and a parameter with direction Output or InputOutput has been added to the <see cref="P:Microsoft.Data.SqlClient.SqlCommand.Parameters" /> collection.`

Wraith2:
Documentation added.

doc/snippets/Microsoft.Data.SqlClient/SqlCommand.xml
Wraith2: Done. 
The enclave tests seem to be failing for some reason.

DavoudEshtehari: You should add `ConditionalFact` over your tests. Because those agents don't set up `TCPConnectionString`.
I suggest preparing a set of connection strings to feed up your tests or simply use `DataTestUtility.ConnectionStrings`.

Wraith2: Doh, done.

DavoudEshtehari (approved):
🎉

johnnypham (approved):

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ParameterTest/ParametersTest.cs
```diff
@@ -520,5 +520,232 @@ private enum MyEnum
...
+                    {
+                        reader.Read();
+
+                        int firstOutput = reader.GetInt32(0);
+                        int secondOutput = reader.GetInt32(1);
+
+                        Assert.Equal(firstInput, secondOutput);
+                        Assert.Equal(secondInput, firstOutput);
+                    }
+
```
cheenamalhotra, line 560:
```suggestion
```

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ParameterTest/ParametersTest.cs
```diff
@@ -520,5 +520,232 @@ private enum MyEnum
...
+                    }
+                    catch (SqlException sqlEx)
+                    {
+                        sqlException = sqlEx;
+                    }
+
+                    Assert.NotNull(sqlException);
+                    Assert.Contains("Must declare the scalar variable",sqlException.Message);
+                    Assert.Contains("@DoesNotExist", sqlException.Message);
+
```
cheenamalhotra, line 590:
```suggestion
```

cheenamalhotra (approved):
Just some empty lines, overall LGTM 👍

cheenamalhotra:
cc @David-Engel 
Docs review

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ParameterTest/ParametersTest.cs
Wraith2: fixed.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ParameterTest/ParametersTest.cs
Wraith2: fixed.

David-Engel (approved):
I like it. 👍

Wraith2:
That's my first surface area addition to the library. Everything I've done before has been invisible to users. Woo!

ErikEJ:
@Wraith2 Congrats!  And my first time naming one! 

Do you have a benchmarking suite, would be interesting to test 1.0 against 3.0?

Wraith2:
I've been relying on the benchmarks that @KristianWedberg posted in the original issue and has kindly been running on the interim builds to verify the changes https://github.com/KristianWedberg/sql-batch-insert-performance

ErikEJ:
@Wraith2 I meant general benchmarks that could highlight the general performance improvements made over time, not related to this PR.

Wraith2:
Ah, no. The closest i've got to that is the https://github.com/aspnet/DataAccessPerformance benchmark which recreates the TE Fortunes benchmark. Everything else has been highly situational and micro benched. In general my focus has been memory usage on common paths like opening connections, executing a reader, reading valuetypes etc.

Wraith2:
@ErikEJ looks like 4.0.0 has been released so your new api is public, let loose the blog post! ;)

ErikEJ:
Yay!

Wraith2:
@KristianWedberg i also had some followup thoughts on this. Do you re-use SqlParameter classes a lot when inserting data? And if so does the object boxing overhead of valuetypes show up in profiles?

mburbea:
It'd be nice if there was a way to avoid it, but I doubt it matters too
much compared to IO overheads. I regularly use SqlBulkCopy with large
number of wide rows of mostly doubles and the only thing that tends to show
up on my profiling is `WriteToServerAsync()`

Wraith2:
I think there is a way that the boxing could be avoided but it's only profitable with with repeated uses of the same parameter. If you're seeing WriteToServerAsync you're probably running a cpu profile in which case you'd see boxing as GC time.

KristianWedberg:
@Wraith2 Yes, the SqlParameters get heavily reused. I pre-create a few batch insert statements (e.g. 1, 32, and 1024 rows) with associated SqlParameters, and generate a function to quickly and repeatedly set the parameters with new values for each new batch.

As you noted during the investigation, value types will be boxed and generates GC pressure. This didn't affect the performance in any of my benchmarks (I also tested with pre-boxed value types), but if bottlenecked on client CPU cycles this could of course matter to some degree. The amount of boxing is simply one per value type inserted per inserted row, and I'm guessing all generation 0, so the impact should be easy to estimate.

I could implement this pre-boxing also in my production code, but that's currently on a fairly low priority. How were you thinking  of addressing this from your end?

Wraith2:
I've got an experimental branch which allows SqlParameter to detect and use `StrongBox<T> : where T is [a supported value type]`  in the Value field. As long as you keep hold of the StrongBox you can read and write from it's Value field without boxing. Obviously that's only beneficial if you re-use the parameter value multiple times since the StrongBox is a heap allocation.

KristianWedberg:
Gotcha. In my scenario I'm guessing I would store the `StrongBox<T>` instances in one array per column in the insert batch (since they would be of the same type), and then generate the code to assign these instances instead of the parameters.

Bit of work, but certainly doable. There is however nothing SqlClient specific with this improvement - it would be really sweet if this was an optional API on `DbParameter` so that any provider could implement it - any chance of that?

Wraith2:
Other providers wouldn't need it, most of them have a `DbParameter<T>` construct of their own or no need for it. In SqlClient we can't replace SqlParameter easily or make it generic without large changes. Since the DbParameter is largely abstract every provider has it's own logic to deal with the values so there's no common mechanism we could usefully expose.

Being able to quietly understand a generic argument container like would be backwards compatible and perform better when opted into. There's a small amount of complexity around nulls but as long as it's properly documented it wouldn't be difficult for anyone who has used direct ADO in the past to deal with.

KristianWedberg:
Even so, why not design it so that it (optionally) _can_ be reused across providers, even if it means they might have both their own API and the portable API? 

I might not be able to justify implementing this for a non-portable API, since I'm generating the parameter setting code with expression trees, and doing that differently for different providers could be more complexity than it's worth...

Wraith2:
That would need runtime implementation planning and work and then pushing out to all providers. So far that level of co-operation hasn't been possible. We can see what @roji thinks but I expect it would be a no-go for a general provider feature because as i said other providers can handle it better already.

roji:
> Other providers wouldn't need it, most of them have a DbParameter<T> construct of their own or no need for it. In SqlClient we can't replace SqlParameter easily or make it generic without large changes.

FWIW I'm not aware of another provider apart from Npgsql which has `DbParameter<T>`. https://github.com/dotnet/runtime/issues/17446 is about adding a generic parameter API to ADO.NET; I'm planning to maybe take a stab at this for 7.0, ideally there would be a default implementation which delegates to non-generic DbParameter and which therefore boxes (but could providers could implement it without boxing).

>  SqlClient we can't replace SqlParameter easily or make it generic without large changes.

Note that at least in the current Npgsql implementation, `NpgsqlParameter<T>` doesn't replace the non-generic NpgsqlParameter - it's just there alongside it. So there wouldn't be any breaking changes or anything to do the same in SqlClient. You can look at [what I did in Npgsql](https://github.com/npgsql/npgsql/blob/main/src/Npgsql/NpgsqlParameter%60.cs) for inspiration, though it would be good to lock down a good design at the ADO.NET level first before implementing.

> Even so, why not design it so that it (optionally) can be reused across providers, even if it means they might have both their own API and the portable API?

Well, first of all, what would there actually be here to design across providers, in System.Data? SqlClient can simply support writing when  SqlParameter.Value is set to `StrongBox<T>` - this wouldn't require any additional API surface here, right? Or am I missing something?

Furthermore, if SqlClient does implement this, then it seems like a generic `SqlParameter<T>` could easily be implemented on top of it... In other words, when somebody sets `SqlParameter<T>.Value` (which would be of type T), that could simply set the value on an internally-managed `StrongBox<T>`, no?

I guess what I'm saying, is that if SqlClient can support writing `StrongBox<T>`, then it's a bit hard to see why it couldn't also just add a `SqlParameter<T>` implementation directly, which is a better API to expose to users... There may be some details here that I'm missing though.

Wraith2:
> I guess what I'm saying, is that if SqlClient can support writing StrongBox<T>, then it's a bit hard to see why it couldn't also just add a SqlParameter<T> implementation directly, which is a better API to expose to users... There may be some details here that I'm missing though.

In technical terms there is some difference but it could be worked out. The SqlParameter implementation leans heavily on reflection to identify the type so a lot of internal logic would need to change if we no longer have the field. The hack i've done sort of sidesteps it but still needs the `object value` internal field it just contains the strongbox so setting it has no downside.

In practical terms getting the interface prototyped, approved, implemented in the BCL, implemented in SqlClient and then available to consumers is a multi-year effort and so far we've been unable to get The batching Api more than halfway through that process in SqlClient despite the other major providers having completed it. 

Hacks in this one library that don't affect the external interface are much "easier" to get implemented than trying to change api surface. Here it might only take a year.

roji:
> In practical terms getting the interface prototyped, approved, implemented in the BCL, implemented in SqlClient and then available to consumers is a multi-year effort and so far we've been unable to get The batching Api more than halfway through that process in SqlClient despite the other major providers having completed it.

The batching API was delivered in .NET 6.0. I wouldn't estimate that a generic `DbParameter<T>` in System.Data is a multi-year effort (though it may slip due to prioritization considerations).

If you feel that eliminating parameter boxing is urgent/high-value enough, then you can indeed go ahead and add support for StrongBox. Given that there's an intention to solve this at the System.Data level, I personally don't see a reason to rush this - it's been this way since forever, it's an optimization that probably isn't terribly important in the SqlClient context (because of the other perf issues), and importantly, when the System.Data API is done you'd be stuck maintaining both the standard `DbParameter<T>` and the SqlClient-specific StrongBox support (which at that point would be obsolete). So I personally wouldn't do this - there's enough other important work - but this would of course be a SqlClient decision.

PR #1017, opened 2021-04-01
Fix TvpTest intermittent failure with Azure SQL Server

DavoudEshtehari:
**Unexpected exception message**: `Invalid command sent to ExecuteXmlReader. The command must return an Xml result.`

PR #1046, opened 2021-04-23
Tests | Fix TvpTest intermittent failures

DavoudEshtehari:
Extending the PR #1017 approach in similar scenarios.

### DIFF
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ParameterTest/StreamInputParam.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ParameterTest/StreamInputParam.cs
@@ -457,17 +457,16 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests
                     byte[] data = new byte[dataSize];
                     rand.NextBytes(data);
                     MemoryStream ms = new MemoryStream(data, false);
-                    cmd.CommandText = "insert into #blobs (Id, blob) values (1, @blob)";
+                    // Include a delay to allow time for cancellation
+                    cmd.CommandText = "WAITFOR DELAY '00:00:05'; insert into #blobs (Id, blob) values (1, @blob)";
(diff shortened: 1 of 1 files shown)


src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ParameterTest/StreamInputParam.cs
```diff
@@ -504,17 +503,16 @@ private static void ImmediateCancelText()
                     StringBuilder sb = new StringBuilder();
                     for (int i = 0; i < 1000000; i++)
                         sb.Append(i);
-                    cmd.CommandText = "insert into #blobs (Id, blob) values (1, @blob)";
+                    // Include a delay to allow time for cancellation
+                    cmd.CommandText = "WAITFOR DELAY '00:00:05'; insert into #blobs (Id, blob) values (1, @blob)";
```
JRahnama, line 507: Curious about the delay time. How did we end up to that exact number of 5 seconds? Is it based on experience or a fact or I am missing something?

DavoudEshtehari: It's based on experience; you can find it in another test in the same class that it's already running well.

cheenamalhotra: Can we avoid this delay or reduce it to 1-2 secs? It eventually adds up to test execution time 👀

DavoudEshtehari: These tests are supposed to be canceled immediately, and this waiting time is just for safety. I'm afraid reducing them may increase the chance of failure.

JRahnama (approved):

PR #358, opened 2019-12-18
SqlBulkCopy - generics to avoid boxing

cmeyertons:
This PR is in reference to #353 

This is an attempt to reduce garbage created during a SqlBulkCopy operation for applications that are using SqlBulkCopy to stream batches of data out during a larger overall process.

The overall approach was to be more respectful of the `IDataReader` interface and use it's typed methods if possible to eliminate uneccesary boxing.  The unboxed value is passed around generically and converted back to its underlying type using a `ValueTypeConverter`.

I did observe some performance degradation when passing around by value, kicking off the idea to pass everything by `ref` -- this is more in line with how `SqlBulkCopy` previously behaved, minus the boxing.

While I do see less garbage created, I am seeing some disappointing performance results (abbreviated Benchmark.NET results):

```
|                Method |     Mean ||    Gen 0 | Gen 1 | Gen 2 |  Allocated |
|---------------------- |---------:||---------:|------:|------:|-----------:|
| NewBulkCopy_NewReader | 387.4 ms ||        - |     - |     - | 12017576 B |
| OldBulkCopy_NewReader | 161.9 ms || 250.0000 |     - |     - | 26412872 B |
```

There are still a considerable amount of heap allocations occurring, primarily during `SqlDecimal` -- I have logged the following issue to address: dotnet/runtime#1034

I think there is something different with the way my dlls are compiled?  I'm seeing much slower times in places where I did not make any changes, such as `TdsParser.WriteTokenLength`

My WriteTokenLength: 144ms
Old WriteTokenLength: 30ms

Any assistance in understanding the perf degradation is welcomed and appreciated!

Issue #846, opened 2020-12-11
SqlDataReader.GetSqlDateTime does not support DATETIME2


bjornbouetsmith:
### Describe the bug
It seems like DATETIME2 sql server datatype is not fully or correctly implemented in SqlDataReader.

using method GetDateTime - it handles DATETIME2 correctly

using method GetSqlDateTime  - it throws an 
System.InvalidCastException:

```
Exception message: Specified cast is not valid 
Stack trace:  at Microsoft.Data.SqlClient.SqlBuffer.get_SqlDateTime()
```

### To reproduce

```c#
 [TestMethod]
        public void DateTime2Mapping()
        {
            string Statement = $@"SELECT CAST('2020-12-11 09:17:12.123456' AS DATETIME2) AS Date";
            
            var builder = new SqlConnectionStringBuilder()
            {
                InitialCatalog = "master",
                DataSource = @"any sql connection",
                IntegratedSecurity = true,
            };

            var sqlConnection = new SqlConnection(builder.ConnectionString);
            sqlConnection.Open();
            using (SqlCommand command = new SqlCommand(Statement))
            {
                command.CommandType = CommandType.Text;
                command.Connection = sqlConnection;
                command.CommandText = Statement;
                using (var reader = command.ExecuteReader())
                {
                    while (reader.Read())
                    {
                        var dtColDate = reader.GetSqlDateTime(0);
                        Assert.IsFalse(dtColDate.IsNull);
                        Console.WriteLine(dtColDate.Value);
                    }
                }
            }
        }
```

### Expected behavior
I expect that GetSqlDateTime handles DATETIME2 and returns a SqlDateTime struct

### Further technical details
Microsoft.Data.SqlClient version: 2.1.0
.NET target: Framework 4.7.2
SQL Server version: 15.0.4079.3 (2019)
Operating system: Windows 10

**Additional context**
Why I found this issue, is because I was profiling our code and saw that SqlDataReader.IsDbNull took up a lot of time since we need to check every column for null, since GetXX methods does not handle DBNull - and then I found SqlDataReader.GetSqlXX methods which handles DBNull.

Basically trying to get rid of the overhead in SqlDataReader
```c#
 override public bool IsDBNull(int i)
        {
            {
                CheckHeaderIsReady(columnIndex: i);

                SetTimeout(_defaultTimeoutMilliseconds);

                ReadColumnHeader(i);    // header data only
            }

            return _data[i].IsNull;
        }
```

by swithing to
[...]

PR #1048, opened 2021-04-26
SqlBulkCopy - Leverage Generics to Eliminate Boxing

cmeyertons:
@Wraith2 @cheenamalhotra - This is a re-creation of #358 - there were multiple merge conflicts that were difficult to resolve.

- Leverage Generic Code Paths to eliminate boxing of value types as the value is being written to the database
- Benchmark test that validates the boxing is eliminated (this only works in Release mode as the Debug JIT does not perform the necessary optimizations)
- Extension Method for Generic Cast instead of a GenericConverter type -- this makes the code a bit less verbose + easier to read.

Possible breaking changes:
- `IDataReader.GetFieldType` is invoked in SqlBulkCopy that was not invoked previously.  Custom data reader implementations that do not properly implement this method could break.
- `IDataReader.IsDbNull` is now invoked in the IDataReader code path for every value (previously it was only invoked if streaming was enabled).  This is necessary to invoke the proper value method (GetInt32, etc.) and handle potential nullable value types.

TODO:
- Migrate this code into `netfx` (I want to wait for all tests to pass before I spend any effort doing this)


src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/SqlBulkCopyTest/NoBoxingValuesTypes.cs
```diff
@@ -0,0 +1,487 @@
...
+            var report = summary.Reports.First();
+
+            Assert.Equal(1, report.AllMeasurements.Count);
+            Assert.True(report.GcStats.BytesAllocatedPerOperation < totalBytesWhenBoxed);
+#endif
+        }
+
+        public class NoBoxingValueTypesBenchmark
+        {
+            [Benchmark]
```
cheenamalhotra, line 105: Hi @cmeyertons 

I highly appreciate your contribution here, but please hold on to committing Performance test and benchmarking code to Manual Tests project in this PR. We're going to introduce a new project specifically for Perf testing, that's where we will add benchmark tests.
_(Till then please upload this test as a file attachment, we'll include it to the new perf project)_

/cc @DavoudEshtehari

cmeyertons: @cheenamalhotra  - okay sounds good no problem -- i'm going to momentarily leave it in to test that the benchmark is successful in `netfx` if that's alright with you.

cheenamalhotra:
Hi @cmeyertons 

We appreciate your efforts. 

> I have implemented a custom data reader implementation in the NoBoxingValueTypes test -- I would've preferred to leverage the FastMember library to reduce the amount of code required, but that package was not available in the currently configured Nuget sources.

**FastMember** is not an official Microsoft package, and neither provides a unique solution that keeps us blocked. Yes it does simplify code, but that's not necessary to do the same job. We cannot justify the need to depend on this package for just code enhancements. Please continue to use the existing reflection design in tests to add your tests.

cmeyertons:
@cheenamalhotra based on #1046 -- i'm guessing that some of my test failures are expected at this point?

Wraith2:
You'll still have to check them individually and decide if they're code related of infra, apart from TVPMain, that's just flaky as hell.

cheenamalhotra:
@cmeyertons 

Yes, we have some random failures which should go away once we rerun pipelines, don't bother about those ones. We're trying to fix them, but as Wraith mentioned, they're flaky..

cmeyertons:
@cheenamalhotra  @Wraith2  alright tests are looking much better -- in the process of re-patching the code onto this branch, I did find the bug that was causing tests to fail.

If I could get a preliminary approval on how the code looks now (the potential breaking changes, using the extension method over the GenericConverter class), I can get started on the `netfx` port (I _really_ don't want to change manage across both with a bunch of iterations)

Wraith2:
> If I could get a preliminary approval on how the code looks now (the potential breaking changes, using the extension method over the GenericConverter class), I can get started on the netfx port

It's been 16 months, you've been trying to get this merged so I'd really hope that someone can find the time to review it and give constructive feedback. You've been far more patient than anyone should need to be on this.

cheenamalhotra:
@cmeyertons 

> IDataReader.IsDbNull is now invoked in the IDataReader code path for every value (previously it was only invoked if streaming was enabled). This is necessary to invoke the proper value method (GetInt32, etc.) and handle potential nullable value types.

As you specifically mentioned this now in your PR, I'm a little worried about the `IDataReader.IsDbNull` call that will be made on every value. I suspect this might have a negative impact on performance for large rows. I have previously heard reports on this call causing performance lag (e.g. https://github.com/dotnet/SqlClient/issues/846#issue-762198571 and also from internal teams when using 200+ columns) so doing so for every type may cause significant performance drop, specially on existing customer applications that rely on current behavior.

We of course need to write a use case to prove that, so I'll be looking more into that tomorrow. But if you can also test it out and come up with a use-case to find any significant performance drop with this PR, we may have to revisit the design a bit to make it an optional behavior. But if our test results do not suggest any performance degradation, we'd be happy to take it as is.

Wraith2:
I've investigated the perf impact of `IsDBNull` in the past and I found that the cost of the seek to the field was being attributed to it which made `IsDBNull` seem heavier than it is and `GetFieldValue` calls seem lighter . To check if a column value is null you have to seek to the column header which will run the parser if you haven't already touched the field and at that point IO can also occur. When you call `GetFieldValue` on that same column the seek should already have happened from the previous call so you usually just pick up the value in the `SqlBuffer`. 

The fastest call is one that never happens so it would be good to get some numbers. For this PR can you demonstrate (with a BenchmarkDotNet bench if possible) that the original/current implementation is still slower than the PR branch with a reasonable column width of say 32 or 64? You'll need to give it a runtime of at least a minute ime to make sure the GC kicks in for the boxes that are being removed, without the GC the overhead of the new approach won't be properly balanced against the waste of the current one and your PR will look artificially poor, been there made that mistake.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/SqlBulkCopyTest/NoBoxingValuesTypes.cs
cmeyertons: [NoBoxingValuesTypes.zip](https://github.com/dotnet/SqlClient/files/6384644/NoBoxingValuesTypes.zip)

@cheenamalhotra ^ attached a zip of the file (uncommented out)

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
```diff
@@ -965,37 +985,66 @@ private object GetValueFromSourceRow(int destRowIndex, out bool isSqlType, out b
...
                     }
                     else
                     {
-                        isDataFeed = false;
-
                         IDataReader rowSourceAsIDataReader = (IDataReader)_rowSource;
-
-                        // Only use IsDbNull when streaming is enabled and only for non-SqlDataReader
-                        if ((_enableStreaming) && (_sqlDataReaderRowSource == null) && (rowSourceAsIDataReader.IsDBNull(sourceOrdinal)))
+                        // previously, IsDbNull was only invoked in a non-streaming scenario with a non-SqlDataReader.
```
cmeyertons, line 994: @cheenamalhotra @Wraith2 starting a conversation here to keep this tied together -- I will create a benchmark around this logic that inserts a million records in a loop until i hit about the minute mark.

I will do this in a couple of other branches in my repo and will post some links to the code here -- should be able to get this done today.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
```diff
@@ -1260,6 +1321,66 @@ private SourceColumnMetadata GetColumnMetadata(int ordinal)
...
                 }
             }
+            else if (_rowSourceType == ValueSourceType.IDataReader)
+            {
+                isSqlType = false;
+                isDataFeed = false;
+
+                Type t = ((IDataReader)_rowSource).GetFieldType(ordinal);
+
+                if (t == typeof(bool))
```
cmeyertons, line 1331: @cheenamalhotra @Wraith2 

Thinking about this in the context of the `IsDbNull` check above -- I think we need to add an OR check to the corresponding nullable type (e.g. `t == typeof(bool) || t == typeof(bool?`)

I think we have two routes we can go here.

1. We add the OR check and invoke `IsDbNull` on each value.  Pros: prevents nullable types from being boxed.  Cons: have to invoke IsDbNull (per impact TBD) and requires data reader implementations to correct determine result of `IsDbNull`
2. Leave this as is and we don't have to invoke `IsDbNull` on each value.  Nullable value types would get boxed.

I would prefer option 1 if possible.

I will measure the option 1 approach as discussed and we will see how it shakes out.

Wraith2: I don't believe that GetFieldType can return a Nullable<bool>, at least it can't in the SqlClient implementation. We don't conflate sql and language nulls (rightly imo) so I don't think you need to change this. If a non-SqlClient data source provides a Nullable<bool> then I think boxing is probably appropriate.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
cmeyertons: If IsDbNull proves to be problematic, there could be an optimization around only invoking it for fields that are possibly null (e.g. if `GetFieldType` returns `typeof(int)` we don't need to check

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
Wraith2: A nullable int column should have GetFieldType of int, nullability is not part of the type in sql.

cmeyertons:
@cheenamalhotra @Wraith2 after much further investigation, it appears that migrating the code paths from `object value` to `T value` appears to have a detrimental impact on performance  -- I see about a 10% increase across the board.

You can view my the benchmarking code in my repo in the `benchmarks/current` and `benchmarks/new` branches.  I am bulk copying appx 40 columns

Some observations:
- This does not appear to be solely related to `IsDbNull` -- i see the 10% uptick on various submethods that were genericized.
- This might be due to passing `decimal` by value (128 bits instead of 64).  I did attempt to change all of the signatures to `ref` and that did not alleviate the problem.
- Maybe leveraging the `in` modifier would help -- however, this change is probably not backwards compatible.
- Some extra boxing is forced to occur during `CoerceValue` if coercion is necessary -- eliminating boxing would require a generic version of `Convert.ChangeType` which is a framework change.
- I did uncover a small performance change that I can submit in a subsequent PR

Inevitably, the boxing cost is much lower than I initially anticipated.  While this was a good and fun personal exercise in getting into the lowers details of the SqlClient library, I'm at a deadend on how to improve performance at this point.

# CURRENT
``` ini

BenchmarkDotNet=v0.12.1, OS=Windows 10.0.18363.1500 (1909/November2018Update/19H2)
Intel Xeon E-2176M CPU 2.70GHz, 1 CPU, 12 logical and 6 physical cores
.NET Core SDK=5.0.202
  [Host]     : .NET Core 3.1.14 (CoreCLR 4.700.21.16201, CoreFX 4.700.21.16208), X64 RyuJIT
  Job-ZNZZJB : .NET Core 3.1.14 (CoreCLR 4.700.21.16201, CoreFX 4.700.21.16208), X64 RyuJIT

IterationCount=5  LaunchCount=1  WarmupCount=0

```
|   Method |    Mean |   Error |  StdDev |       Gen 0 |     Gen 1 | Gen 2 | Allocated |
|--------- |--------:|--------:|--------:|------------:|----------:|------:|----------:|
| BulkCopy | 17.91 s | 1.270 s | 0.330 s | 364000.0000 | 2000.0000 |     - |   2.13 GB |

# NEW

``` ini
BenchmarkDotNet=v0.12.1, OS=Windows 10.0.18363.1500 (1909/November2018Update/19H2)
Intel Xeon E-2176M CPU 2.70GHz, 1 CPU, 12 logical and 6 physical cores
.NET Core SDK=5.0.202
  [Host]     : .NET Core 3.1.14 (CoreCLR 4.700.21.16201, CoreFX 4.700.21.16208), X64 RyuJIT
  Job-UIKJLZ : .NET Core 3.1.14 (CoreCLR 4.700.21.16201, CoreFX 4.700.21.16208), X64 RyuJIT

IterationCount=5  LaunchCount=1  WarmupCount=0

```
|   Method |    Mean |   Error |  StdDev |       Gen 0 |     Gen 1 | Gen 2 | Allocated |
|--------- |--------:|--------:|--------:|------------:|----------:|------:|----------:|
| BulkCopy | 20.01 s | 2.480 s | 0.384 s | 178000.0000 | 2000.0000 |     - |   1.04 GB |

Wraith2:
The memory is a lot better which was the initial goal. The speed drop is a bit troubling. Does any particular hotspot stick out?

cheenamalhotra:
Hi @cmeyertons 

I extended your project to include more info (+10 columns +Async APIs +SqlDataReader +Threading Diagnoser).
I also used the same data to keep things comparable. Below are the results for ~50 columns:
My source changes are here: [fork: benchmarks/new](https://github.com/cheenamalhotra/SqlClient/tree/benchmarks/new)

### CURRENT - IDataReader

|        Method |    Mean |    Error |  StdDev | Completed Work Items |       Gen 0 |     Gen 1 | Allocated |
|-------------- |--------:|---------:|--------:|---------------------:|------------:|----------:|----------:|
|      BulkCopy | 22.82 s | 12.746 s | 1.972 s |               2.0000 | 646000.0000 | 2000.0000 |   2.52 GB |
| BulkCopyAsync | 22.48 s |  1.601 s | 0.416 s |             250.0000 | 646000.0000 | 2000.0000 |   2.52 GB |

### NEW - IDataReader

|        Method |    Mean |    Error |  StdDev | Completed Work Items |       Gen 0 |     Gen 1 | Allocated |
|-------------- |--------:|---------:|--------:|---------------------:|------------:|----------:|----------:|
|      BulkCopy | <sup>1</sup> 25.86 s | 13.320 s | 2.061 s |               4.0000 | 294000.0000 | 2000.0000 |   <sup>3</sup> 1.15 GB |
| BulkCopyAsync | <sup>1</sup> 26.16 s |  1.946 s | 0.505 s | <sup>2</sup> 645.0000 | 294000.0000 | 2000.0000 |  <sup>3</sup> 1.15 GB |

<sup>1</sup> Performance degradation is approx. 13% (sync) &19% (async).
<sup>2</sup> Async flow consumes significantly more threads here.
<sup>3</sup> There's no doubt allocation has improved.

--------

### CURRENT - SqlDataReader

|        Method |    Mean |    Error |  StdDev |       Gen 0 |     Gen 1 | Allocated | Completed Work Items |
|-------------- |--------:|---------:|--------:|------------:|----------:|----------:|---------------------:|
|      BulkCopy | 28.81 s |  2.297 s | 0.355 s | 371000.0000 | 3000.0000 |   1.45 GB |              10.0000 |
| BulkCopyAsync | 65.16 s | 19.408 s | 5.040 s | 802000.0000 | 5000.0000 |   3.04 GB |         3924955.0000 |

### NEW - SqlDataReader

|        Method |    Mean |   Error |  StdDev | Completed Work Items |       Gen 0 |     Gen 1 | Allocated |
|-------------- |--------:|--------:|--------:|---------------------:|------------:|----------:|----------:|
|      BulkCopy | <sup>1</sup> 34.56 s | 2.044 s | 0.531 s |               3.0000 | 371000.0000 | 3000.0000 |   1.45 GB |
| BulkCopyAsync | <sup>1</sup> 70.56 s | 2.968 s | 0.459 s |         3925037.0000 | 804000.0000 | 7000.0000 |   3.04 GB |

<sup>1</sup> Performance degraded by approx. 20% (sync) and 8% (async), even though there's no change in allocation.
Was this intended or can you improve allocations for `SqlDataReader` too?

cmeyertons:
@cheenamalhotra @Wraith2 

> 1 Performance degraded by approx. 20% (sync) and 8% (async), even though there's no change in allocation.
Was this intended or can you improve allocations for SqlDataReader too?

Looking at the SqlDataReader internals, it looks like it holds data in a SqlBuffer object, so it might already be boxed -- i was anticipating less allocations due to invocations through `GetSqlDecimal` etc.

I've refactored some code paths to closer align the # of method invocations to the original -- refactored `ConvertWriteValueAsync` back to more similar of the original `ConvertValue` (changed to `ConvertValueIfNeeded`)

I lifted out the null checking from `ConvertValueIfNeeded` to eliminate the method dispatch for nulls (this is a small perf improvement that could be made to the existing codebase)

Some of my suspicions at this point are:
- The `IfNeeded` pattern is adding the amount of local memory used (instead of writing over the existing value, we're passing back a new, second value) -- even if this is null most of the time, there could be extra cost there.  I see jumps in both `SqlBulkCopy.ConvertValueIfNeeded` and `SqlParameter.CoerceValueIfNeeded`
- I'm wondering if this pattern introduces more CPU cache misses as you are hitting different code paths for each column in the dataset, incurring more context switching -- I have no evidence of this, just a hunch.
- There are a couple of methods that I could additionally inline for the `object` vs `T` use case -- `TdsParser.WriteValueWithWait` and `SqlBulkCopy.DoWriteValueAsync` -- these are the only extra invocations at this point that I can see.  That would require some code duplication however.

I will work on getting your benchmarks into my repo later today -- @cheenamalhotra  thanks for enhancing them, much appreciated it!

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
mburbea: If you're inspecting the SchemaTable there is IsNullable field, if it's not present you should be able to skip the dbnull checks.

cmeyertons:
Hi team, closing this PR to get it off my list.  I no longer am working with MSSQL and it appears this experiment hampered performance rather than helped it (most likely due to CPU caching instructions, etc).

I sincerely appreciate everyone's help throughout this attempted contribution and learned a ton!

PR #1057, opened 2021-04-30
Share SqlNormalizer

Wraith2:
Fairly simple change. I cleaned up sqlnorm which were functionally identical but had minor code differences between netcore and netfx and then moved the virtually identical resulting file to shared and added in #if netframework around the 3 places that CAS is used. I'd really rather have a tiny amount of ifdefs than have two separate versions of these files


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SqlNormalizer.cs
```diff
@@ -20,23 +20,23 @@ internal sealed class FieldInfoEx : IComparable
...
-            FieldInfoEx otherF = other as FieldInfoEx;
-            if (otherF == null)
+            if (other is FieldInfoEx otherFieldInfoEx)
             {
-                return -1;
+                return _offset.CompareTo(otherFieldInfoEx._offset);
             }
-            return _offset.CompareTo(otherF._offset);
+            return -1;           
         }
```
DavoudEshtehari, line 40: Just as a suggestion: using `is` pattern.
```
public int CompareTo(object other) 
     => other is FieldInfoEx otherF ? _offset.CompareTo(otherF._offset): -1;
```

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SqlNormalizer.cs
```diff
@@ -66,9 +69,12 @@ internal BinaryOrderedUdtNormalizer(Type t, bool isTopLevelUdt)
                 _isTopLevelUdt = true;
             }
 
+            // top level udt logic is disabled until we decide
+            // what to do about nested udts
             _isTopLevelUdt = true;
+            //      else
+            //        this.m_isTopLevelUdt = isTopLevelUdt;
```
DavoudEshtehari, line 76:
```suggestion
            //        this._isTopLevelUdt = isTopLevelUdt;
```

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SqlNormalizer.cs
```diff
@@ -66,9 +69,12 @@ internal BinaryOrderedUdtNormalizer(Type t, bool isTopLevelUdt)
                 _isTopLevelUdt = true;
```
DavoudEshtehari, line 69: I don't see any usage for `isTopLevelUdt` parameter in this function. Seems it'd been used in the `else` section. 🤔

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SqlNormalizer.cs
```diff
@@ -129,12 +136,12 @@ private object DeNormalizeInternal(Type t, Stream s)
         {
             object result = null;
             //if nullable and not the top object, read the null marker
-            if (!_isTopLevelUdt && typeof(INullable).IsAssignableFrom(t))
+            if (!_isTopLevelUdt && typeof(System.Data.SqlTypes.INullable).IsAssignableFrom(t))
```
DavoudEshtehari, line 139: There are two more usage for `INullable` interface in this file; If you like to use the whole name of it like this, please apply the same method for all of them and remove the using namespace from the top of file.

Wraith2: The two files were a mix of styles. Which do you prefer?

DavoudEshtehari: I'd say `if (!_isTopLevelUdt && typeof(INullable).IsAssignableFrom(t))` to keep it identical like the other usages.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SqlNormalizer.cs
Wraith2: Done.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SqlNormalizer.cs
Wraith2: Yes. I left it there in the comments in case it's needed later.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SqlNormalizer.cs
Wraith2: Done.

DavoudEshtehari (approved):
Thanks for the quick response.

cheenamalhotra (approved):

PR #934, opened 2021-02-26
Simplify SNIProxy

Wraith2:
SniProxy is used as a single instance and called like this:
```csharp
internal override uint SniGetConnectionId(ref Guid clientConnectionId) => SNIProxy.GetInstance().GetConnectionId(Handle, ref clientConnectionId);
```
This involves calling the GetInstance() method to get the single instance and then calling a function on that instance which only uses the parameters that are passed in. In these cases the functions could be static or merged into the callsite if there is only one use of the function.

The only thing that can't be removed from this single instance pattern is the GetLastError functions. I've investigated what it would take to remove it and it's very very complicated and involves significant changes to the StateObject interface between shared and platform implementations. I choose not to do that now.

PR #1060, opened 2021-05-03
Perf: Static delegates

Wraith2:
Now we've got c#9 i've audited all lambda expressions in the library and changed some of them over to using static delegates which allows the compiler to cache the delegate on first use. 

I also adjusted some other callsites to use state parameters to reduce the size of the closure slightly or allow onFailure, or onCancellation delegates to be static. 

I tried to apply a consistent style to all the lambda usage layout spacing named params etc. It feels better to me overall but let me know if you want a different style.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
@@ -9,6 +9,7 @@ using System.Data;
 using System.Data.Common;
 using System.Data.SqlTypes;
 using System.Diagnostics;
+using System.Runtime.CompilerServices;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
@@ -2902,11 +2903,11 @@ namespace Microsoft.Data.SqlClient
                 {
                     if (_isAsyncBulkCopy)
                     {
-                        CancellationTokenRegistration regReconnectCancel = new CancellationTokenRegistration();
+                        StrongBox<CancellationTokenRegistration> regReconnectCancel = new StrongBox<CancellationTokenRegistration>(new CancellationTokenRegistration());
                         TaskCompletionSource<object> cancellableReconnectTS = new TaskCompletionSource<object>();
                         if (cts.CanBeCanceled)
                         {
-                            regReconnectCancel = cts.Register(static (object tcs) => ((TaskCompletionSource<object>)tcs).TrySetCanceled(), cancellableReconnectTS);
+                            regReconnectCancel.Value = cts.Register(static (object tcs) => ((TaskCompletionSource<object>)tcs).TrySetCanceled(), cancellableReconnectTS);
                         }
                         AsyncHelper.ContinueTaskWithState(reconnectTask, cancellableReconnectTS,
                             state: cancellableReconnectTS,
@@ -2926,7 +2927,7 @@ namespace Microsoft.Data.SqlClient
                             state: regReconnectCancel,
                             onSuccess: (object state) =>
                             {
-                                ((CancellationTokenRegistration)state).Dispose();
+                                ((StrongBox<CancellationTokenRegistration>)state).Value.Dispose();
                                 if (_parserLock != null)
                                 {
                                     _parserLock.Release();
@@ -2936,8 +2937,8 @@ namespace Microsoft.Data.SqlClient
                                 _parserLock.Wait(canReleaseFromAnyThread: true);
                                 WriteToServerInternalRestAsync(cts, source);
                             },
-                            onFailure: static (Exception _, object state) => ((CancellationTokenRegistration)state).Dispose(),
-                            onCancellation: static (object state) => ((CancellationTokenRegistration)state).Dispose(),
+                            onFailure: static (Exception _, object state) => ((StrongBox<CancellationTokenRegistration>)state).Value.Dispose(),
+                            onCancellation: static (object state) => ((StrongBox<CancellationTokenRegistration>)state).Value.Dispose(),
                             exceptionConverter: (ex) => SQL.BulkLoadInvalidDestinationTable(_destinationTableName, ex));
                         return;
                     }
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
@@ -1078,14 +1078,19 @@ namespace Microsoft.Data.SqlClient
             if (_isAsyncBulkCopy && _dbDataReaderRowSource != null)
             {
                 // This will call ReadAsync for DbDataReader (for SqlDataReader it will be truly async read; for non-SqlDataReader it may block.)
-                return _dbDataReaderRowSource.ReadAsync(cts).ContinueWith((t) =>
-                {
-                    if (t.Status == TaskStatus.RanToCompletion)
+                return _dbDataReaderRowSource.ReadAsync(cts).ContinueWith(
+                    static (Task<bool> task, object state) =>
                     {
-                        _hasMoreRowToCopy = t.Result;
-                    }
-                    return t;
-                }, TaskScheduler.Default).Unwrap();
+                        SqlBulkCopy sqlBulkCopy = (SqlBulkCopy)state;
+                        if (task.Status == TaskStatus.RanToCompletion)
+                        {
+                            sqlBulkCopy._hasMoreRowToCopy = task.Result;
+                        }
+                        return task;
+                    },
+                    state: this,
+                    scheduler: TaskScheduler.Default
+                ).Unwrap();
             }
             else
             { // This will call Read for DataRows, DataTable and IDataReader (this includes all IDataReader except DbDataReader)
@@ -1940,7 +1945,7 @@ namespace Microsoft.Data.SqlClient
                         {
                             AsyncHelper.ContinueTaskWithState(writeTask, tcs,
                                 state: tcs,
-                                onSuccess: state => ((TaskCompletionSource<object>)state).SetResult(null)
+                                onSuccess: static (object state) => ((TaskCompletionSource<object>)state).SetResult(null)
                             );
                         }
                     }, ctoken); // We do not need to propagate exception, etc, from reconnect task, we just need to wait for it to finish.
@@ -1948,7 +1953,7 @@ namespace Microsoft.Data.SqlClient
                 }
                 else
                 {
-                    AsyncHelper.WaitForCompletion(reconnectTask, BulkCopyTimeout, () => { throw SQL.CR_ReconnectTimeout(); }, rethrowExceptions: false);
+                    AsyncHelper.WaitForCompletion(reconnectTask, BulkCopyTimeout, static () => throw SQL.CR_ReconnectTimeout(), rethrowExceptions: false);
                 }
             }
 
@@ -1969,27 +1974,32 @@ namespace Microsoft.Data.SqlClient
                 if (resultTask != null)
                 {
                     finishedSynchronously = false;
-                    return resultTask.ContinueWith((t) =>
-                    {
-                        try
+                    return resultTask.ContinueWith(
+                        static (Task t, object state) =>
                         {
-                            AbortTransaction(); // If there is one, on success transactions will be committed.
-                        }
-                        finally
-                        {
-                            _isBulkCopyingInProgress = false;
-                            if (_parser != null)
+                            SqlBulkCopy sqlBulkCopy = (SqlBulkCopy)state;
+                            try
                             {
-                                _parser._asyncWrite = false;
+                                sqlBulkCopy.AbortTransaction(); // If there is one, on success transactions will be committed.
                             }
-                            if (_parserLock != null)
+                            finally
                             {
-                                _parserLock.Release();
-                                _parserLock = null;
+                                sqlBulkCopy._isBulkCopyingInProgress = false;
+                                if (sqlBulkCopy._parser != null)
+                                {
+                                    sqlBulkCopy._parser._asyncWrite = false;
+                                }
+                                if (sqlBulkCopy._parserLock != null)
+                                {
+                                    sqlBulkCopy._parserLock.Release();
+                                    sqlBulkCopy._parserLock = null;
+                                }
                             }
-                        }
-                        return t;
-                    }, TaskScheduler.Default).Unwrap();
+                            return t;
+                        }, 
+                        state: this,
+                        scheduler: TaskScheduler.Default
+                    ).Unwrap();
                 }
                 return null;
             }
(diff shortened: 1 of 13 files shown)


jinek:
I think you are confusing [anonymous methods](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/delegate-operator) and [expression trees](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/expression-trees/) both of which can be created using [lambda expressions](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/lambda-expressions).

Anonymous methods are compiled by c# compiler once at compilation time, while expression trees can be [compiled](https://docs.microsoft.com/en-us/dotnet/api/system.linq.expressions.expression-1.compile?view=net-5.0) to a delegate at runtime.
I've created small example to demonstrate how anonymous methods are compiled:
image
You can also investigated that using windbg tool which shows the method table information for a given type-object.

Static anonymous function has been introduced to restrict the access to instance and local variables if it is not intended so by the design.
However, one of the main benefits of anonymous delegates and local functions is the possibility to capture the variables and instance fields of a method so you don't have to pass them as parameters.
And this is my point here: you are increasing the code base to remove the use of the feature which has been introduced to decrease the codebase:
image

Wraith2:
Captures contribute to the size and number of hidden classes generated by the compiler. Passing using existing state parameter placement decreases or obviates the need for captures at all. Using static lambdas allows the compiler to cache the delegate again avoiding per-call allocations.

I do know what I'm doing here. Try running a memory profile of an async test case with and without the PR and see what difference it makes.

jinek:
> Captures contribute to the size and number of hidden classes generated by the compiler. Passing using existing state parameter placement decreases or obviates the need for captures at all. Using static lambdas allows the compiler to cache the delegate again avoiding per-call allocations.

If you are talking about delegate object itself then does it make sense to do optimization in this code? Because probably even the average connection string is several times bigger than a delegate and captured state together.

> I do know what I'm doing here. Try running a memory profile of an async test case with and without the PR and see what difference it makes.

Could you please share the result of your experiment? 
(because from your PR the only metric seen at the moment has negative impact)

Wraith2:
> If you are talking about delegate object itself then does it make sense to do optimization in this code? Because probably even the average connection string is several times bigger than a delegate and captured state together.

Each connection string is allocated once and re-used. The paths through these methods allocate on each call. If you call them once you'll see a very minor improvement, if you call them thousands or millions of times that adds up considerably. It's a question of scale.

Results from running the https://github.com/aspnet/DataAccessPerformance which is an implementation of the TE Fortunes benchmark.
| Test | Type | Threads | Througput | StdDev | branch
| --------------------------- | ----- | -- | ----- | ---- | -----
| ado-sqlclient+async+16 | async | 16 | 26411 | 1040 | perf-staticdelegates
| ado-sqlclient+async+16 | async | 16 | 26165 | 1101 | main

So a 0.9% throughput increase from some very minor changes.

These changes barely impact the size of the code in comparison to the overall size of the codebase. In general the size of the code is far less relevant than the correctness and speed. If your main concern is code size at the driver level I thing your priorities are wrong.

Wraith2:
@cmeyertons there are a number of changes in bulk copy here and they may improve the performance enough to outweigh the problem in https://github.com/dotnet/SqlClient/pull/1048 if you want to give them a try.

jinek:
> Each connection string is allocated once and re-used. The paths through these methods allocate on each call. If you call them once you'll see a very minor improvement, if you call them thousands or millions of times that adds up considerably. It's a question of scale.

Sure, I completely agree scaling is important as well as performance overall. It's just, even if we run simple non-query request with new connection, it consumes about 25Kb, and empty delegate is 32B . Compiler does optimization and caches the delegate if possible whether it's static or not, it not necessary creates hidden classes if there is no need to capture the stack, right? 
 
> Results from running the https://github.com/aspnet/DataAccessPerformance which is an implementation of the TE Fortunes benchmark.
> Test 	Type 	Threads 	Througput 	StdDev 	branch
> ado-sqlclient+async+16 	async 	16 	26411 	1040 	perf-staticdelegates
> ado-sqlclient+async+16 	async 	16 	26165 	1101 	main
> 
> So a 0.9% throughput increase from some very minor changes.
> 
> These changes barely impact the size of the code in comparison to the overall size of the codebase. In general the size of the code is far less relevant than the correctness and speed. If your main concern is code size at the driver level I think your priorities are wrong.

That interesting tool uses DateTime.UtcNow to measure time intervals, Task.Wait for synchronization etc. 
But there are methodologies of measurements, this topic is not new. I'd suggest you to measure with something like https://benchmarkdotnet.org/ and check if your improvements do more than the measurement error.
Also I believe, you, while being an expirienced dev, could investigate some of the allocations which you can find in mem profiler tools, for example, I would check why it allocates many bytes, may be it does some redundant IO. Or, for example, what is going on in _SqlMetadata 
image

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlUtil.cs
```diff
@@ -64,7 +64,7 @@ internal static Task CreateContinuationTaskWithState(Task task, object state, Ac
             {
                 var completion = new TaskCompletionSource<object>();
                 ContinueTaskWithState(task, completion, state,
-                    onSuccess: (continueState) =>
+                    onSuccess: (object continueState) =>
```
DavoudEshtehari, line 67: Same question here for using `Tuple`.

DavoudEshtehari (commented):
Just to double-check; The following methods have used lambda expression. Please look at them if they're worth being static:

- Net Core:
  - SqlBulkCopy.**WriteToServerInternalRestAsync**
  - SqlCommand.**GetParameterEncryptionDataReaderAsync**
  - SqlConnection.**OnError**

Wraith2:
- SqlBulkCopy.WriteToServerInternalRestAsync : Closes over 4 parameters including `this`. We can't avoid the allocation and a 4 tuple is less readable than the closure. I haven't profiled `SqlBulkCopy` scenarios so I don't have information that makes me think that changing this is beneficial at the moment.
- GetParameterEncryptionDataReaderAsync: The result is cached so it isn't per-call. I haven't started profiling always encrypted scenarios yet so I don't know if this is hot, the cache means it should not be.
- OnError: only 2 parameters so it could be tupled easily and fairly clearly but the allocation still can't be avoided. We're about to throw an exception which is going to allocation the exception, it's message, error collection, error collection contents, stacktrace. Attempting to save memory at this point only hurts code clarity it does nothing to benefit the user or us.

In general the more parameters are captured by a closure the less likely it is that it is worth converting. Tuples hurt code clarity and we can deal with some of that for improved performance of users but there is a limit somewhere and more than 3 parameters is where I currently feel readability suffers more than users benefit. This can in change (and may have done since I've been working on perf) with new information like profiling results and discussions.

The changes made to convert to static delegates in this PR were not driven by profile results so they should retain clarity as much as possible while improving performance. I think I would only reduce clarity further if I had a very strong argument that it would improve performance.

Let me know what you think and whether I've gone too far in reducing clarity. There may be cases where I've converted to tuples in the past that it might be better to use closures for clarity.

DavoudEshtehari (approved):

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
```diff
@@ -1078,14 +1079,19 @@ private Task ReadFromRowSourceAsync(CancellationToken cts)
...
-                {
-                    if (t.Status == TaskStatus.RanToCompletion)
+                return _dbDataReaderRowSource.ReadAsync(cts).ContinueWith(
+                    static (Task<bool> task, object state) =>
                     {
-                        _hasMoreRowToCopy = t.Result;
-                    }
-                    return t;
-                }, TaskScheduler.Default).Unwrap();
+                        SqlBulkCopy sqlBulkCopy = (SqlBulkCopy)state;
```
JRahnama, line 1085: why are we creating a new instance of SqlBulkCopy? _hasMoreRowToCopy is the outer variable in this case. Are we trying to prevent  getting a wrong value of _hasMoreRowTocpy changed by other threads?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
```diff
@@ -1969,27 +1975,32 @@ private Task WriteRowSourceToServerAsync(int columnCount, CancellationToken ctok
...
+                    return resultTask.ContinueWith(
+                        static (Task t, object state) =>
                         {
-                            AbortTransaction(); // If there is one, on success transactions will be committed.
-                        }
-                        finally
-                        {
-                            _isBulkCopyingInProgress = false;
-                            if (_parser != null)
+                            SqlBulkCopy sqlBulkCopy = (SqlBulkCopy)state;
```
JRahnama, line 1981: same question here about a new` SqlBulkCopy`.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
```diff
@@ -1969,27 +1975,32 @@ private Task WriteRowSourceToServerAsync(int columnCount, CancellationToken ctok
                 if (resultTask != null)
                 {
                     finishedSynchronously = false;
-                    return resultTask.ContinueWith((t) =>
-                    {
-                        try
+                    return resultTask.ContinueWith(
+                        static (Task t, object state) =>
```
JRahnama, line 1979: NIT: in some places you have chosen `(Task t, object state)` and in some other places `(t, state)`. Can we use one  convention?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlUtil.cs
```diff
@@ -145,7 +145,7 @@ internal static Task CreateContinuationTaskWithState(Task task, object state, Ac
         )
         {
             task.ContinueWith(
-                tsk =>
+                (Task tsk, object state2) =>
```
JRahnama, line 148: NIT: a better naming convention would be helpful here for state2.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
```diff
@@ -1940,15 +1946,15 @@ private Task WriteRowSourceToServerAsync(int columnCount, CancellationToken ctok
...
+                                onSuccess: static (object state) => ((TaskCompletionSource<object>)state).SetResult(null)
                             );
                         }
                     }, ctoken); // We do not need to propagate exception, etc, from reconnect task, we just need to wait for it to finish.
                     return tcs.Task;
                 }
                 else
                 {
-                    AsyncHelper.WaitForCompletion(reconnectTask, BulkCopyTimeout, () => { throw SQL.CR_ReconnectTimeout(); }, rethrowExceptions: false);
+                    AsyncHelper.WaitForCompletion(reconnectTask, BulkCopyTimeout, static () => throw SQL.CR_ReconnectTimeout(), rethrowExceptions: false);
```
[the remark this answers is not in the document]
Wraith2: Yes, that is the purpose. There is also a useful side effect of the way that the compiler implements it because it's a static delegate means that the delegate can be cached. This [sharplab example](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA+ABATARgLABQGAzAATakDCpA3oYaY+WRiqQEIAUAlA03QSZCAggGcAngDswACRgAbAA4woAOgDqAQwCWAFwBi0ShAC2i+TF3aIkzpICu8+QBpSABlcYcSUj1IBeAD5SXQALKAgAd1IAZQBFABlVSgAlAH0UmEhJSSzdABVtExgIe10eV1gwiMiAUQQwGEUrG1EQUgAzTXlRGG4Abj5GAF9CUYJCEnIsUjEpWQVlKEIBISmvH1ZSLT1DKGMzCxbbCGAAKzzSWGzcsF0MJGdtSV0QopKyzxwZzTvrSVcwAgEHkV0s4Si9UazX+om4tHG4yEkxYM3iCRWQ2Y5G8pChTWOVHSmRueUKxVK5XhNAwAHZSLloviYTYeP1EUA=) shows that you get a hidden static class generated and that the action delegate is created the first time it is needed and re-used every time it is called after that. So we get lazy single allocation for free with a nice syntax.

```csharp
    public void B()
    {
       AsyncHelper.WaitForCompletion(null, 0, static () => throw SQL.CR_ReconnectTimeout(), rethrowExceptions: false);
    }
```

is lowered to:

```csharp
    [Serializable]
    [CompilerGenerated]
    private sealed class <>c
    {
        public static readonly <>c <>9 = new <>c();

        public static Action <>9__0_0;

        internal void <B>b__0_0()
        {
            throw SQL.CR_ReconnectTimeout();
        }
    }

    public void B()
    {
        AsyncHelper.WaitForCompletion(null, 0, <>c.<>9__0_0 ?? (<>c.<>9__0_0 = new Action(<>c.<>9.<B>b__0_0)), false);
    }
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
Wraith2: `_hasMoreRowToCopy ` is a field and is implicitly `this._hasMoreRowToCopy` when used in the original. When changed to static there is no `this` reference and the sqlBulkCopy variable is the same object passed in as state.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
Wraith2: I tried to standardize on `(Task t, object state)`, can you point out where I missed?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlUtil.cs
Wraith2: Any suggestions? It's the a state param which is just going to be cast to the correct type and it only needs to be named differently from the one in the parent scope to prevent confusion.

JRahnama:
@Wraith2 Can you trigger a build by making a small change in your PR and make the tests to run once more? I have tried from our side and somethings are missing...

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlUtil.cs
JRahnama: that is fine.

Wraith2:
Fixed up a couple of anonymous typed lambdas.

JRahnama:
LGTM, I am waiting for the Enclave tests to finish and see if  the failure is fixed after I cleaned up the databases.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
JRahnama: Those were not necessarily from your changes, but were included. We can look at them later we have more time I think.

JRahnama (approved):

Wraith2:
I'd like to get https://github.com/dotnet/SqlClient/pull/934 https://github.com/dotnet/SqlClient/pull/925 and https://github.com/dotnet/SqlClient/pull/902 finished to unblock other work if possible.

PR #1068, opened 2021-05-12
Fixes corrupted connection issue when an exception occurs during RPC execution with TVP types

cheenamalhotra:
**Repro**
The test case added here reproduces the issue with current driver.

**Root Cause:**
Looks like call to `FailureCleanup(stateObj, e);` has been missed when exception is handled which contributes to connection corruption for future queries.

This issue also exists in all System.Data.SqlClient drivers.
cc @saurabh500 @corivera

### DIFF
--- a/build.proj
+++ b/build.proj
@@ -38,7 +38,7 @@
     <FunctionalTests Include="**/tools/Microsoft.DotNet.XUnitExtensions/Microsoft.DotNet.XUnitExtensions.csproj" />
     <FunctionalTests Include="**/tools/CoreFx.Private.TestUtilities/CoreFx.Private.TestUtilities.csproj" />
     <FunctionalTests Include="**/ManualTests/SQL/UdtTest/UDTs/Address/Address.csproj" />
-    <FunctionalTests Include="**/Microsoft.Data.SqlClient.Tests.csproj" />
+    <FunctionalTests Include="**/FunctionalTests/Microsoft.Data.SqlClient.Tests.csproj" />
 
     <ManualTests Condition="$(ReferenceType.Contains('NetStandard'))" Include="**/NSLibrary/Microsoft.Data.SqlClient.NSLibrary.csproj" />
     <ManualTests Include="**/ManualTests/SQL/UdtTest/UDTs/Address/Address.csproj" />
--- a/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
+++ b/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
@@ -10441,14 +10441,7 @@ namespace Microsoft.Data.SqlClient
                 }
                 catch (Exception e)
                 {
-                    // UNDONE - should not be catching all exceptions!!!
-                    if (!ADP.IsCatchableExceptionType(e))
-                    {
-                        throw;
-                    }
-
                     FailureCleanup(stateObj, e);
-
                     throw;
                 }
                 FinalizeExecuteRPC(stateObj);
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ParameterTest/TvpTest.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ParameterTest/TvpTest.cs
@@ -17,6 +17,7 @@ using System.Threading.Tasks;
 using System.Transactions;
 using Microsoft.Data.SqlClient.Server;
 using Xunit;
+using System.Linq;
 
 namespace Microsoft.Data.SqlClient.ManualTesting.Tests
 {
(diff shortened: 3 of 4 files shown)


src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ParameterTest/TvpTest.cs
```diff
@@ -82,6 +83,97 @@ public void TestPacketNumberWraparound()
...
+                var parameters = new List<SqlParameter>() { sqlParam };
+                const string SQL = @"SELECT * FROM information_schema.COLUMNS cols INNER JOIN  @ids Ids on Ids.id = cols.ORDINAL_POSITION";
+                using SqlCommand cmd = new(SQL, connection);
+                cmd.CommandTimeout = 100;
+                AddCommandParameters(cmd, parameters);
+                new SqlDataAdapter(cmd).Fill(new("BadFunc"));
+            }
+            catch
+            {
+                // Ignore this exception as it's deliberately introduced.
```
Wraith2, line 126: It might be worth asserting that the expected exception did occur so it can't be broken in some other way in future.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -10441,14 +10441,7 @@ internal Task TdsExecuteSQLBatch(string text, int timeout, SqlNotificationReques
                 }
                 catch (Exception e)
                 {
-                    // UNDONE - should not be catching all exceptions!!!
-                    if (!ADP.IsCatchableExceptionType(e))
```
saurabh500, line 10445: Is this the right thing to do? The code was earlier checking if the exception is not catchable and then throw it instead of doing a FailureCleanup. I suspect this may cause issues in some other scenarios down the line. 
Can't this be made more specific to the use case you are trying to tackle?

saurabh500: Are you looking to handle NullReferenceException here ?

saurabh500: If we are trying to handle NullReferenceException, then we should be looking at fixing the root cause of the NRE.

cheenamalhotra: The exception caught here comes from delegate implemented by user. In this case, the exception in first "bad" case occurs here: https://github.com/dotnet/SqlClient/pull/1068/files#diff-96a138d0ea44725f6c47597239572099a62c531a08ab225ed51e961ff2362505R104

Which means, user can introduce any exception and no matter what we handle we cannot cover all exceptions.

Furthermore, this change is in "TdsExecuteRPC", and if any exception occurs in here the command execution has failed, the StateObj should ideally be cleaned up for next use, since we cache StateObj in the driver and next command will pick it up.

Wraith2: That sounds reasonable to me. Can you put a comment explaining that in the catch block so that no-one is tempted to revert the change? Just explain that user code can bubble exceptions to that point and we can't protect users against themselves in that case or something.

saurabh500: @cheenamalhotra it looks like the user is passing in invalid values, causing the exception. I am trying to understand how the user can introduce *any* exception. 
Can you explain the context a little bit more around why we received an NRE in the first place when user added the null value? Is this really a user error since null can be passed as a valid input for data. Is the NRE due to a response from the server, which we failed to classify or is it because we got a null value which was de-referenced before sending it to the server?

saurabh500: BTW the intention of my questions is to understand the scenario better. My hunch is that your fix may be the only way to go, however if I have to port it to .Net framework (system.data.dll), I am trying to understand the scenario correctly and so that I can defend this. If in this discussion, we realize that a more localized fix can be made for this use case, then that would be ideal. 
For .Net framework, most likely, I will be adding an app context switch where this fix is opt-in since the fix touches something fundamental. I believe, with your fix, for *any* error, we are trying to discard the connection and I wonder if for some large scale applications that would cause pool performance issues.
If we receive more cases for this issue, then I could, in future, consider turning on this fix in .Net framework by default. 
These decisions would be guided by your assessment on whether this is an actual user error or not. So far my understanding is that passing in a "null" may not be a user error.

cheenamalhotra: The action delegate to fetch 'id' and 'value' from 'x' (as pasted above) is user defined, where NRE occurs.

If you debug this test, you'll notice it's surface area.

saurabh500: The part I am struggling to understand is that if there is a NRE in the case above, then the call stack shouldn't come from the driver, which is what the CSS had reported. 

```
BadFunc Exception is Object reference not set to an instance of an object.
   at System.Data.SqlClient.TdsParser.TdsExecuteRPC(SqlCommand cmd, _SqlRPC[] rpcArray, Int32 timeout, Boolean inSchema, SqlNotificationRequest notificationRequest, TdsParserStateObject stateObj, Boolean isCommandProc, Boolean sync, TaskCompletionSource`1 completion, Int32 startRpc, Int32 startParam)
   at System.Data.SqlClient.SqlCommand.RunExecuteReaderTds(CommandBehavior cmdBehavior, RunBehavior runBehavior, Boolean returnStream, Boolean async, Int32 timeout, Task& task, Boolean asyncWrite, Boolean inRetry, SqlDataReader ds, Boolean describeParameterEncryptionRequest)
```

saurabh500: I had a chat with Cheena, and she provided the explanation for the call stack.

saurabh500 (approved):

build.proj
```diff
@@ -38,7 +38,7 @@
     <FunctionalTests Include="**/tools/Microsoft.DotNet.XUnitExtensions/Microsoft.DotNet.XUnitExtensions.csproj" />
     <FunctionalTests Include="**/tools/CoreFx.Private.TestUtilities/CoreFx.Private.TestUtilities.csproj" />
     <FunctionalTests Include="**/ManualTests/SQL/UdtTest/UDTs/Address/Address.csproj" />
-    <FunctionalTests Include="**/Microsoft.Data.SqlClient.Tests.csproj" />
+    <FunctionalTests Include="**/FunctionalTests/Microsoft.Data.SqlClient.Tests.csproj" />
```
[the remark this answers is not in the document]
cheenamalhotra: I had few issues running tests locally with current setting. Fixing this way makes things work.

JRahnama (approved):

PR #1070, opened 2021-05-12
Vulnerability fix | System.Text.Encodings.Web Version override

cheenamalhotra:
Change assessed by Component detection tool: [Build #31825](https://dev.azure.com/sqlclientdrivers-ci/sqlclient/_build/results?buildId=31825&view=results)

### DIFF
--- a/src/Microsoft.Data.SqlClient/netfx/src/Microsoft.Data.SqlClient.csproj
+++ b/src/Microsoft.Data.SqlClient/netfx/src/Microsoft.Data.SqlClient.csproj
@@ -581,6 +581,7 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="$(MicrosoftSourceLinkGitHubVersion)" PrivateAssets="All" />
+    <PackageReference Include="System.Text.Encodings.Web" Version="$(SystemTextEncodingsWebVersion)" />
     <PackageReference Include="System.Security.Cryptography.Algorithms">
       <Version>$(SystemSecurityCryptographyAlgorithmsVersion)</Version>
     </PackageReference>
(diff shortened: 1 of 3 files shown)


johnnypham (approved):
Does it not need to be added to the nuspec?

src/Microsoft.Data.SqlClient/netfx/src/Microsoft.Data.SqlClient.csproj
```diff
@@ -581,6 +581,9 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="$(MicrosoftSourceLinkGitHubVersion)" PrivateAssets="All" />
+    <PackageReference Include="System.Text.Encodings.Web">
+      <Version>$(SystemTextEncodingsWebVersion)</Version>
+    </PackageReference>
```
DavoudEshtehari, line 586: NIT: Just prefering single line reference.
```suggestion
    <PackageReference Include="System.Text.Encodings.Web" Version="$(SystemTextEncodingsWebVersion)" />
```

DavoudEshtehari (approved):

cheenamalhotra:
> Does it not need to be added to the nuspec?

No, since the vulnerability arises from underlying library, we don't want to force override dependencies for all customers. We override this type of dependency for our product build only.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft.Data.SqlClient.csproj
cheenamalhotra: I tried it before, this does not work in NetFx project.

JRahnama (approved):

PR #1078, opened 2021-05-19
Set ColumnEncryptionKeyCacheTtl to zero when key store provider is registered and used globally

johnnypham:
When a provider is registered globally, its `ColumnEncryptionKeyCacheTtl` property should be set to zero. Before decrypting with a global provider, the property will again be set to zero. Any CEK caching implementation should check this value before caching CEKs and not cache if the value is zero. A globally registered provider should only use the global CEK cache in the driver.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnection.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnection.cs
@@ -98,8 +98,6 @@ namespace Microsoft.Data.SqlClient
         /// </summary>
         private static IReadOnlyDictionary<string, SqlColumnEncryptionKeyStoreProvider> s_globalCustomColumnEncryptionKeyStoreProviders;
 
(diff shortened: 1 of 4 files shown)


cheenamalhotra (approved):

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnection.cs
```diff
@@ -314,9 +314,10 @@ public static void RegisterColumnEncryptionKeyStoreProviders(IDictionary<string,
                     throw SQL.CanOnlyCallOnce();
                 }
 
-                if (customProviders.ContainsKey(s_akvProviderName))
+                // to prevent conflicts between CEK caches, global providers should not use their own CEK caches
+                foreach (SqlColumnEncryptionKeyStoreProvider provider in customProviders.Values)
                 {
-                    customProviders[s_akvProviderName].ColumnEncryptionKeyCacheTtl = new TimeSpan(0);
```
cheenamalhotra, line 319: Guess `s_akvProviderName` is no longer needed.

PR #1088, opened 2021-05-28
Tests | Add Local DB App Name config property

cheenamalhotra:
Allows configuring Local Db App name from config.json for testing.


src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/LocalDBTest/LocalDBTest.cs
```diff
@@ -14,7 +14,7 @@ public static class LocalDBTest
         [ConditionalFact(nameof(IsLocalDBEnvironmentSet))]
```
DavoudEshtehari, line 14: Could you remove `SupportsLocalDb` property from the config file and check if `DataTestUtility.LocalDbAppName` has value instead of that?

DavoudEshtehari (approved):

PR #1101, opened 2021-06-04
Give system key store providers precedence over instance-level providers

johnnypham:
System providers will be checked before command and connection providers.

### DIFF
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/ApiShould.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/ApiShould.cs
@@ -2242,6 +2242,31 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests.AlwaysEncrypted
             }
         }
 
+        // On Windows, "_fixture" will be type SQLSetupStrategyCertStoreProvider
+        // On non-Windows, "_fixture" will be type SQLSetupStrategyAzureKeyVault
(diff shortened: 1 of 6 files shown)


src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/ApiShould.cs
```diff
@@ -2242,6 +2242,31 @@ public void TestCommandCustomKeyStoreProviderDuringAeQuery(string connectionStri
             }
         }
 
+        // On Windows, "_fixture" will be type SQLSetupStrategyCertStoreProvider
+        // On non-Windows, "_fixture" will be type SQLSetupStrategyAzureKeyVault
+        // Test will pass on both but only SQLSetupStrategyCertStoreProvider is a system provider
+        [ConditionalTheory(typeof(DataTestUtility), nameof(DataTestUtility.AreConnStringSetupForAE))]
+        [ClassData(typeof(AEConnectionStringProvider))]
+        public void TestSystemProvidersHavePrecedenceOverInstanceLevelProviders(string connectionString)
```
cheenamalhotra, line 2250: Can we have 1 more test for Connection instance registration similarly?

Wraith2:
Why? this will mean that there is no way to override a system provider, is that the intention?

cheenamalhotra:
System providers are non-overridable by design, even with current static registration API.

Wraith2:
Ok.

cheenamalhotra (approved):

DavoudEshtehari (approved):

JRahnama (approved):

Issue #1105, opened 2021-06-08
Using SqlDataAdapter.Fill on parameter-less stored procedure command with always encrypted w. secure encalves gives NullReferenceException


pigiax:
## Description
If you are using a SQL server connection with Always Encrypted with secure enclaves, if you try to use SqlDataAdapter.Fill to fill a DataSet, using an SqlCommand which is a stored procedure without parameters it goes in error with NullReferenceException on BuildStoredProcedureStatementForColumnEncryption

## To reproduce
Tested with "Microsoft.Data.SqlClient" Version "2.1.3" and "Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider" Version "2.0.0"

```
private static void SqlError()
{
	using (var dbConnection = new SqlConnection("Data Source=[server];Initial Catalog=[db];Column Encryption Setting=Enabled;Attestation Protocol=AAS;Enclave Attestation Url=[secureEnclaveUrl];User ID=[user];Password=[pwd]"))
	{
		dbConnection.Open();
		DataSet ds = new DataSet("Results");

		IDbCommand dbCommand = dbConnection.CreateCommand();
		dbCommand.CommandText = "[YourStoredProcedure]";
		dbCommand.CommandType = CommandType.StoredProcedure;
		
		if(dbCommand.Parameters != null)
		{
			//This is a workarond which triggers the parameters lazy creation
			//https://github.com/dotnet/SqlClient/blob/3df7de613aecee3a8b5229333c84142e6e83ae9d/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlCommand.cs#L889
		}
		using (var da = new SqlDataAdapter(dbCommand as SqlCommand))
		{
			//This works
			da.Fill(ds);
		}

		dbCommand = dbConnection.CreateCommand();
		dbCommand.CommandText = "[YourStoredProcedure]";
		dbCommand.CommandType = CommandType.StoredProcedure;

		using (var da = new SqlDataAdapter(dbCommand as SqlCommand))
		{
			//NullReferenceException on BuildStoredProcedureStatementForColumnEncryption
			//https://github.com/dotnet/SqlClient/blob/3df7de613aecee3a8b5229333c84142e6e83ae9d/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs#L5839
			da.Fill(ds);
		}
	}
}
```

## Expected behavior
[...]

PR #1115, opened 2021-06-15
Do not add parameters for sp_describe_parameter_encryption when not provided

johnnypham:
Fixes [#1105 ](https://github.com/dotnet/SqlClient/issues/1105)

The exception occurs when the driver tries to execute `sp_describe_parameter_encryption` but no user parameters have been provided.

Current behaviour for always encrypted with secure enclaves:

When executing a stored procedure that does not require an input parameter:

- **unexpected: fails with NRE. Driver attempts to add parameters to execute `sp_describe_parameter_encryption` but parameters are null. Should succeed by sending empty parameter list for `sp_describe_parameter_encryption`**

When executing a stored procedure that requires an input parameter but none are provided by the user:

- **unexpected: fails with NRE. Driver attempts to add parameters to execute `sp_describe_parameter_encryption` but parameters are null. Should fail with "`Microsoft.Data.SqlClient.SqlException (0x80131904): Procedure or function '[name of sp]' expects parameter '@[name of param]', which was not supplied.`"**

This change fixes both unexpected behaviours above.


cheenamalhotra (approved):

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/ApiShould.cs
```diff
@@ -878,6 +878,53 @@ public void TestExecuteReaderWithCommandBehavior(string connection, CommandBehav
...
+                sqlCommand.CommandText = $"CREATE PROCEDURE {procWithoutParams} AS SELECT * FROM [{_tableName}];";
+                sqlCommand.ExecuteNonQuery();
+                sqlCommand.CommandText = $"CREATE PROCEDURE {procWithParam} @id INT AS SELECT * FROM [{_tableName}] WHERE CustomerId = @id";
+                sqlCommand.ExecuteNonQuery();
+
+                sqlCommand.CommandText = procWithoutParams;
+                sqlCommand.CommandType = CommandType.StoredProcedure;
+                using (SqlDataReader reader = sqlCommand.ExecuteReader())
+                {
+                    Assert.Equal(3, reader.VisibleFieldCount);
```
DavoudEshtehari, line 906: To avoid future side effects, I suggest you resolve the number of fields from the table or mention the exact fields in the `Select` statement instead of `*`. Since this value has multiple usages, it'd better assign it to a field first.

DavoudEshtehari (approved):

Issue #1121, opened 2021-06-16
Null Reference in GetServerTransactionLevel 


ShaneCourtrille:
I'm investigating an exception occurring when our system is under load and wanted to know if anyone has any idea what could cause the following?  From reading the code it looks like the _innerConnection.Owner must be null but not sure what would cause that?

```
System.NullReferenceException:
   at Microsoft.Data.SqlClient.SqlInternalTransaction.GetServerTransactionLevel (Microsoft.Data.SqlClient, Version=2.0.20168.4, Culture=neutral, PublicKeyToken=23ec7fc2d6eaa4a5)
   at Microsoft.Data.SqlClient.SqlInternalTransaction.CheckTransactionLevelAndZombie (Microsoft.Data.SqlClient, Version=2.0.20168.4, Culture=neutral, PublicKeyToken=23ec7fc2d6eaa4a5)
   at Microsoft.Data.SqlClient.SqlInternalTransaction.Commit (Microsoft.Data.SqlClient, Version=2.0.20168.4, Culture=neutral, PublicKeyToken=23ec7fc2d6eaa4a5)
   at Microsoft.Data.SqlClient.SqlTransaction.Commit (Microsoft.Data.SqlClient, Version=2.0.20168.4, Culture=neutral, PublicKeyToken=23ec7fc2d6eaa4a5)
   at YesSql.Session.CommitTransaction (YesSql.Core, Version=0.0.0.0, Culture=neutral, PublicKeyToken=51cdb8c1938e4789)
   at YesSql.Session+<CommitAsync>d__39.MoveNext (YesSql.Core, Version=0.0.0.0, Culture=neutral, PublicKeyToken=51cdb8c1938e4789)
```

sebastienros:
`_innerConnection.Owner` point to this code: 
https://referencesource.microsoft.com/#System.Data/fx/src/data/System/Data/ProviderBase/DbConnectionInternal.cs,294

Which is a `WeakReference` so I believe the getter should create a local copy of `Target` and check for nullability.

https://referencesource.microsoft.com/#System.Data/fx/src/data/System/Data/ProviderBase/DbConnectionInternal.cs,39

sebastienros:
Could also show a race in the pooling logic since reports of this issue seem to relate to high load.

Wraith2:
use of `WeakReference` should be converted to `WeakReference<T>` if possible because the generic version has a usage pattern designed to avoid this sort of problem with time of check time of use dsparity.

[...]

PR #1122, opened 2021-06-16
Change WeakReference to WeakReference<T>

Wraith2:
Reading https://github.com/dotnet/SqlClient/issues/1121 I remembered that we're using `WeakReference` and not `WeakReference<T>`, the generic version has a different api shape using `bool isAlive = TryGetTarget(out T target)`  to prevent the possibility of a GC between `IsAlive` and `Target` property accesses. Note that `TryGetTarget` returns false for null which is why some null checks are no longer needed.


johnnypham (approved):

cheenamalhotra (approved):

Dzoge:
Is this issue also fixed in System.Data.SqlClient package?

Wraith2:
Unlikely. The `System.Data.SqlClient` package is deprecated in favour of this library and the only things that get back ported are security and high impact stability changes. This is simply a modernization.

PR #1128, opened 2021-06-18
Reset cached async state on disposing a command

DavoudEshtehari:
**The root cause of issue #971**: The cached status of an async operation that is kept in the related connection didn't reset to be ready for the next operation.

- Porting more handled exceptions from **Net Fx**.
- Including more traces.


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
```diff
@@ -2404,28 +2440,41 @@ long firstAttemptStart
...
                 cachedAsyncState.SetActiveConnectionAndResult(completion, nameof(EndExecuteReader), _activeConnection);
                 _stateObj.ReadSni(completion);
             }
+            catch (System.OutOfMemoryException e)
+            {
+                _activeConnection.Abort(e);
+                completion.TrySetException(e);
+                throw;
+            }
+            catch (System.StackOverflowException e)
```
Wraith2, line 2457: Both StackOverflow exception and OOM leave the runtime in a possibly unstable state and trying to call further functions in the catch of either may fail because of the condition that caused the first exception. I don't fault the completeness but I do doubt the effectiveness.

Wraith2:
Looks reasonable. Is it worth checking that the cached async state is correctly reset when the connection is returned to the pool?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
DavoudEshtehari: Interesting, Do you mean they haven't been copied by purpose and we don't need them? Don't you think it's worth keeping them for a partially stable situation inside an unstable runtime state? Or runtime maybe improves itself to have a more stable condition in these situations in the future?

Wraith2: The runtime will never be able to leave you in a consistent state with those exceptions, or ThreadAbortException. I'm told that those exceptions aren't even catchable in core/net5+. I'd advise taking them out, no matter what happens if those exceptions occur there is no way to be safe.

DavoudEshtehari: Ok, I'll remove them because we don't want to include the useless codes. But still, a question remains in my mind that would be appreciable if you had an answer for it: What is the aim of those types if we're not able to do anything after catching them!

David-Engel: I've always assumed that code is simply a best-effort to at least **_try_** and clean things up a bit on those errors. As Wraith said, there are no guarantees, but I'm okay with trying...

Wraith2: Ok. There is likely to be no harm in keeping them, as I said I checked with one of the core devs who told me they aren't catchable so they're not harmful. 
Perhaps just a comment that we don't expect them to be hit or that we know the may not work on core so if someone sees them they don't remove them in future assuming we didn't know?

johnnypham (approved):

JRahnama (approved):

Issue #1118, opened 2021-06-16
Issue with System.Data.SqlClient.SqlException differences between Win/Mac Fx/Core


odhanson:
Hi

We are facing an issue with difference in behavior between our Windows (based on FX) and Mac (based on .NET 5) implementation for connecting to SQL server. We are trying to make an encrypted connection to a server that doesn't support it.

In both environments we will receive a SqlException, but in windows, we receive -2146893019 as the Number of the error, and in Mac it is 0. We use this number to identify the specific exception (that the user is trying to connect to a non-secure server), but on Mac this is broken.

Is there a bug here in the .NET core implementation ? Is there an alternative way to get this done ? 

Some more details. 
In windows, we receive the following error message:
"{"A connection was successfully established with the server, but then an error occurred during the login process. (provider: SSL Provider, error: 0 - The certificate chain was issued by an authority that is not trusted.)"}"
ErrorCode: -2146232060
Source: ".Net SqlClient Data Provider"
Number: -2146893019 
InnerException: {"The certificate chain was issued by an authority that is not trusted"} (System.Exception {System.ComponentModel.Win32Exception})

On Mac, we receive:
"A connection was successfully established with the server, but then an error occurred during the pre-login handshake. (provider: TCP Provider, error: 35 - An internal exception was caught)"
ErrorCode: -2146232060
Number: 0
InnerException: {System.Security.Authentication.AuthenticationException: The remote certificate was rejected by the provided RemoteCertificateValidationCallback. at System.Net.Security.SslStream.SendAuthResetSignal(ProtocolToken message, ExceptionDispatchInfo exception) at System.Net.Security.SslStream.ForceAuthenticationAsync[TIOAdapter](TIOAdapter adapter, Boolean receiveFirst, Byte[] reAuthenticationData, Boolean isApm) at System.Net.Security.SslStream.ProcessAuthentication(Boolean isAsync, Boolean isApm, CancellationToken cancellationToken) at System.Net.Security.SslStream.AuthenticateAsClient(SslClientAuthenticationOptions sslClientAuthenticationOptions) at System.Net.Security.SslStream.AuthenticateAsClient(String targetHost, X509CertificateCollection clientCertificates, SslProtocols enabledSslProtocols, Boolean checkCertificateRevocation) at System.Net.Security.SslStream.AuthenticateAsClient(String targetHost) at System.Data.SqlClient.SNI.SNITCPHandle.EnableSsl(UInt32 options)}
[...]

PR #1130, opened 2021-06-21
Update error code when certificate validation fails in managed SNI

johnnypham:
Fixes [#1118](https://github.com/dotnet/SqlClient/issues/1118)

Managed SNI errors are given a `nativeError` of zero, which is used for `SqlException.Number`. `nativeError` should be consistent with the same error produced in sni.dll.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIError.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIError.cs
@@ -24,24 +24,24 @@ namespace Microsoft.Data.SqlClient.SNI
 
         public SNIError(SNIProviders provider, uint nativeError, uint sniErrorCode, string errorMessage)
         {
-            this.lineNumber = 0;
-            this.function = string.Empty;
+            lineNumber = 0;
+            function = string.Empty;
             this.provider = provider;
             this.nativeError = nativeError;
-            this.sniError = sniErrorCode;
+            sniError = sniErrorCode;
             this.errorMessage = errorMessage;
-            this.exception = null;
+            exception = null;
         }
 
         public SNIError(SNIProviders provider, uint sniErrorCode, Exception sniException, uint nativeErrorCode = 0)
         {
-            this.lineNumber = 0;
-            this.function = string.Empty;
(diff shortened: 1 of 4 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIError.cs
```diff
@@ -30,12 +33,12 @@ public SNIError(SNIProviders provider, uint nativeError, uint sniErrorCode, stri
...
         }
 
-        public SNIError(SNIProviders provider, uint sniErrorCode, Exception sniException)
+        public SNIError(SNIProviders provider, uint sniErrorCode, Exception sniException, uint nativeErrorCode = 0)
         {
             this.lineNumber = 0;
             this.function = string.Empty;
             this.provider = provider;
-            this.nativeError = 0;
+            this.nativeError = nativeErrorCode;
```
JRahnama, line 41: NIT: `this` keyword is not necessary as the variable name is not same as the fieldname. In fact the only part that needs `this` keyword is `provider` which field's name and argument's name is exactly the same.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
```diff
@@ -434,7 +437,7 @@ private static Socket Connect(string serverName, int port, TimeSpan timeout, boo
 
                     // If we have already got a valid Socket, or the platform default was prefered
                     // we won't do the second traversal.
-                    if (availableSocket != null || ipPreference == SqlConnectionIPAddressPreference.UsePlatformDefault) 
+                    if (availableSocket != null || ipPreference == SqlConnectionIPAddressPreference.UsePlatformDefault)
```
JRahnama, line 440: NIT: 
```suggestion
                    if (availableSocket is not null || ipPreference == SqlConnectionIPAddressPreference.UsePlatformDefault)
```

JRahnama:
LGTM. Couple of minor suggestions.

JRahnama (approved):

cheenamalhotra (approved):

Issue #1126, opened 2021-06-18
.NET Core SqlConnection ConnectTimout 15 less than not work?


OdaShinsuke:
## .NET Core SqlConnection ConnectTimout 15 less than not work?

Microsoft.Data.SqlConnection ConnectTimeout behaves differently between .NET Framework and .NET Core.

Environment:
* Windows 10
* SQL Server 2019 Developer Edition (service is stopped)
* .NET Core 5 / 3.1、.NET Framework 4.8
* Microsoft.Data.SqlClient 3.0.0

## To reproduce

**Prerequisites**
Stop the SQL Server service.

**.NET Core 5 or 3.1**
```cs
using Microsoft.Data.SqlClient;
using System;
using System.Diagnostics;

namespace ConnectTimeoutNotWorkCore
{
    class Program
    {
        static void Main(string[] args)
        {
            Console.WriteLine(typeof(Program).FullName);
            OpenError(1);
            OpenError(60);
            Console.ReadKey();
            /*
ConnectTimeoutNotWorkCore.Program
Start
Error Number:10061
16672
Done
Start
Error Number:10061
65981
Done
             */
        }
        static void OpenError(int connectTimeout)
        {
            var builder = new SqlConnectionStringBuilder()
            {
                DataSource = "localhost\\MSSQL2019",
                InitialCatalog = "master",
                IntegratedSecurity = true,
                ConnectTimeout = connectTimeout
            };
            var stopwatch = new Stopwatch();
            try
            {
                using (var conn = new SqlConnection(builder.ConnectionString))
                {
                    Console.WriteLine("Start");
                    stopwatch.Start();
                    conn.Open();
                }
            }
            catch (SqlException e)
            {
                Console.WriteLine($"Error Number:{e.Number}");
            }
            finally
            {
                stopwatch.Stop();
                Console.WriteLine(stopwatch.ElapsedMilliseconds);
                Console.WriteLine("Done");
            }
        }
    }
}
```
**.NET Framework 4.8**
```cs
using Microsoft.Data.SqlClient;
using System;
using System.Diagnostics;

namespace ConnectTimeoutNotWorkFx48
{
    class Program
    {
        static void Main(string[] args)
        {
            Console.WriteLine(typeof(Program).FullName);
            OpenError(1);
            OpenError(60);
            Console.ReadKey();
            /*
ConnectTimeoutNotWorkFx48.Program
Start
Error Number:258
1768
Done
Start
Error Number:258
60019
Done
             */
[...]

PR #1131, opened 2021-06-21
WIP | Fixing issue with ignoreSniOpenTimeout in NetCore

JRahnama:
While working on #1126  I noticed in NetFx:

SqlInternalConnectionTds at line 1762 and 1766
```
Boolean disableTnir = ShouldDisableTnir(connectionOptions); 
Boolean isParallel = connectionOptions.MultiSubnetFailover || (connectionOptions.TransparentNetworkIPResolution && !disableTnir);
```

Basically, `isParallel` is mostly coming true if we do not disable `TransparentNetworkIPResolution` default value.
further in the code, we attemp  `AttemptOneLogin` we pass the value for bool` ignoreSniOpenTimeout` as `!isParallel` comes to false and in result when we get to `TdsParserSafeHandles` we ignore the part for `timeout = Timeout.Infinite;`

However, in Net Core since there is no `TransparentNetworkIPResolution` we only check for `connectionOptions.MultiSubnetFailover  `and by sending `!connectionOptions.MultiSubnetFailover ` we set the timeout to `Timeout.Infinite`. This will cause the expected exception  not to happen even if the connection timeout is set to some values lower than 15 seconds.

I am looking for tests to add as this needs to stop SqlServices as the repro suggests.

Changes in NetFx are only some simplifications as I was looking into them.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
@@ -1521,7 +1521,7 @@ namespace Microsoft.Data.SqlClient
                     AttemptOneLogin(serverInfo,
                                     newPassword,
                                     newSecurePassword,
-                                    !connectionOptions.MultiSubnetFailover,    // ignore timeout for SniOpen call unless MSF
+                                    connectionOptions.MultiSubnetFailover,
                                     connectionOptions.MultiSubnetFailover ? intervalTimer : timeout);
 
                     if (connectionOptions.MultiSubnetFailover && null != ServerProvidedFailOverPartner)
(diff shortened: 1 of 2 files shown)


Wraith2:
There are some differences between netcore and netfx like this where I don't think the code differences are helpful. In this situation I think the code for netfx was correct and if it were used on netcore where MultiSubnetFailover can never be set it would have given the correct behaviour, is that right do you think? If so is it worth updating the netcore code to be the same as the netfx code?

JRahnama:
> There are some differences between netcore and netfx like this where I don't think the code differences are helpful.

I agree. Existing of `TransparentNetworkIPResolution` makes the nature of `SqlInternalConnectionTds` in netfx different than the one in netcore. Mostly all f its objects are considered as parallel where in netcore it is not the case. there are several cases in netfx in the very same class that it checks for `if (isParallel)` which in netfx we enter the if statement but in netcore we do not.  In my opinion we do not need that variable, `ignoreSniOpenTimeout` , at all. In netfx if you look a bit upper that in the code around line 1839 there is a comment which says:

>   //  for non-failover, non-MSF scenarios to avoid breaking changes as part of a QFE.  Consider fixing timeout
                    //  handling in next full release and removing ignoreSniOpenTimeout parameter.

Wraith2:
> I am strongly against making netcore as same as netfx, without deep investigations

Agreed. If it is done it must be carefully and well tested. Having a single shared implementation may be better for maintenance. Something to consider in the future and not this PR.

cheenamalhotra:
I think we need to understand MultiSubnetFailover cases with failover cluster setup before confirming this is the right change here. I'm also curious to know why they added "ignoreSniOpenTimeout" at the first place that's being applied by default! I agree TNIR broke the deal in NetFx but "ignoring lower timeout" has always been a problem case also in NetFx whenever it comes to action. So before we change it to reverse, we need to what adverse affects it could have when someone connects with **MultiSubnetFailover** enabled. It will then cause timeout issues there.

Issue #1110, opened 2021-06-13
SQLCommand Error on Visual Studio Designer (WinForms)


stavrinc:
On a C# project (.NET Framework 4.8) when I try to create a control SqlCommand using the designer I get the following error

**Failed to create component 'SqlCommand'. The error message follows: 'System.invalidCastException: Unable to cast object of type 'Microsoft.Data.SqlClient.SqlCommand' to type 'System.Data.SqlClient.SqlCommand'.**

The same problem appears with Microsoft.Data.SqlClient V3 and V2.1.3 and older.

All other controls like SQLDataAdapter are ok.
![SQLCOMMAND_ERROR](https://user-images.githubusercontent.com/85819270/121797582-c1147180-cc29-11eb-8110-e0aa22f9701e.jpg)

Thanks in advance for any suggestion

johnnypham:
Hi @stavrinc, can you provide a repro app or more details? Are you trying to add a custom user control? How is `SqlCommand` being used in the control?

Wraith2:
I think the problem is that the designer component in visual studio is expecting to get a System.Data.SqlClient instance to design over but the instance is a Microsoft.Data.SqlClient ovject so the cast fails. You can see in the stack trace that it's trying to do design time stuff and then hits Microsoft.VSDesigner.Data.VS.SqlCommandDesigner and falls over, so i think visual studio would need to be updates to recognise this library instead of only the system version.

johnnypham:
Thanks @Wraith2, it's out of our control then. @stavrinc if you can provide more details, maybe we can come up with a workaround.

stavrinc:
> Thanks @Wraith2, it's out of our control then. @stavrinc if you can provide more details, maybe we can come up with a workaround.

What details exactly. It is very simple bug. When I create the SQLCommand button on a form I get the above error.

Wraith2:
The easiest thing to work with is a really really simple repro. Can you provide a sample project that shows the bug, so simple that you can just give someone the zip and tell them to open the designer, click a button and have it crash.

johnnypham:
> What details exactly. It is very simple bug. When I create the SQLCommand button on a form I get the above error.

I don't see a SqlCommand button so I'm assuming it's a custom user control. I'm not seeing the exception when I use the SqlClient library in a custom user control. Like Wraith2 said, please provide a basic application that others can run, otherwise it's difficult to figure out what's wrong.

Wraith2:
[...]

PR #1132, opened 2021-06-22
Remove designer attribute from SqlCommand and SqlDataAdapter

johnnypham:
Fixes [#1110](https://github.com/dotnet/SqlClient/issues/1110)

### DIFF
--- a/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlCommand.cs
+++ b/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlCommand.cs
@@ -32,8 +32,8 @@ namespace Microsoft.Data.SqlClient
     [
     DefaultEvent("RecordsAffected"),
     ToolboxItem(true),
-    Designer("Microsoft.VSDesigner.Data.VS.SqlCommandDesigner, " + AssemblyRef.MicrosoftVSDesigner),
     DesignerCategory("")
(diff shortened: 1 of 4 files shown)


cheenamalhotra:
For some reason they are also in netcore ref class, might as well clean that up.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlCommand.cs
```diff
@@ -32,8 +32,8 @@ namespace Microsoft.Data.SqlClient
     [
     DefaultEvent("RecordsAffected"),
     ToolboxItem(true),
-    Designer("Microsoft.VSDesigner.Data.VS.SqlCommandDesigner, " + AssemblyRef.MicrosoftVSDesigner),
     DesignerCategory("")
+    // TODO: Add designer attribute when Microsoft.VSDesigner.Data.VS.SqlCommandDesigner uses Microsoft.Data.SqlClient
```
cheenamalhotra, line 36: Can we add TODOs in netfx/ref too?

cheenamalhotra (approved):

Wraith2:
Is there a VS issue opened to track adding this library to the types VS knows about?

cheenamalhotra:
There isn't any, I was thinking the same. I might get some conversations started internally, but it doesn't seem they would be replacing existing classes anyways. Most likely new set of classes would be available in future VS - then we'll include those here.

DavoudEshtehari (approved):
Could you add TODO comments in `netcore/ref` file too?

PR #1133, opened 2021-06-23
cleanup SqlDataRecord

Wraith2:
Follow-up from https://github.com/dotnet/SqlClient/pull/1024 , This implements the suggested and agreed changes from that PR feedback.

1) remove `EnsureSubclassOverride` because there is no way to create an SqlDataRecord without the internal buffer that it is checking being set.
2) Ensure that all callsites with an ordinal parameter directly or indirectly (through `GetMetaData` or `GetSmiMetaData`) call `ThrowIfInvalidOrdinal`
3) remove the unused internal ctor
4) mark most fields as readonly

### DIFF
--- a/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/Server/SqlDataRecord.cs
+++ b/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/Server/SqlDataRecord.cs
@@ -14,13 +14,13 @@ namespace Microsoft.Data.SqlClient.Server
     /// <include file='../../../../../../../../doc/snippets/Microsoft.Data.SqlClient.Server/SqlDataRecord.xml' path='docs/members[@name="SqlDataRecord"]/SqlDataRecord/*' />
     public class SqlDataRecord : IDataRecord
     {
-        private SmiRecordBuffer _recordBuffer;
-        private SmiContext _recordContext;
-        private SmiExtendedMetaData[] _columnSmiMetaData;
-        private SmiEventSink_Default _eventSink;
-        private SqlMetaData[] _columnMetaData;
+        private readonly SmiRecordBuffer _recordBuffer;
+        private readonly SmiContext _recordContext;
+        private readonly SmiExtendedMetaData[] _columnSmiMetaData;
+        private readonly SmiEventSink_Default _eventSink;
+        private readonly SqlMetaData[] _columnMetaData;
+        private readonly bool _usesStringStorageForXml;
         private FieldNameLookup _fieldNameLookup;
-        private bool _usesStringStorageForXml;
 
         static readonly SmiMetaData s_maxNVarCharForXml = new SmiMetaData(
             SqlDbType.NVarChar,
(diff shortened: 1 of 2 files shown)


johnnypham (approved):

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/Server/SqlDataRecord.cs
```diff
@@ -283,22 +196,20 @@ public virtual bool IsDBNull(int ordinal)
         //  ISqlRecord implementation
         public virtual SqlMetaData GetSqlMetaData(int ordinal)
         {
-            EnsureSubclassOverride();
+            ThrowIfInvalidOrdinal(ordinal);
```
cheenamalhotra, line 199: Could we call this on all APIs accepting ordinal? As I see the internal ValueUtilSmi does not do any significant checks.

Wraith2: As I mentioned in the original description there are either direct or indirect calls to `ThrowIfInvalidOrdinal` on every entry point now because `GetMetaData` and `GetSmiMetaData` call `ThrowIfInvalidOrdinal`, so we don't need a direct call each time it'll just cause performance noise, we'll get the exception on any ordinal (unless you see one I missed).

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/Server/SqlDataRecord.cs
```diff
@@ -326,122 +236,57 @@ public virtual int GetSqlValues(object[] values)
...
-        }
+        public virtual SqlSingle GetSqlSingle(int ordinal) => ValueUtilsSmi.GetSqlSingle(_eventSink, _recordBuffer, ordinal, GetSmiMetaData(ordinal));
 
         /// <include file='../../../../../../../../doc/snippets/Microsoft.Data.SqlClient.Server/SqlDataRecord.xml' path='docs/members[@name="SqlDataRecord"]/GetSqlDouble/*' />
-        public virtual SqlDouble GetSqlDouble(int ordinal)
-        {
-            EnsureSubclassOverride();
-            return ValueUtilsSmi.GetSqlDouble(_eventSink, _recordBuffer, ordinal, GetSmiMetaData(ordinal));
-        }
+        public virtual SqlDouble GetSqlDouble(int ordinal) => ValueUtilsSmi.GetSqlDouble(_eventSink, _recordBuffer, ordinal, GetSmiMetaData(ordinal));
```
cheenamalhotra, line 269: As in here?
Neither of `GetSmiMetaData()` or `GetSqlDouble()` validate ordinal to throw `IndexOutOfRange` exception.

cheenamalhotra: Sorry I noticed it's now in new changes, my bad 🙈

cheenamalhotra (approved):

TrayanZapryanov:
@Wraith2 I've looked at netcore SqlDataRecord class and found several interesting things:
1) Several things looks unused 
image
2) The code in netcore SmiEventSink_Default class is quite strange - at least for me :) . I can't understand who will fill _errors and _warnings fields. If they are not used , maybe it is worth creating a static dummy class which does nothing in **ProcessMessagesAndThrow**.
3) _usesStringStorageForXml seems to have a constant "true" value

Wraith2:
1) Yes, to the compiler in netcore builds. They might be used directly int he netfx build or used through reflection or inheritance. My current focus is on merging as much of the netcore and netfx codebases as possible and there are many peculiar or nonsensical things that will need resolving once that first task is finished.
2) Yes, as far as I can tell it's never used and when using SqlDataRecord the message sink functions turn up in profiles. Again, with deeper investigation on both codebases they may be possible to dummy out as you said, I just haven't tried to go deep yet. If you want to do so it would be welcome.
3) check if it's ever used in netfx, if not and there is no path to set it through inheritance then we could get rid of it.

PR #1135, opened 2021-06-23
Add sample on how to register and reuse SqlColumnEncryptionAzureKeyVaultProvider across multiple SqlCommand objects

johnnypham:


doc/samples/RegisterCustomKeyStoreProvider_Example.cs
```diff
@@ -0,0 +1,47 @@
+﻿// <Snippet1>
+class Program
```
cheenamalhotra, line 2: Please complete this class with using imports so it can be compiled as-is.

cheenamalhotra (approved):

JRahnama (approved):

PR #1141, opened 2021-06-27
Replace WeakReference with WeakReference<T>

Wraith2:
Follow-up from https://github.com/dotnet/SqlClient/pull/1122 , I checked for more uses of `WeakReference` and replaced them with `WeakReference<T>`.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalTransaction.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalTransaction.cs
@@ -38,7 +38,7 @@ namespace Microsoft.Data.SqlClient
         private int _openResultCount;           // passed in the MARS headers
         private SqlInternalConnection _innerConnection;
         private bool _disposing;                 // used to prevent us from throwing exceptions while we're disposing
-        private WeakReference _parent;                    // weak ref to the outer transaction object; needs to be weak to allow GC to occur.
+        private WeakReference<SqlTransaction> _parent;                    // weak ref to the outer transaction object; needs to be weak to allow GC to occur.
 
         private static int _objectTypeCount; // EventSource counter
         internal readonly int _objectID = Interlocked.Increment(ref _objectTypeCount);
@@ -58,7 +58,7 @@ namespace Microsoft.Data.SqlClient
 
             if (null != outerTransaction)
             {
-                _parent = new WeakReference(outerTransaction);
+                _parent = new WeakReference<SqlTransaction>(outerTransaction);
             }
 
             _transactionId = transactionId;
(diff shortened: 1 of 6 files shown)


johnnypham (approved):

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/sqlinternaltransaction.cs
```diff
@@ -591,20 +590,16 @@ internal void Zombie()
 
         private void ZombieParent()
         {
-            if (null != _parent)
+            if (_parent != null && _parent.TryGetTarget(out SqlTransaction parent))
```
cheenamalhotra, line 593: As per current design, should we ensure `_parent` is set to `null` even if `_parent.TryGetTarget` returns false?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalTransaction.cs
```diff
@@ -533,20 +533,16 @@ internal void Zombie()
 
         private void ZombieParent()
         {
-            if (null != _parent)
+            if (_parent != null && _parent.TryGetTarget(out SqlTransaction parent))
```
cheenamalhotra, line 536: Same here

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/sqlinternaltransaction.cs
Wraith2: if parent is null TryGetTarget will return false, I checked the code in runtime. The corollary is also true, if TryGetTarget returns true then parent is guaranteed not to be null.

cheenamalhotra: I meant reverse, if **parent** is **not** null, will **TryGetTarget** always return **true**?
If no, then we have a case where parent will not be zombied to null, if there was no target attached.

Wraith2: > if parent is not null, will TryGetTarget always return true?
yes.

Wraith2:
```csharp
        public bool TryGetTarget([MaybeNullWhen(false), NotNullWhen(true)] out T target)
        {
            // Call the worker method that has more performant but less user friendly signature.
            T? o = this.Target;
            target = o!;
            return o != null;
        }
```

https://github.com/dotnet/runtime/blob/efd818b90d61a81b73b20f3d1720279f993faa2c/src/libraries/System.Private.CoreLib/src/System/WeakReference.T.cs#L54-L60

You can tell from the naming that is is old, o wouldn't pass api review today.

cheenamalhotra: Well, I'd say it depends on whether there was an item added to it, but is not going to return "true" _by design_. Just to keep things consistent with current design, I'd recommend putting it in second block and setting parent to _null_ never-the-less.

Wraith2: Done.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalTransaction.cs
Wraith2: Done.

cheenamalhotra (approved):

Issue #1151, opened 2021-07-06
Cannot connect when OS has TLS1.3 (and TLS1.2) enabled


Rans4ckeR:
### Describe the bug
After enabling TLS1.3 in the OS (for both client and server) the connection fails.
TLS1.2 is also enabled.
Client & server run on the same machine.

```
Exception.GetType: Microsoft.Data.SqlClient.SqlException
Exception.Message: A connection was successfully established with the server, but then an error occurred during the login process. (provider: SSL Provider, error: 0 - No process is on the other end of the pipe.)
Exception.Source: Core Microsoft SqlClient Data Provider
Exception.TargetSite: Void OnError(Microsoft.Data.SqlClient.SqlException, Boolean, System.Action`1[System.Action])
SqlException.Errors:
SqlError.Number: 233
SqlError.State: 0
SqlError.Class: 20
SqlError.Server: .
SqlError.Message: A connection was successfully established with the server, but then an error occurred during the login process. (provider: SSL Provider, error: 0 - No process is on the other end of the pipe.)
SqlError.Procedure: 
SqlError.LineNumber: 0
ExternalException.ErrorCode: -2146232060
ExternalException.ErrorCode Hex: 0x80131904
Win32Exception.Message: Unknown error (0x80131904)
HResult Facility: FACILITY_URT
HResult Code: 6404
HResult Severity: Failure
HResult Severe Failure: No Severe Failure
HResult Customer: Microsoft-defined
HResult Mapped NT Status Value: No NT status value mapped.
HResult Values: HRESULT values are status values.
Exception.StackTrace:    at Microsoft.Data.SqlClient.SqlInternalConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction) in Microsoft.Data.SqlClient.dll:token 0x6000fcc+0x27
   at Microsoft.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObject stateObj, Boolean callerHasConnectionLock, Boolean asyncClose) in Microsoft.Data.SqlClient.dll:token 0x6000ab1+0x174
   at Microsoft.Data.SqlClient.TdsParserStateObject.ThrowExceptionAndWarning(Boolean callerHasConnectionLock, Boolean asyncClose) in Microsoft.Data.SqlClient.dll:token 0x6001354+0x0
[...]

PR #1168, opened 2021-07-13
Fix | Exclude unsupported protocols

DavoudEshtehari:
This change covers `Managed SNI`.
There is an internal PR for `Native SNI` too.

- To use OS enabled client protocols instead, you must enable the following switch:
```
  "Switch.Microsoft.Data.SqlClient.UseSystemDefaultSecureProtocols"
```
related issue #1151


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNINpHandle.cs
```diff
@@ -355,8 +355,7 @@ public override uint EnableSsl(uint options)
                 _validateCert = (options & TdsEnums.SNI_SSL_VALIDATE_CERTIFICATE) != 0;
                 try
                 {
-
-                    _sslStream.AuthenticateAsClient(_targetServer);
+                    _sslStream.AuthenticateAsClient(_targetServer, null, SupportedProtocols, true);
```
ErikEJ, line 358: Should this test code also be updated? https://github.com/dotnet/SqlClient/blob/2c2f100d8ab9a1f6ee8c048a260d77e6a738e1e7/src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.EndPoint/TDSParser.cs#L94

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIHandle.cs
```diff
@@ -11,6 +12,11 @@ namespace Microsoft.Data.SqlClient.SNI
     /// </summary>
     internal abstract class SNIHandle
     {
+        /// <summary>
+        /// TLS 1.3 (not fully supported) in addition to obsolete and insecure protocols are excluded.
+        /// </summary>
+        protected readonly SslProtocols SupportedProtocols = SslProtocols.Tls12 | SslProtocols.Tls11 | SslProtocols.Tls;
```
ErikEJ, line 18: Is this a breaking change?

DavoudEshtehari: Since managed SNI is newer than the native SNI, I tried to remove the obsolete protocols here. The main purpose of this PR was just to exclude Tls 1.3; So, we decided to avoid a breaking change right now and include the obsolete SSL protocols.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNINpHandle.cs
DavoudEshtehari: Server protocol is set to Tls 1.2; So, the client has to obey it. I don't think modifying it will affect anything.

cheenamalhotra (approved):

JRahnama:
LGTM. Is it possible to have some tests for this?

JRahnama (approved):

cheenamalhotra:
Just FYI - requires Native SNI release and version upgrade as well to fix the issue completely.

BUILDGUIDE.md
```diff
@@ -257,6 +257,12 @@ To use this feature, you must enable the following AppContext switch at applicat
 
 **"Switch.Microsoft.Data.SqlClient.LegacyRowVersionNullBehavior"**
 
+## Enabling OS secure protocols preference
+
+Tls 1.3 has been taken out in view of the fact that the driver lacks full support. To jump back on the OS preferences as same as earlier, you must enable the following AppContext switch on application's startup:
```
ErikEJ, line 262: TLS ?

David-Engel: Yes, ditto.
```suggestion
TLS 1.3 has been excluded due to the fact that the driver lacks full support. To enable OS preferences as before, enable the following AppContext switch on application startup:
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIHandle.cs
```diff
@@ -11,6 +12,16 @@ namespace Microsoft.Data.SqlClient.SNI
     /// </summary>
     internal abstract class SNIHandle
     {
+        /// <summary>
+        /// Exclude Tls 1.3 (not fully supported).
```
David-Engel, line 16:
```suggestion
        /// Exclude TLS 1.3 (not fully supported).
```

BUILDGUIDE.md
```diff
@@ -257,6 +257,12 @@ To use this feature, you must enable the following AppContext switch at applicat
...
+
+**"Switch.Microsoft.Data.SqlClient.EnableSecureProtocolsByOS"**
```
David-Engel, line 264: I would name the switch `Switch.Microsoft.Data.SqlClient.UseSystemDefaultSecureProtocols`.

Issue #1164, opened 2021-07-12
InvalidCastException when using DateTime as parameter value for DbType.Time


natehitze:
I'm not sure if this is a bug or something that would be nice to have documented in the porting advice document. I didn't write this code so I'm not sure if passing DateTime was a requirement for System.Data at some point or not :smile: 

System.Data.SqlClient worked fine with this:

```csharp
p.DbType = DbType.Time;
if (val != null)
{
    var v = (LocalTime)val;
    p.Value = new DateTime(1900, 1, 1,
                           v.Hour,
                           v.Minute,
                           v.Second,
                           v.Millisecond,
                           DateTimeKind.Unspecified);
}
```

In Microsoft.Data.SqlClient, using that parameter gave me this exception:

```
System.InvalidCastException: Failed to convert parameter value from a DateTime to a TimeSpan.
at Microsoft.Data.SqlClient.SqlParameter.CoerceValue(System.Object value, Microsoft.Data.SqlClient.MetaType destinationType, System.Boolean& coercedToDataFeed, System.Boolean& typeChanged, System.Boolean allowStreaming) in Microsoft.Data.SqlClient.SqlParameter line 745
at Microsoft.Data.SqlClient.SqlParameter.GetCoercedValue() in Microsoft.Data.SqlClient.SqlParameter line 86
at Microsoft.Data.SqlClient.SqlParameter.Validate(System.Int32 index, System.Boolean isCommandProc) in Microsoft.Data.SqlClient.SqlParameter line 139
at Microsoft.Data.SqlClient.SqlCommand.BuildParamList(Microsoft.Data.SqlClient.TdsParser parser, Microsoft.Data.SqlClient.SqlParameterCollection parameters, System.Boolean includeReturnValue) in Microsoft.Data.SqlClient.SqlCommand0
at Microsoft.Data.SqlClient.SqlCommand.BuildExecuteSql(System.Data.CommandBehavior behavior, System.String commandText, Microsoft.Data.SqlClient.SqlParameterCollection parameters, Microsoft.Data.SqlClient._SqlRPC& rpc) in Microsoft.Data.SqlClient.SqlCommand line 132
[...]

Issue #434, opened 2020-02-18
Change in handling of Date/DateTime values after migration from v1.0.19269.1 to v1.1.1.


dev-ska:
On v1.0: the DbType.Date parameter, which has a DateTime value, is saved as a DateTime in the database.
On v1.1: the DbType.Date parameter, which has a DateTime value, is saved as a Date in the database (a time part is truncated).
The database column is of type datetime2.

Table declaration:
```
CREATE TABLE SqlClientTest
(
	DateTimeColumn datetime2 NOT NULL
)
GO
```

Code:
```
using (var connection = new SqlConnection(connectionString))
{
	var command = new SqlCommand($"insert into SqlClientTest (DateTimeColumn) values (@value)", connection);
	var parameter = command.CreateParameter();

	parameter.ParameterName = "@value";
	parameter.Value = DateTime.Now;
	parameter.DbType = DbType.Date;

	command.Parameters.Add(parameter);

	connection.Open();
	command.ExecuteNonQuery(); // running on v1.1, DateTimeColumn has no time part
	connection.Close();
}
```

Was this a functional change?

ErikEJ:
Yes, I believe so: https://github.com/dotnet/SqlClient/pull/269

dev-ska:
@ErikEJ Thanks!

cheenamalhotra:
@dev-ska 

Since you're specifying "parameter.DbType = DbType.Date;"  you're seeing this difference.
You need to specify "parameter.DbType = DbType.DateTime" in order to receive the time part.

dev-ska:
Thanks, it's understandable. The question was raised when the code (with an old history) successfully storing datetime's using SqlType.Date started to fail after switching to the new package version.

ErikEJ:
Hmm.. So a breaking change caused by a wrong mapping (bug) in your code.

dev-ska:
Yes, I'm unsure about the original reason for mapping date times to DbType.Date. Might be something related to that the code should support SQL and Oracle databases. I've switched to the DbType.DateTime mapping and it works as it used to with SQL. But this is a subject for further investigation on how the oracle provider behaves with the mapping update. The eventual goal is to keep the previous behaviour on both database types.

PR #269, opened 2019-10-16
Get and set expected value for DdType.Date and DbType.Time

ErikEJ:
Breaks support for SQL Server 2005 (I believe)

fixes #5

PR #1173, opened 2021-07-14
Document difference in DateTime vs TimeSpan value for DbType.Time parameter

natehitze:
Added the only difference in functionality I've found so far, from #1164.

### DIFF
--- a/porting-cheat-sheet.md
+++ b/porting-cheat-sheet.md
@@ -36,6 +36,7 @@ For .NET Framework projects it may be necessary to include the following in your
 | System.Data.SqlClient | Microsoft.Data.SqlClient |
 |--|--|
 | Can use DateTime object as value for SqlParameter with type `DbType.Time`. | Must use TimeSpan object as value for SqlParameter with type `DbType.Time`. |
+| Using DateTime object as value for SqlParameter with type `DbType.Date` would send date and time to SQL Server. | DateTime object's time components will be truncated when sent to SQL Server using `DbType.Date`. |
 
 ## Contribute to this Cheat Sheet
 
--- a/porting-cheat-sheet.md
+++ b/porting-cheat-sheet.md
@@ -31,6 +31,12 @@ For .NET Framework projects it may be necessary to include the following in your
 </configuration>
 ```
 
+## Functionality Changes
+
+| System.Data.SqlClient | Microsoft.Data.SqlClient |
+|--|--|
+| Can use DateTime object as value for SqlParameter with type `DbType.Time`. | Must use TimeSpan object as value for SqlParameter with type `DbType.Time`. |
+
 ## Contribute to this Cheat Sheet
 
 We would love the SqlClient community to help enhance this cheat sheet by contributing experiences and challenges faced when porting their applications.


porting-cheat-sheet.md
```diff
@@ -31,6 +31,12 @@ For .NET Framework projects it may be necessary to include the following in your
 </configuration>
 ```
 
+## Functionality Changes
+
+| System.Data.SqlClient | Microsoft.Data.SqlClient |
+|--|--|
+| Can use DateTime object as value for SqlParameter with type `DbType.Time`. | Must use TimeSpan object as value for SqlParameter with type `DbType.Time`. |
```
cheenamalhotra, line 38: Also applies to `DbType.Date`, please add that too.
ref: https://github.com/dotnet/SqlClient/pull/269

natehitze: I think the `DbType.Date` change is little less straightforward since there's no easy equivalent of "Use TimeSpan instead" for dates. But hopefully what I've added makes sense.

cheenamalhotra:
Hey @ErikEJ 

Would like to request your review too 😊

ErikEJ:
Should all breaking changes be documented on this page? Maybe better to link to the breaking changes in release notes:

https://github.com/dotnet/SqlClient/blob/main/release-notes/2.0/2.0.0.md#breaking-changes-1

https://github.com/dotnet/SqlClient/blob/main/release-notes/3.0/3.0.0.md#breaking-changes-1

I wonder if this should have been documented as a partly breaking change in 1.1 release notes? 

There are hints here: https://github.com/dotnet/SqlClient/pull/269#issuecomment-549483004 and here: https://github.com/dotnet/SqlClient/issues/434 that maybe it should.

Also the 1.1 release notes do not seem to be consolidated with the 1.1 previews: https://github.com/dotnet/SqlClient/blob/main/release-notes/1.1/1.1.0.md (so the fix for #5 is not even mentioned)

cheenamalhotra:
That's unfortunate :(

I agree giving link to breaking changes is definitely a good idea. But also calling out these important changes in comparison to System.Data.SqlClient that matter the most while porting would also help readers catch stuff they should really care about.

We can take care of remaining changes separately, this PR can focus only on Date/Time related changes. I wanted to run by you to ensure details are correct.

ErikEJ:
I think merging this without linking to the breaking changes sections would be misleading.

Otherwise LGTM

natehitze:
IMO linking to the breaking changes actually could be misleading in the context of this document, unless I'm misunderstanding the point of the breaking changes doc. My assumption is that breaking changes sections documents what breaks between versions of this library which may or may not apply to a conversion from System.Data.SqlClient (e.g. breaking changes in features that were only introduced in Microsoft.Data.SqlClient). 

My suggestion would be to comb through the release notes and copy breaking changes that apply from there to here... but I don't know enough about the libraries to accomplish that.

ErikEJ:
I had a look through, and none of the other changes seem important!

ErikEJ (approved):

cheenamalhotra (approved):

PR #1174, opened 2021-07-16
Activity tracking + Event Source support for AKV Provider

cheenamalhotra:
Introduces Event Source tracing support in AKV Provider, with name:

    `Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider.EventSource`

Enables activity tracking with `System.Diagnostics.Trace.CorrelationManager.ActivityId`

### DIFF
--- a/src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/AKVEventSource.cs
+++ b/src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/AKVEventSource.cs
@@ -15,6 +15,8 @@ namespace Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider
         // Defines the singleton instance for the Resources ETW provider
         public static AKVEventSource Log = new();
 
+        private AKVEventSource() { }
+
         // Initialized static Scope IDs
         private static long s_nextScopeId = 0;
 
@@ -68,9 +70,10 @@ namespace Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider
         {
             if (IsTraceEnabled())
             {
-                WriteTrace(memberName, p1 == null ? message :
+                WriteTrace(string.Format("Caller: {0}, Message: {1}", memberName,
+                    p1 == null ? message :
                     string.Format(message, p1.ToString().Length > 10 ?
-                    p1.ToString().Substring(0, 10) + "..." : p1));
+                    p1.ToString().Substring(0, 10) + "..." : p1)));
             }
         }
         [NonEvent]
@@ -93,23 +96,20 @@ namespace Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider
         #endregion
 
         #region Write Events
-        [Event(1, Message = "Caller: {0}, Message: {1}", Level = EventLevel.Informational, Keywords = Keywords.Trace)]
-        internal void WriteTrace(string caller, string message) => WriteEvent(1, caller, message);
+        [Event(1, Level = EventLevel.Informational, Keywords = Keywords.Trace)]
+        internal void WriteTrace(string message) => WriteEvent(1, message);
 
-        [Event(2, Message = "Caller: {0}", Level = EventLevel.Informational, Task = Tasks.Scope, Opcode = EventOpcode.Start, Keywords = Keywords.Scope)]
+        [Event(2, Level = EventLevel.Informational, Task = Tasks.Scope, Opcode = EventOpcode.Start, Keywords = Keywords.Scope)]
         internal long ScopeEnter(string caller)
         {
             SetCurrentThreadActivityId(GetCorrelationActivityId());
             long scopeId = Interlocked.Increment(ref s_nextScopeId);
-            WriteEvent(2, caller + $", Entered Scope: {scopeId}", scopeId);
+            WriteEvent(2, string.Format("Entered Scope: {0}, Caller: {1}", scopeId, caller));
             return scopeId;
         }
 
-        [Event(3, Message = "Exited Scope: {0}, Correlation Activity Id: {1}", Level = EventLevel.Informational, Task = Tasks.Scope, Opcode = EventOpcode.Stop, Keywords = Keywords.Scope)]
-        internal void ScopeExit(long scopeId)
-        {
-            WriteEvent(3, scopeId);
-        }
+        [Event(3, Level = EventLevel.Informational, Task = Tasks.Scope, Opcode = EventOpcode.Stop, Keywords = Keywords.Scope)]
+        internal void ScopeExit(long scopeId) => WriteEvent(3, scopeId);
         #endregion
     }
 
--- a/src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/AKVEventSource.cs
+++ b/src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/AKVEventSource.cs
@@ -1,0 +1,127 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Diagnostics;
+using System.Diagnostics.Tracing;
+using System.Threading;
+
+namespace Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider
+{
+    [EventSource(Name = "Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider.EventSource")]
+    internal class AKVEventSource : EventSource
+    {
+        // Defines the singleton instance for the Resources ETW provider
+        public static AKVEventSource Log = new();
+
+        // Initialized static Scope IDs
+        private static long s_nextScopeId = 0;
+
+        // Provides Correlation Manager Activity Id for current scope.
+        private static Guid GetCorrelationActivityId()
+        {
+            if (Trace.CorrelationManager.ActivityId == Guid.Empty)
+            {
+                Trace.CorrelationManager.ActivityId = Guid.NewGuid();
+            }
+            return Trace.CorrelationManager.ActivityId;
+        }
+
+        #region Keywords
+        public class Keywords
+        {
+            /// <summary>
+            /// Captures basic application flow trace events.
+            /// </summary>
+            internal const EventKeywords Trace = (EventKeywords)1;
+
+            /// <summary>
+            /// Captures basic application scope entering and exiting events.
+            /// </summary>
+            internal const EventKeywords Scope = (EventKeywords)2;
+        }
+        #endregion
+
+        #region Tasks
+        /// <summary>
+        /// Tasks supported by AKV Provider's EventSource implementation
+        /// </summary>
+        public class Tasks
+        {
+            /// <summary>
+            /// Task that tracks trace scope.
+            /// </summary>
+            public const EventTask Scope = (EventTask)1;
+        }
+        #endregion
+
+        [NonEvent]
+        internal bool IsTraceEnabled() => Log.IsEnabled(EventLevel.Informational, Keywords.Trace);
+
+        [NonEvent]
+        internal bool IsScopeEnabled() => Log.IsEnabled(EventLevel.Informational, Keywords.Scope);
+
+        #region Event Methods
+        [NonEvent]
+        internal void TryTraceEvent(string message, object p1 = null, [System.Runtime.CompilerServices.CallerMemberName] string memberName = "")
+        {
+            if (IsTraceEnabled())
+            {
+                WriteTrace(memberName, p1 == null ? message :
+                    string.Format(message, p1.ToString().Length > 10 ?
+                    p1.ToString().Substring(0, 10) + "..." : p1));
+            }
+        }
+        [NonEvent]
+        internal long TryScopeEnterEvent(string memberName)
+        {
+            if (IsScopeEnabled())
+            {
+                return ScopeEnter(memberName);
+            }
+            return default;
+        }
+        [NonEvent]
+        internal void TryScopeExitEvent(long scopeId)
+        {
+            if (Log.IsScopeEnabled())
+            {
+                ScopeExit(scopeId);
+            }
+        }
+        #endregion
+
+        #region Write Events
+        [Event(1, Message = "Caller: {0}, Message: {1}", Level = EventLevel.Informational, Keywords = Keywords.Trace)]
+        internal void WriteTrace(string caller, string message) => WriteEvent(1, caller, message);
+
+        [Event(2, Message = "Caller: {0}", Level = EventLevel.Informational, Task = Tasks.Scope, Opcode = EventOpcode.Start, Keywords = Keywords.Scope)]
+        internal long ScopeEnter(string caller)
+        {
+            SetCurrentThreadActivityId(GetCorrelationActivityId());
+            long scopeId = Interlocked.Increment(ref s_nextScopeId);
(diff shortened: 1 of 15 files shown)


johnnypham:
There are traces for cache misses but none for cache hits (local cache). But it looks like it could be difficult to differentiate between a cache hit for unwrapping a CEK and and verifying a signature, so this is just a suggestion.

Have you checked that perf is not affected?

Otherwise, LGTM.

JRahnama:
LGTM.

cheenamalhotra:
> Have you checked that perf is not affected?

Performance compared before and after - no significant change.

src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/LocalCache.cs
```diff
@@ -77,9 +77,12 @@ internal TValue GetOrCreate(TKey key, Func<TValue> createItem)
                     AbsoluteExpirationRelativeToNow = TimeToLive
                 };
 
+                AKVEventSource.Log.TryTraceEvent("Encryption key added to local cache.");
```
johnnypham, line 80: This cache can contain both keys and signature verification results.

src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/LocalCache.cs
```diff
@@ -77,9 +77,12 @@ internal TValue GetOrCreate(TKey key, Func<TValue> createItem)
...
+
                 _cache.Set(key, cacheEntry, cacheEntryOptions);
             }
 
+            AKVEventSource.Log.TryTraceEvent("Encryption key entry fetched from local cache.");
```
johnnypham, line 85: This will be logged even if the entry did not exist in the cache.

cheenamalhotra: Yep - it will be added above to cache above so its safe to say. I added that knowingly.

src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/LocalCache.cs
cheenamalhotra: Below wer're adding key, with a new cache Entry, does this statement need to change?

johnnypham: The message says an encryption key was added but this cache is currently used for both encryption keys and signature verification results. I don't think there's an easy way to differentiate between the two.

src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/LocalCache.cs
johnnypham: If there is a cache miss and `Func<TValue> createItem` is used to create the cache entry, the event implies that the entry was fetched from the cache, i.e. that it already existed.

cheenamalhotra: I can handle that, but it will be preceded with "new cache entry added" message so that would imply a new entry was  fetched and added to cache, I'm fine eitherway.

src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/LocalCache.cs
cheenamalhotra: I only meant to say "The **key** for the cache entry" is added to the cache.
Not what it finds as a result of "createItem()". I can reword it better or simply ignore this message then?

johnnypham: I think it should remain since caching can be disabled and this event could be helpful in that case. A generic message would be best.

src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/LocalCache.cs
johnnypham: Sure, that makes sense.

src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/AKVEventSource.cs
```diff
@@ -0,0 +1,127 @@
...
+using System.Diagnostics.Tracing;
+using System.Threading;
+
+namespace Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider
+{
+    [EventSource(Name = "Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider.EventSource")]
+    internal class AKVEventSource : EventSource
+    {
+        // Defines the singleton instance for the Resources ETW provider
+        public static AKVEventSource Log = new();
```
DavoudEshtehari, line 16: It's a good practice to hide the default ctor in a singleton approach by explicitly defining a private ctor. It seems we missed this part on SqlClientEventSource too.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
```diff
@@ -789,7 +789,7 @@ public class TraceEventListener : EventListener
 
             protected override void OnEventSourceCreated(EventSource eventSource)
             {
-                if (eventSource.Name.Equals("Microsoft.Data.SqlClient.EventSource"))
+                if (eventSource.Name.StartsWith("Microsoft.Data.SqlClient"))
```
DavoudEshtehari, line 792: By adding support for more than one event source, I suggest including the event source name in the list to verify if all events run per event source.

Also, could you add a test for `AKVEventSource`, please?

DavoudEshtehari (commented):
Have you skipped including the type name in tracing information on purpose? If don't, could you include it in all traces in addition to the caller name?

src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/LocalCache.cs
```diff
@@ -61,11 +61,13 @@ internal TValue GetOrCreate(TKey key, Func<TValue> createItem)
         {
             if (TimeToLive <= TimeSpan.Zero)
             {
+                AKVEventSource.Log.TryTraceEvent("Key caching found disabled, fetching key information.");
                 return createItem();
             }
 
             if (!_cache.TryGetValue(key, out TValue cacheEntry))
             {
+                AKVEventSource.Log.TryTraceEvent("Fetching key information.");
```
johnnypham, line 70: Shouldn't this be a generic message?

src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/AzureSqlKeyCryptographer.cs
```diff
@@ -208,10 +216,13 @@ private void CreateKeyClient(Uri vaultUri)
...
         private void ParseAKVPath(string masterKeyPath, out Uri vaultUri, out string masterKeyName, out string masterKeyVersion)
         {
-            Uri masterKeyPathUri = new Uri(masterKeyPath);
+            Uri masterKeyPathUri = new(masterKeyPath);
             vaultUri = new Uri(masterKeyPathUri.GetLeftPart(UriPartial.Authority));
             masterKeyName = masterKeyPathUri.Segments[2];
             masterKeyVersion = masterKeyPathUri.Segments.Length > 3 ? masterKeyPathUri.Segments[3] : null;
+
+            AKVEventSource.Log.TryTraceEvent("Received Key Name: {0}", masterKeyName);
+            AKVEventSource.Log.TryTraceEvent("Received Key Version: {0}", masterKeyVersion);
```
johnnypham, line 225: Any security concerns in logging the entire version?

cheenamalhotra: All parameters are split at 10 character length universally:
https://github.com/dotnet/SqlClient/pull/1174/files#diff-fdef6a270d141d4cf0ddd2de0debdba3e6c66dc8f99fbc072dcbf3a8b770f59bR76

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/AKVUnitTests.cs
```diff
@@ -21,31 +23,84 @@ public static class AKVUnitTests
...
+            int currentScope = 0;
+
+            // Validate event data captured.
+            Assert.All(eventData, item =>
+            {
+                Assert.Equal(DataTestUtility.AKVEventSourceName, item.EventSource.Name);
+                Assert.Equal(threadActivityId, item.ActivityId);
+                Assert.Equal(EventLevel.Informational, item.Level);
+                Assert.NotNull(item.Payload);
+                Assert.True(item.Payload.Count == 1);
```
johnnypham, line 39: Can you use `Assert.Equal` instead?

johnnypham (approved):

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/AKVUnitTests.cs
cheenamalhotra:
```suggestion
                Assert.Equal(1, item.Payload.Count);
```

cheenamalhotra: It didn't work quite well:

```
error xUnit2013: Do not use Assert.Equal() to check for collection size.
```

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/AKVUnitTests.cs
```diff
@@ -21,31 +23,84 @@ public static class AKVUnitTests
...
+            int currentScope = 0;
+
+            // Validate event data captured.
+            Assert.All(eventData, item =>
+            {
+                Assert.Equal(DataTestUtility.AKVEventSourceName, item.EventSource.Name);
+                Assert.Equal(threadActivityId, item.ActivityId);
+                Assert.Equal(EventLevel.Informational, item.Level);
+                Assert.NotNull(item.Payload);
+                Assert.Equal(1, item.Payload.Count);
```
cheenamalhotra, line 39:
```suggestion
                Assert.True(item.Payload.Count == 1);
```

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/TestFixtures/DatabaseHelper.cs
```diff
@@ -15,19 +17,151 @@ class DatabaseHelper
...
+            while (sqlDataReader.Read())
             {
-                sqlCommand.Parameters.AddWithValue(@"CustomerId", customer.Id);
-                sqlCommand.Parameters.AddWithValue(@"FirstName", customer.FirstName);
-                sqlCommand.Parameters.AddWithValue(@"LastName", customer.LastName);
+                if (sqlDataReader.FieldCount == 3)
+                {
+                    Assert.True(sqlDataReader.GetInt32(0) == 45, "Employee id didn't match.");
+                    Assert.True(sqlDataReader.GetString(1) == @"Microsoft", "Employee FirstName didn't match.");
+                    Assert.True(sqlDataReader.GetString(2) == @"Corporation", "Employee LastName didn't match.");
```
DavoudEshtehari, line 50: NIT: Generally, my preference is using `Assert.Equal` instead of `Assert.True/False` for a better report on failures in each possible situation.

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/AKVUnitTests.cs
DavoudEshtehari:
```suggestion
                Assert.Single(item.Payload);
```
+ Use `Empty` if you don't expect any item.

DavoudEshtehari (approved):

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/TestFixtures/DatabaseHelper.cs
cheenamalhotra:
```suggestion
                    Assert.Equal(45, sqlDataReader.GetInt32(0), "Employee id didn't match.");
                    Assert.Equal(@"Microsoft", sqlDataReader.GetString(1), "Employee FirstName didn't match.");
                    Assert.Equal(@"Corporation", sqlDataReader.GetString(2), "Employee LastName didn't match.");
```

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
```diff
@@ -63,11 +63,13 @@ public static class DataTestUtility
...
         public const string AKVKeyName = "TestSqlClientAzureKeyVaultProvider";
-
+        public const string EventSourcePrefix = "Microsoft.Data.SqlClient";
+        public const string MDSEventSourceName = "Microsoft.Data.SqlClient.EventSource";
+        public const string AKVEventSourceName = "Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider.EventSource";
         private const string ManagedNetworkingAppContextSwitch = "Switch.Microsoft.Data.SqlClient.UseManagedNetworkingOnWindows";
 
         private static Dictionary<string, bool> AvailableDatabases;
-        private static TraceEventListener TraceListener;
+        public static BaseEventListener TraceListener;
```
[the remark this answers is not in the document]
cheenamalhotra: It's the TraceListener we enable traces from, it may not need to be public anymore so I can change it back to private.
```suggestion
        private static BaseEventListener TraceListener;
```

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
```diff
@@ -783,22 +785,46 @@ public static string RetrieveValueFromConnStr(string connStr, string[] keywords)
...
+        public class MDSEventListener : BaseEventListener
         {
-            public List<int> IDs = new List<int>();
+            public override string _name => MDSEventSourceName;
+        }
+
+        public class BaseEventListener : EventListener
+        {
+            public List<int> IDs = new();
+            public List<EventWrittenEventArgs> EventData = new();
```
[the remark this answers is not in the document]
cheenamalhotra:
```suggestion
            public readonly List<int> IDs = new();
            public readonly List<EventWrittenEventArgs> EventData = new();
```

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
```diff
@@ -783,22 +785,46 @@ public static string RetrieveValueFromConnStr(string connStr, string[] keywords)
...
             protected override void OnEventWritten(EventWrittenEventArgs eventData)
             {
-                IDs.Add(eventData.EventId);
+                if (_name == eventData.EventSource.Name)
+                {
+                    IDs.Add(eventData.EventId);
+                    if (EventData != null)
+                    {
+                        EventData.Add(eventData);
+                    }
```
cheenamalhotra, line 826:
```suggestion
                    EventData.Add(eventData);
```

PR #1179, opened 2021-07-20
Fix | Improve decimal conversion from SqlDecimal to .Net Decimal

### LINKED ISSUE #95: SqlClient fails to convert decimal(38,20) to .Net decimal for values within the range of .Net's decimal type
This program

```
        {
            using (var con = new SqlConnection("server=.;database=tempdb;Integrated Security=true"))
            {
                con.Open();
                var cmd = new SqlCommand("select cast(4210862852.86 as decimal(38,20))  val", con);
                using (SqlDataReader rdr = cmd.ExecuteReader())
                {
                    rdr.Read();

                    var val = rdr.GetDecimal(0);
                    Console.WriteLine(val);
                }
            }
        }
```

fails with 

System.OverflowException: 'Conversion overflows.'

at [SqlBuffer.cs](https://github.com/dotnet/corefx/blob/master/src/System.Data.SqlClient/src/System/Data/SqlClient/SqlBuffer.cs) line 254:

```
                if (StorageType.Decimal == _type)
                {
                    if (_value._numericInfo.data4 != 0 || _value._numericInfo.scale > 28)
                    {
                        throw new OverflowException(SQLResource.ConversionOverflowMessage);
                    }
```
This value is well within the range of a .NET decimal, and it's a value that the SqlClient can _write_ using a SqlParameter.  So it should be converted to .NET decimal. 

The byte pattern in SQL and TDS for this value is:

`0x261400010000D877FB4DEE8B51699A5005000000`

And it looks like SqlBuffer refuses to convert any value with a non-zero value in the last 4 bytes.  But a .NET Decimal is stored as a 12-byte integer, along with a sign and a scaling factor.

So the apparent intent is to determine if the SqlDecimal uses a larger-than-12 byte integer and trigger the overflow.

So this conversion has a bug.  It looks like the SqlClient is confused on the byte ordering SQL Server is using, and the last 4 bytes of this buffer are not the 4 least-significant bytes of the 16-byte integer embedded in the decimal.  

This number requires uses 12 (base-10) decimal digits to store, so it requires fewer than 12 (base-256) bytes to store.

https://docs.microsoft.com/en-us/dotnet/api/system.decimal?redirectedfrom=MSDN&view=netframework-4.8

DavoudEshtehari:
Fixes #95

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBuffer.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBuffer.cs
@@ -226,9 +226,10 @@ namespace Microsoft.Data.SqlClient
                                                        _value._numericInfo._data4);
 
                             // Remove trail zeros after the point from the end to calculate the valuable floating part
-                            string[] valueParts = sqlValue.ToString().TrimEnd('0').Split('.');
-                            int scale = valueParts[1].Length;
-                            int precision = valueParts[0].Length + scale + (sqlValue.IsPositive ? 0 : -1);
+                            ReadOnlySpan<char> strValue = sqlValue.ToString().AsSpan().TrimEnd('0');
+                            int integral = strValue.IndexOf('.');
+                            int scale = strValue.Length - integral - 1;
+                            int precision = integral + scale + (sqlValue.IsPositive ? 0 : -1);
 
                             // Out of range
                             if ((precision > 29 || scale > 28) && !(precision == 29 && scale == 28))
@@ -241,7 +242,7 @@ namespace Microsoft.Data.SqlClient
                                 // ex: (precision == 29 && scale == 28)
                                 //  valid:   (+/-)7.1234567890123456789012345678
                                 //  invalid: (+/-)8.1234567890123456789012345678
-                                bool tryMaxPrec = (valueParts[0].Length == 1 || (valueParts[0].Length == 2 && !_value._numericInfo._positive));
+                                bool tryMaxPrec = integral == 1 || (integral == 2 && !_value._numericInfo._positive);
                                 int precConvert = 29;
                                 int scaleConvert;
 
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBuffer.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBuffer.cs
@@ -205,6 +205,7 @@ namespace Microsoft.Data.SqlClient
             }
         }
 
+        #region Decimal
         internal decimal Decimal
         {
             get
@@ -215,64 +216,44 @@ namespace Microsoft.Data.SqlClient
                 {
                     if (_value._numericInfo._data4 != 0 || _value._numericInfo._scale > 28)
                     {
+                        // Only removing trailing zeros from a decimal part won't hit its value!
                         if (_value._numericInfo._scale > 0)
                         {
-                            SqlDecimal sqlValue = new(_value._numericInfo._precision,
-                                                       _value._numericInfo._scale,
-                                                       _value._numericInfo._positive,
-                                                       _value._numericInfo._data1,
-                                                       _value._numericInfo._data2,
-                                                       _value._numericInfo._data3,
-                                                       _value._numericInfo._data4);
+                            int zeroCnt = FindTrailingZerosAndPrec((uint)_value._numericInfo._data1, (uint)_value._numericInfo._data2, 
+                                                                   (uint)_value._numericInfo._data3, (uint)_value._numericInfo._data4, 
+                                                                   _value._numericInfo._scale, out int precision);
 
-                            // Remove trail zeros after the point from the end to calculate the valuable floating part
-                            ReadOnlySpan<char> strValue = sqlValue.ToString().AsSpan().TrimEnd('0');
-                            int integral = strValue.IndexOf('.');
-                            int scale = strValue.Length - integral - 1;
-                            int precision = integral + scale + (sqlValue.IsPositive ? 0 : -1);
+                            int minScale = _value._numericInfo._scale - zeroCnt; // minimum possible sacle after removing the trailing zeros.
 
-                            // Out of range
-                            if ((precision > 29 || scale > 28) && !(precision == 29 && scale == 28))
+                            if (zeroCnt > 0 && minScale <= 28 && precision <= 29)
                             {
-                                throw new OverflowException(SQLResource.ConversionOverflowMessage);
-                            }
-                            else
-                            {
-                                // Precision could be 28 or 29
-                                // ex: (precision == 29 && scale == 28)
-                                //  valid:   (+/-)7.1234567890123456789012345678
-                                //  invalid: (+/-)8.1234567890123456789012345678
-                                bool tryMaxPrec = integral == 1 || (integral == 2 && !_value._numericInfo._positive);
-                                int precConvert = 29;
-                                int scaleConvert;
+                                SqlDecimal sqlValue = new(_value._numericInfo._precision, _value._numericInfo._scale, _value._numericInfo._positive, 
+                                                          _value._numericInfo._data1, _value._numericInfo._data2, 
+                                                          _value._numericInfo._data3, _value._numericInfo._data4);
 
-                                if (!tryMaxPrec && precision != 29)
+                                int integral = precision - minScale;
+                                int newPrec = 29;
+
+                                if (integral != 1 && precision != 29)
                                 {
-                                    precConvert = 28;
+                                    newPrec = 28;
                                 }
-                                scaleConvert = precConvert - (precision - scale);
 
                                 try
                                 {
-                                    return SqlDecimal.ConvertToPrecScale(sqlValue, precConvert, scaleConvert).Value;
+                                    // Precision could be 28 or 29
+                                    // ex: (precision == 29 && scale == 28)
+                                    // valid:   (+/-)7.1234567890123456789012345678
+                                    // invalid: (+/-)8.1234567890123456789012345678
+                                    return SqlDecimal.ConvertToPrecScale(sqlValue, newPrec, newPrec - integral).Value;
                                 }
                                 catch (OverflowException)
                                 {
-                                    if (tryMaxPrec && scale < 28)
-                                    {
-                                        return SqlDecimal.ConvertToPrecScale(sqlValue, precConvert - 1, scaleConvert).Value;
-                                    }
-                                    else
-                                    {
-                                        throw;
-                                    }
+                                    throw new OverflowException(SQLResource.ConversionOverflowMessage);
                                 }
                             }
                         }
-                        else
-                        {
-                            throw new OverflowException(SQLResource.ConversionOverflowMessage);
-                        }
+                        throw new OverflowException(SQLResource.ConversionOverflowMessage);
                     }
                     return new decimal(_value._numericInfo._data1, _value._numericInfo._data2, _value._numericInfo._data3, !_value._numericInfo._positive, _value._numericInfo._scale);
                 }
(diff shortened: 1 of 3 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBuffer.cs
```diff
@@ -215,7 +215,63 @@ internal decimal Decimal
...
+                            SqlDecimal sqlValue = new(_value._numericInfo._precision,
+                                                       _value._numericInfo._scale,
+                                                       _value._numericInfo._positive,
+                                                       _value._numericInfo._data1,
+                                                       _value._numericInfo._data2,
+                                                       _value._numericInfo._data3,
+                                                       _value._numericInfo._data4);
+
+                            // Remove trail zeros after the point from the end to calculate the valuable floating part
+                            string[] valueParts = sqlValue.ToString().TrimEnd('0').Split('.');
```
Wraith2, line 229: You could use span operations here to avoid substring and array allocations.

DavoudEshtehari: Thanks for the suggestion. I'll check it.

tannergooding: You can also divide by `1.0000000000000000000000000000m` (that is `var normalized = x / 1.0000000000000000000000000000m`) to remove all trailing zeros.
Division by `1` will remove trailing zeros, up to the number of trailing zeros the `1` has. So `x / 1m` keeps all trailing zeros, `x / 1.0m` removes up to 1 trailing zero, `x / 1.00m` removes up to 2, etc (you can have up to 29 digits total so `1.0000000000000000000000000000m` removes all trailing zeros).

Multiplication does the inverse and will add trailing zeros up to that many. So `x * 1m` adds nothing, `x * 1.0m` adds up to 1 trailing zero, `x * 1.00m` adds up to 2, etc

In both cases, this removes or adds the zero in the corresponding position (that is `2.1m * 1.00m` returns `2.10m` not `2.100m`)

DavoudEshtehari: Interesting approach @tannergooding, It works fine with **decimal** data type, while our effort is to extract possible **decimal** value from a **SqlDecimal**.
It has two steps: 
1. Resolving the shortest valuable floating part by removing the trailing zeros after point from the `SqlDecimal`.
2. Resolving a valid `decimal` with the most possible scale from the `SqlDecimal`.

In which step do you think it'd be better to use the arithmetic approach if it were possible and why?

Also, I found the following behaviors by **SqlDecimal**:
- on SSMS:
```
DECLARE @d1 decimal(38, 37); 
DECLARE @d2 decimal(38, 37); 
SET @d1 = 0.1;
SET @d2 = 1;

SELECT @d1, @d2, @d1/@d2
```
Output:
> 0.1000000000000000000000000000000000000 | 1.0000000000000000000000000000000000000 | 0.100000

- on .Net app (net core & net fx):
```
SqlDecimal.ConvertToPrecScale(new SqlDecimal(0.1), 38, 37)/SqlDecimal.ConvertToPrecScale(new SqlDecimal(1), 38, 37)
```
> System.OverflowException

Wraith2: I'm still not sure why this can't just be:

```csharp
if (_value._numericInfo._data4 != 0)
{
    SqlDecimal sqlValue = new(_value._numericInfo._precision,
                                _value._numericInfo._scale,
                                _value._numericInfo._positive,
                                _value._numericInfo._data1,
                                _value._numericInfo._data2,
                                _value._numericInfo._data3,
                                _value._numericInfo._data4);

    // max scale of System.Decimal is 28 so try this division to normalize into that range
    // and then throw our own error message if the subsequent conversion to System.Decimal fails because it is still out of range.
    try
    {
        return (decimal)(sqlValue / new SqlDecimal(1.0000_0000_0000_0000_0000_0000_0000m)); 
    }
    catch (OverflowException)
    {
        throw new OverflowException(SQLResource.ConversionOverflowMessage);
    }
}
else
{
    return new decimal(
        _value._numericInfo._data1,
        _value._numericInfo._data2,
        _value._numericInfo._data3,
        !_value._numericInfo._positive,
        _value._numericInfo._scale
    );
}
```
which uses the inbuilt operators on SqlDecimal to do the division that @tannergooding explained and then throws our custom error message if it would still overflow. Does this miss a case? It certainly makes some of the out of range cases in your new test start to pass.

DavoudEshtehari: Thank you for your attention to this discussion. My goal is to resolve a decimal value with the most possible precision even they're zero like `4210862852.860000000000000000` instead of `4210862852.86`.

Even with ignoring this requirement, by running this code you can find **OverFlowException** which comes from the SqlDecimal division issue as I've mentioned earlier besides some vague values like `0.1234567890` instead of `0.1234567890123456789012345678`.

- Trying with SSMS:
`SELECT 0.12345678901234567890123456780000000000 / 1.0000000000000000000000000000`
  Output: **0.1234567890**
  Expected: **0.1234567890123456789012345678**

Wraith2: > My goal is to resolve a decimal value with the most possible precision even they're zero like 4210862852.860000000000000000 instead of 4210862852.86

This is exact numerics so those are the same value. The trailing zeroes do not represent additional information as they could with a float they're simply an artefact of the chosen display format. System.Decimal does not include scale and precision values and that's the conversion we're dealing with here so as long as the precision we change it to is large enough to contain the required value there is no loss of information and it is acceptable.

> Trying with SSMS:
SELECT 0.12345678901234567890123456780000000000 / 1.0000000000000000000000000000

This is because the SqlDecimal result of the division is decimal(10,38) which is the default. I expect that this is modelling the implementation in SqlServer exactly. To me this is an unexpected result and I would have thought that it would keep as much precision as possible. So it looks like we can't use the SqlDecimal type to fix the problem it has to be done with bit twiddling somehow. 

My main problem with this approach is the string allocation. It should never be the best way to deal with the number. Somewhere somehow it is possible to get the information needed without roundtripping through a display format. 

I think in the case is the number of trailing zeroes is what is needed and it's used to work out if we can reduce the scale so that the new scale and precision can be fitted into a System.Decimal. So it we've got a decimal(30,30) and we two traling zeroes we can reduce both scale and precision to 28 which we know we can represent. Is that the correct understanding?

DavoudEshtehari: > My goal is to resolve a decimal value with the most possible precision even they're zero, like 4210862852.860000000000000000 instead of 4210862852.86

Even though these two are equal in value, from the scientific point of view the first number is more precise than the second number.

> I think in the case is the number of trailing zeroes is what is needed and it's used to work out if we can reduce the scale so that the new scale and precision can be fitted into a System.Decimal. So it we've got a decimal(30,30) and we two traling zeroes we can reduce both scale and precision to 28 which we know we can represent. Is that the correct understanding?

Correct, we need to know the valuable floating part without trailing zeros to verify if it's in the decimal range or not.
In addition, the final value, even with the acceptable precision and scale could be out of decimal boundary: check [here](https://github.com/dotnet/SqlClient/pull/1179/files#diff-23f8362bf395b2b82bd09e7d7e117aea0979061624bb67d96ba7d4a03a4e0022R241).

Wraith2: > > My goal is to resolve a decimal value with the most possible precision even they're zero, like 4210862852.860000000000000000 instead of 4210862852.86
> 
> Even though these two are equal in value, from the scientific point of view the first number is more precise than the second number.

Either they're equal or they're not. They represent the same value so the amount of precision is irrelevant, especially if the additional precision prevents you from representing it as a System.Decimal as we're trying to do here. I can appreciate wanting to keep as much precision as possible but it should be balanced against the goal of allowing as many values as possible to be represented. 

For any given value if the choice is to lose some unnecessary precision (trailing zeroes) or throw an error then I think it is better to drop the precision because it loses no data and allows users to continue using their data.

JRahnama: What I noticed from my conversation with @DavoudEshtehari, I can say he is trying to keep the number as precise as possible for accuracy purposes. For example 0.568789899999999999999 could be round up to 0.5687899 and this is not same as 0.5687990000000. We, the driver, should not changed any data from client. We just try to fit that in Sql boundaries, however by dividing the value we have modified it and I do not think that should happen. But again this is my idea. I would call for @David-Engel's opinion to conclude the conversations and provide a closure at this matter.

Wraith2: >  We, the driver, should not changed any data from client.

Agreed. In your example we would not round 0.56878989999999999999 to 0.5687899 because it is a loss of data. That would throw an exception. 

We Can convert 0.56878990000000 to 0.5687899 because they represent the same value. The precision is not important if we are representing the same value. Keeping as much precision as we can, so taking off only as many trailing zeroes as we need to but not all of them, isn't a problem.

The problem that I see is that using a string roundtrip per decimal is a massive perf issue and there's no way that a user can expect or account for without knowing the data beforehand. We really need to find a way to do this from the binary representation.

JRahnama: Just to share here I am testing:
```
int[] bits = decimal.GetBits(<decimal number goes here>);

            int sign = bits[3] & (1 << 31);
            int decimalPoints = (bits[3] >> 16) & 0x1f;
```
the last bit is the sign. From here we can move forward and divide by 10 for top bits, middle bits and bottom bits. and get rid of trailing zeros I believe.  Just wanted to share with you. I will search and test more to have another suggested solution.

Wraith2: Yup. Now, how to divide by 10 in a base 2 representation...

Wraith2: Would it be possible to take the formatting logic from SqlDecimal and instead of formatting to a string write to a stackalloc or rented char[]? we know the max possible length in chars. Once written we can do the same logic of checking for trailing digits but we've avoided the string allocation. 

I can't find any satisfactory method for dividing a base 2 representation by 10.

JRahnama:
```
 decimal testNumber = -7.92281625142643370M;
            int[] bits = decimal.GetBits(testNumber);
            int sign = bits[3] >>= 31;
            int exp = (bits[3] >> 16) & 0x17; // Bits 0 to 15, the lower word, are unused and must be zero.
                                              // Bits 16 to 23 must contain an exponent between 0 and 28, which indicates the power of 10 to divide the integer number.
                                              // Bits 24 to 30 are unused and must be zero.
                                              // Bit 31 contains the sign: 0 mean positive, and 1 means negative.

            uint highBits = (uint)bits[2];  // high bits
            uint middleBits = (uint)bits[1]; // Middle bits
            uint lowdBits = (uint)bits[0];  // low bits


            while (exp > 0 && ((highBits % 5) * 6 + (middleBits % 5) * 6 + lowdBits) % 10 == 0) // I am still testing this part 
            {
                highBits= DivideByTen((uint)0, highBits, out uint remainder);
                middleBits = DivideByTen(remainder, middleBits, out remainder);
                lowdBits = DivideByTen(remainder, lowdBits, out _);
                exp--;
            }


        }

        private static uint DivideByTen(uint highBits, uint lowBits, out uint remainder)
        {
            ulong total = highBits;
            total <<= 32;
            total = total | (ulong)lowBits;

            remainder = (uint)(total % 10L);
            return (uint)(total / 10L);
        }
```

JRahnama: I think this could be the solution, it may need a bit of modification.

JRahnama: Ok, I managed getting results out of it. @Wraith2 and @DavoudEshtehari can you test this:
```
 class Program
    {
        {
            decimal testNumber = -7.92281625142643370M;
            int[] bits = decimal.GetBits(testNumber);
            int sign = bits[3] & (1 << 31);
            int exp = (bits[3] >> 16) & 0x1f;// Bits 0 to 15, the lower word, are unused and must be zero.
                                              // Bits 16 to 23 must contain an exponent between 0 and 28, which indicates the power of 10 to divide the integer number.
                                              // Bits 24 to 30 are unused and must be zero.
                                              // Bit 31 contains the sign: 0 mean positive, and 1 means negative.

            uint highBits = (uint)bits[2]; // high bits
            uint middleBits = (uint)bits[1]; // Middle bits
            uint lowBits = (uint)bits[0]; // low bits

            while (exp > 0 && ((highBits % 5) * 6 + (middleBits % 5) * 6 + lowBits) % 10 == 0)
            {
                highBits = DivideByTen((uint)0, highBits, out uint remainder);
                middleBits = DivideByTen(remainder, middleBits, out remainder);
                lowBits = DivideByTen(remainder, lowBits, out _);
                exp--;
            }

            bits[0] = (int)lowBits;
            bits[1] = (int)middleBits;
            bits[2] = (int)highBits;
            bits[3] = (exp << 16) | sign;

            Console.WriteLine(new decimal(bits));
        }

        private static uint DivideByTen(uint highBits, uint lowBits, out uint remainder)
        {
            ulong total = highBits;
            total <<= 32;
            total = total | (ulong)lowBits;

            remainder = (uint)(total % 10L);
            return (uint)(total / 10L);
        }
    }
```

Wraith2: How about using inbuilt BigInteger? 

```
SqlDecimal sqlDecimal = new SqlDecimal(45644810000);
BigInteger bigInt = new BigInteger(sqlDecimal.BinData, isUnsigned: true, isBigEndian: false);
int trailingCount = 0;
for (int index = 1; index < 32; index++)
{
	if (BigInteger.Remainder(bigInt, (ulong)Math.Pow(10,index)) != 0)
	{
		break;
	}
	trailingCount += 1;
}
```

returns 4 which is correct and it's using already validated maths operations so we're not writing our own. I still don't like the allocations it would cause because it's a struct and it's using arrays internally but it bothers me a lot less than going through a string display format to do the work.

JRahnama: @Wraith2 this is good. The only problem is because BigInteger does not have upper and lower boundaries an OOM exception might be thrown..... We need to be careful of that part.

Wraith2: My idea is that we put the unscaled 128 bit integer into it so we know we've got an upper bound and that the BigInteger type can support it, it'll go a lot bigger than we need it to. So we use that to identify how many trailing zeroes we can remove which we've agreed determines how much we can de-scale the value and then we'll know if we hit the representation limit for the System.Decimal and we can convert or throw.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlBuffer.cs
```diff
@@ -288,6 +269,85 @@ internal decimal Decimal
...
+        /// <param name="data2">An 32-bit unsigned integer which will be combined with data1, data3, and data4</param>
+        /// <param name="data3">An 32-bit unsigned integer which will be combined with data1, data2, and data4</param>
+        /// <param name="data4">An 32-bit unsigned integer which will be combined with data1, data2, and data3</param>
+        /// <param name="scale">The number of decimal places</param>
+        /// <param name="valuablePrecision">OUT |The number of digits without trailing zeros</param>
+        /// <returns>Number of trailing zeros</returns>
+        private static int FindTrailingZerosAndPrec(uint data1, uint data2, uint data3, uint data4, byte scale, out int valuablePrecision)
+        {
+            // Make local copy of data to avoid modifying input.
+            uint[] rgulNumeric = new uint[4] { data1, data2, data3, data4 };
```
Wraith2, line 285: Can i suggest `Span<uint> rgulNumeric= stackalloc uint[4]{data1, data2, data3, data4};` and theh change `SqlDecimalDivBy` to take a span as well, avoids the array allocation

Kaur-Parminder (approved):

JRahnama (approved):

PR #1181, opened 2021-07-21
Enable PoolBlockingPeriod for .NET Standard + expose GetColumnSchema

cheenamalhotra:
Contains below changes:

1. Below API was not exposed in .NET Standard DLLs, even though functionality was available.
   - SqlDataReader.GetColumnSchema()

2. Pool Blocking Period has no API limitations in .NET Standard 2.0 and 2.1, tested using "-p:ReferenceType=NetStandard", hence enabling it here. Would appreciate feedback/concerns on this change.

cc @saurabh500 @David-Engel


David-Engel (commented):
Approved from a functional perspective. I don't know why it was originally excluded from netstandard.

johnnypham (approved):

JRahnama (approved):

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.NetCoreApp.cs
```diff
@@ -5,35 +5,6 @@
 // Changes to this file must follow the http://aka.ms/api-review process.
```
DavoudEshtehari, line 5: Does it require passing the [API review process](http://aka.ms/api-review)?

DavoudEshtehari (approved):

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.NetCoreApp.cs
cheenamalhotra: Well, that's legacy link came from dotnet/corefx, not applicable to this library anymore.

DavoudEshtehari: So, It's good if we remove it.

PR #1182, opened 2021-07-21
Fix bug with LegacyRowVersionNullBehavior

### LINKED ISSUE #1175: LegacyRowVersionNullBehavior causes InvalidCastException when using SqlDataReader.GetFieldValue[T]
### Describe the bug

I updated my project, which uses EF Core 5.0.8, to use the latest version SqlClient, 3.0.0, and got an InvalidCastException exception. Turning off `SqlServerDbContextOptionsBuilder.EnableRetryOnFailure()` prevents the error, but I need it on.

Investigating further, the issue seems to come from `SqlDataReader.GetFieldValue[T](Int32 i)`.

```
Exception message: Unable to cast object of type 'System.DBNull' to type 'System.Byte[]'.

Stack trace: 
System.InvalidCastException: Unable to cast object of type 'System.DBNull' to type 'System.Byte[]'.
   at Microsoft.Data.SqlClient.SqlDataReader.GetFieldValueFromSqlBufferInternal[T](SqlBuffer data, _SqlMetaData metaData)
   at Microsoft.Data.SqlClient.SqlDataReader.GetFieldValueInternal[T](Int32 i)
   at Microsoft.Data.SqlClient.SqlDataReader.GetFieldValue[T](Int32 i)
```

### To reproduce

I've recreated the issue in a repo: https://github.com/andygjp/NullRowVersionIssue

### Expected behavior
It should not throw an exception - the GetFieldValueFromSqlBufferInternal() function needs to have the same behaviour as TryReadColumnInternal(Int32, Bool), ie respect LocalAppContextSwitches.LegacyRowVersionNullBehavior

### Further technical details
```
dotnet --info
.NET SDK (reflecting any global.json):
 Version:   5.0.204
 Commit:    84d1fe1bb7

Runtime Environment:
 OS Name:     fedora
 OS Version:  34
 OS Platform: Linux
 RID:         fedora.34-x64
 Base Path:   /usr/lib64/dotnet/sdk/5.0.204/

Host (useful for support):
  Version: 5.0.7
  Commit:  556582d964

.NET SDKs installed:
  3.1.116 [/usr/lib64/dotnet/sdk]
  5.0.204 [/usr/lib64/dotnet/sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 3.1.16 [/usr/lib64/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 5.0.7 [/usr/lib64/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 3.1.16 [/usr/lib64/dotnet/shared/Microsoft.NETCore.App]
  Microsoft.NETCore.App 5.0.7 [/usr/lib64/dotnet/shared/Microsoft.NETCore.App]

```

cheenamalhotra:
Fixes #1175 as reported.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
@@ -2914,7 +2914,8 @@ namespace Microsoft.Data.SqlClient
                             object rawValue = GetValueFromSqlBufferInternal(data, metaData);
                             if (data.IsNull && !LocalAppContextSwitches.LegacyRowVersionNullBehavior)
                             {
-                                rawValue = DBNull.Value;
+                                // Return "null" if data is null and cannot be cast to DBNull (legacy support for GetFieldValue<byte[]>)
+                                rawValue = typeofT == typeof(DBNull) ? DBNull.Value : null;
                             }
                             return (T)rawValue;
                         }
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
@@ -2911,7 +2911,12 @@ namespace Microsoft.Data.SqlClient
                     {
                         try
                         {
-                            return (T)GetValueFromSqlBufferInternal(data, metaData);
+                            object rawValue = GetValueFromSqlBufferInternal(data, metaData);
+                            if (data.IsNull && !LocalAppContextSwitches.LegacyRowVersionNullBehavior)
+                            {
+                                rawValue = DBNull.Value;
+                            }
+                            return (T)rawValue;
                         }
                         catch (InvalidCastException)
                         {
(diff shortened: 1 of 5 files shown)


Wraith2:
Can you add the covering tests the user provided please.

Wraith2:
What should the new behaviour be for `GetFieldValue<byte[]>` on a null RowVersion value? It can't return an `INullable` but because byte[] can be null we could return a `null`.

cheenamalhotra:
> What should the new behaviour be for `GetFieldValue<byte[]>` on a null RowVersion value? It can't return an `INullable` but because byte[] can be null we could return a `null`.

Addressed that too.

Wraith2:
Ok, I'm fine with returning null for NULL there, just wondering singl ei was converting the new tests for the current behaviour to make sure it was well defined.

I think I prefer my goto fix code better but I'm fine with approving this version if you prefer it.

cheenamalhotra:
>Ok, I'm fine with returning null for NULL there, just wondering singl ei was converting the new tests for the current behaviour to make sure it was well defined.

I decided to revert this back to throw exception - as ideally this is "expected" with the breaking change we have when `LegacyRowVersionNullBehavior` is not enabled. As changing it to give null can cause other types to avoid exception too when cast cannot be done for null column value.

> I think I prefer my goto fix code better but I'm fine with approving this version if you prefer it.

I don't mind yours either, the one I have now is a less impactful and less complex to understand.. I'll leave it as is.

Wraith2:
My only complaint is that the goto version avoids checking the same variable twice. On each path you only check what is needed. In the if-based construction you have to access the legacy switch twice and use isnull twice. If you and the team prefer this version for code clarity that's ok though.

Wraith2 (approved):

JRahnama:
LGTM, plus I liked the approach to C# 9's new features.

JRahnama (approved):

JRahnama (approved):

DavoudEshtehari (commented):
Could you extend the tests and include `Timestamp` too?

cheenamalhotra:
> Could you extend the tests and include Timestamp too?

Timestamp is already included when we run this query in tests:
`"select cast(null as rowversion) rv";`

I'll take care of other improvements, I merely extended tests to cover new cases..

DavoudEshtehari (approved):

roji:
Should this be considered for backporting to 3.0.1?

ErikEJ:
@roji most important if 3.0 is LTS, I guess?

roji:
@ErikEJ I think it isn't though, no? Isn't 4.0 the LTS?

DavoudEshtehari:
This is one of the candidates for backporting to 3.0.1.

PR #1185, opened 2021-07-23
Tests | Adding Kerberos Testing with Integrated Authentication

JRahnama:
-p:KerberosBuild=true should be used when building tests.


cheenamalhotra (approved):
Nit: Small change requested, rest looks good.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/KerberosTests/KerberosTicketManager/KerberosTicketManager.cs
```diff
@@ -0,0 +1,48 @@
...
+                        FileName = s_cmdPrompt,
+                        Arguments = isInit? $"-c {command}" : $"-c {command} -p:{DataTestUtility.KerberosDomainPassword}"
+                    }
+                };
+                proc.Start();
+            }
+            catch (Exception)
+            {
+                throw;
+            }
```
cheenamalhotra, line 45: Nit: Maybe unnecessary?

cheenamalhotra (approved):

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/KerberosTests/KerberosTest.cs
```diff
@@ -0,0 +1,61 @@
...
+    public class KerberosTests
+    {
+        [PlatformSpecific(TestPlatforms.AnyUnix)]
+        [ConditionalTheory(typeof(DataTestUtility), nameof(DataTestUtility.IsKerberosTest))]
+        [ClassData(typeof(ConnectionStringsProvider))]
+        public void IsKerBerosSetupTest(string connection)
+        {
+            try
+            {
+                Task t = Task.Run(() => KerberosTicketManagemnt.Init(DataTestUtility.KerberosDomainUser)).ContinueWith((i) =>
```
johnnypham, line 22: Can we use async/await here?

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/KerberosTests/KerberosTest.cs
```diff
@@ -0,0 +1,61 @@
...
+                    {
+                        conn.Open();
+                        using var command = new SqlCommand("SELECT auth_scheme from sys.dm_exec_connections where session_id = @@spid", conn);
+                        using SqlDataReader reader = command.ExecuteReader();
+                        while (reader.Read())
+                        {
+                            Assert.Equal("KERBEROS", reader.GetString(0));
+                        }
+                    }
+                    catch (SqlException ex)
```
johnnypham, line 35: Both of these try/catch blocks are unnecessary. If an exception is unhandled, I think xunit always prints the exception.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
```diff
@@ -73,6 +74,11 @@ public static class DataTestUtility
         private static Dictionary<string, bool> AvailableDatabases;
         private static BaseEventListener TraceListener;
 
+        //Kerberos variables
+        public static readonly string KerberosDomainUser = null;
+        internal static readonly string KerberosDomainPassword = null;
+        public static readonly bool IsKerberos = false;
```
johnnypham, line 80: This public field and the public property below are redundant. We could encapsulate the logic in a single property like: 
`public static bool IsKerberos => !string.IsNullOrEmpty(KerberosDomainUser) && !string.IsNullOrEmpty(KerberosDomainPassword);`

Also, DataTestUtility.IsKerberos doesn't really make a lot of sense to me.

johnnypham (approved):

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/KerberosTests/KerberosTest.cs
JRahnama: Since I needed the first step to run and complete before next step, the best option would be to go in synchronously from beginning. Changed accordingly.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/KerberosTests/KerberosTest.cs
JRahnama: done.

PR #1186, opened 2021-07-24
Simplify managed SNI receive callback use

Wraith2:
I investigated why there seem to be two ways to set the callbacks which are executed on async reads and writes to packets. It looks like the original method was to pass the callback as a parameter to the function and then later because of the need to support MARS there needed to be a way to store set the callback before that call so it could override the parameter which was made nullable.

This PR removes the ability to pass the callback as part of the call and requires that it always be set by calling SetReceiveCallback. Doing this removes the need to pass the callback as the state parameter to the async call meaning that the packet can be used as state instead which allows the callback to be static because everything needed to finish the call is now part of the packet itself.


Wraith2:
Freshly rebased and the real CI is green, all ready to go.

David-Engel (approved):
LGTM

DavoudEshtehari (approved):
Thank you, It's pretty much cleaner. Just some unnecessary usings in `SNITcpHandle`, `SNIpacket`, and `SNINpHandle` should be removed.

Wraith2:
Usings cleaned up in the files mentioned.

PR #1188, opened 2021-07-25
Simplify event tracing

Wraith2:
This replaces a lot of try finally blocks used to implement log traces with using blocks that do the same. This makes the code visually cleaner. There are a couple of minor style changes scattered around as well but there are no functional changes.


johnnypham (approved):

cheenamalhotra (approved):
This is very helpful @Wraith2 ❤️

PR #1194, opened 2021-07-28
Perf: Cache delegates for non query and xml reader calls

Wraith2:
When using `ExecuteNonQueryAsync` and `ExecuteXmlReaderAsync` begin and end delegates are passed to `Task<T>.Factory.FromApm` and because these are instance methods they cannot be cached meaning we have two delegates allocated per call. ExecuteNonQueryAsync is a _very_ commonly used method so even small improvements here can have a large impact overall. 

I followed the general pattern introduced and used in ExecuteReaderAsync. I created static versions of the begin and end functions needed for each of the FromApm calls and made sure that the instance was passed as a state variable, the static functions simply unpack the instance from state and call the correct method on the instance. The static delegates are then cached as static readonly and re-used every time a call is made. 

Note that the netfx version of ExecuteReaderAsync has not been altered because it will be dealt with by https://github.com/dotnet/SqlClient/pull/1084


Wraith2:
I'm closing this because there is a cleaner and lower impact way to do this once https://github.com/dotnet/SqlClient/pull/1084 has been merged.

PR #1202, opened 2021-08-03
Combine: Move SqlCollation to shared, make readonly and cache last used

Wraith2:
Investigating perf on ExecuteNonQuery i saw that collation is being sent from the server and an SqlCollation being created each time. This can be avoided. 

This PR extracts SqlCollection from the file it was in. Moves it to it's own shared file. Makes it readonly so that it can be cached safely without risk of mutation and changes the methods of construction to allow that readonly change. The creation of the collation is now checked against the last used collation and if they are identical they previously created instance is used.


johnnypham (approved):

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlParameter.cs
```diff
@@ -413,7 +409,11 @@ public SqlCompareOptions CompareInfo
                 {
                     throw ADP.ArgumentOutOfRange(nameof(CompareInfo));
                 }
-                collation.SqlCompareOptions = value;
+
+                if (collation == null || collation.SqlCompareOptions != value)
+                {
+                    _collation = SqlCollation.FromLCIDAndSort(collation?.LCID ?? 0, value);
+                }
```
JRahnama, line 416: how would this behave in async/concurrent calls? If I am not wrong, a new member of `SqlParameter` would be created not `SqlCollation` and we just do the clean up by using `FromLCIDAndSort` for sync calls. Am I correct?

JRahnama: > Does that answer? I admit I don't fully understand the question.

Thanks it did answer my question. Apparently I missed a part of the code and I got myself confused as well.

JRahnama (approved):

PR #1210, opened 2021-08-13
Update `Encrypt` property default value to true

DavoudEshtehari:
- **Encrypt** default value = `true`.
- Improving the pre-login process from a client without encryption possibility ([TDS Pre-login](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-tds/60f56408-0188-4cd5-8b90-25c6f2423868)):
  - Supported values using Native SNI:      **ENCRYPT_OFF**, **ENCRYPT_ON**, and **ENCRYPT_NOT_SUP**
  - Supported values using Managed SNI: **ENCRYPT_OFF**, and **ENCRYPT_ON****

<html>
<body>

Value sent by client | Value returned by server when server is set to ENCRYPT_OFF | Value returned by server when server is set to ENCRYPT_ON | Value returned by server when server is set to ENCRYPT_NOT_SUP
-- | -- | -- | --
ENCRYPT_OFF | ENCRYPT_OFF | ENCRYPT_REQ | ENCRYPT_NOT_SUP
ENCRYPT_ON | ENCRYPT_ON | ENCRYPT_ON | ENCRYPT_NOT_SUP (connection terminated)
ENCRYPT_NOT_SUP | ENCRYPT_NOT_SUP | ENCRYPT_REQ (connection terminated) | ENCRYPT_NOT_SUP

</body>
</html>

> [Complementary explanation](https://github.com/dotnet/SqlClient/pull/1210#issuecomment-898658021)


ErikEJ:
What issue is this solving?

roji:
Doesn't this require the server to have proper SSL certificates? If so, that seems quite breaking for everyone currently happily not using SSL in internal networks...

Wraith2:
It depends on the value of TrustServerCertificate see https://docs.microsoft.com/en-us/sql/relational-databases/native-client/features/using-encryption-without-validation?view=sql-server-ver15 by default default is to trust the server so default configuration self signed will still allow access. EnableEncryption just forces the use of ssl and doesn't have implications for client or server certificate validation.

DavoudEshtehari:
> Doesn't this require the server to have proper SSL certificates? If so, that seems quite breaking for everyone currently happily not using SSL in internal networks...

This is all about the client environment configuration and possible security breaches. Also, the customers who enforce the encryption on the server don't face this breach.

roji:
> by default default is to trust the server so default configuration self signed will still allow access

Are you sure about that? Both that table and checking the default value of SqlConnectionStringBuilder.TrustServerCertificate seem to indicate that self-signed certificates aren't trusted by default, and one has to opt in explicitly by setting `Trust Server Certificate=true`. This makes sense, since by default trusting certificates which cannot be validated wouldn't be very secure.

DavoudEshtehari:
> It depends on the value of TrustServerCertificate see https://docs.microsoft.com/en-us/sql/relational-databases/native-client/features/using-encryption-without-validation?view=sql-server-ver15 by default default is to trust the server so default configuration self signed will still allow access. EnableEncryption just forces the use of ssl and doesn't have implications for client or server certificate validation.

Any connection to the server at least needs a secure login that means both sides of the connection should be possible to encrypt. So, If one side wasn't able to encrypt, establishing a secure connection shouldn't be possible.

Regarding the use of `TrustServerCertificate`, by attempting to open a connection with `Encrypt=true` and `TrustServerCertificate=false` you will get an exception which is expected, while if you try it just after disabling all secure protocols on the client it'll connect except using Managed SNI.

DavoudEshtehari:
> What issue is this solving?

[Comment's ](https://github.com/dotnet/SqlClient/pull/1210#issue-711919185) updated.

Wraith2:
> > What issue is this solving?
> 
> [Comment's ](https://github.com/dotnet/SqlClient/pull/1210#issue-711919185) updated.

Ok. but I think the point was to ask why the change is being made. Is it a top down decision to require encryption at all times?

DavoudEshtehari:
> the point was to ask why the change is being made.

It includes two changes:

1. Replacing the Encrypt default value by true in the connection string.
2. Improving the pre-login process to avoid the probable security breach as it's discussed above.

> Is it a top down decision to require encryption at all times?

If you asked for Encrypt default value, the quick answer is yes. But if you asked for the encryption process, I delegate it to @David-Engel.

David-Engel:
There are two issues being addressed here:
1. Default Encrypt to True. This is for security. Similar to http/https, if the client starts with allowing non-encrypted connections, it will always be susceptible to MITM attacks. Even if the server is configured to require encryption, there can be a MITM altering the server's response to say it doesn't require encryption. The MITM can then proxy the connection. `client <-plain text-> MITM <-encrypted-> server = the connection is compromised`.

    Security has been encouraging us for years to change the default behavior of client drivers to be secure by default and we have resisted, knowing that it is a breaking change for most users. It's easy enough for devs to add Encrypt = false to all their connection strings, if they need to. We just want to make sure they understand the choice they are making and they are making it deliberately. With cloud computing becoming more and more common, it's not safe to leave the default value of Encrypt equal to false.

2. The less-breaking, but still important, fix here is to ensure connections fail if the client does not have any encryption libraries available and either Encrypt = true or the server requires encryption. SqlClient + native SNI is the only MS driver we've found that will successfully connect in that scenario.

ErikEJ:
Thanks, makes sense doing this for version 4!

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -464,6 +469,11 @@ internal void ProcessPendingAck(TdsParserStateObject stateObj)
                 _physicalStateObj.AssignPendingDNSInfo(serverInfo.UserProtocol, FQDNforDNSCahce, ref _connHandler.pendingSQLDNSObject);
             }
 
+            if (encrypt && !EncryptClientPossible)
+            {
+                _encryptionOption = EncryptionOptions.NOT_SUP;
```
cheenamalhotra, line 474: 1. _encryptionOption should be set to NOT_SUP even when "encrypt" is false.
2. If "encrypt" is true instead of setting the option value to NOT_SUP, can you throw below error directly and terminate?
https://github.com/dotnet/SqlClient/pull/1210/files#diff-9468409fbd86dc6afecde79621ca15d128de0597cb0d36ce8464b469bc62a5b8R690

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -674,8 +684,13 @@ private void SendPreLoginHandshake(byte[] instanceName, bool encrypt)
                     case (int)PreLoginOptions.ENCRYPT:
                         if (_encryptionOption == EncryptionOptions.NOT_SUP)
```
cheenamalhotra, line 685: This if condition could be avoided if error is handled for this usecase beforehand.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -877,8 +892,11 @@ private PreLoginHandshakeStatus ConsumePreLoginHandshake(bool encrypt, bool trus
                             case (EncryptionOptions.OFF):
                                 if (serverOption == EncryptionOptions.OFF)
                                 {
-                                    // Only encrypt login.
-                                    _encryptionOption = EncryptionOptions.LOGIN;
+                                    if (EncryptClientPossible)
```
cheenamalhotra, line 895: Is there a use case this change is solving?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserSafeHandles.cs
```diff
@@ -18,7 +18,8 @@ internal sealed partial class SNILoadHandle : SafeHandle
         internal readonly SNINativeMethodWrapper.SqlAsyncCallbackDelegate WriteAsyncCallbackDispatcher = new SNINativeMethodWrapper.SqlAsyncCallbackDelegate(WriteDispatcher);
 
         private readonly uint _sniStatus = TdsEnums.SNI_UNINITIALIZED;
-        private readonly EncryptionOptions _encryptionOption;
+        private readonly EncryptionOptions _encryptionOption = EncryptionOptions.OFF;
+        private bool? _encryptClientPossible = null;
```
cheenamalhotra, line 22: Can we rename this to "_clientOSEncryptionSupport" instead?
Similarly "ClientOSEncryptionSupport" instead of "EncryptClientPossible"

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
DavoudEshtehari: If the idea of throwing an exception even by encrypt=false is approved we can raise the exception here as you mentioned.

cheenamalhotra: No I meant:

```
if (!EncryptClientPossible)
{
    if(encrypt)
    {
        throw exception;
    }
    _encryptionOption = EncryptionOptions.NOT_SUP;
}
```

DavoudEshtehari: If we're good by throwing an exception on either encrypt is `false` or `true` with the message "The instance of SQL Server you attempted to connect to requires encryption but this machine does not support it.", it's possible to handle those conditions here.

cheenamalhotra: As discussed offline, please update design to respect specifications for encryption options supported by driver: https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-tds/60f56408-0188-4cd5-8b90-25c6f2423868

JRahnama:
LGTM. I wait for other teammate opinion.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -464,6 +469,18 @@ internal void ProcessPendingAck(TdsParserStateObject stateObj)
                 _physicalStateObj.AssignPendingDNSInfo(serverInfo.UserProtocol, FQDNforDNSCahce, ref _connHandler.pendingSQLDNSObject);
             }
 
+            if (!ClientOSEncryptionSupport)
+            {
+                //If encryption is required, an error will throw.
```
cheenamalhotra, line 474:
```suggestion
                //If encryption is required, an error will be thrown.
```

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -644,6 +648,18 @@ internal void ProcessPendingAck(TdsParserStateObject stateObj)
             // for DNS Caching phase 1
             AssignPendingDNSInfo(serverInfo.UserProtocol, FQDNforDNSCahce);
 
+            if(!ClientOSEncryptionSupport)
+            {
+                //If encryption is required, an error will throw.
```
cheenamalhotra, line 653:
```suggestion
                //If encryption is required, an error will be thrown.
```

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -1209,11 +1225,12 @@ private void SendPreLoginHandshake(byte[] instanceName, bool encrypt, bool clien
...
                                     _encryptionOption = EncryptionOptions.ON | (_encryptionOption & ~EncryptionOptions.OPTIONS_MASK);
                                 }
-
+                                // NOT_SUP: No encryption.
                                 break;
 
                             case (EncryptionOptions.NOT_SUP):
-                                if ((serverOption & EncryptionOptions.OPTIONS_MASK) == EncryptionOptions.REQ)
+                                // Server required encryption
+                                if ((serverOption & EncryptionOptions.OPTIONS_MASK) != EncryptionOptions.NOT_SUP)
```
cheenamalhotra, line 1233: This is technically same as `if ((serverOption & EncryptionOptions.OPTIONS_MASK) == EncryptionOptions.REQ)` as these are the only 2 options supported by server when client sends `EncryptionOptions.NOT_SUP`

image

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -1189,7 +1205,7 @@ private void SendPreLoginHandshake(byte[] instanceName, bool encrypt, bool clien
                         switch (_encryptionOption & EncryptionOptions.OPTIONS_MASK)
                         {
                             case (EncryptionOptions.ON):
-                                if (serverOption == EncryptionOptions.NOT_SUP)
+                                if ((serverOption & EncryptionOptions.OPTIONS_MASK) != EncryptionOptions.ON)
```
cheenamalhotra, line 1208: Here too I feel it's same as before as there are only 2 possibilities here: ON or NOT_SUP, condition can be left untouched. Options Mask case can be taken care when it's enabled, need not be modified in this PR.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -983,7 +999,7 @@ private void SendPreLoginHandshake(byte[] instanceName, bool encrypt, bool clien
                     case (int)PreLoginOptions.ENCRYPT:
                         if (_encryptionOption == EncryptionOptions.NOT_SUP)
                         {
-                            // If OS doesn't support encryption, inform server not supported.
+                            //If encryption is not required.
```
cheenamalhotra, line 1002:
```suggestion
                            //If OS doesn't support encryption and encryption is not required, inform server "not supported" by client.
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -885,11 +902,12 @@ private PreLoginHandshakeStatus ConsumePreLoginHandshake(bool encrypt, bool trus
...
                                     _encryptionOption = EncryptionOptions.ON;
                                 }
-
+                                // NOT_SUP: No encryption.
                                 break;
 
                             case (EncryptionOptions.NOT_SUP):
-                                if (serverOption == EncryptionOptions.REQ)
+                                // Server required encryption
+                                if (serverOption != EncryptionOptions.NOT_SUP)
```
cheenamalhotra, line 910: Same here, can be left untouched.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -865,7 +882,7 @@ private PreLoginHandshakeStatus ConsumePreLoginHandshake(bool encrypt, bool trus
                         switch (_encryptionOption)
                         {
                             case (EncryptionOptions.ON):
-                                if (serverOption == EncryptionOptions.NOT_SUP)
+                                if (serverOption != EncryptionOptions.ON)
```
cheenamalhotra, line 885: Same here, can be left untouched.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -674,7 +691,7 @@ private void SendPreLoginHandshake(byte[] instanceName, bool encrypt)
                     case (int)PreLoginOptions.ENCRYPT:
                         if (_encryptionOption == EncryptionOptions.NOT_SUP)
                         {
-                            // If OS doesn't support encryption, inform server not supported.
+                            //If encryption is not required.
```
cheenamalhotra, line 694:
```suggestion
                            //If OS doesn't support encryption and encryption is not required, inform server "not supported" by client.
```

cheenamalhotra (commented):
LGTM, just some minor suggestions.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
DavoudEshtehari: Correct, just my point was it could be more descriptive, which means we expect to see `NOT_SUP` here. If you still prefer the old condition I can revert it.
(Same explanation for the similar comments.)

cheenamalhotra: I would go by old, as it clearly defines "when" to terminate connection.

Malcolm-Stewart:
Someone mentioned that TrustServerCertificate defaults to true. That is incorrect, unless that was also changed for this driver.
When Encrypt=true, then the certificate is also validated.
This means that changing the default will break many scenarios.
With Encrypt=False, SSL/TLS is still used to encrypt the login packet, but the certificate is not validated.
With Encrypt=true, SSL/TLS is used to encrypt the login packet and all subsequent packets, but the certificate is validated first.
Systems using SQL Server's self-generated certificate will fail after this change and the use will have to buy a certificate or create a self-signed certificate.

David-Engel:
> Someone mentioned that TrustServerCertificate defaults to true. That is incorrect, unless that was also changed for this driver.

I suspect they misspoke. I confuse TrustServerCertificate all the time, thinking of it as "validate server certificate" in my head.

> When Encrypt=true, then the certificate is also validated.
> This means that changing the default will break many scenarios.
> With Encrypt=False, SSL/TLS is still used to encrypt the login packet, but the certificate is not validated.

Correct. That will still be the case.

> With Encrypt=true, SSL/TLS is used to encrypt the login packet and all subsequent packets, but the certificate is validated first.

With the default TrustServerCertificate value, correct.

> Systems using SQL Server's self-generated certificate will fail after this change and the user will have to buy a certificate or create a self-signed certificate, or change their connection string. If the change was made for scenarios where the user could not modify the connection string to enable encryption, then they are left having to get new certificates.

Given that applications must be on Microsoft.Data.SqlClient to get this change, that means the application is almost certainly under current development. I don't think there is high risk that someone will be blocked by not being able to change the Encrypt setting for an application that is under current development. If their application is truly stuck, they will likely be on System.Data.SqlClient or an earlier version of MDS forever. Additionally, if this argument continues to keep us from making the change, it will never be made and security will never improve. We do understand it may be disruptive, but security has been pushing for us to do this for years.

Malcolm-Stewart:
Okay, so if the change is only for Microsoft.Data.SqlClient, then that is understandable. If System.Data and ODBC/OLE DB are unchanged, then I do not have any objection to that.
I was just worried that legacy apps might get caught up in the change.

cheenamalhotra (approved):

David-Engel (approved):

davidsampson-hv:
This caused downtime for our production Logic Apps talking to SQL as we are using signed SSL certs on SQL and hadn't declared trustservercertificates. I am 100% onboard with the change I just wish this had been communicated to users of Logic Apps for example. We have no ability/visibility to track potential changes like this via the underlying Logic Apps stack, we rely on Microsoft to surface and communicate breaking changes in advance, it's clear from this issue that this was known to be a breaking change before it was merged.

bodzilla:
> This caused downtime for our production Logic Apps talking to SQL as we are using self-signed SSL certs on SQL and hadn't declared trustservercertificates. I am 100% onboard with the change I just wish this had been communicated to users of Logic Apps for example. We have no ability/visibility to track potential changes like this via the underlying Logic Apps stack, we rely on Microsoft to surface and communicate breaking changes in advance, it's clear from this issue that this was known to be a breaking change before it was merged.
> 
> I assume as this was merged a year ago but we only experienced the change a couple of weeks ago that Logic Apps have only just updated to this version of dotnet.

We have experienced this same type of issue during testing of our own apps which utilise the Serilog MSSQL sink - had it not been for the fact that their changelog included the breaking change link for this lib we would've been in a bit of a pickle, so appreciate both ends declaring this - although it perhaps highlights the need for our end to look deeper into checking dependencies of dependencies before pushing something out for testing.

xorinzor:
It would've been nice if an example was included in the documentation on how to revert to the original behaviour.

Kalakaarboyz1:
[Update Encrypt property default value to true](https://github.com/dotnet/SqlClient/pull/1210#top)

greengumby:
Using SqlClient V4 I am expecting the connection string to require `TrustServerCertificate=true` however when using a VM hosted in the Azure Portal connecting to an Azure SQL Database this is not the case.  Any ideas why?

Malcolm-Stewart:
TrustServerCertificate is not required if the server's certificate is implicitly trusted by the client, e.g. it is issued by a certificate authority that has a root certificate already installed in the trusted root store of the client.

PR #1223, opened 2021-08-24
Move unused files causing heavier "AnyOS" binaries

cheenamalhotra:


Wraith2:
`AAsyncCallContext` is surprising but if it builds it's fine.

cheenamalhotra:
The "AnyOS" binaries we build for lib\netcore and lib\netstandard don't contain any source code, they just throw "Not Supported Exception". That's why anything added to those binaries (via default itemgroup without conditions) is going to only add up to size and has no use.

Any new file (containing source-code) should be always added to an ItemGroup with a "non-AnyOS" condition in this case.
When you add a new file directly via Visual Studio it won't recognize the right conditions, manual intervention is needed when adding new files.

DavoudEshtehari (approved):

JRahnama (approved):

Wraith2:
Good to know. I'll try to be more careful in future.

PR #1226, opened 2021-08-24
Fix regression in CancelAsyncConnections test using Native SNI

cheenamalhotra:
A change in PR #933 caused `CancelAsyncConnections` test to fail for Native SNI. Using _Debug_ configuration locally helped reproduce the failing assertion consistently. This test did not fail in public CI pipelines but it failed today in our internal build, which is surprising and we're looking into it.

@Wraith2 
I'm patching the packet release code block to be applicable to only Managed SNI so we can proceed to release, but it should be investigated in Managed SNI layer as there are many debug assertion failures that indicate towards design issues when running the `CancelAsyncConnections` test.


Wraith2:
Ok. Seems reasonable. I noted this location in the original PR as a point of interest.
As we've discussed before we really need to work towards running tests in Debug mode so we can see these sorts of fault.

DavoudEshtehari (approved):

JRahnama (approved):

PR #1237, opened 2021-08-30
FIX | Adding support for sharedInstances to managed SNI

### LINKED ISSUE #1231: LocalDB shared instance not supported
Connection to a shared instance of localdb fails. Shared instances use names in format '.\\*name*'. Hence, a connection string contains a datasource like '(localdb)\\.\\*name*' with **two** backslashes. 

But [here](https://github.com/dotnet/SqlClient/blob/2c5fbdde41f6d2adc9b694e822ec319eb88b3552/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIProxy.cs#L481) the assumption is made, that splitting the datasource value by backslash yields **exactly two** tokens.

```c#
string[] tokensByBackSlash = workingDataSource.Split(BackSlashCharacter);

error = false;

// All LocalDb endpoints are of the format host\instancename where host is always (LocalDb) (case-insensitive)
if (tokensByBackSlash.Length == 2 && LocalDbHost.Equals(tokensByBackSlash[0].TrimStart()))
{
```

### Exception message
```
Cannot connect to SQL Server Browser. Ensure SQL Server Browser has been started.
```
The exception message is a bit misleading. The initial error is, that the datasource name is not recognized as a LocalDB instance.

### Stack trace
```
System.Exception: Cannot connect to SQL Server Browser. Ensure SQL Server Browser has been started.
       ---> System.Net.Sockets.SocketException (11001): No such host is known.
         at System.Net.Dns.GetHostEntryOrAddressesCore(String hostName, Boolean justAddresses)
         at System.Net.Dns.GetHostAddresses(String hostNameOrAddress)
         at System.Net.Sockets.UdpClient.GetEndpoint(String hostname, Int32 port)
         at System.Net.Sockets.UdpClient.SendAsync(Byte[] datagram, Int32 bytes, String hostname, Int32 port)
         at Microsoft.Data.SqlClient.SNI.SSRP.SendUDPRequest(String browserHostname, Int32 port, Byte[] requestPacket)
         at Microsoft.Data.SqlClient.SNI.SSRP.GetPortByInstanceName(String browserHostName, String instanceName)
```

### To reproduce

Try to connect to a shared instance. From what I saw, it does not even have to exist to let you reproduce the error. Just use a connection string with the '.\\*name*' syntax:

```c#
var _connectionString = "Data Source=(localdb)\\.\\MyLocalDB_Shared;Initial Catalog=MyTable;Integrated Security=True;Connect Timeout=30;Encrypt=False;TrustServerCertificate=False;ApplicationIntent=ReadWrite;MultiSubnetFailover=False;"

var connection = new SqlConnection(_connectionString);
await connection.OpenAsync(cancellationToken).ConfigureAwait(false);
```

### Expected behavior

The  data source should be recognized as LocalDB instance.

### Further technical details
Microsoft.Data.SqlClient version: 3.0.0
.NET target: 5.0
SQL Server version: 15.0.4153 with cumulative update SQLServer2019-KB5004524-x64.exe installed
Operating system: Windows 10 Pro, version 2004, build 19041.1165

JRahnama:
fixes #1231 

Manged SNI did not have sharedInstances support for lcoaldb.

1. in SNI Proxy for `localdb` the assumption is that when we split the connection string by `\\` it always returns an array with length of 2 however when the connection string is set for a shared instance we get an array with a length of 3 since it is in a format of `(localdb)\.\shared Instance name`

Based on [MS Docs](https://docs.microsoft.com/en-us/sql/database-engine/configure-windows/sql-server-express-localdb?view=sql-server-ver15#connect-to-a-shared-instance-of-localdb) all shared instances connection strings should be in a format of 
`(localdb)\.\ "shared instance's name"
`
a shared instance could be created as below:
Make sure you have installed SqlLocalDB

1. Open Developer Command Prompt
2. Type: `SqlLocalDB create "" (you can choose any name)`
3. Type: `SqlLocalDB share " your instance name " "provide a name for shared instance"`
4. `SqlLocalDB start "name your instance"`
5. Continue either on SSMS or CLI with SqlCmd: for SSMS use (localdb)\your instancename
6. `Create LOGIN [Domain\User];` (run this if the login is not created by default for the user otherwise it fails) 
7. `GRANT LOGIN TO [Domain\User]`

in the same format one can unshare and delete a shared instance.

One can log into shared instances using ssms with the format mentioned above. Also, Visual studio provides the connection string when you go to View > SQL Server Object Explorer and select local. (It only shows localdb and shared isntances if you have already created one and it is running).

Added: `(localdb)\.` will login the user into the default `localdb`.


JRahnama:
> Regex is much faster in 5 but this is a really trivial case and what you really want is just to split the string and then count the number of resulting strings. Regex isn't needed or much easier here.

Ok. seems fair. It was an over kill. I've changed it.

Wraith2:
Is `(localdb)\LocalDbAppName` a valid name? It looks like it should be from your test input but if I replicate the function `GetLocalDBInstance` it returns false, perhaps you wanted to negate that `if (string.IsNullOrEmpty(..))` check?

Wraith2:
Is case important in the output instanceName? because if it isn't I think I've got a version of the function which only allocates the output string. If i have to lower case it then I'll need to allocate a stringbuilder.

Wraith2:
Have a look, see what you think: https://gist.github.com/Wraith2/b2dbcf6af7d21c48620dc00eb0879364

JRahnama:
> Have a look, see what you think: https://gist.github.com/Wraith2/b2dbcf6af7d21c48620dc00eb0879364

I like the idea, but `System.Range` is not supported in `NetCoreApp 2.1`. Although the support is dropped for that version, but it will take some time for us to get there.

I was thinking about this, what is preventing us from getting the very last index of the array?
We can get `instanceName = input[^1]` or `instanceName = tokensByBackSlash[tokensByBackSlash.Length-1]`

Wraith2:
You can simply replace the range expressions with Slice, they're all start offset only so the single arg version works.

We can get the last index if the split works but then what happens if a user passes us a string with 4 slashes in? Invalid yes but will it behave the same? More importantly I was trying to avoid the allocation that the split causes, it's an array and 3 strings all of them on the heap and we know that 2 of the strings are going to get thrown away instantly. 

The split version is fine if you want to stick with that. It's not a hot path so it's not going to cause anyone a problem. I just like to see if there are efficient ways to implement things like this, as I did with the azure endpoint parsing.

JRahnama:
@Wraith2, FYI ReadOnlySpan is not supported in `netstandard 2.0`, but installing `System.Memory` solves the issue.
for `ReadOnlySpan.StartsWith` Netstandard2.0 does not accept an `string` type as an argument had to be `ReadOnlySpan<char>` type. All done. Thanks for the support.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIProxy.cs
```diff
@@ -474,34 +474,41 @@ private void PopulateProtocol()
...
+                    }
+                }
+                if (!input.IsEmpty)
+                {
+                    instanceName = input.Trim().ToString();
                 }
                 else
                 {
-                    SNILoadHandle.SingletonInstance.LastError = new SNIError(SNIProviders.INVALID_PROV, 0, SNICommon.LocalDBNoInstanceName, Strings.SNI_ERROR_51);
+                    // set the sni error here
```
Wraith2, line 510: You'll want to put the SNI error back in here 😀

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
```diff
@@ -443,7 +452,7 @@ public static void DropDatabase(SqlConnection sqlConnection, string dbName)
             }
         }
 
-        public static bool IsLocalDBInstalled() => !string.IsNullOrEmpty(LocalDbAppName?.Trim());
+        public static bool IsLocalDBInstalled() => !string.IsNullOrEmpty(LocalDbAppName?.Trim()) && !string.IsNullOrEmpty(LocalDbSharedInstanceName?.Trim());
```
cheenamalhotra, line 455: Could you split this check into 2?
- IsLocalDBInstalled
- IsSharedLocalDBInstalled

So we could still run local db tests without shared local db setup available.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/LocalDBTest/LocalDBTest.cs
```diff
@@ -11,47 +11,55 @@ public static class LocalDBTest
...
+        [ConditionalTheory(nameof(IsLocalDBEnvironmentSet))]
+        [MemberData(nameof(LocalDbSourceProvider))]
+        public static void LocalDBConnectionTest(string connectionString)
         {
-            SqlConnectionStringBuilder builder = new SqlConnectionStringBuilder(@$"server=(localdb)\{DataTestUtility.LocalDbAppName}");
-            builder.IntegratedSecurity = true;
-            builder.ConnectTimeout = 2;
+            SqlConnectionStringBuilder builder = new(connectionString)
+            {
+                IntegratedSecurity = true,
```
cheenamalhotra, line 20: This test should only be run if "SupportsIntegratedSecurity" is true.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/LocalDBTest/LocalDBTest.cs
```diff
@@ -11,47 +11,55 @@ public static class LocalDBTest
...
+        [MemberData(nameof(LocalDbSourceProvider))]
+        public static void LocalDBMarsTest(string connectionString)
         {
-            SqlConnectionStringBuilder builder = new SqlConnectionStringBuilder(@$"server=(localdb)\{DataTestUtility.LocalDbAppName}");
-            builder.IntegratedSecurity = true;
-            builder.MultipleActiveResultSets = true;
-            builder.ConnectTimeout = 2;
+            SqlConnectionStringBuilder builder = new SqlConnectionStringBuilder(connectionString)
+            {
+                IntegratedSecurity = true,
```
cheenamalhotra, line 34: Also this one.

src/Microsoft.Data.SqlClient/tests/tools/Microsoft.Data.SqlClient.TestUtilities/Config.cs
```diff
@@ -24,6 +24,7 @@ public class Config
         public string AzureKeyVaultClientId = null;
         public string AzureKeyVaultClientSecret = null;
         public string LocalDbAppName = null;
+        public string LocalDbSharedInstanceName = null;
```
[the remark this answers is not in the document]
JRahnama: done.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
JRahnama: done.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/LocalDBTest/LocalDBTest.cs
JRahnama: done.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/LocalDBTest/LocalDBTest.cs
JRahnama: done.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
```diff
@@ -150,6 +153,12 @@ static DataTestUtility()
                     AEConnStringsSetup.Add(TCPConnectionString);
                 }
             }
+            if (!string.IsNullOrEmpty(LocalDbAppName))
+            {
+                LocalDbDataSources.Add(@$"server=(localdb)\{LocalDbAppName}");
+                LocalDbDataSources.Add(@$"server=(localdb)\.\{LocalDbSharedInstanceName}");
```
cheenamalhotra: Please check for null LocalDbSharedInstanceName before adding this one.

DavoudEshtehari (approved):

cheenamalhotra (approved):

BUILDGUIDE.md
```diff
@@ -118,6 +118,7 @@ Manual Tests require the below setup to run:
   |AzureKeyVaultClientId | (Optional) "Application (client) ID" of an Active Directory registered application, granted access to the Azure Key Vault specified in `AZURE_KEY_VAULT_URL`. Requires the key permissions Get, List, Import, Decrypt, Encrypt, Unwrap, Wrap, Verify, and Sign. | _{Client Application ID}_ |
   |AzureKeyVaultClientSecret | (Optional) "Client Secret" of the Active Directory registered application, granted access to the Azure Key Vault specified in `AZURE_KEY_VAULT_URL` | _{Client Application Secret}_ |
   |LocalDbAppName | (Optional) If Local Db Testing is supported, this property configures the name of Local DB App instance available in client environment. Empty string value disables Local Db testing. | Name of Local Db App to connect to.|
+  |LocalDbSharedInstanceName | (Optional) If Local Db testing is supported and an Instance of that is shared, this property configures the name of Shared Local Db Instance to connect to. | Name of Shared Local Db Instance. |
```
johnnypham, line 121:
```suggestion
  |LocalDbSharedInstanceName | (Optional) If LocalDB testing is supported and the instance is shared, this property configures the name of the shared instance of LocalDB to connect to. | Name of shared instance of LocalDB. |
```

Nit: LocalDB is the official name so we should probably use that.

PR #1240, opened 2021-08-31
Feature | Support for SqlFileStream for .NET Standard 2.0+ on Windows

cheenamalhotra:
Also enabled tests with config file modifications.

Changes are tested with NETSTANDARD reference against both .NET Framework 4.6.1 and .NET Core 2.1:

image

_SqlFileStream stays unsupported on Unix as with .NET Core._


src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
```diff
@@ -152,6 +153,62 @@ static DataTestUtility()
...
 
+        public static string SetupFileStreamDB()
+        {
+            try
+            {
+                if (!IsFileStreamDbSetup())
+                {
+                    if (!FileStreamDirectory.EndsWith("\\"))
+                    {
+                        FileStreamDirectory += "\\\\";
```
David-Engel, line 164: Why the double `\`s on the end of the path?
```suggestion
                        FileStreamDirectory += "\\";
```

cheenamalhotra: I guess I escaped twice

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlCommandTest.cs
```diff
@@ -184,7 +184,7 @@ public void Constructor4()
             Assert.Null(cmd.Container);
             Assert.True(cmd.DesignTimeVisible);
             Assert.Null(cmd.Notification);
-#if NETFX
+#if NETFX && !NETSTANDARDREFERNCE
```
DavoudEshtehari, line 187: I guess line [169](https://github.com/dotnet/SqlClient/blob/4c63e9ecd1d2097047e6c2bb66f957fc16ff81e2/src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlCommandTest.cs#L169) should be updated too.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlCommandTest.cs
```diff
@@ -240,7 +240,7 @@ public void Clone()
             Assert.Null(cmd.Connection);
             Assert.False(cmd.DesignTimeVisible);
             Assert.Same(notificationReq, cmd.Notification);
-#if NETFX
+#if NETFX && !!NETSTANDARDREFERNCE
```
DavoudEshtehari, line 243:
```suggestion
#if NETFX && !NETSTANDARDREFERNCE
```

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
```diff
@@ -44,10 +44,11 @@ public static class DataTestUtility
         public static bool EnclaveEnabled { get; private set; } = false;
         public static readonly bool TracingEnabled = false;
         public static readonly bool SupportsIntegratedSecurity = false;
-        public static readonly bool SupportsFileStream = false;
         public static readonly bool UseManagedSNIOnWindows = false;
         public static readonly bool IsAzureSynapse = false;
         public static Uri AKVBaseUri = null;
+        public static string FileStreamDirectory = null;
+        private static string _fileStreamDBName = null;
```
DavoudEshtehari, line 51: I believe we could manage this value inside the related tests and taking this field out of here.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
```diff
@@ -152,6 +153,60 @@ static DataTestUtility()
...
+            }
+            catch (SqlException e)
+            {
+                Console.WriteLine("File Stream database could not be setup. " + e.Message);
+                FileStreamDirectory = null;
+            }
+            return _fileStreamDBName;
+        }
+
+        internal static void DropFileStreamDb()
```
DavoudEshtehari, line 194: Could you pass the required values as parameters here and move this function into the related test class?

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
```diff
@@ -44,10 +44,11 @@ public static class DataTestUtility
         public static bool EnclaveEnabled { get; private set; } = false;
         public static readonly bool TracingEnabled = false;
         public static readonly bool SupportsIntegratedSecurity = false;
-        public static readonly bool SupportsFileStream = false;
         public static readonly bool UseManagedSNIOnWindows = false;
         public static readonly bool IsAzureSynapse = false;
         public static Uri AKVBaseUri = null;
+        public static string FileStreamDirectory = null;
```
DavoudEshtehari, line 50: NIT: I suggest using property instead of a field here to verify its value.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
```diff
@@ -152,6 +153,60 @@ static DataTestUtility()
             }
         }
 
+        public static string SetupFileStreamDB()
```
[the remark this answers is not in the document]
cheenamalhotra: Since this method is called from outside classes, passing those would mean code like this:

```cs
DataTestUtility.SetupFileStreamDB(DataTestUtility.FileStreamDirectory, DataTestUtility.TCPConnectionString)
```

I prefer accessing local properties directly.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
cheenamalhotra: I added it here so we create only 1 such DB at a time.

cheenamalhotra: I've modified the design to take this to test level.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
cheenamalhotra: This value is tied to config.json, I didn't want to modify it a lot and keep design consistent with current properties.

DavoudEshtehari (approved):
👍

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlTypes/SqlFileStream.Windows.cs
```diff
@@ -689,8 +689,13 @@ static private string InitializeNtPath(string path)
...
             string uniqueId = Guid.NewGuid().ToString("N");
-            return System.IO.PathInternal.IsDeviceUNC(path) ? string.Format(CultureInfo.InvariantCulture, @"{0}\{1}", path.Replace(@"\\.", @"\??"), uniqueId)
-                                                            : string.Format(CultureInfo.InvariantCulture, @"\??\UNC\{0}\{1}", path.Trim('\\'), uniqueId);
+#if NETSTANDARD
+            return System.IO.PathInternal.IsDeviceUNC(path.AsSpan())
+#else
+            return System.IO.PathInternal.IsDeviceUNC(path)
+#endif
+                ? string.Format(CultureInfo.InvariantCulture, @"{0}\{1}", path.Replace(@"\\.", @"\??"), uniqueId)
+                : string.Format(CultureInfo.InvariantCulture, @"\??\UNC\{0}\{1}", path.Trim('\\'), uniqueId);
```
JRahnama, line 698: I have seen usage of `string.Format` and string `interpolation` in this changes. Can we stay with string interpolation ?
string.Format has some limitation which has been addressed in string interpolation.
 - Every time String.Format is called, it needs to parse the composite format string to find all the literal portions of the text, all of the format items, and their specifiers and alignments; somewhat ironically in the case of string interpolation, the C# compiler already had to do such parsing in order to parse the interpolated string and generate the String.Format, yet it has to be done again at run-time for each call.
- These APIs all accept arguments typed as System.Object, which means that any value types end up getting boxed in order to be passed in as an argument.
- There are String.Format overloads that accept up to three individual arguments, but for cases where more than three are needed, there’s a catch-all overload that accepts a params Object[]. That means any number of arguments more than three allocates an array.

More could be read [here](https://devblogs.microsoft.com/dotnet/string-interpolation-in-c-10-and-net-6/)

cheenamalhotra: I didn't modify the file, it's as-is.

JRahnama (approved):

PR #1251, opened 2021-09-03
Drop tables in Dispose() for ConversionTests

### LINKED ISSUE #1230: Test AlwaysEncrypted.ConversionTests.ConversionSmallerToLarger... fails
The test Microsoft.Data.SqlClient.ManualTesting.Tests.AlwaysEncrypted.ConversionTests.ConversionSmallerToLargerInsertAndSelectBulk is failing over multiple PRs with the error:

```
System.AggregateException : One or more errors occurred. (Cannot access destination table 'AE_large_type_enc_112eff3e_f052_464d_8943_170b2b2d3f30'.) (Cannot drop column encryption key 'AE_CEK_0762a097_513a_4c7d_8442_f280a9212167' because the key is referenced by column 'AE_large_type_enc_112eff3e_f052_464d_8943_170b2b2d3f30.Column1'.)
---- System.InvalidOperationException : Cannot access destination table 'AE_large_type_enc_112eff3e_f052_464d_8943_170b2b2d3f30'.
-------- Microsoft.Data.SqlClient.SqlException : Execution Timeout Expired.  The timeout period elapsed prior to completion of the operation or the server is not responding.
The request failed to run because the batch is aborted, this can be caused by abort signal sent from client, or another request is running in the same session, which makes the session busy.
Operation cancelled by user.
------------ System.ComponentModel.Win32Exception : The wait operation timed out
---- Microsoft.Data.SqlClient.SqlException : Cannot drop column encryption key 'AE_CEK_0762a097_513a_4c7d_8442_f280a9212167' because the key is referenced by column 'AE_large_type_enc_112eff3e_f052_464d_8943_170b2b2d3f30.Column1'.
```

which looks like it's a test setup issue.

johnnypham:
Attempt to fix #1230 

The queries to drop the tables are moved from the end of each test method to `Dispose()`, to drop the tables before dropping the CEKs.

### DIFF
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/ConversionTests.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/ConversionTests.cs
@@ -34,6 +34,7 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests.AlwaysEncrypted
         private ColumnEncryptionKey columnEncryptionKey;
         private SqlColumnEncryptionCertificateStoreProvider certStoreProvider = new SqlColumnEncryptionCertificateStoreProvider();
         private List<DbObject> _databaseObjects = new List<DbObject>();
+        private List<string> _tables = new();
 
         private class ColumnMetaData
         {
@@ -55,7 +56,7 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests.AlwaysEncrypted
 
         public ConversionTests()
         {
-            if(certificate == null)
+            if (certificate == null)
             {
                 certificate = CertificateUtility.CreateCertificate();
             }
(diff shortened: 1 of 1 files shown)


DavoudEshtehari (approved):
It seems reasonable, and I couldn't see any issue with it. Just be cautious! We had a known intermittent issue with AE tests, which were related to the `Dispose` functionality in the fixture on facing any exception.

cc @JRahnama

JRahnama:
> It seems reasonable, and I couldn't see any issue with it. Just be cautious! We had a known intermittent issue with AE tests, which were related to the `Dispose` functionality in the fixture on facing any exception.

The issue is a known issue, as far as I could research, in `XUnit` when using `IClassFixture`. The dispose is not called till all the tests related to the fixture is ran to completion. If the tests are stopped or aborted at the middle no Dispose is called to clean up after.

JRahnama (approved):

Issue #1250, opened 2021-09-03
DbCommand.ExecuteReaderAsync throws TaskCanceledException with wrong CancellationToken


Coder3333:
I am witnessing this behavior in packages\netstandard.library\2.0.3\build\netstandard2.0\ref\netstandard.dll when running from a .Net Core 3.1 application.  

When I call DbCommand.ExecuteReaderAsync with a CancellationToken and cancel it before it finishes, an exception is thrown that contains a TaskCanceledException buried down in the InnerExceptions.  I expect that behavior, but the issue is that the TaskCanceledException.CancellationToken is not the CancellationToken that I provided to ExecuteReaderAsync.  I do not know where this other CancellationToken is coming from, but I thought we were only supposed to suppress TaskCanceledException in our applications if it is for the provided CancellationToken, so I end up with a lot more unhandled exceptions than I want.  

Is it safe to suppress any TaskCanceledException, no matter what CancellationToken it contains?  Where does this mystery CancellationToken come from, and could ExecuteReaderAsync just return the CancellationToken given to it?

```c#
                using (var cts = new CancellationTokenSource())
                {
                    using (var cmd = db.GetSqlStringCommand(queryString))
                    {
                        cts.Cancel();
                        try
                        {
                            using (var rdr = cmd.ExecuteReaderAsync(cts.Token).Result)
                            {
                            }
                        }
                        catch (Exception ex3)
                        {
//  if you inspect ex3, you will find a TaskCanceledException buried in InnerExceptions, but its CancellationToken is not cts.Token
                            throw;
                        }
                    }
                }

```

Additional Comments:
I have tested this with "System.Data.SqlClient.SqlCommand" and "Microsoft.Data.SqlClient.SqlCommand".  The result is the same  in both cases where the TaskCanceledException holds a new CancellationToken and the TaskCanceledException is wrapped in an AggregateException, so I have to dig down in the exception stack to find the TaskCanceledException , which means I cannot just catch TaskCanceledException.

System.AggregateException: One or more errors occurred. (A task was canceled.)
 ---> System.Threading.Tasks.TaskCanceledException: A task was canceled.
   --- End of inner exception stack trace ---
   at System.Threading.Tasks.Task.ThrowIfExceptional(Boolean includeTaskCanceledExceptions)
   at System.Threading.Tasks.Task`1.GetResultCore(Boolean waitCompletionNotification)
   at System.Threading.Tasks.Task`1.get_Result()

roji:
[...]

Issue #26, opened 2019-05-15
Cancelling an async SqlClient operation throws SqlException, not TaskCanceledException


roji:
When using a cancellation token to cancel an SqlClient async operation, an SqlException is thrown rather than TaskCanceledException. This seems like problematic behavior, here's some analysis.

On the problematic side, standard async APIs generally throw TaskCanceledException, and SqlClient is doing something different and therefore unexpected. Also, as a result, canceled SqlClient tasks don't have status Canceled, they're Faulted instead. This makes differentiating cancellations from real exceptions more difficult, and the problem is compounded if the SqlClient call is just part of a larger async operation (which again, will be faulted instead of canceled unless some manual specific handling is involved).

The argument for the current behavior seems to be that all of SqlClient's server and network errors are raised as SqlException, making it easy to catch all database errors under a single exception type. On the other hand, it seems that cancellation isn't really an error condition as it's been requested by the user and is expected to occur. The TaskCanceledException is simply a mechanism for conveying the cancellation up the stack.

For the record, this was originally reported by @mikkeljohnsen for Npgsql in https://github.com/npgsql/npgsql/issues/1146. Whatever is decided here for SqlClient will also be adopted for Npgsql (which currently behaves like SqlClient in this respect).

cheenamalhotra:
Acknowledged.
As it seems to be a breaking API change, we'll consider fixing it with other API changes, when that happens.

virzak:
I ran into this as well.

Having to use a hack to suppress. Is there a better way to handle this in the meantime?

```csharp
catch (OperationCanceledException)
{
}
// Fixme:
// Remove this catch when issue is resolved: https://github.com/dotnet/SqlClient/issues/26
catch (Microsoft.Data.SqlClient.SqlException sqlex)
{
    var errors = sqlex.Errors;
    var isCancelled = false;
    foreach (var error in errors)
    {
        if (!(error is Microsoft.Data.SqlClient.SqlError sqlError))
            continue;
        if (sqlError.Message == "Operation cancelled by user.")
        {
            isCancelled = true;
        }
    }

    if (!isCancelled)
        throw;
}
```

proff:
@virzak, you used cancellationTokenfor cancel operation, so yo can use it for exception check:
```
catch (SqlException ex) when (cancellationToken.IsCancellationRequested && ex.Number == 0 && ex.State == 0 && ex.Class == 11)
{
    throw new OperationCanceledException(cancellationToken);
}
```

[...]

PR #1253, opened 2021-09-03
WIP | Try to fix Cancelation Task exception

JRahnama:
Possible fix for #1250 
If approved, same fixes should go to `SqlConnection`. I have added to SqlConnection so far, but waiting to see others opinion to add it to `InternalOpenWithRetryLogic`.
@roji and @Wraith2 could this be a possible fix for this issue?

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
@@ -2643,7 +2643,12 @@ namespace Microsoft.Data.SqlClient
         /// <include file='../../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlCommand.xml' path='docs/members[@name="SqlCommand"]/ExecuteReaderAsync[@name="CancellationToken"]/*'/>
         new public Task<SqlDataReader> ExecuteReaderAsync(CancellationToken cancellationToken)
         {
-            return ExecuteReaderAsync(CommandBehavior.Default, cancellationToken);
+            Task<SqlDataReader> resultTask = ExecuteReaderAsync(CommandBehavior.Default, cancellationToken);
+            if (resultTask.IsCanceled)
+            {
+                cancellationToken.ThrowIfCancellationRequested();
+            }
+            return resultTask;
         }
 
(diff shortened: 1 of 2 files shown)


Wraith2:
I haven't investigated but it's worth a try.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
```diff
@@ -2643,7 +2643,12 @@ protected override Task<DbDataReader> ExecuteDbDataReaderAsync(CommandBehavior b
         /// <include file='../../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlCommand.xml' path='docs/members[@name="SqlCommand"]/ExecuteReaderAsync[@name="CancellationToken"]/*'/>
         new public Task<SqlDataReader> ExecuteReaderAsync(CancellationToken cancellationToken)
         {
-            return ExecuteReaderAsync(CommandBehavior.Default, cancellationToken);
+            Task<SqlDataReader> resultTask = ExecuteReaderAsync(CommandBehavior.Default, cancellationToken);
+            if (resultTask.IsCanceled)
```
roji, line 2647: The right solution here would be to go to where the TaskCanceledException is originally thrown (i.e. somewhere inside ExecuteReaderAsync), and make sure the correct exception with the correct token is thrown there.

(By the way, I don't think this code would work because of #26 - SqlClient currently doesn't throw TaskCanceledException but rather SqlException, so resultTask here would be faulted rather than canceled. But even if it did work, this is better fixed at the source)

johnnypham: >SqlClient currently doesn't throw TaskCanceledException but rather SqlException, so resultTask here would be faulted rather than canceled

A TaskCanceledException is thrown and the task is canceled if the token is canceled before execution starts (i.e. at the beginning of ExecuteReaderAsync). If the token is canceled during execution, a SqlException is thrown and the task is faulted. A lot of effort would be required to fix this properly. I already put a few weeks into it but was eventually blocked.

Wraith2: Am I right that we'd need to somehow include the cancellation token in the SNIError so that it could be included when we throw?

PR #1260, opened 2021-09-10
POC for TokenCredential support

christothes:
This is a POC for discussion for adding support for `TokenCredential` in SqlClient without exposing the type in the core package.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -361,3 +361,5 @@ MigrationBackup/
 
 # Config Json file
 **/config.json
+
+.idea/
--- a/src/Microsoft.Data.SqlClient/add-ons/AzureActiveDirectoryAuthenticationProvider/AzureActiveDirectoryAuthenticationProvider.cs
+++ b/src/Microsoft.Data.SqlClient/add-ons/AzureActiveDirectoryAuthenticationProvider/AzureActiveDirectoryAuthenticationProvider.cs
@@ -1,0 +1,89 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Azure.Core;
+
+namespace Microsoft.Data.SqlClient
+{
+    /// <summary>
+    ///
+    /// </summary>
+    public sealed class AzureActiveDirectoryAuthenticationProvider : SqlAuthenticationProvider
+    {
+        private static readonly string s_defaultScopeSuffix = "/.default";
+        private readonly string _type = typeof(AzureActiveDirectoryAuthenticationProvider).Name;
+        private readonly SqlClientLogger _logger = new();
+        // private readonly string _applicationClientId = ActiveDirectoryAuthentication.AdoClientId;
+        private readonly TokenCredential _credential;
+
+        /// <summary>
+        ///
+        /// </summary>
+        public AzureActiveDirectoryAuthenticationProvider(TokenCredential credential)
+        {
+            _credential = credential;
+        }
+
+        /// <summary>
+        ///
+        /// </summary>
+        /// <param name="authentication"></param>
+        /// <returns></returns>
+        public override bool IsSupported(SqlAuthenticationMethod authentication) => authentication switch
+        {
+            SqlAuthenticationMethod.ActiveDirectoryTokenCredential => true,
+            SqlAuthenticationMethod.ActiveDirectoryIntegrated => true,
+            SqlAuthenticationMethod.ActiveDirectoryPassword => true,
+            SqlAuthenticationMethod.ActiveDirectoryInteractive => true,
+            SqlAuthenticationMethod.ActiveDirectoryServicePrincipal => true,
+            SqlAuthenticationMethod.ActiveDirectoryDeviceCodeFlow => true,
+            SqlAuthenticationMethod.ActiveDirectoryManagedIdentity => true,
+            SqlAuthenticationMethod.ActiveDirectoryMSI => true,
+            SqlAuthenticationMethod.ActiveDirectoryDefault => true,
+            _ => false
+        };
+
+        /// <summary>
+        ///
+        /// </summary>
+        /// <param name="authentication"></param>
+        public override void BeforeLoad(SqlAuthenticationMethod authentication)
+        {
+            _logger.LogInfo(_type, "BeforeLoad", $"being loaded into SqlAuthProviders for {authentication}.");
+        }
+
+        /// <summary>
+        ///
+        /// </summary>
+        /// <param name="authentication"></param>
+        public override void BeforeUnload(SqlAuthenticationMethod authentication)
+        {
+            _logger.LogInfo(_type, "BeforeUnload", $"being unloaded from SqlAuthProviders for {authentication}.");
+        }
+
+        /// <summary>
+        ///
+        /// </summary>
+        /// <param name="parameters"></param>
+        /// <returns></returns>
+        /// <exception cref="Exception"></exception>
+        public override async Task<SqlAuthenticationToken> AcquireTokenAsync(SqlAuthenticationParameters parameters)
+        {
+            CancellationTokenSource cts = new();
+
+            // Use Connection timeout value to cancel token acquire request after certain period of time.
+            cts.CancelAfter(parameters.ConnectionTimeout * 1000); // Convert to milliseconds
+
+            string scope = parameters.Resource.EndsWith(s_defaultScopeSuffix) ? parameters.Resource : parameters.Resource + s_defaultScopeSuffix;
+            string[] scopes = { scope };
+            TokenRequestContext tokenRequestContext = new(scopes);
+            string clientId = string.IsNullOrWhiteSpace(parameters.UserId) ? null : parameters.UserId;
+            AccessToken accessToken = await _credential.GetTokenAsync(tokenRequestContext, cts.Token);
+            return new SqlAuthenticationToken(accessToken.Token, accessToken.ExpiresOn);
+        }
+    }
+}
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
@@ -1315,6 +1315,7 @@ namespace Microsoft.Data.SqlClient
                 || ConnectionOptions.Authentication == SqlAuthenticationMethod.ActiveDirectoryManagedIdentity
                 || ConnectionOptions.Authentication == SqlAuthenticationMethod.ActiveDirectoryMSI
                 || ConnectionOptions.Authentication == SqlAuthenticationMethod.ActiveDirectoryDefault
+                || ConnectionOptions.Authentication == SqlAuthenticationMethod.ActiveDirectoryTokenCredential
                 // Since AD Integrated may be acting like Windows integrated, additionally check _fedAuthRequired
                 || (ConnectionOptions.Authentication == SqlAuthenticationMethod.ActiveDirectoryIntegrated && _fedAuthRequired))
             {
@@ -2124,6 +2125,7 @@ namespace Microsoft.Data.SqlClient
                          || ConnectionOptions.Authentication == SqlAuthenticationMethod.ActiveDirectoryManagedIdentity
                          || ConnectionOptions.Authentication == SqlAuthenticationMethod.ActiveDirectoryMSI
                          || ConnectionOptions.Authentication == SqlAuthenticationMethod.ActiveDirectoryDefault
+                         || ConnectionOptions.Authentication == SqlAuthenticationMethod.ActiveDirectoryTokenCredential
                          || (ConnectionOptions.Authentication == SqlAuthenticationMethod.ActiveDirectoryIntegrated && _fedAuthRequired),
                          "Credentials aren't provided for calling MSAL");
             Debug.Assert(fedAuthInfo != null, "info should not be null.");
@@ -2369,6 +2371,7 @@ namespace Microsoft.Data.SqlClient
                         case SqlAuthenticationMethod.ActiveDirectoryManagedIdentity:
                         case SqlAuthenticationMethod.ActiveDirectoryMSI:
                         case SqlAuthenticationMethod.ActiveDirectoryDefault:
+                        case SqlAuthenticationMethod.ActiveDirectoryTokenCredential:
                             if (_activeDirectoryAuthTimeoutRetryHelper.State == ActiveDirectoryAuthenticationTimeoutRetryState.Retrying)
                             {
                                 _fedAuthToken = _activeDirectoryAuthTimeoutRetryHelper.CachedToken;
--- a/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
+++ b/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
@@ -2572,6 +2572,7 @@ namespace Microsoft.Data.SqlClient
                          || ConnectionOptions.Authentication == SqlAuthenticationMethod.ActiveDirectoryManagedIdentity
                          || ConnectionOptions.Authentication == SqlAuthenticationMethod.ActiveDirectoryMSI
                          || ConnectionOptions.Authentication == SqlAuthenticationMethod.ActiveDirectoryDefault
+                         || ConnectionOptions.Authentication == SqlAuthenticationMethod.ActiveDirectoryTokenCredential
                          || ConnectionOptions.Authentication == SqlAuthenticationMethod.ActiveDirectoryDeviceCodeFlow
                          || (ConnectionOptions.Authentication == SqlAuthenticationMethod.ActiveDirectoryIntegrated && _fedAuthRequired),
                          "Credentials aren't provided for calling MSAL");
--- a/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsEnums.cs
+++ b/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsEnums.cs
@@ -245,6 +245,7 @@ namespace Microsoft.Data.SqlClient
         public const byte MSALWORKFLOW_ACTIVEDIRECTORYDEVICECODEFLOW = 0x03; // Using the Interactive byte as that is the closest we have
         public const byte MSALWORKFLOW_ACTIVEDIRECTORYMANAGEDIDENTITY = 0x03; // Using the Interactive byte as that's supported for Identity based authentication
         public const byte MSALWORKFLOW_ACTIVEDIRECTORYDEFAULT = 0x03; // Using the Interactive byte as that is the closest we have to non-password based authentication modes
+        public const byte MSALWORKFLOW_ACTIVEDIRECTORYTOKENCREDENTIAL = 0x03; // Using the Interactive byte as that is the closest we have to non-password based authentication modes
 
         public enum ActiveDirectoryWorkflow : byte
         {
@@ -255,6 +256,7 @@ namespace Microsoft.Data.SqlClient
             DeviceCodeFlow = MSALWORKFLOW_ACTIVEDIRECTORYDEVICECODEFLOW,
             ManagedIdentity = MSALWORKFLOW_ACTIVEDIRECTORYMANAGEDIDENTITY,
             Default = MSALWORKFLOW_ACTIVEDIRECTORYDEFAULT,
+            TokenCredential = MSALWORKFLOW_ACTIVEDIRECTORYTOKENCREDENTIAL,
         }
 
         // The string used for username in the error message when Authentication = Active Directory Integrated with FedAuth is used, if authentication fails.
@@ -1124,8 +1126,12 @@ namespace Microsoft.Data.SqlClient
         /// <include file='../../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlAuthenticationMethod.xml' path='docs/members[@name="SqlAuthenticationMethod"]/ActiveDirectoryDefault/*'/>
         ActiveDirectoryDefault,
 #if ADONET_CERT_AUTH
-        SqlCertificate
+        SqlCertificate,
 #endif
+        /// <summary>
+        ///
+        /// </summary>
+        ActiveDirectoryTokenCredential
     }
     // This enum indicates the state of TransparentNetworkIPResolution
     // The first attempt when TNIR is on should be sequential. If the first attempt failes next attempts should be parallel.
--- a/src/Microsoft.Data.SqlClient.sln
+++ b/src/Microsoft.Data.SqlClient.sln
@@ -200,6 +200,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Data.SqlClient.Do
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Data.SqlClient.PerformanceTests", "Microsoft.Data.SqlClient\tests\PerformanceTests\Microsoft.Data.SqlClient.PerformanceTests.csproj", "{599A336B-2A5F-473D-8442-1223ED37C93E}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.Data.SqlClient.AzureAadAuthProvider", "Microsoft.Data.SqlClient\add-ons\AzureActiveDirectoryAuthenticationProvider\Microsoft.Data.SqlClient.AzureAadAuthProvider.csproj", "{48800BF7-13D9-48C9-A2CC-6F8017DF21B6}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
@@ -472,6 +474,18 @@ Global
 		{599A336B-2A5F-473D-8442-1223ED37C93E}.Release|x64.Build.0 = Release|x64
 		{599A336B-2A5F-473D-8442-1223ED37C93E}.Release|x86.ActiveCfg = Release|x86
 		{599A336B-2A5F-473D-8442-1223ED37C93E}.Release|x86.Build.0 = Release|x86
+		{48800BF7-13D9-48C9-A2CC-6F8017DF21B6}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{48800BF7-13D9-48C9-A2CC-6F8017DF21B6}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{48800BF7-13D9-48C9-A2CC-6F8017DF21B6}.Debug|x64.ActiveCfg = Debug|x64
+		{48800BF7-13D9-48C9-A2CC-6F8017DF21B6}.Debug|x64.Build.0 = Debug|x64
+		{48800BF7-13D9-48C9-A2CC-6F8017DF21B6}.Debug|x86.ActiveCfg = Debug|x86
+		{48800BF7-13D9-48C9-A2CC-6F8017DF21B6}.Debug|x86.Build.0 = Debug|x86
+		{48800BF7-13D9-48C9-A2CC-6F8017DF21B6}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{48800BF7-13D9-48C9-A2CC-6F8017DF21B6}.Release|Any CPU.Build.0 = Release|Any CPU
+		{48800BF7-13D9-48C9-A2CC-6F8017DF21B6}.Release|x64.ActiveCfg = Release|x64
+		{48800BF7-13D9-48C9-A2CC-6F8017DF21B6}.Release|x64.Build.0 = Release|x64
+		{48800BF7-13D9-48C9-A2CC-6F8017DF21B6}.Release|x86.ActiveCfg = Release|x86
+		{48800BF7-13D9-48C9-A2CC-6F8017DF21B6}.Release|x86.Build.0 = Release|x86
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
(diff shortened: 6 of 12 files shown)


src/Microsoft.Data.SqlClient/add-ons/AzureActiveDirectoryAuthenticationProvider/AzureActiveDirectoryAuthenticationProvider.cs
```diff
@@ -0,0 +1,89 @@
...
+{
+    /// <summary>
+    ///
+    /// </summary>
+    public sealed class AzureActiveDirectoryAuthenticationProvider : SqlAuthenticationProvider
+    {
+        private static readonly string s_defaultScopeSuffix = "/.default";
+        private readonly string _type = typeof(AzureActiveDirectoryAuthenticationProvider).Name;
+        private readonly SqlClientLogger _logger = new();
+        // private readonly string _applicationClientId = ActiveDirectoryAuthentication.AdoClientId;
```
AlexGhiondea, line 20: remove?

src/NuGet.config
```diff
@@ -3,5 +3,6 @@
   <packageSources>
     <clear />
     <add key="dotnet-public" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json" />
+    <add key="dnceng" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json" />
```
David-Engel, line 6: Revert. We can only reference one feed in our config per security standards.

I think I figured out why this always works in CI but not VS (the missing RemoteExecutor package error). Since the CI is all command-line driven, the build tests task has its own [Nuget.config](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/tests/NuGet.config) with a reference to this feed. So if you build via command line it always works. If you build locally via command line first, then VS, VS works because the package is cached from the command line build.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsEnums.cs
```diff
@@ -0,0 +1,1191 @@
+// Licensed to the .NET Foundation under one or more agreements.
```
David-Engel, line 1: Was this file added by mistake? It already exists in the shared code.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/DbConnectionStringCommon.cs
```diff
@@ -361,7 +361,8 @@ internal static ApplicationIntent ConvertToApplicationIntent(string keyword, obj
             "ActiveDirectoryDeviceCodeFlow",
             "ActiveDirectoryManagedIdentity",
             "ActiveDirectoryMSI",
-            "ActiveDirectoryDefault"
+            "ActiveDirectoryDefault",
+            "ActiveDirectoryTokenCredential"
```
David-Engel, line 365: We don't need to add a new Authentication option. We can simply key off whether AccessTokenCallback is null or not.
```suggestion
            "ActiveDirectoryDefault"
```

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/DbConnectionStringCommon.cs
```diff
@@ -516,7 +517,7 @@ internal static string ColumnEncryptionSettingToString(SqlConnectionColumnEncryp
 
         internal static bool IsValidAuthenticationTypeValue(SqlAuthenticationMethod value)
         {
-            Debug.Assert(Enum.GetNames(typeof(SqlAuthenticationMethod)).Length == 10, "SqlAuthenticationMethod enum has changed, update needed");
+            Debug.Assert(Enum.GetNames(typeof(SqlAuthenticationMethod)).Length == 11, "SqlAuthenticationMethod enum has changed, update needed");
```
David-Engel, line 520:
```suggestion
            Debug.Assert(Enum.GetNames(typeof(SqlAuthenticationMethod)).Length == 10, "SqlAuthenticationMethod enum has changed, update needed");
```

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsEnums.cs
```diff
@@ -282,6 +285,7 @@ public enum ActiveDirectoryWorkflow : byte
             DeviceCodeFlow = MSALWORKFLOW_ACTIVEDIRECTORYDEVICECODEFLOW,
             ManagedIdentity = MSALWORKFLOW_ACTIVEDIRECTORYMANAGEDIDENTITY,
             Default = MSALWORKFLOW_ACTIVEDIRECTORYDEFAULT,
+            TokenCredential = MSALWORKFLOW_ACTIVEDIRECTORYTOKENCREDENTIAL,
```
David-Engel, line 288: Revert. Not referenced.

David-Engel (commented):
We can eliminate the need for a new Authentication option by simply keying off whether AccessTokenCallback is null or not. So the connection string/example usage would be simplified:
 
> ```cs
> var cred = new DefaultAzureCredential();
> var myTokenCallback = async (ctx, cancellationToken) =>
> {
> 	string scope = ctx.Resource.EndsWith(s_defaultScopeSuffix) ? ctx.Resource : ctx.Resource + s_defaultScopeSuffix;
> 	var token = await cred.GetTokenAsync(new TokenRequestContext(new[] { scope }), cancellationToken);
> 	return new SqlAuthenticationToken(token.Token, token.ExpiresOn);
> }
> using (SqlConnection sqlConnection = new SqlConnection(
>     "Server=name.database.windows.net;Database=db;")
>         { AccessTokenCallback = myTokenCallback })
> {
> 	sqlConnection.Open();
> 	Console.WriteLine("Connected successfully!")
> }
> ```

src/Microsoft.Data.SqlClient/netfx/src/Microsoft.Data.SqlClient.csproj
```diff
@@ -687,7 +690,16 @@
...
+    <EmbeddedResource Include="Resources\Strings.de.resx" />
+    <EmbeddedResource Include="Resources\Strings.es.resx" />
+    <EmbeddedResource Include="Resources\Strings.fr.resx" />
+    <EmbeddedResource Include="Resources\Strings.it.resx" />
+    <EmbeddedResource Include="Resources\Strings.ja.resx" />
+    <EmbeddedResource Include="Resources\Strings.ko.resx" />
+    <EmbeddedResource Include="Resources\Strings.pt-BR.resx" />
+    <EmbeddedResource Include="Resources\Strings.ru.resx" />
+    <EmbeddedResource Include="Resources\Strings.zh-Hans.resx" />
+    <EmbeddedResource Include="Resources\Strings.zh-Hant.resx" />
```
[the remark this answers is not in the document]
christothes: thanks - I believe these happen automatically when a new string is added

cheenamalhotra:
> Discussion update - we'll refine further by implementing the delegate without the need for a `SqlAuthenticationProvider` using the hash of the delegate as the connection pool key.

@christothes Can you cover the advantage of using the AccessTokenCallback over SqlAuthenticationProvider implementation in terms of access token/pool key - I recall there was a discussion long time ago, but let's capture that in this PR description?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/AzureADTokenRequestContext.cs
```diff
@@ -0,0 +1,25 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System.Threading;
+
+namespace Microsoft.Data.SqlClient
+{
+    /// <summary>
+    ///
+    /// </summary>
+    public class AzureADTokenRequestContext
```
David-Engel, line 12: Good point on not using AAD. AccessTokenRequestContext might be even better here. We don't have the term "Azure" in any of our public APIs.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -2818,43 +2839,65 @@ internal SqlFedAuthToken GetFedAuthToken(SqlFedAuthInfo fedAuthInfo)
...
+                            {
+                                authParamsBuilder.WithUserId(ConnectionOptions.UserID);
+                                SqlAuthenticationParameters parameters = authParamsBuilder;
+                                CancellationTokenSource cts = new();
+                                // Use Connection timeout value to cancel token acquire request after certain period of time.(int)
+                                if (_timeout.MillisecondsRemaining < Int32.MaxValue)
+                                {
+                                    cts.CancelAfter((int)_timeout.MillisecondsRemaining);
+                                }
+                                _fedAuthToken = Task.Run(async () => await _accessTokenCallback(new AzureADTokenRequestContext(parameters.Resource), cts.Token)).GetAwaiter().GetResult().ToSqlFedAuthToken();
```
cheenamalhotra, line 2894: As discussed, please pass `authParamsBuilder` directly, just like how `AcquireTokenAsync` is being called.. so implementation can take care of performing cancellations with `timeout` available.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnection.cs
```diff
@@ -688,6 +694,35 @@ public string AccessToken
             }
         }
 
+        /// <summary>
+        ///
+        /// </summary>
+        public Func<AzureADTokenRequestContext, CancellationToken, Task<SqlAuthenticationToken>> AccessTokenCallback
```
cheenamalhotra, line 700: Please also include a sample on how to use this API in the [samples](https://github.com/dotnet/SqlClient/tree/main/doc/samples) folder demonstrating basic working example, which can be later added to public documentations.

/cc @DavoudEshtehari @David-Engel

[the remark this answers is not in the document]
christothes: done

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnection.cs
```diff
@@ -1024,6 +1059,49 @@ private void CheckAndThrowOnInvalidCombinationOfConnectionOptionAndAccessToken(S
...
+            {
+                throw ADP.InvalidMixedUsageOfAccessTokenAndTokenCallback();
+            }
+        }
+
+        // CheckAndThrowOnInvalidCombinationOfConnectionOptionAndAccessTokenCallback: check if the usage of AccessTokenCallback has any conflict
+        //  with the keys used in connection string and credential
+        //  If there is any conflict, it throws InvalidOperationException
+        //  This is to be used setter of ConnectionString and AccessTokenCallback properties
+        private void CheckAndThrowOnInvalidCombinationOfConnectionOptionAndAccessTokenCallback(SqlConnectionString connectionOptions)
```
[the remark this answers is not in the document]
christothes: Other than user Id and Password, is there any other interesting combo?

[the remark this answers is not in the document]
cheenamalhotra: You can try with SqlCredential object set on SqlConnection object too. It should be validated same way as UID/PWD on connection string. Secondly setting 'Authentication' property on connection string along with callback should not be allowed too.. So we should be targeting all authentication modes to not conflict with the callback same way as with AccessToken string property.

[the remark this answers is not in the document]
christothes: I think we are doing that now

doc/snippets/Microsoft.Data.SqlClient/SqlConnection.xml
```diff
@@ -129,6 +129,21 @@ using (SqlConnection connection = new SqlConnection(connectionString))
             <value>The access token for the connection.</value>
             <remarks>To be added.</remarks>
         </AccessToken>
+        <AccessTokenCallback>
```
DavoudEshtehari, line 132: Can you add the expected exception as well?

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -4,6 +4,10 @@
 
 // NOTE: The current Microsoft.VSDesigner editor attributes are implemented for System.Data.SqlClient, and are not publicly available.
 // New attributes that are designed to work with Microsoft.Data.SqlClient and are publicly documented should be included in future.
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+
```
DavoudEshtehari, line 10: NIT, following other declarations, uses a full name instead of adding using keyword.

src/Microsoft.Data.SqlClient/netcore/src/Resources/Strings.resx
```diff
@@ -1941,4 +1941,13 @@
...
     <value>Socket did not throw expected '{0}' with error code '{1}'.</value>
   </data>
+  <data name="ADP_InvalidMixedUsageOfAccessTokenAndTokenCallback" xml:space="preserve">
+    <value>Cannot set the AccessToken property if the AccessTokenCallback has been set.</value>
+  </data>
+  <data name="ADP_InvalidMixedUsageOfAccessTokenCallbackAndIntegratedSecurity" xml:space="preserve">
+    <value>Cannot set the AccessTokenCallback property if the 'Integrated Security' connection string keyword has been set to 'true' or 'SSPI'.</value>
+  </data>
+  <data name="ADP_InvalidMixedUsageOfAuthenticationAndTokenCallback" xml:space="preserve">
+    <value>Cannot set the AccessTokenCallback property if 'Authentication' has been specified in the connection string.</value>
```
DavoudEshtehari, line 1951:
```suggestion
    <value>Cannot set the AccessTokenCallback property if 'Authentication=Active Directory Default' has been specified in the connection string.</value>
```

src/Microsoft.Data.SqlClient/netcore/src/Resources/Strings.Designer.cs
```diff
@@ -438,6 +447,24 @@ internal class Strings {
...
+        ///   Looks up a localized string similar to Cannot set the AccessTokenCallback property if the &apos;Integrated Security&apos; connection string keyword has been set to &apos;true&apos; or &apos;SSPI&apos;..
+        /// </summary>
+        internal static string ADP_InvalidMixedUsageOfAccessTokenCallbackAndIntegratedSecurity {
+            get {
+                return ResourceManager.GetString("ADP_InvalidMixedUsageOfAccessTokenCallbackAndIntegratedSecurity", resourceCulture);
+            }
+        }
+        
+        /// <summary>
+        ///   Looks up a localized string similar to Cannot set the AccessTokenCallback property if &apos;Authentication&apos; has been specified in the connection string..
```
DavoudEshtehari, line 460:
```suggestion
        ///   Looks up a localized string similar to Cannot set the AccessTokenCallback property if &apos;Authentication=Active Directory Default&apos; has been specified in the connection string..
```

src/Microsoft.Data.SqlClient/netfx/src/Resources/Strings.resx
```diff
@@ -4635,4 +4635,13 @@
...
   </data>
-</root>
+  <data name="ADP_InvalidMixedUsageOfAccessTokenAndTokenCallback" xml:space="preserve">
+    <value>Cannot set the AccessToken property if the AccessTokenCallback has been set.</value>
+  </data>
+  <data name="ADP_InvalidMixedUsageOfAccessTokenCallbackAndIntegratedSecurity" xml:space="preserve">
+    <value>Cannot set the AccessTokenCallback property if the 'Integrated Security' connection string keyword has been set to 'true' or 'SSPI'.</value>
+  </data>
+  <data name="ADP_InvalidMixedUsageOfAuthenticationAndTokenCallback" xml:space="preserve">
+    <value>Cannot set the AccessTokenCallback property if 'Authentication' has been specified in the connection string.</value>
```
DavoudEshtehari, line 4645:
```suggestion
    <value>Cannot set the AccessTokenCallback property if 'Authentication=Active Directory Default' has been specified in the connection string.</value>
```

src/Microsoft.Data.SqlClient/netfx/src/Resources/Strings.Designer.cs
```diff
@@ -915,6 +924,24 @@ internal class Strings {
...
+        ///   Looks up a localized string similar to Cannot set the AccessTokenCallback property if the &apos;Integrated Security&apos; connection string keyword has been set to &apos;true&apos; or &apos;SSPI&apos;..
+        /// </summary>
+        internal static string ADP_InvalidMixedUsageOfAccessTokenCallbackAndIntegratedSecurity {
+            get {
+                return ResourceManager.GetString("ADP_InvalidMixedUsageOfAccessTokenCallbackAndIntegratedSecurity", resourceCulture);
+            }
+        }
+        
+        /// <summary>
+        ///   Looks up a localized string similar to Cannot set the AccessTokenCallback property if &apos;Authentication&apos; has been specified in the connection string..
```
DavoudEshtehari, line 937:
```suggestion
        ///   Looks up a localized string similar to Cannot set the AccessTokenCallback property if &apos;Authentication=Active Directory Default&apos; has been specified in the connection string..
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnection.cs
```diff
@@ -685,11 +693,38 @@ public string AccessToken
...
+                }
+
+                if (value != null)
+                {
+                    // Check if the usage of AccessToken has any conflict with the keys used in connection string and credential
+                    CheckAndThrowOnInvalidCombinationOfConnectionOptionAndAccessTokenCallback((SqlConnectionString)ConnectionOptions);
+                }
+                else
+                {
+                    throw new ArgumentNullException(nameof(AccessTokenCallback), "Callback cannot be null.");
```
DavoudEshtehari, line 720: This message should be moved to the resources like other exceptions.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnection.cs
```diff
@@ -739,7 +741,34 @@ public string AccessToken
...
+                }
+
+                if (value != null)
+                {
+                    // Check if the usage of AccessToken has any conflict with the keys used in connection string and credential
+                    CheckAndThrowOnInvalidCombinationOfConnectionOptionAndAccessTokenCallback((SqlConnectionString)ConnectionOptions);
+                }
+                else
+                {
+                    throw new ArgumentNullException(nameof(AccessTokenCallback), "Callback cannot be null.");
```
DavoudEshtehari, line 767: Same here.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -479,6 +481,11 @@ internal SqlConnectionTimeoutErrorInternal TimeoutErrorInternal
                 _accessTokenInBytes = System.Text.Encoding.Unicode.GetBytes(accessToken);
             }
 
+            if (accessTokenCallback != null)
+            {
+                _accessTokenCallback = accessTokenCallback;
+            }
```
DavoudEshtehari, line 487: The null check doesn't seem necessary here.
```suggestion
            _accessTokenCallback = accessTokenCallback;
```

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -487,6 +490,11 @@ static SqlInternalConnectionTds()
                 _accessTokenInBytes = System.Text.Encoding.Unicode.GetBytes(accessToken);
             }
 
+            if (accessTokenCallback != null)
+            {
+                _accessTokenCallback = accessTokenCallback;
+            }
```
DavoudEshtehari, line 496: Same note here.
```suggestion
            _accessTokenCallback = accessTokenCallback;
```

src/Microsoft.Data.SqlClient/netfx/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -4,6 +4,10 @@
 
 // NOTE: The current Microsoft.VSDesigner editor attributes are implemented for System.Data.SqlClient, and are not publicly available.
 // New attributes that are designed to work with Microsoft.Data.SqlClient and are publicly documented should be included in future.
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+
```
DavoudEshtehari, line 10: Same note here.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectivityTests/AADConnectionTest.cs
```diff
@@ -551,6 +553,72 @@ public static void ActiveDirectoryDefaultWithPasswordMustFail()
...
+                "Authentication=ActiveDirectoryDefault";
+            InvalidOperationException e = Assert.Throws<InvalidOperationException>(() =>
+            {
+                using (SqlConnection conn = new SqlConnection(connStrWithNoCred))
+                {
+                    conn.AccessTokenCallback = (ctx, token) =>
+                        Task.FromResult(new SqlAuthenticationToken("my token", DateTimeOffset.MaxValue));
+                    conn.Open();
+
+                    Assert.True(conn.State == System.Data.ConnectionState.Open);
```
DavoudEshtehari, line 571: I'd rather it because of more informative reporting content in case of failure.
```suggestion
                    Assert.NotEqual(System.Data.ConnectionState.Open, conn.State);
```

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectivityTests/AADConnectionTest.cs
```diff
@@ -551,6 +553,72 @@ public static void ActiveDirectoryDefaultWithPasswordMustFail()
...
+            string[] credKeys = { "Authentication", "User ID", "Password", "UID", "PWD" };
+            string connStr = DataTestUtility.RemoveKeysInConnStr(DataTestUtility.AADPasswordConnectionString, credKeys);
+            var cred = new DefaultAzureCredential();
+            const string defaultScopeSuffix = "/.default";
+            using (SqlConnection conn = new SqlConnection(connStr))
+            {
+                conn.AccessTokenCallback = (ctx, cancellationToken) =>
+                {
+                    string scope = ctx.Resource.EndsWith(defaultScopeSuffix) ? ctx.Resource : ctx.Resource + defaultScopeSuffix;
+                    var token = cred.GetToken(new TokenRequestContext(new[] { scope }), cancellationToken);
```
DavoudEshtehari, line 591:
```suggestion
                    AccessToken token = cred.GetToken(new TokenRequestContext(new[] { scope }), cancellationToken);
```

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectivityTests/AADConnectionTest.cs
```diff
@@ -551,6 +553,72 @@ public static void ActiveDirectoryDefaultWithPasswordMustFail()
...
+            var cred = new DefaultAzureCredential();
+            const string defaultScopeSuffix = "/.default";
+            using (SqlConnection conn = new SqlConnection(connStr))
+            {
+                conn.AccessTokenCallback = (parms, cancellationToken) =>
+                {
+                    Assert.Equal(userId, parms.UserId);
+                    Assert.Equal(pwd, parms.Password);
+                    string scope = parms.Resource.EndsWith(defaultScopeSuffix) ? parms.Resource : parms.Resource + defaultScopeSuffix;
+                    var token = cred.GetToken(new TokenRequestContext(new[] { scope }), cancellationToken);
```
DavoudEshtehari, line 615:
```suggestion
                    AccessToken token = cred.GetToken(new TokenRequestContext(new[] { scope }), cancellationToken);
```

.gitignore
```diff
@@ -361,3 +361,5 @@ MigrationBackup/
 
 # Config Json file
 **/config.json
+
+.idea/
```
DavoudEshtehari, line 365: Can you add a comment here and explain the purpose of adding it if it's not added mistakenly?

doc/snippets/Microsoft.Data.SqlClient/SqlConnection.xml
christothes: Since this is a callback provided by the developer user, it could be anything 😄

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnection.cs
christothes: I will just eliminate the 2nd arg, like we do in most of the other places we throw ANE.

doc/snippets/Microsoft.Data.SqlClient/SqlConnection.xml
DavoudEshtehari: That's right. I don't mean the application's exceptions, you can include the exceptions that you expect to happen and already have handled.

christothes: I must be missing something - the only exceptions that can occur would be the ones generated by the contents of the callback, which are supplied by the user. Which exceptions did you have in mind that we would expect or already have handled?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnection.cs
```diff
@@ -685,11 +693,38 @@ public string AccessToken
...
+                }
+
+                if (value != null)
+                {
+                    // Check if the usage of AccessToken has any conflict with the keys used in connection string and credential
+                    CheckAndThrowOnInvalidCombinationOfConnectionOptionAndAccessTokenCallback((SqlConnectionString)ConnectionOptions);
+                }
+                else
+                {
+                    throw new ArgumentNullException(nameof(AccessTokenCallback));
```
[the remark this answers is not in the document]
christothes: It must be set when creating the SqlConnection, as shown in the PR description:

```c#
using (SqlConnection sqlConnection = new SqlConnection(
    "Server=name.database.windows.net;Database=db;")
        { AccessTokenCallback = myTokenCallback })
{
	sqlConnection.Open();
	Console.WriteLine("Connected successfully!")
}
```

doc/snippets/Microsoft.Data.SqlClient/SqlConnection.xml
DavoudEshtehari: `InvalidOperationException` and `ArgumentNullException` from the AccessTokenCallback's setter.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnection.cs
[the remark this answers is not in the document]
DavoudEshtehari: Assume you have a SqlConnection with the AccessTokenCallback. When you close it, it should be possible to be reused with different configurations without providing an AccessTokenCallback. With the current implementation, you will get an ArgumentNullException, which is not desired.
Or attempting to set the AccessTokenCallback to null before opening the SqlConnection, ends with the same exception.

```c#
SqlConnection sqlConnection = new SqlConnection();
sqlConnection.AccessTokenCallback = myTokenCallback;
sqlConnection.AccessTokenCallback = null; // throws ArgumentNullException
```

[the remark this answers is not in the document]
christothes: Good point - It sounds like we need to allow setting null to accommodate this, unless you have a better idea.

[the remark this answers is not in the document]
DavoudEshtehari: The null validation can be done before establishing a connection in combination with the connection string properties to verify if it's required.
Also, I'd include the connection status while a callback is going to be assigned.

cc @David-Engel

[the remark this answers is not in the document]
David-Engel: Davoud brought this up to me today and he's right. We need to consider re-use of a SqlConnection. Technically, you can close a connection, change properties (connection string, Server, Authentication, etc), and re-open it. Not a widely used pattern, but supported.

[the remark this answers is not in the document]
christothes: @David-Engel Are you suggesting that setting the callback should close and re-open the connection if it was already open, or that we should throw if it is already open?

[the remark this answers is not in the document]
David-Engel: No. We simply need to allow the AccessTokenCallback property to be set to null here (remove the else case). The use case would be a SqlConnection object re-use one where, after the connection has been closed, some other auth mechanism is being used for a subsequent connection.

David-Engel (approved):
I'm good with the current state of this feature. It'll be nice to get this one merged.

DavoudEshtehari (approved):

David-Engel (approved):

aburakab:
Currently, I'm using `SqlClient` to connect to **Dynamics 365** using **Service Principal Credentials**. 
After the latest update for the `Azure.Identity` Nuget package. It's impossible to use `SqlClient` or maybe an extra step that I should be added for the **Access Token**, as I can see here for the `Default Azure Credentials.` 
I added some details in this [Stackoverflow question](https://stackoverflow.com/questions/78325276/latest-azure-identity-package-caused-an-error-with-sqlclient-and-dynamics-365)

JRahnama:
> Currently, I'm using `SqlClient` to connect to **Dynamics 365** using **Service Principal Credentials**. After the latest update for the `Azure.Identity` Nuget package. It's impossible to use `SqlClient` or maybe an extra step should be added for the **Access Token**, as I can see here for the **Default Azure Credentials**. I added some details in this [Stackoverflow question](https://stackoverflow.com/questions/78325276/latest-azure-identity-package-caused-an-error-with-sqlclient-and-dynamics-365)

Two questions:

1. Have you tried to use the latest stable release of Azure.Identity (1.11.2)?
2. Can you provide a sample repro please?

PR #1265, opened 2021-09-15
Merging common code bases DbConnectionStringCommon

DavoudEshtehari:
Common types:
1. DbConnectionStringDefaults
2. DbConnectionStringKeywords
3. DbConnectionStringSynonyms
4. DbConnectionStringBuilderUtil

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/DbConnectionStringCommon.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/DbConnectionStringCommon.cs
@@ -10,165 +10,6 @@ using Microsoft.Data.SqlClient;
 
 namespace Microsoft.Data.Common
 {
-
-    /*
-        internal sealed class NamedConnectionStringConverter : StringConverter {
-
-            public NamedConnectionStringConverter() {
-            }
-
-            public override bool GetStandardValuesSupported(ITypeDescriptorContext context) {
-                return true;
-            }
-
-            public override bool GetStandardValuesExclusive(ITypeDescriptorContext context) {
-                // Although theoretically this could be true, some people may want to just type in a name
-                return false;
-            }
-
-            public override StandardValuesCollection GetStandardValues(ITypeDescriptorContext context) {
-                StandardValuesCollection standardValues = null;
-                if (null != context) {
-                    DbConnectionStringBuilder instance = (context.Instance as DbConnectionStringBuilder);
-                    if (null != instance) {
-                        string myProviderName = instance.GetType().Namespace;
-
-                        List<string> myConnectionNames = new List<string>();
-                        foreach(System.Configuration.ConnectionStringSetting setting in System.Configuration.ConfigurationManager.ConnectionStrings) {
-                            if (myProviderName.EndsWith(setting.ProviderName)) {
-                                myConnectionNames.Add(setting.ConnectionName);
-                            }
-                        }
-                        standardValues = new StandardValuesCollection(myConnectionNames);
-                    }
-                }
-                return standardValues;
-            }
-        }
-    */
-
-
-    [Serializable()]
-    internal sealed class ReadOnlyCollection<T> : System.Collections.ICollection, ICollection<T>
-    {
-        private T[] _items;
-
-        internal ReadOnlyCollection(T[] items)
-        {
-            _items = items;
-#if DEBUG
-            for (int i = 0; i < items.Length; ++i)
-            {
-                Debug.Assert(null != items[i], "null item");
-            }
-#endif
-        }
-
-        public void CopyTo(T[] array, int arrayIndex)
-        {
-            Array.Copy(_items, 0, array, arrayIndex, _items.Length);
-        }
-
-        void System.Collections.ICollection.CopyTo(Array array, int arrayIndex)
-        {
-            Array.Copy(_items, 0, array, arrayIndex, _items.Length);
-        }
-
-
-        IEnumerator<T> IEnumerable<T>.GetEnumerator()
-        {
-            return new Enumerator<T>(_items);
-        }
-
-        public System.Collections.IEnumerator GetEnumerator()
-        {
-            return new Enumerator<T>(_items);
-        }
-
-        bool System.Collections.ICollection.IsSynchronized
-        {
-            get { return false; }
-        }
-
-        Object System.Collections.ICollection.SyncRoot
-        {
-            get { return _items; }
-        }
-
-        bool ICollection<T>.IsReadOnly
-        {
-            get { return true; }
-        }
-
-        void ICollection<T>.Add(T value)
-        {
-            throw new NotSupportedException();
-        }
-
-        void ICollection<T>.Clear()
-        {
-            throw new NotSupportedException();
-        }
-
-        bool ICollection<T>.Contains(T value)
-        {
-            return Array.IndexOf(_items, value) >= 0;
-        }
-
-        bool ICollection<T>.Remove(T value)
-        {
-            throw new NotSupportedException();
-        }
-
-        public int Count
-        {
-            get { return _items.Length; }
-        }
-
-        [Serializable()]
-        internal struct Enumerator<K> : IEnumerator<K>, System.Collections.IEnumerator
-        { // based on List<T>.Enumerator
-            private K[] _items;
-            private int _index;
-
-            internal Enumerator(K[] items)
-            {
-                _items = items;
-                _index = -1;
-            }
-
-            public void Dispose()
-            {
-            }
-
-            public bool MoveNext()
-            {
-                return (++_index < _items.Length);
-            }
-
-            public K Current
-            {
-                get
-                {
-                    return _items[_index];
-                }
-            }
-
-            Object System.Collections.IEnumerator.Current
-            {
-                get
-                {
-                    return _items[_index];
-                }
-            }
-
-            void System.Collections.IEnumerator.Reset()
-            {
-                _index = -1;
-            }
-        }
-    }
-
     internal static class DbConnectionStringBuilderUtil
     {
 
@@ -190,15 +31,15 @@ namespace Microsoft.Data.Common
                     else if (StringComparer.OrdinalIgnoreCase.Equals(tmp, "false") || StringComparer.OrdinalIgnoreCase.Equals(tmp, "no"))
                         return false;
                 }
-                return Boolean.Parse(svalue);
+                return bool.Parse(svalue);
             }
             try
             {
-                return ((IConvertible)value).ToBoolean(CultureInfo.InvariantCulture);
+                return Convert.ToBoolean(value, CultureInfo.InvariantCulture);
             }
             catch (InvalidCastException e)
             {
-                throw ADP.ConvertFailed(value.GetType(), typeof(Boolean), e);
+                throw ADP.ConvertFailed(value.GetType(), typeof(bool), e);
             }
         }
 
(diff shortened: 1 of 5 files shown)


Wraith2:
What s the reason for a dedicated `ReadOnlyCollection<T>` type? Can we not use the standard bcl provided one? https://apisof.net/catalog/9ff7feaa-ce3f-c2d6-3b5d-4cf21ce96643 and if not can there be a comment explaining the behaviour we need from our own version in the code?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/DbConnectionStringCommon.cs
```diff
@@ -220,63 +61,59 @@ internal static bool ConvertToIntegratedSecurity(object value)
...
         {
             try
             {
-                return ((IConvertible)value).ToString(CultureInfo.InvariantCulture);
+                return Convert.ToString(value, CultureInfo.InvariantCulture);
             }
             catch (InvalidCastException e)
             {
-                throw ADP.ConvertFailed(value.GetType(), typeof(String), e);
+                throw ADP.ConvertFailed(value.GetType(), typeof(string), e);
```
Wraith2, line 96: When looking at this in the diff I wondered if value was guaranteed to be not-null here. What do you think of enabling nullable reference types for some smaller files and seeing if they help us see any possible correctness problems?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/DbConnectionStringCommon.cs
```diff
@@ -1147,20 +962,23 @@ internal static SqlConnectionIPAddressPreference ConvertToIPAddressPreference(st
...
-                || value.StartsWith("subject:", StringComparison.OrdinalIgnoreCase)
-                || value.StartsWith("sha1:", StringComparison.OrdinalIgnoreCase);
-        }
     }
 
     internal static class DbConnectionStringDefaults
     {
-        // all
-        //        internal const string NamedConnection           = "";
         private const string _emptyString = "";
```
Wraith2, line 969: `_emptyString` is pointless. Because the runtime uses string interning all instances of the empty string will refer to the same instance and the only way to change the value of that instance is using unsafe code, if someone does this they should expect a crash and we should not attempt to protect against it. We can just use the literal `""` or `string.Empty` where we need an empty string. There is a similar constant in the ADP class which will eventually be removed.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/DbConnectionStringCommon.cs
```diff
@@ -1147,20 +962,23 @@ internal static SqlConnectionIPAddressPreference ConvertToIPAddressPreference(st
...
+            "Core Microsoft SqlClient Data Provider";
+#endif
+        internal const string AttachDBFilename = _emptyString;
+        internal const int CommandTimeout = 30;
+        internal const int ConnectTimeout = 15;
+
+#if NETFRAMEWORK
         // Odbc
         internal const string Driver = _emptyString;
         internal const string Dsn = _emptyString;
```
Wraith2, line 984: We create two constants with empty string values but I can't think of any good reason to do that instead of just using empty string. These lines are just confusing, weird old code.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/DbConnectionStringCommon.cs
```diff
@@ -1304,10 +1098,15 @@ internal static class DbConnectionStringKeywords
...
         internal const string MinPoolSize = "Min Pool Size";
+        internal const string PoolBlockingPeriod = "Pool Blocking Period";
     }
 
     internal static class DbConnectionStringSynonyms
     {
+#if NETFRAMEWORK
+        //internal const string TransparentNetworkIPResolution = TRANSPARENTNETWORKIPRESOLUTION;
+        internal const string TRANSPARENTNETWORKIPRESOLUTION = "transparentnetworkipresolution";
+#endif
```
Wraith2, line 1109: Weird casing behaviour. If these are strictly internal strings we could audit them to see why they're forced upper and lower case and see if we can normalize the behaviour somehow.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/DbConnectionStringCommon.cs
```diff
@@ -919,18 +734,18 @@ internal static SqlConnectionColumnEncryptionSetting ConvertToColumnEncryptionSe
         /// <returns></returns>
         internal static bool TryConvertToAttestationProtocol(string value, out SqlConnectionAttestationProtocol result)
         {
-            if (StringComparer.InvariantCultureIgnoreCase.Equals(value, AttestationProtocolHGS))
+            if (StringComparer.InvariantCultureIgnoreCase.Equals(value, nameof(SqlConnectionAttestationProtocol.HGS)))
```
Wraith2, line 737: I like the `nameof(thing)` rather than having extra constants 👍

DavoudEshtehari:
> What s the reason for a dedicated `ReadOnlyCollection<T>` type? Can we not use the standard bcl provided one? https://apisof.net/catalog/9ff7feaa-ce3f-c2d6-3b5d-4cf21ce96643 and if not can there be a comment explaining the behaviour we need from our own version in the code?

Thank you for bringing it up. I didn't find any reason for using `ReadOnlyCollection<T>` too. Just brought it here for more discussion.
cc @cheenamalhotra

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/DbConnectionStringCommon.cs
DavoudEshtehari: This block catches the `InvalidCastException` here, which means it couldn't be null at this point; in addition, `Conver.Toxxx()` functions we've used in this type don't raise an exception in facing with null input parameter.
Regarding your question on the second part, could you elaborate on it with more details?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/DbConnectionStringCommon.cs
DavoudEshtehari: The `const` is a compile-time value, not runtime! Using a common constant to declare the string empty would prevent possible miss-assignment. I don't think using `_emptyString` had any drawback here.
Using `const` instead of `readonly` would be a quick solution for ADP.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/DbConnectionStringCommon.cs
DavoudEshtehari: We've ignored the casing in string comparison for connection properties. It was camel-case in some cases using `NetCore` while `NetFx` used lower-case. I tried to keep them unified by porting from 'NetFx' and having the lowest impact on code changes for the sake of review.
My personal preference would be using camel-case cause of more legibility if we managed to modify them.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/DbConnectionStringCommon.cs
DavoudEshtehari: I like to gain benefit of strong types. 😉

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/DbConnectionStringCommon.cs
Wraith2: Using either a const or readonly doesn't matter, there is only one empty string unless you Create one and no-one should do that. 

Using all the extra names for "" obscures the value. If you see `DbConnectionStringDefaults.Driver` you'll think it has some special meaning, it does not, it's just "". It's harmful to readability, it's no harmful at compile or runtime.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/DbConnectionStringCommon.cs
Wraith2: >  having the lowest impact on code changes for the sake of review.

When I've tried this it hasn't helped. Unless you're absolutely sure you'll have chance to do a second pass over the files involved it's better to make everything as coherent as possible in the ones you touch. It's better to change small numbers of files to their best version than work on more files incompletely.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/DbConnectionStringCommon.cs
Wraith2: > Regarding your question on the second part, could you elaborate on it with more details?

https://devblogs.microsoft.com/dotnet/nullable-reference-types-in-csharp/ we can use that per file or even per scope to help get safer code.

cheenamalhotra:
> > What s the reason for a dedicated `ReadOnlyCollection<T>` type? Can we not use the standard bcl provided one? https://apisof.net/catalog/9ff7feaa-ce3f-c2d6-3b5d-4cf21ce96643 and if not can there be a comment explaining the behaviour we need from our own version in the code?
> 
> Thank you for bringing it up. I didn't find any reason for using `ReadOnlyCollection<T>` too. Just brought it here for more discussion.
> cc @cheenamalhotra

I agree we could use the standard BCL type already available.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/DbConnectionStringCommon.cs
DavoudEshtehari: I removed null-referenced old constants. Pruning the stale keywords could be done in a separate PR.

Wraith2: excellent, thanks :)

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/DbConnectionStringCommon.cs
DavoudEshtehari: I replaced the `_emptyString` with `""`.
Regarding the argument about const string default values and empty string, I agree it'd obscure the value but consider it has some benefits like having reference to all usages in the one place, which means code maintenance would be much easier and safer.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/DbConnectionStringCommon.cs
DavoudEshtehari: Thank you @Wraith2, by enabling it I managed to find more possible improvements in this file.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/DbConnectionStringCommon.cs
DavoudEshtehari: We already have related functional tests which were passed. We could rich those tests If you think we've missed some tests here.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/DbConnectionStringCommon.cs
```diff
@@ -1147,102 +926,63 @@ internal static SqlConnectionIPAddressPreference ConvertToIPAddressPreference(st
             }
         }
         #endregion
-
-        internal static bool IsValidCertificateValue(string value)
```
johnnypham, line 1151: This method is used in SqlConnectionString in a `#if ADONET_CERT_AUTH` conditional block.

cheenamalhotra (approved):

PR #1271, opened 2021-09-21
Read makecert.exe path from config.json + AdapterTest improvements

johnnypham:
These changes are required as we migrate pipelines to 1ES hosted pools.

- path to makecert.exe can be configured in config.json to prevent test failures when makecert is not found in the path environment variable
- improve random names of database objects in AdapterTest


src/Microsoft.Data.SqlClient/tests/tools/Microsoft.Data.SqlClient.TestUtilities/Config.cs
```diff
@@ -38,6 +38,7 @@ public class Config
         public bool IsDNSCachingSupportedTR = false;  // this is for the tenant ring
         public string EnclaveAzureDatabaseConnString = null;
         public string UserManagedIdentityClientId = null;
+        public string MakecertPath = null;
```
cheenamalhotra, line 41: Please also update config.default.json file + Buildguide for future use.

johnnypham: Updated.

cheenamalhotra (approved):

BUILDGUIDE.md
```diff
@@ -123,6 +123,7 @@ Manual Tests require the below setup to run:
   |FileStreamDirectory | (Optional) If File Stream is enabled on SQL Server, pass local directory path to be used for setting up File Stream enabled database. |  `D:\\escaped\\absolute\\path\\to\\directory\\` |
   |UseManagedSNIOnWindows | (Optional) Enables testing with Managed SNI on Windows| `true` OR `false`|
   |IsAzureSynpase | (Optional) When set to 'true', test suite runs compatible tests for Azure Synapse/Parallel Data Warehouse. | `true` OR `false`|
+  |MakecertPath | The location of makecert.exe. This is not required if the path is present in the PATH environment variable. | Location of makecert.exe on machine |
```
DavoudEshtehari, line 126:
```suggestion
  |MakecertPath | The location of makecert.exe. This is not required if the path is present in the PATH environment variable. | Full path of makecert.exe on machine |
```

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/AdapterTest/AdapterTest.cs
```diff
@@ -853,7 +855,7 @@ public void UpdateRefreshTest()
                     {
                         ExecuteNonQueryCommand(spDropInsert);
                         ExecuteNonQueryCommand("DROP TABLE " + _tempTable);
-                        ExecuteNonQueryCommand("DROP TABLE ID_" + _tempTable);
+                        ExecuteNonQueryCommand("DROP TABLE " + identTableName);
```
DavoudEshtehari, line 858: NIT: To drop an object I'd suggest using the DataTestUtility defined functions.

DavoudEshtehari (approved):

BUILDGUIDE.md
johnnypham: I just noticed the value above for FileStreamDirectory : `D:\\escaped\\absolute\\path\\to\\directory\\`. Should we do something similar here? e.g.  `D:\\escaped\\absolute\\path\\to\\makecert.exe\\`

DavoudEshtehari: Regarding your change at [CertificateUtilityWin](https://github.com/dotnet/SqlClient/pull/1271/files#diff-a0826e84a92307f80be30889258cc82faaa5fe0d009e6fe366dea0e8c38137b4R34) you're supposed to read a full path that includes `makecert.exe`. So, if it was correct the "Location" word couldn't clearly explain your purpose here!

PR #1272, opened 2021-09-22
Drop .NET Core 2.1 (now Out of Support)

cheenamalhotra:
### .NET Core 2.1 End of life
.NET Core 2.1 reached end of life in August 2021. **Microsoft.Data.SqlClient v4.0** drops dedicated support for .NET Core 2.1, and fallbacks to .NET Standard 2.0 for consuming applications. Customers are recommended to migrate their apps to use supported versions of .NET Core 3.1+.

Reference: [.NET Support Policy](https://dotnet.microsoft.com/platform/support/policy/dotnet-core)


lcheunglci (approved):
Looks good to me

JRahnama:
LGTM.

NIT: There are a couple of places that needs to be considered  if we want to take those mentioned netcore2.1  out:
1. SNIStream.cs line 10 in the comments 
2. SqlClientEventSource line 22 in the `#region`

JRahnama (approved):

johnnypham (approved):

Issue #1261, opened 2021-09-11
Possible mergeable classes for merging code bases


DavoudEshtehari:
The main aim of this ticket is to clarify the project progress and having contribution with who is interested in it.

At the first step, we've prepared a list of possible mergeable files with their types which maybe need some pruning. Then, priority could be declared before starting to merge.

<details>
<summary>Mergable files list</summary>
<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:dt="uuid:C2F41010-65B3-11d1-A29F-00AA00C14882"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=OneNote.File>
<meta name=Generator content="Microsoft OneNote 15">
</head>

<body lang=en-US style='font-family:Calibri;font-size:11.0pt'>

<div style='direction:ltr'>

File & PR | Type | Group | Priority
-- | -- | -- | --
DbConnectionStringCommon.cs   #1265 | DbConnectionStringDefaults    DbConnectionStringKeywords    DbConnectionStringSynonyms    DbConnectionStringBuilderUtil | Common |  1  
DbConnectionOptions.Common.cs   (core)   DbConnectionOptions.cs #1279 | DbConnectionOptions    KEY    SYNONYM | Common |  1  
AdapterUtil.cs   AdapterUtil.Drivers.cs   (core)   AdapterUtil.SqlClient.cs   (core) #1306 | ADP | Common |  1  
StringsHelper.cs #1288 | StringsHelper | Common | 1 
DbConnectionClosed.cs | DbConnectionClosed    DbConnectionBusy    DbConnectionClosedBusy    DbConnectionClosedConnecting    DbConnectionClosedNeverOpened    DbConnectionClosedPreviouslyOpened    DbConnectionOpenBusy | ProviderBase |  
DbConnectionInternal.cs | DbConnectionInternal | ProviderBase |  
DbConnectionFactory.cs | DbConnectionFactory | ProviderBase |  
DbConnectionPool.cs #2812 | PendingGetConnection    PoolWaitHandles    State    TransactedConnectionList    TransactedConnectionPool | ProviderBase |  
DbConnectionPoolGroup.cs #1371 | DbConnectionPoolGroup | ProviderBase |  
[...]

PR #1280, opened 2021-09-23
Move to shared SqlCachedBuffer

lcheunglci:
Relating to issue #1261 .  I moved the SqlCacheBuffer.cs from netcore to shared src and updated the reference in the netfx csproj.  I did notice that in the `ToXmlReader` method was implemented differently between the two.  I looked at the helper method `SqlTypeWorkarounds.SqlXmlCreateSqlXmlReader(...)` and it seems to be doing the same functionality without using reflection.


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCachedBuffer.cs
```diff
@@ -140,7 +138,7 @@ public bool IsNull
         {
             get
             {
-                return (_cachedBytes == null) ? true : false;
+                return (_cachedBytes == null);
```
Wraith2, line 141: You can probably make this and a few others expression bodied properties. If the getter is a single expression or a simple ternary so it's clear from a single line of text that it doesn't I usually do so. 
I think my rule of thumb is if I feel I'd need a line break or it's likely to need scrolling to see the end of the expression I'll leave it as statement body. See what you think.

lcheunglci: Thanks! I've add a new commit to change those getter accessors to single line expression bodied definitions.

lcheunglci:
I decided to remove the `ToXmlReader` method because there's no references to it anywhere in the code and tests.  After further comparison between the reflection implementation in netfx and the helper method in netcore, they both seems different but since it's an internal method and it's not being used, I feel like it's safe to remove.

Wraith2:
Seems reasonable. We'll see id it affects any of the tests (which sometimes use private reflection) and if nothing is broken then it'll be a good cleanup.

johnnypham (approved):

Kaur-Parminder (approved):

Wraith2:
Unless there are merge conflicts I wouldn't suggest keeping all your PR's up to date with main. It will take you a lot of time and it can be months before the team are ready to review.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCachedBuffer.cs
```diff
@@ -58,7 +56,7 @@ internal static bool TryCreate(SqlMetaDataPriv metadata, TdsParser parser, TdsPa
                     break;
```
DavoudEshtehari, line 56: NIT: Even though it's not mandatory for single-line if-statements, it could prevent misunderstanding, and I suggest using braces even for single-line statements like this.
```suggestion
                {
                    break;
                }
```

DavoudEshtehari (commented):
Modifiers could be reordered on [ToString()](https://github.com/dotnet/SqlClient/pull/1280/files#diff-2e044cae1bdfcabc70deb9c953279908321f4739ea4d6a35f323e787955d135fL105): `public override string ToString()`

DavoudEshtehari (approved):

PR #1152, opened 2021-07-06
FIX | Add Component Model annotations to SqlConnectionStringBuilder class in netcore's  src folder

JRahnama:
fixes #1142

PR #1284, opened 2021-09-24
Move into Shared for SqlCommandBuilder.cs

lcheunglci:
Relates to #1261 .  I merged the SqlCommandBuilder.cs from netfx into netcore and then move it into the shared src.  The netfx version uses `System.ComponentModel` and some attributes from it, which I assume were for WPF or WinForms and there were some DEBUG ifdefs, so I ifdef'ed them for NETFX.


Wraith2:
I'm pretty sure we want the designer attributes in all builds. The difference may have come from the early version of netcore not having those attributes but since we don't support anything before 3.1 now if those attributes are present we should keep them. 

The only problem you might have is if the resources they use are localized and in that case we'll have to see what the MS team want to do, I know there's a long term goal for getting the netfx localized resources working in netcore.

lcheunglci:
That's good to know. In that case, I should wait for #1272 and merge those in this PR before I remove the ifdefs

Wraith2:
It'll depend on the attribute, if you check them on apisof.net for example [DesignerSerializationVisibilityAttribute](https://apisof.net/catalog/0a8c0b1e-1f65-8a3b-1ba0-727df90892f5) you can see it's in NS2.0 and netcore 2 onwards so it should be ok to do now. Perhaps just take out the iffdefs locally and see if it compiles?

lcheunglci:
Thanks @Wraith2 , it looks like majority of the attributes builds fine with the exception of `ResDescriptionAttribute(StringsHelper.ResourceNames.SqlCommandBuilder_DataAdapter)` where the string doesn't exist in the StringHelper in the netcore project. I wonder if it's ok to copy it out of the `Strings.resx` from netfx as a string literal or should I leave it with the ifdef NETFX

Wraith2:
Hmm, have a look at https://github.com/dotnet/SqlClient/pull/1152 and see if there's any precedent there that can help. I'd say just add the resource.

src/Microsoft.Data.SqlClient/netcore/src/Resources/Strings.Designer.cs
```diff
@@ -3525,6 +3525,15 @@ internal class Strings {
...
         
+        /// <summary>
+        ///   Looks up a localized string similar to The DataAdapter for which to automatically generate SqlCommands..
+        /// </summary>
+        internal static string SqlCommandBuilder_DataAdapter {
+            get {
+                return ResourceManager.GetString("SqlCommandBuilder_DataAdapter", resourceCulture);
+            }
+        }
+
```
DavoudEshtehari, line 3536: We don't include the resources like this in `netcore` which are just used in attributes; They're supposed to be merged in the future and we have a temporary place for them. You can add a constant for it in [StringsHelper.ResourceNames](https://github.com/lcheunglci/SqlClient/blob/MergeShared-SqlCommandBuilder/src/Microsoft.Data.SqlClient/netcore/src/Resources/StringsHelper.cs#L75). So, you can revert this change.

src/Microsoft.Data.SqlClient/netcore/src/Resources/Strings.resx
```diff
@@ -1932,4 +1932,7 @@
   <data name="SQL_ParameterDirectionInvalidForOptimizedBinding" xml:space="preserve">
     <value>Parameter '{0}' cannot have Direction Output or InputOutput when EnableOptimizedParameterBinding is enabled on the parent command.</value>
   </data>
-</root>
\ No newline at end of file
+  <data name="SqlCommandBuilder_DataAdapter" xml:space="preserve">
+    <value>The DataAdapter for which to automatically generate SqlCommands.</value>
+  </data>
+</root>
```
DavoudEshtehari, line 1938: It's not required. Please revert it.

src/Microsoft.Data.SqlClient/netcore/src/Resources/StringsHelper.cs
```diff
@@ -158,6 +158,7 @@ internal class ResourceNames
             internal const string TCE_DbConnectionString_AttestationProtocol = @"Specifies an attestation protocol for its corresponding enclave attestation service.";
             internal const string TCE_DbConnectionString_IPAddressPreference = @"Specifies an IP address preference when connecting to SQL instances.";
             internal const string SqlConnection_ServerProcessId = @"Server Process Id (SPID) of the active connection.";
+            internal const string SqlCommandBuilder_DataAdapter = @"The DataAdapter for which to automatically generate SqlCommands.";
```
DavoudEshtehari, line 161: It's not required. Please revert it.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommandBuilder.cs
```diff
@@ -14,25 +15,31 @@
 
 namespace Microsoft.Data.SqlClient
 {
-    /// <include file='../../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlCommandBuilder.xml' path='docs/members[@name="SqlCommandBuilder"]/SqlCommandBuilder/*'/>
+    /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlCommandBuilder.xml' path='docs/members[@name="SqlCommandBuilder"]/SqlCommandBuilder/*'/>
+    [DesignerCategory("")]
```
DavoudEshtehari, line 19: NIT: The default constructor has the same behavior.
```suggestion
    [DesignerCategory()]
```

src/Microsoft.Data.SqlClient/netfx/src/Microsoft.Data.SqlClient.csproj
```diff
@@ -288,6 +288,10 @@
     <Compile Include="..\..\src\Microsoft\Data\SqlClient\SqlColumnEncryptionKeyStoreProvider.cs">
       <Link>Microsoft\Data\SqlClient\SqlColumnEncryptionKeyStoreProvider.cs</Link>
     </Compile>
+    <Compile Include="..\..\src\Microsoft\Data\SqlClient\SqlCommandBuilder.cs">
+      <Link>Microsoft\Data\SqlClient\SqlCommandBuilder.cs</Link>
+      <SubType>Component</SubType>
```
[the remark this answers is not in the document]
lcheunglci: I'm actually not sure why it's added.  It's automatically added by visual studios whenever build/save the project, so I end up leaving it there.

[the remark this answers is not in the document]
lcheunglci: Nevermind, it looks like I am able to remove it and it doesn't get automatically added back after build now.

src/Microsoft.Data.SqlClient/netcore/src/Resources/StringsHelper.cs
lcheunglci: Without this it does not compile with the merged [ResDescriptionAttribute(StringsHelper.ResourceNames.SqlCommandBuilder_DataAdapter)]

src/Microsoft.Data.SqlClient/netfx/src/Microsoft.Data.SqlClient.csproj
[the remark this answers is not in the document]
Wraith2: It does add it back in and I remember fighting against VS re-adding it before and we found some way to prevent it happening. I'll try to find the PR and see if we need to do the same thing on netfx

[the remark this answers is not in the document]
lcheunglci: That explains why I was able to remove it this time because `[DesignerCategory()]` was there on line 19

DavoudEshtehari (approved):

Issue #1290, opened 2021-09-27
FormatException is thrown when opening a connection if Tracing is enabled


NadeemAfana:
### Describe the bug
An exception of type `FormatException` is thrown intermittently when opening a connection.

The method overload `TraceEvent<T0, T1>(string message, T0 args0, T1 args1, [System.Runtime.CompilerServices.CallerMemberName] string memberName = "")` takes precedence over the generic overload `TraceEvent<T0, T1, T2>(string message, T0 args0, T1 args1, T2 args2)` when the last argument is a string. An exception of type `System.FormatException: Index (zero based) must be greater than or equal to zero and less than the size of the argument list.` is thrown, for example, from the method call in `TryGetFedAuthTokenLocked`.

```
Exception message: Index (zero based) must be greater than or equal to zero and less than the size of the argument list.
Stack trace:
    at System.Text.StringBuilder.AppendFormatHelper(IFormatProvider provider, String format, ParamsArray args)
   at System.String.FormatHelper(IFormatProvider provider, String format, ParamsArray args)
   at Microsoft.Data.SqlClient.SqlClientEventSource.TraceEvent[T0,T1](String message, T0 args0, T1 args1, String memberName) in H:\\tsaagent1\\_work\\18\\s\\src\\Microsoft.Data.SqlClient\\src\\Microsoft\\Data\\SqlClient\\SqlClientEventSource.cs:line 363
   at Microsoft.Data.SqlClient.SqlInternalConnectionTds.TryGetFedAuthTokenLocked(SqlFedAuthInfo fedAuthInfo, DbConnectionPoolAuthenticationContext dbConnectionPoolAuthenticationContext, SqlFedAuthToken& fedAuthToken) in H:\\tsaagent1\\_work\\18\\s\\src\\Microsoft.Data.SqlClient\\netfx\\src\\Microsoft\\Data\\SqlClient\\SqlInternalConnectionTds.cs:line 2712
   at Microsoft.Data.SqlClient.SqlInternalConnectionTds.OnFedAuthInfo(SqlFedAuthInfo fedAuthInfo) in H:\\tsaagent1\\_work\\18\\s\\src\\Microsoft.Data.SqlClient\\netfx\\src\\Microsoft\\Data\\SqlClient\\SqlInternalConnectionTds.cs:line 2636
[...]

PR #1291, opened 2021-09-27
Fixes FormatException when Tracing is enabled

NadeemAfana:
This is a fix for the bug #1290 by simply removing the redundant argument from the overload.


cheenamalhotra (approved):

lcheunglci (approved):
Looks good to me

cheenamalhotra:
The use case is related to refreshing access token before expiry which will be a long wait in test suite.
The changes look reasonable, we should prevent overriding of methods with optional params when using the generic types. The right method that should have been called in this case is on L371.

Wraith2:
Ok, making the tests take even longer isn't a good idea.

JRahnama (approved):

JRahnama:
@shueybubbles I will add this to our list.

PR #1292, opened 2021-09-27
Move to Shared for SqlConnectionPoolKey.cs

lcheunglci:
Relates to #1261 . Merged the netfx version of SqlConnectionPoolKey.cs into netcore, and move it to the share src. I did notice that the netfx uses the `ICloneable` interface and the netcore version overrides the `Clone` method, and I went with the netcore version and verified it still compiled successfully on netfx.


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionPoolKey.cs
```diff
@@ -102,18 +111,19 @@ internal SqlClientOriginalNetworkAddressInfo OriginalNetworkAddressInfo
...
-                    _originalNetworkAddressInfo == key._originalNetworkAddressInfo);
+            return (obj is SqlConnectionPoolKey key
+                && _credential == key._credential
+                && ConnectionString == key.ConnectionString
+#if NETFRAMEWORK
+                && _serverCertificateValidationCallback == key._serverCertificateValidationCallback
+                && _clientCertificateRetrievalCallback == key._clientCertificateRetrievalCallback
+                && _originalNetworkAddressInfo == key._originalNetworkAddressInfo
+#endif
+                && string.CompareOrdinal(_accessToken, key._accessToken) == 0);
```
johnnypham, line 126: This last condition should come before the #if block?

johnnypham (approved):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionPoolKey.cs
lcheunglci: Thanks! I've moved the access token check to line 121 before the ifdef block in the next commit.

DavoudEshtehari:
My preference is to avoid `#if #else` directives as much as possible. Please, see this [comment](https://github.com/dotnet/SqlClient/issues/1261#issuecomment-933769626) for more explanation.

cheenamalhotra:
I agree, please combine code together in a single if-def as much as possible, or in a separate partial class with `*****.netfx.cs` / `*****.netcore.cs` file, depending on the amount of per-framework code.

lcheunglci:
Thanks! I revised my changes based on @DavoudEshtehari comments from https://github.com/dotnet/SqlClient/issues/1261#issuecomment-933838859 renaming the shared src with the `.Common.cs` prefix and made it a partial class so that the netcore and netfx version `SqlConnectionPoolKey.cs` can be back in its original location without using ifdefs

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionPoolKey.cs
```diff
@@ -148,5 +127,49 @@ private void CalculateHashCode()
...
+            }
+            else if (_accessToken != null)
+            {
+                unchecked
+                {
+                    _hashValue = _hashValue * 17 + _accessToken.GetHashCode();
+                }
+            }
+        }
+#endif // NETFRAMEWORK
```
DavoudEshtehari, line 173: NIT: I see it's the end of the Netfx **#if-directive**. Instead of this comment I'd suggest wrapping each section in a `#region` and `#endregion` block.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionPoolKey.cs
```diff
@@ -79,49 +45,62 @@ internal string AccessToken
...
-            get
-            {
-                return _clientCertificateRetrievalCallback;
-            }
+            _credential = key.Credential;
+            _accessToken = key.AccessToken;
+            _serverCertificateValidationCallback = key._serverCertificateValidationCallback;
+            _clientCertificateRetrievalCallback = key._clientCertificateRetrievalCallback;
+            CalculateHashCode();
         }
```
DavoudEshtehari, line 65: This function could be taken out to the common section. Plz, follow the [old commit](https://github.com/dotnet/SqlClient/pull/1292/commits/4d1f09530317d3ac5264b183ed1dd94db7af7041#diff-813dd48970098a3e3745790fdd8a33f0a9d98162936ebe899fbec804d91f19edR51) and avoid redundancy.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionPoolKey.cs
```diff
@@ -79,49 +45,62 @@ internal string AccessToken
...
-                    _clientCertificateRetrievalCallback == key._clientCertificateRetrievalCallback &&
-                    _originalNetworkAddressInfo == key._originalNetworkAddressInfo);
+            return (obj is SqlConnectionPoolKey key
+                && _credential == key._credential
+                && ConnectionString == key.ConnectionString
+                && string.CompareOrdinal(_accessToken, key._accessToken) == 0
+                && _serverCertificateValidationCallback == key._serverCertificateValidationCallback
+                && _clientCertificateRetrievalCallback == key._clientCertificateRetrievalCallback
+                && _originalNetworkAddressInfo == key._originalNetworkAddressInfo);
         }
```
DavoudEshtehari, line 92: This function could be taken out to the common section; like [here](https://github.com/dotnet/SqlClient/pull/1292/commits/4d1f09530317d3ac5264b183ed1dd94db7af7041#diff-813dd48970098a3e3745790fdd8a33f0a9d98162936ebe899fbec804d91f19edR116).

lcheunglci:
> NIT: I'd suggest to follow the same pattern for all the similar changes in a file; My preference is to see the final if-else block at the end right before closing the class definition.

I did consider that but I found that having the fields, properties and constructors at the top easier to follow than to find it mixed at the end.

JRahnama (approved):

Issue #1264, opened 2021-09-13
SqlDependency.Start throws NRE against Azure SQL Database


ascott18:
### Describe the bug

Calling `SqlDependency.Start` against an Azure SQL Database produces a NRE, rather than `SQL.SqlDependencyDatabaseBrokerDisabled()`

This is because the query to detect the service broker returns zero rows on Azure SQL, rather than a single row with a boolean. `SqlDependencyListener` is casting `(bool)null`.

```
Exception message:
System.NullReferenceException: 'Object reference not set to an instance of an object.'
Stack trace:
   at System.Runtime.CompilerServices.CastHelpers.Unbox(Void* toTypeHnd, Object obj)
   at SqlDependencyProcessDispatcher.SqlConnectionContainer..ctor(SqlConnectionContainerHashHelper hashHelper, String appDomainKey, Boolean useDefaults) in /_/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDependencyListener.cs:line 103
   at SqlDependencyProcessDispatcher.Start(String connectionString, String& server, DbConnectionPoolIdentity& identity, String& user, String& database, String& queueService, String appDomainKey, SqlDependencyPerAppDomainDispatcher dispatcher, Boolean& errorOccurred, Boolean& appDomainStart, Boolean useDefaults) in /_/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDependencyListener.cs:line 1592
   at SqlDependencyProcessDispatcher.Start(String connectionString, String queue, String appDomainKey, SqlDependencyPerAppDomainDispatcher dispatcher) in /_/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDependencyListener.cs:line 1528
   at Microsoft.Data.SqlClient.SqlDependency.Start(String connectionString, String queue, Boolean useDefaults) in /_/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDependency.cs:line 489
   at Microsoft.Data.SqlClient.SqlDependency.Start(String connectionString, String queue) in /_/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDependency.cs:line 394
```

### To reproduce

```c#
string connectionString = "<any azure sql database>"
SqlDependency.Start(connectionString);
```

### Expected behavior
[...]

PR #1294, opened 2021-09-27
FIX | db_id returns null on Azure SQL Database.

JRahnama:
closes Issue #1264.

`db_id()` returns null on Azure SQL Database. The proposed fix adds a nullable boolean and assign the values from the reader to it.

Also `ExecuteScalar`  is replaced with `ExecuteReader(CommandBehavior.SingleRow)`. 

The reason to do so was avoiding the cast on the result of  returned object  from `ExecuteScalar` and since db_id() returns the database ID of the current database we only need one row.


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDependencyListener.cs
```diff
@@ -100,7 +101,15 @@ internal SqlConnectionContainer(SqlConnectionContainerHashHelper hashHelper, str
...
                 };
 
-                if (!(bool)_com.ExecuteScalar())
+                // db_id() returns the database ID of the current database hence it will always be one line result
+                using SqlDataReader reader = _com.ExecuteReader(CommandBehavior.SingleRow);
+                if (reader.Read() && reader[0] is not null)
+                {
+                    dbId = reader.GetBoolean(0);
+                }
+                reader.Close();
```
cheenamalhotra, line 110: Can we using bracketed `using` for `reader`? You wouldn't need to call this in that case.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlDependencyListener.cs
```diff
@@ -138,11 +138,20 @@ internal SqlConnectionContainer(SqlConnectionContainerHashHelper hashHelper, str
...
-                if (!(bool)_com.ExecuteScalar())
+                // db_id() returns the database ID of the current database hence it will always be one line result
+                using SqlDataReader reader = _com.ExecuteReader(CommandBehavior.SingleRow);
+                if (reader.Read() && reader[0] is not null)
+                {
+                    dbId = reader.GetBoolean(0);
+                }
+                reader.Close();
+
+                if (dbId is null || !dbId.Value)
```
cheenamalhotra, line 154: Can we add test for code that currently throws NRE?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDependencyListener.cs
Wraith2: In general I'm not a fan of the new unscoped using syntax. I can understand that in quick projects it has some value but for highly formal code like a base library I think it's better to be explicit, what do you think?

JRahnama: Agreed. I will fix it in the next commit.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlDependencyListener.cs
Wraith2: This would be shorter as `!value.GetValueOrDefault()`, I'm not sure if it's clearer and perf isn't a major concern just a general one.

JRahnama: Regarding the test, it needs more investigation as this only happen on Azure SQL DB and even in this case it may return null.

GetValueOrDefault returns a value even if the HasValue property is false. This also needs more cautious investigation performance wise and also returned value as well.

Wraith2: > GetValueOrDefault returns a value even if the HasValue property is false

Yup, and it does so in a single call with a smaller number of instructions than doing the null check then value check. As long as you have that leading negation it would work. But as I said it's hardly a performance problem so leave it as it already is.

ascott18:
Copying this over from #1264: The issue is not that db_id() returns null - it doesn't. The issue is that, [as documented](https://docs.microsoft.com/en-us/sql/t-sql/functions/db-id-transact-sql?view=sql-server-ver15#remarks), its value is nonsensical on Azure SQL.

image

David-Engel (approved):

DavoudEshtehari (approved):

DavoudEshtehari:
Please, resolve the conflict.

PR #1288, opened 2021-09-25
Merging common code bases StringsHelper

DavoudEshtehari:
- Related issue #1261 
- Common type: **StringsHelper**

PR #1299, opened 2021-09-28
Move to Shared for SqlDependency.cs

lcheunglci:
Relates to #1261 .  I merged the netfx version of SqlDependency into the netcore version and move it to the shared src.  I was originally going to allow the resource attributes but that would require an update to the StringsHelper which is currently being merged in #1288 so I left it with the ifdef NETFRAMEWORK, but once that PR is merged, I could revised this.  Also I updated the file to conform the coding style but I left two IDE0063 info messages because I found the using statement in line 512 and 545 more readable, but the intellisense recommended change is with C# 8 syntax without the extra scope body, so if that is preferred I am more than happy to change it.


Wraith2:
I much prefer usings with scope. If the MS team agree I think we'd need to update the repo editor config to make it stop suggesting the implicitly scoped version.

johnnypham:
> I much prefer usings with scope. If the MS team agree I think we'd need to update the repo editor config to make it stop suggesting the implicitly scoped version.
I have no preference as long as it's readable.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlDependency.cs
```diff
@@ -6,7 +6,16 @@
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Globalization;
+#if NETFRAMEWORK
```
johnnypham, line 9: Imo, it's a little neater to have one #if for netfx, rather than trying to maintain alphabetical order.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlDependency.cs
```diff
@@ -305,11 +380,17 @@ public SqlDependency(SqlCommand command, string options, int timeout)
                         {
                             if (_dependencyFired)
                             { // If fired, fire the new event immediately.
+#if NETFRAMEWORK
```
johnnypham, line 383: Should these events be netfx only? I think it was accidentally left out from netcore.

lcheunglci: I don't think it affects netcore since it's only adds extra logging, I'll remove the ifdef and update the string format to keep it consistent in the next commit.

johnnypham (approved):

Kaur-Parminder (approved):

PR #1300, opened 2021-09-29
Move to Shared for SqlException.cs

lcheunglci:
Relates to #1261 .  Merged the SqlException.cs from netfx into netcore, move it into the shared src and updated it to conform with the coding style.


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlException.cs
```diff
@@ -82,67 +85,69 @@ public SqlErrorCollection Errors
...
             get
             {
-                return this._clientConnectionId;
+                return _clientConnectionId;
             }
         }
 
+#if NETFRAMEWORK
         /*virtual protected*/
         private bool ShouldSerializeErrors()
```
johnnypham, line 99: I couldn't find this method used anywhere. Let's remove it?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlException.cs
```diff
@@ -82,67 +85,69 @@ public SqlErrorCollection Errors
...
         { // MDAC 65548
             return ((null != _errors) && (0 < _errors.Count));
         }
+#endif
 
-        /// <include file='..\..\..\..\..\..\..\doc\snippets\Microsoft.Data.SqlClient\SqlException.xml' path='docs/members[@name="SqlException"]/Class/*' />
+        /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlException.xml' path='docs/members[@name="SqlException"]/Class/*' />
         public byte Class
         {
             get { return Errors.Count > 0 ? Errors[0].Class : default; }
```
johnnypham, line 108: Since you mentioned style, most of these getters can be simplified.

johnnypham (approved):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlException.cs
lcheunglci: sounds good, I'll do that in the next commit.

PR #1306, opened 2021-10-01
Merging common code bases AdapterUtil

DavoudEshtehari:
- Related issue #1261.
- Common type: **ADP**
  - All the common code-bases are pruned and moved into a new file with the best implementation with new styles.
  - Each project has its own specific implementation in a separate file.
  - By a minor change in StringsHelper I managed to replace `Format()` with `GetString()` for current functions in **NetCore**, and avoided splitting code.


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
```diff
@@ -617,7 +617,7 @@ public override string CommandText
             get
             {
                 string value = _commandText;
-                return ((null != value) ? value : ADP.StrEmpty);
+                return ((null != value) ? value : ADP.s_strEmpty);
```
Wraith2, line 620: Can we just get rid of empty string constants in favour of using "" or string.Empty?

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/Common/SafeNativeMethods.cs
```diff
@@ -143,7 +143,7 @@ sealed internal class Wrapper
             [ResourceConsumption(ResourceScope.Process, ResourceScope.Process)]
             static internal void ClearErrorInfo()
             { // MDAC 68199
-                SafeNativeMethods.SetErrorInfo(0, ADP.PtrZero);
+                SafeNativeMethods.SetErrorInfo(0, ADP.s_ptrZero);
```
Wraith2, line 146: Same as empty string, there's really no point in using a reference to a variable which is zero instead of just using the literal zero and letting the compiler and jit see it as a const.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/ProviderBase/DbConnectionPool.cs
```diff
@@ -485,7 +485,7 @@ protected override bool ReleaseHandle()
                             DbConnectionPoolIdentity identity,
                             DbConnectionPoolProviderInfo connectionPoolProviderInfo)
         {
-            Debug.Assert(ADP.IsWindowsNT, "Attempting to construct a connection pool on Win9x?");
+            Debug.Assert(ADP.s_isWindowsNT, "Attempting to construct a connection pool on Win9x?");
```
Wraith2, line 488: I know it's just following the naming rules but it looks bad. Can it be made a property or perhaps just not follow the naming convention in some cases like this for readability?

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/ProviderBase/DbConnectionPoolGroup.cs
```diff
@@ -189,7 +189,7 @@ internal DbConnectionPool GetConnectionPool(DbConnectionFactory connectionFactor
             DbConnectionPool pool = null;
             if (null != _poolGroupOptions)
             {
-                Debug.Assert(ADP.IsWindowsNT, "should not be pooling on Win9x");
+                Debug.Assert(ADP.s_isWindowsNT, "should not be pooling on Win9x");
```
Wraith2, line 192: Is Win9x supported by this library? If not do we need the assert? and if we don't can get get rid of IsWindowsNT since everything we support is windows nt based?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
DavoudEshtehari: I wanted to remove it if there wasn't the [comment](https://github.com/dotnet/SqlClient/pull/1306/files#diff-28fd361d833bb12ffb3d5462abcd11f1fa22c9157c923510e76bf1c4c9c0932eR512) over its definition. My question is how's it possible for who has an answer!?

from the code:
```
 // security issue, don't rely upon public static readonly values
 internal static readonly string s_strEmpty = ""; // String.Empty
```

Wraith2: All instances of "" will be interned by the runtime, if you have one you have all of them. The only way to create a different one would be to to use string.Create() in a way that prevents the interned string check. So even though we intend to have our own copy of the constant we won't, we'll still get the exact same one as is publicly available.

Moreover the only ways to change a string are unsafe and unsupported so if you do any of them you put the entire runtime in a state it isn't going to be able to cope with. If you somehow changes "" to be non-empty you would break many many other pieces of software.

lcheunglci (approved):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/AdapterUtilCommon.cs
```diff
@@ -0,0 +1,1202 @@
+// Licensed to the .NET Foundation under one or more agreements.
```
cheenamalhotra, line 1: Can this file name stay "AdapterUtil"? Any reason to call it AdapterUtilCommon?

cheenamalhotra (approved):
Overall looks good to me.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/AdapterUtilCommon.cs
Wraith2: If there are AdapterUtil.cs in the specific projects it would be `AdapterUtil.common.cs` following @DavoudEshtehari 's logic in https://github.com/dotnet/SqlClient/issues/1261#issuecomment-933838859

DavoudEshtehari: Apparently, I missed the dot. 😉

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/Common/SafeNativeMethods.cs
Wraith2: If we could use net6 then we'd want to use `(nint)0` but we can't so either `new IntPtr(0)` or `IntPtr.Zero` so we don't need our own constant.

DavoudEshtehari: I obsess about this change and [taking out the Win9x assessment](https://github.com/dotnet/SqlClient/pull/1306#discussion_r719870325), and I'm not sure if it's a good idea to include them in this PR if they already have been approved.
cc @cheenamalhotra

Wraith2: The objective is to merge. If it's not clear that changing to use IntPtr or taking out those assertions is correct at this time then I'm fine with leaving them in and coming back later, they aren't incorrect or causing a problem.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
DavoudEshtehari: Done. I preferred to use `""` to avoid the suspected security issue.

johnnypham (approved):

PR #1308, opened 2021-10-04
Move into Shared SqlDependencyListener.cs

lcheunglci:
Relates to #1261 . Merged netfx to netcore and move it to shared src. While cleaning up the coding style errorlist, I left 2 IDE0063 (Line 957 and Line 1076) info messages because I chose to keep the scope bodied using statement and there was a IDE0079 suggesting that I remove the suppression of `CA2100:ReviewSqlQueriesForSecurityVulnerabilities` on Line 968, which I left.  I may need to merge PR #1299 because there's a small change to the property to the property to retrieve the static instance of `SingletonProcessDispatcher`.


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlDependencyListener.cs
```diff
@@ -976,6 +1031,13 @@ private void TearDownAndDispose()
                     {
                         _stopped = true;
                         _con.Dispose(); // Close and dispose connection.
+#if NETFRAMEWORK
+                        //dispose windows identity
+                        if (_windowsIdentity != null)
+                        {
+                            _windowsIdentity.Dispose();
+                        }
```
johnnypham, line 1039:
```suggestion
                        _windowsIdentity?.Dispose();
```

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlDependencyListener.cs
```diff
@@ -645,7 +676,31 @@ private void Restart(object unused)
...
+#endif 
                             _con.Open();
+#if NETFRAMEWORK
+                                }
+                                finally
+                                {
+                                    if (null != context)
+                                    {
+                                        context.Undo();
+                                    }
```
johnnypham, line 696:
```suggestion
                                    context?.Undo();
```

johnnypham:
Some events were using `SqlClientEventSource.Log.TryNotificationTraceEvent` in netfx and `SqlClientEventSource.Log.TryTraceEvent` in netcore. Not sure which one we should use. Maybe someone else can comment.

lcheunglci:
> Some events were using `SqlClientEventSource.Log.TryNotificationTraceEvent` in netfx and `SqlClientEventSource.Log.TryTraceEvent` in netcore. Not sure which one we should use. Maybe someone else can comment.

I noticed that too. I'm not sure if we need to unify those and if so which one should we be using.

cheenamalhotra (changes requested):
Please reset branch to contain only PR related changes.

PR #1312, opened 2021-10-04
Fix | Disable encryption when connecting to SQL Local DB

cheenamalhotra:
This change disables encryption when connection to SQL Local DB is being made as SQL Local DB does not support encryption.

### DIFF
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/LocalDBTest/LocalDBTest.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/LocalDBTest/LocalDBTest.cs
@@ -25,6 +25,15 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests
 
         [SkipOnTargetFramework(TargetFrameworkMonikers.Uap)] // No Registry support on UAP
         [ConditionalFact(nameof(IsLocalDBEnvironmentSet))]
+        public static void LocalDBEncryptionNotSupportedTest()
+        {
+            // Encryption is not supported by SQL Local DB.
+            // But connection should succeed as encryption is disabled by driver.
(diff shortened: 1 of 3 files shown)


src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/LocalDBTest/LocalDBTest.cs
```diff
@@ -87,6 +118,7 @@ private static void OpenConnection(string connString)
         {
             using SqlConnection connection = new(connString);
             connection.Open();
+            Assert.Equal(System.Data.ConnectionState.Open, connection.State);
```
JRahnama, line 121: Can we instead use `try/catch` block to get a better exception handling of the thrown exception?

JRahnama:
LGTM. Just a minor suggestion.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/LocalDBTest/LocalDBTest.cs
cheenamalhotra: There is no exception expected here.

JRahnama: Correct me if I am wrong, if the assertion fails which means the connection state is not open, therefor an exception will be thrown on running the command.

cheenamalhotra: Yep - that's a test failure, so assert will fail.

JRahnama (approved):

johnnypham (approved):

PR #1313, opened 2021-10-04
Move to Shared - SqlSer.cs

Kaur-Parminder:
This is part of Code merge for issue #1261. I have created .common.cs and moved the uncommon code to private methods.


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/Server/SqlSer.cs
```diff
@@ -151,72 +30,19 @@ internal static SqlUserDefinedTypeAttribute GetUdtAttribute(Type t)
...
+        private void SerializeFrameworkSpecific(Stream s, object o)
         {
             BinaryWriter w = new BinaryWriter(s);
             ((IBinarySerialize)o).Write(w);
         }
 
-        // Prevent inlining so that reflection calls are not moved
-        // to a caller that may be in a different assembly that may
-        // have a different grant set.
-        [MethodImpl(MethodImplOptions.NoInlining)]
```
Wraith2, line 218: This attribute and it's explanation seem important to keep.

Kaur-Parminder: @Wraith2 This method, attribute and comment has been moved to SqlSer.Common.cs

Kaur-Parminder: @Wraith2  This methods, attribute and comment has been moved to SqlSer.Common.cs

Wraith2: Ok. Diff comprehension failure on my part then.

Kaur-Parminder: Understood.

Kaur-Parminder:
> There's not that much code remaining in the framework specific files. Why don't we just move it to the common file?

@johnnypham I just removed the netfx and netcore classes and updated SqlSer.cs in common project with the three functions. 
1. Both Serialize and DeSerialize functions uses If defs, added return statements in if block.
2 .GetUdtAttribute function in netfx used Reflection to throw same exception so just kept the .NetCore one.

JRahnama (approved):

DavoudEshtehari (commented):
The `SerializationHelperSql9` could be sealed: `internal sealed class SerializationHelperSql9`
Also, I see some minor possible improvements like `unnecessary assignment` and `redundant member`. You can find them in the `error list` window.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SqlSer.cs
```diff
@@ -209,7 +207,16 @@ internal BinarySerializeSerializer(Type t) : base(t)
         public override void Serialize(Stream s, object o)
         {
             BinaryWriter w = new BinaryWriter(s);
+
+#if NETFRAMEWORK
+            if (o is Microsoft.SqlServer.Server.IBinarySerialize)
+            {
+                ((SqlServer.Server.IBinarySerialize)o).Write(w);
+                return;
+            }
```
DavoudEshtehari, line 216:
```suggestion
            if (o is Microsoft.SqlServer.Server.IBinarySerialize bs)
            {
                bs.Write(w);
                return;
            }
```

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SqlSer.cs
```diff
@@ -220,8 +227,17 @@ public override object Deserialize(Stream s)
...
             object instance = Activator.CreateInstance(_type);
             BinaryReader r = new BinaryReader(s);
-            ((IBinarySerialize)instance).Read(r);
+
+#if NETFRAMEWORK
+            if (instance is Microsoft.SqlServer.Server.IBinarySerialize)
+            {
+                ((SqlServer.Server.IBinarySerialize)instance).Read(r);
+                return instance;
+            }
```
DavoudEshtehari, line 236:
```suggestion
            if (instance is Microsoft.SqlServer.Server.IBinarySerialize bs)
            {
                bs.Read(r);
                return instance;
            }
```

DavoudEshtehari (commented):
I'd suggest using `Dictionary` instead of `Hashtable` to avoid boxing and unboxing for [s_types2Serializers](https://github.com/dotnet/SqlClient/blob/04593ff73078c40e3bf83d231b44be935455763b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SqlSer.cs#L54):
- `private static ConcurrentDictionary<Type, Serializer> s_types2Serializers;`

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SqlSer.cs
```diff
@@ -50,13 +52,14 @@ internal static void Serialize(Stream s, object instance)
...
-        [ThreadStatic]
-        private static Hashtable s_types2Serializers;
+        private static ConcurrentDictionary<Type, Serializer> s_types2Serializers;
 
         private static Serializer GetSerializer(Type t)
         {
             if (s_types2Serializers == null)
-                s_types2Serializers = new Hashtable();
+            {
+                s_types2Serializers = new();
```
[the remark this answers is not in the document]
Wraith2: The rule used in /runtime is that use of `var` or `new()` is only permitted if the type is duplicated in the same statement. So either `var s_types2Serializers = new ConcurrentDictionary<Type, Serializer>()` or `ConcurrentDictionary<Type, Serializer> s_types2Serializers = new ()` would be allowed (with var preferred I think) but not the form used here.

If you rebase to master the style rules were updated in a recent PR to remove the new() fixer suggestion in the main library projects.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SqlSer.cs
```diff
@@ -4,14 +4,15 @@
 
 using System;
 using System.Collections;
```
DavoudEshtehari, line 6:
```suggestion
```

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SqlSer.cs
```diff
@@ -50,16 +52,21 @@ internal static void Serialize(Stream s, object instance)
...
+            Serializer s = null;
             if (s_types2Serializers == null)
-                s_types2Serializers = new Hashtable();
+            {
+                s_types2Serializers = new ConcurrentDictionary<Type, Serializer>();
+            }
 
-            Serializer s = (Serializer)s_types2Serializers[t];
-            if (s == null)
+            if (s_types2Serializers.ContainsKey(t))
```
DavoudEshtehari, line 65: This `if-else` block could be replaced with the following block:
```
if(!s_types2Serializers.TryGetValue(t, out s))
{
    s = GetNewSerializer(t);
    s_types2Serializers[t] = s;
}
```

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SqlSer.cs
```diff
@@ -50,16 +52,21 @@ internal static void Serialize(Stream s, object instance)
         //
         // Use a per-thread cache, so that there are no synchronization
         // issues when accessing cache entries from multiple threads.
-        [ThreadStatic]
-        private static Hashtable s_types2Serializers;
+        private static ConcurrentDictionary<Type, Serializer> s_types2Serializers;
 
         private static Serializer GetSerializer(Type t)
         {
+            Serializer s = null;
```
DavoudEshtehari, line 59: NIT: Unnecessary assignment.
```suggestion
            Serializer s;
```

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SqlSer.cs
Kaur-Parminder: TryGetValue will return Default value if there is no value for the key, condition will never be false and compiler will not enter if block. Although, I think I should keep the check if(s==null) after contains key check because we don't check  null value before entry to dictionary.

DavoudEshtehari (approved):

PR #1314, opened 2021-10-04
Move into Shared SqlNotificationEventArgs

lcheunglci:
Relates to #1261. Merge the netcore version into shared src and update the reference in the csproj.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
@@ -226,6 +226,9 @@
     <Compile Include="..\..\src\Microsoft\Data\SqlClient\SqlInfoMessageEventHandler.cs">
       <Link>Microsoft\Data\SqlClient\SqlInfoMessageEventHandler.cs</Link>
     </Compile>
+    <Compile Include="..\..\src\Microsoft\Data\SqlClient\SqlNotificationEventArgs.cs">
+      <Link>Microsoft\Data\SqlClient\SqlNotificationEventArgs.cs</Link>
+    </Compile>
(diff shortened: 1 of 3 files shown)


Wraith2:
Can this use init setters to avoid the fields entirely?

Kaur-Parminder (approved):

lcheunglci:
I just realized that C#9 introduced init-only setter.  It's a pretty neat syntax to remove the fields and constructor. I'm down with that if it's ok with everyone.

cheenamalhotra:
> I just realized that [C#9 introduced init-only setter](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/init). It's a pretty neat syntax to remove the fields and constructor. I'm down with that if it's ok with everyone.

We only need to ensure the scope of APIs does not change. i.e. Private setters should stay private and same for public.

lcheunglci:
I just tried to use the new `init` syntax

`
public SqlNotificationType Type { get; init; }
public SqlNotificationInfo Info { get; init; }
public SqlNotificationSource Source { get; init; }
`

and I got compiler errors for netcore and netfx projects. `1>D:\git\SqlClient\src\Microsoft.Data.SqlClient\src\Microsoft\Data\SqlClient\SqlNotificationEventArgs.cs(23,48,23,52): error CS0518: Predefined type 'System.Runtime.CompilerServices.IsExternalInit' is not defined or imported`

Then I found out that C#9 is only available in .NET5 and above and since the project is built using .net framework 4.6.1 and .net 3.1, we won't be able to use `init` only setters.

lcheunglci:
> Both old files can be deleted too?

Thanks for catching that, I missed that for some reason.  I usually merge netfx into netcore first, delete netfx version, update the ref in the netfx csproj verify it builds and then move it over to shared src and update the references before I do the doc path updates, and IDE fixes and when I commit it, it will automatically marks the one for netcore as rename, but somehow it got missed.

johnnypham (approved):

PR #1315, opened 2021-10-04
Move into Shared SqlMetaDataFactory.cs

lcheunglci:
Related to #1261 . Moved the netcore version into shared src and updated the csproj references


Kaur-Parminder (approved):

johnnypham:
Looks like the netfx file is still there.

cheenamalhotra (changes requested):
Linux builds are failing, there seems to be an issue with conflict resolution.

PR #1316, opened 2021-10-05
Move to Shared for SqlQueryMetadataCache.cs

lcheunglci:
Relates to #1261. Move the netcore version into shared src and updated the references in the csprojs.  During the update to the conform to the coding style, I left a IDE1019, pattern matching, on line 64 on the info message list because I find the original way easier to follow than the suggestion.


Wraith2:
The original:
```csharp
            Dictionary<string, SqlCipherMetadata> cipherMetadataDictionary = _cache.Get(cacheLookupKey) as Dictionary<string, SqlCipherMetadata>;

            // If we had a cache miss just return false.
            if (cipherMetadataDictionary is null)
            {
                IncrementCacheMisses();
                return false;
            }
```

the pattern matched alternative:
```csharp
            if (_cache.Get(cacheLookupKey) is not Dictionary<string, SqlCipherMetadata> cipherMetadataDictionary)
            {
                IncrementCacheMisses();
                return false;
            }
'''

Since the `cipherMetadataDictionary` variable is used after the if statement that declares it I agree with you that it is clearer to use the original version.

cheenamalhotra (approved):

johnnypham (approved):

PR #1317, opened 2021-10-05
Move to Shared for SqlEnums.cs

lcheunglci:
Relates to #1261 . I merged the netfx changes into netcore, and move it to shared src.  I updated to the code to conform with the coding style, but left the IDE1006 ones because renaming the static variables with _s and internal read-only fields with _ affected other files i.e. SqlDataReader, SqlDataRecord, SqlUtil, SqlParameter, TdsParser, SqlComand, and more, which are all on the list to be merged, so I wanted to keep the changes small.


johnnypham (approved):
There were some mdac comments that weren't brought over. It was confirmed that they should remain.

Other than that, LGTM.

Kaur-Parminder (approved):

JRahnama (approved):

PR #1322, opened 2021-10-06
Move into Shared for SqlError.cs

lcheunglci:
Relates to #1261 . Merged netfx into netcore for SqlError.cs and move it into shared src and updated the references in the csprojs.  I did notice the SqlError in netfx uses the [Serializable] attribute for the class, and has 2 fields that are marked as [System.Runtime.Serialization.OptionalField] with a version number, but in netcore, they've all been switched over to auto properties so they don't get serialized anyway. I ran the functional tests and manual tests for netfx and it seems to be pass and didn't notice any issues, so hopefully we don't need to use them.


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlError.cs
```diff
@@ -6,9 +6,10 @@
 
 namespace Microsoft.Data.SqlClient
 {
-    /// <include file='../../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlError.xml' path='docs/members[@name="SqlError"]/SqlError/*' />
+    /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlError.xml' path='docs/members[@name="SqlError"]/SqlError/*' />
     public sealed class SqlError
```
DavoudEshtehari, line 10:
```suggestion
    [Serializable]
    public sealed class SqlError
```

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlErrorTest.cs
```diff
@@ -0,0 +1,71 @@
...
+
+            Type sqlErrorType = typeof(SqlError);
+
+            // SqlError only has internal constructors, in order to instantiate this, we use reflection
+            SqlError sqlError =  (SqlError)sqlErrorType.Assembly.CreateInstance(
+                sqlErrorType.FullName,
+                false,
+                BindingFlags.Instance | BindingFlags.NonPublic,
+                null,
+                new object[] { 0, (byte)0x00, FATAL_ERROR_CLASS, null, msg, "", 0, null },
```
DavoudEshtehari, line 64: Being more specific.
```suggestion
                new object[] { 100, (byte)0x00, FATAL_ERROR_CLASS, "ServerName", msg, "ProcedureName", 10, null },
```

DavoudEshtehari (approved):

Kaur-Parminder (approved):

johnnypham (approved):

johnnypham (approved):

PR #1349, opened 2021-10-14
Merging SqlConnectionStringBuilder code bases

DavoudEshtehari:
- related issue: #1261.
- Common Type: **SqlConnectionStringBuilder**
  - I used if-directives for separation.
  - I rearranged members and grouped them in **Fields**, **Private Methods**, and **Public APIs**.
  - Null reference members have been removed.
  - I chose the better implementation from the existing codebases besides applying some code improvements.


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionStringBuilder.cs
```diff
@@ -38,26 +42,34 @@ private enum Keywords
...
             UserInstance,
+#if NETFRAMEWORK
             ContextConnection,
+#endif
             TransactionBinding,
             ApplicationIntent,
             MultiSubnetFailover,
+#if NETFRAMEWORK
             TransparentNetworkIPResolution,
+#endif
```
Wraith2, line 72: Is the order of keywords important or could they be reordered so that all the netfx ones are in a single block?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionStringBuilder.cs
```diff
@@ -169,1766 +184,1676 @@ static SqlConnectionStringBuilder()
...
+        public override object this[string keyword]
         {
-            get { return _integratedSecurity; }
-            set
+            get
             {
-                SetValue(DbConnectionStringKeywords.IntegratedSecurity, value);
-                _integratedSecurity = value;
+                Keywords index = GetIndex(keyword);
+                return GetAt(index);
```
Wraith2, line 921: `get => GetAt(GetIndex(keyword));` ?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionStringBuilder.cs
```diff
@@ -169,1766 +184,1676 @@ static SqlConnectionStringBuilder()
...
-        [ResCategoryAttribute(StringsHelper.ResourceNames.DataCategory_Security)]
-        [ResDescriptionAttribute(StringsHelper.ResourceNames.DbConnectionString_Authentication)]
-        [RefreshPropertiesAttribute(RefreshProperties.All)]
-        public SqlAuthenticationMethod Authentication
-        {
-            get { return _authentication; }
             set
             {
-                if (!DbConnectionStringBuilderUtil.IsValidAuthenticationTypeValue(value))
+                if (null != value)
```
Wraith2, line 925: `value is not null` or `value != null` ?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionStringBuilder.cs
DavoudEshtehari: As you know, the Enum order would be important if we used the member values. I didn't check its references and tried to keep it's order like it was. I'll check to find whether we could group them together.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionStringBuilder.cs
Wraith2: The new one is an improvement, I just prefer the c# idomatic way of using variable == const rather than the C way. 😁

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionStringBuilder.cs
DavoudEshtehari: done

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionStringBuilder.cs
```diff
@@ -169,1766 +176,1672 @@ static SqlConnectionStringBuilder()
...
+
+        private static bool ConvertToIntegratedSecurity(object value) => DbConnectionStringBuilderUtil.ConvertToIntegratedSecurity(value);
+
+        private static SqlAuthenticationMethod ConvertToAuthenticationType(string keyword, object value) => DbConnectionStringBuilderUtil.ConvertToAuthenticationType(keyword, value);
+
+        private static string ConvertToString(object value) => DbConnectionStringBuilderUtil.ConvertToString(value);
+
+        private static ApplicationIntent ConvertToApplicationIntent(string keyword, object value) => DbConnectionStringBuilderUtil.ConvertToApplicationIntent(keyword, value);
+
+        /// <summary>
```
johnnypham, line 287: These xml comments for the private method aren't really helpful, let's remove them?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionStringBuilder.cs
```diff
@@ -169,1766 +176,1672 @@ static SqlConnectionStringBuilder()
...
             }
         }
 
-        /// <include file='..\..\..\..\..\..\..\doc\snippets\Microsoft.Data.SqlClient\SqlConnectionStringBuilder.xml' path='docs/members[@name="SqlConnectionStringBuilder"]/CurrentLanguage/*' />
-        [DisplayName(DbConnectionStringKeywords.CurrentLanguage)]
-        [ResCategoryAttribute(StringsHelper.ResourceNames.DataCategory_Initialization)]
-        [ResDescriptionAttribute(StringsHelper.ResourceNames.DbConnectionString_CurrentLanguage)]
-        [RefreshPropertiesAttribute(RefreshProperties.All)]
-        public string CurrentLanguage
+        internal sealed class SqlConnectionStringBuilderConverter : ExpandableObjectConverter
```
johnnypham, line 696: This class isn't used in netcore.

Wraith2: It may not be used directly but if it's ever referenced it can be created through the component provider infrastructure. It's probably best to keep it around.

DavoudEshtehari: It wasn't but I've included it in netcore to narrow the diverge.

PR #1354, opened 2021-10-18
Moved to Shared - SqlParameter

Kaur-Parminder:
This is part of #1261 code merge effort. Moved the file to common folder. added some code style changes and updated Resource strings.


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlParameter.cs
```diff
@@ -317,8 +330,8 @@ string xmlSchemaCollectionName
             SqlDbType = dbType;
             Size = size;
             Direction = direction;
-            PrecisionInternal = precision;
-            ScaleInternal = scale;
```
Wraith2, line 321: The difference between precision and scale calculations between netcore and netfx was the main reason that I hadn't merged these two files. It looked to me as if the netcore version was later than the netfx version because netfx was simpler so netcore looked like a fix for something specific but it wasn't clear what behaviour it was fixing. If that's the case then this merge would regress some behaviour for someone.

It would be good to get some deep dive info on what the difference between the netfx and netcore behviour means for this.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlParameter.cs
```diff
@@ -1620,7 +1637,11 @@ internal SmiParameterMetaData GetMetadataForTypeInfo()
         internal SmiParameterMetaData MetaDataForSmi(out ParameterPeekAheadValue peekAhead)
         {
             peekAhead = null;
-            MetaType mt = ValidateTypeLengths(true /* Yukon or newer */ );
+            MetaType mt = ValidateTypeLengths(
+#if NETFRAMEWORK
```
Wraith2, line 1641: Yukon is SQL 2005 which isn't supported anymore. So everything we will ever touch from this library is yukon or later which is why netcore didn't contain this code anymore. I don't think keeping this code is useful.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlParameter.cs
Kaur-Parminder: I have kept both framework properties to avoid any regression.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlParameter.cs
Kaur-Parminder: I will do it in separate PR after discussing internally, as this check is in many files not just here.

johnnypham (approved):

David-Engel (approved):

PR #1355, opened 2021-10-18
Tests | Address random error on AADPasswordWithWrongPassword test case

cheenamalhotra:
We should not compare error message text in this case, as it may change based on when this test runs.
This is widely observed to fail randomly.

e.g. https://dev.azure.com/sqlclientdrivers-ci/sqlclient/_build/results?buildId=42647&view=results


src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectivityTests/AADConnectionTest.cs
```diff
@@ -208,8 +208,8 @@ public static void AADPasswordWithWrongPassword()
 
             SqlException e = Assert.Throws<SqlException>(() => ConnectAndDisconnect(connStr));
```
DavoudEshtehari, line 209:
```suggestion
            Assert.Throws<SqlException>(() => ConnectAndDisconnect(connStr));
```

DavoudEshtehari (approved):

lcheunglci (approved):

PR #1357, opened 2021-10-18
Revert mars state machine changes

### LINKED ISSUE #1234: Concurrent connection usage and MARS
As part of running some compatibility checks between EF Core 6.0 and SqlClient 3.0.0 and 4.0.0-preview1.21237.2 (https://github.com/dotnet/efcore/issues/25766), I came across what looks like a considerable perf regression. While the EF test suite completed in around 4m30s with SqlClient 3.0.0, it took around 12 minutes with 4.0.0-preview1.21237.2.

Some digging seems to point at test [QueryBugsTest.Thread_safety_in_relational_command_cache](https://github.com/dotnet/efcore/blob/release/6.0/test/EFCore.SqlServer.FunctionalTests/Query/QueryBugsTest.cs#L7090); with 3.0.0 the total run completes in 20 seconds (including all startup), whereas with 4.0.0-preview1.21237.2 it takes around **5m15s**. This test runs 100 parallel queries - no big data is being transferred, and purely sync I/O is used.

To run this, you can check out the EF Core code base, run build.cmd to get the latest dotnet core, go inside test/EFCore.SqlServer.FunctionalTests and run `dotnet test --filter Thread_safety_in_relational_command_cache`

cheenamalhotra:
Fixes #1234 
+ Test added to reproduce failures with change from #933.

P.S. we've been having this issue in pipelines too, they're failing a lot more recently.
At this point for 4.0 GA it's safest to revert it.


src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/MARSTest/MARSTest.cs
```diff
@@ -639,5 +639,64 @@ public static async Task MarsScenarioClientJoin()
...
+                            0, 300,
+                            i =>
+                            {
+                                using var cmd = new SqlCommand
+                                {
+                                    Connection = conn,
+                                    CommandText = @$"
+                                SELECT [l].[Id], [l].[IsDeleted]
+                                FROM [{table}] AS [l]
+                                WHERE ([l].[IsDeleted] = CAST(0 AS bit)) AND [l].[Id] IN (1, 2, 3)"
```
DavoudEshtehari, line 682: Did you run the query on an empty table by purpose or it's supposed to have some records?

cheenamalhotra: I believe this is a dummy query, I used as-is from https://github.com/dotnet/SqlClient/issues/1234#issuecomment-910111717 😄
Data is not required to get it to fail.

Wraith2:
Reverting for 4.0 release and then putting back in or just reverting and leaving out entirely?

cheenamalhotra:
> Reverting for 4.0 release and then putting back in or just reverting and leaving out entirely?

We can accept again with fix incorporated for failing test.

cheenamalhotra:
> Who will be doing that? If we want to evolve the managed implementation then this mars code needs to change but it's a big complex job and waiting for me to work it all out doesn't seems sensible.

Sure, currently it's not in our plan as it's not highly requested change and is an improvement that needs redesign. 
We will look into it as an enhancement when we have capacity post GA. You can create an issue to be able to track it for future.

Wraith2:
> We will be considering other design improvements too if we come up something better, that’s why merging this exact change is uncertain right now.

If someone can provide a better solution that sounds good to me. Unfortunately I don't believe that will happen based on my experience working on this library so far. Even important user affecting and often reported issues like async string perf sit untouched for months to years. To come up with something better someone would have to make the effort and take the time to do so, and that just won't happen.

johnnypham (approved):

Wraith2:
Now that 4.0 has shipped can this reversion be considered again. The intent of the original fix is still good it just needs attention on edge cases that wasn't available before.

JRahnama:
@Wraith2, we are interested in the change/PR, but cannot merge it again without fixing the issue it caused before. If you can address the issue we will gladly take it back. Please provide enough tests for edge cases as well. We have also added pipelines for net 6 testing.

Wraith2:
All on my own again then.

JRahnama:
> All on my own again then.

Well, we ,as a team, and I personally appreciate your contribution. You have been a great help to us.  If there is any part that you need us to look into, ping me or other team members and we will jump in. Currently, all other team members are busy with EOY tasks and reports and some other features to add into the driver.

JRahnama:
@Wraith2 @DavoudEshtehari is on vacation and will be back next week. Will you be interested in having a call and go through the changes and failures of this PR and come up with a plan sometime next week?

Wraith2:
Sure. Drop me an email invite of when's best for you guys.

Issue #1332, opened 2021-10-12
ActiveDirectoryDefault authentication ignores environment variables used by DefaultAzureCredential


smichtch:
### Describe the bug
`SqlAuthenticationMethod.ActiveDirectoryDefault` does not fall back to environment variables normally used by `DefaultAzureCredentialOptions`.
Specifically the code below is clobbering all the environment variable values read by the `DefaultAzureCredentialOptions` constructor even if the SQL connection string isn't overriding for those keys:
https://github.com/dotnet/SqlClient/blob/d9c2e7c2e8422c334bd08db345c9036c82ac3d79/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ActiveDirectoryAuthenticationProvider.cs#L125-L135

### To reproduce
Azure service using  `AZURE_CLIENT_ID`

Include a complete code listing (or project/solution) that we can run to reproduce the issue.

Partial code listings, or multiple fragments of code, will slow down our response or cause us to push the issue back to you to provide code to reproduce the issue.

```c#
Console.WriteLine("Hello World!");
```

### Expected behavior
`SqlAuthenticationMethod.ActiveDirectoryDefault` token acquisition behavior matches that of `DefaultAzureCredential`.

### Further technical details
Microsoft.Data.SqlClient version: 3.0.1
.NET target: N/A
SQL Server version: N/A
Operating system: N/A

**Additional context**

cheenamalhotra:
Hi @smichtch 

The reason why that happens is because we set client Id for ManagedIdentityClientId which could also be an empty string.. We can look into improving it to only set those options if they have a value set.

I'll try it and let you know soon.

PR #1360, opened 2021-10-19
Active Directory Default authentication improvements

cheenamalhotra:
Addresses #1332 and other improvements:
- Rename `tenantId` to `audience` as the parameters accept broader values than GUID tenant Id.
- Set Credential Options optionally only when valid data is available, allows users to specify Environment variables
- Test case added for Kusto Samples DB.

cc @David-Engel

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ActiveDirectoryAuthenticationProvider.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ActiveDirectoryAuthenticationProvider.cs
@@ -157,7 +157,7 @@ namespace Microsoft.Data.SqlClient
                     defaultAzureCredentialOptions.ManagedIdentityClientId = clientId;
                     defaultAzureCredentialOptions.SharedTokenCacheUsername = clientId;
                 }
-                AccessToken accessToken = await new DefaultAzureCredential(defaultAzureCredentialOptions).GetTokenAsync(tokenRequestContext, cts.Token);
+                AccessToken accessToken = await new DefaultAzureCredential(defaultAzureCredentialOptions).GetTokenAsync(tokenRequestContext, cts.Token).ConfigureAwait(false);
                 SqlClientEventSource.Log.TryTraceEvent("AcquireTokenAsync | Acquired access token for Default auth mode. Expiry Time: {0}", accessToken.ExpiresOn);
                 return new SqlAuthenticationToken(accessToken.Token, accessToken.ExpiresOn);
             }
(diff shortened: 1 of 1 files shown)


lcheunglci (approved):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ActiveDirectoryAuthenticationProvider.cs
```diff
@@ -119,26 +119,41 @@ public override async Task<SqlAuthenticationToken> AcquireTokenAsync(SqlAuthenti
...
 
+            string authority = parameters.Authority;
             int seperatorIndex = parameters.Authority.LastIndexOf('/');
-            string tenantId = parameters.Authority.Substring(seperatorIndex + 1);
-            string authority = parameters.Authority.Remove(seperatorIndex + 1);
 
-            TokenRequestContext tokenRequestContext = new TokenRequestContext(scopes);
+            // Authority Url does not always contain "<tenantId>", e.g. for Kusto cluster dbs
+            // Optionally split Tenant Id in such cases.
+            if (Guid.TryParse(parameters.Authority.Substring(seperatorIndex + 1), out Guid tenantId))
```
smichtch, line 127: The `tenantId` isn't always a GUID, can be `"microsoft.com"`, `"ame.gbl"`, ... or `"common"`. In the latter case token will be issued by the home tenant of the user that you signin with.
Also could resolve it under `if (parameters.AuthenticationMethod == SqlAuthenticationMethod.ActiveDirectoryDefault)` since SPA auth uses the full authority URL now.

cheenamalhotra: After reading through [documentation](https://docs.microsoft.com/en-us/azure/active-directory/develop/msal-client-application-configuration#how-to-specify-the-audience-in-your-codeconfiguration), it seems to work fine either way... we need not change this part and can continue as-is. The term `tenantId` was misleading and our thought was that it's only applicable to `Guid` values. It makes sense to call it `audience` instead. I've updated code accordingly.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ActiveDirectoryAuthenticationProvider.cs
```diff
@@ -118,27 +118,45 @@ public override async Task<SqlAuthenticationToken> AcquireTokenAsync(SqlAuthenti
...
-                    VisualStudioCodeTenantId = tenantId,
-                    VisualStudioTenantId = tenantId,
+                    SharedTokenCacheTenantId = audience,
+                    VisualStudioCodeTenantId = audience,
+                    VisualStudioTenantId = audience,
                     ExcludeInteractiveBrowserCredential = true // Force disabled, even though it's disabled by default to respect driver specifications.
                 };
+
+                // Optionally set clientId when available
+                if (clientId != null)
```
JRahnama, line 155: NIT: just to keep/follow same coding style:
```suggestion
                if (clientId is not null)
```

DavoudEshtehari (approved):

johnnypham (approved):

PR #1366, opened 2021-10-21
Use Buffer.BlockCopy with byte[]

Kaur-Parminder:
This change is based on changes done https://github.com/dotnet/corefx/commit/3ab3bd6efcba669bf04d42f738fe5a495ec62940. I have replaced Array.Copy wherever there is byte[]. Buffer.BlockCopy has less overhead than Array.Copy when copying byte[]s.


johnnypham (approved):

DavoudEshtehari (approved):

JRahnama:
working with bytes are more difficult than working with indexes and may cause debugging to be more complicated, but since the data is already in bytes no issue here. Looks good to me.

JRahnama (approved):

PR #1375, opened 2021-10-27
Test | Adds TestPacketNumberWraparoundwithAwaitAsync and fixes the issue with TestConnectionIsSafeToReuse in TVPTest

JRahnama:
This PR adds a new test to use `await/async` to run `RunPacketNumberWraparound`. acopy of  the existing test `TestPacketNumberWraparound ` which is an active issue. It also fixes the issue with `TestConnectionIsSafeToReuse` in TVP tests.

### DIFF
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ParameterTest/TvpTest.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ParameterTest/TvpTest.cs
@@ -53,8 +53,7 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests
         }
 
         [ConditionalFact(typeof(DataTestUtility), nameof(DataTestUtility.AreConnStringsSetup), nameof(DataTestUtility.IsNotAzureServer))]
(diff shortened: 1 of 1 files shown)


JRahnama:
> What's the problem with `TestPacketNumberWraparound`? I wrote that test to cover a fix I made while the code was in corefx [dotnet/corefx#34709](https://github.com/dotnet/corefx/pull/34709) The ActiveIssue number 5531 doesn't match up with an issue in this repository or corefx

The test fails in its current setup, but when I turned it to async/await pattern it worked pretty good, but apparently it changed the purpose of the test, so I reverted that back. I am going to close this PR soon. Regarding the Active issue, there is an intermittent failure when it runs on netcore. I will investigate it deeper on Tuesday when I am back to work.

Wraith2:
There was some discussion of the test and an explanation of the intent on the original issue I linked to which might help. Timing is important in that particular test because it's designed to run through 255+ packets to complete a roundtrip through a byte sized counter, if the CI is too slow it will signal as a fault even on netcore which is why it's set to run for a long time.

Issue #1376, opened 2021-10-29
SqlNotificationInfo.None enum value missing, causing unnecessary exceptions.


ascott18:
### Describe the bug
In SqlDependencyListener, when an incoming XML message is parsed, sometimes these messages look like this:

``` xml
<qn:QueryNotification xmlns:qn="http://schemas.microsoft.com/SQL/Notifications/QueryNotification" id="4579" type="change" source="timeout" info="none" database_id="9" sid="0x01050000000000051500000002AAFCB5FFBA57356D2C83B7E9030000"><qn:Message>52cdaaaf-f5e5-4342-8c10-dfec05cf4950;ec06ef71-ec94-4e56-a11e-c1905ebf4cf9</qn:Message></qn:QueryNotification>
```

The `info` attribute is parsed at https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlDependencyListener.cs#L1133-L1167. However, the enum that the value is parsed into does not included a value "None", causing an exception to be thrown. It is ultimately handled by the containing try/catch, but there are a few issues with this:

- Exception flow and handling incurs a performance cost.
- The exception is System.ArgumentException and is thrown from mscorlib, which means that suppressing it in Visual Studio's Exception Settings causes _all_ such exceptions to be supressed, making debugging more difficult when there are other AgumentExceptions thrown by mscorlib that I _do_ want to break on.

![image](https://user-images.githubusercontent.com/5017521/139485093-3bd6ad00-f978-4f40-b46c-6574d6212cd0.png)

### To reproduce
This is a bit weird to reproduce. The first few notifications where will be of `Info == Error`. But sometimes, it will be `"none"` in the XML, which then throws and gets turned into `Unknown`. 

Set exception settings in Visual Studio to break on all CLR exceptions.

```c#
// See https://aka.ms/new-console-template for more information
using Microsoft.Data.SqlClient;

var connectionString = "Server=localhost;Database=BugReproDb;Trusted_Connection=True;Command Timeout=5";
SqlDependency.Start(connectionString);
var connection = new SqlConnection(connectionString);
connection.Open();
var cmd1 = connection.CreateCommand();
cmd1.CommandText = "IF OBJECT_ID('TestTable') IS NULL CREATE TABLE TestTable (c1 int);";
cmd1.ExecuteNonQuery();
cmd1.Dispose();

[...]

PR #1378, opened 2021-11-05
Avoid throwing exception when receiving invalid SqlNotificationInfo value

johnnypham:
Addresses #1376 

When receiving an invalid `SqlNotificationInfo` value, an exception would be thrown because the call to `Enum.Parse` fails. We can use `Enum.TryParse` to avoid this. The result will still be `SqlNotificationInfo.Unknown` either way.


ascott18 (approved):
I can confirm after testing that the exception is no longer thrown. The "none" attribute value still comes out as "Unknown" as expected, but without having VS break into an exception when exception settings are all enabled and Just My Code is disabled.

cheenamalhotra:
Can you add a test case here to cover this code block?

johnnypham:
> Can you add a test case here to cover this code block?

I was unable to reproduce it but this is a server-side bug so it shouldn't be occurring. If it does, we'll avoid throwing an unnecessary exception.

DavoudEshtehari (approved):

Kaur-Parminder (approved):

Issue #305, opened 2019-11-13
Integrated security - manual kerberos negotiation


aaomidi:
I would like to be able to manually do the kerberos negotiation with something like: https://github.com/SteveSyfuhs/Kerberos.NET

Then include the security token in the connection string for SqlClient and use that to connect to databases. Is something like this possible?

cheenamalhotra:
Hi @aaomidi 

Have you tried using [accessToken](https://docs.microsoft.com/en-us/dotnet/api/microsoft.data.sqlclient.sqlconnection.accesstoken?view=sqlclient-dotnet-core-1.0) connection property?

aaomidi:
I have not! I'll try to use that :)


Is there documentation available explaining what that token does?

aaomidi:
I came across this "patch" for SqlClient, https://github.com/macsux/sql-server-kerberos-poc/blob/master/SqlClientPlay/Program.cs I'm guessing it would accomplish the same thing?

cheenamalhotra:
This accessToken will be passed instead of credentials to perform authentication, you can try using it as:

```csharp
string ConnectionString =@"Data Source={serverName}; Initial Catalog={db};"
SqlConnection conn = new SqlConnection(ConnectionString);
conn.AccessToken = "Your access token"
conn.Open();
```

aaomidi:
@charlienilsson I was curious what is "AccessToken" - like where would this come from? What is the value that's expected there?

cheenamalhotra:
> I came across this "patch" for SqlClient, https://github.com/macsux/sql-server-kerberos-poc/blob/master/SqlClientPlay/Program.cs I'm guessing it would accomplish the same thing?

This is using reflection with driver's SNI and passing GSS Credentials to acquire ticket, which is a hack, I wouldn't recommend as a permanent solution.

aaomidi:
Yeah I was curious if this AccessToken is basically looking for, say the kerberos negotiated client token.

cheenamalhotra:
AccessToken is a JWT/OAuth2 token which is acquired from authorizing parties that SQL Server recognizes. It's mostly talked about in terms of Azure DB authentication. [Docs](https://docs.microsoft.com/en-us/azure/sql-database/sql-database-aad-authentication-configure#azure-ad-token)

For Kerberos, I haven't tested myself but thought maybe you got something similar to experiment with, if not then that's not going to be possible.

aaomidi:
[...]

PR #1379, opened 2021-11-08
[DRAFT] [Issue 305] Introduce a hook in Microsoft.Data.SqlClient to allow different mechanisms to acquire a Kerberos ticket

srdan-bozovic-msft:


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/SqlClient?pullRequest=1379) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/SqlClient?pullRequest=1379) before we can accept your contribution.<br/><br/>:x: srdan-bozovic-msft [sign now](https://cla.dotnetfoundation.org/dotnet/SqlClient?pullRequest=1379)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/SqlClient?pullRequest=1379) it.</sub>

srdan-bozovic-msft:
resolves #305

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -135,6 +138,9 @@ sealed internal class SqlInternalConnectionTds : SqlInternalConnection, IDisposa
         SqlFedAuthToken _fedAuthToken = null;
         internal byte[] _accessTokenInBytes;
 
+        // Kerberos ticket provided by OnRetrieveKerberosTicket
+        internal byte[] _kerberosTicketInBytes;
```
Wraith2, line 142: I realise that this follow the naming convention from the line above, the same logic will apply to that field as well if someone added it today. 
Do we need the `InBytes` suffix? Do we have any other form of the same data (string etc) that is used? If not then it could just be called `_kerberosTicket`

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -3327,4 +3344,404 @@ internal void SetDerivedNames(string protocol, string serverName)
...
+        // possible scenarios:
+        // (localdb)\<instance name>
+        // or (localdb)\. which goes to default localdb
+        // or (localdb)\.\<sharedInstance name>
+        internal static string GetLocalDBInstance(string dataSource, out bool error)
+        {
+            string instanceName = null;
+
+            // ReadOnlySpan is not supported in netstandard 2.0, but installing System.Memory solves the issue
+            ReadOnlySpan<char> input = dataSource.AsSpan().TrimStart();
```
Wraith2, line 3457: Thanks for using spans, almost always better for perf 😁

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -3327,4 +3344,404 @@ internal void SetDerivedNames(string protocol, string serverName)
...
+            if (string.IsNullOrEmpty(ServerName) || IsLocalHost(ServerName))
+            {
+                ServerName = _connectionProtocol == Protocol.Admin ?
+                    Environment.MachineName : DefaultHostName;
+            }
+        }
+
+        private bool InferConnectionDetails()
+        {
+            string[] tokensByCommaAndSlash = _dataSourceAfterTrimmingProtocol.Split(BackSlashCharacter, CommaSeparator);
```
Wraith2, line 3521: Could this use span to avoid the string and array allocations? We had some interesting back and forth on using spans while parsing in https://github.com/dotnet/SqlClient/pull/1237 that might be useful to review.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/KerberosTicketRetrievalCallback.cs
```diff
@@ -0,0 +1,13 @@
...
+// See the LICENSE file in the project root for more information.
+
+
+namespace Microsoft.Data.SqlClient
+{
+        /// <summary>
+        /// A callback to provide kerberos ticket for given service principal name (SPN) on demand.
+        /// </summary>
+        /// <returns></returns>
+        public delegate byte[] KerberosTicketRetrievalCallback(string serverPrincipalName);
```
Wraith2, line 12: Does this need to be a new type, or is there a reason to use a type over a `Func<byte[],string>` ? 

I'm also no sure about the naming. Is it really a callback? it feels more like a provider or an adapter. The user will set the this property of the connection to and it will be called by the connection to get the ticket to use. So we're not really calling back to anything instead we're providing a way to get an external value by some means. `KerberosTicketProvider` ? something like that?

srdan-bozovic-msft: This follows conventions for this library ( https://github.com/dotnet/SqlClient/search?q=callback ). Team is also exploring changing this into adapter pattern design already used in FedAuth flow.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
srdan-bozovic-msft: This is c/p of a class from netcore path. This needs some refactoring and unification as it's useful for both necore and netfx.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
srdan-bozovic-msft: It's definitely worth exploring. Everything said above applies.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
srdan-bozovic-msft: Agree with the comment. As you've noticed correctly the idea was to follow the established convention. Team may consider name refactoring.

MichelZ:
3rd anniversary of this PR! Congratulations :)

cheenamalhotra:
Closing the PR for now as stale, please fix conflicts and re-open when ready for review.

Issue #1380, opened 2021-11-08
SqlParameter ignores scale when explicitly set to 0


EamonHetherton:
### Describe the bug
When creating a DateTime2 SqlParameter with scale explicitly set to zero, the parameter is sent to the SQL server as datetime2(7)

`new SqlParameter("@p1", System.Data.SqlDbType.DateTime2) { Scale = 0};` -> `@p1 datetime2(7)`

Miss-matched parameter types can cause performance issues in SQL. in addition the use of datetime2(0) instead of datetime2(7) in the database saves 2 bytes per field.

My real world example of this is a table with hundred of millions of rows where the datetime data does not have sub-second precision so we are using datetime2(0) as the data type and that a specific database query does 200-300 times as much IO when using a parameter type of datetime2(7) than datetime2(0).

### To reproduce

```c#
[Fact]
public async Task DemonstrateBrokenScale()
{
    var date = new DateTime(2021, 11, 01, 23, 59, 59, 999);
    var dateRoundedToSeconds = date.AddMilliseconds((Math.Round((double)date.Millisecond / 1000) * 1000) - date.Millisecond);

    var sqlConnection = new SqlConnection("Data Source=.;Integrated Security=True;");
    await sqlConnection.OpenAsync();
    await using var command = sqlConnection.CreateCommand();
    command.CommandType = System.Data.CommandType.Text;
    command.CommandText = "select @p1";
    
    var p1 = new SqlParameter("@p1", System.Data.SqlDbType.DateTime2) { Scale = 0, Value = date };
    command.Parameters.Add(p1);

    var result = await command.ExecuteScalarAsync();
    
    Assert.Equal(dateRoundedToSeconds, (DateTime)result); //should get back value rounded up to whole seconds (SQL server behaviour)
}
```

This will produce the following sql: 
`exec sp_executesql N'select @p1',N'@p1 datetime2(7)',@p1='2021-11-01 23:59:59.9990000'`

### Expected behavior
it should create a parameter of type datetime2(0), i'm a little unsure if it should round the value or leave that to the server:

`exec sp_executesql N'select @p1',N'@p1 datetime2(0)',@p1='2021-11-01 23:59:59.9990000'`
or
`exec sp_executesql N'select @p1',N'@p1 datetime2(0)',@p1='2021-11-02 00:00:00'`

### Further technical details

**Additional context**
I believe this has been a problem for some time and the issue is noted in a comment in the source:
[...]

PR #1381, opened 2021-11-08
#1380 support explicitly set scale of zero for datetime2

EamonHetherton:
...datetimeoffset and time

### DIFF
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/TracingTests/DiagnosticTest.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/TracingTests/DiagnosticTest.cs
@@ -25,9 +25,8 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests
     public class DiagnosticTest
     {
         private const string BadConnectionString = "data source = bad; initial catalog = bad; integrated security = true; connection timeout = 1;";
-        private static readonly string s_tcpConnStr = DataTestUtility.TCPConnectionString ?? string.Empty;
 
-        [ConditionalFact(typeof(DataTestUtility), nameof(DataTestUtility.AreConnStringsSetup))]
+        [Fact]
         public void ExecuteScalarTest()
         {
             RemoteExecutor.Invoke(() =>
@@ -41,14 +40,14 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests
                         cmd.CommandText = "SELECT [name], [state] FROM [sys].[databases] WHERE [name] = db_name();";
 
                         conn.Open();
-                        var output = cmd.ExecuteScalar();
+                        cmd.ExecuteScalar();
                     }
                 });
                 return RemoteExecutor.SuccessExitCode;
             }).Dispose();
         }
 
-        [ConditionalFact(typeof(DataTestUtility), nameof(DataTestUtility.AreConnStringsSetup))]
+        [Fact]
         public void ExecuteScalarErrorTest()
         {
             RemoteExecutor.Invoke(() =>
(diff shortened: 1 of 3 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/SqlClient?pullRequest=1381) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/SqlClient?pullRequest=1381) before we can accept your contribution.<br/><br/>:x: EamonHetherton [sign now](https://cla.dotnetfoundation.org/dotnet/SqlClient?pullRequest=1381)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/SqlClient?pullRequest=1381) it.</sub>

Wraith2:
I've mentioned my misgivings on the issue so I won't repeat them here. 
The implementation looks correct. Can you add test(s)?

EamonHetherton:
I've two different points to discuss (supporting existing behaviour and tests), so I'll do them in two separate comments.

Regarding the existing behaviour and minimising the impact of a change, I think it is still worth discussing and maybe coming to a consensus. There are three realistic options I see:

1. Change the behaviour as discussed in [#1380](https://github.com/dotnet/SqlClient/issues/1380#issuecomment-963628525).
2) Do 1 and create an AppContextSwitch to opt-in to the existing behaviour.
3) Do 1 and create an AppContextSwitch to opt-in to the new behaviour.

Then there is a second dimension of netfx vs netcore. With all the work to merge code bases I suspect there is not a lot of appetite to diverge behavior on the NETFRAMEWORK switch but it's worth asking. It seems anecdotally to me at least that the threshold for breaking changes is higher in netfx than netcore but not sure if that is actually policy or where I'd find out. There could be an option to retain existing behavior on netfx by default with an op-in switch for the correct behaviour and just the correct behaviour on netcore?

Your thoughts?

Wraith2:
> (the SqlParameter class is not yet common across netfx and netcore)

There's a PR for it, https://github.com/dotnet/SqlClient/pull/1354 but PR's are merged in order so whichever gets merged first the other will need to resolve the conflicts.

> Then there is a second dimension of netfx vs netcore.

My opinion is that we should minimize the observable differences between netcore and netfx so whatever is done it should be the same. We have plenty of problems between windows and linux implementation at the network layer which it's not obvious to users why they are different, it's just easier for everyone to avoid that sort of situation.

> There are three realistic options I see:
My choice would be #2 (Do 1 and create an AppContextSwitch to opt-in to the existing behaviour.)  but it's down to the team decision, I just hang around here because I have no hobbies.

EamonHetherton:
As for tests, I did try :)

The SqlParameter class is sealed and the methods involved here are all internal or private apart from the Scale property itself which already returns whatever value is set so I couldn't write a failing test with the current code.

I followed the callers of GetActualScale to see if I could get to it by proxy and again I was blocked by the SqlCommand class being sealed and relevant methods being internal. Beyond that I was getting to the point of just about having to execute a command with a valid connection and trying to interrogate the produced sql which is pretty far from unit test land.

Removing the sealed attribute would allow me to create a derived class for testing but implications for existing code that may depend on the SqlParameter being sealed are unknown so not comfortable with that approach.

So the simplest approach to being able to test this is to add an "[assembly: InternalsVisibleTo("FunctionalTests")]" attribute.  I'll head down that road unless it is likely to get rejected?

Wraith2:
I wouldn't think there is a sensible way to add a functional test (functional tests don't require a database, manual tests do) because as you've said it requires either surface area changes or internal shenanigans which are better avoided.

You can just add a test case to the ManualTests which writes a value to the database like your original issue report and then verify that the scale has been correctly written can't you?

JRahnama:
@EamonHetherton currently we are on code freeze due to our next GA release. We will resume back on PRs after release.

PR #1377, opened 2021-11-01
Test - Add tests for code coverage part 1

lcheunglci:
Instead of making a giant PR with as much code coverage tests as possible, I'm going to split them up into parts and try to open one PR per week or depending on the number of tests added/modified.  I added [code coverage](https://dev.azure.com/sqlclientdrivers-ci/sqlclient/_build/results?buildId=43083&view=codecoverage-tab) to the following classes:

- SqlCollation
- SqlConnectionStringBuilder
- SqlClientLogger
- SqlClientFactory
- SqlParameterCollection
- SqlCommandSet

PR #1384, opened 2021-11-09
Test - Add tests for code coverage part2

lcheunglci:
Continuing from PR #1377 , this adds code [coverage](https://dev.azure.com/sqlclientdrivers-ci/sqlclient/_build/results?buildId=43083&view=codecoverage-tab) to the following classes:

- SqlCommandBuilder
- SqlMetaData
- SqlDependency

SqlMetaData is at 92.8% coverage, and there are certain code blocks that's not possible to add coverage, so I can see some areas where it can be cleaned up, but my focus here is for the tests. SqlDependency is at 75% coverage, and it's a bit tricky to invoke certain exceptions.

### DIFF
--- a/src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlCommandBuilderTest.cs
+++ b/src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlCommandBuilderTest.cs
@@ -3,8 +3,11 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
+using System.Collections.Generic;
 using System.Data;
 using System.Data.Common;
+using System.Reflection;
+using System.Threading;
 using Xunit;
 
 namespace Microsoft.Data.SqlClient.Tests
@@ -64,6 +67,8 @@ namespace Microsoft.Data.SqlClient.Tests
         {
             SqlCommandBuilder cb = new SqlCommandBuilder();
             Assert.Equal(".", cb.CatalogSeparator);
+            cb.CatalogSeparator = ".";
+            Assert.Equal(".", cb.CatalogSeparator);
         }
 
         [Theory]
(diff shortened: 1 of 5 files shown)


src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlMetaDataTest.cs
```diff
@@ -2,14 +2,742 @@
...
+            {
+                SqlMetaData metaData = new SqlMetaData("col1", SqlDbType.Decimal, precision, scale, true, true, SortOrder.Ascending, 0);
+            });
+            Assert.NotNull(ex);
+            Assert.NotEmpty(ex.Message);
+            Assert.Contains("precision", ex.Message, StringComparison.OrdinalIgnoreCase);
+        }
+
+        // TODO: This condition can never bet met because SqlDecimal.MaxPrecision == SqlDecimal.MaxScale
+        // and there's a check that scale cannot exceed precision, so we cannot this exception.
```
cheenamalhotra, line 524:
```suggestion
        // TODO: This condition can never be met because SqlDecimal.MaxPrecision == SqlDecimal.MaxScale
        // and there's a check that scale cannot exceed precision, so we cannot test this exception.
```

cheenamalhotra (approved):

johnnypham:
Removed some unnecessary asserts. Everything else LGTM.

johnnypham (approved):

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlCommandBuilderTest.cs
```diff
@@ -64,6 +67,8 @@ public void CatalogSeparator()
         {
             SqlCommandBuilder cb = new SqlCommandBuilder();
             Assert.Equal(".", cb.CatalogSeparator);
+            cb.CatalogSeparator = ".";
+            Assert.Equal(".", cb.CatalogSeparator);
```
Kaur-Parminder, line 71: @lcheunglci if Assertion line 69 is correct, Do we need line 71?

lcheunglci: This is to add code coverage in the CatalogSeparator property that checks if a "." is set even though it does default to it.

PR #1401, opened 2021-11-22
Update obsolete api calls in .NET 6

johnnypham:
https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.rngcryptoserviceprovider?view=net-6.0

https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x509certificate2.privatekey?view=net-6.0#System_Security_Cryptography_X509Certificates_X509Certificate2_PrivateKey

### DIFF
--- a/src/Microsoft.Data.SqlClient/tests/FunctionalTests/AlwaysEncryptedTests/Utility.cs
+++ b/src/Microsoft.Data.SqlClient/tests/FunctionalTests/AlwaysEncryptedTests/Utility.cs
@@ -95,8 +95,8 @@ namespace Microsoft.Data.SqlClient.Tests.AlwaysEncryptedTests
         {
             // Generate random bytes cryptographically.
             byte[] randomBytes = new byte[length];
-            RNGCryptoServiceProvider rngCsp = new RNGCryptoServiceProvider();
-            rngCsp.GetBytes(randomBytes);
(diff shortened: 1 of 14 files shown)


src/Microsoft.Data.SqlClient/tests/FunctionalTests/AlwaysEncryptedTests/Utility.cs
```diff
@@ -348,8 +348,9 @@ internal static byte[] DecryptDataUsingAED(byte[] encryptedCellBlob, byte[] key,
             return decryptedData;
         }
 
+#if NETFRAMEWORK
         /// <summary>
```
johnnypham, line 352: This method is only called in netfx. In net5, accessing the PrivateKey property throws an exception. Now in net6, it won't even compile, so we can just compile it for netfx.

Kaur-Parminder (approved):

JRahnama (approved):

DavoudEshtehari:
I found more usages should be updated. Check from [here](https://github.com/dotnet/SqlClient/search?q=RNGCryptoServiceProvider).

johnnypham:
> I found more usages should be updated. Check from [here](https://github.com/dotnet/SqlClient/search?q=RNGCryptoServiceProvider).

Updated those too.

PR #1404, opened 2021-11-24
Include derived types of ArgumentException in DataStreamTest when reading past end of stream 

johnnypham:
[ArgumentOutOfRangeException is thrown in .NET 6 instead of ArgumentException](https://github.com/dotnet/runtime/issues/61994)

### DIFF
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataStreamTest/DataStreamTest.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataStreamTest/DataStreamTest.cs
@@ -1555,8 +1555,12 @@ CREATE TABLE {tableName} (id INT, foo VARBINARY(MAX))
                             DataTestUtility.AssertThrowsWrapper<ArgumentNullException>(() => stream.Read(null, 0, 1));
                             DataTestUtility.AssertThrowsWrapper<ArgumentOutOfRangeException>(() => stream.Read(buffer, -1, 2));
                             DataTestUtility.AssertThrowsWrapper<ArgumentOutOfRangeException>(() => stream.Read(buffer, 2, -1));
-                            Assert.ThrowsAny<ArgumentException>(() => stream.Read(buffer, buffer.Length, buffer.Length));
-                            Assert.ThrowsAny<ArgumentException>(() => stream.Read(buffer, int.MaxValue, int.MaxValue));
+
+                            // ArgumentException is thrown in net5 and earlier. ArgumentOutOfRangeException in net6 and later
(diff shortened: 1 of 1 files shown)


Kaur-Parminder (approved):

DavoudEshtehari (commented):
Could you be more specific and assess the result of the test if it's either `ArgumentException` or `ArgumentOutOfRangeException`? Using `ThrowsAny` includes the types which they're not valid and if we hit them in the future we won't catch them.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataStreamTest/DataStreamTest.cs
```diff
@@ -1555,8 +1555,12 @@ private static void ReadStream(string connectionString)
...
                             DataTestUtility.AssertThrowsWrapper<ArgumentOutOfRangeException>(() => stream.Read(buffer, 2, -1));
-                            DataTestUtility.AssertThrowsWrapper<ArgumentException>(() => stream.Read(buffer, buffer.Length, buffer.Length));
-                            DataTestUtility.AssertThrowsWrapper<ArgumentException>(() => stream.Read(buffer, int.MaxValue, int.MaxValue));
+
+                            // ArgumentException is thrown in net5 and earlier. ArgumentOutOfRangeException in net6 and later
+                            ArgumentException ex = Assert.ThrowsAny<ArgumentException>(() => stream.Read(buffer, buffer.Length, buffer.Length));
+                            Assert.True(ex.GetType() == typeof(ArgumentException) || ex.GetType() == typeof(ArgumentOutOfRangeException));
+                            ex = Assert.ThrowsAny<ArgumentException>(() => stream.Read(buffer, int.MaxValue, int.MaxValue));
+                            Assert.True(ex.GetType() == typeof(ArgumentException) || ex.GetType() == typeof(ArgumentOutOfRangeException));
```
DavoudEshtehari, line 1563:
```suggestion
                            Assert.True(ex.GetType() == typeof(ArgumentException) || ex.GetType() == typeof(ArgumentOutOfRangeException),
                                        "Expected: ArgumentException in net5 and earlier. ArgumentOutOfRangeException in net6 and later.");
                            ex = Assert.ThrowsAny<ArgumentException>(() => stream.Read(buffer, int.MaxValue, int.MaxValue));
                            Assert.True(ex.GetType() == typeof(ArgumentException) || ex.GetType() == typeof(ArgumentOutOfRangeException),
                                        "Expected: ArgumentException in net5 and earlier. ArgumentOutOfRangeException in net6 and later.");
```

DavoudEshtehari (approved):

PR #1022, opened 2021-04-06
Sync Crypto api usage

Wraith2:
Follow up from https://github.com/dotnet/SqlClient/pull/887

This merges some of the crypto related files and moved platform specific crypto into a single file `AlwaysEncryptedKeyConverter` which now has Cng and CrossPlatform versions. It is not possible to use the cross platform version on all build targets for three reasons, 
1) netfx does not have a way to create RSA or ECDiffieHelman from their parameters types without using the Cng class constructors directly.
2) netfx does not have the same ECDiffieHelman classes as the cross platform api does which means keys have to handled differently.
3) netstandard2.0 does not have ECDiffieHelman at all

These differences are encapsulated in the `AlwaysEncryptedKeyConverter` class now and consumers should be able to use or extend that class instead of needing to spread different crypto usage throughout multiple classes.

These changes make it possible to share a lot of the enclave support classes which should make maintaining them easier.
I have no way to test a lot of this locally (beyond compiling) so we'll need to rely on the CI tests.

PR #1419, opened 2021-12-02
Rename SimulatorEnclaveProvider - "None" attestation protocol, phase 1

johnnypham:
**Summary of feature**
Currently, VBS enclaves are supported on-prem and SGX enclaves on Azure VMs. VBS support for Azure is in the works but strong attestation is not possible on Azure VMs due to restrictions against accessing the host machine's Trusted Platform Module (TPM). 
A new attestation protocol called "None" will be allowed in the connection string, allowing users to forgo enclave attestation when using VBS enclaves: `Attestation Protocol = None;` 

**Phase 1 changes**
The existing `SimulatorEnclaveProvider` performs no attestation and only derives the shared secret required to set up a secure enclave session, which is exactly what is needed. This PR renames it to the proposed `NoneAttestationEnclaveProvider` and the file can now be shared across netcore and netfx.

The simulator can be built with 
```bash
msbuild -p:Configuration="Release" -p:BuildSimulator=true
````

### CODE REVIEW

File: src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/NoneAttestationEnclaveProvider.cs [hunk 1]
```diff
@@ -29,18 +25,15 @@ internal override void GetEnclaveSession(EnclaveSessionParameters enclaveSession
         // Gets the information that SqlClient subsequently uses to initiate the process of attesting the enclave and to establish a secure session with the enclave.
         internal override SqlEnclaveAttestationParameters GetAttestationParameters(string attestationUrl, byte[] customData, int customDataLength)
         {
-            // The key derivation function and hash algorithm name are specified when key derivation is performed
-            ECDiffieHellman clientDHKey = ECDiffieHellman.Create();
-            clientDHKey.KeySize = 384;
-
-            return new SqlEnclaveAttestationParameters(2, new byte[] { }, clientDHKey);
+            ECDiffieHellman clientDHKey = KeyConverter.CreateECDiffieHellman(DiffieHellmanKeySize);
```
johnnypham, line 28: This change is required to share the file. See #1022

Wraith2: Could `new byte[] { }` be `Array.Empty<byte>()` or is it not available in all supported targets?

johnnypham: I think it'll work, I'll let the CI determine that 😄

Kaur-Parminder (approved):

src/Microsoft.Data.SqlClient/netfx/src/Microsoft.Data.SqlClient.csproj
```diff
@@ -629,9 +632,6 @@
     <Compile Include="Microsoft\Data\SqlClient\Server\sqlser.cs" />
     <Compile Include="Microsoft\Data\Sql\SqlGenericUtil.cs" />
   </ItemGroup>
-  <ItemGroup Condition="'$(BuildSimulator)' == 'true'">
-    <Compile Include="Microsoft\Data\SqlClient\SimulatorEnclaveProvider.cs" />
-  </ItemGroup>
```
JRahnama, line 634: is removing the condition intentional?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/NoneAttestationEnclaveProvider.cs
```diff
@@ -3,21 +3,17 @@
...
-using System.Collections.Concurrent;
 
 namespace Microsoft.Data.SqlClient
 {
-    internal class SimulatorEnclaveProvider : EnclaveProviderBase
+    internal class NoneAttestationEnclaveProvider : EnclaveProviderBase
     {
         private static readonly int EnclaveSessionHandleSize = 8;
+        private const int DiffieHellmanKeySize = 384;
+        private const int NoneAttestationProtocolId = 2;
```
DavoudEshtehari, line 16: This value equals to `SqlConnectionAttestationProtocol.SIM` which builds only with `ENCLAVE_SIMULATOR=true` at build time. Could you add some notes to explain the source of the value **2**?

DavoudEshtehari (approved):

src/Microsoft.Data.SqlClient/netfx/src/Microsoft.Data.SqlClient.csproj
johnnypham: Yes, in phase 2, the newly named  `NoneAttestationEnclaveProvider` will now actually be used within the driver when `Attestation Protocol = None;` is specified. Before it was only used for testing purposes when the simulator flag was set.

JRahnama (approved):

PR #1433, opened 2021-12-14
Fix | SqlLocalDB instance pipename issue with Encryption

### LINKED ISSUE #1395: SqlException thrown related to encryption not being supported for SQL LocalDB named pipes
### Describe the bug

Using Microsoft.Data.SqlClient 4.0.0 with SQL LocalDB and named pipes does not work when attempting to connect to the instance.

I maintain a library that provides interop with the SQL LocalDB APIs, and upgrading its tests to version 4.0.0 is causing a number of its [tests to fail](https://github.com/martincostello/sqllocaldb/runs/4263534086?check_suite_focus=true).

An example of one of these failing tests is [here](https://github.com/martincostello/sqllocaldb/blob/4bf3f68798ff48b0f8999ea661fc11a39aa15fc1/tests/SqlLocalDb.Tests/SqlLocalDbApiTests.cs#L248-L287).

```
Exception message: The instance of SQL Server you attempted to connect to does not support encryption.
Stack trace:

[xUnit.net 00:01:35.98]     Can_Manage_SqlLocalDB_Instances [FAIL]
  Failed Can_Manage_SqlLocalDB_Instances [8 s]
  Error Message:
   Microsoft.Data.SqlClient.SqlException : The instance of SQL Server you attempted to connect to does not support encryption.
  Stack Trace:
   at Microsoft.Data.SqlClient.TdsParser.ConsumePreLoginHandshake(Boolean encrypt, Boolean trustServerCert, Boolean integratedSecurity, Boolean& marsCapable, Boolean& fedAuthRequired)
   at Microsoft.Data.SqlClient.TdsParser.Connect(ServerInfo serverInfo, SqlInternalConnectionTds connHandler, Boolean ignoreSniOpenTimeout, Int64 timerExpire, Boolean encrypt, Boolean trustServerCert, Boolean integratedSecurity, Boolean withFailover, SqlAuthenticationMethod authType)
   at Microsoft.Data.SqlClient.SqlInternalConnectionTds.AttemptOneLogin(ServerInfo serverInfo, String newPassword, SecureString newSecurePassword, Boolean ignoreSniOpenTimeout, TimeoutTimer timeout, Boolean withFailover)
   at Microsoft.Data.SqlClient.SqlInternalConnectionTds.LoginNoFailover(ServerInfo serverInfo, String newPassword, SecureString newSecurePassword, Boolean redirectedUserInstance, SqlConnectionString connectionOptions, SqlCredential credential, TimeoutTimer timeout)
   at Microsoft.Data.SqlClient.SqlInternalConnectionTds.OpenLoginEnlist(TimeoutTimer timeout, SqlConnectionString connectionOptions, SqlCredential credential, String newPassword, SecureString newSecurePassword, Boolean redirectedUserInstance)
   at Microsoft.Data.SqlClient.SqlInternalConnectionTds..ctor(DbConnectionPoolIdentity identity, SqlConnectionString connectionOptions, SqlCredential credential, Object providerInfo, String newPassword, SecureString newSecurePassword, Boolean redirectedUserInstance, SqlConnectionString userConnectionOptions, SessionData reconnectSessionData, Boolean applyTransientFaultHandling, String accessToken, DbConnectionPool pool)
   at Microsoft.Data.SqlClient.SqlConnectionFactory.CreateConnection(DbConnectionOptions options, DbConnectionPoolKey poolKey, Object poolGroupProviderInfo, DbConnectionPool pool, DbConnection owningConnection, DbConnectionOptions userOptions)
   at Microsoft.Data.ProviderBase.DbConnectionFactory.CreatePooledConnection(DbConnectionPool pool, DbConnection owningObject, DbConnectionOptions options, DbConnectionPoolKey poolKey, DbConnectionOptions userOptions)
   at Microsoft.Data.ProviderBase.DbConnectionPool.CreateObject(DbConnection owningObject, DbConnectionOptions userOptions, DbConnectionInternal oldConnection)
   at Microsoft.Data.ProviderBase.DbConnectionPool.UserCreateRequest(DbConnection owningObject, DbConnectionOptions userOptions, DbConnectionInternal oldConnection)
   at Microsoft.Data.ProviderBase.DbConnectionPool.TryGetConnection(DbConnection owningObject, UInt32 waitForMultipleObjectsTimeout, Boolean allowCreate, Boolean onlyOneCheckConnection, DbConnectionOptions userOptions, DbConnectionInternal& connection)
   at Microsoft.Data.ProviderBase.DbConnectionPool.WaitForPendingOpen()
   at MartinCostello.SqlLocalDb.SqlLocalDbApiTests.Can_Manage_SqlLocalDB_Instances() in D:\a\sqllocaldb\sqllocaldb\tests\SqlLocalDb.Tests\SqlLocalDbApiTests.cs:line 287
   at MartinCostello.SqlLocalDb.SqlLocalDbApiTests.Can_Manage_SqlLocalDB_Instances() in D:\a\sqllocaldb\sqllocaldb\tests\SqlLocalDb.Tests\SqlLocalDbApiTests.cs:line 288
```

### To reproduce

Clone the [martincostello/sqllocaldb](https://github.com/martincostello/sqllocaldb) repo on a Windows machine and checkout the [`dependabot/nuget/Microsoft.Data.SqlClient-4.0.0`](https://github.com/martincostello/sqllocaldb/tree/dependabot/nuget/Microsoft.Data.SqlClient-4.0.0) branch and run `build.ps1` to run the library's build and test script.

### Expected behavior

The tests pass.

### Further technical details

Microsoft.Data.SqlClient version: `4.0.0`
.NET target: `net6.0`
SQL Server version: SQL LocalDB 13
Operating system: Windows 2019

JRahnama:
fixes #1395 

This PR fixes the issue by adding the support for Instance pipe name of localdb.


ErikEJ:
Does this also cover Shared instance?

JRahnama:
@ErikEJ I think so, Every localdb (shared or not) is having a pipe name instance
image
that pipe name is good for shared instances as well. 
We were only checking for `(localdb)` as an identifier of localDB, however as per mentioned issue, some frameworks are using name pipe internally. This was not a problem up until we added `Encrypt=true` and now it tries to send and encrypted connection to localDB which is not supported.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIProxy.cs
```diff
@@ -381,6 +380,7 @@ internal class DataSource
         private const string Slash = @"/";
         private const string PipeToken = "pipe";
         private const string LocalDbHost = "(localdb)";
+        private const string LocalDbHost_NamedInstance = @"np:\\.\pipe\LOCALDB#";
```
DavoudEshtehari, line 383: NIT: please, update it in other files too.
```suggestion
        private const string LocalDbHost_NP = @"np:\\.\pipe\LOCALDB#";
```

DavoudEshtehari (commented):
Could you add `LocalDbSharedInstanceName` into the `config.json` and `config.default.json` files?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/LocalDBAPI.cs
```diff
@@ -2,33 +2,38 @@
...
 using System.Runtime.InteropServices;
 using System.Text;
 
 namespace Microsoft.Data
 {
     internal static partial class LocalDBAPI
     {
-        private const string const_localDbPrefix = @"(localdb)\";
+        private const string LocalDbPrefix = @"(localdb)\";
+        private const string LocalDbPrefix_namedInstance = @"np:\\.\pipe\LOCALDB#";
```
DavoudEshtehari, line 13: Could you update this for consistency?
```suggestion
        private const string LocalDbPrefix_NP = @"np:\\.\pipe\LOCALDB#";
```

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/LocalDBAPI.cs
```diff
@@ -19,8 +19,9 @@ namespace Microsoft.Data
 
     internal static class LocalDBAPI
     {
-        const string const_localDbPrefix = @"(localdb)\";
-        const string const_partialTrustFlagKey = "ALLOW_LOCALDB_IN_PARTIAL_TRUST";
+        private const string LocalDbPrefix = @"(localdb)\";
+        private const string LocalDbPrefix_namedInstance = @"np:\\.\pipe\LOCALDB#";
```
DavoudEshtehari, line 23:
```suggestion
        private const string LocalDbPrefix_NP = @"np:\\.\pipe\LOCALDB#";
```

DavoudEshtehari (approved):

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/LocalDBAPI.cs
```diff
@@ -2,33 +2,38 @@
...
-            if (!serverName.StartsWith(const_localDbPrefix, StringComparison.OrdinalIgnoreCase))
-                return null;
-            string instanceName = serverName.Substring(const_localDbPrefix.Length).Trim();
-            if (instanceName.Length == 0)
-                return null;
-            else
-                return instanceName;
+            string instanceName = null;
+            bool isLocalDb = serverName.StartsWith(LocalDbPrefix) || serverName.StartsWith(LocalDbPrefix_NP);
+            if (isLocalDb)
```
Kaur-Parminder, line 25: All StartsWith calls in this method requires StringComparison.OrdinalIgnoreCase.

johnnypham: Also, in the old code, `TrimStart()` is called on the server name but is missing in the new code?

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/LocalDBTest/LocalDBTest.cs
```diff
@@ -13,14 +15,18 @@ public static class LocalDBTest
         private static readonly string s_localDbConnectionString = @$"server=(localdb)\{DataTestUtility.LocalDbAppName}";
         private static readonly string[] s_sharedLocalDbInstances = new string[] { @$"server=(localdb)\.\{DataTestUtility.LocalDbSharedInstanceName}", @$"server=(localdb)\." };
         private static readonly string s_badConnectionString = $@"server=(localdb)\{DataTestUtility.LocalDbAppName};Database=DOES_NOT_EXIST;Pooling=false;";
+        private static readonly string s_commandPrompt = "cmd.exe";
+        private static readonly string s_sqlLocalDbInfo = @$"/c SqlLocalDb info {DataTestUtility.LocalDbAppName}";
+        private static readonly string s_startLocalDbCommand = @$"/c SqlLocalDb start {DataTestUtility.LocalDbAppName}";
+        private static readonly string s_localDbNamedPipeConnectionString = @$"server={GetLocalDbNamedPipe()}";
```
johnnypham, line 21: Should we test for local db names with leading white space?

Kaur-Parminder: @johnnypham Are you referring to space between /c SqlLocalDb ? This is probably because of https://stackoverflow.com/questions/515309/what-does-cmd-c-mean

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/LocalDBAPI.cs
JRahnama: I added the TrimStart and also added the OrdinalIgnoreCase for comparison. Thanks for catching it

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/LocalDBTest/LocalDBTest.cs
JRahnama: so, with adding TrimStart I dont think we need to test for leading whitespaces. Correct?

Kaur-Parminder (approved):

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/LocalDBAPI.cs
```diff
@@ -10,25 +10,32 @@ namespace Microsoft.Data
...
         private delegate int LocalDBFormatMessageDelegate(int hrLocalDB, uint dwFlags, uint dwLanguageId, StringBuilder buffer, ref uint buflen);
 
         // check if name is in format (localdb)\<InstanceName - not empty> and return instance name if it is
+        // localDB can also have a format of np:\\.\pipe\LOCALDB#<some number>\tsql\query
         internal static string GetLocalDbInstanceNameFromServerName(string serverName)
         {
-            if (serverName == null)
-                return null;
+            string instanceName = null;
             serverName = serverName.TrimStart(); // it can start with spaces if specified in quotes
```
johnnypham, line 25: There is no more null check for serverName so this could throw a NRE.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/LocalDBAPI.cs
```diff
@@ -10,25 +10,32 @@ namespace Microsoft.Data
...
+            bool isLocalDb = serverName.StartsWith(LocalDbPrefix, StringComparison.OrdinalIgnoreCase) || serverName.StartsWith(LocalDbPrefix_NP, StringComparison.OrdinalIgnoreCase);
+            if (isLocalDb)
+            {
+                if (serverName.StartsWith(LocalDbPrefix, StringComparison.OrdinalIgnoreCase))
+                {
+                    instanceName = serverName.Substring(LocalDbPrefix.Length).Trim();
+                }
+                else
+                {
+                    instanceName = serverName.TrimEnd();
```
johnnypham, line 35: So if the server name starts with LocalDbPrefix_NP, we will TrimEnd? I'm confused why this TrimEnd is needed.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/LocalDBTest/LocalDBTest.cs
JRahnama: not for local db, but we have tests for leading spaces in general as server name.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/LocalDBAPI.cs
```diff
@@ -30,16 +31,28 @@ internal static class LocalDBAPI
...
+                        {
+                            instanceName = serverName;
+                        }
+                    }
+                    else
+                    {
+                        instanceName = serverName.TrimEnd();
+                    }
+                }
+            }
```
DavoudEshtehari, line 54: Same note here.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/LocalDBAPI.cs
```diff
@@ -10,25 +10,39 @@ namespace Microsoft.Data
...
-                return null;
-            serverName = serverName.TrimStart(); // it can start with spaces if specified in quotes
-            if (!serverName.StartsWith(const_localDbPrefix, StringComparison.OrdinalIgnoreCase))
-                return null;
-            string instanceName = serverName.Substring(const_localDbPrefix.Length).Trim();
-            if (instanceName.Length == 0)
-                return null;
-            else
-                return instanceName;
+            string instanceName = null;
```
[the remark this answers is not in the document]
JRahnama: that is the approach we have taken in netcore, but I did not want to change the original code. If it is required I can change that.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/LocalDBAPI.cs
```diff
@@ -10,25 +10,41 @@ namespace Microsoft.Data
...
-                return null;
-            serverName = serverName.TrimStart(); // it can start with spaces if specified in quotes
-            if (!serverName.StartsWith(const_localDbPrefix, StringComparison.OrdinalIgnoreCase))
-                return null;
-            string instanceName = serverName.Substring(const_localDbPrefix.Length).Trim();
-            if (instanceName.Length == 0)
-                return null;
-            else
-                return instanceName;
+            if (serverName is not null)
```
johnnypham, line 24: The end of the server name is never trimmed. If a server name of "(localdb)\exampleInstance       " is used, then "exampleInstance       " is returned. Will it still connect?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/LocalDBAPI.cs
```diff
@@ -10,25 +10,41 @@ namespace Microsoft.Data
...
+            {
+                // it can start with spaces if specified in quotes
+                // Memory allocation is reduced by using ReadOnlySpan
+                ReadOnlySpan<char> input = serverName.AsSpan().TrimStart();
+                bool isLocalDb = input.StartsWith(LocalDbPrefix.AsSpan().Trim(), StringComparison.OrdinalIgnoreCase)
```
johnnypham, line 29: Why are the constants being trimmed here?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/LocalDBAPI.cs
```diff
@@ -10,25 +10,41 @@ namespace Microsoft.Data
...
+                    {
+                        input = input.Slice(LocalDbPrefix.Length);
+                        if (!input.IsEmpty)
+                        {
+                            return input.ToString();
+                        }
+                    }
+                    else
+                    {
+                        return input.ToString();
```
johnnypham, line 43: If the input does not start with LocalDbPrefix_NP, then the whole input is returned? It doesn't get sliced or trimmed?

JRahnama: yes. the input will be the entire servername starting with `np:\\.\pipe\LOCALDB#<number>\tsql\query`

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/LocalDBAPI.cs
JRahnama: I changed it to Trim. Same as managed SNI approach.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/LocalDBAPI.cs
JRahnama: fixed.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/LocalDBAPI.cs
JRahnama: if the input does not start with `LocalDbPrefix_NP` or `LocalDbPrefix` null will be returned

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/LocalDBAPI.cs
JRahnama: changed it to trim

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/LocalDBAPI.cs
```diff
@@ -10,25 +10,41 @@ namespace Microsoft.Data
...
-            if (instanceName.Length == 0)
-                return null;
-            else
-                return instanceName;
+            if (serverName is not null)
+            {
+                // it can start with spaces if specified in quotes
+                // Memory allocation is reduced by using ReadOnlySpan
+                ReadOnlySpan<char> input = serverName.AsSpan().Trim();
+                bool isLocalDb = input.StartsWith(LocalDbPrefix.AsSpan(), StringComparison.OrdinalIgnoreCase)
```
johnnypham, line 29: This bool is redundant. It could just be:

```csharp
  if (input.StartsWith(LocalDbPrefix.AsSpan(), StringComparison.OrdinalIgnoreCase))
                {
                        input = input.Slice(LocalDbPrefix.Length);
                        if (!input.IsEmpty)
                        {
                            return input.ToString();
                        }
                    }
else if (input.StartsWith(LocalDbPrefix_NP.AsSpan(), StringComparison.OrdinalIgnoreCase))
 {
                        return input.ToString();
`

Comment (JRahnama) [reply src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/LocalDBAPI.cs:29]: but I will change it right away.

File: src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/LocalDBAPI.cs [hunk 14]
```diff
@@ -10,25 +10,37 @@ namespace Microsoft.Data
...
+                {
+                    input = input.Slice(LocalDbPrefix.Length);
+                    if (!input.IsEmpty)
+                    {
+                        return input.ToString();
+                    }
+                }
+                else if (input.StartsWith(LocalDbPrefix_NP.AsSpan(), StringComparison.OrdinalIgnoreCase))
+                {
+                    return input.ToString();
```
johnnypham, line 39: I think the input should also be checked if it's empty here. I'm getting different errors when using the prefixes as the server:

`(localdb)\////`;
(provider: SNI_PN11, error: 51 - An instance name was not specified while connecting to a Local Database Runtime. Specify an instance name in the format (localdb)\instance_name.)

`np:\\.\pipe\LOCALDB#/`:
(provider: Named Pipes Provider, error: 40 - Could not open a connection to SQL Server)

JRahnama: this is user error. The error is pretty much explanatory and that is the behavior from before for (localdb).
1- (localdb)\.
2-(localdb)\Instance name
3- (localdb)\.\shared instance name

JRahnama: for named pipe:
the only part changing in `np:\\.\pipe\LOCALDB#SH452F48\tsql\query` is the number inside it

JRahnama: we could use other methods such as RegEx, but that would be an overkill. The fact that connection string starts with `np:\\.\pipe\LOCALDB` is enough to conclude that user is trying LocalDB

johnnypham: In the case of `(localdb)\` , we tell the user that the server name is not in the correct format, i.e. the instance name is missing. In the case of `np:\\.\pipe\LOCALDB#/`, shouldn't we also tell them that the instance name is missing?

JRahnama: > In the case of `(localdb)\` , we tell the user that the server name is not in the correct format, i.e. the instance name is missing. In the case of `np:\\.\pipe\LOCALDB#/`, shouldn't we also tell them that the instance name is missing?

I dont think so. Because that pattern is not created by user. It is in a folder in LocalDB inside Program File folder. So, you cant just create whatever you want. One can read the name from the folder using a process or copy paste it, whereas (localdb) could have some user mistakes. Plus are you testing in managed SNI?

johnnypham: Yes, it is the same in managed SNI.

The whole point of error handling is that we can't assume that the user knows what they're doing and got their named pipe server name from the correct source. If we assumed that the user does the right thing every time, there would be no need for error handling.

The error message when trying to connect to `np:\\.\pipe\LOCALDB#/` states that a connection could not be opened because the server was not found or was inaccessible. We should be more specific and say that the connection failed because the provided server name was not even in the correct format.

johnnypham (approved):

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/LocalDBAPI.cs
David-Engel: I see merit in aligning the errors produced between tcp and np scenarios. However, I wouldn't block this PR since we are not changing the behavior here. The behavior can be addressed in the future so that we don't delay fixing this issue for the upcoming hotfix.

PR #1442, opened 2021-12-22
Fix all docpaths to unix format paths

Wraith2:
A simple find-replace on windows style paths to unix style paths. This will make it easier to see code differences in the large files that need merging like SqlDataReader.

@DavoudEshtehari this contributes to merge,


DavoudEshtehari (approved):

Wraith2:
This is the result of find and replace on a very specific doc string and the library won't compile if the paths are wrong. Given that the CI is green and this cannot cause any behavioural changes just how much reviewing does it need?

JRahnama (approved):

johnnypham:
> This is the result of find and replace on a very specific doc string and the library won't compile if the paths are wrong. Given that the CI is green and this cannot cause any behavioural changes just how much reviewing does it need?

I had to count every single slash and period. Jk, approved.

johnnypham (approved):

PR #1461, opened 2022-01-11
Test | Add lock when using ClearSqlConnectionGlobalProviders

johnnypham:
`s_globalCustomColumnEncryptionKeyStoreProviders` can only be set once using `SqlConnection.RegisterColumnEncryptionKeyStoreProviders()`. It has to be cleared after any test sets it, so that other tests can set it again. Because the functional tests run in parallel, we sometimes see a `Key store providers cannot be set more than once.` error. We can add a lock to ensure only one test at a time is setting and clearing `s_globalCustomColumnEncryptionKeyStoreProviders`.

Repro: 

https://github.com/johnnypham/SqlClient/tree/testfailurerepro

Run the two tests with the appropriate filter: 
```bash
dotnet test "src\Microsoft.Data.SqlClient\tests\FunctionalTests\Microsoft.Data.SqlClient.Tests.csproj" -p:Platform="AnyCPU" -p:ReferenceType=Project -p:Configuration="Release" -p:TestTargetOS="Windowsnetcoreapp" --no-build --logger:"console;verbosity=normal" --filter "TestCanSetGlobalProvidersOnlyOnce|TestExceptionsFromCustomKeyStore"

### DIFF
--- a/src/Microsoft.Data.SqlClient/tests/FunctionalTests/AlwaysEncryptedTests/SqlColumnEncryptionCertificateStoreProviderShould.cs
+++ b/src/Microsoft.Data.SqlClient/tests/FunctionalTests/AlwaysEncryptedTests/SqlColumnEncryptionCertificateStoreProviderShould.cs
@@ -92,11 +92,6 @@ namespace Microsoft.Data.SqlClient.Tests.AlwaysEncryptedTests
         /// </summary>
         private const int CipherTextStartIndex = IVStartIndex + IVLengthInBytes;
 
-        /// <summary>
-        /// SetCustomColumnEncryptionKeyStoreProvider can be called only once in a process. To workaround that, we use this flag.
-        /// </summary>
-        private static bool s_testCustomEncryptioKeyStoreProviderExecutedOnce = false;
-
         [Theory]
         [InvalidDecryptionParameters]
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/Microsoft.Data.SqlClient/tests/FunctionalTests/AlwaysEncryptedTests/SqlColumnEncryptionCertificateStoreProviderShould.cs [hunk 1]
```diff
@@ -92,11 +92,6 @@ public class SqlColumnEncryptionCertificateStoreProviderWindowsShould : IClassFi
         /// </summary>
         private const int CipherTextStartIndex = IVStartIndex + IVLengthInBytes;
 
-        /// <summary>
-        /// SetCustomColumnEncryptionKeyStoreProvider can be called only once in a process. To workaround that, we use this flag.
-        /// </summary>
-        private static bool s_testCustomEncryptioKeyStoreProviderExecutedOnce = false;
```
johnnypham, line 98: This field was never even changed anywhere.

DavoudEshtehari (approved):

Kaur-Parminder (approved):

JRahnama (approved):

PR #1482, opened 2022-01-19
Test | Added connection Open check to DataUtility methods for Db entity cleanups

Kaur-Parminder:
Fix for Intermittent failure for CollatedDataReaderTest. (https://dev.azure.com/sqlclientdrivers-ci/sqlclient/_build/results?buildId=46069&view=ms.vss-test-web.build-test-results-tab&runId=497922&resultId=100031&paneView=debug).
The connection closes before cleanup in finally block. I updated the DataUtility methods to open connection to do mandatory cleanup if there is ever connection failure and wrapped the command execution in try catch block as they are mostly invoked from finally block.

### DIFF
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataReaderTest/DataReaderTest.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataReaderTest/DataReaderTest.cs
@@ -125,9 +125,9 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests
             // Remove square brackets
             var dbName = databaseName.Substring(1, databaseName.Length - 2);
             using SqlConnection con = new(DataTestUtility.TCPConnectionString);
+            con.Open();
             try
             {
-                con.Open();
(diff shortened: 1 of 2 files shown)


src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataReaderTest/DataReaderTest.cs
```diff
@@ -165,9 +156,27 @@ public static void CollatedDataReaderTest()
...
+        }
+
+        private static void CreateCollatedDB(SqlConnection con,string databaseName)
+        {
+            try
+            {
+                if (con.State != ConnectionState.Open)
+                {
+                    con.Open();
+                }
```
DavoudEshtehari, line 170: Beign more specific.
```suggestion
                if (con.State ==ConnectionState.Connecting)
                {
                    Thread.Sleep(300);
                }
                else if (con.State == ConnectionState.Closed || con.State == ConnectionState.Broken)
                {
                    con.Open();
                }
```

David-Engel:
No longer needed after changes from #1658.

Issue #1506, opened 2022-02-04
CleanupExecuteReaderAsync OperationId is empty GUID


AndrewP-GH:
### WriteCommandAfter: CleanupExecuteReaderAsync has empty GUID as OperationId
Hello!

I'm writing the library for tracing Sql.Client. And I ran into a problem that the `CleanupExecuteReaderAsync` operation (or correctly call it _command_), which is produced as "Microsoft.Data.SqlClient.WriteCommandAfter" event, contains `Guid.Empty` OperationId. I've been confused with this behaviour because it is "after" command.
 

Event example:
```
{
    OperationId: 00000000 - 0000 - 0000 - 0000 - 000000000000,
    Operation: "CleanupExecuteReaderAsync",
    ConnectionId: 910 f6790 - b10b - 4143 - b723 - 915229 b16bb5,
    Command: SqlCommand {
        Connection: SqlConnection {
            RetryLogicProvider: SqlRetryLogicProvider {
                Retrying: null,
                RetryLogic: SqlRetryLogic {
                    PreCondition: null,
                    NumberOfTries: 1,
                    Current: 0,
                    RetryIntervalEnumerator: SqlNoneIntervalEnumerator {
                        GapTimeInterval: 00: 00: 00,
                        MaxTimeInterval: 00: 00: 00,
                        MinTimeInterval: 00: 00: 00,
                        Current: 00: 00: 00
                    },
                    TransientPredicate: "System.Predicate`1[System.Exception]"
                }
            },
            StatisticsEnabled: False,
            ConnectionString: "",
            ConnectionTimeout: 15,
            CommandTimeout: 30,
            AccessToken: null,
            Database: "",
            DataSource: "",
            PacketSize: 8000,
            ClientConnectionId: 910 f6790 - b10b - 4143 - b723 - 915229 b16bb5,
            ServerVersion: "The property accessor threw an exception: InvalidOperationException",	// this cause because SqlConnection.get_ServerVersion() uses SqlConnection.GetOpenTdsConnection() that throws ADP.ClosedConnectionError() if SqlInternalConnectionTds innerConnection is null when I dump this event
            ServerProcessId: 0,
            State: Closed,
            WorkstationId: "pay-994d47d-qwls4",
            Credential: null,
            FireInfoMessageEventOnUserErrors: False,
            CanCreateBatch: False,
            Site: null,
            Container: null
        },
        RetryLogicProvider: SqlRetryLogicProvider {
            Retrying: null,
[...]

PR #1511, opened 2022-02-09
Set _parentOperationStarted in ExecuteReaderAsync continuation

ahydrax:
This fixes https://github.com/dotnet/SqlClient/issues/1506.

The problem was there https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs#L2777
`_parentOperationStarted ` flag has to be set explicitly after `ExecuteReaderAsync` i.e. in its continuation.

Reproduction code is available here: https://gist.github.com/AndrewP-GH/1e1ad8d2b864ed549e31e5f03fee015e


Wraith2:
Does this prevent the event that occurs after the parent is unset from having a guid at all? Is that better than saving the parent guid and using it later?

ahydrax:
@Wraith2 
> Does this prevent the event that occurs after the parent is unset from having a guid at all?

Since its hard to write a non-flaky test for that case I can only assume that. At least now it looks like a correct control flow: we for sure unset the flag right after child operation completes. I tried to launch a repro code several times and got no issues with that.
I was also thinking about attaching `_parentOperationStarted = false;` in a separated continuation to outer task. But from my perspective it just adds an extra overhead for TaskScheduler. What do you think?

> Is that better than saving the parent guid and using it later?

It might be also a solution to be discussed.

ahydrax:
@Wraith2 could you help me with test pipeline? https://dev.azure.com/sqlclientdrivers-ci/sqlclient/_build/results?buildId=47677&view=logs&jobId=6399d89d-3d30-5fff-3e92-07bbaf7ebe2f&j=6399d89d-3d30-5fff-3e92-07bbaf7ebe2f&t=d757e160-f926-47ff-8096-49a8c7da6621 it looks like tests didn't even start

Wraith2:
I don't have CI access either. You can push a tiny change commit, close and re-open the PR or just wait if you think they're false positives and the MS team will probably re-run them when they get to review.

ahydrax:
@Wraith2 the pipeline is green now. Could you assign or tag someone who can review this?

Wraith2:
I'm an external contributor like you, I don't have any permissions. The MS team will pick it up when they can.

ahydrax:
Any news? We would like to have fixed tracing 😸 
@Wraith2 @johnnypham @DavoudEshtehari @David-Engel

DavoudEshtehari:
@ahydrax, attempting to run the [repro](https://gist.github.com/AndrewP-GH/1e1ad8d2b864ed549e31e5f03fee015e) with this change using a .NET 6.0 console app, I got the NRE while it doesn't break with the latest source code. 
Can you verify if it's only me?

```
Unhandled exception. System.AggregateException: One or more errors occurred. (Object reference not set to an instance of an object.)
 ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.Data.SqlClient.SqlConnection.OpenAsync(CancellationToken cancellationToken) in C:\xxx\src\Microsoft.Data.SqlClient\netcore\ref\Microsoft.Data.SqlClient.cs:line 892
   at ConsoleApp1.Program.<>c.<<RunTest>b__1_0>d.MoveNext() in C:\xxx\ConsoleApp1\Program.cs:line 36
   at ConsoleApp1.Program.<>c.<<RunTest>b__1_0>d.MoveNext() in C:\xxx\ConsoleApp1\Program.cs:line 40
   at System.Threading.Tasks.Parallel.<>c__50`1.<<ForEachAsync>b__50_0>d.MoveNext()
   at ConsoleApp1.Program.RunTest() in C:\xxx\ConsoleApp1\Program.cs:line 33
   at System.Threading.Tasks.Task.ThrowIfExceptional(Boolean includeTaskCanceledExceptions)
   at System.Threading.Tasks.Task.Wait(Int32 millisecondsTimeout, CancellationToken cancellationToken)
   at System.Threading.Tasks.Task.Wait()
   at ConsoleApp1.Program.Main(String[] args) in C:\xxx\ConsoleApp1\Program.cs:line 14
```

David-Engel:
I've run the repro with and without the fix. In both cases, I still get empty GUIDs in the repro. There are usually fewer empties with the fix, but something is still missing. For reference, without the fix, I see 20-30 empty GUIDs. With the fix, it's 10-15.

ahydrax:
Thanks for update!
@David-Engel could you tell me which framework/os you used to run the repro? I don't get empty guids on windows 11 x64

DavoudEshtehari:
I got the empty GUIDs on Win 10 (version 21H2) and `.NET 6.0` using a core i7 CPU.

ahydrax:
I tried to reproduce this bug on `5.1.0` and seems it's fixed now.

Issue #1501, opened 2022-02-02
Should AlwaysEncryptedAttestationException be public?


kmscode:
Reference file: src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AlwaysEncryptedAttestationException.cs

Intermittently / Randomly (though more frequently when attempting to stress test my applications code I will receive the below AlwaysEncryptedAttestationException. However, I cannot directly catch this specific exception - I assume due to the class being internal.

Ideally I would like to retry in such a scenario.

Should this class be public?

![image](https://user-images.githubusercontent.com/16654335/152079993-a3e4f5db-0182-4f2f-a0e5-2368434d678f.png)

```
Microsoft.Data.SqlClient.AlwaysEncryptedAttestationException: The validation of an attestation token failed. Cannot retrieve a public key from the attestation public key endpoint, or the retrieved key has an invalid format. Error details: 'A task was canceled.'.
 ---> System.AggregateException: One or more errors occurred. (IDX20803: Unable to obtain configuration from: 'https://name.region.attest.azure.net/.well-known/openid-configuration'.)
 ---> System.InvalidOperationException: IDX20803: Unable to obtain configuration from: 'https://name.region.attest.azure.net/.well-known/openid-configuration'.
 ---> System.IO.IOException: IDX20804: Unable to retrieve document from: 'https://name.region.attest.azure.net/.well-known/openid-configuration'.
 ---> System.Threading.Tasks.TaskCanceledException: The request was canceled due to the configured HttpClient.Timeout of 100 seconds elapsing.
 ---> System.TimeoutException: A task was canceled.
 ---> System.Threading.Tasks.TaskCanceledException: A task was canceled.
   at System.Threading.Tasks.TaskCompletionSourceWithCancellation`1.WaitWithCancellationAsync(CancellationToken cancellationToken)
   at System.Net.Http.HttpConnectionPool.GetHttp11ConnectionAsync(HttpRequestMessage request, Boolean async, CancellationToken cancellationToken)
   at System.Net.Http.HttpConnectionPool.SendWithVersionDetectionAndRetryAsync(HttpRequestMessage request, Boolean async, Boolean doRequestAuth, CancellationToken cancellationToken)
   at System.Net.Http.DiagnosticsHandler.SendAsyncCore(HttpRequestMessage request, Boolean async, CancellationToken cancellationToken)
[...]

PR #1515, opened 2022-02-11
Replace AlwaysEncryptedAttestationException with SqlException

johnnypham:
Addresses #1501 

The exception type is not public and is thrown for various error scenarios during enclave attestation. We can replace it with a catchable SqlException in a default state with the original message and inner exception. This is the same behaviour in related drivers.


Wraith2 (approved):
Looks good to me.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AzureAttestationBasedEnclaveProvider.cs
```diff
@@ -213,7 +213,7 @@ public AzureAttestationInfo(byte[] attestationInfo)
                 }
                 catch (Exception exception)
                 {
-                    throw new AlwaysEncryptedAttestationException(string.Format(Strings.FailToParseAttestationInfo, exception.Message));
+                    throw SQL.AttestationFailed(string.Format(Strings.FailToParseAttestationInfo, exception.Message));
```
Kaur-Parminder, line 216: exceptions can be passed as inner exception to SqlException? not just the message. same on line 429 and 504

Wraith2: It is possible that passing the full exception would expose an internal exception type to users. 

If the exception that can be caught here might come from users code then it seems like a good idea to pass it back to them as the inner exception because it might provide extra context information about the failure. If the exception can only be internal to this library then I think simply concatenating the message is sufficient.

DavoudEshtehari (approved):

Kaur-Parminder (approved):

JRahnama (approved):

PR #1544, opened 2022-03-13
Add and use TryReadChars method

Wraith2:
In https://github.com/dotnet/SqlClient/issues/593#issuecomment-1041190401 @panoskj pointed out that there was a path in the example reproduction which was copying characters from a buffer one at a time in a loop and that this was inefficient. I've done a little digging and this is an attempt to speed it up. Note that not all strings are copied this way.

Unfortunately the fix isn't as simple as copying into from the byte[]. We need to make sure that if we reach a packet boundary that we call a function capable of fetching the next packet. We also need to be aware that we could end up with a partial character in the input buffer and that needs special handling. The strategy I've settled on for this PR is to identify how much can be copied in bulk from the current packet and then pick up any remainder or simply force a new packet by calling TryReadChar() as the code currently does. This should result in most data being read in large blocks with single chars being picked up at existing boundaries which should be faster if not the fastest possible.

/cc @panoskj could you take a look and see if you agree with this and possibly run it through your benches/profiler to see if it improves the numbers you saw?

### DIFF
--- a/src/Microsoft.Data.SqlClient.sln
+++ b/src/Microsoft.Data.SqlClient.sln
@@ -205,6 +205,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{4F3CD363-B1E
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.SqlServer.Server", "Microsoft.SqlServer.Server\Microsoft.SqlServer.Server.csproj", "{A314812A-7820-4565-A2A8-ABBE391C11E4}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SqlBench3.0", "..\..\dev\SqlClient-Integrated\SqlBench3.0\SqlBench3.0.csproj", "{1BCC919A-CA34-4A5C-B0C3-69A16FA844D7}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
@@ -489,6 +491,18 @@ Global
 		{A314812A-7820-4565-A2A8-ABBE391C11E4}.Release|x64.Build.0 = Release|Any CPU
 		{A314812A-7820-4565-A2A8-ABBE391C11E4}.Release|x86.ActiveCfg = Release|Any CPU
 		{A314812A-7820-4565-A2A8-ABBE391C11E4}.Release|x86.Build.0 = Release|Any CPU
+		{1BCC919A-CA34-4A5C-B0C3-69A16FA844D7}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{1BCC919A-CA34-4A5C-B0C3-69A16FA844D7}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{1BCC919A-CA34-4A5C-B0C3-69A16FA844D7}.Debug|x64.ActiveCfg = Debug|Any CPU
+		{1BCC919A-CA34-4A5C-B0C3-69A16FA844D7}.Debug|x64.Build.0 = Debug|Any CPU
+		{1BCC919A-CA34-4A5C-B0C3-69A16FA844D7}.Debug|x86.ActiveCfg = Debug|Any CPU
+		{1BCC919A-CA34-4A5C-B0C3-69A16FA844D7}.Debug|x86.Build.0 = Debug|Any CPU
+		{1BCC919A-CA34-4A5C-B0C3-69A16FA844D7}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{1BCC919A-CA34-4A5C-B0C3-69A16FA844D7}.Release|Any CPU.Build.0 = Release|Any CPU
+		{1BCC919A-CA34-4A5C-B0C3-69A16FA844D7}.Release|x64.ActiveCfg = Release|Any CPU
+		{1BCC919A-CA34-4A5C-B0C3-69A16FA844D7}.Release|x64.Build.0 = Release|Any CPU
+		{1BCC919A-CA34-4A5C-B0C3-69A16FA844D7}.Release|x86.ActiveCfg = Release|Any CPU
+		{1BCC919A-CA34-4A5C-B0C3-69A16FA844D7}.Release|x86.Build.0 = Release|Any CPU
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
@@ -525,6 +539,7 @@ Global
 		{B93A3149-67E8-491E-A1E5-19D65F9D9E98} = {0CC4817A-12F3-4357-912C-09315FAAD008}
 		{599A336B-2A5F-473D-8442-1223ED37C93E} = {0CC4817A-12F3-4357-912C-09315FAAD008}
 		{A314812A-7820-4565-A2A8-ABBE391C11E4} = {4F3CD363-B1E6-4D6D-9466-97D78A56BE45}
+		{1BCC919A-CA34-4A5C-B0C3-69A16FA844D7} = {28E5EFE6-C9DD-4FF9-9FEC-532F72DFFA6E}
 	EndGlobalSection
 	GlobalSection(ExtensibilityGlobals) = postSolution
 		SolutionGuid = {01D48116-37A2-4D33-B9EC-94793C702431}
--- a/src/Microsoft.Data.SqlClient.sln
+++ b/src/Microsoft.Data.SqlClient.sln
@@ -205,8 +205,6 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{4F3CD363-B1E
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.SqlServer.Server", "Microsoft.SqlServer.Server\Microsoft.SqlServer.Server.csproj", "{A314812A-7820-4565-A2A8-ABBE391C11E4}"
 EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SqlBench3.0", "..\..\dev\SqlClient-Integrated\SqlBench3.0\SqlBench3.0.csproj", "{1BCC919A-CA34-4A5C-B0C3-69A16FA844D7}"
-EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
@@ -491,18 +489,6 @@ Global
 		{A314812A-7820-4565-A2A8-ABBE391C11E4}.Release|x64.Build.0 = Release|Any CPU
 		{A314812A-7820-4565-A2A8-ABBE391C11E4}.Release|x86.ActiveCfg = Release|Any CPU
 		{A314812A-7820-4565-A2A8-ABBE391C11E4}.Release|x86.Build.0 = Release|Any CPU
-		{1BCC919A-CA34-4A5C-B0C3-69A16FA844D7}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
-		{1BCC919A-CA34-4A5C-B0C3-69A16FA844D7}.Debug|Any CPU.Build.0 = Debug|Any CPU
-		{1BCC919A-CA34-4A5C-B0C3-69A16FA844D7}.Debug|x64.ActiveCfg = Debug|Any CPU
-		{1BCC919A-CA34-4A5C-B0C3-69A16FA844D7}.Debug|x64.Build.0 = Debug|Any CPU
-		{1BCC919A-CA34-4A5C-B0C3-69A16FA844D7}.Debug|x86.ActiveCfg = Debug|Any CPU
-		{1BCC919A-CA34-4A5C-B0C3-69A16FA844D7}.Debug|x86.Build.0 = Debug|Any CPU
-		{1BCC919A-CA34-4A5C-B0C3-69A16FA844D7}.Release|Any CPU.ActiveCfg = Release|Any CPU
-		{1BCC919A-CA34-4A5C-B0C3-69A16FA844D7}.Release|Any CPU.Build.0 = Release|Any CPU
-		{1BCC919A-CA34-4A5C-B0C3-69A16FA844D7}.Release|x64.ActiveCfg = Release|Any CPU
-		{1BCC919A-CA34-4A5C-B0C3-69A16FA844D7}.Release|x64.Build.0 = Release|Any CPU
-		{1BCC919A-CA34-4A5C-B0C3-69A16FA844D7}.Release|x86.ActiveCfg = Release|Any CPU
-		{1BCC919A-CA34-4A5C-B0C3-69A16FA844D7}.Release|x86.Build.0 = Release|Any CPU
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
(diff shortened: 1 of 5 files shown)


panoskj:
I have not run any benchmarks with it yet, but it seems like a reasonable change. It should speed up both sync and async paths.

It also looks like you could replace usages of TryReadPlpUnicodeChars with TryReadStringWithEncoding (passing Encoding.Unicode as encoding parameter and doubled string size), if the result will be converted to a string anyway. This is the case inside TryReadSqlStringValue for example. This way you can avoid the case of "partially read characters" altogether. It looks like TryReadPlpUnicodeChars is only needed for SqlDataReader.GetChars actually.

Though, this change cannot improve the async path much, because as I mentioned in [#593 (comment)](https://github.com/dotnet/SqlClient/issues/593#issuecomment-1041190401), its time complexity is _quadratic_. No matter how fast you make the copying itself, all packets in the replay buffer will be copied into the characters buffer for each packet received again and again. 

The optimal solution would copy each packet once instead. See https://github.com/panoskj/SqlClient/tree/wip/issue/593 for a proof of concept. Async runs about as fast as the sync does, even for very large columns, e.g. 50MB. Note that I haven't reviewed the last commit much, it wasn't meant to be uploaded yet.

In conclusion, perhaps we should merge the netcore and netfx files before applying a dozen of new patches to them. But except for the pending merge, I agree with this change.

Wraith2:
> It also looks like you could replace usages of TryReadPlpUnicodeChars with TryReadStringWithEncoding (passing Encoding.Unicode as encoding parameter and doubled string size), if the result will be converted to a string anyway. This is the case inside TryReadSqlStringValue for example. This way you can avoid the case of "partially read characters" altogether. It looks like TryReadPlpUnicodeChars is only needed for SqlDataReader.GetChars actually.

Possibly. Identifying whether the behaviour is the same can be time consuming.

> Though, this change cannot improve the async path much, because as I mentioned in https://github.com/dotnet/SqlClient/issues/593#issuecomment-1041190401, its time complexity is quadratic. No matter how fast you make the copying itself, all packets in the replay buffer will be copied into the characters buffer for each packet received again and again.

We both have branches which can linearize that, just waiting on the tree merges.

> In conclusion, perhaps we should merge the netcore and netfx files before applying a dozen of new patches to them. But except for the pending merge, I agree with this change.

I've love to. Tree merging is going slowly though. If I can get a small PR through to improve performance I'll try it.

Wraith2:
It looks like this doesn't work with some of the bulkcopy tests.

Wraith2:
I updated without making any material change and the CI is mostly clean. The only failures are in AE and while they do look look related I have no way to replicate that environment to debug them.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -12392,18 +12394,19 @@ private bool TryReadPlpUnicodeCharsChunk(char[] buff, int offst, int len, TdsPar
                 return true;
             }
 
-            charsRead = len;
+            int charsToRead = len;
 
             // stateObj._longlenleft is in bytes
-            if ((stateObj._longlenleft >> 1) < (ulong)len)
-                charsRead = (int)(stateObj._longlenleft >> 1);
+            if ((stateObj._longlenleft / 2) < (ulong)len)
```
David-Engel, line 12400: Why the change from '>> 1' to '/ 2'?

Wraith2: I needed a petty change to force the CI to run. I think the division is clearer and that the jit will do the same thing but I'm not going to argue strongly either way.

David-Engel: Bit shifting has always been confusing to me but after some "learning", I get it now. I have no problems with this either way.

src/Microsoft.Data.SqlClient.sln
```diff
@@ -204,6 +204,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{4F3CD363-B1E
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.SqlServer.Server", "Microsoft.SqlServer.Server\Microsoft.SqlServer.Server.csproj", "{A314812A-7820-4565-A2A8-ABBE391C11E4}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SqlBench3.0", "..\..\dev\SqlClient-Integrated\SqlBench3.0\SqlBench3.0.csproj", "{1BCC919A-CA34-4A5C-B0C3-69A16FA844D7}"
```
lcheunglci, line 207: I don't think project exists the the repo

Wraith2: It doesn't, sln changes reverted.

DavoudEshtehari (commented):
@Wraith2 LGTM, do you have any plan to port it into netfx too?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
```diff
@@ -1465,6 +1465,52 @@ internal bool TryReadChar(out char value)
...
+        internal bool TryReadChars(char[] chars, int charsOffset, int charsCount, out int charsCopied)
+        {
+            charsCopied = 0;
+            while (charsCopied < charsCount)
+            {
+                // check if the current buffer contains some bytes we need to copy and copy them
+                //  in a block
+                int bytesToRead = Math.Min(
+                    (charsCount - charsCopied) * 2,
+                    unchecked((_inBytesRead - _inBytesUsed) & (int)0xFFFFFFFE) // it the result is odd take off the 0 to make it even
```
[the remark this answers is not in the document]
Wraith2: N specific plans. I'm still hoping that the need to do ports will be removed by the codebase merge which is in progress.

DavoudEshtehari:
PR #1520 is the prerequisite of this change.

Wraith2:
Prerequisites are met...

panoskj:
I see you haven't modified any existing code inside TdsParserStateObject - you only added a new method. This means, you can now update this PR to add the new method inside the merged file (src\Microsoft\Data\SqlClient\TdsParserStateObject.cs) instead of inside the corefx version. Of course, to do this, you will have to merge main branch into this one (to get #1520's changes).

Moreover, according to [this comment](https://github.com/dotnet/SqlClient/pull/1843#issuecomment-1322890512), there may be someone working on merging TdsParser already, but there is no PR yet (they are waiting for #1844, which [will help them a lot](https://github.com/dotnet/SqlClient/pull/1844#discussion_r1030882488)). You could port the changes to TryReadPlpUnicodeCharsChunk to netfx in order to help them. Alternatively, you could merge TryReadPlpUnicodeCharsChunk into a new common file (just merge this one method and leave the rest). That being said, I haven't checked whether TryReadPlpUnicodeCharsChunk is the same in corefx and netfx - if there are significant changes, porting/merging may be too hard. But I suspect they are identical.

Wraith2:
rebased, moved to shared and cleaned up slightly

Wraith2:
CI failures are infrastructure issues or unreliable reliability test failures. nothing appears linked to the changes made here.

JRahnama:
this one looks related:
image

Wraith2:
passes locally:
image

David-Engel:
The failure is only on Linux with AE and it is definitely right in your new code. The call stack goes through TryReadPlpUnicodeChars -> TryReadPlpUnicodeCharsChunk (strangely not shown in the stack - release optimized out?) [your code changes] -> TryReadChars [your code changes] -> TryReadChar -> TryReadByteArray -> TryReadByteArray.

```
Failed Microsoft.Data.SqlClient.ManualTesting.Tests.AlwaysEncrypted.SqlBulkCopyTruncation.BulkCopyNVarchar(connectionString: "Data Source=tcp:10.0.0.4;Database=NORTHWIND;UID=sa"...) [35 s]
     Assert.Throws() Failure
  Expected: typeof(System.InvalidOperationException)
  Actual:   typeof(System.ArgumentOutOfRangeException): Specified argument was out of the range of valid values.
  ---- System.ArgumentOutOfRangeException : Specified argument was out of the range of valid values.
    Stack Trace:
       at Microsoft.Data.SqlClient.TdsParserStateObject.TryReadByteArray(Span`1 buff, Int32 len, Int32& totalRead) in /mnt/vss/_work/1/s/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs:line 400
     at Microsoft.Data.SqlClient.TdsParserStateObject.TryReadByteArray(Span`1 buff, Int32 len) in /mnt/vss/_work/1/s/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs:line 354
     at Microsoft.Data.SqlClient.TdsParserStateObject.TryReadChar(Char& value) in /mnt/vss/_work/1/s/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs:line 470
     at Microsoft.Data.SqlClient.TdsParserStateObject.TryReadChars(Char[] chars, Int32 charsOffset, Int32 charsCount, Int32& charsCopied) in /mnt/vss/_work/1/s/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs:line 468
     at Microsoft.Data.SqlClient.TdsParser.TryReadPlpUnicodeChars(Char[]& buff, Int32 offst, Int32 len, TdsParserStateObject stateObj, Int32& totalCharsRead, Boolean supportRentedBuff, Boolean& rentedBuff) in /mnt/vss/_work/1/s/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs:line 12621
     at Microsoft.Data.SqlClient.TdsParser.TryReadSqlStringValue(SqlBuffer value, Byte type, Int32 length, Encoding encoding, Boolean isPlp, TdsParserStateObject stateObj) in /mnt/vss/_work/1/s/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs:line 5706
     at Microsoft.Data.SqlClient.TdsParser.TryReadSqlValue(SqlBuffer value, SqlMetaDataPriv md, Int32 length, TdsParserStateObject stateObj, SqlCommandColumnEncryptionSetting columnEncryptionOverride, String columnName, SqlCommand command) in /mnt/vss/_work/1/s/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs:line 6083
     at Microsoft.Data.SqlClient.SqlDataReader.TryReadColumnInternal(Int32 i, Boolean readHeaderOnly, Boolean forStreaming) in /mnt/vss/_work/1/s/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs:line 3915
     at Microsoft.Data.SqlClient.SqlDataReader.GetValueInternal(Int32 i) in /mnt/vss/_work/1/s/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs:line 2713
     at Microsoft.Data.SqlClient.SqlDataReader.GetValue(Int32 i) in /mnt/vss/_work/1/s/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs:line 2665
     at Microsoft.Data.SqlClient.SqlBulkCopy.GetValueFromSourceRow(Int32 destRowIndex, Boolean& isSqlType, Boolean& isDataFeed, Boolean& isNull) in /mnt/vss/_work/1/s/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs:line 956
     at Microsoft.Data.SqlClient.SqlBulkCopy.ReadWriteColumnValueAsync(Int32 col) in /mnt/vss/_work/1/s/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs:line 2158
     at Microsoft.Data.SqlClient.SqlBulkCopy.CopyColumnsAsync(Int32 col, TaskCompletionSource`1 source) in /mnt/vss/_work/1/s/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs:line 2232
     at Microsoft.Data.SqlClient.SqlBulkCopy.CopyRowsAsync(Int32 rowsSoFar, Int32 totalRows, CancellationToken cts, TaskCompletionSource`1 source) in /mnt/vss/_work/1/s/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs:line 2400
     at Microsoft.Data.SqlClient.SqlBulkCopy.CopyBatchesAsyncContinued(BulkCopySimpleResultSet internalResults, String updateBulkCommandText, CancellationToken cts, TaskCompletionSource`1 source) in /mnt/vss/_work/1/s/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs:line 2570
     at Microsoft.Data.SqlClient.SqlBulkCopy.CopyBatchesAsync(BulkCopySimpleResultSet internalResults, String updateBulkCommandText, CancellationToken cts, TaskCompletionSource`1 source) in /mnt/vss/_work/1/s/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs:line 2500
     at Microsoft.Data.SqlClient.SqlBulkCopy.WriteToServerInternalRestContinuedAsync(BulkCopySimpleResultSet internalResults, CancellationToken cts, TaskCompletionSource`1 source) in /mnt/vss/_work/1/s/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs:line 2769
     at Microsoft.Data.SqlClient.SqlBulkCopy.WriteToServerInternalRestAsync(CancellationToken cts, TaskCompletionSource`1 source) in /mnt/vss/_work/1/s/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs:line 2997
     at Microsoft.Data.SqlClient.SqlBulkCopy.WriteToServerInternalAsync(CancellationToken ctoken) in /mnt/vss/_work/1/s/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs:line 3056
     at Microsoft.Data.SqlClient.SqlBulkCopy.WriteRowSourceToServerAsync(Int32 columnCount, CancellationToken ctoken) in /mnt/vss/_work/1/s/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs:line 1974
     at Microsoft.Data.SqlClient.SqlBulkCopy.WriteToServer(DbDataReader reader) in /mnt/vss/_work/1/s/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs:line 1648
     at Microsoft.Data.SqlClient.ManualTesting.Tests.AlwaysEncrypted.SqlBulkCopyTruncation.DoBulkCopy(String sourceTable, String targetTable, String connectionString) in /mnt/vss/_work/1/s/src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/SqlBulkCopyTruncation.cs:line 582
     at Microsoft.Data.SqlClient.ManualTesting.Tests.AlwaysEncrypted.SqlBulkCopyTruncation.<>c__DisplayClass12_0.<BulkCopyNVarchar>b__0() in /mnt/vss/_work/1/s/src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/SqlBulkCopyTruncation.cs:line 345
```

Wraith2:
Sure, but I can't repro so I'm looking at complex guesswork.

The only changed method is `TryReadChars`. What I changed it to do was work out how many chars it could copy completely and do them in a block. If it then has a partial char it falls back to the usual method of calling the existing `TryReadChar` because that can initiate a packet read. 

In the callstack above we can see that it's copied all it can as a block (if any) and is now calling into `TryReadChar` to handle the partial character, so it's likely we've got 1 byte of a char. That makes sense because `TryReadChar` then ends up calling into `TryReadByteArray` and it will only do that if there aren't enough bytes left in `_inBuff`.

So the error isn't in new code. It's in existing code. It's just passed through new code. Why didn't it fail before? no idea. I have a suspicion but no evidence. I'll add some debugging exception information and see if it fails with interesting values.

Wraith2:
Count someone from the MS side check the CI machines? the enclave steps are either timing out or just plain not working.

Wraith2:
Fixed, quick, merge while it's green!

David-Engel (approved):
Lets get this in sooner than later so it gets some miles on it during preview.

JRahnama (approved):

Wraith2:
> Lets get this in sooner than later so it gets some miles on it during preview.

Good idea. @panoskj do you have any benches you want to try with it since you found the original issue?

PR #1547, opened 2022-03-15
Improve error message when adding wrong type to SqlParameterCollection

### LINKED ISSUE #1546: SqlParameterCollection.Add() gives confusing error message when mixing drivers
### Describe the bug

Trying to add a `System.Data.SqlClient.SqlParameter` to a `Microsoft.Data.SqlClient.SqlParameterCollection` gives this error message:

```
The SqlParameterCollection only accepts non-null SqlParameter type objects, not SqlParameter objects.
```

This mistake can come up when migrating to the new driver or when using the new driver for ADO.NET Code and the old driver when working with EF6. If you're aware of the situation and history of the two drivers it is easy to intuit what this message means, but for the uninitiated this can be pretty confusing.

### To reproduce

```c#
var command = new Microsoft.Data.SqlClient.SqlCommand();
command.Parameters.Add(new System.Data.SqlClient.SqlParameter());
```

### Expected behavior

A message like the following would be more clear:

```
The SqlParameterCollection only accepts non-null Microsoft.Data.SqlClient.SqlParameter type objects, not System.Data.SqlClient.SqlParameter objects.
```

This seems trivial to fix; just changing https://github.com/dotnet/SqlClient/blob/865ac03ea9e0d362b38429348b8694ec960ae3a2/src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/AdapterUtil.cs#L787 to use `Type.ToString()/Type.FullName` instead of `Type.Name` for the expected and actual types. **I'd be happy to submit a PR if there is interest**.

### Further technical details
Microsoft.Data.SqlClient version: 4.1.0
.NET target: net5.0
SQL Server version: n/a
Operating system: Windows 10

ErikEJ:
(to help conversions from System.Data.SqlClient)

fixes #1546

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/AdapterUtil.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/AdapterUtil.cs
@@ -784,7 +784,7 @@ namespace Microsoft.Data.Common
             => IndexOutOfRange(StringsHelper.GetString(Strings.ADP_CollectionIndexString, itemType.Name, propertyName, propertyValue, collection.Name));
 
         internal static InvalidCastException CollectionInvalidType(Type collection, Type itemType, object invalidValue)
-            => InvalidCast(StringsHelper.GetString(Strings.ADP_CollectionInvalidType, collection.Name, itemType.Name, invalidValue.GetType().Name));
+            => InvalidCast(StringsHelper.GetString(Strings.ADP_CollectionInvalidType, collection.Name, itemType.FullName, invalidValue.GetType().FullName));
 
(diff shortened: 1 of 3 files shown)


DavoudEshtehari:
Thanks @ErikEJ
Could you address the `ParametersTest.CodeCoverageSqlClient` failure too?

Wraith2:
It doesn't, you're just getting columbo'ed ("just one more thing...") since code coverage is poor in places and you're editing the file.

ErikEJ:
Happy to do a separate PR to increase code coverage. 

Is there an issue for that?

johnnypham (approved):

PR #1551, opened 2022-03-18
Change ConditionalTheory for TestKeysFromCertificatesCreatedWithMultipleCryptoProviders

johnnypham:
The test does not require TCP and NP connection strings. These connection strings aren't provided in the enclave pipelines, so this test was not being run.


johnnypham:
makecert.exe doesn't seem to be accessible on 1es agents. The test may have to be rewritten.

DavoudEshtehari (approved):

Kaur-Parminder (approved):

Issue #988, opened 2021-03-16
Null Reference Exception


msallin:
We discovered the following `NullReferenceException`. 
The command which is executed is executed every hour and deletes some records.
The code runs already for about a year exactly like this and we only saw this exception once.

One interesting detail: 
The incoming HTTP request is logged at 8:15:07.000 AM and the exception at 8:17:10.049 AM

```
Exception message: System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.Data.SqlClient.SNI.TdsParserStateObjectManaged.ReleasePacket(PacketHandle syncReadPacket)
   at Microsoft.Data.SqlClient.TdsParserStateObject.ReadSniSyncOverAsync()
   at Microsoft.Data.SqlClient.TdsParserStateObject.TryReadNetworkPacket()
   at Microsoft.Data.SqlClient.TdsParserStateObject.TryPrepareBuffer()
   at Microsoft.Data.SqlClient.TdsParserStateObject.TryReadByte(Byte& value)
   at Microsoft.Data.SqlClient.TdsParser.TryRun(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj, Boolean& dataReady)
   at Microsoft.Data.SqlClient.TdsParser.Run(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj)
   at Microsoft.Data.SqlClient.TdsParser.ProcessAttention(TdsParserStateObject stateObj)
   at Microsoft.Data.SqlClient.TdsParserStateObject.ResetCancelAndProcessAttention()
   at Microsoft.Data.SqlClient.TdsParserStateObject.CloseSession()
   at Microsoft.Data.SqlClient.SqlCommand.EndExecuteNonQueryAsync(IAsyncResult asyncResult)
   at System.Threading.Tasks.TaskFactory`1.FromAsyncCoreLogic(IAsyncResult iar, Func`2 endFunction, Action`1 endAction, Task`1 promise, Boolean requiresSynchronization)
--- End of stack trace from previous location ---
   at Microsoft.EntityFrameworkCore.Storage.RelationalCommand.ExecuteNonQueryAsync(RelationalCommandParameterObject parameterObject, CancellationToken cancellationToken)
[...]

PR #1555, opened 2022-03-20
Rework `TdsParserStateObjectManaged` with nullable annotations

Wraith2:
I was contacted by someone in the c# discord server who was experiencing https://github.com/dotnet/SqlClient/issues/988 with version 3.0.1 of this library. If the previous fix was correct that shouldn't be possible so I had a look at the code in `TdsParserStateObjectManaged.ReleasePacket`:

```csharp
internal override void ReleasePacket(PacketHandle syncReadPacket)
{
    SNIPacket packet = syncReadPacket.ManagedPacket;
    if (packet != null)
    {
        SNIHandle handle = Handle;
        handle.ReturnPacket(packet);
    }
}
```

There are only two variables involved and the only one we haven't checked for null is `handle` so it's pretty obvious what's going on. Looking at this it's clear that c#8 nullable reference type annotations would have shown this as a problem so I thought I'd rewrite the file with nullable enabled to see if it caused any interesting bugs or changes to show up. This PR is the result.

Nullable annotations are a compile time feature and do not add any runtime checking. It is possible to pass nulls to functions defined in nullable enabled files even if those functions are decorated with non-null context. That means that these changes only introduce new exceptions where an explicit throw is added, there are no hidden null checks.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObjectManaged.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObjectManaged.cs
@@ -14,7 +14,7 @@ using Microsoft.Data.Common;
 
 namespace Microsoft.Data.SqlClient.SNI
 {
-    internal class TdsParserStateObjectManaged : TdsParserStateObject
+    internal sealed class TdsParserStateObjectManaged : TdsParserStateObject
     {
         private SNIMarsConnection? _marsConnection;
         private SNIHandle? _sessionHandle;
@@ -46,7 +46,7 @@ namespace Microsoft.Data.SqlClient.SNI
             }
             else
             {
-                throw ADP.InvalidOperation("physicalConnection type is incorrect");
+                throw new ArgumentException(StringsHelper.GetString(StringsHelper.SNI_IncorrectPhysicalConnectionType));
             }
         }
 
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObjectManaged.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObjectManaged.cs
@@ -2,8 +2,13 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+#nullable enable
+
 using System;
 using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
+using System.Runtime.CompilerServices;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Data.Common;
 
@@ -11,17 +16,15 @@ namespace Microsoft.Data.SqlClient.SNI
 {
     internal class TdsParserStateObjectManaged : TdsParserStateObject
     {
-        private SNIMarsConnection _marsConnection;
-        private SNIHandle _sessionHandle;
-        private SspiClientContextStatus _sspiClientContextStatus;
+        private SNIMarsConnection? _marsConnection;
+        private SNIHandle? _sessionHandle;
+        private SspiClientContextStatus? _sspiClientContextStatus;
 
         public TdsParserStateObjectManaged(TdsParser parser) : base(parser) { }
 
         internal TdsParserStateObjectManaged(TdsParser parser, TdsParserStateObject physicalConnection, bool async) :
             base(parser, physicalConnection, async)
         { }
-
-        internal SNIHandle Handle => _sessionHandle;
 
         internal override uint Status => _sessionHandle != null ? _sessionHandle.Status : TdsEnums.SNI_UNINITIALIZED;
 
@@ -36,15 +39,25 @@ namespace Microsoft.Data.SqlClient.SNI
         protected override void CreateSessionHandle(TdsParserStateObject physicalConnection, bool async)
         {
             Debug.Assert(physicalConnection is TdsParserStateObjectManaged, "Expected a stateObject of type " + GetType());
-            TdsParserStateObjectManaged managedSNIObject = physicalConnection as TdsParserStateObjectManaged;
-            _sessionHandle = managedSNIObject.CreateMarsSession(this, async);
-            SqlClientEventSource.Log.TryTraceEvent("TdsParserStateObjectManaged.CreateSessionHandle | Info | State Object Id {0}, Session Id {1}", _objectID, _sessionHandle?.ConnectionId);
+            if (physicalConnection is TdsParserStateObjectManaged managedSNIObject)
+            {
+                _sessionHandle = managedSNIObject.CreateMarsSession(this, async);
+                SqlClientEventSource.Log.TryTraceEvent("TdsParserStateObjectManaged.CreateSessionHandle | Info | State Object Id {0}, Session Id {1}", _objectID, _sessionHandle?.ConnectionId);
+            }
+            else
+            {
+                throw ADP.InvalidOperation("physicalConnection type is incorrect");
+            }
         }
 
         internal SNIMarsHandle CreateMarsSession(object callbackObject, bool async)
         {
             SqlClientEventSource.Log.TryTraceEvent("TdsParserStateObjectManaged.CreateMarsSession | Info | State Object Id {0}, Session Id {1}, Async = {2}", _objectID, _sessionHandle?.ConnectionId, async);
-            return _marsConnection.CreateMarsSession(callbackObject, async);
+            if (_marsConnection is null)
+            {
+                ThrowClosedConnection();
+            }
+            return _marsConnection!.CreateMarsSession(callbackObject, async);
         }
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObjectManaged.cs
```diff
@@ -36,15 +39,25 @@ protected override bool CheckPacket(PacketHandle packet, TaskCompletionSource<ob
...
 
         internal SNIMarsHandle CreateMarsSession(object callbackObject, bool async)
         {
             SqlClientEventSource.Log.TryTraceEvent("TdsParserStateObjectManaged.CreateMarsSession | Info | State Object Id {0}, Session Id {1}, Async = {2}", _objectID, _sessionHandle?.ConnectionId, async);
-            return _marsConnection.CreateMarsSession(callbackObject, async);
+            if (_marsConnection is null)
+            {
+                ThrowClosedConnection();
+            }
+            return _marsConnection!.CreateMarsSession(callbackObject, async);
```
Wraith2, line 60: This non-null assertion suffix is required only for netstandard2.0 builds, on netcore 2.1 and later correctly understand the control flow and understand that the variable is not null.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObjectManaged.cs
```diff
@@ -347,8 +382,26 @@ internal override uint GenerateSspiClientContext(byte[] receivedBuff, uint recei
...
         }
+
+        private SNIHandle GetSessionSNIHandleHandleOrThrow()
+        {
+            SNIHandle? sessionHandle = _sessionHandle;
+            if (sessionHandle is null)
+            {
+                ThrowClosedConnection();
+            }
+            return sessionHandle!;
```
Wraith2, line 396: This non-null assertion is only needed on netstandard2.0, netcore 2.1 and later understand the flow correctly.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObjectManaged.cs
```diff
@@ -347,8 +382,26 @@ internal override uint GenerateSspiClientContext(byte[] receivedBuff, uint recei
...
+        }
+
+#if NETCOREAPP
+        [DoesNotReturn]
+#endif
```
Wraith2, line 401: We could remove this ifdef if we added a nuget reference which implements the reference, there is one available. I'm not sure if it is a good idea to do this because the nullable attributes are a compile time feature so passing the dependency on to users doesn't seem useful to me.

roji:
@Wraith2 it is pretty common practice to [turn on NRTs only for recent TFMs](https://github.com/npgsql/npgsql/blob/main/Directory.Build.props#L25), to avoid the issues above, e.g. since the netstandard2.0 isn't yet properly annotated. For nullability-related attributes which were only introduced later (e.g. [DoesNotReturn]), you can have those attributes defined in the project once behind #if, so that you don't need to do conditional compilation every time you want to use them ([here's the Npgsql example](https://github.com/npgsql/npgsql/blob/main/src/Shared/CodeAnalysis.cs)).

This should land you in a place where you can use NRTs freely and without hacks or code smells, while getting almost perfect coverage through the new TFMs. There's always the chance of unverified code that's conditionally compiled only in older TFMs, but that's generally very rare, and code quality is usually more important.

Wraith2:
> you can have those attributes defined in the project once behind #if, so that you don't need to do conditional compilation every time you want to use them ([here's the Npgsql example](https://github.com/npgsql/npgsql/blob/main/src/Shared/CodeAnalysis.cs)).

I considered this but my concern is what happens if multiple assemblies do the same thing? Say for example npgsql and sqlclient get loaded into the same appdomain/alc, doesn't it cause a crash because it's multiply defined?

roji:
>> you can have those attributes defined in the project once behind #if, so that you don't need to do conditional compilation every time you want to use them ([here's the Npgsql example](https://github.com/npgsql/npgsql/blob/main/src/Shared/CodeAnalysis.cs)).

> 
> I considered this but my concern is what happens if multiple assemblies do the same thing? Say for example npgsql and sqlclient get loaded into the same appdomain/alc, doesn't it cause a crash because it's multiply defined?

Having multiple types in the same namespace isn't an issue, provided they live in different assemblies. .NET types are fully-qualified by their assembly name. These would just be internal types which never get used or exposed externally.

In fact, the NRT NullableAttribute (which is what gets generated based on NRT annotations) is synthesized by Roslyn into assemblies it builds (there's no NullableAttribute in the BCL), so there's already lots of copies of that type all around.

Wraith2:
Ok, that works for me. I've integrated the definitions into the shared codebase so when we get to merging the managed implementation into netfx we'll be able to use the same file. This removes all the postfix non-null assertion decorations which is pleasing.

David-Engel (approved):
LGTM

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObjectManaged.cs
```diff
@@ -36,14 +39,24 @@ protected override bool CheckPacket(PacketHandle packet, TaskCompletionSource<ob
...
-            _sessionHandle = managedSNIObject.CreateMarsSession(this, async);
-            SqlClientEventSource.Log.TryTraceEvent("TdsParserStateObjectManaged.CreateSessionHandle | Info | State Object Id {0}, Session Id {1}", _objectID, _sessionHandle?.ConnectionId);
+            if (physicalConnection is TdsParserStateObjectManaged managedSNIObject)
+            {
+                _sessionHandle = managedSNIObject.CreateMarsSession(this, async);
+                SqlClientEventSource.Log.TryTraceEvent("TdsParserStateObjectManaged.CreateSessionHandle | Info | State Object Id {0}, Session Id {1}", _objectID, _sessionHandle?.ConnectionId);
+            }
+            else
+            {
+                throw ADP.InvalidOperation("physicalConnection type is incorrect");
```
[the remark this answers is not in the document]
Wraith2: Done.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObjectManaged.cs
```diff
@@ -2,27 +2,30 @@
...
 using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
+using System.Runtime.CompilerServices;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Data.Common;
 
 namespace Microsoft.Data.SqlClient.SNI
 {
     internal class TdsParserStateObjectManaged : TdsParserStateObject
```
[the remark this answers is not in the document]
Wraith2: Done. Though it really doesn't make any difference because there is no way for a user to legally derive from an internal class. They'd have to use reflection trickery and if they're doing that they can get around sealing anyway.

src/Microsoft.Data.SqlClient/src/System/Diagnostics/CodeAnalysis.cs
```diff
@@ -0,0 +1,101 @@
...
+using System;
+using System.Diagnostics.CodeAnalysis;
+
+#pragma warning disable 1591
+
+namespace System.Diagnostics.CodeAnalysis
+{
+#if NETSTANDARD2_0
+    [AttributeUsageAttribute(AttributeTargets.Field | AttributeTargets.Parameter | AttributeTargets.Property)]
+    sealed class AllowNullAttribute : Attribute
```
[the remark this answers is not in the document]
Wraith2: Done throughout the file.

src/Microsoft.Data.SqlClient/src/System/Diagnostics/CodeAnalysis.cs
```diff
@@ -0,0 +1,101 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Diagnostics.CodeAnalysis;
+
+#pragma warning disable 1591
+
```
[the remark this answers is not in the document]
Wraith2: Removed.

src/Microsoft.Data.SqlClient/netcore/src/Resources/Strings.Designer.cs
```diff
@@ -1923,6 +1923,15 @@ internal class Strings {
             }
         }
         
+        /// <summary>
+        ///   Looks up a localized string similar to incorrect physicalConnection type.
```
David-Engel, line 1927:
```suggestion
        ///   Looks up a localized string similar to Incorrect physicalConnection type.
```

src/Microsoft.Data.SqlClient/netcore/src/Resources/Strings.resx
```diff
@@ -1932,6 +1932,9 @@
   <data name="SQL_ParameterDirectionInvalidForOptimizedBinding" xml:space="preserve">
     <value>Parameter '{0}' cannot have Direction Output or InputOutput when EnableOptimizedParameterBinding is enabled on the parent command.</value>
   </data>
+  <data name="SNI_IncorrectPhysicalConnectionType" xml:space="preserve">
+    <value>incorrect physicalConnection type</value>
```
David-Engel, line 1936:
```suggestion
    <value>Incorrect physicalConnection type.</value>
```

DavoudEshtehari (approved):

Issue #1402, opened 2021-11-23
Unable to connect to SQL Server with encryption enabled with 4.0


jmezach:
### Describe the bug
We've recently upgraded one of our applications from .NET 5 to 6 and also updated Microsoft.Data.SqlClient from 3.0.1 to 4.0.0. After this upgrade we are unable to connect to our SQL Server running on Windows from a Linux container running on Kubernetes based on the `dotnet/aspnet:6.0-alpine` image. Connecting to the SQL Server gives us the following exception:

```
Microsoft.Data.SqlClient.SqlException (0x80131904): A connection was successfully established with the server, but then an error occurred during the pre-login handshake. (provider: TCP Provider, error: 35 - An internal exception was caught)
 ---> System.Security.Authentication.AuthenticationException: The remote certificate was rejected by the provided RemoteCertificateValidationCallback.
   at System.Net.Security.SslStream.SendAuthResetSignal(ProtocolToken message, ExceptionDispatchInfo exception)
   at System.Net.Security.SslStream.CompleteHandshake(SslAuthenticationOptions sslAuthenticationOptions)
   at System.Net.Security.SslStream.ForceAuthenticationAsync[TIOAdapter](TIOAdapter adapter, Boolean receiveFirst, Byte[] reAuthenticationData, Boolean isApm)
   at System.Net.Security.SslStream.AuthenticateAsClient(SslClientAuthenticationOptions sslClientAuthenticationOptions)
   at System.Net.Security.SslStream.AuthenticateAsClient(String targetHost, X509CertificateCollection clientCertificates, SslProtocols enabledSslProtocols, Boolean checkCertificateRevocation)
   at Microsoft.Data.SqlClient.SNI.SNITCPHandle.EnableSsl(UInt32 options)
   at Microsoft.Data.ProviderBase.DbConnectionPool.TryGetConnection(DbConnection owningObject, UInt32 waitForMultipleObjectsTimeout, Boolean allowCreate, Boolean onlyOneCheckConnection, DbConnectionOptions userOptions, DbConnectionInternal& connection)
   at Microsoft.Data.ProviderBase.DbConnectionPool.TryGetConnection(DbConnection owningObject, TaskCompletionSource`1 retry, DbConnectionOptions userOptions, DbConnectionInternal& connection)
[...]

PR #1559, opened 2022-03-24
Fix | Skip the CRL check during authenticaiton

DavoudEshtehari:
Passing the `true` as the last parameter in [AuthenticateAsClient(String, X509CertificateCollection, SslProtocols, Boolean)](https://docs.microsoft.com/en-us/dotnet/api/system.net.security.sslstream.authenticateasclient?view=net-6.0#system-net-security-sslstream-authenticateasclient(system-string-system-security-cryptography-x509certificates-x509certificatecollection-system-security-authentication-sslprotocols-system-boolean)) method enables the Certificate Revocation List (CRL) check during authentication. If the CRL wasn't accessible from where the request was sent, it will fail to send the request to the SQL Server.

- related issue #1402

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -166,7 +166,3 @@ dotnet_code_quality.ca1802.api_surface = private, internal
 
 [*.cs]
 dotnet_code_quality.CA2100.excluded_type_names_with_derived_types = Microsoft.Data.SqlClient.ManualTesting.Tests.*
-[*.cs]
-dotnet_code_quality.CA2000.excluded_type_names_with_derived_types = Microsoft.Cci.* | Microsoft.Cci.Extensions.* | Microsoft.Cci.Writers.Syntax.*
-[*.cs]
-dotnet_code_quality.excluded_type_names_with_derived_types = Microsoft.Cci.* | Microsoft.Cci.Extensions.* | Microsoft.Cci.Writers.Syntax.*
--- a/.editorconfig
+++ b/.editorconfig
@@ -166,3 +166,7 @@ dotnet_code_quality.ca1802.api_surface = private, internal
 
 [*.cs]
 dotnet_code_quality.CA2100.excluded_type_names_with_derived_types = Microsoft.Data.SqlClient.ManualTesting.Tests.*
+[*.cs]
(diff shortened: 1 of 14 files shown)


David-Engel (approved):
I'd note in the description that passing `false` instead of `true` to that API changes the behavior of the call back to the way it was prior to the changes in #1168.

David-Engel (approved):

roji:
It may be worth adding a connection string parameter to allow opting into the CRL check, for users which do want it (CRLs can be a very important security thing).

johnnypham (approved):

Kaur-Parminder (approved):

David-Engel:
> It may be worth adding a connection string parameter to allow opting into the CRL check, for users which do want it (CRLs can be a very important security thing).

@roji 
I don't think that is necessary. The previous method we were calling, [AuthenticateAsClient(string targetHost)](https://source.dot.net/#System.Net.Security/System/Net/Security/SslStream.cs,387), just called the new method with a `false` parameter:
[AuthenticateAsClient(targetHost, null, SecurityProtocol.SystemDefaultSecurityProtocols, false);](https://source.dot.net/#System.Net.Security/System/Net/Security/SslStream.cs,387)
So by changing false to true in the previous PR, I believe we are forcing an immediate CRL check on every connection (bad for performance). I also assume, that by passing false, as the previous method was, we are only suppressing the CRL check for the immediate connection. The underlying platform can still update the CRL list when it deems it is necessary/appropriate.

roji:
> So by changing false to true in the previous PR, I believe we are forcing an immediate CRL check on every connection (bad for performance).

A CRL check on every (physical) connection can be critical for security, otherwise an application may accept certificates that have been revoked. In the common case, I don't expect the performance of physical connection opening to be critical - that's why pooling is there, and SSL/TLS already typically involved multiple round trips between the client and server.

Note that I'm not advocating CRL necessarily be enabled by default - just to allow users to opt into it if they so wish, for added security (FWIW this is what Npgsql does, [see docs](https://www.npgsql.org/doc/connection-string-parameters.html#security-and-encryption). After all, SslStream does expose this bool flag, so it's just a matter of allowing users to determine what SqlClient passes into it. If that makes sense, I can open a new issue for that.

PR #1588, opened 2022-04-18
Add support for aliases on netcore+Windows

David-Engel:
This brings support for aliases on netcore+Windows. It also merges TdsParserStaticMethods.cs into common code.

Functionally, this should bring alias support to netcore and netstandard when running on Windows.


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionString.cs
```diff
@@ -1102,20 +1102,6 @@ internal PoolBlockingPeriod ConvertValueToPoolBlockingPeriod()
             }
         }
 
-#if NETFRAMEWORK
```
David-Engel, line 1105: This diff is a bit misleading. This "deleted" section simply moved down and the NetLib code moved up so that the NetLib code is available on netcore.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStaticMethods.cs
```diff
@@ -103,62 +111,92 @@ internal static byte[] ObfuscatePassword(string password)
...
+                bLo = (byte)(password[i] & 0x0f);
+                bHi = (byte)(password[i] & 0xf0);
+                password[i] = (byte)(((bHi >> 4) | (bLo << 4)) ^ 0xa5);
+            }
+            return password;
         }
 
+        private const int NoProcessId = -1;
+        private static int s_currentProcessId = NoProcessId;
+        internal static int GetCurrentProcessIdForTdsLoginOnly()
```
David-Engel, line 136: I chose the netcore implementation of GetCurrentProcessIdForTdsLoginOnly over the netfx one that relied on the deprecated method. I don't believe it will cause any noticeable behavior differences on netfx.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStaticMethods.cs
```diff
@@ -103,62 +111,92 @@ internal static byte[] ObfuscatePassword(string password)
...
+            }
+            return s_currentProcessId;
+        }
 
-        [SecurityPermission(SecurityAction.Assert, Flags = SecurityPermissionFlag.UnmanagedCode)]
-        [ResourceExposure(ResourceScope.None)] // SxS: we use this method for TDS login only
-        [ResourceConsumption(ResourceScope.Process, ResourceScope.Process)]
-        static internal Int32 GetCurrentThreadIdForTdsLoginOnly()
+
+        internal static int GetCurrentThreadIdForTdsLoginOnly()
```
David-Engel, line 153: I chose the netcore implementation of GetCurrentThreadIdForTdsLoginOnly over the netfx one that relied on the deprecated method. I don't believe it will cause any noticeable behavior differences on netfx.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStaticMethods.cs
```diff
@@ -103,62 +111,92 @@ internal static byte[] ObfuscatePassword(string password)
...
-            const string key = "NetworkAddress";
-            const string localKey = "NetworkAddressLocal";
-            const string folder = "SOFTWARE\\Description\\Microsoft\\Rpc\\UuidTemporaryData";
-
-            int result = 0;
-            byte[] nicAddress = null;
-
-            object temp = ADP.LocalMachineRegistryValue(folder, localKey);
-            if (temp is int)
+            if (s_nicAddress != null)
```
David-Engel, line 164: I chose a merger of the netfx and netcore implementations of GetNetworkPhysicalAddressForTdsLoginOnly. Netcore had the optimization of caching s_nicAddress but did not have the registry lookup when running on Windows. Since this introduces registry lookups when on Windows, we can bring that functionality to netcore since users generally expect behavior parity when migrating from netfx to netcore.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
```diff
@@ -828,6 +833,40 @@ public static string RetrieveValueFromConnStr(string connStr, string[] keywords)
             return res;
         }
 
+        public static bool ParseDataSource(string dataSource, out string hostname, out int port, out string instanceName)
```
David-Engel, line 836: I put this in DataTestUtility because this is the same code as in the SSRP PR. If this is merged, the SSRP test can be refactored to reference this shared code instead of using its own.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectivityTests/ConnectivityTest.cs
```diff
@@ -368,5 +371,76 @@ public static void ConnectionOpenDisableRetry()
...
+        {
+            if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                throw new Exception("Alias test only valid on Windows");
+            }
+
+            if (!CanCreateAliases())
+            {
+                throw new Exception("Unable to create aliases in this environment. Windows + Admin + non-instance data source required.");
+            }
```
DavoudEshtehari, line 422: I'm wondering how the conditional fact lets the test run when I don't have the registry key! And the test throws the "Unable to create aliases in this environment. Windows + Admin + non-instance data source required." exception, whilst it's passed one time.

DavoudEshtehari (commented):
All the test runs were skipped because the `CanCreateAliases` didn't return `true`.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
David-Engel: Doing it in both at the same time will likely result in code conflicts. After one of these PRs is merged, the other can be updated.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectivityTests/ConnectivityTest.cs
```diff
@@ -368,5 +371,76 @@ public static void ConnectionOpenDisableRetry()
             duration = timer.Elapsed;
             Assert.True(duration.Seconds > 5, $"Connection Open() with retries took less time than expected. Expect > 5 sec with transient fault handling. Took {duration.Seconds} sec.");                //    sqlConnection.Open();
         }
+
+        private const string ConnectToPath = "SOFTWARE\\Microsoft\\MSSQLServer\\Client\\ConnectTo";
+        private static bool CanCreateAliases()
+        {
+            if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ||
+                !DataTestUtility.IsTCPConnStringSetup())
```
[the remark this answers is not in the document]
David-Engel: Nope. Aliases can be used for those endpoints, too.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectivityTests/ConnectivityTest.cs
```diff
@@ -368,5 +371,76 @@ public static void ConnectionOpenDisableRetry()
...
+            }
+
+            using (WindowsIdentity identity = WindowsIdentity.GetCurrent())
+            {
+                WindowsPrincipal principal = new(identity);
+                if (!principal.IsInRole(WindowsBuiltInRole.Administrator))
+                {
+                    return false;
+                }
+            }
```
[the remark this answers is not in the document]
David-Engel: Your user has admin privileges, but did you actually "Run as administrator"? (UAC prompt and everything?) When UAC is enabled on Windows, that block will return false for a user unless you "Run as administrator". I went through a few iterations of code, including your suggestion, before landing on what is here.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectivityTests/ConnectivityTest.cs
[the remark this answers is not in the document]
David-Engel: I found the ConditionalFact to be unreliable in all situations so I added this as a precaution. For example, running locally (as administrator) versus in the pipeline produced different behaviors on different configurations.

David-Engel: Yes. This is what I meant by the ConditionalFact being inconsistent across different environments. I remember I spent quite some time trying to figure out why it was not working the way it should and landed on this solution.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
David-Engel: Done

DavoudEshtehari (approved):

DavoudEshtehari (approved):

JRahnama (approved):

PR #1590, opened 2022-04-22
Test | Replacing Binary formatter with DataContractSerializer 

Kaur-Parminder:
Addressing build error (ReferrenceType = NetStandard) 'BinaryFormatter.Serialize(Stream, object)' is obsolete: 'BinaryFormatter serialization is obsolete and should not be used.


johnnypham:
Is `DataContractSerializer` being used properly? `DataContractAttribute` and `DataMemberAttribute` attributes have to be applied to classes and members that should be serialized but there aren't any of these attributes in `SqlError`.

Kaur-Parminder:
> Is `DataContractSerializer` being used properly? `DataContractAttribute` and `DataMemberAttribute` attributes have to be applied to classes and members that should be serialized but there aren't any of these attributes in `SqlError`.

Test is using only `primitive types` for validation in SqlError class.

lcheunglci (approved):

cheenamalhotra (approved):

PR #1593, opened 2022-04-26
Added Properties ReferenceType and TargetNetStandardVersion to RunTest command

Kaur-Parminder:
Added Properties ReferenceType and TargetNetStandardVersion to RunTest command to support NetStandard Test build for piprlines


Kaur-Parminder:
added the runtime parameters to RunTest command.

Issue #1591, opened 2022-04-25
Retry not working with SqlDataAdapter.Fill(table) (.net framework)


MichaelAzzer:
When use retry with command and execute command with DataReader
like 
```
 using (SqlDataReader reader = command.ExecuteReader())
                    {
                        table.Load(reader);
                        
                    }
```
it's working
but if I used SqlDataAdapter to fill table it's not working

JRahnama:
@MichaelAzzer we will look at it and will get back to you.

lcheunglci:
Hi, is there a sample project that I can use to reproduce this?

MichaelAzzer:
> Hi, is there a sample project that I can use to reproduce this?

Yes, I prepared a simple sample to produce the problem

this zipped file contains WinForms sample with database script.sql
[sample.zip](https://github.com/dotnet/SqlClient/files/8572773/sample.zip)

lcheunglci:
Thanks @MichaelAzzer , I was able to reproduce it locally with your sample project.  I'll take a look and debug it.

MichaelAzzer:
> Thanks @MichaelAzzer , I was able to reproduce it locally with your sample project. I'll take a look and debug it.
It's nice to hear that, thank you very much for your time

lcheunglci:
Hi @MichaelAzzer , I noticed that the exception being thrown is a timeout exception:
```
Microsoft.Data.SqlClient.SqlException: 'Execution Timeout Expired.  The timeout period elapsed prior to completion of the operation or the server is not responding.'
```
I change the `command.CommandTimeout = 1;` to `command.CommandTimeout = 5;` and it worked.  I also tried the "working" version with `SqlDataReader` and it also threw the same timeout exception until I changed the command timeout.

Can you try changing the command timeout to a longer duration and try again and let me know if it works?

MichaelAzzer:
> Hi @MichaelAzzer , I noticed that the exception being thrown is a timeout exception:
> 
> ```
> Microsoft.Data.SqlClient.SqlException: 'Execution Timeout Expired.  The timeout period elapsed prior to completion of the operation or the server is not responding.'
> ```
> 
> I change the `command.CommandTimeout = 1;` to `command.CommandTimeout = 5;` and it worked. I also tried the "working" version with `SqlDataReader` and it also threw the same timeout exception until I changed the command timeout.
> 
> Can you try changing the command timeout to a longer duration and try again and let me know if it works?

Yes, you're right 
[...]

PR #1599, opened 2022-04-29
Docs - update SqlDataAdapter docs re: RetryLogicProvider not being supported

lcheunglci:
This is for issue #1591 where `RetryLogicProvider` from the `SQLCommand` is not supported for the `SqlDataAdapter.Fill` because `System.Data.Common.DbDataAdapter` does not allow it.  Therefore, we could mention in documentation the `RetryLogicProvider` is not supported.


lcheunglci:
As discussed in #1591, this is no longer needed.

PR #1603, opened 2022-05-03
Merge to shared - TdsParserStaticMethods

lcheunglci:
Relates to #1261. Merged netfx into netcore for TdsParserStaticMethods, and moving it to shared src and updating file to conform with coding style.


Wraith2:
I'm not a big fan of using #ifdef's in the main files. I've not checked if there's a better way to do this but if there is it'd be better to have .netfx.cs and .netcore.cs files with parts of the partial class in them.

DavoudEshtehari:
Closing this PR since #1588 addresses the same merging.

PR #1604, opened 2022-05-04
Merge to shared - TdsParserSafeHandles

lcheunglci:
Relates to #1261. Merged netfx into netcore for TdsParserSafeHandles, and moving it to shared src and update the project references.


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserSafeHandles.cs
```diff
@@ -144,10 +161,17 @@ internal sealed class SNIHandle : SafeHandle
...
+			#if NETFRAMEWORK
+            TransparentNetworkResolutionState transparentNetworkResolutionState,
+            int totalTimeout,
+			#endif
             SqlConnectionIPAddressPreference ipPreference,
             SQLDNSInfo cachedDNSInfo)
             : base(IntPtr.Zero, true)
         {
+#if NETFRAMEWORK
+            RuntimeHelpers.PrepareConstrainedRegions();
```
Wraith2, line 173: I don't think this needs to be ifdeffed, or if it does we should make it so that it doesn't. Constrained region calls on netcore are nops.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserSafeHandles.cs
```diff
@@ -156,11 +180,21 @@ internal sealed class SNIHandle : SafeHandle
...
                 if (ignoreSniOpenTimeout)
                 {
+                    // UNDONE: ITEM12001110 (DB Mirroring Reconnect) Old behavior of not truly honoring timeout presevered 
+                    //  for non-failover scenarios to avoid breaking changes as part of a QFE.  Consider fixing timeout
+                    //  handling in next full release and removing ignoreSniOpenTimeout parameter.
                     timeout = Timeout.Infinite; // -1 == native SNIOPEN_TIMEOUT_VALUE / INFINITE
                 }
 
+#if NETFRAMEWORK
+                int transparentNetworkResolutionStateNo = (int)transparentNetworkResolutionState;
```
Wraith2, line 190: I don't know what the plans are for transparent network resolution. It may be a feature that should come to netcore eventually. It might be an idea to have a new compiler constant for it, @David-Engel is probably the person who would know.

David-Engel: The current sentiment around TNIR is that we can bring connection string and SqlConnectionStringBuilder parity to netcore but we should not actually implement the TNIR logic. Apparently it caused more support issues than it solved for Azure. Customers that really need the driver to initiate parallel connection attempts against each IP DNS resolves to should use MSF.

lcheunglci:
Thanks for the review @Wraith2 and I made the revisions. I noticed that there were some compiler errors on Linux because this file was meant for windows only and I placed it in the common ItemGroup.  I also noticed that some existing files have the .windows suffix, not sure if I should add the windows suffix to this to keep it consistent.

Wraith2:
If it's windows specific then please add the .windows.cs suffix.

DavoudEshtehari (approved):

David-Engel (approved):

PR #1606, opened 2022-05-06
Fix | SqlDataAdapter with RetryLogicProvider

lcheunglci:
As mentioned in #1591 , due to a the way `SqlCommand` implements the `ExecuteReader()` with `new` keyword, it is not able to retrieve the `SqlDataReader` with the `RetryLogicProvider` when `Fill` is invoked from `DbDataAdapter`, so in order to allow the retry logic to work with `SqlDataAdapter.Fill`, we have to new all the 5 `public` `Fill` methods to resolve the `SqlDataReader` and use the `protected` `Fill` overload from `DataAdapter` in order for the `RetryLogicProvider` to work.  I'm still in the middle of adding unit/integrated tests for this as there's no code coverage as the sample provide requires modification to the stored procedures to with one that provides a delay.


David-Engel (changes requested):
For the docs, we need to add a doc section for each new public API to SqlDataAdapter.xml.

For each new public API, we also need to add them to the ref files for netcore and netfx.
src\Microsoft.Data.SqlClient\netcore\ref\Microsoft.Data.SqlClient.cs
src\Microsoft.Data.SqlClient\netfx\ref\Microsoft.Data.SqlClient.cs

lcheunglci:
Thanks for the suggestions, I'm still looking through the `SqlDataAdapter.xml` and trying to figure out how to add the documentation to it.  I was hoping I could be able to reference an existing method such as the `constructor` or `AddToBatch`, but can't see how it generate the method details e.g.
```
        <AddToBatch>
            <param name="command">To be added.</param>
            <summary>To be added.</summary>
            <returns>To be added.</returns>
            <remarks>To be added.</remarks>
        </AddToBatch>
```
I tried to look for the documentation in [runtime ](https://github.com/dotnet/runtime)repo and the [sql docs](https://github.com/MicrosoftDocs/sql-docs) repo to see if there's any examples, but I can't seem to find any.  If you can direct me to an example that would be great.

DavoudEshtehari (changes requested):
The related issue only occurs using netfx. I'd suggest investigating its difference with netcore.

lcheunglci:
> The related issue only occurs using netfx. I'd suggest investigating its difference with netcore.

You're right! It looks like it works without the changes in netcore. I'll need to investigate why the behavior diverged between netfx and netcore.

PR #1608, opened 2022-05-09
Feature | Adding TDS 8 support

JRahnama:
TDS8 support is in preview mode for SQL server 2022. This PR is to add client-side support.

Preview MS-TDS spec changes:
https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-winprotlp/8a9c667b-2825-46a8-8066-a80681233c33

Behavior:
To use TDS 8, users should specify Encrypt=Strict in the connection string. Strict mode disables TrustServerCertificate (always treated as False in Strict mode). HostNameInCertificate has been added to help some Strict mode scenarios. In the future, the plan is also to add a ServerCertificate setting, which points to a certificate file that will be directly compared to the server certificate for validation. This will allow secure use of any certificate (self-signed, for example) on a connection without opening up the connection to MITM attacks by blanket trusting of all certificates (as TrustServerCertificate=true would do).

The currently proposed code also retains API backwards compatibility in SqlConnectionStringBuilder.Encrypt. Encrypt has changed from a bool to a SqlConnectionEncryptOption object, but implicit conversion operators have been added so that it can still be treated like a bool. Comparing SqlConnectionEncryptOption to a bool will return true for SqlConnectionEncryptOption.Mandatory or SqlConnectionEncryptOption.Strict. Converting a bool to SqlConnectionEncryptOption will produce SqlConnectionEncryptOption.Mandatory for true and SqlConnectionEncryptOption.Optional for false.


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
```diff
@@ -249,8 +258,13 @@ public SNITCPHandle(string serverName, int port, long timerExpire, bool parallel
...
 
-                    _sslOverTdsStream = new SslOverTdsStream(_tcpStream, _connectionId);
-                    _sslStream = new SNISslStream(_sslOverTdsStream, true, new RemoteCertificateValidationCallback(ValidateServerCertificate));
+                    Stream stream = _tcpStream;
+                    if (!_isTDSS)
+                    {
+                        _sslOverTdsStream = new SslOverTdsStream(_tcpStream, _connectionId);
+                        stream = _sslOverTdsStream;
+                    }
+                    _sslStream = new SNISslStream(stream, true, new RemoteCertificateValidationCallback(ValidateServerCertificate));
```
Wraith2, line 267: Is `ValidateServerCertificate` an instance method that we need to create a delegate for each time we create the connection? If it's static or takes a context parameter we could avoid the allocation.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
```diff
@@ -581,10 +595,31 @@ public override uint EnableSsl(uint options)
...
-                    _sslStream.AuthenticateAsClient(_targetServer, null, SupportedProtocols, false);
-                    _sslOverTdsStream.FinishHandshake();
+                    if (_isTDSS)
+                    {
+#if NETCOREAPP 
+
+                        SslClientAuthenticationOptions sslClientOptions = new()
+                        {
+                            TargetHost = _serverIndicationName,
+                            ApplicationProtocols = new List<SslApplicationProtocol>() { new(TdsEnums.TDS8) },
```
Wraith2, line 608: Does this need to be a per instance collection? Is if going to be modified by the caller? if it isn't going to be modified we could have a single static instance that we reuse. If it has to be instantiated consider using the ctor that specifies the initial size because the default initial size is 5 and if we know this will almost never grow we waste some memory each time.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
```diff
@@ -631,9 +669,18 @@ private bool ValidateServerCertificate(object sender, X509Certificate cert, X509
...
                 return true;
             }
+            string serverNameToValidate;
+            if (_isTDSS && !string.IsNullOrEmpty(_hostNameInCertificate))
+            {
+                serverNameToValidate = _hostNameInCertificate;
+            }
+            else
+            {
+                serverNameToValidate = _hostNameInCertificate;
```
Wraith2, line 679: The if and else look identical

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -671,7 +673,7 @@ internal void ProcessPendingAck(TdsParserStateObject stateObj)
 
             // UNDONE - send "" for instance now, need to fix later
             SqlClientEventSource.Log.TryTraceEvent("<sc.TdsParser.Connect|SEC> Sending prelogin handshake");
-            SendPreLoginHandshake(instanceName, encrypt, !string.IsNullOrEmpty(certificate), useOriginalAddressInfo);
+            SendPreLoginHandshake(instanceName, encrypt, isTDS8!, string.IsNullOrEmpty(certificate), useOriginalAddressInfo);
```
Wraith2, line 676: Was the `!` supposed to stay with the `string.IsNullOrEmpty`? At the moment it's become an nullable annotation.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParserHelperClasses.cs
```diff
@@ -29,6 +29,7 @@ internal enum EncryptionOptions
         NOT_SUP,
         REQ,
         LOGIN,
+        Strict, //TDS8
```
Wraith2, line 32: Should this be all uppercase like the others?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionStringBuilder.cs
```diff
@@ -201,6 +207,8 @@ private static string[] CreateValidKeywords()
                 { DbConnectionStringKeywords.CurrentLanguage, Keywords.CurrentLanguage },
                 { DbConnectionStringKeywords.DataSource, Keywords.DataSource },
                 { DbConnectionStringKeywords.Encrypt, Keywords.Encrypt },
+                { DbConnectionStringKeywords.IsTDS8, Keywords.IsTDS8 },
```
Wraith2, line 210: This means that they keyword `IsTDS8` will be exposed to users in the connection string, correct? If so is that a decision that has been made and set in stone? I don't this most users of the library separate sql server and tds so the naming is non-obvious to most users. Is it possible this could be changed to something like ProtocolVersion?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNINpHandle.cs
```diff
@@ -312,8 +322,29 @@ public override uint EnableSsl(uint options)
                 _validateCert = (options & TdsEnums.SNI_SSL_VALIDATE_CERTIFICATE) != 0;
                 try
                 {
-                    _sslStream.AuthenticateAsClient(_targetServer, null, SupportedProtocols, false);
-                    _sslOverTdsStream.FinishHandshake();
+                    if (_isTDS8)
+                    {
+#if NETCOREAPP
+                        SslApplicationProtocol TDS8 = new("tds/8.0");
```
Wraith2, line 328: Nit: naming would be `tds8` since it's a local not a public member. I think I'd probably just call it `protocol` since tds is the value not the description of the variable that holds it.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -409,6 +414,13 @@ internal void ProcessPendingAck(TdsParserStateObject stateObj)
                 SqlClientEventSource.Log.TryTraceEvent("TdsParser.Connect | SEC | SSPI or Active Directory Authentication Library loaded for SQL Server based integrated authentication");
             }
 
+            // if Strict encryotion is chosen TDS8 should be used and trust server certificate should be false.
```
Wraith2, line 417: typo

Wraith2 (commented):
Looks ok in general. 
Have you considered using an int protocol version instead of a bool? It seems sensible to think there may be later versions and make switching on the version cleaner.

doc/snippets/Microsoft.Data.SqlClient/SqlConnectionEncryptionOption.xml
```diff
@@ -0,0 +1,30 @@
+﻿<?xml version="1.0"?>
+<docs>
+  <members name="SqlConnectionEncryptionOption">
+    <SqlConnectionEncryptionOption>
+      <summary>
+
+        Note that these settings cannot be used to bypass encryption and gain access to plaintext data. For details, see <see href="https://docs.microsoft.com/sql/relational-databases/security/encryption/always-encrypted-database-engine">Always Encrypted (Database Engine)</see>.
```
David-Engel, line 7:
```suggestion
        These options are used to control encryption behavior of the communication between the server and the client.</see>.
```

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -994,7 +1006,15 @@ public sealed partial class SqlConnectionStringBuilder : System.Data.Common.DbCo
         /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnectionStringBuilder.xml' path='docs/members[@name="SqlConnectionStringBuilder"]/Encrypt/*'/>
         [System.ComponentModel.DisplayNameAttribute("Encrypt")]
         [System.ComponentModel.RefreshPropertiesAttribute(System.ComponentModel.RefreshProperties.All)]
-        public bool Encrypt { get { throw null; } set { } }
+        public SqlConnectionEncryptionOption Encrypt { get { throw null; } set { } }
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnectionStringBuilder.xml' path='docs/members[@name="SqlConnectionStringBuilder"]/IsTDS8/*'/>
+        [System.ComponentModel.DisplayNameAttribute("IsTDS8")]
+        [System.ComponentModel.RefreshPropertiesAttribute(System.ComponentModel.RefreshProperties.All)]
+        public bool IsTDS8 { get { throw null; } set { } }
```
David-Engel, line 1013: Why are we exposing this new public API? It looks like this is something we'd only use internally.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNINpHandle.cs
David-Engel: I would define a static list of client-supported protocols somewhere above (`private static clientApplicationProtocols s_Protocols = new List<SslApplicationProtocol>() { new SslApplicationProtocol("tds/8.0") }`) that you can reference on line 333 and remove this local variable completely.

EDIT: I now see that Wraith already suggested the same thing in SNITcpHandle.cs. We can use the same static reference here and there.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
```diff
@@ -117,14 +120,20 @@ public override int ProtocolVersion
         /// <param name="parallel">Parallel executions</param>
         /// <param name="ipPreference">IP address preference</param>
         /// <param name="cachedFQDN">Key for DNS Cache</param>
-        /// <param name="pendingDNSInfo">Used for DNS Cache</param>        
-        public SNITCPHandle(string serverName, int port, long timerExpire, bool parallel, SqlConnectionIPAddressPreference ipPreference, string cachedFQDN, ref SQLDNSInfo pendingDNSInfo)
+        /// <param name="pendingDNSInfo">Used for DNS Cache</param>
+        /// <param name="isTDSS">Support TDS8.0</param>
```
David-Engel, line 124:
```suggestion
        /// <param name="isTDS8">Support TDS8.0</param>
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
```diff
@@ -117,14 +120,20 @@ public override int ProtocolVersion
...
+        /// <param name="hostNameInCertificate">Host Name in Certoficate</param>
+        /// <param name="serverNameIndication"></param>
+        public SNITCPHandle(string serverName, int port, long timerExpire, bool parallel, SqlConnectionIPAddressPreference ipPreference, string cachedFQDN, ref SQLDNSInfo pendingDNSInfo, bool isTDSS, string hostNameInCertificate, string serverNameIndication)
         {
             using (TrySNIEventScope.Create(nameof(SNITCPHandle)))
             {
                 SqlClientEventSource.Log.TrySNITraceEvent(nameof(SNITCPHandle), EventType.INFO, "Connection Id {0}, Setting server name = {1}", args0: _connectionId, args1: serverName);
 
                 _targetServer = serverName;
+                _isTDSS = isTDSS;
```
David-Engel, line 134:
```suggestion
                _isTDS8 = isTDS8;
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
```diff
@@ -581,10 +595,31 @@ public override uint EnableSsl(uint options)
             using (TrySNIEventScope.Create(nameof(SNIHandle)))
             {
                 _validateCert = (options & TdsEnums.SNI_SSL_VALIDATE_CERTIFICATE) != 0;
+
                 try
                 {
-                    _sslStream.AuthenticateAsClient(_targetServer, null, SupportedProtocols, false);
-                    _sslOverTdsStream.FinishHandshake();
+                    if (_isTDSS)
```
David-Engel, line 601:
```suggestion
                    if (_isTDS8)
```

David-Engel (commented):
Not done reviewing, but wanted to submit some partial feedback.

As we discussed offline, we should also try to maintain a backwards compatible connection string, where possible. Specifically, with the new encrypt enum and Optional/Mandatory values, when we output a connection string, we should continue to output `Encrypt=true` or `Encrypt=false` for Mondatory/Optional respectively. This will preserve some backwards compatibility between versions in scenarios where an application is passing connection strings around. This is more important since Encrypt is such a common/well-used option.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionStringBuilder.cs
Wraith2: @David-Engel it'd be good to get your opinion on this.

David-Engel: Ah, yes. I hadn't gotten down that far in my review. No, this should definitely be removed. I'm pretty sure it is a leftover artifact of the original POC. It should not be included in this implementation. I made a similar comment on the ref source file.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
JRahnama: can you check the new changes I made?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNINpHandle.cs
JRahnama: can you check the new changes I made?

Wraith2: in `SNIPhysicalHandle` create a new field `private static s_clientApplicationProtocols s_Protocols = new List<SslApplicationProtocol>(1) { new SslApplicationProtocol("tds/8.0") };` as David suggested and then in `SNINpHandle` and `SNITcpHandle` use it like this:
```
SslClientAuthenticationOptions sslClientOptions = new()
{
    TargetHost = _serverNameIndication,
    ApplicationProtocols = SNIPhysicalHandle.s_clientApplicationProtocols ,
    EnabledSslProtocols = SupportedProtocols,
    ClientCertificates = null,
};
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNINpHandle.cs
```diff
@@ -312,8 +322,29 @@ public override uint EnableSsl(uint options)
...
+
+                        SslClientAuthenticationOptions sslClientOptions = new()
+                        {
+                            TargetHost = _serverNameIndication,
+                            ApplicationProtocols = new List<SslApplicationProtocol>() { TDS8 },
+                            EnabledSslProtocols = SupportedProtocols,
+                            ClientCertificates = null,
+                        };
+                        _sslStream.AuthenticateAsClientAsync(sslClientOptions).Wait();
```
Wraith2: This is using an async function and then awaiting it synchronously, if @roji sees you doing this you'll make him sad. See if there is a naturally synchronous `AuthenticateAsClient` method you can use instead and avoid sync-over-async.

David-Engel (commented):
Another partial review. In general, where we were previously passing `bool encrypt` internally and we are now passing `SqlConnectionEncryptionOption encrypt` plus `bool isTDS8`, we could simplify the code by just passing `bool encrypt` plus `bool isTDS8` (or more accurately `bool tlsFirst`). This would make for fewer code changes related to the change of the encrypt parameter type.

Alternatively, just pass `SqlConnectionEncryptionOption encrypt` and base all the `if (isTDS8)` logic off of `if (encrypt == SqlConnectionEncryptionOption.Strict)`.

Since I haven't been all the way through the code, feel free to point to to why either of the above wouldn't work.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNINpHandle.cs
JRahnama: there is no AuthenticateAsClient that accepts SslClientAuthenticationOption as an argument in netcore 3.1. It started from net5 and after.

Wraith2: Netcore 3.1 goes out of support at the end of this year (it's LTS) and net5 is already unsupported. https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core

You can use compiler ifdefs to use the current pattern only on 3.1 and lower and then use the sync overload from net5 onwards. In this case I think it's acceptable to do that because it will avoid more threading problems.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
```diff
@@ -844,8 +844,23 @@ private SNINativeMethodWrapper.ConsumerInfo CreateConsumerInfo(bool async)
...
+            bool async,
+            bool fParallel,
+            TransparentNetworkResolutionState transparentNetworkResolutionState,
+            int totalTimeout,
+            SqlConnectionIPAddressPreference ipPreference,
+            string cachedFQDN,
+            bool tlsFirst = false,
+            string hostNameInCertificate = "",
+            string databaseName = "",
+            ApplicationIntent applicationIntent = ApplicationIntent.ReadWrite)
```
Wraith2, line 863: This is a lot of parameters. It might be worth just making a class to hold them.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -1149,10 +1155,122 @@ private void SendPreLoginHandshake(byte[] instanceName, SqlConnectionEncryptionO
...
+                bool valid = serverCallback(serverCert);
+                if (!valid)
+                {
+                    throw SQL.InvalidServerCertificate();
+                }
+            }
+
+            // create a new packet encryption changes the internal packet size Bug# 228403
+            try
+            { } // EmptyTry/Finally to avoid FXCop violation
```
Wraith2, line 1263: Which violation is flagged? The only reason i know of that try catches like this are used are in pre-prepared regions.

Wraith2 (approved):
Looks ok to me apart from the todos

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/DbConnectionStringCommon.cs
```diff
@@ -810,6 +867,65 @@ internal static SqlConnectionAttestationProtocol ConvertToAttestationProtocol(st
...
+                // try again after remove leading & trailing whitespaces.
+                sValue = sValue.Trim();
+                if (TryToConvertToSqlConnectionEncryptionOption(sValue, out SqlConnectionEncryptionOption result))
+                {
+                    return result;
+                }
+                // string values must be valid
+                throw ADP.InvalidConnectionOptionValue(keyword);
+            }
+            else
```
lcheunglci, line 887: I think we can remove the `else` here because the condition above is a guard clause since it either returns the result or throws an exception, and that's one less layer of indention for readability

lcheunglci, line 887: I think we can remove the `else` here because the condition above is a guard clause since it either returns the result or throws an exception, and that's one less layer of indention for readability

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -954,8 +965,28 @@ internal void BestEffortCleanup()
...
+            if (tlsFirst)
+            {
+                //Always validate the certificate when in strict encryption mode
+                uint info = TdsEnums.SNI_SSL_VALIDATE_CERTIFICATE | TdsEnums.SNI_SSL_USE_SCHANNEL_CACHE | TdsEnums.SNI_SSL_SEND_ALPN_EXTENSION;
+
+                EnableSsl(info, true, integratedSecurity, serverCallback, clientCallback);
+
+                // Since encryption has already been negotiated, we need to set encryption not supported in
+                // prelogin so that we don't try to negotiate encryption again during ConsumePreLoginHandshake.
+                _encryptionOption = EncryptionOptions.NOT_SUP;
```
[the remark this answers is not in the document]
David-Engel: I know it's a bit odd, hence the code comments. But the byte value of the enum is sent to the server in the payload. So I'm hesitant to add a value to it that doesn't correspond to anything the server knows about.

If you think about the encryption option as a TDS feature and we are now doing TDS _inside_ TLS, we don't want to support encryption again within TDS.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
David-Engel: I almost took this out. It was in the existing code (see deleted line 1375), but I suspect it no longer applies. I think I'll just go ahead and remove it.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/DbConnectionStringCommon.cs
David-Engel: Done.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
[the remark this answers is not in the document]
Wraith2: Ok, in that case the only way out would be to have two enums and then a mapping in and out which would probably be more confusing. So leave as is.

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -500,6 +501,33 @@ public enum SqlConnectionIPAddressPreference
         /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnectionIPAddressPreference.xml' path='docs/members[@name="SqlConnectionIPAddressPreference"]/UsePlatformDefault/*' />
         UsePlatformDefault = 2
     }
+    /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnectionEncryptOption.xml' path='docs/members[@name="SqlConnectionEncryptOption"]/SqlConnectionEncryptOption/*'/>
+    public struct SqlConnectionEncryptOption
+    {
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnectionEncryptOption.xml' path='docs/members[@name="SqlConnectionEncryptOption"]/Optional/*' />
+        public static SqlConnectionEncryptOption Optional { get { throw null; } }
```
DavoudEshtehari, line 508: NIT
```suggestion
        public static SqlConnectionEncryptOption Optional => throw null;
```

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -500,6 +501,33 @@ public enum SqlConnectionIPAddressPreference
...
+
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnectionEncryptOption.xml' path='docs/members[@name="SqlConnectionEncryptOption"]/Mandatory/*' />
+        public static SqlConnectionEncryptOption Mandatory { get { throw null; } }
```
DavoudEshtehari, line 511:
```suggestion
        public static SqlConnectionEncryptOption Mandatory => throw null;
```

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -500,6 +501,33 @@ public enum SqlConnectionIPAddressPreference
         /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnectionIPAddressPreference.xml' path='docs/members[@name="SqlConnectionIPAddressPreference"]/UsePlatformDefault/*' />
         UsePlatformDefault = 2
     }
+    /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnectionEncryptOption.xml' path='docs/members[@name="SqlConnectionEncryptOption"]/SqlConnectionEncryptOption/*'/>
+    public struct SqlConnectionEncryptOption
```
DavoudEshtehari, line 505:
```suggestion
    public class SqlConnectionEncryptOption
```

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -500,6 +501,33 @@ public enum SqlConnectionIPAddressPreference
...
+    {
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnectionEncryptOption.xml' path='docs/members[@name="SqlConnectionEncryptOption"]/Optional/*' />
+        public static SqlConnectionEncryptOption Optional { get { throw null; } }
+
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnectionEncryptOption.xml' path='docs/members[@name="SqlConnectionEncryptOption"]/Mandatory/*' />
+        public static SqlConnectionEncryptOption Mandatory { get { throw null; } }
+
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnectionEncryptOption.xml' path='docs/members[@name="SqlConnectionEncryptOption"]/Strict/*' />
+        public static SqlConnectionEncryptOption Strict { get { throw null; } }
```
DavoudEshtehari, line 514:
```suggestion
        public static SqlConnectionEncryptOption Strict => throw null;
```

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -500,6 +501,33 @@ public enum SqlConnectionIPAddressPreference
...
+
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnectionEncryptOption.xml' path='docs/members[@name="SqlConnectionEncryptOption"]/BoolToOption/*' />
+        public static implicit operator SqlConnectionEncryptOption(bool value) => value ? SqlConnectionEncryptOption.Mandatory : SqlConnectionEncryptOption.Optional;
```
DavoudEshtehari, line 517:
```suggestion
        public static implicit operator SqlConnectionEncryptOption(bool value) => throw null;
```

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -500,6 +501,33 @@ public enum SqlConnectionIPAddressPreference
...
+
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnectionEncryptOption.xml' path='docs/members[@name="SqlConnectionEncryptOption"]/OptionToBool/*' />
+        public static implicit operator bool(SqlConnectionEncryptOption value) => !Optional.Equals(value);
```
DavoudEshtehari, line 520:
```suggestion
        public static implicit operator bool(SqlConnectionEncryptOption value) => throw null;
```

src/Microsoft.Data.SqlClient/netfx/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -918,6 +918,34 @@ public enum SqlConnectionIPAddressPreference
         UsePlatformDefault = 2
     }
 
+    /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnectionEncryptOption.xml' path='docs/members[@name="SqlConnectionEncryptOption"]/SqlConnectionEncryptOption/*'/>
+    public struct SqlConnectionEncryptOption
```
DavoudEshtehari, line 922:
```suggestion
    public class SqlConnectionEncryptOption
```

src/Microsoft.Data.SqlClient/netfx/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -918,6 +918,34 @@ public enum SqlConnectionIPAddressPreference
...
+    {
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnectionEncryptOption.xml' path='docs/members[@name="SqlConnectionEncryptOption"]/Optional/*' />
+        public static SqlConnectionEncryptOption Optional { get { throw null; } }
```
DavoudEshtehari, line 925: NIT
```suggestion
        public static SqlConnectionEncryptOption Optional => throw null;
```

src/Microsoft.Data.SqlClient/netfx/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -918,6 +918,34 @@ public enum SqlConnectionIPAddressPreference
...
+
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnectionEncryptOption.xml' path='docs/members[@name="SqlConnectionEncryptOption"]/Mandatory/*' />
+        public static SqlConnectionEncryptOption Mandatory { get { throw null; } }
```
DavoudEshtehari, line 928:
```suggestion
        public static SqlConnectionEncryptOption Mandatory => throw null;
```

src/Microsoft.Data.SqlClient/netfx/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -918,6 +918,34 @@ public enum SqlConnectionIPAddressPreference
...
+
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnectionEncryptOption.xml' path='docs/members[@name="SqlConnectionEncryptOption"]/Strict/*' />
+        public static SqlConnectionEncryptOption Strict { get { throw null; } }
```
DavoudEshtehari, line 931:
```suggestion
        public static SqlConnectionEncryptOption Strict => throw null;
```

src/Microsoft.Data.SqlClient/netfx/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -918,6 +918,34 @@ public enum SqlConnectionIPAddressPreference
...
+
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnectionEncryptOption.xml' path='docs/members[@name="SqlConnectionEncryptOption"]/BoolToOption/*' />
+        public static implicit operator SqlConnectionEncryptOption(bool value) => value ? SqlConnectionEncryptOption.Mandatory : SqlConnectionEncryptOption.Optional;
```
DavoudEshtehari, line 934:
```suggestion
        public static implicit operator SqlConnectionEncryptOption(bool value) => throw null;
```

src/Microsoft.Data.SqlClient/netfx/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -918,6 +918,34 @@ public enum SqlConnectionIPAddressPreference
...
+
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnectionEncryptOption.xml' path='docs/members[@name="SqlConnectionEncryptOption"]/OptionToBool/*' />
+        public static implicit operator bool(SqlConnectionEncryptOption value) => !Optional.Equals(value);
```
DavoudEshtehari, line 937:
```suggestion
        public static implicit operator bool(SqlConnectionEncryptOption value) => throw null;
```

David-Engel (approved):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionString.cs
```diff
@@ -400,15 +406,15 @@ internal SqlConnectionString(string connectionString) : base(connectionString, G
...
+            if (_encrypt == SqlConnectionEncryptOption.Optional)
             {    // Support legacy registry encryption settings
                 const string folder = "Software\\Microsoft\\MSSQLServer\\Client\\SuperSocketNetLib";
                 const string value = "Encrypt";
 
                 object obj = ADP.LocalMachineRegistryValue(folder, value);
                 if ((obj is int iObj) && (iObj == 1))
                 {         // If the registry key exists
-                    _encrypt = true;
+                    _encrypt = SqlConnectionEncryptOption.Mandatory;
```
DavoudEshtehari, line 417: Shouldn't we support the `Strict` from the registry?

DavoudEshtehari (commented):
Generally, I believe there are some opportunities to add more event source logs, especially in the new paths that would be better if we could include them with this change.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParserSafeHandles.cs
```diff
@@ -152,9 +152,7 @@ internal sealed class SNIHandle : SafeHandle
             SqlConnectionIPAddressPreference ipPreference,
             SQLDNSInfo cachedDNSInfo,
             bool tlsFirst,
```
DavoudEshtehari, line 154: Did you forget to use the `tlsFirst` parameter or it's unnecessary?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionString.cs
David-Engel: I thought about it. I can't even find documentation for the way the setting currently behaves in netfx. I'm dis-inclined to extend it further since we will need something new in the future for system-level-forced-strict-encryption that is cross-plat.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -393,10 +394,10 @@ internal void ProcessPendingAck(TdsParserStateObject stateObj)
                     authType == SqlAuthenticationMethod.NotSpecified ? SqlAuthenticationMethod.SqlPassword.ToString() : authType.ToString());
             }
 
-            // Encryption is not supported on SQL Local DB - disable it for current session.
-            if (connHandler.ConnectionOptions.LocalDBInstance != null && encrypt)
+            // Encryption is not supported on SQL Local DB - disable it if they have only specified Mandatory
+            if (connHandler.ConnectionOptions.LocalDBInstance != null && encrypt == SqlConnectionEncryptOption.Mandatory)
```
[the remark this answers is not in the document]
David-Engel: I thought about this one. I believe connections to localdb should fail if encrypt is set to strict. The reason the logic is comparing only to Mandatory is that Mandatory is our new default and we don't want the new default to fail for all the existing localdb scenarios that don't specify an encrypt mode. If someone has explicitly set strict mode, we should fail against localdb.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -530,10 +533,10 @@ internal void ProcessPendingAck(TdsParserStateObject stateObj)
             if (connHandler.ConnectionOptions.LocalDBInstance != null)
             {
                 LocalDBAPI.CreateLocalDBInstance(connHandler.ConnectionOptions.LocalDBInstance);
-                if (encrypt)
+                if (encrypt == SqlConnectionEncryptOption.Mandatory)
```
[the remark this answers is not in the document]
David-Engel: Same comment as on the netcore version. We only want to automatically change to optional if encrypt=true/mandatory because it likely means they are using the new default value. If they have set encrypt=strict, they have explicitly set it and localdb connections will fail since they don't support encryption.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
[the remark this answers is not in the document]
DavoudEshtehari: Could we clarify this in documentation to avoid possible confusion in the future?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/DbConnectionStringCommon.cs
```diff
@@ -696,6 +696,38 @@ internal static SqlConnectionColumnEncryptionSetting ConvertToColumnEncryptionSe
...
+            if (StringComparer.InvariantCultureIgnoreCase.Equals(value, nameof(SqlConnectionEncryptOption.Optional))
+                || StringComparer.InvariantCultureIgnoreCase.Equals(value, "false")
+                || StringComparer.InvariantCultureIgnoreCase.Equals(value, "no"))
+            {
+                result = SqlConnectionEncryptOption.Optional;
+                return true;
+            }
+            else if (StringComparer.InvariantCultureIgnoreCase.Equals(value, nameof(SqlConnectionEncryptOption.Mandatory))
+                           || StringComparer.InvariantCultureIgnoreCase.Equals(value, "true")
+                           || StringComparer.InvariantCultureIgnoreCase.Equals(value, "yes"))
```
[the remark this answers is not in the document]
David-Engel: I removed the unused method completely.

DavoudEshtehari (approved):

David-Engel (approved):

PR #1609, opened 2022-05-11
Test | Add unit tests to TDS8 in netcore

lcheunglci:
This is in addition to #1608 to add unit tests to the new TDS8 properties in the connection string.


lcheunglci:
I'm going to close this PR and create a PR directly into @JRahnama fork instead, so it gets integrated in the same PR.

PR #1625, opened 2022-05-26
Fix | Handle NRE on Azure federated authentication

DavoudEshtehari:
Applications owned and operated by Microsoft must get approval from the API owner before requesting tokens for that API, otherwise, the token won't be resolved.


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -2440,9 +2440,15 @@ internal SqlFedAuthToken GetFedAuthToken(SqlFedAuthInfo fedAuthInfo)
                         }
                         else
                         {
+                            SqlClientEventSource.Log.TryTraceEvent("<sc.SqlInternalConnectionTds.GetFedAuthToken.MsalServiceException error:> Timeout: {0}", serviceException.ErrorCode);
                             break;
```
David-Engel, line 2444: There's another potential point here for GerFedAuthToken to return a null token. If we've timed out in a cycle of 429 errors, this break will exit the loop without populating the token. We should throw an exception instead of breaking. Something along the lines of "Connection timed out while retrieving an access token. Last error: {serviceException.StatusCode}: {serviceException.Message}".

David-Engel (approved):

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlUtil.cs
```diff
@@ -656,6 +656,11 @@ static internal Exception ActiveDirectoryDeviceFlowTimeout()
             return ADP.TimeoutException(Strings.SQL_Timeout_Active_Directory_DeviceFlow_Authentication);
         }
 
+        internal static Exception ActiveDirectoryTokenRetrievingTimeout(string authenticaton, string errorCode, Exception exception)
+        {
+            return ADP.TimeoutException(StringsHelper.GetString(Strings.AAD_Token_Retrieving_Timeout, authenticaton, errorCode, exception?.Message), exception);
+        }
```
Kaur-Parminder, line 662: NIT: All methods including this in this class can use body expression. same for NetCore

Kaur-Parminder (commented):
SqlInternalConnectionTds and SqlUtil has a lot of new language version VS suggestions to apply.

DavoudEshtehari:
> SqlInternalConnectionTds and SqlUtil has a lot of new language version VS suggestions to apply.
> NIT: All methods including this in this class can use body expression. same for NetCore

They're a lot to improve, and if you check the [possible merging codebases](https://github.com/dotnet/SqlClient/issues/1261) those are listed there. So, I'd rather don't mix them up with this change.

Issue #1634, opened 2022-06-04
Fix where diagnostic listener SqlDiagnosticsListener.OnAfterCommand runs to avoid implict concurrent execution


davidfowl:
### Describe the bug

See https://github.com/dotnet/aspnetcore/issues/41924 for more background and https://github.com/dotnet/aspnetcore/issues/41924#issuecomment-1146646674 for the context.

When you run an async sql SqlCommand (via ExecuteReaderAsync), the implementation fires the diagnostics listener callback **after** running the user's continuation. This allows the code to run concurrently with the user code, which can introduce accidental parallelism and break in various ways. The linked issues explain how this breaks in a very common app insights scenario.

```
Microsoft.AspNetCore.Http.DefaultHttpRequest.get_Cookies()+8f 
Microsoft.ApplicationInsights.AspNetCore.TelemetryInitializers.WebSessionTelemetryInitializer.UpdateRequestTelemetryFromPlatformContext(Microsoft.ApplicationInsights.DataContracts.RequestTelemetry, Microsoft.AspNetCore.Http.HttpContext)+3c 
Microsoft.ApplicationInsights.AspNetCore.TelemetryInitializers.WebSessionTelemetryInitializer.OnInitializeTelemetry(Microsoft.AspNetCore.Http.HttpContext, Microsoft.ApplicationInsights.DataContracts.RequestTelemetry, Microsoft.ApplicationInsights.Channel.ITelemetry)+c4 
Microsoft.ApplicationInsights.AspNetCore.TelemetryInitializers.TelemetryInitializerBase.Initialize(Microsoft.ApplicationInsights.Channel.ITelemetry)+fa 
Microsoft.ApplicationInsights.TelemetryClient.Initialize(Microsoft.ApplicationInsights.Channel.ITelemetry)+2f9 
Microsoft.ApplicationInsights.TelemetryClient.Track(Microsoft.ApplicationInsights.Channel.ITelemetry)+35 
Microsoft.ApplicationInsights.DependencyCollector.Implementation.SqlClientDiagnostics.SqlClientDiagnosticSourceListener.AfterExecuteHelper(System.Collections.Generic.KeyValuePair`2, Microsoft.ApplicationInsights.Common.PropertyFetcher, Microsoft.ApplicationInsights.Common.PropertyFetcher, Microsoft.ApplicationInsights.Common.PropertyFetcher)+18b 
[...]

PR #1637, opened 2022-06-05
Bug: fix naming, order and formatting for diagnostics

Wraith2:
Fixes https://github.com/dotnet/SqlClient/issues/1634

We were not being consistent about calling diagnostic message output functions before we set the result/exception/cancel methods on the TaskCompletionSource that the users is waiting on. This allowed a rare race condition to occur in libraries consuming diagnostic events and attempting to correlate it to their own current context. 

This PR does 3 things:
1) ensures that all disgnostic message methods in async methods are called before tcs methods.
2) changes the name of the diagnostic source to be consistent with the other uses in the library and the applivable naming rules, it should be `s_diagnosticListener` not `_diagnosticListener`
3) reformats the task continuation lambdas so it's clearer to understand the scope.

N.B. netfx does not use diagnostics so no changes are required for that part of the project.

/cc @okolvik-avento if possible you could try the output artifacts nuget package once the CI run has completed?

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
@@ -114,7 +114,7 @@ namespace Microsoft.Data.SqlClient
         private static bool _forceInternalEndQuery = false;
 #endif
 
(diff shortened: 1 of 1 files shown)


okolvik-avento:
@Wraith2 Running tests now

okolvik-avento:
@Wraith2 Works great! No exceptions after over 500k requests.

okolvik-avento:
Over 5 million requests now without exceptions

David-Engel (approved):
LGTM

PR #1678, opened 2022-07-22
Change | Using default system protocols for TDS8 on Managed code

JRahnama:
This simple PR changes the previous behavior of the driver and omits using supported TLS protocols and uses the latest and the most secure TLS version on the client machine on managed code.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIHandle.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIHandle.cs
@@ -36,7 +36,7 @@ namespace Microsoft.Data.SqlClient.SNI
             {
                 TargetHost = serverNameIndication,
                 ApplicationProtocols = s_tdsProtocols,
-                EnabledSslProtocols = s_supportedProtocols,
+                EnabledSslProtocols = SslProtocols.None,
                 ClientCertificates = certificate
             };
             await sslStream.AuthenticateAsClientAsync(sslClientOptions, token);


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIHandle.cs
```diff
@@ -36,7 +36,7 @@ protected static async Task AuthenticateAsClientAsync(SslStream sslStream, strin
             {
                 TargetHost = serverNameIndication,
                 ApplicationProtocols = s_tdsProtocols,
-                EnabledSslProtocols = s_supportedProtocols,
+                EnabledSslProtocols = SslProtocols.None,
```
David-Engel, line 39: None is the default so we can simply remove the line.
```suggestion
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIHandle.cs
```diff
@@ -36,7 +36,7 @@ protected static async Task AuthenticateAsClientAsync(SslStream sslStream, strin
             {
```
David-Engel, line 36: On line 21, I would also change the comment to:
```C#
/// Exclude TLS 1.3 in TLS-over-TDS modes (TDS 7.4 and below)
```

JRahnama: done.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIHandle.cs
JRahnama: done. I changed it in the original code since I was changing the comment.

David-Engel (approved):

lcheunglci (approved):

DavoudEshtehari (approved):

Issue #1682, opened 2022-07-27
Memory leak when an exception is raised during execution of SqlCommand with CancellationToken


olegkv:
### Description
When using CancellationToken, an exception raised by sqlCommand.ExecuteNonQueryAsync() causes memory leaks.
When not using CancellationToken, an exception raised by sqlCommand.ExecuteNonQueryAsync() does not cause any memory leaks.

### More details about context
When we run a background process which executes sql commands in regular (scheduled) manner (execute-sleep-execute), and using global CancellationToken, process would go out of memory if some commands generate errors. 

### To reproduce
Run the code below. Comment and uncomment calls of ExecuteNonQueryAsync to see the difference between using CancellationToken and not using it.

```c#
CancellationToken stoppingToken = new CancellationTokenSource().Token;
using var connection = new SqlConnection("Data Source=localhost");
while (!stoppingToken.IsCancellationRequested)
{
	try
	{
		using var command = connection.CreateCommand();
		string myString = new string('*', 1000000);
		command.CommandText = "select * from Table1 where Name = @p1";
		var tvpParam = command.Parameters.AddWithValue("@p1", myString);
		tvpParam.SqlDbType = SqlDbType.VarChar;
		//memory leak
		await command.ExecuteNonQueryAsync(stoppingToken);
		//no memory leak
		//await command.ExecuteNonQueryAsync();
	}
	catch
	{
	}
}
```

### Expected behavior
When using CancellationToken, memory usage should be the same as in the case of not using CancellationToken

### Further technical details
Microsoft.Data.SqlClient version: (4.1.0)
.NET target: (Net 5 and 6)
Operating system: (Windows)

ErikEJ:
There is no (4.8.3) version of Microsoft.Data.SqlClient. are you testing against the deprecated System.Data.SqlClient ?

olegkv:
Corrected SqlClient version. It is version 4.1.0 of Microsoft.Data.SqlClient.

olegkv:
On first glance, it looks like you just need to add 
a call to registration.Dispose();
to this catch:
https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs#L2573

Because if BeginExecuteNonQueryAsync throws an exception, ContinueWith (which calls registration.Dispose()) will not be called.
FromAsync documentation:
[...]

PR #1692, opened 2022-08-04
Convert ExecuteNonQueryAsync to use async context object

Wraith2:
fixes https://github.com/dotnet/SqlClient/issues/1682

This addresses the incorrect handling of cancellation token registrations in `SqlCommand` methods `ExecuteNonQuery` and `ExecuteReader`, I didn't change `ExecuteScalar` because it should be done in a separate PR and it's just `ExecuteReader` wearing a fake nose a glasses, people should use `ExexuteReader`

Background information: There are a number of async api calls on `SqlDataReader` which have been changed to avoid the use of language provided lambda closures and instead use explicit context objects. This process makes the code more complicated because it splits apart the code into multiple functions so that instance delegate allocations can be avoided. This pattern is no longer needed if we use language provided static lambdas and the code can be much cleaner and better contained. This PR converts the ExecuteNonQuery function to use this explicity context pattern and uses static lambdas to doi t.

This PR also changes all existing context objects to use an explicit type for the IDispoable so that we can avoid boxing the disposable. This change forces any existing context types which were declared like this: `AAsyncContext<TOwner,TResult>` to be split apart into two types, 
1) `AAsyncContext<TOwner,TResult,TDisposable>` which has a strong disposable type and is used at call sites that know what the disposable is, mostly setup locations.  
2)  `AAsyncBaseContext<TOwner,TResult>` which is used in the async functions that don't need to know what the type of the disposable is so we can avoid duplicating functions like `InvokeAsync`

The specific changes to fix the issue reported are 
1) change `InternalExecuteNonQueryAsync` to use an explicity context object 
2) make disposable idempotent by including a disposed flag. Previously it was not possible to reach disposable from more than one path so knowledge of disposal was implicit, now it is explicit.
3) call context dispose from catch blocks ensuring that the registration is correctly cleaned up on all paths

/cc @olegkv

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/AAsyncCallContext.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/AAsyncCallContext.cs
@@ -42,7 +42,6 @@ namespace Microsoft.Data.SqlClient
         {
             TDisposable copyDisposable = _disposable;
             _disposable = default;
-            _isDisposed = true;
             copyDisposable?.Dispose();
         }
     }
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/AAsyncCallContext.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/AAsyncCallContext.cs
@@ -17,37 +17,68 @@ namespace Microsoft.Data.SqlClient
     // CONSIDER creating your own Set method that calls the base Set rather than providing a parameterized ctor, it is friendlier to caching
     // DO NOT use this class' state after Dispose has been called. It will not throw ObjectDisposedException but it will be a cleared object
 
-    internal abstract class AAsyncCallContext<TOwner, TTask> : IDisposable
+    internal abstract class AAsyncCallContext<TOwner, TTask, TDisposable> : AAsyncBaseCallContext<TOwner,TTask>
         where TOwner : class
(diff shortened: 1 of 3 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/AAsyncCallContext.cs
```diff
@@ -17,38 +17,69 @@ namespace Microsoft.Data.SqlClient
...
+        {
+            base.Set(owner, source);
+            _disposable = disposable;
+        }
+
+        protected override void DisposeCore()
+        {
+            TDisposable copyDisposable = _disposable;
+            _disposable = default;
+            _isDisposed = true;
```
olegkv, line 45: isDisposed is already set by parent class in ClearCore()
ReadAsyncCallContext class is another descendant which seems to do it in more consistent way, it does not touch _isDisposed and relies on parent class instead. Maybe isDisposed should be made private..

lcheunglci:
@Wraith2 Looks like there's a merge conflict. Let us know if you would like us to resolve it for you?

David-Engel:
> Looks like there's a merge conflict. Let us know if you would like us to resolve it for you.

I took care of it. It was a small change from #1781.

David-Engel:
That'll teach me to use the GitHub conflict UI. I broke it. I'm looking at it.

Wraith2:
Rebased, checked it still makes sense and addressed the feedback above.

David-Engel (approved):

JRahnama (approved):

PR #1688, opened 2022-07-29
Fix null binary rowversion and add test coverage

Wraith2:
fixes https://github.com/dotnet/SqlClient/issues/1684

The fix is very simple and simply adds a a null check that was already present in all the other SqlType returning properties on SqlBuffer. Test coverage is added for SqlBinary and SqlBytes in the existing test.

/cc @rclabo, @dbrownems

PR #1712, opened 2022-08-11
[Hotfix 4.1.1] | Fix null binary rowversion and add test coverage (#1688)

DavoudEshtehari:
Ports #1688


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBuffer.cs
```diff
@@ -625,6 +625,10 @@ internal SqlBinary SqlBinary
             {
                 if (StorageType.SqlBinary == _type)
                 {
+                    if (IsNull)
```
David-Engel, line 628: We also need to port this to the netfx SqlBuffer.cs file. The fix in main only had one file because the file had been merged by that point.

David-Engel (approved):

lcheunglci (approved):

JRahnama (approved):

PR #1766, opened 2022-09-17
Fix async deadlock issue when sending attention fails due to network failure

cheenamalhotra:
In an edge case scenario, when a network failure occurs when drives attempts to send an attention signal, the driver fails to throw exception and terminate connection, leaving the connection open indefinitely, when async APIs are used to query the database. This can cause applications to go into hang due to driver getting into a deadlock situation as the async callbacks won't clear.

The issue is associated with the synchronous timeout flow that kicks off on ReadAsyncCallback, which will only be executed when timeout threads are not able to acquire thread from Threadpool and execute until callback returns on the other thread and finds out that the timeout is expired but attention is not yet triggered.

This PR fixes the issue by enabling 'asyncClose' flag on ThrowExceptionAndWarning methods when performing timeout activity via 'ReadAsyncCallback" as the native SNI needs to clear the callback on termination of connection. This was a missed opportunity in PR #906 that caused this edge case scenario to affect a few customers.

I tried adding tests, but it won't be possible to neatly add a test, without causing a lot of code changes, therefore skipping here. Linking repro below which can be easily executed and tested with the PR.

Link to Repro App: https://gist.github.com/cheenamalhotra/94d501e093a0392520fe2095c4bb7491

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
@@ -55,7 +55,7 @@ namespace Microsoft.Data.SqlClient
 
         private const int AttentionTimeoutSeconds = 5;
 
-        private static readonly ContextCallback s_readAdyncCallbackComplete = ReadAsyncCallbackComplete;
+        private static readonly ContextCallback s_readAsyncCallbackComplete = ReadAsyncCallbackComplete;
 
         // Ticks to consider a connection "good" after a successful I/O (10,000 ticks = 1 ms)
         // The resolution of the timer is typically in the range 10 to 16 milliseconds according to msdn.
@@ -2337,9 +2337,9 @@ namespace Microsoft.Data.SqlClient
             }
         }
 
-        private bool OnTimeoutSync()
+        private bool OnTimeoutSync(bool asyncClose = false)
         {
-            return OnTimeoutCore(TimeoutState.Running, TimeoutState.ExpiredSync);
+            return OnTimeoutCore(TimeoutState.Running, TimeoutState.ExpiredSync, asyncClose);
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
```diff
@@ -2337,9 +2337,9 @@ private void OnTimeoutAsync(object state)
             }
         }
 
-        private bool OnTimeoutSync()
+        private bool OnTimeoutSync(bool asyncClose = false)
```
DavoudEshtehari, line 2340: NIT
```suggestion
        private bool OnTimeout(bool asyncClose = false)
```

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
```diff
@@ -2401,9 +2401,9 @@ private void OnTimeoutAsync(object state)
             }
         }
 
-        private bool OnTimeoutSync()
+        private bool OnTimeoutSync(bool asyncClose = false)
```
DavoudEshtehari, line 2404:
```suggestion
        private bool OnTimeout(bool asyncClose = false)
```

DavoudEshtehari (approved):
LGTM 🎉

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
cheenamalhotra: FYI - AsyncClose has nothing to do with OnTimeout being run synchronously. It's weird, but AsyncClose is specifically needed when OnTimeoutSync gets called from ReadAsyncCallback.

lcheunglci (approved):

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
```diff
@@ -3603,7 +3604,7 @@ private Task SNIWritePacket(SNIHandle handle, SNIPacket packet, out UInt32 sniEr
                 {
                     SqlClientEventSource.Log.TryTraceEvent("<sc.TdsParser.WritePacket|Info> write async returned error code {0}", (int)sniError);
                     AddError(_parser.ProcessSNIError(this));
-                    ThrowExceptionAndWarning(callerHasConnectionLock);
+                    ThrowExceptionAndWarning(callerHasConnectionLock, false);
```
David-Engel, line 3607: @cheenamalhotra Did we purposely call `ThrowExceptionAndWarning` with the `asyncClose` parameter hard coded to `false` here? In the same line in netcore, we passed the local value of `asyncClose` to the same parameter. ([Line 3573 in the netcore TdsParserStateObject.cs file above in this PR](https://github.com/dotnet/SqlClient/pull/1766/files#diff-e15040521566aad8d026726b5b8269aeeecdfb7d8e92e3f85c52f37be3dc752fR3573).)

This seems like a mistake and I've changed this line to match netcore in PR #2073.

PR #1771, opened 2022-09-23
Make SqlConnectionEncryptOption string parser public

cheenamalhotra:
It would be useful to have these parsers public to be able to parse string values directly when needed, instead of consumer apps writing their own.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
@@ -506,6 +506,10 @@ namespace Microsoft.Data.SqlClient
     /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnectionEncryptOption.xml' path='docs/members[@name="SqlConnectionEncryptOption"]/SqlConnectionEncryptOption/*'/>
     public sealed class SqlConnectionEncryptOption
     {
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnectionEncryptOption.xml' path='docs/members[@name="SqlConnectionEncryptOption"]/Parse/*' />
+        public static SqlConnectionEncryptOption Parse(string value) => null;
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnectionEncryptOption.xml' path='docs/members[@name="SqlConnectionEncryptOption"]/TryParse/*' />
+        public static bool? TryParse(string value, out SqlConnectionEncryptOption result) => throw null;
         /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnectionEncryptOption.xml' path='docs/members[@name="SqlConnectionEncryptOption"]/Optional/*' />
         public static SqlConnectionEncryptOption Optional => throw null;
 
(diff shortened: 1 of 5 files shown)


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionEncryptOption.cs
```diff
@@ -54,6 +56,21 @@ internal static SqlConnectionEncryptOption Parse(string value)
...
+        /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnectionEncryptOption.xml' path='docs/members[@name="SqlConnectionEncryptOption"]/TryParse/*' />
+        public static bool TryParse(string value, out SqlConnectionEncryptOption result)
+        {
+            result = null;
+            try
+            {
+                result = Parse(value);
+                return true;
+            }
+            catch
```
David-Engel, line 68: Exceptions are expensive. I would move the Parse logic into TryParse, refactoring it a bit to not throw an exception. Parse can then call TryParse and throw an exception if it fails.

David-Engel (approved):

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -506,6 +506,10 @@ public enum SqlConnectionIPAddressPreference
     /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnectionEncryptOption.xml' path='docs/members[@name="SqlConnectionEncryptOption"]/SqlConnectionEncryptOption/*'/>
     public sealed class SqlConnectionEncryptOption
     {
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnectionEncryptOption.xml' path='docs/members[@name="SqlConnectionEncryptOption"]/Parse/*' />
+        public static SqlConnectionEncryptOption Parse(string value) => null;
```
DavoudEshtehari, line 510:
```suggestion
        public static SqlConnectionEncryptOption Parse(string value) => throw null;
```

src/Microsoft.Data.SqlClient/netfx/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -923,6 +923,12 @@ public enum SqlConnectionIPAddressPreference
     /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnectionEncryptOption.xml' path='docs/members[@name="SqlConnectionEncryptOption"]/SqlConnectionEncryptOption/*'/>
     public sealed class SqlConnectionEncryptOption
     {
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnectionEncryptOption.xml' path='docs/members[@name="SqlConnectionEncryptOption"]/Parse/*' />
+        public static SqlConnectionEncryptOption Parse(string value) => null;
```
DavoudEshtehari, line 927:
```suggestion
        public static SqlConnectionEncryptOption Parse(string value) => throw null;
```

DavoudEshtehari (commented):
`(SqlConnectionEncryptOption)null` returns `True`. Just raising it for more discussion.
If it's by design it'd better be mentioned in the documentation, plus extending the code coverage.

cc @David-Engel

David-Engel:
> `(SqlConnectionEncryptOption)null` returns `True`. Just raising it for more discussion. If it's by design it'd better be mentioned in the documentation, plus extending the code coverage.

@DavoudEshtehari 

Do you have a use case for that scenario? It wasn't intentional (it wasn't even considered), but since that's the behavior, we should maintain it by adding it to the tests.

DavoudEshtehari:
> Do you have a use case for that scenario? It wasn't intentional (it wasn't even considered), but since that's the behavior, we should maintain it by adding it to the tests.

No, just a random test. 

I'd suggest these 2 options to improve this part:
1- Overriding nullable bool instead of a bool type to support `null`:
```
public static implicit operator bool?(SqlConnectionEncryptOption value) => !Optional?.Equals(value);
```
2- Keep it as it is instead of interpreting `null` to `false` not `true`:
```
public static implicit operator bool(SqlConnectionEncryptOption value) => value is not null && !Optional.Equals(value);
```

lcheunglci:
I realize I missed the discussion before approving the changes, but the changes is relatively safe and tests were included to validate them so perhaps the suggestion from @DavoudEshtehari could be handled in separate PR.

PR #1307, opened 2021-10-04
Move to Shared for the Microsoft.Data.SqlClient Reference project

lcheunglci:
This is related to #1261 but it's not in the list.  @cheenamalhotra mentioned that we would like to merge the two reference projects into one as the long term goal is to have one codebase.  I merged the netfx csproj into netcore and combined the Microsoft.Data.SqlClient.cs.  I ran MSBUILD on release and generated the nuget package and compared it before it was merged.  The file size remained the same and I used [NuGet Package Explorer](https://github.com/NuGetPackageExplorer/NuGetPackageExplorer) to view the contents of the package and it looked the same.

PR #1776, opened 2022-09-27
Add missing HostNameInCertificate property in NetFx Ref

cheenamalhotra:


David-Engel (approved):

lcheunglci (approved):

ErikEJ:
I have seen this happen before with another new connection string property. Wonder if it would be possible to author a general test to prevent this from happening again?

cheenamalhotra:
Maybe time to revive https://github.com/dotnet/SqlClient/pull/1307 and enable "Package" reference tests to be run once in a while..

PR #1781, opened 2022-09-30
Fixes ReadAsync() behavior to register Cancellation token action before streaming results

### LINKED ISSUE #1065: Triggering a cancellation token may not cancel a running query based on timing
When an async cancellation token is triggered, a running query is expected to be cancelled at the server; this apparently works fine when the token is triggered during ExecuteReaderAsync, or during ReadAsync.

However, if the token happens to be triggered while no SqlClient code is in progress (e.g. ReadAsync), the next time ReadAsync is called, it would simply return a cancelled Task immediately, without cancelling the query at the server. This means that there's a race condition - depending on when the token triggering happens, the query may be cancelled (if ReadAsync happens to be in progress) or it may not.

In Npgsql, we handled this race conditions by cancelling the (running) server query from ReadAsync, even when it's called with an already-cancelled token. This is being discussed in https://github.com/DapperLib/Dapper/pull/1590#issuecomment-739732477, where Dapper may be working around this issue by registering a callback to call SqlCommand.Cancel when the token is triggered - this should IMHO not be necessary.

An alternative strategy would be to continue monitoring the cancellation token originally passed to ExecuteReaderAsync - even after ExecuteReaderAsync itself has already completed. This would have the advantage of allowing immediate cancellation at the server at any point, and not only when ReadAsync is called.

For a bit more context, I wrote a small [blog post](https://www.roji.org/db-commands-and-cancellation) on ADO.NET and cancellation.

cheenamalhotra:
Fixes #1065 with test case added from [comment](https://github.com/dotnet/SqlClient/issues/1065#issuecomment-1262671004).

cc @jnm2 @roji would appreciate some extra validations!

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
@@ -4733,16 +4733,17 @@ namespace Microsoft.Data.SqlClient
                     return Task.FromException<bool>(ADP.ExceptionWithStackTrace(ADP.DataReaderClosed()));
                 }
 
-                // If user's token is canceled, return a canceled task
-                if (cancellationToken.IsCancellationRequested)
-                {
-                    return Task.FromCanceled<bool>(cancellationToken);
-                }
-
+                // Register first to catch any already expired tokens to be able to trigger cancellation event.
                 IDisposable registration = null;
                 if (cancellationToken.CanBeCanceled)
                 {
                     registration = cancellationToken.Register(SqlCommand.s_cancelIgnoreFailure, _command);
+                }
+
+                // If user's token is canceled, return a canceled task
+                if (cancellationToken.IsCancellationRequested)
+                {
+                    return Task.FromCanceled<bool>(cancellationToken);
                 }
 
                 // Check for existing async
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
@@ -4739,6 +4739,12 @@ namespace Microsoft.Data.SqlClient
                     return Task.FromCanceled<bool>(cancellationToken);
                 }
 
+                IDisposable registration = null;
+                if (cancellationToken.CanBeCanceled)
+                {
+                    registration = cancellationToken.Register(SqlCommand.s_cancelIgnoreFailure, _command);
+                }
+
                 // Check for existing async
                 if (_currentTask != null)
                 {
@@ -4829,12 +4835,6 @@ namespace Microsoft.Data.SqlClient
                     source.SetCanceled();
                     _currentTask = null;
                     return source.Task;
-                }
-
-                IDisposable registration = null;
-                if (cancellationToken.CanBeCanceled)
-                {
-                    registration = cancellationToken.Register(SqlCommand.s_cancelIgnoreFailure, _command);
                 }
 
                 ReadAsyncCallContext context = null;
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataReaderTest/DataReaderCancellationTest.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataReaderTest/DataReaderCancellationTest.cs
@@ -42,7 +42,41 @@ from ThousandRows as A, ThousandRows as B, ThousandRows as C;";
                         }
                     }
                 });
-                Assert.True(stopwatch.ElapsedMilliseconds < 1000, "Cancellation did not trigger on time.");
+                Assert.True(stopwatch.ElapsedMilliseconds < 10000, "Cancellation did not trigger on time.");
+            }
+        }
+
+        /// <summary>
+        /// Test ensures cancellation token is registered before ReadAsync starts processing results from TDS Stream,
+        /// such that when Cancel is triggered, the token is capable of canceling reading further results.
+        /// </summary>
+        /// <returns>Async Task</returns>
+        [ConditionalFact(typeof(DataTestUtility), nameof(DataTestUtility.AreConnStringsSetup))]
+        public static async Task CancelledCancellationTokenIsRespected_ReadAsync()
+        {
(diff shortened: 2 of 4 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
```diff
@@ -4739,6 +4739,12 @@ public override Task<bool> ReadAsync(CancellationToken cancellationToken)
                     return Task.FromCanceled<bool>(cancellationToken);
                 }
 
+                IDisposable registration = null;
+                if (cancellationToken.CanBeCanceled)
```
roji, line 4743: Happy to see your contributions again @cheenamalhotra :)

I don't know enough about this code to fully understand what moving the registration does (but can look more in-depth).

However, the original reason I opened #1065 was the line just above - if IsCancellationRequested is true, ReadAsync returns immediately. That means that if the token is cancelled just before ReadAsync is invoked, it will not actually cancel the query in SQL Server, and it will keep running.

Note that if a token is in the cancelled state, [CancellationToken.Register](https://learn.microsoft.com/en-us/dotnet/api/system.threading.cancellationtoken.register?view=net-7.0#system-threading-cancellationtoken-register(system-action)) immediately runs the callback. That means that it may be enough to reverse the order here, and call Register *before* checking whether IsCancellationRequested is true and exiting.

Note: it's possible that your change still improves the situation and fixes the scenario in https://github.com/dotnet/SqlClient/issues/1065#issuecomment-1262671004; but as long as ReadAsync simply exits early when IsCancellationRequested is true, I don't think #1065 is fixed...

cheenamalhotra: It is nice to be around again 🙂 SqlClient is indeed my favourite team ❤️ I just got notified today and realized this issue was left out in my backlog...

I agree I missed the before call cancellation scenario here. Will move registration up further and add test-case accordingly tomorrow!

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataReaderTest/DataReaderCancellationTest.cs
```diff
@@ -0,0 +1,49 @@
...
+        /// </summary>
+        /// <returns>Async Task</returns>
+        [ConditionalFact(typeof(DataTestUtility), nameof(DataTestUtility.AreConnStringsSetup))]
+        public static async Task CancellationTokenIsRespected_ReadAsync()
+        {
+            const string longRunningQuery = @"
+with TenRows as (select Value from (values (1), (2), (3), (4), (5), (6), (7), (8), (9), (10)) as TenRows (Value)),
+    ThousandRows as (select A.Value as A, B.Value as B, C.Value as C from TenRows as A, TenRows as B, TenRows as C)
+select *
+from ThousandRows as A, ThousandRows as B, ThousandRows as C;";
```
jnm2, line 26: Reading a billion rows could take a long time. Would this be a concern in the test suite, on the off chance that this test is regressed?

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataReaderTest/DataReaderCancellationTest.cs
```diff
@@ -0,0 +1,49 @@
...
+                    using (var command = new SqlCommand(longRunningQuery, connection))
+                    using (var reader = await command.ExecuteReaderAsync(source.Token))
+                    {
+                        while (await reader.ReadAsync(source.Token))
+                        {
+                            source.Cancel();
+                        }
+                    }
+                });
+                Assert.True(stopwatch.ElapsedMilliseconds < 1000, "Cancellation did not trigger on time.");
```
jnm2, line 45: Because I'm used to CI build servers getting random pauses that are longer than a second, 10 seconds is the length of time I've hit on that seems to reliably avoid false positive detections in test suites. (Is there a standard practice within this test suite?)

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
cheenamalhotra: Addressed.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataReaderTest/DataReaderCancellationTest.cs
cheenamalhotra: So it shouldn't regress :)
But I'll wait for the CI builds to say that - I don't think that should be a problem IMO.

jnm2: The danger is that someone works in this area and does regress it within their branch, and every CI job takes 60 minutes to run and times out before they figure out what's going on. Just an inconvenience.

cheenamalhotra: I don't think it should be that bad as it takes around 60 seconds to complete query execution locally. We cannot add a 'WAITFOR DELAY' or any procedure call, as that would cause delay in ExecuteReaderAsync and not ReadAsync. And ReadAsync can only be delayed with a large dataset. If you have a better query, let me know and I can evaluate. We also want to be able to consistently rely in tests, as intermittent test failures are more painful to identify regressions.

JRahnama: > @JRahnama any clue why CI builds are not getting triggered?

Can you commit some changes? Azure had some difficulties resolving Default branch for manual and scheduled builds address which is now solved and should be working.

cheenamalhotra: @JRahnama - I've updated the service connection so it seems to be triggering CI builds now. Just FYI, in case you wonder later 🙂

lcheunglci:
@cheenamalhotra Please backport this to the 5.0.1 as well. Thanks!

lcheunglci (approved):

PR #1786, opened 2022-10-05
clean up StreamingFromServer

SimonCropp:
since it renders way too wide 
image


lcheunglci:
Thanks for the PR, we're currently in the middle of a release, so we'll have to get back to reviewing it later.

PR #1798, opened 2022-10-07
[5.0.1] Hang on infinite timeout and managed SNI

### LINKED ISSUE #1733: Unable to connect to SqlServer instance with managed SNI and Timeout=0 on SqlClient 5.0
### Describe the bug
While attempting to establish the connection to SqlServer instance while having `Timeout=0` in the connection string the app becomes unresponsive. I'm able to reproduce this problem on linux and windows with `AppContext.SetSwitch("Switch.Microsoft.Data.SqlClient.UseManagedNetworkingOnWindows", true)`. It works fine on SqlClient 4.1.0.

### To reproduce

```c#
AppContext.SetSwitch("Switch.Microsoft.Data.SqlClient.UseManagedNetworkingOnWindows", true);

var sw = Stopwatch.StartNew();
using var connection = new SqlConnection("Server=.;Database=master;Integrated Security=false;User ID=sa;Password=SomePassword;Pooling=true;Max Pool Size=1024;TrustServerCertificate=true;Timeout=0");
await connection.OpenAsync();

using var cmd = connection.CreateCommand();
cmd.CommandText = "SELECT 1";
await cmd.ExecuteNonQueryAsync();
Console.WriteLine($"Connection established successfully in {sw.ElapsedMilliseconds}ms. Press any key to continue...");
Console.ReadKey();
```

### Expected behavior
SqlClient should be able to establish a connection with `Timeout=0`, just like in previous verions

### Further technical details
Microsoft.Data.SqlClient 5.0
.NET target: .NET 6.0
SQL Server version: SQL Server 2017
Operating system: CentOS 7, Windows 10

**Additional context**

The problem is because while waiting for the dns, the timeout passed there is the default `TimeSpan`, which is 0, and that's why it fails immediately (it should have been -1):

https://github.com/dotnet/SqlClient/blob/bd21d762e2362256aab49866e5ad9797b4a037b4/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs#L150-L159

https://github.com/dotnet/SqlClient/blob/bd21d762e2362256aab49866e5ad9797b4a037b4/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs#L362

This was introduced in #1578.

JRahnama:
Backporting #1742, which fixes #1733.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNICommon.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNICommon.cs
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
+using System.Net;
 using System.Net.Security;
 using System.Security.Cryptography.X509Certificates;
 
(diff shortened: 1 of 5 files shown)


cheenamalhotra:
Can you link or describe the issue with a repro?

David-Engel (approved):

JRahnama:
> Can you link or describe the issue with a repro?

The regression introduced in https://github.com/dotnet/SqlClient/pull/1578. The issue was first reported at https://github.com/dotnet/SqlClient/issues/1733

cheenamalhotra:
Thanks for bringing this in <3

PR #1801, opened 2022-10-09
Release connection lock before signaling SinglePhaseCommit completion

### LINKED ISSUE #1800: Deadlock when using SinglePhaseCommit with distributed transactions
In .NET 7.0, the OleTx/MSDTC distributed transactions support has been ported over from .NET Framework (https://github.com/dotnet/runtime/issues/715). This means that it will now be possible to use distributed transactions with .NET 7.0, on Windows only.

We've received a bug report (https://github.com/dotnet/runtime/issues/76010) with two separate console programs, enlisting which propagate a transaction. The initiating program (the "outer" transaction) enlists only a single SqlConnection to the transaction, and hangs when the transaction is committed. This happens quite consistently, but not 100% reliably. To reproduce this, simply run the .NET 7 version of the application in https://github.com/dotnet/runtime/issues/76010 in two separate windows; the first one should hang.

When the program hangs, two threads are stuck with the following stack traces:

### Stack 1

```
SinglePhaseEnlistment.Committed()at C:\Users\shrojans\AppData\Roaming\JetBrains\Rider2022.2\resharper-host\SourcesCache\d42a13b4dda85c7ff6fe99f383b4a1b279abd39735cb45449b7401d99c387\SinglePhaseEnlistment.cs:line 62
SqlDelegatedTransaction.SinglePhaseCommit()at C:\Users\shrojans\AppData\Roaming\JetBrains\Rider2022.2\resharper-host\SourcesCache\7c62ff4f1cfc3e61c3e65f286d919f9f0bbb2ef6a822b99cb37623d6d3a891\SqlDelegatedTransaction.cs:line 389
TransactionStateDelegatedCommitting.EnterState()
CommittableTransaction.Commit()
TransactionScope.InternalDispose()
TransactionScope.Dispose()
Program.StartTransaction()
async Program.Main()
AsyncMethodBuilderCore.Start<System.__Canon>()
Program.Main()
Program.<Main>()
```

### Stack 2:

```
DbConnectionInternal.DetachTransaction()at C:\Users\shrojans\AppData\Roaming\JetBrains\Rider2022.2\resharper-host\SourcesCache\71847c97c677567f13e9df7e4f1f44a99f349762bb77c34f819752ee84b71\DbConnectionInternal.cs:line 413
DbConnectionInternal.CleanupConnectionOnTransactionCompletion()at C:\Users\shrojans\AppData\Roaming\JetBrains\Rider2022.2\resharper-host\SourcesCache\71847c97c677567f13e9df7e4f1f44a99f349762bb77c34f819752ee84b71\DbConnectionInternal.cs:line 436
DbConnectionInternal.TransactionCompletedEvent()
TransactionStatePromotedCommitted.EnterState()
InternalTransaction.DistributedTransactionOutcome()
RealOletxTransaction.FireOutcome()
OutcomeEnlistment.InvokeOutcomeFunction()
OletxTransactionManager.ShimNotificationCallback()
PortableThreadPool.CompleteWait()
ThreadPoolWorkQueue.Dispatch()
PortableThreadPool.WorkerThread.WorkerThreadStart()
[Native to Managed Transition]
```

Here is the general flow leading up to this state:

1. When the TransactionScope is disposed, we get to `SqlDelegatedTransaction.SinglePhaseCommit()`, which is the SqlClient part that interacts with System.Transactions.
  a. `SqlDelegatedTransaction.SinglePhaseCommit()` locks `connection` (the SqlInternalConnectionTds)
  b. It then sends Commit to SQL Server. Since the transaction is delegated, the triggers the commit with MSDTC, which causes commit notifications to be sent to the enlisted parties. This includes the running program, which starts thread 2 below running concurrently.
  b. Finally, it proceeds to call `enlistment.Committed()`, which is `SinglePhaseEnlistment.Committed()`, while keeping the lock on SqlInternalConnectionTds. That method then takes `_internalEnlistment.SyncRoot`, which is the InternalTransaction.

2. The above commit causes us to get a notification from the native layer (MSDTC) that the transaction committed (triggered above in 1b).
  a. We get to `InternalTransaction.DistributedTransactionOutcome()`, which locks the InternalTransaction.
  b. We then get to `DbConnectionInternal.DetachTransaction()` (DbConnectionInternal is registered as a listener on the Transaction.TransactionCompleted event), which attempts to lock `this` (the SqlInternalConnectionTds).

So thread 1 locks the SqlInternalConnectionTds and then the InternalTransaction, while thread 2 - which runs concurrently - locks InternalTransaction and then SqlInternalConnectionTds. This produces a deadlock.

I'm not an expert in this code, but moving the call to `enlistment.Committed()` ([code](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDelegatedTransaction.cs#L418)) outside of the lock could be a way to resolve this deadlock.

Note that I couldn't reproduce the deadlock in .NET Framework; there's likely some timing differences which make the deadlock manifest far more frequently on .NET Core, but AFAICT the bug is in Framework as well. Note that this repro uses SinglePhaseCommit - only one connection is enlisted to the TransactionScope in the application. I recommend also checking having two connections to force 2PC - the same bug could be present in that flow as well.

Many thanks to @nathangreaves for providing the original repro code.

/cc @ajcvickers

roji:
This is a possible fix for #1800, introduced by #1042; but it may re-introduce the issue that #1042 was meant to fix.

Fixes #1800

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDelegatedTransaction.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDelegatedTransaction.cs
@@ -348,6 +348,8 @@ namespace Microsoft.Data.SqlClient
 #endif
                 try
                 {
+                    Exception commitException = null;
+
                     lock (connection)
                     {
                         // If the connection is doomed, we can be certain that the
@@ -362,7 +364,6 @@ namespace Microsoft.Data.SqlClient
                         }
                         else
                         {
-                            Exception commitException;
                             try
                             {
                                 // Now that we've acquired the lock, make sure we still have valid state for this operation.
@@ -372,7 +373,6 @@ namespace Microsoft.Data.SqlClient
                                 _connection = null;   // Set prior to ExecuteTransaction call in case this initiates a TransactionEnd event
 
                                 connection.ExecuteTransaction(SqlInternalConnection.TransactionRequest.Commit, null, System.Data.IsolationLevel.Unspecified, _internalTransaction, true);
-                                commitException = null;
                             }
                             catch (SqlException e)
                             {
@@ -420,12 +420,13 @@ namespace Microsoft.Data.SqlClient
                             }
 
                             connection.CleanupConnectionOnTransactionCompletion(_atomicTransaction);
-                            if (commitException == null)
-                            {
-                                // connection.ExecuteTransaction succeeded
-                                enlistment.Committed();
-                            }
                         }
+                    }
+
+                    if (commitException == null)
+                    {
+                        // connection.ExecuteTransaction succeeded
+                        enlistment.Committed();
                     }
                 }
                 catch (System.OutOfMemoryException e)
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/Microsoft.Data.SqlClient.ManualTesting.Tests.csproj
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/Microsoft.Data.SqlClient.ManualTesting.Tests.csproj
@@ -192,6 +192,7 @@
     <Compile Include="SQL\SqlNotificationTest\SqlNotificationTest.cs" />
     <Compile Include="SQL\SqlSchemaInfoTest\SqlSchemaInfoTest.cs" />
     <Compile Include="SQL\SqlStatisticsTest\SqlStatisticsTest.cs" />
+    <Compile Include="SQL\TransactionTest\DistributedTransactionTest.cs" />
     <Compile Include="SQL\TransactionTest\TransactionTest.cs" />
     <Compile Include="SQL\TransactionTest\TransactionEnlistmentTest.cs" />
     <Compile Include="SQL\UdtTest\SqlServerTypesTest.cs" />
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/TransactionTest/DistributedTransactionTest.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/TransactionTest/DistributedTransactionTest.cs
@@ -1,0 +1,38 @@
+﻿using System.Threading.Tasks;
+using System.Transactions;
+using Xunit;
+
+#if NET7_0_OR_GREATER
+
+namespace Microsoft.Data.SqlClient.ManualTesting.Tests
+{
+    [PlatformSpecific(TestPlatforms.Windows)]
+    public class DistributedTransactionTest
+    {
+        [ConditionalFact]
+        public async Task Delegated_transaction_deadlock_in_SinglePhaseCommit()
+        {
+            TransactionManager.ImplicitDistributedTransactions = true;
+            using var transaction = new CommittableTransaction();
+
+            // Uncommenting the following makes the deadlock go away as a workaround. If the transaction is promoted before
+            // the first SqlClient enlistment, it never goes into the delegated state.
+            // _ = TransactionInterop.GetTransmitterPropagationToken(transaction);
+            await using var conn = new SqlConnection(DataTestUtility.TCPConnectionString);
+            await conn.OpenAsync();
+            conn.EnlistTransaction(transaction);
+
+            // Enlisting the transaction in second connection causes the transaction to be promoted.
+            // After this, the transaction state will be "delegated" (delegated to SQL Server), and the commit below will
+            // trigger a call to SqlDelegatedTransaction.SinglePhaseCommit.
+            await using var conn2 = new SqlConnection(DataTestUtility.TCPConnectionString);
+            await conn2.OpenAsync();
+            conn2.EnlistTransaction(transaction);
+
+            transaction.Commit();
+            // We never get here because of the deadlock
+        }
+    }
+}
+
+#endif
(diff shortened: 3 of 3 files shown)


cheenamalhotra:
Hi @roji 

Apologies for the delay, I've been heads down in release work. I tested these changes today with multiple cases and the fix looks good to me overall. 

However, since we don't have `TransactionFlow` attribute support in .NET 7 yet, we cannot replicate the scenario from .NET Framework where external service would interfere with distributed transactions. But I verified the change in .NET Framework, and I don't see any regression with this change either so I would like to believe it should be safe.

I couldn't reproduce this issue in .NET Framework too with the repro provided, so I'm not sure if we want to backport this fix just yet as Framework's been stable with the changes so I would be hesitant to touch it. I think due to lack of Distribution Transaction support in .NET Core, we couldn't really test this back then. But I'll be watchful of if/when `TransactionFlow` attribute support arrives with WCF services, we should validate #237 in .NET 7+ once.

cheenamalhotra (approved):
Would suggest adding a test case for future proofing this scenario with the repro from https://github.com/dotnet/SqlClient/issues/1800#issuecomment-1272482132

roji:
Thanks for reviewing @cheenamalhotra!

> Would suggest adding a test case for future proofing this scenario with the repro from https://github.com/dotnet/SqlClient/issues/1800#issuecomment-1272482132

Giving that this is a timing-sensitive race condition, that repro runs in a `while (true)` loop and isn't suitable as a regression test that always runs, I think... Or are you thinking about a manually-run test? Can you point me to where to add that etc.?

cheenamalhotra:
Hi @roji 

I got the repro simplified and can almost reproduce the issue everytime with this code..

```cs
    string connString = "Server=localhost; Integrated Security=true; Pooling=false;"; // remember to disable pooling

    TransactionManager.ImplicitDistributedTransactions = true;
    using var transaction = new CommittableTransaction();

    // Uncommenting the following makes the deadlock go away as a workaround. If the transaction is promoted before
    // the first SqlClient enlistment, it never goes into the delegated state.
    // _ = TransactionInterop.GetTransmitterPropagationToken(transaction);
    await using var conn = new SqlConnection(connString);
    await conn.OpenAsync();
    conn.EnlistTransaction(transaction);
    await conn.CloseAsync();

    // Enlisting the transaction in second connection causes the transaction to be promoted.
    // After this, the transaction state will be "delegated" (delegated to SQL Server), and the commit below will
    // trigger a call to SqlDelegatedTransaction.SinglePhaseCommit.
    await using var conn2 = new SqlConnection(connString);
    await conn2.OpenAsync();
    conn2.EnlistTransaction(transaction);

    // Enlisting the transaction in third connection causes the deadlock on transaction.Commit() below (issue 1800)
    await using var conn3 = new SqlConnection(connString);
    await conn3.OpenAsync();
    await conn2.CloseAsync();
    conn3.EnlistTransaction(transaction);

    Console.WriteLine("Committing transaction...");
    transaction.Commit();
    // Never gets printed due to deadlock.
    Console.WriteLine("Transaction committed.");
```

There may be 1% chance it may not repro, but even if it does 99% of the time, we still have a possibility of capturing it if anything changes in future. I'd recommend adding this to test case with useful comments. 🙂 

And once again, thank you for looking after this ❤️

roji:
OK, sounds good! Can you point me to where you guys put this kind of test? Or if you prefer, you can also just push a commit doing that to this PR.

cheenamalhotra:
You can add here: https://github.com/dotnet/SqlClient/tree/main/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/TransactionTest

roji:
@cheenamalhotra thanks, I've pushed the test. Some notes:

* The test is under NET7_0_OR_GREATER, but the SqlClient tests currently target net6.0, so it's never run by default. I hacked around the project system and confirmed that when targeting net7.0, the test fails without the fix and passes with it. I'd recommend targeting net7.0 in the tests (possibly in addition to net6.0), but that's beyond the scope of this PR.
* Your proposed test code above - with pooling disabled and a 3rd connection - actually passed without the fix; however, my original code sample (with pooling, two connections) very reliably failed without the fix and passed with it. So I just pushed that.
* Just a note: various other tests under ManualTests/TransactionTest failed for me, I'm hoping that's expected (I'm guessing these aren't run as part of your CI build, hence the name ManualTests?)

@David-Engel and others, I really hope this can be patched in 5.0; the bug effectively prevents people from using distributed transactions with SqlClient on .NET 7 (the primary use case); in EF we won't be able to take a dependency on SqlClient 5.1 (7.0 is out, and patch releases generally can only bump patch versions of dependencies).

lcheunglci (approved):

lcheunglci:
Since we're in the middle of releasing the 5.1 GA, I'll need to wait for approval from @David-Engel before we can merge this.

DavoudEshtehari:
@roji Adding support .NET 7.0 to the pipelines is on our plan, but it has the known issue [#4014](https://github.com/microsoft/vstest/issues/4014) on forwarding MSBuild's properties using `dotnet test` command, which is a blocker.

DavoudEshtehari (approved):

JRahnama:
@roji failures are not related to your changes. I have added windows 11 and apparently there is a failure due to some access restriction for Namedpipes. I am addressing them.

David-Engel (approved):

Issue #1675, opened 2022-07-20
TransactionAbortedException when performing queries in parallel inside a transaction scope


joostmeijles:
### Describe the bug
When performing multiple queries in parallel, and each pair of 2 queries are inside a transaction scope, an unexpected transaction error is thrown (see below).

```
Exception message: System.Transactions.TransactionAbortedException: The transaction has aborted.
Stack trace:
Unhandled exception. System.Transactions.TransactionAbortedException: The transaction has aborted.
 ---> System.Transactions.TransactionPromotionException: Failure while attempting to promote transaction.
 ---> System.InvalidOperationException: The requested operation cannot be completed because the connection has been broken.
   at Microsoft.Data.SqlClient.SqlInternalConnectionTds.ExecuteTransaction(TransactionRequest transactionRequest, String name, IsolationLevel iso, SqlInternalTransaction internalTransaction, Boolean isDelegateControlRequest)
   at Microsoft.Data.SqlClient.SqlDelegatedTransaction.Promote()
   --- End of inner exception stack trace ---
   at Microsoft.Data.SqlClient.SqlDelegatedTransaction.Promote()
   at System.Transactions.TransactionStatePSPEOperation.PSPEPromote(InternalTransaction tx)
   at System.Transactions.TransactionStateDelegatedBase.EnterState(InternalTransaction tx)
   --- End of inner exception stack trace ---
   at System.Transactions.TransactionStateAborted.CheckForFinishedTransaction(InternalTransaction tx)
   at System.Transactions.EnlistableStates.Promote(InternalTransaction tx)
   at System.Transactions.Transaction.Promote()
   at System.Transactions.TransactionInterop.ConvertToDistributedTransaction(Transaction transaction)
   at System.Transactions.TransactionInterop.GetExportCookie(Transaction transaction, Byte[] whereabouts)
   at Microsoft.Data.SqlClient.SqlInternalConnection.GetTransactionCookie(Transaction transaction, Byte[] whereAbouts)
   at Microsoft.Data.SqlClient.SqlInternalConnection.EnlistNonNull(Transaction tx)
   at Microsoft.Data.ProviderBase.DbConnectionInternal.ActivateConnection(Transaction transaction)
   at Microsoft.Data.ProviderBase.DbConnectionPool.PrepareConnection(DbConnection owningObject, DbConnectionInternal obj, Transaction transaction)
[...]

PR #1814, opened 2022-10-24
Improved general connection pool

DavoudEshtehari:
While investigating issue #1675, I noticed that the underlying data structure used at `DBConnectionPool` for storing internal connection objects is a `Stack`, which won't be the best choice in a high-demand multi-thread application.


Wraith2:
Imagine a situation where you have an application which uses a large number of connections for a short time meaning that the pool has a relatively high population and then the app calms down to a steady state there it uses connections more slowly.

In this case with a stack a small number of connections will be pushed and popped off the stack regularly and then the other items will get older and eventually be removed by the pool cleanup because they're inactive.

In this case with a queue all connections are kept alive. You will retain all of the items in the pool not aging any out (depending on the number of items in the pool) so pools will be more highly populated.

David-Engel:
> Imagine a situation where you have an application which uses a large number of connections for a short time meaning that the pool has a relatively high population and then the app calms down to a steady state there it uses connections more slowly.
> 
> In this case with a stack a small number of connections will be pushed and popped off the stack regularly and then the other items will get older and eventually be removed by the pool cleanup because they're inactive.
> 
> In this case with a queue all connections are kept alive. You will retain all of the items in the pool not aging any out (depending on the number of items in the pool) so pools will be more highly populated.

I've been looking at [CleanupCallback()](https://github.com/dotnet/SqlClient/blob/v5.1.0-preview1/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/ProviderBase/DbConnectionPool.cs#L571) trying think about your scenario and I don't think that's how it works. The pool pruning logic doesn't care when a connection was last used. If it is idle  in the pool when the method runs, it is subject to cleanup regardless of when it was last used. Yes, using a stack (as-is) will favor keeping the last used connection as the most used and least likely to be cleaned up, but changing to a queue won't keep idle connections from being pruned.. It will just more evenly distribute connection use across all of the items in the pool.

Wraith2:
It has been some time since I tried to understand the connection pool logic and it's very possible that I misunderstood the structure. I thought this was how it worked. Since you've looked through the code with that scenario in mind and say it's fine then I'm ok with the change. I just wanted to flag up what I thought might be the reason for the stack.

roji:
@DavoudEshtehari am on vacation at the moment, and it's once I'm back it's very unlikely I'll be able to look at this before .NET Conf (at least November 10th). I'll happy to take a look afterwards though, if that's not too far away.

DavoudEshtehari:
@roji thank you for the prompt reply. I wish you the best with .NET Conf, and I hope we could gain the benefit of your thoughts as well.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/ProviderBase/DbConnectionPool.cs
```diff
@@ -1608,7 +1608,7 @@ internal void PutObjectFromTransactedPool(DbConnectionInternal obj)
             // done and all transactions are ended.
             SqlClientEventSource.Log.TryPoolerTraceEvent("<prov.DbConnectionPool.PutObjectFromTransactedPool|RES|CPOOL> {0}, Connection {1}, Transaction has ended.", ObjectID, obj.ObjectID);
 
-            if (_state == State.Running && obj.CanBePooled)
+            if (_state == State.Running && obj.CanBePooled /*&& obj.EnlistedTransaction is null*/)
```
Tornhoof, line 1611: you might want to remove the commented part again.

DavoudEshtehari, line 1611:
```suggestion
            if (_state == State.Running && obj.CanBePooled)
```

cheenamalhotra:
@DavoudEshtehari have you measured performance by run parallel testing with these changes? I tried but I saw the current implementation being much more performant.

Here's my repro: [Test SqlConnection.OpenAsync performance](https://gist.github.com/cheenamalhotra/1441826f98b0158df5d08969df0788eb). Feel free to modify and add output per connection attempt, and the difference would be visible easily when connecting to Azure SQL.

This is an important use case I would recommend keeping close and not regressing.

cheenamalhotra:
Looking at the changes, I don't think this should make any significant difference, negative or positive, so you may disregard my last comment. Slight movement in results is possible due to network jitter so that should be fine.

Stacks should be preferred for pool when you're looking at high usability, such that when accessing a pool element, it need not move other items around unlike queue, so time taken should be less to fetch/add an element in a heavy pool, I think. Applications creating and pooling large number of connections would most likely notice that, so I would be considerate of that.

I would like to understand why you think changing it to queue/bag would benefit a multi-threading app; in case I'm missing a logical reasoning?

DavoudEshtehari:
> Stacks should be preferred for pool when you're looking at high usability, such that when accessing a pool element, it need not move other items around unlike queue, so time taken should be less to fetch/add an element in a heavy pool, I think. Applications creating and pooling large number of connections would most likely notice that, so I would be considerate of that.

To compare the Stack and Queue, I don't have internal implementation knowledge of them. So, I put it for those who have such ability. Even though they don't have significant perf differences, I believe the benefit of using it on a multi-threaded app can be beneficial.
 
> I would like to understand why you think changing it to queue/bag would benefit a multi-threading app; in case I'm missing a logical reasoning?

When I was running the [repro](https://github.com/dotnet/SqlClient/issues/1675#issue-1311193734), I noticed a reduction in the number of failures, plus a difference in the stack trace. You can find it by referring to the [originally reported](https://github.com/dotnet/SqlClient/issues/1675#issue-1311193734) stacktrace and the [result](https://github.com/dotnet/SqlClient/issues/1675#issuecomment-1290233601) of this change.
Also, from the logs, I found the failure occurs when a connection's taken by a new thread that was owned by another thread before returning to the pool.

roji:
This is something we looked into in Npgsql a while ago as well. A long time ago, the pool was implemented as a stack to promote good pruning (for the same reasoning @Wraith2 explains in [this comment](https://github.com/dotnet/SqlClient/pull/1814#issuecomment-1289822596)). However, we ended up moving away from that, and not relying on the data structure for pruning. In other words, the current pruning algorithm samples the idle/busy state several times, and if it determines that pruning should occur, closes *any* idle connection (we don't care whether it's one that happened to get handed out recently or not).

The current Npgsql implementation uses [System.IO.Channels](https://devblogs.microsoft.com/dotnet/an-introduction-to-system-threading-channels/), which is a conceptual queue that always manages efficient handing off of elements (connections) across threads, as well as efficiently waiting when the queue is empty (i.e. Max Pool Size has been reached).

In any case, I'd highly recommend measuring perf via benchmarking before making any actual changes. Would also be happy to continue talking about this topic (pooling implementation is very interesting and not at all trivial).

Issue #1810, opened 2022-10-20
CancellationToken registration leak, regression


tlecomte:
### Describe the bug

After updating to SqlClient 5.0.1, we observe a memory leak in our application. Memory dumps show a huge amount of CancellationTokenSource+CallbackNode entries (we had a dump with 2 GB of those), referencing SqlCommand objects.

We believe this might be due to the changes in:
- https://github.com/dotnet/SqlClient/pull/1785 for 5.0.1 backport
- https://github.com/dotnet/SqlClient/pull/1781 for 5.1.0

With the changes in these PRs, it seems the CancellationToken `registration` is not always disposed. There are several `if` blocks that may return before the `registration` object gets to be saved in the `context`. And exceptions could also be thrown.

### To reproduce

```csharp
using Microsoft.Data.SqlClient; // 5.0.1
using System.Threading;
using System.Threading.Tasks;

public static class Program
{
    public static async Task Main()
    {
        const string query = "SELECT 1";

        // this could be the application lifetime cancellation
        // or another form of global shared cancellation
        using var longRunningCts = new CancellationTokenSource();

        var connectionString = new SqlConnectionStringBuilder
        {
            DataSource = "(local)",
            IntegratedSecurity = true,
            TrustServerCertificate = true,
        }.ToString();

        try
        {
            using (var linkedCts = CancellationTokenSource.CreateLinkedTokenSource(longRunningCts.Token))
            using (var connection = new SqlConnection(connectionString))
            {
                await connection.OpenAsync(linkedCts.Token);

                using (var command = new SqlCommand(query, connection))
                using (var reader = await command.ExecuteReaderAsync(linkedCts.Token))
                {
                    linkedCts.Cancel();

                    while (await reader.ReadAsync(linkedCts.Token))
                    {

                    }
                }
            }
        }
        catch (Exception e)
        {

        }

        // => attach a debugger here
        // linkedCts as well as all the SQL objects have been disposed above
        // however a memory snapshot shows that the CancellationTokenSource+Registrations and CancellationTokenSource+CallbackNode are still there, including the objects they reference
        // so garbage collection cannot happen correctly
    }
}
```

### Expected behavior

No memory leak, CancellationTokenSource+CallbackNode entries are not present in a dump.

### Further technical details

Microsoft.Data.SqlClient version: 5.0.1
.NET target: .NET 6.0.3
SQL Server version: SQL Server 2019
Operating system: Windows 2019

**Additional context**
 -

[...]

PR #1818, opened 2022-10-25
Add | Adding disposable stack temp ref struct and use

Wraith2:
fixes https://github.com/dotnet/SqlClient/issues/1810

Adds a mutable ref struct to hold disposable things. The cancellation token registration is created and put into the struct. If the code reaches a Take method the value is taken by the caller. If the finally block of the using scope is reached while the value is still held the value will be disposed. This ensures that either the registration is passed to the context object or is correctly disposed.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
@@ -5386,9 +5386,6 @@ namespace Microsoft.Data.SqlClient
             internal override Func<Task, object, Task<T>> Execute => s_execute;
         }
 
-
-
-
         /// <summary>
         /// Starts the process of executing an async call using an SqlDataReaderAsyncCallContext derived context object.
         /// After this call the context lifetime is handled by BeginAsyncCall ContinueAsyncCall and CompleteAsyncCall AsyncCall methods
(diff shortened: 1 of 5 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
```diff
@@ -5382,6 +5386,9 @@ protected override void Clear()
             internal override Func<Task, object, Task<T>> Execute => s_execute;
         }
 
+
+
+
```
David-Engel, line 5391: Nit:
```suggestion
```

David-Engel (approved):

lcheunglci (approved):
Thanks @Wraith2 , LGTM

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
Wraith2: Removed.

JRahnama:
@Wraith2 that failure on Mac machine is a bit concerning. Probably if I rerun it that will pass, but this test should not fail randomly and it seems critical to me.

Wraith2:
It was all green before I removed some spaces...

The error is in marc concurrency and despite what the message says, that it shouldn't fail randomly, concurrency tests are hard to write and keep stable. Nothing in this change set should have any effect on concurrency or mars.

JRahnama:
We should spend some time in near future look into Active issues, some of them might not be an issue anymore, and also find a solution for these kinds of intermittent failures.

PR #1821, opened 2022-10-31
TDS8 - Enable retrieval of TLS 1.3 SSL Protocol from SNI on .NET Core

lcheunglci:
I ran into a timeout issue during `WaitForSSLHandShakeToComplete` when I ran .NET Core with TLS 1.3 and noticed a commented-out section where it checks for TLS 1.3 and returning `SslProtocols.None`.  However, I ran into a compile error when uncommenting a check for the ssl protocol while testing .NET Core with TLS1.3 when building the MDS project on .NET standard 2.1 because `SslProtocols.Tls13` does not exist.  The `SslProtocols.Tls13` does exist on .NET Core 3.1 and above, so I added an `if-def` for it, so that I can get past the timeout issue where the retrieval of the protocol version occurs.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObjectNative.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObjectNative.cs
@@ -425,12 +425,16 @@ namespace Microsoft.Data.SqlClient
             var nativeProtocol = (NativeProtocols)nativeProtocolVersion;
 
             /* The SslProtocols.Tls13 is supported by netcoreapp3.1 and later
-             * This driver does not support this version yet!
+             * This driver does not support this version yet! */
+#if NETCOREAPP3_1_OR_GREATER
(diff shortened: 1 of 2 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObjectNative.cs
```diff
@@ -425,12 +425,16 @@ internal override uint WaitForSSLHandShakeToComplete(out int protocolVersion)
             var nativeProtocol = (NativeProtocols)nativeProtocolVersion;
 
             /* The SslProtocols.Tls13 is supported by netcoreapp3.1 and later
-             * This driver does not support this version yet!
+             * This driver does not support this version yet! */
+#if NETCOREAPP3_1_OR_GREATER
```
JRahnama, line 429: Since we are not supporting netcoreapp3.1 it should suffice if we just want to check for NETCOREAPP and omit the version.
```suggestion
#if NETCOREAPP
```

DavoudEshtehari (approved):

JRahnama (approved):

PR #1824, opened 2022-11-02
Change | Revert excluding unsupported protocols

DavoudEshtehari:
Regarding issue #1151 and related PR #1168, the proposed changes are not required anymore since Windows 10 and/or SQL Server 2019 have fixed the issue while Microsoft.Data.SqlClient has been adding support for TLS 1.3.

- Note: Native SNI changes are proposed in the internal [Pull Request 3527](https://sqlclientdrivers.visualstudio.com/ADO.Net/_git/Microsoft.Data.SqlClient.sni/pullrequest/3527): Revert PR 2947 Remove UseSystemDefaultSecureProtocols

PR #1828, opened 2022-11-04
Experimental: Add ARM64 support when targeting .NET Framework

David-Engel:
The CI pipelines don't run tests on ARM64 + .NET Framework and I'm having a hard time getting an ARM64 Windows VM to build + run MDS tests from the command line with VS 2022 Preview 6 installed.

Test can be run through the VS GUI, though, you just have to manually filter through expected failures. Doing that identified a bug in the build options of the ARM SNI binaries that resulted in asserts popping up even when using the release binaries. That will be addressed separately in a future SNI release.

Consider this "feature" experimental, for now.


src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/Interop/SNINativeManagedWrapperX64.cs
```diff
@@ -89,7 +89,7 @@ internal static class SNINativeManagedWrapperX64
         internal static extern uint SNIGetInfoWrapper([In] SNIHandle pConn, SNINativeMethodWrapper.QTypes QType, out ProviderEnum provNum);
 
         [DllImport(SNI, CallingConvention = CallingConvention.Cdecl, EntryPoint = "SNIInitialize")]
-        internal static extern uint SNIInitialize([In] IntPtr pmo);
+        internal static extern uint SNIInitialize([In] bool useSystemDefaultSecureProtocols, [In] IntPtr pmo);
```
David-Engel, line 92: This is a partial revert of the changes in #1824. The native SNI interface here still has 2 parameters and I don't know why things work in x86/x64 with only 1. Without this revert, I get an assert when running in ARM64 mode because pmo is not what the native SNIInitialize function expects. Oddly, I _don't_ get an assert when running the exact same code in x86 or x64 mode, which explains why the change passed CI, but not why it doesn't assert.

lcheunglci: I believe @DavoudEshtehari found it odd that the SNI revert for the `useSystemDefaultSecureProtocols` was passing on the pipeline before the new revert SNI changes was available as it won't be available until the next SNI preview release.

lcheunglci: We can create a new PR with the sni version bump to reapply the removal of `useSystemDefaultSecureProtocols`.

David-Engel: To explain the reason things "work" in x86 and x64 tests, the pmo parameter has a default value of NULL in the SNIInitialize method so there was no interface issue. Since we hard code pmo to 0 in MDS, the previous change just ended up sending false/0 for useSystemDefaultSecureProtocols and pmo became NULL. In SNI, pmo has an assert(pmo == 0) on it but asserts are suppressed in release x64/x86 builds and pmo doesn't actually do anything in our SNI code paths (we could remove it completely) so there was no negative effect. The SNI arm release builds were missing the NDEBUG setting so asserts were firing and popping up windows.

DavoudEshtehari (approved):

JRahnama (approved):

Matteo-T:
Nice! When can I expect this to show up in MDS3.x? Pretty please... :)

PR #1843, opened 2022-11-21
Merge more methods from TdsParserStateObject 

bhugot:
more merge after #1520 It would maybe help to start working on fix the async bug.
I think we should be able to do more.


panoskj:
@bhugot @JRahnama as I hinted in #1520 I have also merged about 2000 more lines of this file. I was actually about to make a PR about it.

I can't review this right now, but at a glance it looks like you have made changes to more files than I did, but I have merged a little bit more of the TdsParserStateObject file. My version also splits the merge in smaller commits too.

So, I will make the PR and leave it up to you to decide what to do with it.

bhugot:
@panoskj to bad didn't see your branch on this on your repo. I think you started the move and so your are more legit to it. If that can help I started moving TdsParser the same way. So after #1844 merge I could do the PR for TdsParser

panoskj:
> @panoskj to bad didn't see your branch on this on your repo. 

@bhugot my bad, I should have added a link to it at least. 

Anyway, I was trying to compare your version to mine, but it isn't so easy. It looks like we have a different position/order for some methods, which messes up the diff viewer. But I can see we have merged some different parts of the file (e.g. you have merged more snapshot related methods, but still I have merged more code in total).

We could merge both our versions, but since these PRs are already huge, it is probably best to go for a third PR along with the rest of the file. The important methods for fixing the async bug are already merged by both PRs anyway.

Thanks for the feedback on my PR by the way, I will check the code and come back with answers later.

PR #1536, opened 2022-03-07
Remove package references to runtime and ref pack files when targeting .NET 6+.

AraHaan:
In .NET 6+ minimize network load on restore by using the ones from the ref and runtime packs over the package versions.

PR #1849, opened 2022-11-24
Remove redundant NuGet package references.

teo-tsirpanis:
This PR removes dependencies to NuGet packages that are not needed for each framework we target. Some packages are useless for anything other than .NET Standard 1.x and cause an explosion of transitive dependencies (such as `System.IO`), others provide functionality that became inbox in newer frameworks (such as `System.Buffers` for .NET Standard 2.1 and .NET 6, or `Microsoft.Win32.Registry` for .NET 6), and others were unused (such as `System.Security.Cryptography.Algorithms` which provides `IncrementalHash` on older .NET Framework versions which is not used by this library).

I didn't change any executable code, if CI passes it should be OK. Let's see if I missed anything.

Fixes NuGet/Home#12240.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
@@ -950,7 +950,7 @@
     </EmbeddedResource>
   </ItemGroup>
   <ItemGroup>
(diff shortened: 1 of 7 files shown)


teo-tsirpanis:
Overlaps with #1536; this one is simpler and more targeted.

DavoudEshtehari (approved):

lcheunglci (approved):
LGTM, thanks for your contributions

PR #1860, opened 2022-12-04
Switched to SDK project.

bhugot:
In order to ease merging netfx and netcore I transformed the netfx project to use sdk project. 
Also auto included files from common part but some files to ignore. 
Suffixed files from Common with .common in the link, but maybe you would prefer I remove this.


bhugot:
I plan to do the same auto include from common for netcore project.

ErikEJ:
@JRahnama Has this not happened in the meantime?

JRahnama:
> @JRahnama Has this not happened in the meantime?

netfx in converted to be SDK style in PR #2413, code merge is still in progress. Changed could be traced at #1261

Issue #1864, opened 2022-12-07
Heavy GC thrashing in TdsParser


masonwheeler:
### Describe the bug
While running and processing a large (multi-GB) query, the debugger shows the GC running almost constantly.  After reducing allocations as much as possible in my code, I ran it through JetBrains DotMemory, and it said that the process was allocating almost 5 GB of `char[]` buffers in `TdsPaser.TryReadPlpUnicodeChars`.  The stack trace it gave shows that this is being called as part of reading data from a result set, apparently to read a string buffer and turn it into a string.

Would it be possible to cache and reuse the char buffer rather than (I assume; the relevant code doesn't appear to be in this repo) allocating a new buffer for each TryReadPlpUnicodeChars invocation?

### To reproduce

* Grab and restore [this StackOverflow data dump](https://downloads.brentozar.com/StackOverflow2013_201809117.7z).
* Create a `SqlConnection` and open a connection to the database.
* Run the following:
```
void Process(SqlConnection conn)
{
	using var cmd = conn.CreateCommand();
	cmd.CommandText = "select * from dbo.Posts";
	using var reader =  cmd.ExecuteReader();
	var buffer = new object[reader.FieldCount];
	while (reader.Read())
	{
		reader.GetValues(buffer);
	}
}
```

### Expected behavior
As we're processing one row at a time, memory usage should remain fairly low.

### Observed behavior:
Constant GC pressure.  Profiling shows the vast majority of it comes from char buffer allocations as described above.

### Further technical details
Microsoft.Data.SqlClient version: nuget 4.1.0
.NET target: .NET 7
SQL Server version: SQL Server 15.0.2095.3
Operating system: Windows 10

lcheunglci:
Thanks for bringing this to our attention. We'll get back to you soon.

Wraith2:
I'm not sure a 10Gib compressed dataset counts as a "minimal" reproduction ;)

masonwheeler:
Fair enough.  This is the testing dataset I was using when I ran across this problem; I used it here because it clearly shows that this is a big, serious problem when working with large amounts of data.  Thing is, that's something that can't easily be demonstrated *without* actually working with a large amount of data, so it's admittedly a bit in conflict with the notion of a simple, minimal reproduction.

Honestly, any dataset that contains a lot of VARCHARs would show the issue.

Wraith2:
[...]

PR #1866, opened 2022-12-09
Add array rental capability in TryReadPlpUnicodeChars

Wraith2:
fixes https://github.com/dotnet/SqlClient/issues/1864

This changes TryReadPlpUnicodeChars in two ways. 
1) it adds the ability to specify that the calling context can support a rented char[] buffer and adds support for using the arraypool inside the function
2) it adds a check for the int.max/2  which is used as a sentinel value in cases where the length is not yet known. Adding this check prevents the attempted use of very very large array pool buffers for small strings and defers the rental of the buffer until the length calculation has been made.

The outcome of this is that in general use the arraypool will be used no intermediate buffers will be dropped to the GC. This does some with a minor performance loss in CPU time spend dealing with the array pool rental mechanics and clearing. See the fixed thread for the details on that. The perf loss is small enough that I think https://github.com/dotnet/SqlClient/pull/1544 will probably make up for it.

/cc @masonwheeler

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
@@ -5684,8 +5684,8 @@ namespace Microsoft.Data.SqlClient
                         if (isPlp)
                         {
                             char[] cc = null;
-
-                            if (!TryReadPlpUnicodeChars(ref cc, 0, length >> 1, stateObj, out length))
+                            bool buffIsRented = false;
+                            if (!TryReadPlpUnicodeChars(ref cc, 0, length >> 1, stateObj, out length, supportRentedBuff: true, rentedBuff: ref buffIsRented))
                             {
                                 return false;
                             }
@@ -5696,6 +5696,16 @@ namespace Microsoft.Data.SqlClient
                             else
                             {
                                 s = "";
+                            }
+                            if (buffIsRented)
+                            {
+                                // do not use clearArray:true on the rented array because it can be massively larger
+                                // than the space we've used and we would incur performance clearing memory that
+                                // we haven't used and can't leak out information.
+                                // clear only the length that we know we have used.
+                                cc.AsSpan(0, length).Clear();
+                                ArrayPool<char>.Shared.Return(cc, clearArray: false);
+                                cc = null;
                             }
                         }
                         else
@@ -12512,8 +12522,9 @@ namespace Microsoft.Data.SqlClient
         internal int ReadPlpUnicodeChars(ref char[] buff, int offst, int len, TdsParserStateObject stateObj)
         {
             int charsRead;
+            bool rentedBuff = false;
             Debug.Assert(stateObj._syncOverAsync, "Should not attempt pends in a synchronous call");
-            bool result = TryReadPlpUnicodeChars(ref buff, offst, len, stateObj, out charsRead);
+            bool result = TryReadPlpUnicodeChars(ref buff, offst, len, stateObj, out charsRead, supportRentedBuff: false, ref rentedBuff);
             if (!result)
             {
                 throw SQL.SynchronousCallMayNotPend();
@@ -12525,8 +12536,9 @@ namespace Microsoft.Data.SqlClient
         // requested length is -1 or larger than the actual length of data. First call to this method
         //  should be preceeded by a call to ReadPlpLength or ReadDataLength.
         // Returns the actual chars read.
-        internal bool TryReadPlpUnicodeChars(ref char[] buff, int offst, int len, TdsParserStateObject stateObj, out int totalCharsRead)
+        internal bool TryReadPlpUnicodeChars(ref char[] buff, int offst, int len, TdsParserStateObject stateObj, out int totalCharsRead, bool supportRentedBuff, ref bool rentedBuff)
         {
+            bool inputWasRentedBuff = rentedBuff;
             int charsRead = 0;
             int charsLeft = 0;
             char[] newbuf;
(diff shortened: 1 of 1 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -12525,8 +12536,9 @@ internal int ReadPlpUnicodeChars(ref char[] buff, int offst, int len, TdsParserS
         // requested length is -1 or larger than the actual length of data. First call to this method
         //  should be preceeded by a call to ReadPlpLength or ReadDataLength.
         // Returns the actual chars read.
-        internal bool TryReadPlpUnicodeChars(ref char[] buff, int offst, int len, TdsParserStateObject stateObj, out int totalCharsRead)
+        internal bool TryReadPlpUnicodeChars(ref char[] buff, int offst, int len, TdsParserStateObject stateObj, out int totalCharsRead, bool supportRentedBuff, ref bool rentedBuff)
         {
+            bool inputWasRentedBuff = rentedBuff;
```
campersau, line 12541: Looks like this is not used?
```suggestion
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -12538,16 +12550,31 @@ internal bool TryReadPlpUnicodeChars(ref char[] buff, int offst, int len, TdsPar
...
+            if (buff == null && stateObj._longlen != TdsEnums.SQL_PLP_UNKNOWNLEN && len < (int.MaxValue >> 1))
             {
-                buff = new char[(int)Math.Min((int)stateObj._longlen, len)];
+                if (supportRentedBuff && len < 1073741824) // 1 Gib
+                {
+                    if (buff != null && rentedBuff)
+                    {
+                        buff.AsSpan(0, offst).Clear();
+                        ArrayPool<char>.Shared.Return(buff, clearArray: false);
+                    }
```
campersau, line 12569: `buff` is always `null` here
```suggestion
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
Wraith2: Yes, this got moved into the loop. Removed.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
Wraith2: It isn't. If the pre loop check finds that that length was passed in as unknown and the caller hasn't provided a buffer then it will hit this piece of code and then rent/allocate a correctly sized buffer once it has calculated the required size. 

This is now the desired path because otherwise the pre-loop rental/allocate will create a `char[int.MaxValue/2]` size buffer which is usually far far too large and not something we want to keep around in the array pool because it will contribute to memory pressure.

campersau: Hm not sure if I am missing something but two lines above there is this check
`if (buff == null && stateObj._longlen != TdsEnums.SQL_PLP_UNKNOWNLEN && len < (int.MaxValue >> 1))`
which means that `buff` can't be `null` here or?
Or is the `buff == null` check wrong?

Wraith2: Ah, my mistake, I thought this was about the loop version. It's the preloop version. Yes, looks like it no longer applies,

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -12538,16 +12549,26 @@ internal bool TryReadPlpUnicodeChars(ref char[] buff, int offst, int len, TdsPar
...
 
-            // If total length is known up front, allocate the whole buffer in one shot instead of realloc'ing and copying over each time
-            if (buff == null && stateObj._longlen != TdsEnums.SQL_PLP_UNKNOWNLEN)
+            // If total length is known up front, the length isn't specified as unknown 
+            // and the caller doesn't pass int.max/2 indicating that it doesn't know the length
+            // allocate the whole buffer in one shot instead of realloc'ing and copying over each time
+            if (buff == null && stateObj._longlen != TdsEnums.SQL_PLP_UNKNOWNLEN && len < (int.MaxValue >> 1))
             {
-                buff = new char[(int)Math.Min((int)stateObj._longlen, len)];
+                if (supportRentedBuff && len < 1073741824) // 1 Gib
```
madelson, line 12562: Is there ever a case where the buffer will be so short that renting won't be worthwhile?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -5697,6 +5697,16 @@ private bool TryReadSqlStringValue(SqlBuffer value, byte type, int length, Encod
...
                             }
+                            if (buffIsRented)
+                            {
+                                // do not use clearArray:true on the rented array because it can be massively larger
+                                // than the space we've used and we would incur performance clearing memory that
+                                // we haven't used and can't leak out information.
+                                // clear only the length that we know we have used.
+                                cc.AsSpan(0, length).Clear();
+                                ArrayPool<char>.Shared.Return(cc, clearArray: false);
+                                cc = null;
```
madelson, line 5708: Is this line necessary? Isn't cc about to go out of scope anyway?

Wraith2: It isn't strictly necessary in the current configuration. Imagine a future refactoring which hoists cc to an outer scope, with this in place it'll be cleared correctly in future. It costs very little.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
Wraith2: How would you determine the answer to this question? Is it ever cheaper to new and GC an array than the overhead of renting? Possibly, I'd expect it to be a very low difference and almost impossible to quantify without a specific weird use case that is pathalogical along that route.

madelson: My understanding from other articles/comments about `ArrayPool` is that it isn't necessarily worthwhile for small arrays (on my machine it takes ~3x longer for arrays of length 10 for example). 

Not sure if that matters here or not. You're probably correct that it won't be the bottleneck.

Wraith2: In terms of speed you're right. Speed isn't really a limiting factor for this library at the moment, network delay is. In general this will has a small speed penalty that would be quantifiable in a microbenchmark but very unlikely to become a bottleneck, it'll have an overall improvement in reducing GC'ed objects which will allow more time to be spend in user code then GC. 

If a pathalogical case turns up then we can determine a more complex solution to avoid it.

David-Engel (approved):

PR #1880, opened 2023-01-06
Draft | Addressing name pipe failures on sync and async calls.

JRahnama:
While working on some intermittent failures on test pipelines on, concurrent test runs for retry logic, I noticed there is  `PipeOptions.Asynchronous ` in[ `SNINPHandle` class ctr](https://github.com/dotnet/SqlClient/blob/f0f2cbc54c6f8690d3e9cd87d3d7cb05c7bf0940/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNINpHandle.cs#L51). That causes the tests to fail at some random runs. There is an option to pass a variable to decide on sync/async calls, but that approach needs more investigation as the value of async passed down to other functions in `TDSParser.Connect` for the functionc call [`_physicalStateObj.CreatePhysicalSNIHandle`](https://github.com/dotnet/SqlClient/blob/f0f2cbc54c6f8690d3e9cd87d3d7cb05c7bf0940/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs#L446) is set to be alwyas true. 

My main concern with this change: what would be the effect of this on async calls that happen after openAsync? will taking out the pipeoptions for asynchronous calls will cause any damage on the async pattern, tasks or threads?

on some other notes:

In the very same class `SNINPHandle` on the [Receive](https://github.com/dotnet/SqlClient/blob/f0f2cbc54c6f8690d3e9cd87d3d7cb05c7bf0940/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNINpHandle.cs#L185) function when the packet length is 0, stream is closed by server, the driver gets the last Win32Exception and throws that as exception. This does not seem like a very good approach to me. There could be some other problem happening in between. For example, `There is no process at the other end of the pipe` when a wrong password is used. How other drivers have solved this issue?

To repro the issue make sure youa re using named pipes and managed SNI is in use. The test that fails intermittently is:  
[ConcurrentExecutation in SqlConnectionReliablityTest](https://github.com/dotnet/SqlClient/blob/f0f2cbc54c6f8690d3e9cd87d3d7cb05c7bf0940/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/RetryLogic/SqlConnectionReliabilityTest.cs#L103)
cc: @roji, @Wraith2 , @David-Engel , @DavoudEshtehari


JRahnama:
while we are at this there is [PipeSecurity ](https://learn.microsoft.com/en-us/dotnet/api/system.io.pipes.pipesecurity.-ctor?view=net-7.0) class which was available in netcoreapp 1.1 and it came back again in net6 and net7. We set ACL in native code, but not in netcore, Discretionary Access Control List (DACL). Something that needs to be considered in my opinion.

Wraith2:
> There is an option to pass a variable to decide on sync/async calls

This is fairly common and much better than attempting to use a heuristic approach. It can be annoying to write but as long as you're only working within the managed implementation and don't need to change the interface used by native it should just turn out to be code plumbing.

JRahnama:
@Wraith2 so you are pro passing a flag from inside OpenAsync and pass it all the way here? Actually, there is a flag, but mid-way its value is set to be true all the time.

Wraith2:
It seems like a reasonable approach. It is at least worth trying.

DavoudEshtehari:
Temporary disabled test [ConcurrentExecution](https://github.com/dotnet/SqlClient/blob/d5df671376b3114e8f7adc570b95073b796b6e16/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/RetryLogic/SqlConnectionReliabilityTest.cs#L101C8-L101C31)

David-Engel:
@JRahnama I think the problem with the test is that it is simply calling async methods in the context of sync methods and we are starving the thread pool. I can repro this locally by increasing the concurrentExecution number. If you SetMinThreads to >= concurrentExecution, the test succeeds. Remember, the default system thread count is # of CPU cores and our test VMs only have 2 cores. Another option is probably to split the test into separate sync and async versions and convert the async one to fully async. But even that option isn't certain with our sync over async issues.

PR #1885, opened 2023-01-11
Create sql

maqz9999:


lcheunglci:
I'm assuming this was created by mistake as it's an empty file and it doesn't seem to fix or address any opened issues.

PR #1891, opened 2023-01-18
Tests | Add HNIC and ServerCertificate connectivity tests for TDS8

lcheunglci:
The tests validate the `HostNameInCertificate` and the `ServerCertificate` functionality by connecting with various combinations of Encryption modes i.e. Encrypt=Mandatory, Encrypt=Optional, Encrypt=Strict with those parameters set.

This PR also includes a PowerShell script that creates a self-signed certificate and updates the SQL Server instance to reference it as a precondition before the TDS8 Connectivity Tests can run.

Note: that the `[ConditionalTheory]` attribute has some issue when I use the conditionals for detecting if the certificate has been installed properly, so I ended up writing an Assert for it.

In addition, only Sql Server 2022 is able to connect with `Encrypt=Strict` for the 2 tests that can successfully connect with HNIC and ServerCertificate.

I'll be attaching a link to the pipeline result in the comments with the test results shortly.

### DIFF
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/TDS8Test/Tds8ConnectivityTest.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/TDS8Test/Tds8ConnectivityTest.cs
@@ -12,16 +12,21 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests.TDS8
 
         // These environment variables are populated from a powershell or bash script.after the certificates are generated
         private const string ENV_CERT_FRIENDLYNAME = "TDS8_Test_Certificate_FriendlyName";
+        private const string ENV_CERT_MISMATCH_FRIENDLYNAME = "TDS8_Test_Certificate_Mismatch_FriendlyName";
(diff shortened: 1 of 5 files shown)


src/Microsoft.Data.SqlClient/tests/ManualTests/TDS8Test/Tds8ConnectivityTest.cs
```diff
@@ -0,0 +1,554 @@
...
+                DataSource = GetDataSourceName(dataSourceType),
+                Encrypt = strict ? SqlConnectionEncryptOption.Strict : SqlConnectionEncryptOption.Mandatory,
+                ServerCertificate = mismatchValidCertificatePath
+            };
+
+            if (strict && GetSqlServerMajorVersion() < 16)
+            {
+                // Connecting in Strict mode with Server Cerficiate is only available in SQL Server 2022; it's expected to fail lower SQL Server versions.
+                SqlException ex = Assert.Throws<SqlException>(() => Connect(builder.ConnectionString));
+                Assert.NotNull(ex);
```
DavoudEshtehari, line 526:
```suggestion
                Assert.Throws<SqlException>(() => Connect(builder.ConnectionString));
```

PR #1897, opened 2023-01-23
Draft | Addressing the issue with wrong exception when column decryption fails.

JRahnama:


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -2040,8 +2041,9 @@ internal bool TryRun(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataRead
                 {
                     case TdsEnums.SQLERROR:
                     case TdsEnums.SQLINFO:
+                    case TdsEnums.SQLDEBUG_CMD:
                         {
-                            if (token == TdsEnums.SQLERROR)
+                            if (token == TdsEnums.SQLERROR || token == TdsEnums.SQLDEBUG_CMD) // This the value of 0x60 send fromm TDS Control token
```
Wraith2, line 2046: I don't understand what the comment is trying to communicate here.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsEnums.cs
```diff
@@ -149,7 +149,7 @@ internal static class TdsEnums
         public const byte SQLDONEINPROC = 0xff;
         public const byte SQLOFFSET = 0x78;
         public const byte SQLORDER = 0xa9;
-        public const byte SQLDEBUG_CMD = 0x60;
+        public const byte SQLDEBUG_CMD = 0x60; // This needs more info
```
Wraith2, line 152: Who needs to provide the info? the SQL Server team? The spec?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
JRahnama: For the comment since this was a draft PR and I needed to make a change to run the pipelines I did not pay much attention to it.

The specs do not provide any info on that TDS token stream. At first, I did not notice that 0x60 is available in TDSEnums. basically, the issue was raised by PS team. They tried to provide an invalid token(AE token not TDS header token) to decrypt an encrypted column. Server returns 96 instead of error token (0xaa).

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsEnums.cs
JRahnama: I have to provide more info and add it here 😄 . This is just a replacement for more info and triggering pipelines.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
Wraith2: If we're getting it from the server then either they need to update the spec or stop sending it. I think updating the spec is easier.

lcheunglci:
Should we close this PR since the fix in https://github.com/dotnet/SqlClient/pull/1948 address this issue as well?

JRahnama:
> Should we close this PR since the fix in #1948 address this issue as well?

I forgot. I should close this.

Issue #1919, opened 2023-02-07
Upgrade Azure.Identity dependency to version 1.8.0


scottaddie:
In v1.8.0 of the Azure Identity library for .NET (_Azure.Identity_), token caching was added for the `ManagedIdentityCredential` type. Token lifetime and refreshing is handled automatically. See https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/identity/Azure.Identity/CHANGELOG.md#180-beta1-2022-10-13. Under the hood, MSAL's in-memory cache is used. This dependency upgrade will help mitigate Azure AD throttling issues that occur from calling it too frequently.

For Java customers, this issue has already been addressed as of the latest stable JDBC driver release. See https://github.com/microsoft/mssql-jdbc/issues/1930.

/cc: @David-Engel

ErikEJ:
@scottaddie So an explicit update to 1.8.0 will also add automagic caching?

lcheunglci:
Thanks @scottaddie for making the suggestion.  I do believe by upgrading this dependency it would also resolve https://github.com/dotnet/SqlClient/issues/1900 and we can consider upgrading it in our next release and I'll bring it up to the team and get back you too soon.

scottaddie:
> @scottaddie So an explicit update to 1.8.0 will also add automagic caching?

That's correct, @ErikEJ. Caching is enabled by default. No app code changes are required to use it.

lcheunglci:
Thanks @scottaddie for the suggestion and Azure Identity 1.8.0 would be included in the 5.2 release.

scottaddie:
@lcheunglci The following issue should also be resolved with this dependency upgrade: https://github.com/dotnet/SqlClient/issues/1495

Issue #1900, opened 2023-01-25
Issue with Microsoft.Data.SqlClient.dll looking for Azure.Identity.dll version 1.7.0 but project has a dependency on version 1.8.0


amitverma009:
Hi All,

Am facing issue with one of my windows  application which is using Microsoft.Data.SqlClient.dll  5.1.0 ( which has an internal dependency on azure.identity.dll ver >1.7.0).

There is a dependency on a nuget in application which is using Azure.Identity.dll 1.8.0.

All project dependencies shows dependency on version 1.8.0 and the same is getting packaged as part of MSI.

Issue : at runtime code is  looking for version 1.7.0 .

any help in understanding the exact cause would be helpful

Thanks 

Amit

lcheunglci:
Hi @amitverma009 , thanks for bringing this issue to our attention.  When we dropped support for .NET Core 3.1 and added support to .NET6, I believe we upgraded our dependency from [Azure Identity 1.6.0 to 1.7.0](https://github.com/dotnet/SqlClient/commit/20d4c199923c9b4ea2ffd44d9304fcb306c5efb5#diff-dcbc98937e117eea4b0ccd63fc36baf30e39fe3f2273699d144c59d5b5584cdb). I'll have to check with the team and see if we should upgrade dependency to the newer version or change the dependency to a min version instead of an exact version. We'll get back to you soon.

amitverma009:
Thanks @lcheunglci .

HamsterExAstris:
Is this Windows application .NET Framework or .NET Core?

If .NET Framework, does your MSI include an `app.config` file (with the same name as the executable) with a binding redirect for Azure.Identity?

amitverma009:
Yes It is in .NET Framework 4.7.2. I have done a binding redirect (manually ) but that is not a long term solution. 

Reason : There is an internal Nuget  

So the nuget is referred in a separate project (class libraray) which is not having any app config. App config is in the main project where I have put in binding redirect.

lcheunglci:
Can you provide us with a sample project to reproduce this? The dependency we have for Azure Identity is >= 1.7.0, so 1.8.0 shouldn't be an issue.

amitverma009:
Hi @lcheunglci  , am afraid I wont be able to provide the exact code. I have done the workaround by adding binding redirect.  Let me try and create another sample project so that i can share that with you.

inthemedium:
[...]

PR #1921, opened 2023-02-08
Chore | Upgrade Azure Identity from 1.7.0 to 1.8.0

lcheunglci:
This upgrade should resolve the following issues with the MSAL caching being enabled by default. 
https://github.com/dotnet/SqlClient/issues/1919
https://github.com/dotnet/SqlClient/issues/1900

Also upgraded some related dependencies.

### DIFF
--- a/tools/props/Versions.props
+++ b/tools/props/Versions.props
@@ -24,10 +24,10 @@
   </PropertyGroup>
   <!-- NetFx and NetCore project dependencies -->
   <PropertyGroup>
-    <AzureIdentityVersion>1.7.0</AzureIdentityVersion>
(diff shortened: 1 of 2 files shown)


ErikEJ:
Why do you need to be explicit about the dependencies of a Dependency??

lcheunglci:
The other dependencies aren't necessarily required in fact none of them has a direct dependency on Azure Identity. I was going the release notes and felt it might be a nice to have and it actually makes more sense that I revert them as they should be upgraded in a sperate PR when it becomes required. Thanks!

JRahnama (approved):

Kaur-Parminder (approved):

PR #1931, opened 2023-02-23
Enable SqlDiagnosticListener on .NET Standard

### LINKED ISSUE #1928: DiagnosticListener no longer works on .NET Core 3.1 or .NET 5
### Describe the bug

SqlClient 5.1.0 dropped support for .NET Core 3.1, and targeted .NET 6 instead.  However, there are still .NET Standard 2.0 and 2.1 targets, so applications running .NET Core 3.1 or .NET 5.0 can still install the newer version - but in doing so they'll get the .NET Standard version of SqlClient.

Unfortunately, the `SqlDiagnosticListener` class [is unimplemented for .NET Standard](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDiagnosticListener.NetStandard.cs).  Thus, if a .NET Core 3.1 or .NET 5.0 application updates to the latest version, they will silently lose all diagnostic messages from Sql Client at runtime.

We found this by a test breaking in our CI.  See https://github.com/getsentry/sentry-dotnet/issues/2189

### To reproduce

Make a .NET Core 3.1 or .NET 5 app that uses `Microsoft.Data.SqlClient` version 5.0.1 and subscribes to diagnostic listeners.  Perform any operation so you can see the diagnostic information.

Then update to `Microsoft.Data.SqlClient` version 5.1.0.  Notice there are no compilation errors, but the diagnostic data is no longer emitted.

### Expected behavior

Any of the following would have been acceptable:
- Diagnostic data still is emitted (requiring implementation of `SqlDiagnosticListener` for .NET Standard)
- The application didn't compile due to the breaking change (which would require some msbuild trickery)
- The version number of the release should have been 6.0.0 to indicate a significant breaking change.

Yes, I understand both .NET Core 3.1 and .NET 5 are out of support, but that doesn't mean it's fine to release changes that will break behavior in them without bumping the major.

mattjohnsonpint:
Enables `SqlDiagnosticListener` for .NET Standard using the existing .NET Core implementation.

Fixes #1928


mattjohnsonpint:
Question: Do you want me to add an explicit package reference to `System.Diagnostics.DiagnosticSource`?  I didn't yet, because it's currently imported transitively (`Microsoft.Data.SqlClient` => `Azure.Identity` => `Azure.Core` => `System.Diagnostics.DiagnosticSource`).

mattjohnsonpint:
Actually, I think I will need to copy some more code over to the netfx project, as it looks like the listener is wired up only for the netcore versions of `SqlCommand.cs`, `SqlConnection.cs`, and `SqlTransaction.cs`.

mattjohnsonpint:
I started copying over the invocation code over to the netfx version, but it turned out to be too much work.  There's a lot of deviation between the two sets of sources, and I'm just not familiar enough with this code to make such a mass change.

It should be possible (and likely easy) for someone who's already acclimated to this project to make such improvements though.  There aren't any hard blockers, as far as I can tell.

I force-pushed to revert the previous attempt.  Now it just focuses on .NET Standard.

mattjohnsonpint:
Anything I need to know about the four check failures? Looks like most of it passed just fine.

lcheunglci:
I don't think the failures are related to your changes as we notice some intermittent failures that happening in certain async tests which a draft PR was opened to address them.

mattjohnsonpint:
Oh, I just noticed that #535 actually split several other source files.  I'll make a few more changes.  Thanks.

mattjohnsonpint:
Ok, the only actual split in #535 was with `SqlAuthenticationProviderManager`, but it appears that the implementation details there were related more to `EventSource` than the `SqlDiagnosticListener`.  Nonetheless, I see that the event source is already implemented for .NET Standard 2.1, just not .NET Standard 2.0.  So the same concern applies.  (.NET Core 3.1+5.0 are .NET Standard 2.1 compliant)

Looking further at the project file, I found that *all* of the files that were in the `netstandard` item group were actually compatible with .NET Standard 2.1.  Other than `SqlAuthenticationProviderManager`, all of them appear to have been split due to APIs not in .NET Standard 2.0 such as `ValueTask` and `new Guid(ReadOnlySpam<byte>)`, etc.

https://github.com/dotnet/SqlClient/blob/85edec97efb8ee50d6d6103087390ad81e24e108/src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj#L517-L524

I moved them to the appropriate item groups, and everything still compiles without error.

After the changes, the only files remaining in the `netcoreapp`-only group are:

https://github.com/dotnet/SqlClient/blob/c53ab683d7533d8116e9132c62b26d72e118362f/src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj#L561-L567

I'm uncertain about these.  They compile on _both_ .NET Standard 2.0 and 2.1, but if that's the case then why were they split to netcoreapp only in the first place?

mattjohnsonpint:
I just pushed another commit that updates the `System.Diagnostics.DiagnosticSource` package dependency.  Although it is brought in transitively already, it's probably better to be explicit incase it gets dropped from the transitive reference in the future.

In doing so, I noticed that it was already being referenced for .NET Core.  That's unnecessary, as it's already built-in there.  It's only needed on the .NET Standard targets.

DavoudEshtehari:
I built the project and tested using the package reference against TF `netcoreapp3.1`, which is not run on the pipeline, and it interestingly looks good 👍🏼. My theory for the netstandard diversion, especially v2.1 is that it wouldn't have been supported at the time those functionalities had been added or they have been classified based on a wrong assumption.

DavoudEshtehari (approved):

Issue #1930, opened 2023-02-22
Could not load type 'SqlGuidCaster' from assembly Microsoft.Data.SqlClient


Alerinos:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

Error:
```
System.Reflection.ReflectionTypeLoadException: „Unable to load one or more of the requested types.
Could not load type 'SqlGuidCaster' from assembly 'Microsoft.Data.SqlClient, Version=5.0.0.0, Culture=neutral, PublicKeyToken=23ec7fc2d6eaa4a5' because it contains an object field at offset 0 that is incorrectly aligned or overlapped by a non-object field.”
```
Code:
```cs
    public static IEnumerable<Assembly> GetAssembliesWithTypesImplementing<T>()
    {
        return AppDomain.CurrentDomain.GetAssemblies()
            .SelectMany(assembly => assembly.GetTypes())
            .Where(type => !type.IsAbstract && typeof(T).IsAssignableFrom(type))
            .Select(type => type.Assembly)
            .Distinct();
    }
```
line:
``assembly.GetTypes()``

The error appeared after upgrading from .net 7 to .net 8 preview 1.
I searched and I don't have such Microsoft.Data.SqlClient package installed

Wraith2:
Even if you aren't using it directly you must have a transitive reference to Microsoft.Data.SqlClient (probably through EFCore) in order to have it present in the assembly list. So can you provide the minimal reproduction that can be debugged which exhibits this problem?

lcheunglci:
Hi @Alerinos  Could you provide us with a sample repro? The provided code snippet is not enough information for us to investigate.

Alerinos:
@Wraith2 @lcheunglci 
I was able to reproduce the problem, here is the repo:
https://github.com/Alerinos/SQLError
Error:
![image](https://user-images.githubusercontent.com/42449535/220736735-e0bbb9a2-4a22-4b2a-b755-0b92aa22d625.png)

Alerinos:
@lcheunglci Let me know if there is a problem, if so is there a quick workaround? The problem occurs only on .net 8 in .net 7 everything is fine.

Wraith2:
Ok. that's interesting. The problem isn't in SqlClient it's to do without how object layouts are computed in the Jit. It looks like a class that's laid out safely in 7 is now not safe in 8. Since 8 is still in very early preview this isn't a pressing problem but i'll see if i can find the people who might know what to do.

[...]

PR #1934, opened 2023-02-26
Add net8 compatibility

Wraith2:
closes https://github.com/dotnet/SqlClient/issues/1930

In netcore builds SqlGuidCaster and the type workaround that used it are not required. The problem they solved has already been change to use the ReadOnlySpan<byte> ctor. The SqlGuid workarounds in netcore are removed. Netfx builds use a more recent approach to the original problem and do not need to change.

I have also added ifdeffed use of the new api's introduced to the ryntime in https://github.com/dotnet/runtime/pull/72724 so if/when a net8 build is created the new methods will be used and the workarounds will be ifdeffed out entirely.

Testing on net was done manually with a small repro app that used reflection to force SqlGuidCaster to be compiled. Before these change the reported TypeLoadException was encountered, after them it is not.

/cc @Alerinos @MichalPetryka

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
@@ -648,6 +648,9 @@
     <Compile Include="Microsoft\Data\SqlClient\TdsParserHelperClasses.cs" />
     <Compile Include="Microsoft\Data\SqlClient\TdsParserStateObject.cs" />
     <Compile Include="Microsoft\Data\SqlClient\TdsParserStateObjectManaged.cs" />
+  </ItemGroup>
+  <!--This will exclude SqlTypeWorkarounds.netcore.cs from Net7 and greater versions-->
+  <ItemGroup Condition="'$(OSGroup)' != 'AnyOS' AND !$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net7.0'))">
     <Compile Include="Microsoft\Data\SqlTypes\SqlTypeWorkarounds.netcore.cs" />
   </ItemGroup>
   <!-- Windows only -->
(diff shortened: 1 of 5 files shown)


JRahnama:
I will do some changes to this PR, such as the csproj file to include SqlTypeWorkarounds to be included if it is net6 only and changing the Net 8 to net 7 for now.

Wraith2:
You can't change the net8 to 7 because this is adding support for new apis that aren't in 7. For the rest, go ahead.

David-Engel:
> You can't change the net8 to 7 because this is adding support for new apis that aren't in 7. For the rest, go ahead.

The new APIs made it into .net7 last year. I think we are good to go on this PR.
https://github.com/dotnet/runtime/commits/v7.0.0/src/libraries/System.Data.Common/src/System/Data/SQLTypes

DavoudEshtehari (approved):

FS-FrankLakmann:
Hello, we just stumbled over  https://github.com/dotnet/SqlClient/issues/1930
One question concerning your release strategy: Why was this fix not added to the LTS 5.1.x branch? Now we have a situation where the only branch with half-decent (not enough but better than the other branches) support lifetime is not compatible to the current LTS version of the dotnet runtime itself. 
Please remember: not everybody does inhouse development and can push releases every month. There are ISVs out there who release once a year and those have to run for 2 years with only minor bugfixes. We need stable and long-term support for the runtime and essential base libs like SqlClient. 
Your LTS schedule should be aligned with the release of the framework LTS-versions and keep compatibility with that version.

Wraith2:
Why is your code, or your vendor code, touching internal types in an assembly it does not own?

FS-FrankLakmann:
I'm not sure what you mean by touching. We just used SqlClient with NHibernate in dotnet 8 for some months in an ordinary line-of-business-application. Everthinig was fine with version 5.1.6. Then we made some changes to entity definitions and data mappings and got the same runtime exception as in #1930. I have no idea what triggered the change but we were not doing anything fancy in the internals.

Wraith2:
The only way to touch SqlGuidCaster  is through reflection accessing internal types inside SqlClient. That's why it wasn't found in testing, it wasn't used and the public surface area wasn't changed. If your dependency, NHibernate was using private reflection then it should not have been.

ErikEJ:
@Wraith2 Looks like this repros:

`assembly.GetTypes()`

PR #1936, opened 2023-02-27
Fix | TDS RPC error on large queries in SqlCommand.ExecuteReaderAsync

lcheunglci:
We found bug when using always encrypted with a large query with at least a parameter in async, it would insert extra bytes into the TDS packet header for the enclave byte per parameter when run in async mode resulting in the following error:
```
The incoming tabular data stream (TDS) remote procedure call (RPC) protocol stream is incorrect. Parameter 2 (""): Data type 0x00 is unknown.
```
It turns out it was fixed in .NET Core but broken in .NET Framework, so after I backport changes i.e. move the logic of adding the enclave byte from once per parameter to once per rpc call, it seems to fix the issue.

### DIFF
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/ApiShould.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/ApiShould.cs
@@ -692,7 +692,7 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests.AlwaysEncrypted
 
         [ConditionalTheory(typeof(DataTestUtility), nameof(DataTestUtility.AreConnStringSetupForAE))]
         [ClassData(typeof(AEConnectionStringProvider))]
-        public async void TestExecuteReaderAsyncWithLargeQuery(string connection)
+        public async void TestExecuteReaderAsyncWithLargeQuery(string connectionString)
         {
             string randomName = DataTestUtility.GetUniqueName(Guid.NewGuid().ToString().Replace("-", ""), false);
             if (randomName.Length > 50)
@@ -705,15 +705,15 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests.AlwaysEncrypted
             // Arrange - drops the table with long name and re-creates it with 52 columns (ID, name, ColumnName0..49)
             try
             {
-                CreateTable(connection, tableName, columnsCount);
+                CreateTable(connectionString, tableName, columnsCount);
                 string name = "nobody";
 
-                using (SqlConnection sqlConnection = new SqlConnection(connection))
+                using (SqlConnection connection = new SqlConnection(connectionString))
                 {
-                    await sqlConnection.OpenAsync();
+                    await connection.OpenAsync();
                     // This creates a "select top 100" query that has over 40k characters
                     using (SqlCommand sqlCommand = new SqlCommand(GenerateSelectQuery(tableName, columnsCount, 10, "WHERE Name = @FirstName AND ID = @CustomerId"),
-                        sqlConnection,
+                        connection,
                         transaction: null,
                         columnEncryptionSetting: SqlCommandColumnEncryptionSetting.Enabled))
                     {
@@ -739,7 +739,7 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests.AlwaysEncrypted
             }
             finally
             {
-                DropTableIfExists(connection, tableName);
+                DropTableIfExists(connectionString, tableName);
             }
         }
 
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/ApiShould.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/ApiShould.cs
@@ -692,7 +692,7 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests.AlwaysEncrypted
 
         [ConditionalTheory(typeof(DataTestUtility), nameof(DataTestUtility.AreConnStringSetupForAE))]
         [ClassData(typeof(AEConnectionStringProvider))]
-        public async void TestExecuteReaderAsyncWithLargeQuery(string connectionString)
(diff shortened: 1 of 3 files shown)


Wraith2:
Looks like a simple fix to a hard to investigate problem. 👍

lcheunglci:
Indeed, the fix was simple but took a long time to get proper environment and repro case, and apparently it happened in all versions of Microsoft.Data.SqlClient including 1.0 but somehow was still working on System.Data.SqlClient.

David-Engel (approved):

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/ApiShould.cs
```diff
@@ -2806,6 +2856,93 @@ private void CleanUpTable(string connString, string tableName)
...
+        {
+            using (var sqlConnection = new SqlConnection(connString))
+            {
+                sqlConnection.Open();
+
+                SqlCommand cmd = new SqlCommand(GenerateCreateQuery(tableName, columnsCount), sqlConnection);
+                cmd.ExecuteNonQuery();
+            }
+        }
+
```
DavoudEshtehari, line 2875: Using data utilities.
```suggestion
        private static void CreateTable(string connString, string tableName, int columnsCount)
            => DataTestUtility.RunNonQuery(connString, GenerateCreateQuery(tableName, columnsCount));
```

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/ApiShould.cs
```diff
@@ -689,6 +690,55 @@ public void TestExecuteReader(string connection)
...
+                        {
+                            Assert.False(sqlDataReader.HasRows, "The table should be empty");
+                        }
+                    }
+                }
+            }
+            catch (Exception ex)
+            {
+                Assert.Fail($"The following exception was thrown: {ex.Message}");
+            }
```
DavoudEshtehari, line 735: Any exceptions should break the test. Swallowing the inner exception hides helpful information on a failure.
```suggestion
            catch
            {
                throw;
            }
```

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/ApiShould.cs
```diff
@@ -2806,6 +2856,93 @@ private void CleanUpTable(string connString, string tableName)
...
+        /// <param name="tableName">The name of the table to be dropped</param>
+        private void DropTableIfExists(string connString, string tableName)
+        {
+            using (var sqlConnection = new SqlConnection(connString))
+            {
+                sqlConnection.Open();
+                SqlCommand cmd = new SqlCommand($"DROP TABLE IF EXISTS {tableName};", sqlConnection);
+                cmd.ExecuteNonQuery();
+            }
+        }
```
DavoudEshtehari, line 2889: Using data utilities.
```suggestion
        private static void DropTableIfExists(string connString, string tableName)
        {
            using var sqlConnection = new SqlConnection(connString);
            sqlConnection.Open();
            DataTestUtility.DropTable(sqlConnection, tableName);
        }
```

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/ApiShould.cs
```diff
@@ -2806,6 +2856,93 @@ private void CleanUpTable(string connString, string tableName)
...
+
+        /// <summary>
+        /// Generates the query for creating a table with the number of bit columns specified.
+        /// </summary>
+        /// <param name="tableName">The name of the table</param>
+        /// <param name="columnsCount">The number of columns for the table</param>
+        /// <returns></returns>
+        private string GenerateCreateQuery(string tableName, int columnsCount)
```
DavoudEshtehari, line 2897:
```suggestion
        private static string GenerateCreateQuery(string tableName, int columnsCount)
```

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/ApiShould.cs
```diff
@@ -2806,6 +2856,93 @@ private void CleanUpTable(string connString, string tableName)
...
+
+        /// <summary>
+        /// Generates the large query with the select top 100 of all the columns repeated multiple times.
+        /// </summary>
+        /// <param name="tableName">The name of the table</param>
+        /// <param name="columnsCount">The number of columns to be explicitly included</param>
+        /// <param name="repeat">The number of times the select query is repeated</param>
+        /// <param name="where">A where clause for additional filters</param>
+        /// <returns></returns>
+        private string GenerateSelectQuery(string tableName, int columnsCount, int repeat = 10, string where = "")
```
DavoudEshtehari, line 2922:
```suggestion
        private static string GenerateSelectQuery(string tableName, int columnsCount, int repeat = 10, string where = "")
```

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/ApiShould.cs
```diff
@@ -689,6 +690,55 @@ public void TestExecuteReader(string connection)
             });
         }
 
+        [ConditionalTheory(typeof(DataTestUtility), nameof(DataTestUtility.AreConnStringSetupForAE))]
+        [ClassData(typeof(AEConnectionStringProvider))]
+        public async void TestExecuteReaderAsyncWithLargeQuery(string connection)
```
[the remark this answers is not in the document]
DavoudEshtehari: This was a partial change to convey the idea. I just wanted to explain to make sure you deliberately put it aside. 😉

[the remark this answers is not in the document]
lcheunglci: thanks, that make sense. I must have accidentally reverted the change when I saw the compiler failure.  I went back in and refactored it properly.

PR #1943, opened 2023-03-05
Replace calls to `Method.GetCurrentMethod().Name` with `nameof()`

kant2002:
That reduce places which marked as trimmer unfriendly. Even if `Method.CurrentMethod().Name` will work after trimming, but tooling cannot decide that in generic cases, see: https://github.com/dotnet/runtime/issues/53242 At this point I would like to replace formatting primitives with `$""` if possible, since that allow Rolsyn build constant string for us at compile time.

Also enabled Trim Analyzer on the build, so all trim warnings can be gradually removed after proper annotations, or other work done.

/cc @Wraith2

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/LocalAppContextSwitches.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/LocalAppContextSwitches.cs
@@ -10,7 +10,6 @@ namespace Microsoft.Data.SqlClient
 {
     internal static partial class LocalAppContextSwitches
     {
-        private const string TypeName = nameof(LocalAppContextSwitches);
         internal const string MakeReadAsyncBlockingString = @"Switch.Microsoft.Data.SqlClient.MakeReadAsyncBlocking";
         internal const string LegacyRowVersionNullString = @"Switch.Microsoft.Data.SqlClient.LegacyRowVersionNullBehavior";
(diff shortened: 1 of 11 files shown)


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/LocalAppContextSwitches.cs
```diff
@@ -30,7 +30,7 @@ static LocalAppContextSwitches()
             catch (Exception e)
             {
                 // Don't throw an exception for an invalid config file
-                SqlClientEventSource.Log.TryTraceEvent("<sc.{0}.{1}|INFO>: {2}", TypeName, MethodBase.GetCurrentMethod().Name, e);
+                SqlClientEventSource.Log.TryTraceEvent("<sc.{0}.{1}|INFO>: {2}", TypeName, "..ctor", e);
```
Wraith2, line 33: shouldn't this be `.cctor` for a static constructor?

Wraith2: Also, can you inline the TypeName, it's a nameof in a variable which is a bit pointless when it's single use.

Wraith2 (approved):

DavoudEshtehari (approved):

lcheunglci (approved):

lcheunglci:
The changes looks good to me. Thanks @kant2002 for your contributions

PR #1948, opened 2023-03-08
Fix | Addressing failure on providing correct error message when symmetric key decryption fails using Always Encrypted.

JRahnama:
The issue is happening when IsDBNull called on an encrypted column with backed keys in Azure key vault. If, for any reason, application fails to decrypt column keys, not the value the actual key for the column, the driver is waiting for the pending data to be received and throws unknown TDS token header, by setting the pending data to false, application will not process the rest to read the data.

```C#
 string command = $"SELECT [SSN] FROM [{s_schema}].[{s_table}]";
            using SqlCommand sqlCommand = new(command, connection);
            try
            {
                connection.Open();
                using SqlCommand cmd = connection.CreateCommand();
                cmd.CommandText = command;
                try
                {
                    using SqlDataReader reader = cmd.ExecuteReader();
                    if (reader.HasRows)
                    {
                        while (reader.Read())
                        {
                            if (!reader.IsDBNull(0))
                            {
                                _ = reader[0];
                                Console.WriteLine("It's not null db");
                            }
                        }
                    }
                }
                catch (Exception ex)
                {
                    Console.WriteLine(ex);
                }
~~~

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
@@ -6080,7 +6080,11 @@ namespace Microsoft.Data.SqlClient
                         }
                         catch (Exception e)
                         {
-                            _physicalStateObj.HasPendingData = false;
+                            if (stateObj is not null)
+                            {
+                                DrainData(stateObj);
+                                stateObj.HasPendingData = false;
+                            }
                             throw SQL.ColumnDecryptionFailed(columnName, null, e);
                         }
                     }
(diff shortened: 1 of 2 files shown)


lcheunglci (approved):

DavoudEshtehari:
You removed the `DrainData(stateObj)` function call later. Can you explain what the implication was?

Also, I am just adding a note not to forget to include tests.

JRahnama:
> You removed the `DrainData(stateObj)` function call later. Can you explain what the implication was?
> 
> Also, I am just adding a note not to forget to include tests.

Adding a test is bit complicated. It needs some modification to prevent the application from obtaining the key to decrypt column key. DrainData is reproducing the issue as there is no data at this point for the column. Basically at this point setting pending data to false will cause the driver to skip looking for the column value and exception will be thrown immediately.

DavoudEshtehari:
What would be the result of this change with MARS enabled, and/or parallel async calls? These are the questions that require more tests.

JRahnama:
> What would be the result of this change with MARS enabled, and/or parallel async calls? These are the questions that require more tests.

So, the issue only happens on ISDBNull to get the column's info and if you dont use that no issue comes up at all. For async and MARS we do have tests to run and read values, I am not sure if ISDBNull has been called inside those tests, but no failure on the pipelines so far. I will try to come up with a solution for the test purposes.

David-Engel (approved):

lcheunglci:
Thanks! I've tested this fix locally and it seems good to me. We'll have to add the test in a separate PR.

Issue #1942, opened 2023-03-04
Improve binary size when using NativeAOT


roji:
As pointed out by @vonzshik in #1941, a minimal SqlClient application weighs 45MB, which is quite a lot. For use in size-sensitive environments, it would be good to reduce this.

Note: this may be related to #1108 - the Azure-specific functionality is currently built-in, making it opt-in would also reduce binary size.

kant2002:
From what I see there following offenders

![image](https://user-images.githubusercontent.com/4257079/222961353-ad8927cc-50ff-4939-9062-66db77221d75.png)

- System.Private.Xml
- System.Text.Json

# System.Private.Xml

This is coming from https://github.com/dotnet/SqlClient/blob/73c6b2ff3ef3b55431cf187e72d5de918caccf76/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs#L2856-L2884

and from this place
https://github.com/dotnet/SqlClient/blob/73c6b2ff3ef3b55431cf187e72d5de918caccf76/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs#L2967-L2978

# System.Text.Json

This one is a bit tricky, since it's indirectly referenced
- Microsoft.IdentityModel.JsonWebTokens.JsonWebTokenHandler coming from
https://github.com/dotnet/SqlClient/blob/73c6b2ff3ef3b55431cf187e72d5de918caccf76/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/EnclaveDelegate.Crypto.cs#L92-L96

Output of WhyDgml if interesting.
```
**** Match 1 ****
() System_Text_Json_System_Text_Json_Utf8JsonReader___ctor
  (call) System_Text_Json_System_Text_Json_JsonDocument__Parse
    (call) System_Text_Json_System_Text_Json_JsonDocument__Parse_5
      (call) System_Text_Json_System_Text_Json_JsonDocument__Parse_3
        (call) System_Text_Json_System_Text_Json_JsonDocument__Parse_4
          (call) Microsoft_IdentityModel_JsonWebTokens_Microsoft_IdentityModel_JsonWebTokens_JsonWebToken__ReadToken
[...]

Issue #1108, opened 2021-06-09
Feature | Split Azure dependent functionality in a separate NuGet Package


SimonCropp:
why does Microsoft.Data.SqlClient v3 reference Azure.Identity?

SimonCropp:
found my answer https://github.com/dotnet/SqlClient/pull/1010. so will close this. it still seems very weird to me that a sql client should have a dependency on a cloud infrastructure library. For hypothetical comparison, would it be ok to also take dependencies on google and aws libraries.

David-Engel:
@SimonCropp Microsoft.Data.SqlClient also provides connectivity to Azure cloud services (Azure SQL DB, Azure Synapse Analytics, etc) that are very similar to SQL Server on-premise. The Azure.Identity library provides authentication functionality for those Azure services.

virzak:
I was also wondering if there is a way to separate functionality into multiple DLLs, i.e.
- Microsoft.Data.SqlClient
- Microsoft.Data.SqlClient.Azure

With .Net 5, a project which uses this library and doesn't use Azure gets at least 3 new unused DLLs in the bin directory:
Azure.Core.dll
Azure.Identity.dll
Microsoft.Bcl.AsyncInterfaces.dll

SimonCropp:
i had the same expectation as @virzak

virzak:
@SimonCropp reopen then? :)

SimonCropp:
@virzak i think i have raised my concerns. if the sql client team think they should be address, then they can re-open the issue

cheenamalhotra:
We are considering this internally, timeline not certain yet. But we can reopen for sure!

mtaylorfsmb:
Agree with others. This library is supposed to be replacement for the standard library already in the framework. Azure support is fine but shouldn't be auto-included in a "core" package. This would be equivalent to having the core framework depend upon Azure as well. If my app is using AWS or GCP then having an Azure dependency looks wrong.

aersamkull:
Agree, too. Mostly because of the dependencies Azure.Identity brings in. There are really many packages needed for Microsoft.Data.SqlClient

MarkPflug:
[...]

PR #1949, opened 2023-03-08
Remove linq

Wraith2:
helps with https://github.com/dotnet/SqlClient/issues/1942

Removes all uses of linq operations in the main binary and addons allowing linq to be trimmed out if no other libraries are using it
I suggest reviewing 1 commit at a time. Some of the enclave stuff is a bit messy and this cleans it up as well as removing the linq.

/cc @kant2002


Wraith2:
@David-Engel 

Is there someone knowledgeable about enclave stuff in the codebase that could help here. I've found something odd that concerns me.

There's a piece of code that was using `data.Skip(offset).Take(count)` and i replaced it with a function call that does the same thing without using linq. My replacement differs in behaviour by requiring that you be able to copy exactly `count` bytes where the linq `Take` function will only copy up-to `count` bytes, if there aren't enough it won't fail.

This caused failures in this code:
```csharp
int secureSessionBufferSize = Convert.ToInt32(secureSessionInfoResponseSize) - sizeof(uint);
byte[] secureSessionBuffer = attestationInfo.Skip(offset).Take(secureSessionBufferSize).ToArray();
EnclaveDHInfo = new EnclaveDiffieHellmanInfo(secureSessionBuffer);
offset += Convert.ToInt32(EnclaveDHInfo.Size);

if (secureSessionBufferSize != secureSessionBuffer.Length)
{
	throw new InvalidOperationException($"enclave data mismatch, expected {secureSessionBufferSize} got {EnclaveDHInfo.Size}");
}
```
Where we calculate how many bytes of data we expect to get and then try to take it and end up 4 bytes short. Either the calculation of how many bytes we should expect to get is wrong or the something else is going on.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SSRP.cs
```diff
@@ -372,5 +396,40 @@ internal static string SendBroadcastUDPRequest()
             }
             return response.ToString();
         }
+
+        private static void SplitIPv4AndIPv6(IPAddress[] input, out IPAddress[] ipv4Addresses, out IPAddress[] ipv6Addresses)
+        {
+            ipv4Addresses = Array.Empty<IPAddress>();
+            ipv6Addresses = Array.Empty<IPAddress>();
+
+            if (input != null && input.Length > 0)
```
kant2002, line 405: Maybe revert this condition and have fast exit? Code above looks a bit unnatural with this condition
```
ipv4Addresses = Array.Empty<IPAddress>();
ipv6Addresses = Array.Empty<IPAddress>();
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIPhysicalHandle.cs
```diff
@@ -83,12 +83,19 @@ public override void ReturnPacket(SNIPacket packet)
...
-                .Skip(3) // trims off the common parts at the top of the stack so you can see what the actual caller was
-                .Take(7) // trims off most of the bottom of the stack because when running under xunit there's a lot of spam
-            );
+            // trims off the common parts at the top of the stack so you can see what the actual caller was
+            // trims off most of the bottom of the stack because when running under xunit there's a lot of spam
+            string[] parts = Environment.StackTrace.Split(new string[] { Environment.NewLine }, StringSplitOptions.None);
+            List<string> take = new List<string>(7);
+            for (int index = 0; take.Count < 7 && index < parts.Length; index++)
+            {
+                if (index > 2)
```
kant2002, line 92: Why not start counting from index 3 immediately?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SSRP.cs
Wraith2: It's like this because they're out parameters. I suppose it's a personal style choice but I always assign the default values for out parameters at the start of the method so that anywhere inside the body I can easily return without worrying about whether they're assigned or not.

roji:
@Wraith2 what kind of size reduction are you seeing with this?

roji:
> I also think it's unlikely to help if sqlclient is used through EFCore because there's no way that linq is getting trimmed from EFCore.

I'm sure that's true; EF hasn't yet undergone a real trimming effort - it's possible (thought not sure) that when we do that in a serious way, LINQ would be needed.

But I'd assume that various dependencies of SqlClient are also bringing in LINQ (e.g. the Azure Identity libs). Personally, for cases where removing LINQ makes the code less maintainable, I'd avoid doing that until I see it provides a tangible gain (but I also think there are at least some cases in this PR where the code doesn't become less maintainable).

Wraith2:
Agreed. I was talking with kant on discord about this and it's possible that the azure libs will be trimmable in future if you can work out the api to do so. 

If a total removal isn't appropriate certain parts of this can be backed out. There's plenty of scope for discussion and this is only a draft.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIPhysicalHandle.cs
kant2002: How so? I see that 
```
for (int index = 3; take.Count < 7 && index < parts.Length; index++)
{
    take.Add(parts[index]);
}
```
would be unrolled like that
```
int index = 3;
start_loop:
if (take.Count < 7 && index < parts.Length)
{
    take.Add(parts[index]);
    index++;
    goto start_loop;
}

```

roji:
> I was talking with kant on discord about this and it's possible that the azure libs will be trimmable in future if you can work out the api to do so.

I'd propose thinking about the approach we've taken in Npgsql (https://github.com/dotnet/SqlClient/issues/1942#issuecomment-1455106580): a "slim" data source builder which, when used, doesn't have size-expensive functionality on by default, unless explicitly opted into.

However, specifically regarding the Azure libs, the very fact that SqlClient has a direct references on them is probably not ideal (#1108, highest-voted issue on the repo). Here specifically, I'd suggest SqlClient expose an auth API that allows 3rd-party packages (like the Azure ones) to register themselves. This would totally decouple SqlClient from Azure, and of course naturally strip them away size-wise unless used.

Wraith2:
As long as backwards compatibility is maintained adding new apis around the DataSource abstraction seems fine to me. That could allow both advanced AE and Enclave functionality as well as extended authentication mechanisms to be trimmable.

David-Engel:
> Is there someone knowledgeable about enclave stuff in the codebase that could help here. I've found something odd that concerns me.
> 
> There's a piece of code that was using `data.Skip(offset).Take(count)` and i replaced it with a function call that does the same thing without using linq. My replacement differs in behaviour by requiring that you be able to copy exactly `count` bytes where the linq `Take` function will only copy up-to `count` bytes, if there aren't enough it won't fail.
> 
> This caused failures in this code:
> 
> ```cs
> int secureSessionBufferSize = Convert.ToInt32(secureSessionInfoResponseSize) - sizeof(uint);
> byte[] secureSessionBuffer = attestationInfo.Skip(offset).Take(secureSessionBufferSize).ToArray();
> EnclaveDHInfo = new EnclaveDiffieHellmanInfo(secureSessionBuffer);
> offset += Convert.ToInt32(EnclaveDHInfo.Size);
> 
> if (secureSessionBufferSize != secureSessionBuffer.Length)
> {
> 	throw new InvalidOperationException($"enclave data mismatch, expected {secureSessionBufferSize} got {EnclaveDHInfo.Size}");
> }
> ```
> 
> Where we calculate how many bytes of data we expect to get and then try to take it and end up 4 bytes short. Either the calculation of how many bytes we should expect to get is wrong or the something else is going on.

@Wraith2 

This doesn't fail outside of your PR, right? Wouldn't the original exception throw if there was some underlying issue with the original code? I've been looking at the code changes and I'd have to debug it to get a better idea of what might be happening but it feels like something is missing. I need to find an enclave environment set up to debug against. The team has just been horribly busy with support issues for the last month or so...

Wraith2:
It isn't a failure outside my PR. I think because while the length calculations are incorrect it isn't meaningful and the code actually does the right thing. 

This is the original code:
```csharp
                    uint secureSessionInfoResponseSize = BitConverter.ToUInt32(attestationInfo, offset);
                    offset += sizeof(uint);

                    SessionId = BitConverter.ToInt64(attestationInfo, offset);
                    offset += sizeof(long);

                    int secureSessionBufferSize = Convert.ToInt32(secureSessionInfoResponseSize) - sizeof(uint);
                    byte[] secureSessionBuffer = attestationInfo.Skip(offset).Take(secureSessionBufferSize).ToArray();
                    EnclaveDHInfo = new EnclaveDiffieHellmanInfo(secureSessionBuffer);
                    offset += Convert.ToInt32(EnclaveDHInfo.Size);
```

I think the sessionId which is read as `sizeof(long)` is incorrectly subtracted from the remaining bytes as `sizeof(uint)` but since we've actually read 8 bytes and advanced by 8 bytes we're just calculating the remaining length incorrectly. When we do Skip(offset).Take(length) the length is 4 bytes longer than we have available but the offset is correct and the lazy behaviour of Take causes it to turn out correct. 

So if the code is working should I change it? I think so. I think it should be changed because the comment above the function we're looking at doesn't accurately express what the code is doing and because the code is working accidentally. As I've proven if you try to change it you risk breaking it by interpreting the code and not carefully inspecting the results in a debugger. Since the only way to debug this is complex and requires enclave functionality to be available I think code would be better.

David-Engel:
@Wraith2 

Looking at the JDBC implementation, I think the existing code is just failing to account for an additional size field after the Session ID, which is not accounted for in the secureSessionInfoResponseSize when calculating the secureSessionBufferSize.
[JDBC reference](https://github.com/microsoft/mssql-jdbc/blob/main/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerAASEnclaveProvider.java#L277)

You're right. It's an error, but one that fortunately works itself out due to the Take() behavior.
```csharp
int secureSessionBufferSize = Convert.ToInt32(secureSessionInfoResponseSize) - sizeof(uint);
```
should be
```csharp
// Account for 4 byte size of enclave ECDH public key followed by 4 byte size of the DH public key signature
int secureSessionBufferSize = Convert.ToInt32(secureSessionInfoResponseSize) - sizeof(uint) - sizeof(uint);
```

I still haven't debugged, so let me know if that works out correctly. 😄

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIPhysicalHandle.cs
Wraith2: Agreed, I was wrong. Changed.

Wraith2:
I revised the code but the CI is apparently off for the weekend already.

David-Engel (approved):
Looks good to me. This takes us in a good direction with regard to trimming and potential AoT support. Thanks for the PR!

lcheunglci:
We just merged https://github.com/dotnet/SqlClient/pull/1943 and it created a small merge conflict with the namespace.  Let us know if you would like us to push changes to your PR branch or if you have moment to fix it. Thanks!

Wraith2:
conflicts resolved and pushed.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AlwaysEncryptedEnclaveProviderUtils.cs
```diff
@@ -18,23 +18,21 @@ public EnclavePublicKey(byte[] payload)
 
     internal class EnclaveDiffieHellmanInfo
     {
-        public int Size { get; private set; }
+        public int Size => sizeof(int) + sizeof(int) + PublicKey?.Length ?? 0 + PublicKeySignature?.Length ?? 0;
```
lcheunglci, line 21: I'm curious how did Size = `payload.Length` from the ctor end up being this calculation.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/VirtualSecureModeEnclaveProvider.cs
```diff
@@ -274,32 +272,25 @@ public EnclaveReportPackageHeader(byte[] payload)
...
-        }
+        public static int SizeInPayload => 6 * sizeof(uint);
     }
 
     // A managed model of struct VBS_ENCLAVE_REPORT
     // https://msdn.microsoft.com/en-us/library/windows/desktop/mt844255(v=vs.85).aspx
     internal class EnclaveReport
     {
-        private int Size { get; set; }
+        private const int EnclaveDataLength = 64;
```
lcheunglci, line 282: I can see that `Size` is only set once in the ctor and this is private, so not sure why did we needed it in the first place so removing it seem fine.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Reliability/SqlConfigurableRetryLogicLoader.cs
```diff
@@ -273,5 +298,26 @@ private static void OnRetryingEvent(object sender, SqlRetryingEventArgs args)
...
+                    {
+                        if (int.TryParse(parts[index], System.Globalization.NumberStyles.AllowLeadingWhite | System.Globalization.NumberStyles.AllowTrailingWhite, null, out int value))
+                        {
+                            set.Add(value);
+                        }
+                    }
+                    return set;
+                }
+            }
+            return null;
```
lcheunglci, line 320: `.Select(x => Convert.ToInt32(x)).ToList()` would return an empty collection instead of null right?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AlwaysEncryptedEnclaveProviderUtils.cs
Wraith2: I think it was a knock on effect of the size calculation being incorrect elsewhere. 

It feels like the author had a problem with being a few bytes off and was using Size to help try and work out where it was, once they found or resolved it they didn't feel the need to go back and change them all around. I changed it so that rather than keeping the number of bytes that were made available we report the actual number of bytes that were read.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/VirtualSecureModeEnclaveProvider.cs
```diff
@@ -4,9 +4,9 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
+using System.IdentityModel.Tokens.Jwt;
```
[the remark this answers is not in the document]
Wraith2: No. I think that's visual studio auto adding usings for extensions. Removed.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/VirtualSecureModeEnclaveProvider.cs
```diff
@@ -308,38 +299,33 @@ public EnclaveReport(byte[] payload)
             ReportVersion = BitConverter.ToUInt32(payload, offset);
             offset += sizeof(uint);
 
-            EnclaveData = payload.Skip(offset).Take(EnclaveDataLength).ToArray();
-            offset += EnclaveDataLength;
+            //EnclaveData = payload.Skip(offset).Take(EnclaveDataLength).ToArray();
+            //offset += EnclaveDataLength;
```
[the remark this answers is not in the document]
Wraith2: Done.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Reliability/SqlConfigurableRetryLogicLoader.cs
Wraith2: True, and in this case the null wouldn't have failed but would cause the default set to be used which would mean  you could no longer set an empty list. Changed to return an empty hashset.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SSRP.cs
```diff
@@ -192,52 +191,77 @@ private static byte[] SendUDPRequest(string browserHostname, int port, byte[] re
...
                 Debug.Assert(ipAddresses.Length > 0, "DNS should throw if zero addresses resolve");
-
+                IPAddress[] ipv4Addresses = null;
+                IPAddress[] ipv6Addresses = null;
                 switch (ipPreference)
                 {
                     case SqlConnectionIPAddressPreference.IPv4First:
                         {
-                            SsrpResult response4 = SendUDPRequest(ipAddresses.Where(i => i.AddressFamily == AddressFamily.InterNetwork).ToArray(), port, requestPacket, allIPsInParallel);
+                            SplitIPv4AndIPv6(ipv4Addresses, out ipv4Addresses, out ipv6Addresses);
```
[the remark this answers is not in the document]
lcheunglci: I agree.  I was wondering about that as well.  It turns out since this code is only run on Linux and requires sudo/administrative privileges to get the correct environment, there's currently no test coverage for this section unfortunately.

lcheunglci (approved):
LGTM, Thanks @Wraith2 for your contributions!

PR #1958, opened 2023-03-21
Partially annotate IL2026

kant2002:
I cannot get rid all of them since rety logic used everywhere in the SqlConnection and SqlCommad and I think there maybe other better ways. Other places is more or less self-contained and did not pollute anything.

/cc @Wraith2

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Common/src/System/Net/Logging/NetEventSource.Common.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Common/src/System/Net/Logging/NetEventSource.Common.cs
@@ -10,6 +10,7 @@
 
 using System.Collections;
 using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
 using System.Diagnostics.Tracing;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
@@ -50,6 +51,8 @@ namespace System.Net
 #endif
     internal sealed partial class NetEventSource : EventSource
     {
+        private const string EventSourceSuppressMessage = "Parameters to this method are primitive and are trimmer safe";
+
         /// <summary>The single event source instance to use for all logging.</summary>
         public static readonly NetEventSource Log = new NetEventSource();
 
@@ -518,6 +521,8 @@ namespace System.Net
         #region Custom WriteEvent overloads
 
         [NonEvent]
+        [SuppressMessage("ReflectionAnalysis", "IL2026:UnrecognizedReflectionPattern",
+                   Justification = EventSourceSuppressMessage)]
         private unsafe void WriteEvent(int eventId, string arg1, string arg2, string arg3, string arg4)
         {
             if (IsEnabled())
@@ -566,6 +571,8 @@ namespace System.Net
         }
 
         [NonEvent]
+        [SuppressMessage("ReflectionAnalysis", "IL2026:UnrecognizedReflectionPattern",
+                   Justification = EventSourceSuppressMessage)]
         private unsafe void WriteEvent(int eventId, string arg1, string arg2, byte[] arg3)
         {
             if (IsEnabled())
@@ -612,6 +619,8 @@ namespace System.Net
         }
 
         [NonEvent]
+        [SuppressMessage("ReflectionAnalysis", "IL2026:UnrecognizedReflectionPattern",
+                   Justification = EventSourceSuppressMessage)]
         private unsafe void WriteEvent(int eventId, string arg1, int arg2, int arg3, int arg4)
         {
             if (IsEnabled())
@@ -651,6 +660,8 @@ namespace System.Net
         }
 
         [NonEvent]
+        [SuppressMessage("ReflectionAnalysis", "IL2026:UnrecognizedReflectionPattern",
+                   Justification = EventSourceSuppressMessage)]
         private unsafe void WriteEvent(int eventId, string arg1, int arg2, string arg3)
         {
             if (IsEnabled())
@@ -688,6 +699,8 @@ namespace System.Net
         }
 
         [NonEvent]
+        [SuppressMessage("ReflectionAnalysis", "IL2026:UnrecognizedReflectionPattern",
+                   Justification = EventSourceSuppressMessage)]
         private unsafe void WriteEvent(int eventId, string arg1, string arg2, int arg3)
         {
             if (IsEnabled())
@@ -725,6 +738,8 @@ namespace System.Net
         }
 
         [NonEvent]
+        [SuppressMessage("ReflectionAnalysis", "IL2026:UnrecognizedReflectionPattern",
+                   Justification = EventSourceSuppressMessage)]
         private unsafe void WriteEvent(int eventId, string arg1, string arg2, string arg3, int arg4)
         {
             if (IsEnabled())
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/Reliability/SqlConfigurableRetryLogicManager.NetCoreApp.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/Reliability/SqlConfigurableRetryLogicManager.NetCoreApp.cs
@@ -4,6 +4,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.Reflection;
 using System.Runtime.Loader;
@@ -21,6 +22,9 @@ namespace Microsoft.Data.SqlClient
         /// and its related assemblies in default assembly load context if they aren't loaded yet.
         /// </summary>
         /// <returns>Resolved type if it could resolve the type; otherwise, the `SqlConfigurableRetryFactory` type.</returns>
+#if NET6_0_OR_GREATER
+        [RequiresUnreferencedCode("The type might be removed")]
+#endif
         private static Type LoadType(string fullyQualifiedName)
         {
             string methodName = nameof(LoadType);
@@ -58,6 +62,9 @@ namespace Microsoft.Data.SqlClient
             return File.Exists(fullPath) ? fullPath : null;
         }
 
+#if NET6_0_OR_GREATER
+        [RequiresUnreferencedCode("Calls System.Runtime.Loader.AssemblyLoadContext.LoadFromAssemblyPath(String)")]
+#endif
         private static Assembly AssemblyResolver(AssemblyName arg)
         {
             string methodName = nameof(AssemblyResolver);
@@ -69,6 +76,10 @@ namespace Microsoft.Data.SqlClient
             return fullPath == null ? null : AssemblyLoadContext.Default.LoadFromAssemblyPath(fullPath);
         }
 
+
+#if NET6_0_OR_GREATER
+        [RequiresUnreferencedCode("Calls System.Reflection.Assembly.ExportedTypes")]
+#endif
         private static Type TypeResolver(Assembly arg1, string arg2, bool arg3)
         {
             IEnumerable<Type> types = arg1?.ExportedTypes;
@@ -97,6 +108,9 @@ namespace Microsoft.Data.SqlClient
         /// <summary>
         /// Load assemblies on request.
         /// </summary>
+#if NET6_0_OR_GREATER
+        [RequiresUnreferencedCode("Calls System.Runtime.Loader.AssemblyLoadContext.LoadFromAssemblyPath(String)")]
+#endif
         private static Assembly Default_Resolving(AssemblyLoadContext arg1, AssemblyName arg2)
         {
             string methodName = nameof(Default_Resolving);
@@ -107,6 +121,6 @@ namespace Microsoft.Data.SqlClient
 
             return target == null ? null : arg1.LoadFromAssemblyPath(target);
         }
-        #endregion
+#endregion
     }
 }
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlClientDiagnosticListenerExtensions.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlClientDiagnosticListenerExtensions.cs
@@ -5,6 +5,7 @@
 using System;
 using System.Data;
 using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
 using System.Runtime.CompilerServices;
 
 namespace Microsoft.Data.SqlClient
@@ -14,6 +15,8 @@ namespace Microsoft.Data.SqlClient
     /// </summary>
     internal static class SqlClientDiagnosticListenerExtensions
     {
+        private const string EventSourceSuppressMessage = "Parameters to this method are statically known and thus are trimmer safe";
+
         public const string DiagnosticListenerName = "SqlClientDiagnosticListener";
 
         private const string SqlClientPrefix = "Microsoft.Data.SqlClient.";
@@ -38,6 +41,8 @@ namespace Microsoft.Data.SqlClient
         public const string SqlAfterRollbackTransaction = SqlClientPrefix + nameof(WriteTransactionRollbackAfter);
         public const string SqlErrorRollbackTransaction = SqlClientPrefix + nameof(WriteTransactionRollbackError);
 
+        [SuppressMessage("ReflectionAnalysis", "IL2026:UnrecognizedReflectionPattern",
+                   Justification = EventSourceSuppressMessage)]
         public static Guid WriteCommandBefore(this SqlDiagnosticListener @this, SqlCommand sqlCommand, SqlTransaction transaction, [CallerMemberName] string operation = "")
         {
             if (@this.IsEnabled(SqlBeforeExecuteCommand))
@@ -62,6 +67,8 @@ namespace Microsoft.Data.SqlClient
                 return Guid.Empty;
         }
 
+        [SuppressMessage("ReflectionAnalysis", "IL2026:UnrecognizedReflectionPattern",
+                   Justification = EventSourceSuppressMessage)]
         public static void WriteCommandAfter(this SqlDiagnosticListener @this, Guid operationId, SqlCommand sqlCommand, SqlTransaction transaction, [CallerMemberName] string operation = "")
         {
             if (@this.IsEnabled(SqlAfterExecuteCommand))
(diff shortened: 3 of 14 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/Reliability/SqlConfigurableRetryLogicManager.NetCoreApp.cs
```diff
@@ -21,6 +22,9 @@ internal sealed partial class SqlConfigurableRetryLogicLoader
         /// and its related assemblies in default assembly load context if they aren't loaded yet.
         /// </summary>
         /// <returns>Resolved type if it could resolve the type; otherwise, the `SqlConfigurableRetryFactory` type.</returns>
+#if NET6_0_OR_GREATER
+        [RequiresUnreferencedCode("The type might be removed")]
```
Wraith2, line 26: Removed by who? is this something that the trimmer might remove or that we, the library authors, should remove?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbMetaDataFactory.cs
```diff
@@ -36,6 +37,9 @@ internal class DbMetaDataFactory
         private const string SqlCommandKey = "SQLCommand";
         private const string PrepareCollectionKey = "PrepareCollection";
 
+#if NET6_0_OR_GREATER
```
Wraith2, line 40: "Metadata loaded from XML stream may require types that have been trimmed away" ?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/Reliability/SqlConfigurableRetryLogicManager.NetCoreApp.cs
kant2002: image

Also this button allow you to directly change my bad English.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbMetaDataFactory.cs
```diff
@@ -36,6 +37,9 @@ internal class DbMetaDataFactory
...
+        [RequiresUnreferencedCode("Metadata loaded from XML stream may require types which was trimmed out")]
```
kant2002, line 41:
```suggestion
        [RequiresUnreferencedCode("Metadata loaded from XML stream may require types that have been trimmed away")]
```

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Reliability/SqlConfigurableRetryLogicManager.cs
```diff
@@ -15,18 +16,27 @@ internal sealed class SqlConfigurableRetryLogicManager
...
+        {
+            ISqlConfigurableRetryConnectionSection cnnConfig = null;
+            ISqlConfigurableRetryCommandSection cmdConfig = null;
 
-                // Fetch the section attributes values from the configuration section of the app config file.
-                cnnConfig = AppConfigManager.FetchConfigurationSection<SqlConfigurableRetryConnectionSection>(SqlConfigurableRetryConnectionSection.Name);
-                cmdConfig = AppConfigManager.FetchConfigurationSection<SqlConfigurableRetryCommandSection>(SqlConfigurableRetryCommandSection.Name);
+            // Fetch the section attributes values from the configuration section of the app config file.
+            cnnConfig = AppConfigManager.FetchConfigurationSection<SqlConfigurableRetryConnectionSection>(SqlConfigurableRetryConnectionSection.Name);
+            cmdConfig = AppConfigManager.FetchConfigurationSection<SqlConfigurableRetryCommandSection>(SqlConfigurableRetryCommandSection.Name);
```
kant2002, line 36: @Wraith2 starting from this two lines we apply `RequiresUnreferencedCode` across codebase. this bubble up to override of `DbDataReader` like `ExecuteScalar`/`ExecuteDbDataReader`, etc. In .NET DbDataReader does not marked with RUC so it is prohibited to mark these methods with RUC in derived classes. 

I'm not sure that App.config code path is actually excercised on .NET variant, but if so, I really would like to have different configuration for that. if no configuration in app.config then only one [instance of class](https://github.com/dotnet/SqlClient/blob/272219b0428e9dce078cc63c21cf108a94ed9050/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Reliability/SqlConfigurableRetryLogicLoader.cs#L134) is available here, so I really would like to streamline instantiation of that class without using all these "reflection" magic somehow.

src/Microsoft.Data.SqlClient/netcore/src/Common/src/System/Net/Logging/NetEventSource.Common.cs
```diff
@@ -518,6 +521,8 @@ private static string Format(FormattableString s)
         #region Custom WriteEvent overloads
 
         [NonEvent]
+        [SuppressMessage("ReflectionAnalysis", "IL2026:UnrecognizedReflectionPattern",
```
teo-tsirpanis, line 524: Should it be `UnconditionalSuppressMessage`? It's available on .NET 5+ so it needs an `#if` guard.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlClientDiagnosticListenerExtensions.cs
```diff
@@ -14,6 +15,8 @@ namespace Microsoft.Data.SqlClient
     /// </summary>
     internal static class SqlClientDiagnosticListenerExtensions
     {
+        private const string EventSourceSuppressMessage = "Parameters to this method are statically known and thus are trimmer safe";
```
DavoudEshtehari, line 18: Can you explain why suppression is required while the parameters are genuinely trimmer-safe? I expect the analyzer to distinguish it. VS recognizes them as an unnecessary suppression that could be removed though.

kant2002: IL trimmer made design decision operate on IL level, when it see virtual call he don't perform full application analysys to prove that this is actuall specific call. So each virtual call should have same annotations as base class. 

I know, probably that can be relaxed, but it is how it is. @vitek-karas maybe give more perspective why this is needed.

kant2002: ooh, here. I should not be that fast in answering. EventSource.Write annotated as RequireUnreferencedCode, and trimmer think that if somebody use such methods it's dangerous. It is expected suppress warning like that

src/Microsoft.Data.SqlClient/netcore/src/Common/src/System/Net/Logging/NetEventSource.Common.cs
vitek-karas: Yes - it MUST be `UnconditionalSuppressMessage`. `SuppressMessage` is "conditional" and the compiler will not emit it to IL - it disappears after compilation. As a result the trimmer doesn't get to see it.

We've been discussing some possibilities to provide an analyzer for these "bugs" but it's difficult (`SuppressMessage` is handled by the compiler intrinsically...). I could not find a tracking issue, so I created https://github.com/dotnet/runtime/issues/93453.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnectionFactory.cs
```diff
@@ -296,6 +297,13 @@ override internal void SetInnerConnectionTo(DbConnection owningObject, DbConnect
             }
         }
 
+#if NET6_0_OR_GREATER
+        [DynamicDependency(DynamicallyAccessedMemberTypes.All, typeof(GroupByBehavior))]
+        [DynamicDependency(DynamicallyAccessedMemberTypes.All, typeof(IdentifierCase))]
+        [DynamicDependency(DynamicallyAccessedMemberTypes.All, typeof(SupportedJoinOperators))]
+        [UnconditionalSuppressMessage("Trimming", "IL2026:Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code",
+            Justification = "All non-primitive types which are contained in the XML stream marked as dynamic dependencies.")]
```
vitek-karas, line 305: In cases like these we would generally recommend adding a unit test which validates the assumption here. So in this case a test which would deserialize the embedded XML and then walk the returned object graph (using reflection) and check that it only encountered expected set of types. You could go even the final step and read the expected set of types from the `DynamicDependencyAttribute` on this method.

The main goal of such a test is to make the code maintainable. If in the future somebody changes the XML and adds a new type into it, the assumption here would break, but there would be no way to tell (nothing would produce a warning). So the unit test acts as the guard of that.

We've done things like this in other repos as well - and it seems to work pretty well.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnectionFactory.cs
```diff
@@ -296,6 +297,13 @@ override internal void SetInnerConnectionTo(DbConnection owningObject, DbConnect
             }
         }
 
+#if NET6_0_OR_GREATER
+        [DynamicDependency(DynamicallyAccessedMemberTypes.All, typeof(GroupByBehavior))]
+        [DynamicDependency(DynamicallyAccessedMemberTypes.All, typeof(IdentifierCase))]
+        [DynamicDependency(DynamicallyAccessedMemberTypes.All, typeof(SupportedJoinOperators))]
```
vitek-karas, line 303: Is it really required to have `All`? `All` is weird in that it applies recursively to all base types and interfaces, so it can bring in a LOT of dependencies. If we're just reading these from XML, maybe some like "public/private properties, constructors and fields" may be enough?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnection.cs
```diff
@@ -121,6 +121,9 @@ private enum CultureCheckState : uint
         [DesignerSerializationVisibility(DesignerSerializationVisibility.Hidden)]
         public SqlRetryLogicBaseProvider RetryLogicProvider
         {
+#if NET6_0_OR_GREATER
+            [RequiresUnreferencedCode("RetryLogicProvider can be read from app.config which is unsafe for trimming")]
```
vitek-karas, line 125: Typical pattern is to define a const string field with the message and the use that. It makes it easy to improve the message and make it consistent across multiple usages. It also makes it clearer if the `RequiresUnreferencedCode` is "propagated" across multiple methods in a call chain - that they all belong together and are for the same reason.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbMetaDataFactory.cs
```diff
@@ -495,6 +499,9 @@ private bool IncludeThisColumn(DataColumn sourceColumn, string[] hiddenColumnNam
             return result;
         }
 
+#if NET6_0_OR_GREATER
+        [RequiresUnreferencedCode("LoadDataSetFromXml uses System.Data.DataSet.ReadXml(Stream)")]
```
vitek-karas, line 503: In general would avoid using messages like this - the message should be telling the caller of this method what is the problem. Describing that the problem is that this method is going to call another method is not very helpful to the caller.

I know that these are private methods, so it doesn't matter much, but that's only true now. 6 months from now if somebody adds a call to this method they will get a warning with this message. And it would effectively require them to go search through the code and across multiple methods to determine the real problem. It's also a good habit, to avoid doing this on a public API.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Sql/SqlDataSourceEnumeratorNativeHelper.cs
```diff
@@ -106,6 +107,10 @@ internal static DataTable GetDataSources()
             return ParseServerEnumString(strbldr.ToString());
         }
 
+#if NET6_0_OR_GREATER
+        [UnconditionalSuppressMessage("Trimming", "IL2026:Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code",
+            Justification = "Filter expression uses only primitive types and trimmer safe")]
+#endif
```
vitek-karas, line 113: I would recommend avoiding suppressions on large methods. These are problematic because:
* The suppression applies to the entire method body. It is typically intended to suppress one or two specific calls, but it covers everything. Later on if somebody adds a new code into the method, and introduces a new trim incompatibility, the suppression will still silence that warning as well - and the justification may be wrong at that time -> introducing a trim incompatibility problem.
* It's hard to read - for example I still can't tell exactly where the suppression applies and which primitive types it's talking about.

I would suggest to refactor the code such that the problematic call is isolated into a local function and the suppression is added onto the local function.

I know it makes the code a little less clean, but it's the price we have to pay for the fact that there's no way to add an attribute (or really any kind of metadata) onto a block of IL instructions.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Reliability/SqlConfigurableRetryLogicManager.cs
```diff
@@ -15,18 +16,27 @@ internal sealed class SqlConfigurableRetryLogicManager
     {
         private const string TypeName = nameof(SqlConfigurableRetryLogicManager);
 
+#if NET6_0_OR_GREATER
+        [UnconditionalSuppressMessage("Trimming", "IL2026:Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code",
+            Justification = "Usages of this property annotated with RequiresUnreferencedCode")]
```
vitek-karas, line 21: This feels wrong - if the usages are all annotated, then why not simply mark this with `RequiresUnreferencedCode`?

Also of note - this suppression will not work in the trimmer. The initialization code is actually part of the static constructor for the class in IL. This is a long standing problem we haven't been able to fix yet.

The simplest solution is probably to add an explicit static constructor, initialize the values there and suppress the warnings there as well.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Reliability/SqlConfigurableRetryLogicManager.cs
vitek-karas: Assuming it's uncommon to use the config file to provide a custom class, this whole thing might be a candidate for a [feature switch](https://github.com/dotnet/designs/blob/main/accepted/2020/feature-switch.md).

Basically disable this capability by default for trimmed apps. If the user enables it explicitly via a property in the csproj file, they would get a warning and would have to deal with it on their own.

But this requires somebody with the knowledge of how this is used - the feature switches are a reasonable solution only for rarely used capabilities. They also degrade the overall experience - instead of providing compile time warnings about trim incompatibilities, the failure will happen only at runtime, so they should be used only when other solutions are not viable.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/VirtualSecureModeEnclaveProvider.cs
```diff
@@ -57,6 +58,8 @@ protected override string GetAttestationUrl(string attestationUrl)
         }
 
         // Makes a web request to the provided url and returns the response as a byte[]
+        [SuppressMessage("ReflectionAnalysis", "IL2026:UnrecognizedReflectionPattern",
+                   Justification = "DataContractJsonSerializer used here deserialize only byte[] which is already kept.")]
```
vitek-karas, line 62: Use the local function refactoring to apply the suppression only onto the specific piece of code (in this case the two lines inside the using statement)

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlClientDiagnosticListenerExtensions.cs
vitek-karas: This is actually not correct and should not be suppressed as-is. This code for example means that OpenTelemetry's SqlClient instrumentation is not trim/AOT compatible. https://github.com/open-telemetry/opentelemetry-dotnet/blob/b091af1899b66a516d753109ce6bf8182be416e0/src/OpenTelemetry.Instrumentation.SqlClient/SqlClientInstrumentation.cs#L56C9-L61

Sorry it fell through cracks on my side, I meant to file an issue around this long time ago... so I finally did it: https://github.com/dotnet/SqlClient/issues/2184. Other libraries are in similar (although not as bad) problem, so we're slowly trying to get this fixed. OTel in its latest version is now fully AOT compatible, with the exception of the SqlClient instrumentation as there's no good way to make that work correctly right now.

As described in the issue - the best solution is to declare all of these payload types as public. That way the consumer (OTel for example) can just cast they payload instance to that type and directly access its properties. That would also fix all of the trimming issues related to in-proc consumers which are using the public types.

Separate issue is all the consumers which are either using reflection or rely on diagnostic event source to do the same. For those the only solution is to explicitely hint the trimmer to preserve certain properties. For example System.Http, ASP.NET and GRPC all do this to some degree. For example HTTP stack:
https://github.com/dotnet/runtime/blob/9c4d51c0a45a129745b2eec9fb5497042cd805fb/src/libraries/System.Net.Http/src/System/Net/Http/DiagnosticsHandler.cs#L325 - this preserves all top-level properties on the payload object
https://github.com/dotnet/runtime/blob/9c4d51c0a45a129745b2eec9fb5497042cd805fb/src/libraries/System.Net.Http/src/System/Net/Http/DiagnosticsHandler.cs#L272-L276 - this preserves some nested properties on one of the payload objects.

This is not a precise solution and never will be, but it's worth preserving at least properties used by OTel and other popular consumers (personally I don't know what they would be though).

David-Engel:
@kant2002 Can you address vitek-karas' comments?

I also recommend pulling/merging from main, too, as there is a fix that is needed to avoid validation pipeline failures now.

PR #1959, opened 2023-03-21
Fix bunch of annontations

kant2002:
Actually remove some trimming issues completely.


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
```diff
@@ -1232,6 +1237,11 @@ override public IEnumerator GetEnumerator()
         }
 
         /// <include file='../../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlDataReader.xml' path='docs/members[@name="SqlDataReader"]/GetFieldType/*' />
+#if NET6_0_OR_GREATER
+        [SuppressMessage("ReflectionAnalysis", "IL2093:MismatchOnMethodReturnValueBetweenOverrides",
```
Wraith2, line 1241: What kind of annotations? and were they shipped in 7?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SmiMetaData.cs
```diff
@@ -942,6 +964,9 @@ ParameterDirection direction
             byte scale,
             long localeId,
             SqlCompareOptions compareOptions,
+#if NET6_0_OR_GREATER
```
Wraith2, line 967: I don't like that this isn't clear that it applies to the parameter following the #endif but i can't see another way to format the code to make it clear :(

kant2002: Yeah, that's unfortunate.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
kant2002: I think issue was that ref assembly was missing this annotations.
https://github.com/dotnet/runtime/blame/main/src/libraries/System.Data.Common/ref/System.Data.Common.cs#L2323-L2324

and yes, it was fixed in .NET 7 (or even .NET 8)

kant2002:
@lcheunglci May you take a look?

kant2002:
@lcheunglci I still have hopes I catch your attention, and move whole thing a bit forward 😄

lcheunglci:
> @lcheunglci I still have hopes I catch your attention, and move whole thing a bit forward 😄

Sorry I haven't been as active lately. Perhaps, @JRahnama @DavoudEshtehari or @Kaur-Parminder when they have a moment could also take a look.

kant2002:
Okay. Sorry to bother, probably misinterpret your like

src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
```diff
@@ -18,8 +18,8 @@
     <DocumentationFile>$(OutputPath)\$(TargetFramework)\Microsoft.Data.SqlClient.xml</DocumentationFile>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <Product>Core $(BaseProduct)</Product>
-    <EnableTrimAnalyzer>true</EnableTrimAnalyzer>
-    <NoWarn>$(NoWarn);IL2026;IL2057;IL2067;IL2070;IL2072;IL2075;IL2077;IL2080;IL2093;IL2111</NoWarn>
+    <EnableTrimAnalyzer Condition="$(TargetFramework)=='net6.0'">true</EnableTrimAnalyzer>
```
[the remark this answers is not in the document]
kant2002: Done! Thanks for the tip.

kant2002:
It looks like failure are from unreliable tests? I have hard time understand how annotations can be the cause.

DavoudEshtehari (approved):
For the record, can you add more context about the way this change can affect and the justifications?

kant2002:
When not ignoring trimming anotations does not affect runtime behavoir of the code, but affect dev workflow.
If somebody working on SqlClient code and would like to add code which will use reflection he has to [annotate that code for trimming](https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/prepare-libraries-for-trimming) and make sure that trimming analyzer would not complain.

For developer who consume SqlClient it will mean that it can use this library with trimming and know that if analyzer do not produce warnings, then they are safe. Right now if you use trimming/NativeAOT you should pray that SqlClient is trimming friendly (I would say it is, but now we will have proof for that)

JRahnama (approved):

PR #1960, opened 2023-03-21
Fix | TransactionScope connection issue when Enlist is enable and SQL MI Connection is set to Redirect

lcheunglci:
We found that there was a bug when using a SQL Managed Instance when the connection type is set from `proxy` to `redirect` where connection pooling is disabled and enlist is enabled would result in `Failed: Error Code: 10054 - A transport-level error has occurred when receiving results from the server. (provider: TCP Provider, error: 0 - An existing connection was forcibly closed by the remote host.)`. By adding a check to ensure `RoutingInfo` is null before `Enlist` is called, it resolves the issue.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
@@ -1235,7 +1235,7 @@ namespace Microsoft.Data.SqlClient
 
             // for non-pooled connections, enlist in a distributed transaction
             // if present - and user specified to enlist
-            if (enlistOK && ConnectionOptions.Enlist)
+            if (enlistOK && ConnectionOptions.Enlist && RoutingInfo == null)
             {
                 _parser._physicalStateObj.SniContext = SniContext.Snix_AutoEnlist;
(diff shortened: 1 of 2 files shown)


lcheunglci:
I find it strange that we do have a test for `pooling=false` and enlist by default is `true`, but this test passed before. https://github.com/dotnet/SqlClient/blob/1bcdb235e3952331698cfa2d45402a627122b357/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/TransactionTest/TransactionTest.cs#L52

David-Engel:
> I find it strange that we do have a test for `pooling=false` and enlist by default is `true`, but this test passed before.
> 
> https://github.com/dotnet/SqlClient/blob/1bcdb235e3952331698cfa2d45402a627122b357/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/TransactionTest/TransactionTest.cs#L52

The test specifically doesn't run against Azure SQL because Azure SQL used to not support distributed transactions. I think it does now, via elastic transactions. So we probably just need to remove the `nameof(DataTestUtility.IsNotAzureServer)` from the ConditionalTheory. There might be other tests we can try enabling, too, although I don't know if there is functionality difference between the different .NET versions or if distributed transaction support has all been enabled by changes on the server-side.

David-Engel (approved):

JRahnama (approved):

PR #1974, opened 2023-03-29
Fix | Dropping reference to Microsoft.Win32.Registry in netcore.

JRahnama:
According to [this announcement](https://github.com/dotnet/runtime/issues/54365) Microsoft.Win32.Registry is shipped with net 6 and after versions. Removing the reference in netcore csproj will eliminate the warning.

### DIFF
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectivityTests/ConnectivityTest.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectivityTests/ConnectivityTest.cs
@@ -389,20 +389,17 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests
         }
 
         [PlatformSpecific(TestPlatforms.Windows)]
-        [ConditionalFact(typeof(DataTestUtility), nameof(DataTestUtility.AreConnStringsSetup))]
+        [ConditionalFact(typeof(DataTestUtility), nameof(DataTestUtility.AreConnStringsSetup), nameof(DataTestUtility.IsSQLAliasSetup))]
         public static void ConnectionAliasTest()
         {
-            if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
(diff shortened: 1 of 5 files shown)


ErikEJ:
What does this fix?

JRahnama:
> What does this fix?

Warnings on build time and also on Visual studio in referenced packages.

David-Engel:
@JRahnama 
Make sure to run the ConnectionAliasTest test manually. It gets skipped in the pipeline since the agent doesn't have permission to create aliases.
image
Alternatively, you could refactor the pipeline and test to rely on a static alias that we set up at the beginning of the pipeline and define in the config.json.

JRahnama:
> @JRahnama Make sure to run the ConnectionAliasTest test manually. It gets skipped in the pipeline since the agent doesn't have permission to create aliases. image Alternatively, you could refactor the pipeline and test to rely on a static alias that we set up at the beginning of the pipeline and define in the config.json.

I did check the test. With modifying `DataSource` from `localhost` to `FQDN` and setting `Encrypt` to false it ran successfully.

lcheunglci (approved):

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectivityTests/ConnectivityTest.cs
```diff
@@ -4,6 +4,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Data;
```
David-Engel, line 7: Remove and sort all usings. A few will go away.

David-Engel (changes requested):

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectivityTests/ConnectivityTest.cs
JRahnama: done

David-Engel (approved):

PR #1979, opened 2023-03-30
Release notes 5.0.2 

lcheunglci:
Add release notes for 5.0.2 Hotfix release.


Kaur-Parminder (approved):

ErikEJ:
Is this not in 5.0.2 ?? https://github.com/dotnet/SqlClient/issues/1810

roji:
Yeah, to second @ErikEJ, #1810 (memory leak because of cancellation token) is quite serious, and it would be good to see a backport of the fix to 5.0.x. FYI the EF SQL Server 7.0 provider depends on 5.0.1, and we generally don't bump minor/major versions of dependencies in our own patch versions. In other words, while EF Core 8.0 may depend on 5.1.x (where the bug is already fixed), current users of EF 7 will still have this issue.

/cc @ajcvickers

lcheunglci:
Thanks @ErikEJ and @roji for pointing that out the dependency from EF Server 7, and looks like we backported #1785, which resulted in that regression in v5.0.1, and I've created backport PR #1980 with the changes from #1810 and I've CC'ed @David-Engel to confirm.

ErikEJ:
@lcheunglci Thanks a lot!

David-Engel (approved):

PR #1980, opened 2023-03-31
[5.0.2] Fix | Adding disposable stack temp ref struct and use

lcheunglci:
Backporting fix to 5.0-servicing branch.  This fixes the memory leak in the CancellationToken introduced from #1785 to address #1810.  These changes are based on #1818.


David-Engel (approved):

ErikEJ:
@David-Engel @lcheunglci Really appreciate your response to community feedback.

lcheunglci:
Thanks @ErikEJ for reminding us. We're glad that you were able to spot it since 5.0 will reach EOS in July, and it's uncertain if there'll be a 5.0.3 in time.

ErikEJ:
@roji Interesting. So EF Core 7 will depend on an unsupported SqlClient version...

Issue #1982, opened 2023-04-05
Transient Fault Handling is disabled by default with SqlConnection.OpenAsync()


cheenamalhotra:
### Describe the bug
When opening a connection with `SqlConnection.OpenAsync()`, transient fault handling is not enabled.
Test with placing breakpoints on below locations:
https://github.com/dotnet/SqlClient/blob/a5ad8382a390ea599740bd3ab51b19601bbd92e0/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnection.cs#L1859
https://github.com/dotnet/SqlClient/blob/a5ad8382a390ea599740bd3ab51b19601bbd92e0/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnection.cs#L1888

`_applyTransientFaultHandling` is set to `false` as `retry` is not `null` in this method when opening connection asynchronously.
When opening connection with `Open()`, `_applyTransientFaultHandling` is set to `true` as expected.

David-Engel:
Thanks for filing the issue and submitting a PR for the fix!

PR #1983, opened 2023-04-05
Fix Transient fault handling issue with OpenAsync

cheenamalhotra:
Addresses #1982 

'Retry', is a completion callback that is completed after `CreateConnection()` completes (which handles transient faults), here:

https://github.com/dotnet/SqlClient/blob/a5ad8382a390ea599740bd3ab51b19601bbd92e0/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/ProviderBase/DbConnectionPool.cs#L1079-L1095

So I don't see a reason to not enable transient fault handling retries when its specified.

### DIFF
--- a/src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlConnectionBasicTests.cs
+++ b/src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlConnectionBasicTests.cs
@@ -7,6 +7,7 @@ using System.Data;
 using System.Data.Common;
 using System.Reflection;
 using System.Security;
+using System.Threading.Tasks;
 using Microsoft.SqlServer.TDS.Servers;
 using Xunit;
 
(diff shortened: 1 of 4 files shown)


David-Engel:
Nice. Thanks!

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlConnectionBasicTests.cs
```diff
@@ -34,6 +35,33 @@ public void IntegratedAuthConnectionTest()
...
+            using SqlConnection connection = new(builder.ConnectionString);
+            try
+            {
+                await connection.OpenAsync();
+                Assert.Equal(ConnectionState.Open, connection.State);
+            }
+            catch (Exception e)
+            {
+                Assert.False(true, e.Message);
+            }
```
DavoudEshtehari, line 62: In the case of failure, the exception has better information.
```suggestion
            await connection.OpenAsync();
            Assert.Equal(ConnectionState.Open, connection.State);
```

DavoudEshtehari (approved):

PR #1985, opened 2023-04-10
Code Health | Cleanup `SNIMARSConnection`

Wraith2:
In https://github.com/dotnet/SqlClient/pull/933 many changes were made to SNIMarsConnection which were then lost when that PR was reverted. This PR implements the harmless and side-effect-less changes such as using scoped event tracing like the rest of SNI does, cleaning up a class name variable and adding a dedicated sync object (there are no locks on instances of this class outside this file, i've checked).

This does not change the demuxing code in any way. I want to get the files cleaned up and then try again with the muxing code in a future PR. this is a simple way to get easy changes in place minimizing review complexity.


Wraith2:
test failure is the usual unreliable reliability test and 3 legs failed to configure their firewall.

David-Engel (approved):

JRahnama (approved):

PR #2003, opened 2023-04-18
[4.0.3] | Upgrade "System.Drawing.Common" version

DavoudEshtehari:
- Addresses the [remote code execution vulnerability](https://github.com/dotnet/core/blob/main/release-notes/5.0/5.0.3/5.0.3.md) that is reported in `System.Drawing.Common` by increasing the version of dependent packages. The following packages depend on this package:
  - System.Configuration.ConfigurationManager
  - System.Runtime.Caching
  - System.Security.Permissions

- `System.Security.Permissions` is added indirectly with `System.Configuration.ConfigurationManager`.


lcheunglci (approved):

JRahnama (approved):

Kaur-Parminder (approved):

DavoudEshtehari:
This vulnerability exists when parsing certain types of graphics files on Linux or MacOS, which does not affect SqlClient. To avoid breaking the current consumers we decided not to include this change. To remediate it, applications should use `System.Drawing.Common` v5.0.3 and about.

Issue #2001, opened 2023-04-17
SqlBulkCopy.RowsCopied overflows because it is an int


madelson:
### Describe the bug
`SqlBulkCopy.RowsCopied` is supposed to return "The number of rows processed in the ongoing bulk copy operation". However, because it is `Int32` rather than `Int64` like `SqlRowsCopiedEventArgs.RowsCopied`, it is prone to overflow into the negative with very large copy operations (something we're currently running into).

This makes it difficult to track the actual number of rows copied.

Furthermore, the internal `int` field behind this property also flows into the event args, resulting in negative values there even though the data type is `long`.

### To reproduce
* Run a SqlBulkCopy operation with > int.MaxValue rows.
* Observe SqlBulkCopy.RowsCopied after the fact (and/or observe the SqlRowsCopied event's args)

### Expected behavior
Ideally, both properties would be `long` and would only overflow after `long.MaxValue` rows copied, which seems extremely unlikely to happen. Given that we probably can't change the type of the `RowsCopied` property at this point, what would be nice to see is:
* Internally the rows are tracked as `long` so that the event args has the right value
* A new `long RowsCopied64` property is added to `SqlBulkCopy` that returns the underlying value as `long`.

### Further technical details
Microsoft.Data.SqlClient version: multiple
.NET target: .NET 6
SQL Server version: SQL Server 2019
Operating system: Windows 10

JRahnama:
@madelson thanks for opening the issue. We will look into this and will get back to you soon.

Wraith2:
> Given that we probably can't change the type of the `RowsCopied` property at this point, what would be nice to see is:
> 
> * Internally the rows are tracked as `long` so that the event args has the right value
> * A new `long RowsCopied64` property is added to `SqlBulkCopy` that returns the underlying value as `long`.

This seems reasonable. 
What do we do with the existing `int RowsCopied` property if the value exceeds int.MaxValue? we can either leave the behaviour as it is currently or clamp it to int.MaxValue.

ErikEJ:
Leave as is, people may rely on the current behaviour

Wraith2:
That works for me. Doing nothing is always my preferred option 😁

ErikEJ:
It should be.

roji:
[...]

PR #2004, opened 2023-04-18
Add RowsCopied64 property, refs, and docs

Wraith2:
Closes https://github.com/dotnet/SqlClient/issues/2001

Changes the `SqlBulkCopy._rowscopied` variable to a long and adds a new `long SqlBulkCopy.RowsCopied64` property to the class. 
The existing `int SqlBulkCopy.RowsCopied` behaviour of going negative is preserved. I have changed the implementation to be `unchecked((int)_rowsCopied);` to make it clear in the source that the unchecked behaviour is deliberate. The unchecked is not strictly required because it is the default compile option. 
Is keeping the existing behaviour desiable? It could be changed to throw an overflow exception. Keeping the existing behaviour is most compatible in case someone has taken a dependency on the existing behaviour. It is not the most correct.

### DIFF
--- a/doc/snippets/Microsoft.Data.SqlClient/SqlBulkCopy.xml
+++ b/doc/snippets/Microsoft.Data.SqlClient/SqlBulkCopy.xml
@@ -503,14 +503,15 @@ and faster to use a Transact-SQL `INSERT … SELECT` statement to copy the data.
 				property.
 			</value>
 			<remarks>
-				<format type="text/markdown"><![CDATA[
+				<format type="text/markdown">
+  <![CDATA[
 
 ## Remarks
 This value is incremented during the <xref:Microsoft.Data.SqlClient.SqlBulkCopy.SqlRowsCopied> event and does not imply that this number of rows has been sent to the server or committed.
 
 This value can be accessed during or after the execution of a bulk copy operation.
 
-This value can become negative if the number of rows exceelds int.MaxValue. If this may happen use the RowsCopied64 property.
+This value will wrap around and become negative if the number of rows exceeds int.MaxValue. Consider using the <xref:Microsoft.Data.SqlClient.SqlBulkCopy.RowsCopied64> property.
 ]]></format>
 			</remarks>
 		</RowsCopied>
--- a/doc/snippets/Microsoft.Data.SqlClient/SqlBulkCopy.xml
+++ b/doc/snippets/Microsoft.Data.SqlClient/SqlBulkCopy.xml
@@ -509,9 +509,30 @@ and faster to use a Transact-SQL `INSERT … SELECT` statement to copy the data.
 This value is incremented during the <xref:Microsoft.Data.SqlClient.SqlBulkCopy.SqlRowsCopied> event and does not imply that this number of rows has been sent to the server or committed.
 
 This value can be accessed during or after the execution of a bulk copy operation.
+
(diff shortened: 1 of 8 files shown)


doc/snippets/Microsoft.Data.SqlClient/SqlBulkCopy.xml
```diff
@@ -509,9 +509,30 @@ and faster to use a Transact-SQL `INSERT … SELECT` statement to copy the data.
 This value is incremented during the <xref:Microsoft.Data.SqlClient.SqlBulkCopy.SqlRowsCopied> event and does not imply that this number of rows has been sent to the server or committed.
 
 This value can be accessed during or after the execution of a bulk copy operation.
+
+This value can become negative if the number of rows exceelds int.MaxValue. If this may happen use the RowsCopied64 property.
```
madelson, line 513: * `exceelds` -> `exceeds` (typo)
* `RowsCopied64` should have some kind of ref tag?
* I feel like this language should be stronger: "This value will wrap around and become negative if... Applications should use RowsCopied64 instead".

Wraith2: Changed to:
`This value will wrap around and become negative if the number of rows exceeds int.MaxValue. Considering using the <xref:Microsoft.Data.SqlClient.SqlBulkCopy.RowsCopied64> property.`

If the tone needs to be stronger than that I'll leave the exact wording to the MS team.

madelson: I like your rewording. As an app developer I would just always use RowsCopied64 now that it exists since there's no cost and real upside.

doc/snippets/Microsoft.Data.SqlClient/SqlBulkCopy.xml
```diff
@@ -503,15 +503,37 @@ and faster to use a Transact-SQL `INSERT … SELECT` statement to copy the data.
...
-				<format type="text/markdown"><![CDATA[
+				<format type="text/markdown">
+  <![CDATA[
 
 ## Remarks
 This value is incremented during the <xref:Microsoft.Data.SqlClient.SqlBulkCopy.SqlRowsCopied> event and does not imply that this number of rows has been sent to the server or committed.
 
 This value can be accessed during or after the execution of a bulk copy operation.
+
+This value will wrap around and become negative if the number of rows exceeds int.MaxValue. Considering using the <xref:Microsoft.Data.SqlClient.SqlBulkCopy.RowsCopied64> property.
```
ErikEJ, line 514: "Consider"

Wraith2: Fixed.

David-Engel (approved):

JRahnama:
LGTM, Is it possible to add some tests as well?

Wraith2:
Possible, yes. But why? it's a tiny property with an entirely obvious implementation.

JRahnama:
> Possible, yes. But why? it's a tiny property with an entirely obvious implementation.

We are trying to improve the code coverage and each line of new code added with no test will cause some shortage in the code coverage percentage. That is the only reason.

Wraith2:
I've altered the two existing tests which cover RowsCopied to also check RowsCopied64.

JRahnama (approved):

PR #2012, opened 2023-04-20
Add | Use Minimum Login Timeout as 1 sec in .NET Core and enable behavior by default

cheenamalhotra:
This PR brings over app context switch "UseOneSecFloorInTimeoutCalculationDuringLogin" to use 1 second instead of 0 seconds as minimum login timeout to prevent indefinite timeouts and potential application hangs.

Also enables the app context switch by default, in both NetFx and NetCore.

**NOTE: [Public documentation for this App Context switch](https://learn.microsoft.com/en-us/sql/connect/ado-net/appcontext-switches?view=sql-server-ver16#enable-a-minimum-timeout-during-login) needs to enable support in .NET Core and .NET Standard when this change gets released.**


Wraith2 (approved):

JRahnama:
LGTM, but why are we adding an app context switch to address a bug? Whoever is facing this edge case situation must have suffered from it. Is this just to avoid breaking changes?

David-Engel:
> LGTM, but why are we adding an app context switch to address a bug? Whoever is facing this edge case situation must have suffered from it. Is this just to avoid breaking changes?

The switch was already there. This just brings it to netcore and turns it on, by default. Keeping it around will let customers turn it off, just in case there was a good reason someone added it to netfx in the past. When we were talking about it, we couldn't think of a good reason the change was put under a context switch.

It seemed like there were two possible side effects of the change under the switch.
- It could slightly extended a connect timeout time. (But who would actually notice this?)
- Connection timeout errors could occur when slow connections right on the edge of your connect timeout might have previously succeeded beyond the timeout. (Fix: Extend your timeout if you were expecting this behavior.)

David-Engel (approved):

JRahnama (approved):

PR #2063, opened 2023-06-15
Dev | Add support of the new authentication API with SSPI token on Managed SNI

DavoudEshtehari:
- Consumed [NegotiateAuthentication](https://learn.microsoft.com/en-us/dotnet/api/system.net.security.negotiateauthentication?view=net-7.0) type on Managed SNI, which has been shipped with .net 7.0 and above.


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObjectManaged.cs
```diff
@@ -384,15 +389,26 @@ internal override uint SetConnectionBufferSize(ref uint unsignedPacketSize)
...
         {
-            if (_sspiClientContextStatus is null)
+#if NET7_0_OR_GREATER
+            _negotiateAuth ??= new(new NegotiateAuthenticationClientOptions { Package = "Negotiate", TargetName = Encoding.Unicode.GetString(_sniSpnBuffer[0]) });
+            sendBuff = _negotiateAuth.GetOutgoingBlob(receivedBuff, out NegotiateAuthenticationStatusCode statusCode)!;
+            SqlClientEventSource.Log.TryTraceEvent("TdsParserStateObjectManaged.GenerateSspiClientContext | Info | Session Id {0}, StatusCode={1}", _sessionHandle?.ConnectionId, statusCode);
+            if (statusCode is not NegotiateAuthenticationStatusCode.Completed and not NegotiateAuthenticationStatusCode.ContinueNeeded)
             {
-                _sspiClientContextStatus = new SspiClientContextStatus();
+                throw new InvalidOperationException(SQLMessage.SSPIGenerateError() + "\n" + statusCode);
```
Wraith2, line 404: Instead of "\n" use Environment.NewLine to get the right one for the executing platform.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObjectManaged.cs
```diff
@@ -384,15 +389,26 @@ internal override uint SetConnectionBufferSize(ref uint unsignedPacketSize)
...
-        internal override uint GenerateSspiClientContext(byte[] receivedBuff, uint receivedLength, ref byte[] sendBuff, ref uint sendLength, byte[][] _sniSpnBuffer)
+        internal override uint GenerateSspiClientContext(byte[] receivedBuff,
+                                                         uint receivedLength,
+                                                         ref byte[] sendBuff,
+                                                         ref uint sendLength,
+                                                         byte[][] _sniSpnBuffer)
         {
-            if (_sspiClientContextStatus is null)
+#if NET7_0_OR_GREATER
+            _negotiateAuth ??= new(new NegotiateAuthenticationClientOptions { Package = "Negotiate", TargetName = Encoding.Unicode.GetString(_sniSpnBuffer[0]) });
```
Wraith2, line 399: Is this really intended to get a string from a single byte of data? That seems like a strange thing to do.

DavoudEshtehari: It's an array of byte array.

Wraith2: Ah, looks ok then.

David-Engel:
@DavoudEshtehari This is great! Can we condition out the NET6-only files/classes/code related to the previous calls so that they aren't even compiled into the NET7+ binary?

David-Engel (approved):

JRahnama (approved):

PR #2068, opened 2023-06-20
Code Health | Enforcing Ordinal for  StringComparison

JRahnama:
@Wraith2: based on the other conversation we had on PR #2065, what do you think on this change request?


ErikEJ:
I was just going to ask if there were other potential  relevant changes - LGTM

Wraith2:
Looks like a good idea to me.

.editorconfig
```diff
@@ -143,6 +143,9 @@ dotnet_diagnostic.CA2100.severity = silent
 # CA1416: Validate platform compatibility
 dotnet_diagnostic.CA1416.severity = silent
 
+# CA1310: Specify StringComparison for correctness
+dotnet_diagnostic.CA1310.severity = error
```
David-Engel, line 147: As you can see from the build results, we should consider putting this in a .editorconfig file in various subfolders to only apply it to libraries that require it.

YES: MDS, MSS, MDS AKV Provider
MAYBE: Test utility libraries
NEEDS EVAL: MDS Test libraries (I'm guessing a lot of places in test code should be specifying the option, but not sure if all require it, so we might not want to impose this rule on tests.)

David-Engel:
To add background for anyone investigating this in the future: In a library like MDS, almost all string comparison operations being done inside the library should not use culture comparison rules (they should use Ordinal). For example, if we are parsing a connection string, we are looking for specific characters, not culture-specific representations of them. For example, if we are looking for a comma (also a thousands separator in some languages), we only want to find a comma, not a period (also a thousands separator in some languages). Default string comparison operations in .NET are culture specific. To be safe, we should always be explicit with which option we want to use inside MDS.

.editorconfig
JRahnama: that is what I was looking for, but nothing on the .editorconfig subfolders helped me. so I went through all the tests and noticed there are almost 17 classes that needs to be converted to ordinal comparison. There were couple of places that ignore case were required. I have changed them all. If you still think that I should go with the editorconfig setup. I will revert these changes and will go with that pattern.

David-Engel (approved):

Kaur-Parminder (approved):

PR #2090, opened 2023-07-17
Test | Fix TdsParserStateObjectHelper debug mode

DavoudEshtehari:
`AsyncDebugScope` related tests are only supported in Debug mode.


David-Engel (approved):
Approved. But is there a way we could detect when variable renames affect tests that reference them via reflection?

DavoudEshtehari:
> Approved. But is there a way we could detect when variable renames affect tests that reference them via reflection?

I'm afraid not for the non-public types and members in a separate assembly without modifying the driver.

Kaur-Parminder (approved):

PR #2095, opened 2023-07-20
Reimplement rework cancel

Wraith2:
Reimplementation of https://github.com/dotnet/SqlClient/pull/956 to see what fails and why. The extra test added in https://github.com/dotnet/SqlClient/pull/1352 passes locally.

/cc @cheenamalhotra in case you're interested or can help pinpoint what's going on here

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
@@ -1479,19 +1479,13 @@ namespace Microsoft.Data.SqlClient
             else
             {
                 ThrowIfReconnectionHasBeenCanceled();
-                // lock on _stateObj prevents races with close/cancel.
-                // If we have already initiate the End call internally, we have already done that, so no point doing it again.
-                if (!_internalEndExecuteInitiated)
+                if (!_internalEndExecuteInitiated && _stateObj != null)
                 {
-                    lock (_stateObj)
-                    {
-                        return EndExecuteNonQueryInternal(asyncResult);
-                    }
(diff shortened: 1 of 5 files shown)


cheenamalhotra:
I started taking a look too, will find time sometime soon to review this. I was also thinking of an alternative way to pass cancellation token via Async context all the way to TdsParser.. Have you considered that as an option?

Wraith2:
The CI failed on the test that was added to show that the rework was flawed. I spent some time trying to identify what's going wrong just by reading the test and looking at the code. This might be wrong because i can't reproduce the test failure locally.

I _think_ that the inner connection is closed but still needs to drain the packet data that it has available. It is put into the connection pool and picked up again causing the old packet data to available to the new query which we see as a difference in the query result output parameter. 

What I could really do with is a reliable reproduction. I'm going to spin up an azure db over the weekend if I can to see if latency allows me to see the problem locally.

> I started taking a look too, will find time sometime soon to review this. I was also thinking of an alternative way to pass cancellation token via Async context all the way to TdsParser.. Have you considered that as an option?

I hadn't considered that. It sounds like a good idea to investigate.

patrick-yates-redgate:
Interested to know if there are any plans to progress this further?

@cheenamalhotra / @Wraith2 given your recent contributions have either of you gotten any further with this issue?

Wraith2:
No. I don't have a way to replicate the problem so i can't investigate further.

patrick-yates-redgate:
> No. I don't have a way to replicate the problem so i can't investigate further.

Thank you for the quick response, much appreciated.

With the original issue #44 being specific to while loops, in relation to this bug do you know if we should be concerned about Cancel causing deadlock for other types of queries?

cheenamalhotra:
Closing stale PRs, please open a new one when ready with feedback implemented.

PR #2107, opened 2023-07-27
Dev | Exposing IsActive as Public API

Kaur-Parminder:
- Added Public method for IsActive (renamed from IsConnectionAlive). Added related documentation and unit tets.


roji:
@Kaur-Parminder would it be possible to get some context on this new API? I'm specifically interested in the exact meaning of "Active" and the intended usage.

Since IsActive() is a synchronous method, I'm assuming it doesn't actually do a roundtrip to SQL Server to check that the connection is still up; if so, then "active" would guarantee that an operation on the connection would actually succeed, right?

In any case, any such check of the connection being alive/usable is inherently prone to race conditions: the connection may very well be usable right when IsActive() is called, but broken 1 microsecond later when the user attempts to actually use it. So exposing such an API would falsely lead users to write code that isn't safe, no?

Otherwise, if IsActive() doesn't actually check the network connection, how does this different from simply checking the connection's state?

Kaur-Parminder:
> @Kaur-Parminder would it be possible to get some context on this new API? I'm specifically interested in the exact meaning of "Active" and the intended usage.
> 
> Since IsActive() is a synchronous method, I'm assuming it doesn't actually do a roundtrip to SQL Server to check that the connection is still up; if so, then "active" ~would~ wouldn't guarantee that an operation on the connection would actually succeed, right?
> 
> In any case, any such check of the connection being alive/usable is inherently prone to race conditions: the connection may very well be usable right when IsActive() is called, but broken 1 microsecond later when the user attempts to actually use it. So exposing such an API would falsely lead users to write code that isn't safe, no?
> 
> Otherwise, if IsActive() doesn't actually check the network connection, how does this different from simply checking the connection's state?

@roji, IsActive name is still to be finalised and is open to discussion. The purpose of API is to check the state of socket , used to communicate with SQL Server, whether it is Connected or Disconnected. There is possibility the Connection is Open and Socket is Disconnected OR Connected.

This is already being used by Connection pool resiliency to see connection is alive/active. Having this visibility would allow to check that state before running things like close window checks and avoid long hangs by those checks incurring ICR delays.

For the race condition, I am working on case to handle it and add Unit tests to cover the scenario.

roji:
Thanks for the context.

Just to be sure: you can indeed probably look at the socket state as an indication that the socket is Disconnected (and therefore there's no point doing anything with that connection).

However, the opposite doesn't hold: if the socket says it's connected, that's no guarantee of any kind that a subsequent operation on that connection would succeed. As I wrote above, this is both because a connected socket does not imply that the connection is up (only actually sending and receiving data checks that), and also because the socket may still be up when IsActive is called, but be down 1 microsecond later.

So if the intent for this property is to be used for negative checks only (i.e. avoid doing something if we know the connection is down), that *could* make sense, but it cannot be used to reliably detect whether a connection is up.

> avoid long hangs by those checks incurring ICR delays

If the socket state really is disconnected, then trying to execute anything on it should fail immediately with an exception (rather than causing a delay). Are you sure you're seeing a delay when the socket is reported as being disconnected?

Because of all the above, I think this feature would generally not be very useful - the common practice is to try whatever operation you need to do, and expect the possibility of failure (which you always need to expect regardless of what yuo do).

roji:
@David-Engel thanks for the added context, that helps to understand the motivation here.

However, if the goal is to avoid long delays because of retries, then this solution (checking the socket state) is very partial. As I wrote above, the socket state only tells you "Disconnected" in a (probably small) subset of scenarios; the connection may very well have been dropped by an intemediate network device without the client machine (and SqlClient) ever knowing about it.

In other words, even you do this and IsActive returns true to SSMS, there's a big chance it would still hang on the timeouts since the connection is down regardles.

> Another option I proposed internally was to add an option similar to the one we added to SqlConnection.Open() to signal that the driver shouldn't use ICR and instead fail fast. But there would be a lot of potential execute methods to change for that.

That seems like the right option to me... If SqlClient has a built-in resiliency check (ICR) and the user (SSMS) isn't interested in it, they should be able to simply disable it, no?

David-Engel:
@roji Discussion is good. 😃 

> If that's done, then when SSMS tries to execute something on it, wouldn't that immediately throw because the connection is broken?

Avoiding that exception is the main purpose of ICR.

Just because the socket is disconnected doesn't mean SqlClient can't successfully restore the connection transparently to the application (avoiding the need for the application to have its own resiliency logic). So we don't want to just doom it. SqlClient won't know it's actually doomed until a reconnect fails. I'd guess most of the time, idle reconnects are successful, since the purpose is to transparently handle _idle_ connection timeouts.

There is an additional subtlety to the scenario. An "idle" connection that ICR can restore is also one that doesn't have an open transaction on it. Since the SSMS check is for `@@TRANSCOUNT > 0`, if the socket is disconnected, ICR wouldn't be able to restore the connection anyway. (ICR could try and still fail relatively quickly if the server is available.)  SSMS can skip the check and just close the window straight away knowing that any open transaction the user had is already lost.

But the main pain point here is to avoid the delay that happens when ICR is attempted against a server that is no longer available. Prior to trying to execute a query on an open connection, SqlClient doesn't know the difference between a disconnected socket on a connection that can be quickly restored against an available server and one against a server that isn't available or where the connection can't be restored.

ajcvickers:
>  SqlClient doesn't know the difference between a disconnected socket on a connection that can be quickly restored against an available server and one against a server that isn't available or where the connection can't be restored.

I think the point is that this can _never be known_ when working with sockets. The socket can never know if there is _really_ a connection to the server until it tries to use it. There is nothing that can be exposed by a flag that will change this general case. At best, it will fix only a subset of cases.

PR #2131, opened 2023-08-21
SqlSequentialStream position

SimonCropp:
Add support for [Stream.Position](https://learn.microsoft.com/en-us/dotnet/api/system.io.stream.position) to SqlSequentialStream 

currently `SqlSequentialStream.Position` throws a NotSupported.

Many APIs that consume Stream assume that a get on `position` will not throw. Most recently for me is the Aspose suite of libraries. But i have been hit by this at least 3 times in the past. each time i have had to wrap SqlSequentialStream in another  stream that keeps track of the position.

i figured i would finally fix the root cause

### DIFF
--- a/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlSequentialStreamSmi.cs
+++ b/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlSequentialStreamSmi.cs
@@ -50,8 +50,16 @@ namespace Microsoft.Data.SqlClient
 
         public override long Position
         {
-            get { throw ADP.NotSupported(); }
-            set { throw ADP.NotSupported(); }
+            get => _position;
+            set
+            {
+                if (_position == value)
+                {
+                    return;
+                }
+
+                throw ADP.NotSupported();
+            }
         }
 
         internal int ColumnIndex
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlSequentialStream.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlSequentialStream.cs
@@ -14,6 +14,7 @@ namespace Microsoft.Data.SqlClient
     {
         private SqlDataReader _reader;  // The SqlDataReader that we are reading data from
         private readonly int _columnIndex;       // The index of out column in the table
+        private long _position;         // Current position in the stream
         private Task _currentTask;      // Holds the current task being processed
         private int _readTimeout;       // Read timeout for this stream in ms (for Stream.ReadTimeout)
         private readonly CancellationTokenSource _disposalTokenSource;    // Used to indicate that a cancellation is requested due to disposal
@@ -54,8 +55,16 @@ namespace Microsoft.Data.SqlClient
 
         public override long Position
         {
-            get => throw ADP.NotSupported();
-            set => throw ADP.NotSupported();
+            get => _position;
+            set
+            {
+                if (_position == value)
+                {
+                    return;
+                }
+
+                throw ADP.NotSupported();
+            }
         }
 
         public override int ReadTimeout
@@ -90,7 +99,9 @@ namespace Microsoft.Data.SqlClient
 
             try
             {
-                return _reader.GetBytesInternalSequential(_columnIndex, buffer, offset, count, _readTimeout);
+                int bytesRead = _reader.GetBytesInternalSequential(_columnIndex, buffer, offset, count, _readTimeout);
+                _position += bytesRead;
+                return bytesRead;
             }
             catch (SqlException ex)
             {
@@ -154,6 +165,7 @@ namespace Microsoft.Data.SqlClient
                             }
                             else
                             {
+                                _position += bytesRead;
                                 completion.SetResult(bytesRead);
                             }
 
@@ -170,7 +182,9 @@ namespace Microsoft.Data.SqlClient
                                 // If we completed, but _reader is null (i.e. the stream is closed), then report cancellation
                                 if ((t.Status == TaskStatus.RanToCompletion) && (CanRead))
                                 {
-                                    completion.SetResult((int)t.Result);
+                                    int result = t.Result;
+                                    _position += result;
+                                    completion.SetResult(result);
                                 }
                                 else if (t.Status == TaskStatus.Faulted)
                                 {
@@ -325,14 +339,20 @@ namespace Microsoft.Data.SqlClient
                 throw ex.InnerException;
             }
 
-            return readTask.Result;
+            int bytesRead = readTask.Result;
+            _position += bytesRead;
+            return bytesRead;
         }
 #else
         public override IAsyncResult BeginRead(byte[] array, int offset, int count, AsyncCallback asyncCallback, object asyncState) =>
             TaskToApm.Begin(ReadAsync(array, offset, count, CancellationToken.None), asyncCallback, asyncState);
 
-        public override int EndRead(IAsyncResult asyncResult) =>
-            TaskToApm.End<int>(asyncResult);
+        public override int EndRead(IAsyncResult asyncResult)
+        {
+            int bytesRead = TaskToApm.End<int>(asyncResult);
+            _position += bytesRead;
+            return bytesRead;
+        }
 #endif
     }
 }
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataStreamTest/DataStreamTest.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataStreamTest/DataStreamTest.cs
@@ -1515,9 +1515,11 @@ CREATE TABLE {tableName} (id INT, foo VARBINARY(MAX))
                             // Basic case
                             using (stream = reader.GetStream(0))
                             {
+                                Assert.Equal(0, stream.Position);
                                 stream.Read(smallBuffer, 0, smallBuffer.Length);
                                 stream.Read(buffer, 2, 2);
 
+                                Assert.Equal(4, stream.Position);
                                 // Testing stream properties
                                 stream.Flush();
                                 DataTestUtility.AssertThrowsWrapper<NotSupportedException>(() => stream.SetLength(1));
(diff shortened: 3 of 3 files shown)


Wraith2:
Remember to add tests for sequential and non-sequential, or perhaps don't allow seeking at all in sequential mode?

SimonCropp:
@Wraith2 i think my changes to DataStreamTest.cs cover both scenarios?

SimonCropp:
@Wraith2 side query: would it make sense for SqlSequentialStream  to partially support seeking? ie if the seek action results in moving the stream forward, that should not throw an exception? if so i can submit a diff PR

Wraith2:
I'm not sure. If you've got the entire stream contents then you can seek forwards and back as you choose, though performance may be poor if you have a large number of packets. 

If you've not got all the packets then you're limited in what you can do, should you allow a seek to a position you don't yet have data for? i'd say that depends on whether you want the seek to block waiting for IO or not. It will also depend on whether you know how much data there is going to be and you might not know that because a text field can be plp, partially length prefixed, meaning you won't know how much data there is until you've got it all.

There are lots of considerations. I don't think seeking is going to work in all situations and I suspect that's why it isn't currently implemented. If you've got a limited set of situations where it can be made to work I don't have a problem with it.

SimonCropp:
@Wraith2 also. what is SqlSequentialStreamSmi.cs for? i modified it, but not sure if i should have. and also where are the Smi specific tests?

Wraith2:
I think the Smi prefixed entities are to do with a richer metadata model, so I suspect SMI stands for Sql Metadata Interface. It either wasn't finished or isn't meant to replace the ado metdata definitions. I flow changes into it where they're needed but otherwise mostly ignore it.

roji:
> would it make sense for SqlSequentialStream to partially support seeking?

My 2c on this: I agree with @Wraith2. NetworkStream does not support seeking, since that really only makes sense for file-based streams and similar. Since an ADO.NET sequential stream is effectively a network stream (with some buffering/framing), I don't think it makes sense to support seeking there either.

If the goal is to allow users to skip and jump forward (the only possible type of seeking that's possible), then they can already do that simply by (repeatedly) reading into some temporary buffer.. Any seeking implementation on on SqlSequentialStream would effectively do the same anyway, so it would just be some sugar that's inconsistent (because seeking back wouldn't be allowed).

However, when not using CommandBehavior.SequentialAccess, I do think it could make sense to support streaming, since the entire data is buffered in memory anyway. But that's quite orthogonal to this.

JRahnama:
@SimonCropp what are the benefits and use case for this PR? What does this address or add? Can you add some context to the description please?

SimonCropp:
@JRahnama done

Wraith2:
Most consumers of stream will assume that Position is valid to read even if CanSeek is false even though this is contrary to the documentation. It makes some sense because if you're returning bytes why can't you say how many  you've returned?

In this case keeping track of the number of bytes returned isn't harmful to perf or difficult to do and even though  the caller is not honouring the contract there is some value to being a good ecosystem citizen and allowing them to do this. I do not think we should support seeking or setting of position at all.

roji:
> Most consumers of stream will assume that Position is valid to read even if CanSeek is false even though this is contrary to the documentation.

That's not the case with NetworkStream, which is a much, much more widespread implementation of Stream. So I don't think that assumption is there.

> It makes some sense because if you're returning bytes why can't you say how many you've returned?

Keep in mind that these APIs quite closely low-level filesystem native APIs. At that level, the corresponding API (e.g. [ftell](https://linux.die.net/man/3/ftell) in POSIX) AFAIK does not return a value for file descriptors representing a socket.

In other words, when FileStream reports its position, it simply does a system call and forwards the information received from the kernel. In our case, though, the tracking must happen within the .NET Stream implementation as an extra thing; that really isn't its job.

> In this case keeping track of the number of bytes returned isn't harmful to perf or difficult to do and even though the caller is not honouring the contract there is some value to being a good ecosystem citizen and allowing them to do this.

Once again, I don't see why it's being a good ecosystem citizen to implement Position, when e.g. NetworkStream does not. I'd be curious to hear the case where this is actually necessary (i.e. in Aspose): either the specific component/use-case really isn't meant to work with streaming streams (NetworkStream, SqlSequentialStream), or Aspose itself is at fault and should be fixed.

Wraith2:
A network stream is a stream over a socket and I wouldn't expect to use it with stream composition. A sequential sql stream over a single field is expected to be contiguous data and makes sense to use in stream compsition, for example wrapping in a json reader, decryptor or decompressor.

roji:
> A network stream is a stream over a socket and I wouldn't expect to use it with stream composition. A sequential sql stream over a single field is expected to be contiguous data and makes sense to use in stream composition, for example wrapping in a json reader, decryptor or decompressor.

I really don't see why NetworkStream is any less composable (or meant to be composed) than SequentialSqlStream; it really is very common to wrap SslStream, BufferedStream, DeflateStream or any other types of composable streams over NetworkStream. In fact, that's probably the way it's used in the majority of cases. One of the main points of the Stream abstraction is to be composable.

In fact, SequentialSqlStream is **itself** an example of stream composition - and over NetworkStream; it may not be implemented that way specifically in SqlClient (I have no idea), but that's not really important; all SequentialSqlStream logically does is stream binary data from the network, taking care of TDS framing, buffering, and other details. In fact, the Stream implementation returned by Npgsql's DbDataReader.GetStream (in sequential mode) really is a thin wrapper over NetworkStream.

I'm still curious to hear why Aspose (or anything else) wants SequentialSqlStream to implement Position.

Wraith2:
I expect that Aspose doesn't agree with your ideas about how streams should or should not be composed. Rather than trying to argue who is wrong I'm in favour of being compatible if it's safe and low cost, which I think this PR is.

roji:
> I expect that Aspose doesn't agree with your ideas about how streams should or should not be composed. Rather than trying to argue who is wrong I'm in favour of being compatible if it's safe and low cost, which I think this PR is.

This isn't about arguing about who's wrong and who's right; NetworkStream already **doesn't** implement Position, and that's the way non-seekable Streams work in .NET - of which SequentialSqlStream. Rather than just making a change that adds non-standard behavior (and adds complexity - regardless how safe you thing it is), it doesn't seem unreasonable to ask why this particular component has this requirement, and how/why it deals with NetworkStream which doesn't meet it.

SimonCropp:
@roji as i mentioned i have hit this multiple times over the years across multiple libs from different vendors. i am sure if we looked through the IL in nuget pakages there would be many cases where libs assume that a stream supports read of position. it is fine to say "well those usages are wrong since canseek is false". but i would argue read on position is not "seeking" and there should be a very good reason to not support it. in the case of sqlsequentialstream, i see no good reason, as is indicated by the relative simplicity of this pr

roji:
* My question wasn't about looking through IL, but rather which component (or "usage story") of Aspose requires this, to try to understand the requirement better. For example, maybe the component in question really was made for working with files, and using it with network data is inappropriate in some way. It's about understanding whether the Aspose Position requirement is some arbitrary bug/limitation which can just be fixed on their end, or whether there's something more interesting behind it.
* Also, understanding a scenario where Position - but not seeking - is required might give us better general understanding of a Stream usage that we're not currently taking into account (which could lead to other changes). In this context I'll repeat again that NetworkStream doesn't support Position - are Aspose users complaining about that? If not, why?
* Importantly, has Aspose support been contacted on this limitation? What was their answer on this?
* SqlClient is a foundational library in the .NET ecosystem, so it's a good idea to be very careful with what new features go in, and about consistency with the ecosystem (e.g. whether non-seeking streams support Position). As simple as this change looks, it does add a bit of complexity, and some bit of perf overhead as well. Yes, that overhead is unlikely to matter for most usages, but adding it for **all** users of SqlClient to address a limitation we currently don't even fully understand in some specific library is not a great way to manage an important foundational library. For example, I can say with some confidence that if this same feature were proposed for NetworkStream, it would very likely be rejected; we should be clear on why it does make sense here and not there.
* There doesn't seem to be any particular reason for this change to go into SqlClient; in other words, it's trivial to introduce a wrapping stream in the middle, keeping this logic out of SqlClient. This would be "pay-per-play" (only Aspose users would need/use it), and decoupling this from SqlClient would allow the "Aspose workaround" Stream to be used with other streams which don't support Position... like NetworkStream (e.g. you could provide a non-standard API to reset the internal count at arbitrary points). Aspose could even add this to their docs if they wanted.
* I know this change seems small and harmless. However, if I had a dollar every time a user asked for some feature in order to work around some problem/bug in their application (or some 3rd-party lib), I'd be a rich man with some very convoluted and inconsistent OSS projects.

SimonCropp:
i feel no matter how much justification i put into this the response will be

> SqlClient is a foundational library in the .NET ecosystem, so it's a good idea to be very careful with what new features go in

The status is: this is causing consumers of SqlClient pain for no good reason.

This change is low risk, means sqlsequentialstream more completely implements the base API, and will reduce future consumers of Stream creating bugs from making the assumption about a read of position being a valid thing to do.

roji:
@SimonCropp first, note that I don't actually make any decisions here, the SqlClient team does. I'm just commenting as an external person who's interested in ADO.NET across drivers etc. One of my problems here is that if we go down this route, we also have to apply the same thing to Npgsql's sequential stream implementation, to the MySQL driver's, to Oracle's... This workaround for Aspose's bug/problem really does not belong in the database driver.

Now, it would still be good to get some answers to the questions I posted above - has Aspose support been contacted and what have they said, why the component in question actually needs this, what are the exact drawbacks of having a wrapping Stream doing this - which would have the advantage of being applicable to all database drivers (not just SqlClient), and even to non-database drivers (e.g. NetworkStream).

On that note, here's a suggestion. Since the Aspose problem would be just as relevant for NetworkStream, try opening an issue (or PR) on https://github.com/dotnet/runtime asking for this same workaround to be built into NetworkStream. I think you'll get objections similar to what I wrote here, but I may be wrong. If the proposal is accepted, it would definitely make sense here as well (I certainly wouldn't argue against it any more); otherwise, we'll have arguments to the contrary which should apply just as well to SqlClient.

What do you think?

roji:
One more note on this (thanks @vonzshik for pointing it out): [the .NET docs on Stream.Position](https://learn.microsoft.com/en-us/dotnet/api/system.io.stream.position?view=net-7.0#remarks) has this to say:

> The stream must support seeking to get or set the position. Use the [CanSeek](https://learn.microsoft.com/en-us/dotnet/api/system.io.stream.canseek?view=net-7.0) property to determine whether the stream supports seeking.

So adding support for a Position getter without supporting seek would be in violation of the Stream contract.

SimonCropp:
@roji  the docs for Stream.Position are flawed. they speak only to setting position. clearly a read of position should not be considered a seek

roji:
> @roji the docs for Stream.Position are flawed. they speak only to setting position. clearly a read of position should not be considered a seek

The docs are very clear, saying that the stream "must support seeking to get or set the position". This doesn't imply that reading is considered a seek, only that a stream must support seeking if getting the position is supported.

Of course, that in itself doesn't mean that the docs can't be changed; this is why I think this conversation belongs in the runtime repo rather than here. This is a wider discussion, and simply modifying the SqlClient stream implementation contrary to the docs (just to satisfy an Aspose limitation) doesn't seem right.

SimonCropp:
for context here is the Aspose bug https://forum.aspose.com/t/support-non-seekable-streams-in-aspose-pdf/209599/5

cheenamalhotra:
@SimonCropp Please initiate a new issue instead to discuss on the requirements and limitations for this API support request.
I'd like to close discussions on the PR until all requirements are agreed upon and API approved.

Closing the PR for now.

SimonCropp:
@cheenamalhotra i have put enough effort into this. if anyone else wants to continue pushing for this to be fixed, they can raise a new issue

PR #2147, opened 2023-09-05
SDL | Changing ReadXml to a more secure overload.

JRahnama:
Code Analysis detected CA3075, Unsafe overload of 'ReadXml' method, for [LoadDataSetFromXml](https://github.com/dotnet/SqlClient/blob/f384d4a25d5c8bb23e8033cc9eec365e5b385cb5/src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbMetaDataFactory.cs#L498). Changing the XmlRead to a more secure overload will solve that issue, by setting XmlResolver to null.

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbMetaDataFactory.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbMetaDataFactory.cs
@@ -506,7 +506,7 @@ namespace Microsoft.Data.ProviderBase
             {
                 XmlResolver = null
             };
-            XmlReader reader = XmlReader.Create(XmlStream, settings);
+            using XmlReader reader = XmlReader.Create(XmlStream, settings);
             _metaDataCollectionsDataSet.ReadXml(reader);
         }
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbMetaDataFactory.cs
```diff
@@ -499,9 +500,14 @@ private void LoadDataSetFromXml(Stream XmlStream)
...
             {
-                Locale = System.Globalization.CultureInfo.InvariantCulture
+                Locale = CultureInfo.InvariantCulture
+            };
+            XmlReaderSettings settings = new()
+            {
+                XmlResolver = null
             };
-            _metaDataCollectionsDataSet.ReadXml(XmlStream);
+            XmlReader reader = XmlReader.Create(XmlStream, settings);
```
Wraith2, line 509: Does the reader live after the `ReadXml(reader)` call? If it doesn't then this could be a `using` block.

DavoudEshtehari (approved):

David-Engel (approved):

PR #2157, opened 2023-09-15
Merge TDSParserStateObject.StateSnapshot stateObj fields to a new class in shared

Wraith2:
There are a number of fields which are used to store data about the state object that are captured and then restored as the snapshot is replayed. Moving these fields to a new object has two benefits. 1) the maintenance of the packet list and the state is separated logically making both easier to understand and 2) if we introduce a continue option to improve performance we will need to be able to store stateobject state for both the replay start and the replay continue points.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObject.netcore.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObject.netcore.cs
@@ -3063,7 +3063,6 @@ namespace Microsoft.Data.SqlClient
 
             internal byte[] _plpBuffer;
 
-
             private int _snapshotInBuffCount;
 
 #if DEBUG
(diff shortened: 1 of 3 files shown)


Wraith2:
The pr that builds on this one, combine-32, has a ~10% async string perf increase so it'd be nice to get this one in sooner rather than later.

David-Engel (approved):
LGTM

David-Engel:
Just a couple comments. Since this is touching core code, I did look and see the few uncovered lines. Now I know what null bitmaps are. Interesting to see it appears that existing tests aren't covering those lines. I also learned earlier this year that ALTMETADATA is related to functionality that was removed from SQL Server long ago. It's probably candidate code for pruning since we would never be able to hit that code against current server versions.

Wraith2:
I've seen null bitmaps around the docs but never knew what alt metadata was about either. I'll put it on my list of things to prune.

Wraith2:
@JRahnama @DavoudEshtehari @arellegue any chance of a quick review?

arellegue (approved):

Wraith2:
It was merged,. here https://github.com/dotnet/SqlClient/pull/2164

ErikEJ:
@Wraith2 Thanks, managed to uncover it!

PR #2033, opened 2023-05-08
Test | Removing internal build of XunitExtensions and using the nuget

JRahnama:
This PR drops the internal build of Microsoft.DotNet.XUnitExtensions and uses the nuget from dotnet/arcade.

Also there is a version update on xunit from 2.4.2 to 2.5.0

Also Microsoft.Net.Test.SDK version is upgraded to 17.7.2

PR #2159, opened 2023-09-16
Add Workload Identity Support

### LINKED ISSUE #2158: Workload Identity Support
With the [deprecation of AAD Pod Identity](https://cloudblogs.microsoft.com/opensource/2022/01/18/announcing-azure-active-directory-azure-ad-workload-identity-for-kubernetes/), app developers working in Kubernetes have been migrating to [Azure Workload Identity](https://github.com/Azure/azure-workload-identity). The `Azure.Identity` library now supports a new `WorkloadIdentityCredential` as of version `1.9.0`, and it would great to take advantage of it in the SQL client library.

While `Authentication = Active Directory Default` is already supported, which would include `WorkloadIdentityCredential` in versions >= `1.9.0`, it would require a small change to help propagate a user-specified ClientId using the existing `User Id` property in connection strings, like [it is done with managed identities today](https://github.com/dotnet/SqlClient/blob/c1984b822f8d31794cd8c85245cc6d7bd1059526/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ActiveDirectoryAuthenticationProvider.cs#L163). Furthermore, it would be advantageous for client code to directly specify the use of Workload Identity, instead of having to wait for the other token mechanisms in the default credential to fail. Today the SQL client supports directly specifying managed identity using `Authentication = Active Directory Managed Identity`, and similarly other Azure libraries are special-casing workload identity as well. I made a [similar change for the `Microsoft.Extensions.Azure` package](https://github.com/Azure/azure-sdk-for-net/pull/37944) to support workload identity specifically.

I would like two things:
1. The specification of `ClientId` via `User Id` in the connection string for workload identity when using `Authentication = Active Directory Default`
2. A new `SqlAuthenticationMethod` value called `ActiveDirectoryWorkloadIdentity` that leverages the `WorkloadIdentityCredential`
    - While it could leverage new parameters in the connection string, I think we could also rely on environment variables to configure it by default, including `AZURE_TENANT_ID`, `AZURE_CLIENT_ID`, and `AZURE_FEDERATED_TOKEN_FILE`. These are the environment variables used today by the `Azure.Identity` library.

In my app code, I could rely on updating `Azure.Identity` myself to a version >= `1.9.0` and use the default credential + environment variables to leverage workload identity.

### Additional context
Going forward, workload identity will be the Microsoft-recommended way to authenticate in AKS.

wsugarman:
Resolves #2158 

In this pull request, I make two changes:
1. Use the `User Id` as the `Client Id` for workload identity if specified when using `Authentication = Active Directory Default`
2. Add a new value `ActiveDirectoryWorkloadIdentity` for the enum `SqlAuthenticationMethod` that operates very similarly to `ActiveDirectoryManagedIdentity` except that it uses `WorkloadIdentityCredential` instead.
    - Like with managed identities, it will use the value of the `User Id` parameter in the connection string for its `Client Id` if specified. But unlike managed identity, `WorkloadIdentityCredentialOptions` defaults its value from environment variables: `AZURE_TENANT_ID`, `AZURE_CLIENT_ID`, and `AZURE_FEDERATED_TOKEN_FILE`. As of writing this description, only the `Client Id` may be overridden by the connection string.

To use `WorkloadIdentityCredential`, I needed to update `Azure.Identity` to `1.9.0` and `Azure.Core` to `1.32.0`.

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ActiveDirectoryAuthenticationProvider.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ActiveDirectoryAuthenticationProvider.cs
@@ -84,7 +84,8 @@ namespace Microsoft.Data.SqlClient
                 || authentication == SqlAuthenticationMethod.ActiveDirectoryDeviceCodeFlow
                 || authentication == SqlAuthenticationMethod.ActiveDirectoryManagedIdentity
                 || authentication == SqlAuthenticationMethod.ActiveDirectoryMSI
-                || authentication == SqlAuthenticationMethod.ActiveDirectoryDefault;
+                || authentication == SqlAuthenticationMethod.ActiveDirectoryDefault
+                || authentication == SqlAuthenticationMethod.ActiveDirectoryWorkloadIdentity;
         }
 
         /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/ActiveDirectoryAuthenticationProvider.xml' path='docs/members[@name="ActiveDirectoryAuthenticationProvider"]/BeforeLoad/*'/>
@@ -162,6 +163,7 @@ namespace Microsoft.Data.SqlClient
                 {
                     defaultAzureCredentialOptions.ManagedIdentityClientId = clientId;
                     defaultAzureCredentialOptions.SharedTokenCacheUsername = clientId;
+                    defaultAzureCredentialOptions.WorkloadIdentityClientId = clientId;
                 }
                 AccessToken accessToken = await new DefaultAzureCredential(defaultAzureCredentialOptions).GetTokenAsync(tokenRequestContext, cts.Token).ConfigureAwait(false);
                 SqlClientEventSource.Log.TryTraceEvent("AcquireTokenAsync | Acquired access token for Default auth mode. Expiry Time: {0}", accessToken.ExpiresOn);
@@ -177,11 +179,23 @@ namespace Microsoft.Data.SqlClient
                 return new SqlAuthenticationToken(accessToken.Token, accessToken.ExpiresOn);
             }
 
-            AuthenticationResult result = null;
             if (parameters.AuthenticationMethod == SqlAuthenticationMethod.ActiveDirectoryServicePrincipal)
             {
                 AccessToken accessToken = await new ClientSecretCredential(audience, parameters.UserId, parameters.Password, tokenCredentialOptions).GetTokenAsync(tokenRequestContext, cts.Token).ConfigureAwait(false);
                 SqlClientEventSource.Log.TryTraceEvent("AcquireTokenAsync | Acquired access token for Active Directory Service Principal auth mode. Expiry Time: {0}", accessToken.ExpiresOn);
+                return new SqlAuthenticationToken(accessToken.Token, accessToken.ExpiresOn);
+            }
+
+            if (parameters.AuthenticationMethod == SqlAuthenticationMethod.ActiveDirectoryManagedIdentity || parameters.AuthenticationMethod == SqlAuthenticationMethod.ActiveDirectoryMSI)
+            {
+                WorkloadIdentityCredentialOptions options = new()
+                {
+                    AuthorityHost = new Uri(authority),
+                    ClientId = clientId,
+                };
+
+                AccessToken accessToken = await new WorkloadIdentityCredential(options).GetTokenAsync(tokenRequestContext, cts.Token).ConfigureAwait(false);
+                SqlClientEventSource.Log.TryTraceEvent("AcquireTokenAsync | Acquired access token for Workload Identity auth mode. Expiry Time: {0}", accessToken.ExpiresOn);
                 return new SqlAuthenticationToken(accessToken.Token, accessToken.ExpiresOn);
             }
 
@@ -209,6 +223,7 @@ namespace Microsoft.Data.SqlClient
 #endif
                 );
 
+            AuthenticationResult result = null;
             IPublicClientApplication app = GetPublicClientAppInstance(pcaKey);
 
             if (parameters.AuthenticationMethod == SqlAuthenticationMethod.ActiveDirectoryIntegrated)
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsEnums.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsEnums.cs
@@ -273,6 +273,7 @@ namespace Microsoft.Data.SqlClient
         public const byte MSALWORKFLOW_ACTIVEDIRECTORYMANAGEDIDENTITY = 0x03; // Using the Interactive byte as that's supported for Identity based authentication
         public const byte MSALWORKFLOW_ACTIVEDIRECTORYDEFAULT = 0x03; // Using the Interactive byte as that is the closest we have to non-password based authentication modes
         public const byte MSALWORKFLOW_ACTIVEDIRECTORYTOKENCREDENTIAL = 0x03; // Using the Interactive byte as that is the closest we have to non-password based authentication modes
+        public const byte MSALWORKFLOW_ACTIVEDIRECTORYWORKLOADIDENTITY = 0x03; // Using the Interactive byte as that's supported for Identity based authentication
 
         public enum ActiveDirectoryWorkflow : byte
         {
@@ -283,6 +284,7 @@ namespace Microsoft.Data.SqlClient
             DeviceCodeFlow = MSALWORKFLOW_ACTIVEDIRECTORYDEVICECODEFLOW,
             ManagedIdentity = MSALWORKFLOW_ACTIVEDIRECTORYMANAGEDIDENTITY,
             Default = MSALWORKFLOW_ACTIVEDIRECTORYDEFAULT,
+            WorkloadIdentity = MSALWORKFLOW_ACTIVEDIRECTORYWORKLOADIDENTITY,
         }
 
         // The string used for username in the error message when Authentication = Active Directory Integrated with FedAuth is used, if authentication fails.
@@ -1206,8 +1208,11 @@ namespace Microsoft.Data.SqlClient
         /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlAuthenticationMethod.xml' path='docs/members[@name="SqlAuthenticationMethod"]/ActiveDirectoryDefault/*'/>
         ActiveDirectoryDefault,
 #if ADONET_CERT_AUTH && NETFRAMEWORK
-        SqlCertificate
+        SqlCertificate,
 #endif
+
+        /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlAuthenticationMethod.xml' path='docs/members[@name="SqlAuthenticationMethod"]/ActiveDirectoryWorkloadIdentity/*'/>
+        ActiveDirectoryWorkloadIdentity,
     }
     // This enum indicates the state of TransparentNetworkIPResolution
     // The first attempt when TNIR is on should be sequential. If the first attempt failes next attempts should be parallel.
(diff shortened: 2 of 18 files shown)


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsEnums.cs
```diff
@@ -1206,8 +1208,11 @@ public enum SqlAuthenticationMethod
         /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlAuthenticationMethod.xml' path='docs/members[@name="SqlAuthenticationMethod"]/ActiveDirectoryDefault/*'/>
         ActiveDirectoryDefault,
 #if ADONET_CERT_AUTH && NETFRAMEWORK
-        SqlCertificate
+        SqlCertificate,
 #endif
+
+        /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlAuthenticationMethod.xml' path='docs/members[@name="SqlAuthenticationMethod"]/ActiveDirectoryWorkloadIdentity/*'/>
```
wsugarman, line 1214: This comment is a bit hard to write because it may be the 10th or 11th value. I could make this doc inclusion conditional?

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectivityTests/AADConnectionTest.cs
```diff
@@ -453,6 +453,27 @@ public static void ActiveDirectoryManagedIdentityWithCredentialsMustFail()
             Assert.Contains(expectedMessage, e.Message);
         }
 
+        [ConditionalFact(nameof(IsAADConnStringsSetup))]
+        public static void ActiveDirectoryWorkloadIdentityWithCredentialsMustFail()
```
wsugarman, line 457: I still need to write test cases for auth when the test environment supports workload identity, but I need to do more research in how to write that.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ActiveDirectoryAuthenticationProvider.cs
```diff
@@ -177,14 +179,26 @@ public override async Task<SqlAuthenticationToken> AcquireTokenAsync(SqlAuthenti
...
             if (parameters.AuthenticationMethod == SqlAuthenticationMethod.ActiveDirectoryServicePrincipal)
             {
                 AccessToken accessToken = await new ClientSecretCredential(audience, parameters.UserId, parameters.Password, tokenCredentialOptions).GetTokenAsync(tokenRequestContext, cts.Token).ConfigureAwait(false);
                 SqlClientEventSource.Log.TryTraceEvent("AcquireTokenAsync | Acquired access token for Active Directory Service Principal auth mode. Expiry Time: {0}", accessToken.ExpiresOn);
                 return new SqlAuthenticationToken(accessToken.Token, accessToken.ExpiresOn);
             }
 
+            if (parameters.AuthenticationMethod == SqlAuthenticationMethod.ActiveDirectoryManagedIdentity || parameters.AuthenticationMethod == SqlAuthenticationMethod.ActiveDirectoryMSI)
+            {
+                WorkloadIdentityCredentialOptions options = new()
```
wsugarman, line 191: All of the properties are initialized using environment variables: https://github.com/Azure/azure-sdk-for-net/blob/99cc57c41c7fa2a3df564ad0c6a745c001cd4517/sdk/identity/Azure.Identity/src/Credentials/WorkloadIdentityCredentialOptions.cs#L13

JRahnama:
@wsugarman,I have not investigated the details yet, but couldn’t’ve achieved this by setting a custom provider? I will dig deeper into this next week, but do we have to add every possible scenario to the driver?

wsugarman:
> @wsugarman,I have not investigated the details yet, but couldn’t’ve achieved this by setting a custom provider? I will dig deeper into this next week, but do we have to add every possible scenario to the driver?

I had no idea this API existed 😅 This would be a great workaround!
I do think the library will benefit from including it in the driver as it replaces more and more managed identity scenarios.

David-Engel:
In addition to creating a custom provider that overrides existing authentication options, there is a new, simpler AccessTokenCallback property added in 5.2 preview 3 that makes implementing new token authentication methods much easier. (Not quite as easy as a connection string adjustment, but a-few-lines-of-code easy.)

```
using Azure.Core;
using Azure.Identity;
...

            string s_defaultScopeSuffix = "/.default";
            DefaultAzureCredentialOptions options = new()
            {
                //set your options here
                ExcludeAzureCliCredential = true,
                ExcludeVisualStudioCredential = true
            };
            var cred = new DefaultAzureCredential(options);
            Func<SqlAuthenticationParameters, CancellationToken, Task<SqlAuthenticationToken>> myTokenCallback = async (ctx, cancellationToken) =>
            {
                string scope = ctx.Resource.EndsWith(s_defaultScopeSuffix) ? ctx.Resource : ctx.Resource + s_defaultScopeSuffix;
                AccessToken token = await cred.GetTokenAsync(new TokenRequestContext(new[] { scope }), cancellationToken);
                return new SqlAuthenticationToken(token.Token, token.ExpiresOn);
            };

            using (SqlConnection sqlConnection = new SqlConnection("Server=name.database.windows.net;Database=db;"))
            {
                sqlConnection.AccessTokenCallback = myTokenCallback;
                sqlConnection.Open();
                Console.WriteLine("Connected successfully!");
            }
```

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsEnums.cs
cheenamalhotra: You can include it above SqlCertificate, does that work?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ActiveDirectoryAuthenticationProvider.cs
cheenamalhotra: If this is documented well, it should work I suppose. Tagging @David-Engel to review this feature to fit well in the AAD stack potentially in sync with other drivers as well.

wsugarman:
> In addition to creating a custom provider that overrides existing authentication options, there is a new, simpler AccessTokenCallback property added in 5.2 preview 3 that makes implementing new token authentication methods much easier. (Not quite as easy as a connection string adjustment, but a-few-lines-of-code easy.)
> 
> ```
> using Azure.Core;
> using Azure.Identity;
> using Microsoft.Data.SqlClient;
> ...
> 
>             string s_defaultScopeSuffix = "/.default";
>             DefaultAzureCredentialOptions options = new()
>             {
>                 //set your options here
>                 ExcludeAzureCliCredential = true,
>                 ExcludeVisualStudioCredential = true
>             };
>             var cred = new DefaultAzureCredential(options);
>             Func<SqlAuthenticationParameters, CancellationToken, Task<SqlAuthenticationToken>> myTokenCallback = async (ctx, cancellationToken) =>
>             {
>                 string scope = ctx.Resource.EndsWith(s_defaultScopeSuffix) ? ctx.Resource : ctx.Resource + s_defaultScopeSuffix;
>                 AccessToken token = await cred.GetTokenAsync(new TokenRequestContext(new[] { scope }), cancellationToken);
>                 return new SqlAuthenticationToken(token.Token, token.ExpiresOn);
>             };
> 
>             using (SqlConnection sqlConnection = new SqlConnection("Server=name.database.windows.net;Database=db;"))
>             {
>                 sqlConnection.AccessTokenCallback = myTokenCallback;
>                 sqlConnection.Open();
>                 Console.WriteLine("Connected successfully!");
>             }
> ```

Oh! That will be great! So, it will join the existing `AccessToken` property, but this will be a more generic callback complete with the auth parameters from the connection string?

That will be super helpful, but I still think a common scenario like workload identity will warrant inclusion in the driver. Otherwise, users will have to create the same workload identity `SqlAuthenticationProvider` or likely create some abstraction around `SqlConnection` that automatically assigns the `AccessTokenCallback` if there a lot of places in the codebase where a `SqlConnection` is created and opened.

David-Engel:
> Oh! That will be great! So, it will join the existing `AccessToken` property, but this will be a more generic callback complete with the auth parameters from the connection string?

Yes. The `AccessToken` property suffers from some major drawbacks that affect connection pooling:

* SqlClient doesn't know when the token will expire, resulting in errors when a connection with an expired token is obtained from the connection pool.
* Since it's part of the pool key, if you set min pool size, those min connections will never be automatically cleaned up unless ClearPool() or ClearAllPools() is called.

The callback gives SqlClient the expiration date and allows it to get a new token when a new physical connection is required, just like the built-in methods.

> That will be super helpful, but I still think a common scenario like workload identity will warrant inclusion in the driver. Otherwise, users will have to create the same workload identity `SqlAuthenticationProvider` or likely create some abstraction around `SqlConnection` that automatically assigns the `AccessTokenCallback` if there a lot of places in the codebase where a `SqlConnection` is created and opened.

Oh definitely. It's just difficult to keep up with all the new Azure AD authentication methods they keep coming up with. So a generic method at least takes the pressure off implementing every new one ASAP. You can write a wrapper around SqlConnection to authenticate however you want and still have connection pooling work correctly.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsEnums.cs
wsugarman: Yeah, I noticed that was done previously too, so no problem.

src/Microsoft.Data.SqlClient.sln
```diff
@@ -11,9 +11,6 @@ EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "TDS.Servers", "Microsoft.Data.SqlClient\tests\tools\TDS\TDS.Servers\TDS.Servers.csproj", "{978063D3-FBB5-4E10-8C45-67C90BE1B931}"
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "TDS", "Microsoft.Data.SqlClient\tests\tools\TDS\TDS\TDS.csproj", "{8DC9D1A0-351B-47BC-A90F-B9DA542550E9}"
-	ProjectSection(ProjectDependencies) = postProject
```
wsugarman, line 14: I included a fix in this PR to clean up the solution file. The project with this GUID was removed as part of this PR: https://github.com/dotnet/SqlClient/pull/2033

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsEnums.cs
```diff
@@ -1212,8 +1214,12 @@ public enum SqlAuthenticationMethod
 
         /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlAuthenticationMethod.xml' path='docs/members[@name="SqlAuthenticationMethod"]/ActiveDirectoryDefault/*'/>
         ActiveDirectoryDefault,
+
+        /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlAuthenticationMethod.xml' path='docs/members[@name="SqlAuthenticationMethod"]/ActiveDirectoryWorkloadIdentity/*'/>
+        ActiveDirectoryWorkloadIdentity,
+
 #if ADONET_CERT_AUTH && NETFRAMEWORK
-        SqlCertificate
+        SqlCertificate,
```
H-Yeo, line 1222: Extra Comma here

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectivityTests/AADConnectionTest.cs
H-Yeo: Sorry, kind of new here, but do you have an update on this? It's 2 months old.

David-Engel: Our automated test environment is unlikely to ever support workload identity unless there is a way to emulate it without setting up an expensive Kubernetes environment.

@wsugarman - Do you think that (emulation) could ever happen?

Either way, I'm okay with manual testing of this feature for now.

wsugarman: I'm unaware of any existing emulation of workload identity, but perhaps it may make its way into other AKS-based offerings that could be simpler to set-up, like Azure Container Apps? I think in the meantime, without AKS, it will need to be manually tested

JRahnama (approved):

wsugarman:
@JRahnama - Thanks for merging it!

Issue #1124, opened 2021-06-17
Deadlock between SinglePhaseEnlistment.Aborted() and DbConnectionInternal.DetachTransaction()


DavidBoike:
### Describe the bug
Starting with 2.1.3, we experienced deadlocking issues when a TransactionScope is disposed. Either the code should fully commit, or it should throw a `TransactionAbortedException` or some such, but instead it just hangs on the end of the `using` block where the `TransactionScope` is disposed.

May be an unintended consequence of https://github.com/dotnet/SqlClient/pull/1049, the deadlock does not occur in 2.1.2 but does occur in 2.1.3 and 3.0.0.

At Particular Software we have a test for our SQL Transport designed to ensure that certain things happen when a transaction is aborted that was deadlocking every single time. We were able to slim this down into a minimal repro.

We have another situation where we were spinning up two connections to the same server to test the database's support for DTC. This was also hanging but it appeared like a race condition. The repro below happens consistently.

### To reproduce

[The code in this gist](https://gist.github.com/DavidBoike/94a1d5e3fc57e735843a33054e6ba03e) is a minimal reproduction of the problem.

1. Run pointed at a local SQL Express instance, or change the connection string to match.
2. As written (using v2.1.3) it will hang and never complete.
3. Changing to v2.1.2 (and being sure to rebuild) will result in an a `TransactionAbortedException` being quickly thrown, caught, and printed to the console, after which the program will complete.

Partial code listings, or multiple fragments of code, will slow down our response or cause us to push the issue back to you to provide code to reproduce the issue.

### Expected behavior
Should throw `TransactionAbortedException` instead of hanging.

### Further technical details
* Microsoft.Data.SqlClient version: 2.1.3 or 3.0.0
* .NET target: .NET Framework 4.7.2
* SQL Server version: Microsoft SQL Server 2019 (RTM-GDR) (KB4583458) - 15.0.2080.9 (X64)   Nov  6 2020 16:50:01   Copyright (C) 2019 Microsoft Corporation  Express Edition (64-bit) on Windows 10 Home 10.0 <X64> (Build 19042: ) (Hypervisor) 
* Operating system: Windows 10 Home (10.0.19042 Build 19042)

cheenamalhotra:
Hi @DavidBoike 

It is indeed a bug for DTC in .NET Framework. 
I'm able to reproduce it as well. Thanks for bringing it to us with repro.

[...]

PR #2161, opened 2023-09-19
Ports Fix deadlock in transaction (#1242) to .NET

danielmarbach:
.NET 7 and higher supports distributed transactions on Windows. For more details see https://github.com/dotnet/runtime/issues/715.

Unfortunately, the code that is responsible for the transactions in the Netcore part of the SqlClient has not been adopted to address https://github.com/dotnet/SqlClient/issues/1124 which was fixed with https://github.com/dotnet/SqlClient/pull/1242. This PR brings over the changes from https://github.com/dotnet/SqlClient/pull/1242 and makes sure the distributed transaction on .NET running on Windows are not deadlocking.

We ran into the same issue as described in https://github.com/dotnet/SqlClient/issues/1124 when trying to enable distributed transaction support on NET8 on Windows.

### DIFF
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
@@ -442,7 +442,7 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests
         public static bool IsSupportingDistributedTransactions()
         {
 #if NET7_0_OR_GREATER
-            return OperatingSystem.IsWindows() && IsNotAzureServer();
+            return OperatingSystem.IsWindows() && System.Runtime.InteropServices.RuntimeInformation.ProcessArchitecture != System.Runtime.InteropServices.Architecture.X86 && IsNotAzureServer();
 #elif NETFRAMEWORK
             return System.Runtime.InteropServices.RuntimeInformation.IsOSPlatform(System.Runtime.InteropServices.OSPlatform.Windows) && IsNotAzureServer();
 #else
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
@@ -444,7 +444,7 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests
 #if NET7_0_OR_GREATER
             return OperatingSystem.IsWindows() && System.Runtime.InteropServices.RuntimeInformation.ProcessArchitecture != System.Runtime.InteropServices.Architecture.X86 && IsNotAzureServer();
 #elif NETFRAMEWORK
-            return System.Runtime.InteropServices.RuntimeInformation.IsOSPlatform(System.Runtime.InteropServices.OSPlatform.Windows) && IsNotAzureServer();
+            return IsNotAzureServer();
 #else
             return false;
 #endif
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
@@ -439,6 +439,17 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests
                 ;
         }
 
+        public static bool IsSupportingDistributedTransactions()
+        {
+#if NET7_0_OR_GREATER
+            return OperatingSystem.IsWindows() && IsNotAzureServer();
+#elif NETFRAMEWORK
+            return System.Runtime.InteropServices.RuntimeInformation.IsOSPlatform(System.Runtime.InteropServices.OSPlatform.Windows) && IsNotAzureServer();
+#else
+            return false;
+#endif
+        }
+
         public static bool IsUsingManagedSNI() => UseManagedSNIOnWindows;
 
         public static bool IsNotUsingManagedSNIOnWindows() => !UseManagedSNIOnWindows;
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/TransactionTest/TransactionEnlistmentTest.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/TransactionTest/TransactionEnlistmentTest.cs
@@ -47,7 +47,7 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests
             RunTestSet(TestCase_ManualEnlistment_Enlist_TxScopeComplete);
         }
 
-        [ConditionalFact(typeof(DataTestUtility), nameof(DataTestUtility.AreConnStringsSetup), nameof(DataTestUtility.IsNotAzureServer))]
+        [ConditionalFact(typeof(DataTestUtility), nameof(DataTestUtility.AreConnStringsSetup), nameof(DataTestUtility.IsSupportingDistributedTransactions))]
         public static void TestEnlistmentPrepare_TxScopeComplete()
         {
             Assert.Throws<TransactionAbortedException>( () =>
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/TransactionTest/TransactionEnlistmentTest.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/TransactionTest/TransactionEnlistmentTest.cs
@@ -47,7 +47,6 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests
             RunTestSet(TestCase_ManualEnlistment_Enlist_TxScopeComplete);
         }
 
-        [SkipOnTargetFramework(TargetFrameworkMonikers.Netcoreapp)]
         [ConditionalFact(typeof(DataTestUtility), nameof(DataTestUtility.AreConnStringsSetup), nameof(DataTestUtility.IsNotAzureServer))]
         public static void TestEnlistmentPrepare_TxScopeComplete()
         {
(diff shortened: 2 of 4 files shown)


danielmarbach:
The test seems to fail only on a few targets. I enabled an existing test from the original PR to now run on NET too.

danielmarbach:
@cheenamalhotra any ideas/pointers?

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/TransactionTest/TransactionEnlistmentTest.cs
```diff
@@ -47,7 +47,6 @@ public static void TestManualEnlistment_Enlist_TxScopeComplete()
             RunTestSet(TestCase_ManualEnlistment_Enlist_TxScopeComplete);
         }
 
-        [SkipOnTargetFramework(TargetFrameworkMonikers.Netcoreapp)]
         [ConditionalFact(typeof(DataTestUtility), nameof(DataTestUtility.AreConnStringsSetup), nameof(DataTestUtility.IsNotAzureServer))]
         public static void TestEnlistmentPrepare_TxScopeComplete()
```
cheenamalhotra, line 51: I would suggest looking more into whether Enlistment prepared in .NET Core?
The test expects that [Prepare](https://github.com/dotnet/SqlClient/blob/c1984b822f8d31794cd8c85245cc6d7bd1059526/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/TransactionTest/TransactionEnlistmentTest.cs#L200) is called that forces a rollback and calling `txScope.Complete();` would lead to an exception.

It seems the failing tests are due to no error being thrown.

danielmarbach: The try/catch and the assert was masking another exception. I reworked it to use Assert.Throws. I forgot that DTC support works on .NET 7 and higher only on windows for now. See https://github.com/dotnet/runtime/issues/715

The remaining work is tracked in https://github.com/dotnet/runtime/issues/71769

I will try to find ways to properly guard the tests there with a conditional fact

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
```diff
@@ -439,6 +439,17 @@ public static bool IsTargetReadyForAeWithKeyStore()
                 ;
         }
 
+        public static bool IsSupportingDistributedTransactions()
+        {
+#if NET7_0_OR_GREATER
+            return OperatingSystem.IsWindows() && IsNotAzureServer();
+#elif NETFRAMEWORK
+            return OperatingSystem.IsWindows() && IsNotAzureServer();
```
danielmarbach, line 447:
```suggestion
            return RuntimeInformation.IsOSPlatform(OSPlatform.Windows) && IsNotAzureServer();
```

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
```diff
@@ -439,6 +439,17 @@ public static bool IsTargetReadyForAeWithKeyStore()
                 ;
         }
 
+        public static bool IsSupportingDistributedTransactions()
+        {
+#if NET7_0_OR_GREATER
+            return OperatingSystem.IsWindows() && IsNotAzureServer();
+#elif NETFRAMEWORK
+            return System.Runtime.InteropServices. RuntimeInformation.IsOSPlatform(OSPlatform.Windows) && IsNotAzureServer();
```
danielmarbach, line 447:
```suggestion
            return System.Runtime.InteropServices. RuntimeInformation.IsOSPlatform(System.Runtime.InteropServices.OSPlatform.OSPlatform.Windows) && IsNotAzureServer();
```

danielmarbach:
One last problem to address, it seems

> Distributed transactions are currently unsupported in 32-bit processes.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
```diff
@@ -439,6 +439,17 @@ public static bool IsTargetReadyForAeWithKeyStore()
                 ;
         }
 
+        public static bool IsSupportingDistributedTransactions()
+        {
+#if NET7_0_OR_GREATER
+            return OperatingSystem.IsWindows() && IsNotAzureServer();
```
danielmarbach, line 445:
```suggestion
            return OperatingSystem.IsWindows() && System.Runtime.InteropServices.RuntimeInformation.ProcessArchitecture != System.Runtime.InteropServices.Architecture.X86 && IsNotAzureServer();
```

danielmarbach:
NET 7 and higher requires non-32 bit processes, see https://github.com/dotnet/runtime/blob/main/src/libraries/System.Transactions.Local/src/System/Transactions/DtcProxyShim/DtcProxyShimFactory.cs#L69-L72. So I used the same check negated (although that is not entirely accurate it is a good enough start)

danielmarbach:
@cheenamalhotra @JRahnama Fingers crossed this is ready to go now. I would appreciate a review

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
```diff
@@ -439,6 +439,17 @@ public static bool IsTargetReadyForAeWithKeyStore()
                 ;
         }
 
+        public static bool IsSupportingDistributedTransactions()
```
danielmarbach, line 442: Decided to use fully qualified names here to avoid having to do `if/def` in the using section

danielmarbach, line 442: I tried to use the same `Is` pattern although I felt is is a bit of a clumsy name. I would have preferred `SupportsDistributedTransactions` but I wanted to make sure it fits the style of other such methods.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
```diff
@@ -439,6 +439,17 @@ public static bool IsTargetReadyForAeWithKeyStore()
...
+        {
+#if NET7_0_OR_GREATER
+            return OperatingSystem.IsWindows() && System.Runtime.InteropServices.RuntimeInformation.ProcessArchitecture != System.Runtime.InteropServices.Architecture.X86 && IsNotAzureServer();
+#elif NETFRAMEWORK
+            return System.Runtime.InteropServices.RuntimeInformation.IsOSPlatform(System.Runtime.InteropServices.OSPlatform.Windows) && IsNotAzureServer();
```
cheenamalhotra, line 447: Nit: .NET Framework is not built/tested on non-windows, so platform check can be avoided.

cheenamalhotra (approved):
LGTM

danielmarbach:
@cheenamalhotra It seems the enclaves tests are running into timeouts but when I look at the stack trace I do not see any of the code I touched being in the stack trace. Is this a red herring?

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
danielmarbach: Isn't it safer to leave it as is? Happy to change it if you wish so

danielmarbach:
Is this something that should also be fixed in the SqlClient in the runtime repo?

danielmarbach:
@cheenamalhotra Do you have any insights on https://github.com/dotnet/SqlClient/pull/2161#issuecomment-1729084037 and https://github.com/dotnet/SqlClient/pull/2161#issuecomment-1729096189? Is there anything else you need from my end to push this forward?

Chucked:
@danielmarbach Any updates?

danielmarbach:
@Chucked why are you pinging me? I'm a community contributor who has gone through the process of fixing this bug but I'm at the mercy of people at Microsoft to merge this which hasn't happened yet nor I have a gotten inputs on the failing tests

JRahnama:
Please bear with us as we’re working on some internal updates. We apologize for the delay and will get back to you as soon as possible.

danielmarbach:
@JRahnama thanks for the reply.

@Chucked no worries. Me too 😊

David-Engel (approved):

bording:
@David-Engel I see you've added this to the 5.2.0-preview4 milestone. Any sort of time frame for when 5.2.0 is going to be released?

Given that this is a bug fix, is there any chance of it being backported and released sooner as a 5.1.2 release perhaps?

David-Engel:
We just set a new target of Dec 7 for 5.2 preview 4. We have some internal fires that have pushed things back and there are still a number of PRs we want to get into a preview before GA. That means 5.2 GA won't be until Q1 2024, at the soonest.

I'd prefer to see this fix get some time under a preview release before going into a hotfix. So it's not going into 5.1.2. But it will likely go into a future hotfix.

bording:
> I'd prefer to see this fix get some time under a preview release before going into a hotfix. So it's not going into 5.1.2. But it will likely go into a future hotfix.

If this was a new fix, I could understand this reasoning. However, this fix is already live in 5.1.1 for the Framework side of things, but it was missed for the Core side.

This bug is actively blocking us and prevents DTC use on .NET, which has been supported on Windows since the release of .NET 7.

Given all of this, any chance of you reconsidering shipping this fix sooner?

SimonCropp:
> Given that this is a bug fix, is there any chance of it being backported and released sooner as a 5.1.2 release perhaps?

big +1 for this. would really like to see this in a patch ASAP

David-Engel:
We have this currently targeting a 5.1 hotfix in early January.

PR #2165, opened 2023-09-21
Deprioritize ALTMETADATA and ALTROW tokens

Wraith2:
The MS-TDS spec identifies that two sql tokens were deprecated in TDS 7.4 which is from around 2010. We still support these tokens in the code but they're checked in a few switch statements before tokens which are still in use and this is a waste of time. This PR moves the two deprecated tokens to the bottom of the list so they'll still work if we somehow encounter them but they'll be the last things checked for. It also puts a note on the definition and uses of the token to indicate that it is deprecated.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
@@ -3147,21 +3147,6 @@ namespace Microsoft.Data.SqlClient
 
                     switch (token)
                     {
-                        case TdsEnums.SQLALTROW:
-                            if (_altRowStatus == ALTROWSTATUS.Null)
-                            {
-                                // cache the regular metadata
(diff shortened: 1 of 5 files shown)


roji:
@Wraith2 AFAIK switch statements over enums are frequently optimized by the compiler (they're not just a series of if/else), so that the placement of the arms doesn't matter (whether this actually happens depends on the actual enum in various ways).

In other words, if this is for documenting/reorganizing the code that's one thing, but if you believe there's an actual perf difference, as always it would be nice to see a benchmark.

Wraith2:
If the only effect is to improve the code then I'm fine with that.

I believe that in the absence of dPGO and PGO that because the tokens are discontiguous and the method is large that the current jit optimizations that can cause reordering of switch conditions will not apply and as a result this will remove a couple of comparisons. To prove that I would need a full local runtime build and the ability to jit-diff the output which is frankly too much work just to disprove some nerd sniping.

David-Engel (approved):

JRahnama (approved):

PR #1844, opened 2022-11-22
Merge common code bases for TdsParserStateObject.cs (2)

panoskj:
Second part of #1520. As I explained in this [comment](https://github.com/dotnet/SqlClient/pull/1843#issuecomment-1322847306), #1843 does more or less the same, but I have already written the code for this PR. Perhaps you will find some use for it.

As for the code itself, you will notice I have left lots of #if directives, intentionally, to highlight the differences between the two versions, before deciding how to resolve them.

PS: either way, please try to merge one of these PRs as soon as possible. They will allow us to partially fix the async bug. A full fix would require ~~SqlDataReader~~ TdsParser too, if I remember correctly.

PR #285, opened 2019-10-24
perf: Add multipacket async read target buffer caching

Wraith2:
closes https://github.com/dotnet/SqlClient/issues/245

Async reads that span multiple packets can allocate large target arrays and drop them repeatedly causing a large memory and cpu problem. This PR uses the async snapshot already present for all async calls to allow storing the presized target buffer between packet receipt decode attempts avoiding re-allocations.

Before:

| Method |         Behavior |      Mean |     Error |     StdDev |       Gen 0 |       Gen 1 |       Gen 2 |   Allocated |
|------- |----------------- |----------:|----------:|-----------:|------------:|------------:|------------:|------------:|
|  **Async** |          **Default** | **448.84 ms** | **8.9453 ms** |  **8.7855 ms** | **309000.0000** | **309000.0000** | **309000.0000** |     **3.98 KB** |
|   Sync |          Default |  17.46 ms | 0.1025 ms |  0.0959 ms |    406.2500 |    406.2500 |    406.2500 | 10244.82 KB |
|  **Async** | **SequentialAccess** | **458.23 ms** | **9.0577 ms** | **10.4309 ms** | **310000.0000** | **310000.0000** | **310000.0000** |     **3.98 KB** |
|   Sync | SequentialAccess |  17.38 ms | 0.0638 ms |  0.0596 ms |    406.2500 |    406.2500 |    406.2500 | 10244.82 KB |

After:

| Method |         Behavior |      Mean |     Error |    StdDev |    Gen 0 |    Gen 1 |    Gen 2 |   Allocated |
|------- |----------------- |----------:|----------:|----------:|---------:|---------:|---------:|------------:|
|  **Async** |          **Default** | **124.68 ms** | **0.7149 ms** | **0.6687 ms** | **800.0000** | **400.0000** | **200.0000** |     **3.95 KB** |
|   Sync |          Default |  16.41 ms | 0.0898 ms | 0.0840 ms | 406.2500 | 406.2500 | 406.2500 | 10244.82 KB |
|  **Async** | **SequentialAccess** | **125.36 ms** | **1.3684 ms** | **1.2800 ms** | **750.0000** | **500.0000** | **250.0000** |     **3.95 KB** |
|   Sync | SequentialAccess |  16.48 ms | 0.0525 ms | 0.0491 ms | 406.2500 | 406.2500 | 406.2500 | 10244.82 KB |

BDN isn't showing LOH allocations which is why the memory numbers look odd. Using profiler results you can see the savings.

Before:
beforePNG

[...]

PR #2168, opened 2023-09-24
Merge common code bases for TdsParserStateObject.cs (3)

panoskj:
Basically the first two commits of #1844. Plus an optional refactoring. Let me know what you think.


panoskj:
I just rebased the branch adding a few commits, in order to follow @David-Engel's suggestion:

> On the "breaking it up" suggestion. Troubleshooting the pipeline failures might be easier by breaking it up, too. I'd first try to make changes to the separate TdsParserStateObject.cs files to bring them closer to identical. Ideas for steps in individual PRs:
> 
> 1. Rename internal variables/properties to all match.
> 
> 2. Reorder methods/member variables to align the file layout as much as possible, keeping identical code together as much as possible. I might even add some regions and/or comments to help future diffs keep their alignment when common code is removed from each file. (No other changes so that reviewers don't have to scrutinize any differences between the large code blocks.)
> 
>  3. Move common code into a new common file. Avoid any changes within moved blocks of code.

Wraith2:
That's still a lot of changes to review with the care we need. I've gone through them all and made notes so I'll put my notes here to help others, function by function.

TryPeekByte - identical in both
TryReadByteArray - functionally identical, netfx adds reliability assert, uses different memory barrier, uses null check for span empty
TryReadByte - functionally identical, netfx adds reliability assert, uses different memory barrier
TryReadChar - functionally identical, netfx uses _bTmp where netcore uses stackalloc, netcore is better
TryReadInt16 - functionally identical, netfx uses _bTmp where netcore uses stackalloc, netcore is better
TryReadInt32 - mostly identical, netfx uses _bTmp where netcore uses stackalloc, netfx uses BinaryConverter where netcore does manual shifting
TryReadInt64 - functionally identical, netfx adds reliability assert
TryReadUInt16 - functionally identical, netfx uses _bTmp where netcore uses stackalloc, netcore is better, no reliability assert?
TryReadUInt32 - functionally identical, netfx uses _bTmp where netcore uses stackalloc, netfx adds reliability assert, netcore is better
TryReadSingle - functionally identical, netfx adds reliability assert
TryReadDouble - functionally identical, netfx adds reliability assert
TryReadString - functionally identical, netfx adds reliability assert
TryReadStringWithEncoding - functionally identical, netfx adds reliability assert
ReadPlpLength - identical in both
TryReadPlpLength - identical in both
ReadPlpBytesChunk - identical in both
TryReadPlpBytes - mostly the same, netcore uses Array.Empty and has support for cached plpbuffer stored in snapshot, netcore is better

It would be good to understand what the reliability assert is supposed to be doing and see if there a way we can make it compile out completely on netcore.

Wraith2 (approved):

panoskj:
Good to see you reviewing this @Wraith2, since you wrote #667 and #285. If it helps:
1. TryReadByteArray: Interlocked.MemoryBarrier is a wrapper for Thread.MemoryBarrier according to its [documentation](https://learn.microsoft.com/en-us/dotnet/api/system.threading.interlocked.memorybarrier?view=net-7.0) (see Remarks section).
2. TryReadByteArray: The null-check that changes to Span.IsEmpty is inside a Debug.Assert condition, so it doesn't affect Release behavior. Moreover, a few lines below there is an if statement using Span.IsEmpty rather than the null-check in both netcore and netfx. I didn't go back in commit history to see how it ended up this way yet, will edit this comment if I do.
3. The reliability assert call is compiled out actually, because I annotated the empty stub with `[Conditional("NETFRAMEWORK")]`. Since the stub's code is only compiled for netcore, the condition will always be false. The reliability assert in netfx is basically a Debug.Assert, checking the that method is running in the correct context (e.g. some reliability functions have been called beforehand).

Wraith2:
My personal preference would be to use Interlocked.MemoryBarrier everywhere if possible. It fits more closely with the logical intention than using the Thread version to me. Either version would work identically so it's not a huge issue.

For spans since they're structs comparing against null has never made sense to me and I'd prefer to have IsEmpty used everywhere if possible. It was probably me that got it wrong but since you're here you can correct it :)

As long as codegen confirms that the reliability call is entirely compiled out i'm fine with it as you've ported it.

panoskj:
> My personal preference would be to use Interlocked.MemoryBarrier everywhere if possible. It fits more closely with the logical intention than using the Thread version to me. Either version would work identically so it's not a huge issue.

I agree, that's why I chose Interlocked.MemoryBarrier.

> For spans since they're structs comparing against null has never made sense to me and I'd prefer to have IsEmpty used everywhere if possible. It was probably me that got it wrong but since you're here you can correct it :)

Comparing against null is a indeed confusing, because of the implicit conversion. You can use Span.Empty instead. However it isn't exactly the same as checking IsEmpty in case of an empty array. In which case, the null-check triggers Debug.Assert while IsEmpty doesn't. But in the end of the day I think it doesn't really matter, since it is just a debug assertion and the code below it handles both null and empty arrays the same way.

> As long as codegen confirms that the reliability call is entirely compiled out i'm fine with it as you've ported it.

I think we could add Debug.Assert or throw an exception inside the stub, to verify it is omitted without having to look at the codegen. An exception would also make tests fail. What do you think?

David-Engel (approved):

David-Engel:
image
What was this? A pull and rebase? A merge on top of your previous commits would have been clear. Please avoid anything that requires a force push on a PR under active review. It makes it really difficult to tell what was changed since the last time someone reviewed. It essentially invalidates all reviewed commits as we can't tell what's changed. This causes a delay in getting your PR reviewed and merged.

panoskj:
@David-Engel My bad, you are right. I will force-push the previously reviewed commits and add a merge commit on top.

David-Engel (approved):

JRahnama (approved):

PR #2170, opened 2023-09-26
Support For Big Endian Systems

saitama951:
The Microsoft SqlClient driver crashes on the s390x. This error stems mostly from the use of BitConverter function used in method definitions across the code and improper byte ordering for big endian systems.

The BitConverter class does conversions according to the endianess of the system it is operating on which causes incorrect reads of little-endian packets sent by the server.


saitama951:
Hi, @Wraith2 @David-Engel can you please review this?
Thanks

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -1763,7 +1760,13 @@ internal byte[] SerializeFloat(float v)
                 throw ADP.ParameterValueOutOfRange(v.ToString());
             }
 
+#if NETCOREAPP
+            var bytes = new byte[4];
+            BinaryPrimitives.WriteInt32LittleEndian(bytes, BitConverter.SingleToInt32Bits(v));
+            return bytes;
```
Wraith2, line 1766: Is there a way to do this using the span we already have instead of allocating a new byte[]? If an array is needed it could be stackalloc-ed since we know it is fixed size.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -1886,7 +1889,13 @@ internal byte[] SerializeDouble(double v)
                 throw ADP.ParameterValueOutOfRange(v.ToString());
             }
 
+#if NETCOREAPP
+            byte[] bytes = new byte[8];
+            BinaryPrimitives.WriteInt64LittleEndian(bytes, BitConverter.DoubleToInt64Bits(v));
+            return bytes;
```
Wraith2, line 1895: same here, can we avoid the array allocation entirely or at least use the stack?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -7490,7 +7530,20 @@ internal Task WriteString(string s, int length, int offset, TdsParserStateObject
...
-            Buffer.BlockCopy(source, sourceOffset, dest, destOffset, charLength * ADP.CharSize);
+            if (!BitConverter.IsLittleEndian)
+            {
+                int desti = 0;
+                for(int srci = 0; srci < charLength; srci++)
+                {
+                    dest[desti + destOffset] = (byte)(source[srci + sourceOffset]);
+                    dest[desti + destOffset+1] = (byte)(source[srci + sourceOffset] >> 8);
+                    desti += 2;
+                }
```
Wraith2, line 7541: why not `BinaryPrimitives.ReverseEndianness(ushort)` ?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObject.netcore.cs
```diff
@@ -1793,6 +1826,16 @@ private void SetBufferSecureStrings()
...
+                            if (!BitConverter.IsLittleEndian)
+                            {
+                                byte temp;
+                                for (int ii = 0; ii < _securePasswords[i].Length * 2; ii += 2)
+                                {
+                                    temp = (byte)data[ii];
+                                    data[ii] = (byte)data[ii + 1];
+                                    data[ii + 1] = (byte)temp;
+                                }
+                            }
```
Wraith2, line 1838: Have you tested this? I wouldn't expect to need to byteswap protocol fields, the spec usually defines a specific endianness for them.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlSequentialTextReader.cs
```diff
@@ -514,6 +514,10 @@ public override void Convert(byte[] bytes, int byteIndex, int byteCount, char[]
                 completed = (bytesUsed == byteCount);
 
                 // BlockCopy uses offsets\length measured in bytes, not the actual array index
+                if (!BitConverter.IsLittleEndian)
+                {
+                    bytes = Encoding.Unicode.GetBytes(Encoding.BigEndianUnicode.GetString(bytes));
```
Wraith2, line 519: Can this byteswap in-place rather than allocating?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
```diff
@@ -488,6 +489,19 @@ internal bool TryStartNewRow(bool isNullCompressed, int nullBitmapColumnsCount =
...
+
+                for (int ii = 0; ii < charsCopied; ii++)
+                {
+                    if (!TryReadChar(out chars[charsOffset + ii]))
+                    {
+                        return false;
+                    }
+                }
+                return true;
+            }
```
Wraith2, line 504: isn't this going to cause a performance problem? in the BE case it'll do char by char reads rather than copying in chunks. It's be better to read the data completely and then do a byteswap just before converting to a string, otherwise any position where we're off-by-one in bytes would do an incorrect swap.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsValueSetter.cs
```diff
@@ -677,8 +690,13 @@ internal void SetDateTimeOffset(DateTimeOffset value)
             int days = utcDateTime.Subtract(DateTime.MinValue).Days;
             short offset = (short)value.Offset.TotalMinutes;
 
+#if NETCOREAPP
+            _stateObj.WriteByteArray(BitConverter.GetBytes(BinaryPrimitives.ReadInt64LittleEndian(BitConverter.GetBytes(time))), length - 5, 0); // time
+            _stateObj.WriteByteArray(BitConverter.GetBytes(BinaryPrimitives.ReadInt32LittleEndian(BitConverter.GetBytes(days))), 3, 0); // date
```
Wraith2, line 695: covert, reverse, convert again? doesn't seem sensible, there should be a better way to do this.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNICommon.cs
```diff
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
+using System.Buffers.Binary;
```
Wraith2, line 6: does the using need to be #ifdefed for netfx or does it exist on all targets?

Wraith2 (changes requested):

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObject.netcore.cs
saitama951: The SqlCredentialTest's from the manual test suite were failing. I assume the bytes of the password data were not rearranged before calling ObuscatePassword (I guress that's were all protocol related stuff is done and then written back into the packet), please correct me I'am wrong.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsValueSetter.cs
saitama951: It's seems that BinaryPrimitive.Read...LittleEndian(byte[]) returns data type value while BinaryPrimitives.Write...LittleEndian(out byte[], datatype value) returns a bool type. there is not really a good way to deal with the help of Read and Write Functions. Although we can do something like this only for BE.
```
BitConvert.GetBytes(BinaryPrimitives.ReverseEndianness(value));
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
saitama951: we have to explicity cast it to (byte) and we lose 1 byte so we would have to shift it again to incorporate in the dest byte array which these two statements already do

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
saitama951: The Function returns an byte[], using stackalloc will free the memory after the return

saitama951:
@Wraith2 Thank you for the suggestions, I have made the changes. not sure why the CI is failing.

JRahnama:
@saitama951 failures are not related to your changes. Somehow starting SqlBrowser task fails on the pipelines. We will look into them.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNICommon.cs
```diff
@@ -59,10 +62,17 @@ public void Read(byte[] bytes)
         {
             SMID = bytes[0];
             flags = bytes[1];
+#if NETCOREAPP
+            sessionId = BinaryPrimitives.ReadUInt16LittleEndian(new ReadOnlySpan<byte>(bytes, 2, 2));
```
panoskj, line 66: nit: `bytes.AsSpan(2, 2)` would be a bit more readable in my opinion, these lines are already too long.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
panoskj: Maybe it will be slightly faster if you perform one read operation rather than two (e.g. `char c = source[srci + sourceOffset]` than use `c`). I have no idea whether the compiler does this optimization automatically.

panoskj: You could also do `BinaryPrimitives.WriteInt16LittleEndian(dest.AsSpan(desti + destOffset), (short)source[srci + sourceOffset]);` in the loop.

Since the system is not little endian, writing in little endian effectively reverses the endianness. I think @Wraith2 meant something along these lines.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObject.netcore.cs
```diff
@@ -1793,6 +1828,16 @@ private void SetBufferSecureStrings()
...
```
panoskj, line 1840: An alternative way to do this would be:
```
for (int i = 0; i < data.Length; i += 2)
{
    var span = data.AsSpan(i);
    var value = BinaryPrimitives.ReadInt16LittleEndian(span);
    BinaryPrimitives.WriteInt16BigEndian(span, value);
}
```

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsValueSetter.cs
```diff
@@ -586,9 +589,17 @@ internal void SetDateTime(DateTime value)
                     if (SqlDbType.DateTime2 == _metaData.SqlDbType)
                     {
                         long time = value.TimeOfDay.Ticks / TdsEnums.TICKS_FROM_SCALE[_metaData.Scale];
+#if NETCOREAPP
+                        _stateObj.WriteByteArray(BitConverter.GetBytes(BinaryPrimitives.ReadInt64LittleEndian(BitConverter.GetBytes(time))), (int)_metaData.MaxLength - 3, 0);
```
panoskj, line 593: I think what you're looking for is something like this:
```
Span<byte> result = stackalloc byte[8];
BinaryPrimitives.WriteInt64LittleEndian(result, time);
_stateObj.WriteByteArray(result.AsSpan(0, (int)_metaData.MaxLength - 3));
```
You can follow this example to fix the rest of the places where you used BitConverter.GetBytes twice. Should address @Wraith2's [concerns](https://github.com/dotnet/SqlClient/pull/2170/commits/4114d8af84b06f519b2f3e1c453f729b39d56d74#r1338730042).

On a side note, this snippet uses 8 bytes from stack instead of 8 bytes from heap, so it seems to be an improvement over the original code in terms of performance.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.NetCoreApp.cs
```diff
@@ -4,16 +4,29 @@
...
 namespace Microsoft.Data.SqlClient
 {
     internal sealed partial class TdsParser
     {
         internal static void FillGuidBytes(Guid guid, Span<byte> buffer) => guid.TryWriteBytes(buffer);
 
-        internal static void FillDoubleBytes(double value, Span<byte> buffer) => BitConverter.TryWriteBytes(buffer, value);
+        internal static void FillDoubleBytes(double value, Span<byte> buffer) => 
+#if NETCOREAPP
+        BinaryPrimitives.WriteDoubleLittleEndian(buffer, value);
```
panoskj, line 19: Is WriteDoubleBigEndian available in all build targets? The [documentation](https://learn.microsoft.com/en-us/dotnet/api/system.buffers.binary.binaryprimitives.writedoublebigendian?view=net-8.0) says it is supported since .NET 5. Same as ReadDoubleBigEndian, which you are not using.

You also used WriteSingleLittleEndian in the next few lines.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -1743,14 +1746,10 @@ internal void WriteInt(int v, TdsParserStateObject stateObj)
 
         internal static void WriteInt(Span<byte> buffer, int value)
         {
-#if NETCOREAPP
-            BitConverter.TryWriteBytes(buffer, value);
```
panoskj, line 1747: Why remove the netcore special case rather than using BinaryPrimitives.WriteInt32LittleEndian in place of BitConverter.TryWriteBytes?

Wraith2:
I'm a little concerned about the number of preprocessor regions. If there's any way to extract common patterns into functions so we have fewer ifdefs it'd be nice. Otherwise without the ability to test i'll go with your changes as long as you've got the tests working.

Out of interest are you using managed networking? or is there a big endian native sni.dll ?

saitama951:
> I'm a little concerned about the number of preprocessor regions. If there's any way to extract common patterns into functions so we have fewer ifdefs it'd be nice. Otherwise without the ability to test i'll go with your changes as long as you've got the tests working.
> 
> Out of interest are you using managed networking? or is there a big endian native sni.dll ?

I agree with you  the if defs are unnecessarily populating the code, I do see the csproj includes a package System.Memory which supports BinaryPrimitives for all the targets but some functions are not supported with netstandard2.0 so I have added a file with the compatibility support, let me know your thoughts on this.
to answer your second question it uses the managed networking dll.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.NetCoreApp.cs
panoskj: Exactly what I was thinking. I didn't try to compile it to verify there is an error, but it looks like it.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
panoskj: The idea is to use the most performant code for methods such as WriteInt because some of the code you modified can run in a hot path/loop and affect performance. I'm not sure about WriteInt and the difference here is probably negligible but the places where you call GetBytes twice in the same row will certainly affect performance in some use cases, for example.

Wraith2: If it's possible to give both big and little endian paths good performance and switch between inline them using BitConverter.IsLittleEndian then that is the preferred method. The jit will eliminate any unused code paths.

If getting equal (at best) or adequate (at worst) perf for big endian systems requires a preprocessor definition and that change is done in multiple places, then abstracting that to a another method is preferred.

The overall goal is to get equal performance wherever possible. 
We should not degrade performance for existing targets to enable new targets unless it is unavoidable.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNICommon.cs
```diff
@@ -59,10 +60,10 @@ public void Read(byte[] bytes)
...
             SMID = bytes[0];
             flags = bytes[1];
-            sessionId = BitConverter.ToUInt16(bytes, 2);
-            length = BitConverter.ToUInt32(bytes, 4) - SNISMUXHeader.HEADER_LENGTH;
-            sequenceNumber = BitConverter.ToUInt32(bytes, 8);
-            highwater = BitConverter.ToUInt32(bytes, 12);
+            sessionId = BinaryPrimitives.ReadUInt16LittleEndian(new ReadOnlySpan<byte>(bytes, 2, 2));
+            length = BinaryPrimitives.ReadUInt32LittleEndian(new ReadOnlySpan<byte>(bytes, 4, 4)) - SNISMUXHeader.HEADER_LENGTH;
+            sequenceNumber = BinaryPrimitives.ReadUInt32LittleEndian(new ReadOnlySpan<byte>(bytes, 8, 4));
+            highwater = BinaryPrimitives.ReadUInt32LittleEndian(new ReadOnlySpan<byte>(bytes, 12, 4));
```
Wraith2, line 66: Note for other reviewers. the MC-SMP spec which defines the smux header says in 2.1 "All integer fields are represented in little-endian byte order.", these are the LENGTH, SEQNO and WNW fields.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
saitama951: done

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsValueSetter.cs
saitama951: I tried this suggestion it seems to be throwing a compilation error. it seems `Span<byte> doesn't have a overloading method AsSpan()` .  as a turnaround how about this?
```
 unsafe
 {
      var result = stackalloc byte[8];
      BinaryPrimitives.WriteInt64LittleEndian(new Span<byte>(result,8), time);
      _stateObj.WriteByteSpan(new Span<byte>(result,(int)_metaData.MaxLength - 3)); 
}
```
@panoskj

panoskj: I'm also not sure why you added `unsafe`.

saitama951: `result` is a byte * so it should be used in a unsafe context. right?

saitama951: done

saitama951:
@panoskj I have made the changes. Thank you for the suggestions

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -7490,7 +7495,20 @@ internal Task WriteString(string s, int length, int offset, TdsParserStateObject
...
         {
-            Buffer.BlockCopy(source, sourceOffset, dest, destOffset, charLength * ADP.CharSize);
+            if (!BitConverter.IsLittleEndian)
+            {
+                int desti = 0;
+                var span = dest.AsSpan();
+                for(int srci = 0; srci < charLength; srci++)
+                {
+                    BinaryPrimitives.WriteUInt16LittleEndian(span.Slice(desti + destOffset), (ushort)source[srci + sourceOffset]);
+                    desti+=2;
```
panoskj, line 7505: Still some missing spaces. Also in for loop above.

saitama951: done thanks

saitama951:
Is this good to merge?

Wraith2:
You need to wait for the MS team to have time to review before it can be merged. That may take several months.

panoskj (changes requested):

panoskj:
@saitama951 I added some comments the other day but I forgot to make them visible. One of them is a bug that has to be fixed. And yeah, you have to wait for MS team to review them too.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -5837,8 +5842,8 @@ internal bool DeserializeUnencryptedValue(SqlBuffer value, byte[] unencryptedByt
                             return false;
                         }
 
-                        mid = BitConverter.ToInt32(unencryptedBytes, 0);
-                        lo = BitConverter.ToUInt32(unencryptedBytes, 4);
+                        mid = BinaryPrimitives.ReadInt32LittleEndian(unencryptedBytes);
+                        lo = BinaryPrimitives.ReadUInt32LittleEndian(unencryptedBytes);
```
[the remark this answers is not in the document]
saitama951: oops, done thanks

David-Engel (approved):
In general, I'm fine with the changes.

Note: We don't have big endian systems to test on, so we'll be reliant on external reports to know if anything breaks in those environments in the future.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObject.netcore.cs
```diff
@@ -1793,6 +1794,15 @@ private void SetBufferSecureStrings()
                             str = Marshal.SecureStringToBSTR(_securePasswords[i]);
                             byte[] data = new byte[_securePasswords[i].Length * 2];
                             Marshal.Copy(str, data, 0, _securePasswords[i].Length * 2);
+                            if (!BitConverter.IsLittleEndian)
+                            {
+                                var span = data.AsSpan();
+                                for (int ii = 0; ii < _securePasswords[i].Length * 2; ii += 2)
+                                {
+                                    var value = BinaryPrimitives.ReadInt16LittleEndian(span.Slice(ii));
```
DavoudEshtehari, line 1802:
```suggestion
                                Span<byte> span = data.AsSpan();
                                for (int ii = 0; ii < _securePasswords[i].Length * 2; ii += 2)
                                {
                                    short value = BinaryPrimitives.ReadInt16LittleEndian(span.Slice(ii));
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -1744,7 +1745,7 @@ internal void WriteInt(int v, TdsParserStateObject stateObj)
         internal static void WriteInt(Span<byte> buffer, int value)
         {
 #if NETCOREAPP
-            BitConverter.TryWriteBytes(buffer, value);
+            BinaryPrimitives.WriteInt32LittleEndian(buffer, value);
```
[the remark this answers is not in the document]
saitama951: That's a good point, I didn't notice that, in my opinion it is better to preserve the try- variants there were previously and maintain the behaviour.

DavoudEshtehari:
My concern is how this change is going to impact the performance by impacting the hot paths. Does anybody have a perf test to compare it with before the change?

saitama951:
@DavoudEshtehari 
I did manage to run the performance tests and here are the results
note:- the base= without the patch and diff= with the patch
```
summary:
better: 1, geomean: 3.491
worse: 42, geomean: 1.254
total diff: 43

| Slower                                                                                        | diff/base | Base Median (ns) | Diff Median (ns) | Modality  |
| --------------------------------------------------------------------------------------------- | ---------:| ----------------:| ----------------:| ---------- |
| Microsoft.Data.SqlClient.PerformanceTests.SqlCommandRunner.ExecuteScalar                      |      1.52 |      81632944.25 |     123795769.75 | several?  |
| Microsoft.Data.SqlClient.PerformanceTests.SqlCommandRunner.ExecuteXmlReader                   |      1.41 |     362968055.00 |     511413731.50 | several?  |
| Microsoft.Data.SqlClient.PerformanceTests.DataTypeReaderRunner.DateTimeOffset                 |      1.40 |      17359631.75 |      24297528.00 | several?  |
| Microsoft.Data.SqlClient.PerformanceTests.DataTypeReaderRunner.SmallDateTime                  |      1.36 |      19801378.00 |      27024967.50 | bimodal   |
| Microsoft.Data.SqlClient.PerformanceTests.DataTypeReaderRunner.Decimal                        |      1.35 |      18220311.50 |      24638132.50 | several?  |
| Microsoft.Data.SqlClient.PerformanceTests.DataTypeReaderRunner.VarBinary                      |      1.35 |      24514042.75 |      33120146.50 |           |
| Microsoft.Data.SqlClient.PerformanceTests.DataTypeReaderRunner.NChar                          |      1.35 |      18399459.50 |      24820209.00 | several?  |
| Microsoft.Data.SqlClient.PerformanceTests.DataTypeReaderRunner.NVarChar                       |      1.34 |      25595037.50 |      34186686.75 | several?  |
| Microsoft.Data.SqlClient.PerformanceTests.DataTypeReaderRunner.Datetime                       |      1.33 |      18858414.50 |      25097611.75 | several?  |
| Microsoft.Data.SqlClient.PerformanceTests.DataTypeReaderRunner.Xml                            |      1.33 |      30359710.00 |      40349804.00 | several?  |
| Microsoft.Data.SqlClient.PerformanceTests.DataTypeReaderRunner.Char                           |      1.31 |      18560508.00 |      24309393.25 |           |
| Microsoft.Data.SqlClient.PerformanceTests.SqlCommandRunner.ExecuteNonQuery                    |      1.31 |      97676723.75 |     127612822.00 | several?  |
| Microsoft.Data.SqlClient.PerformanceTests.DataTypeReaderRunner.Datetime2                      |      1.29 |      19728887.00 |      25509330.50 | several?  |
| Microsoft.Data.SqlClient.PerformanceTests.DataTypeReaderRunner.Binary                         |      1.29 |      18213765.00 |      23410165.50 | several?  |
| Microsoft.Data.SqlClient.PerformanceTests.DataTypeReaderAsyncRunner.NVarCharAsync             |      1.28 |      30038761.25 |      38505020.00 |           |
| Microsoft.Data.SqlClient.PerformanceTests.DataTypeReaderRunner.Real                           |      1.27 |      18352333.00 |      23353771.75 | several?  |
| Microsoft.Data.SqlClient.PerformanceTests.DataTypeReaderRunner.Date                           |      1.26 |      18424424.50 |      23286289.25 | several?  |
| Microsoft.Data.SqlClient.PerformanceTests.DataTypeReaderRunner.Float                          |      1.26 |      18732205.75 |      23654323.75 |           |
| Microsoft.Data.SqlClient.PerformanceTests.DataTypeReaderRunner.SmallMoney                     |      1.25 |      19230988.00 |      24130540.50 |           |
| Microsoft.Data.SqlClient.PerformanceTests.DataTypeReaderRunner.Numeric                        |      1.24 |      19406590.00 |      24090994.00 | several?  |
| Microsoft.Data.SqlClient.PerformanceTests.DataTypeReaderAsyncRunner.VarCharAsync              |      1.24 |      29144578.50 |      36085790.00 | bimodal   |
| Microsoft.Data.SqlClient.PerformanceTests.DataTypeReaderRunner.VarChar                        |      1.24 |      25736796.75 |      31856117.50 | multimodal|
| Microsoft.Data.SqlClient.PerformanceTests.DataTypeReaderAsyncRunner.NumericAsync              |      1.24 |      24404761.50 |      30154729.50 |           |
| Microsoft.Data.SqlClient.PerformanceTests.DataTypeReaderAsyncRunner.BitAsync                  |      1.22 |      23455534.50 |      28669823.25 | bimodal   |
| Microsoft.Data.SqlClient.PerformanceTests.DataTypeReaderAsyncRunner.CharAsync                 |      1.21 |      24139974.50 |      29258859.50 | bimodal   |
| Microsoft.Data.SqlClient.PerformanceTests.DataTypeReaderAsyncRunner.DateTimeOffsetAsync       |      1.21 |      23872583.00 |      28934241.75 |           |
| Microsoft.Data.SqlClient.PerformanceTests.DataTypeReaderRunner.Time                           |      1.21 |      20120527.00 |      24290295.25 | several?  |
| Microsoft.Data.SqlClient.PerformanceTests.DataTypeReaderAsyncRunner.IntAsync                  |      1.20 |      24145428.25 |      29087973.25 |           |
| Microsoft.Data.SqlClient.PerformanceTests.DataTypeReaderAsyncRunner.DecimalAsync              |      1.20 |      25246572.00 |      30361482.00 |           |
| Microsoft.Data.SqlClient.PerformanceTests.DataTypeReaderAsyncRunner.Datetime2Async            |      1.20 |      25484344.75 |      30597239.50 |           |
| Microsoft.Data.SqlClient.PerformanceTests.DataTypeReaderAsyncRunner.DatetimeAsync             |      1.20 |      25987078.75 |      31198285.00 |           |
| Microsoft.Data.SqlClient.PerformanceTests.DataTypeReaderAsyncRunner.VarBinaryAsync            |      1.20 |      30502607.50 |      36583319.25 |           |
| Microsoft.Data.SqlClient.PerformanceTests.DataTypeReaderAsyncRunner.SmallIntAsync             |      1.20 |      24918030.75 |      29813608.25 | bimodal   |
| Microsoft.Data.SqlClient.PerformanceTests.SqlBulkCopyRunner.BulkCopy_IDataReader(Columns: 25) |      1.19 |     632961417.50 |     753515670.50 | several?  |
| Microsoft.Data.SqlClient.PerformanceTests.DataTypeReaderAsyncRunner.SmallDateTimeAsync        |      1.18 |      26242335.25 |      30973506.75 | several?  |
| Microsoft.Data.SqlClient.PerformanceTests.DataTypeReaderAsyncRunner.TimeAsync                 |      1.18 |      25614011.75 |      30228362.25 |           |
| Microsoft.Data.SqlClient.PerformanceTests.DataTypeReaderAsyncRunner.SmallMoneyAsync           |      1.18 |      25276536.00 |      29796326.75 | multimodal|
| Microsoft.Data.SqlClient.PerformanceTests.DataTypeReaderAsyncRunner.BigIntAsync               |      1.16 |      25780506.25 |      30006454.00 | bimodal   |
| Microsoft.Data.SqlClient.PerformanceTests.DataTypeReaderAsyncRunner.UniqueIdentifierAsync     |      1.16 |      24704125.75 |      28671626.50 |           |
| Microsoft.Data.SqlClient.PerformanceTests.DataTypeReaderAsyncRunner.NCharAsync                |      1.16 |      24424800.50 |      28304558.75 | bimodal   |
| Microsoft.Data.SqlClient.PerformanceTests.DataTypeReaderAsyncRunner.XmlAsync                  |      1.15 |      34468910.50 |      39579307.75 | several?  |
| Microsoft.Data.SqlClient.PerformanceTests.DataTypeReaderAsyncRunner.BinaryAsync               |      1.14 |      25671757.00 |      29162259.00 |           |

| Faster                                                               | base/diff | Base Median (ns) | Diff Median (ns) | Modality|
| -------------------------------------------------------------------- | ---------:| ----------------:| ----------------:| -------- |
| Microsoft.Data.SqlClient.PerformanceTests.DataTypeReaderRunner.Money |      3.49 |      84750981.50 |      24273529.00 | bimodal |
```

I did manage to run the perf profiler to look out for hotspots in the code for the first couple of benchmarks.
what I could find is that most of the hotspots are in `BinaryPrimitives` i.e `ReadUInt32`

I'm looking into the other regressions as well 
NOTE :- There might be false positives among these, since these are running on a VM  (both server and client).

saitama951:
@DavoudEshtehari 
I'm looking into the perf tests.

saitama951:
@DavoudEshtehari 
Hi,
I did run the performance tests and I'am seeing some noise and inconsistency in the results with the patch and without the patch. any guidance on this would help.
Thanks

Wraith2:
They're quite noisy. Unless the differences are very large I wouldn't worry about it too much.

DavoudEshtehari:
@saitama951 Make sure there is no heavy background process running while your test is running.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/BitConverterCompat.cs
```diff
@@ -0,0 +1,18 @@
+// See the LICENSE file in the project root for more information.
+
+using System;
+
+namespace Microsoft.Data.SqlClient
+{
+    internal static class BitConverterCompat
```
H-Yeo, line 7: Did we truncate a word here/spelling mistake? Are you trying to say Compatible/Compact? I think it's okay to make it more clear if you need to say Compatible.

DavoudEshtehari:
@saitama951 May I ask to share the perf project and share the result from your perf test result?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/BitConverterCompat.cs
saitama951: @H-Yeo Yes, This was meant to be Compatible (since certain functions of bitconverter were not supported in older frameworks). I will make the necessary change .Thanks

saitama951:
@DavoudEshtehari Sorry for the delay in response, I'am running the performance tests present in the [SqlClient](https://github.com/dotnet/SqlClient/tree/main/src/Microsoft.Data.SqlClient/tests/PerformanceTests)  and was using the [ResultComparer](https://github.com/dotnet/performance/tree/main/src/tools/ResultsComparer) tool to compare the results with and without the patch. Initially I was facing a lot of issue in generating a consist data with various runs. As if now I'am able to pertain the consistency, here are the perf results 
[run-1.xlsx](https://github.com/dotnet/SqlClient/files/13591003/run-1.xlsx)
[run-2.xlsx](https://github.com/dotnet/SqlClient/files/13591004/run-2.xlsx)
[run-3.xlsx](https://github.com/dotnet/SqlClient/files/13591005/run-3.xlsx)
[run-4.xlsx](https://github.com/dotnet/SqlClient/files/13591006/run-4.xlsx)

JRahnama (approved):

JRahnama:
@saitama951 can you address the conflict so we can proceed with the preview release today?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
```diff
@@ -1152,7 +1152,6 @@ public bool TryReadByteArray(Span<byte> buff, int len)
         // Every time you call this method increment the offset and decrease len by the value of totalRead
         public bool TryReadByteArray(Span<byte> buff, int len, out int totalRead)
         {
-            TdsParser.ReliabilitySection.Assert("unreliable call to ReadByteArray");  // you need to setup for a thread abort somewhere before you call this method
```
panoskj, line 1147: These shouldn't be removed.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
```diff
@@ -1174,7 +1173,7 @@ public bool TryReadByteArray(Span<byte> buff, int len, out int totalRead)
             }
 #endif
 
-            Debug.Assert(buff.IsEmpty || buff.Length >= len, "Invalid length sent to ReadByteArray()!");
+            Debug.Assert(buff == null || buff.Length >= len, "Invalid length sent to ReadByteArray()!");
```
panoskj, line 1176: This shouldn't be changed.

panoskj (changes requested):
You have reverted some changes I introduced with #2168 . Note this file is shared between netcore and netfx versions.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
saitama951: yes, there was some mess up,I'am adding them back

panoskj:
@saitama951 The last 3 commits look good now. I haven't reviewed the merge commit at all, it will take me some time.

JRahnama (approved):

PR #2202, opened 2023-11-01
Test | Skip Host Name In Certificate Unit Test for Named Instance Sql Server

arellegue:


arellegue:
This PR is no longer needed as another PR replaces it,

Issue #2222, opened 2023-11-16
Missing functionality in batch API


mgravell:
Relates to https://github.com/dotnet/SqlClient/pull/1825

Testing local build (`main`) with Dapper shows a few problems, i.e. "it doesn't work"; severity: IMO critical, this should absolutely be fixed before next build; the batch API is not usable by non-provider scenarios without it

1. `SqlConnection` fails to `override` the `CanCreateBatch` and `CreateDbBatch()` members (net6+)
2. `SqlBatchCommand` *does* `override` the `CanCreateParameter` and `CreateParameter()` members (net8+), but: the library does not currently build for net8, so: this code is never actually compiled as far as I can see

Suggestions:

for `SqlConnection`:

``` csharp
#if NET6_0_OR_GREATER
        /// <inheritdoc/>
        public override bool CanCreateBatch => true;

        /// <inheritdoc/>
        protected override SqlBatch CreateDbBatch() => new SqlBatch(this);
#endif
```

(you may have local rules on docs, especially since we're doing type variance here)

---

in the csproj:

- add `net8.0` to `<TargetFrameworks>`

---

tests:

for net6+

``` csharp
using var cnn = new SqlConnection();
Assert.True(cnn.CanCreateBatch);
var batch = cnn.CreateBatch();
Assert.Is<SqlBatch>(batch);
Assert.Same(cnn, batch.Connection);
Assert.Null(batch.Transaction);
```

for net8+

``` csharp
var cmd = new SqlBatchCommand();
Assert.True(cnn.CanCreateParameter);
Assert.Is<SqlParameter>(cnn.CreateParameter());
```

mgravell:
note that adding net8 triggers a few `[Obsolete]` warnings, CS0672; requires

```
#if NET8_0_OR_GREATER
        [Obsolete]
#endif
```

on

- `SqlException.GetObjectData`
- `SqlException.ctor(SerializationInfo, StreamingContext)`
- `OperationAbortedException.ctor(SerializationInfo, StreamingContext)`

I also get a build error from `Microsoft.Data.SqlClient.notsupported.cs` via `GetnAPI`; not sure what to do about that one, but: it needs fixing

Wraith2:
I'll take a look.

mgravell:
awesome; I'll go ahead and do what I can with my locally hacked version that targets net7 with the first half of the fixes; I can't test the net8 parameter fix, but... that's the small end of the wedge for this

PR #2223, opened 2023-11-16
SqlBatch: Add missing overrides and tests

Wraith2:
fixes https://github.com/dotnet/SqlClient/issues/2222

Add missing overrides for SqlConnection CanCreateBatch and CreateDbBatch() and tests
Add test for SqlBatchCommand CanCreateParameter and CreateParameter() wrapped in #ifdef
Adding a net8 build is out of scope for this PR. That needs approval at very least from the MS side. When it is added the net8 features and tests related to the sql batch feature should light up automatically.

@David-Engel we'll want to get this in before the next preview. 

/cc @mgravell

### DIFF
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/Batch/BatchTests.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/Batch/BatchTests.cs
@@ -12,7 +12,7 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests
 {
     public static class BatchTests
     {
-        [SkipOnTargetFramework(TargetFrameworkMonikers.NetFramework)]
+
         [ConditionalFact(typeof(DataTestUtility), nameof(DataTestUtility.AreConnStringsSetup))]
         public static void MissingCommandTextThrows()
         {
@@ -24,7 +24,6 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests
             }
         }
 
-        [SkipOnTargetFramework(TargetFrameworkMonikers.NetFramework)]
         [ConditionalFact(typeof(DataTestUtility), nameof(DataTestUtility.AreConnStringsSetup))]
         public static void MissingConnectionThrows()
         {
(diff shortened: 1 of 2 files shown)


roji:
@Wraith2 I think you also want to override the related methods missing on SqlClientFactory ([here are the original specs](https://github.com/dotnet/runtime/issues/28633)):

```c#
public virtual DbBatch CreateBatch() => throw new NotSupportedException();
public virtual DbBatchCommand CreateBatchCommand() => throw new NotSupportedException();
    
// Covers both CreateBatch and CreateBatchCommand
public virtual bool CanCreateBatch => false;
```

Wraith2:
Those are in a separate file:
https://github.com/dotnet/SqlClient/blob/d30d9d107eca0f11b735a550aa187e57167858c9/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlClientFactory.NetCoreApp.cs#L9-L18

Wraith2:
They should light up as soon as the team add a net8 build.

David-Engel:
Do these APIs need to be in the ref assemblies, too?

Wraith2:
They're overrides so I wouldn't expect to see them in the ref assembly since they're defined on the base class and we didn't merge our own copies of the base classes.

David-Engel (approved):

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/Batch/BatchTests.cs
```diff
@@ -12,6 +12,7 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests
 {
     public static class BatchTests
     {
+        [SkipOnTargetFramework(TargetFrameworkMonikers.NetFramework)]
```
DavoudEshtehari, line 15: NIT, skipping the target framework doesn't seem necessary since this file only included with net6.0 compatible versions in the csproj file.

Wraith2: removed them all in this file

DavoudEshtehari (approved):

PR #2224, opened 2023-11-17
Fix | Don't validate server certificate when using Authentication option but not encrypting

David-Engel:
When using the authentication option (Authentication=SqlPassword or other value), the netfx code would try to validate the server certificate regardless of whether or not encryption was negotiated. This resulted in the requirement to set TrustServerCertificate=true, even if client and server encryption was off, in order for connections to succeed when the server doesn't have a public certificate provisioned. This also manifests in the same manner when using Authentication=ActiveDirectoryIntegrated. In a Kerberos flow, that option behaves like IntegratedSecurty=true.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
+++ b/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
@@ -1485,12 +1485,10 @@ namespace Microsoft.Data.SqlClient
                             }
 
                             // Validate Certificate if Trust Server Certificate=false and Encryption forced (EncryptionOptions.ON) from Server.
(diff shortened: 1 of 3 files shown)


Matteo-T (approved):
Looks reasonably good to me :)

arellegue (approved):

DavoudEshtehari:
Could we have a unit test to cover this case if already there isn't one?

David-Engel:
> Could we have a unit test to cover this case if already there isn't one?

The test ADIntegratedUsingSSPI (renamed in this PR from the inaccurate ADInteractiveUsingSSPI) was exercising this bad path in our Kerberos pipelines but the old ones are configured with TrustServerCertificate=true to avoid this issue. The new Kerberos pipeline wasn't using TrustServerCertificate=true, so it was hitting this issue.

I added another test to specifically / force Authentication=SqlPassword and TrustServerCertificate=false to help ensure this doesn't regress in localhost pipelines.

DavoudEshtehari (approved):

Issue #2184, opened 2023-10-13
Consider changing diagnostic events to use public declared types as their payload


vitek-karas:
### Is your feature request related to a problem? Please describe.
The events produced by SqlClient to the diagnostic listener pass a payload object which is consumed by several different tools/libraries. The events use mostly anonymous types, the consumer has to use reflection to access the relevant data.

This has distinct disadvantages in performance impact and trim/AOT compatibility. Currently for example Open Telemetry library uses several different tools to overcome these:

* It includes a relatively complicated reflection based "cache" of accessors to make the access to the properties as fast as possible
* It has to code defensively to make it work in cases the reflection fails
* The properties may be trimmed away because there's no attempt in SqlClient to hint the trimmer to preserve them

Some other libraries use similar approach. For example, GRPC uses declared but internal types and makes an effort to preserve the necessary properties. But even that approach as its problems. The trimmer compatibility is not verifiable by any tools, since it relies on the combination of attributes in the GRPC code and careful reflection usage in the consuming library. The consuming library has to use suppressions which is always fragile. For more details on the similar GRPC problem see: https://github.com/grpc/grpc-dotnet/issues/2154

### Describe the solution you'd like
The shape of the payload object is effectively a public contract, since the consuming libraries hardcode the name of the properties and then cast the value of those properties to public types. For example [this code](https://github.com/open-telemetry/opentelemetry-dotnet/blob/865bcb64a5c67d713ded373a2839c49dba07dfc2/src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlClientDiagnosticListener.cs#L78-L84) gets the `Command` property from the payload instance, using reflection to get to it. The returned instance is still typed as `System.Object`, so it has to use additional reflection to get access to its properties and so on. If the library (SqlClient) changed the name of the property, it would be an effective breaking change, even though technically it's not changing any public property.

It seems it would be beneficial to consider this a public contract/API in all meanings of that. Mainly declare the type of the payload object as a public type so that the consumer library can just cast the payload instance and access everything on it directly. No need for complicated and less performant reflection based solutions.

### Describe alternatives you've considered
There are no good alternatives, every other solution will probably need to rely on reflection and will run into the same problems as described above.

### Additional context
[...]

PR #2226, opened 2023-11-17
Strong typed diagnostics

Wraith2:
fixes https://github.com/dotnet/SqlClient/issues/2184

Currently when emitting diagnostic events we use anonymous types. The only way to access the information in those types is using reflection which makes them unusable in reflection-free situations such as trimmed or Native AOT applications.

This PR replaces the anonymous types with dedicated types for each event. The new types are put into a new Microsoft.Data.SqlClient.Diagnostics namespace because no users are ever like to want to deal with those types directly and their names would be confusing it they commonly appears in intellisense listings. Documentation is added for all public members. Ref assembly is updated. Internal references to the names of the events have been updated (please check these carefully). 
The names of the events have moved from internal constants to public constants because I feel that the names provide part of the public contract for accessing these diagnostics. If there is pushback on this particular part of the PR I could fairly easily be persuaded to revert them. I'm not sure of it.

The changes I've made are based on the changes contributed to asp.net by @benaadams in https://github.com/dotnet/aspnetcore/pull/11730 . They provide strongly typed properties and optimized access through `IReadOnlyList<KeyValuePair<string,object>>` so callers do not need to take a direct dependency on our library they can simply try the cast to that interface and cleanly access all the data. I considered also adding `IReadOnlyDictionary<string,object>` allowing direct access to known named properties without a direct dependency. I can do that if it will be used by callers but wanted to have a discussion about it first.

/cc @kant2002 for trimming, @mbakalov for appinsights/otel oversight


Wraith2:
If you check the original issue you'll see that this is a suggested method of removing the anonymous types which require the use of reflection. So it doesn't make this library a lot safer but it allows consumer binaries to be safer consuming this one and so propagates trimability up the dependency stack.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlClientDiagnostic.cs
```diff
@@ -0,0 +1,676 @@
...
+
+        /// <inheritdoc/>>
+        IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
+
+        /// <include file='../../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlClientDiagnostic.xml' path='docs/members[@name="SqlClientDiagnostic"]/TryGetCommonProperty/*'/>
+        protected bool TryGetCommonProperty(int index, out KeyValuePair<string, object> property)
+        {
+            switch (index)
+            {
+                case 0:
```
JRahnama, line 70: I saw these index numbers are being used a lot in this class, would it be possible to pull them up as constant fields with more descriptive names just for code readability purpose?

Wraith2: I tried that while I was working on the first class I implemented and found that it hurt readability more than it helped.

Consider either:
```csharp
	private const int ConnectionIdIndex = 0;
	private const int TransactionIdIndex = 1;
	private const int CommandIdIndex = 2

        protected sealed override KeyValuePair<string, object> GetDerivedProperty(int index) => index switch
        {
            ConnectionIdIndex => new KeyValuePair<string, object>(nameof(ConnectionId), ConnectionId),
            TransactionIdIndex => new KeyValuePair<string, object>(nameof(TransactionId), TransactionId),
            CommandIdIndex => new KeyValuePair<string, object>(nameof(Command), Command),
            _ => throw new IndexOutOfRangeException(nameof(index)),
        };
```

or:

```csharp
        protected sealed override KeyValuePair<string, object> GetDerivedProperty(int index) => index switch
        {
	    const int ConnectionIdIndex = 0;
	    const int TransactionIdIndex = 1;
	    const int CommandIdIndex = 2

            ConnectionIdIndex => new KeyValuePair<string, object>(nameof(ConnectionId), ConnectionId),
            TransactionIdIndex => new KeyValuePair<string, object>(nameof(TransactionId), TransactionId),
            CommandIdIndex => new KeyValuePair<string, object>(nameof(Command), Command),
            _ => throw new IndexOutOfRangeException(nameof(index)),
        };
```

Both are much longer and require you to mentally match up the constants by name when their only realy importantce is that they're sequential. They're also used in only on place (because each class has different properties) so the constants are only ever used once. I settled with just using the numbers directly because it was shorter and overall clearer.

For the same reasons I'm not really totally happy with the Name constants because I feel they make things less readable. In the case of the name the idea of having a unique required string that people will need to copy paste to use the individual class but that string not being defined in the class felt more wrong than the additional verbosity of having it locally.

mbakalov:
Confirming we are good with this for otel - thank you! Checking AppInsights next...

Wraith2:
Any appinsights feedback? not expecting it to fail if otel worked but it'd be good to have confirmation.

JRahnama:
failures are not related to your changes. I have fixed the pipelines, if you can push something that should pass.

mbakalov:
Thanks! I agree on the right way. Will check with AppInsights and Otel if it is something that can be changed in the future.

Wraith2:
Base class removed and everything implemented directly on the new classes. Once the CI finishes can you retest on appinsights please.

mbakalov:
@Wraith2: confirm it works ok now for both Otel and AppInsights, thanks!

Wraith2:
@David-Engel any particular blocker on this?

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -1901,6 +1903,316 @@ public sealed class SqlConfigurableRetryFactory
...
+        protected const int CommonPropertyCount = 3;
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlClientDiagnostic.xml' path='docs/members[@name="SqlClientDiagnostic"]/ctor/*'/>
+        protected SqlClientDiagnostic(System.Guid operationId, string operation, long timestamp) { }
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlClientDiagnostic.xml' path='docs/members[@name="SqlClientDiagnostic"]/OperationId/*'/>
+        public System.Guid OperationId => throw null;
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlClientDiagnostic.xml' path='docs/members[@name="SqlClientDiagnostic"]/Operation/*'/>
+        public string Operation => throw null;
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlClientDiagnostic.xml' path='docs/members[@name="SqlClientDiagnostic"]/Timestamp/*'/>
+        public long Timestamp => throw null;
+        /// <inheritdoc/>>
```
David-Engel, line 1922: We found out we can't use inheritdoc because it doesn't work during API doc publishing. We can only inherit from API docs that are part of our own doc repo. 😞

Wraith2: Replaced with copied text from the base methods.

doc/snippets/Microsoft.Data.SqlClient/SqlClientDiagnostic.xml
```diff
@@ -0,0 +1,296 @@
...
+    <Timestamp>
+      <summary>The timestamp of the event.</summary>
+    </Timestamp>
+  </members>
+  <members name="SqlClientCommandBefore">
+    <SqlClientCommandBefore>
+      <summary>Contains diagnostic information emitted before a command is executed.</summary>
+    </SqlClientCommandBefore>
+    <Name>
+      <summary>The name of the event that that needs to be enabled for the event to be raised.</summary>
```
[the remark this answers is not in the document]
Wraith2: All fixed.

JRahnama:
@Wraith2 and of course a conflict had to happen here right on this PR when we are ready to merge.

Wraith2:
rebasing now. give me a minute to resolve things and check it.

Wraith2:
Done, we're in the hands of the CI now.

David-Engel (approved):

cheenamalhotra:
@Wraith2 Can you provide more details why you chose to keep Abstract class "only" in the ref project, while the SqlClientDiagnostic implementations in the main assembly? Why are API surfaces not consistent or was it an oversight?

Secondly, what is the rationale behind Abstract class if the types that emit telemetry don't implement it?

Wraith2:
It's a mistake. The abstract class was used to remove some duplication but one of the consumers only works with direct properties and ignores inherited ones. We need to correct the ref assembly to match the implementation.

Issue #2225, opened 2023-11-17
Result of System.AppContext.GetData is not being cached


JulianRooze:
In profiling our application under high load, we've found that calls to System.AppContext.GetData(string) make up ~3% of our total CPU time, making it a top 5 CPU consumer function overall for us. All these calls originate from this project.

The main argument it's being called with is `"Switch.Microsoft.Data.SqlClient.MakeReadAsyncBlocking"` which originates from this file:

https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/LocalAppContextSwitches.cs

I've filed a runtime [issue](https://github.com/dotnet/runtime/issues/94894) to potentially make AppContext.GetData more performant for reads, but one commenter in that issue pointed out that it's likely unintentional that it's being called this often as there are bool fields to cache the return value but it's only being assigned to and not read in two cases. 

I think the performance here can be improved by using the same kind of pattern as in the other properties of that class;

```csharp
 public static bool LegacyRowVersionNullBehavior
        {
            get
            {
                if (s_legacyRowVersionNullBehavior is null)
                {
                    bool result;
                    result = AppContext.TryGetSwitch(LegacyRowVersionNullString, out result) && result;
                    s_legacyRowVersionNullBehavior = result;
                }
                return s_legacyRowVersionNullBehavior.Value;
            }
        }
``` 

### Expected behavior

The AppContext.GetData is only performed once and then cached.

Wraith2:
Seems reasonable. I'll implement it.

JulianRooze:
Thanks so much @Wraith2 ! 🙌

PR #2227, opened 2023-11-17
cache all appcontext switches

Wraith2:
fixes https://github.com/dotnet/SqlClient/issues/2225

Change any uncached appcontext switches to use caching like the others.

/cc @JulianRooze


David-Engel (approved):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/LocalAppContextSwitches.cs
```diff
@@ -15,8 +15,8 @@ internal static partial class LocalAppContextSwitches
 
         private static bool? s_legacyRowVersionNullBehavior;
         private static bool? s_suppressInsecureTLSWarning;
-        private static bool s_makeReadAsyncBlocking;
-        private static bool s_useMinimumLoginTimeout;
+        private static bool? s_makeReadAsyncBlocking;
```
jkotas, line 18: bool? is a structure with two fields. It is not updated atomically. The lock-free reads and writes below have race conditions.

If you want the reads to be lock-free, you can either:
- Change the type of these caches to `bool` and initialize them eagerly in a static constructors
or
- Change the type of these caches to atomically updated integer with 3 possible values uninitialized/false/true. Here is an example of similar pattern in dotnet/runtime: https://github.com/dotnet/runtime/blob/b88c74a109903de213b049177cbc0481a689477f/src/libraries/System.Private.CoreLib/src/System/Globalization/TextInfo.cs#L22C22-L27

arellegue (approved):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/LocalAppContextSwitches.cs
Wraith2: Thanks. I've updated all the variables to use the tristate enum pattern.

David-Engel (approved):

jkotas:
Failing tests:
```
  [xUnit.net 00:01:04.53]       System.InvalidCastException : Unable to cast object of type 'Tristate' to type 'System.Nullable`1[System.Boolean]'.
  [xUnit.net 00:01:04.53]       Stack Trace:
  [xUnit.net 00:01:04.53]         D:\a\_work\1\s\src\Microsoft.Data.SqlClient\tests\ManualTests\SQL\DataReaderTest\DataReaderTest.cs(330,0): at Microsoft.Data.SqlClient.ManualTesting.Tests.DataReaderTest.SetLegacyRowVersionNullBehavior(Nullable`1 value)
  [xUnit.net 00:01:04.53]         D:\a\_work\1\s\src\Microsoft.Data.SqlClient\tests\ManualTests\SQL\DataReaderTest\DataReaderTest.cs(301,0): at Microsoft.Data.SqlClient.ManualTesting.Tests.DataReaderTest.CheckLegacyNullRowVersionIsEmptyArray()
  ```

**Comment by Wraith2** [comment]
Thanks. Fixed and added note to tests and source to indicate that private reflection is being used. I wrote that test so I should have remembered.

File: src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/LocalAppContextSwitches.cs [hunk 2]
```diff
@@ -8,15 +8,23 @@ namespace Microsoft.Data.SqlClient
...
         internal const string MakeReadAsyncBlockingString = @"Switch.Microsoft.Data.SqlClient.MakeReadAsyncBlocking";
         internal const string LegacyRowVersionNullString = @"Switch.Microsoft.Data.SqlClient.LegacyRowVersionNullBehavior";
         internal const string SuppressInsecureTLSWarningString = @"Switch.Microsoft.Data.SqlClient.SuppressInsecureTLSWarning";
         internal const string UseMinimumLoginTimeoutString = @"Switch.Microsoft.Data.SqlClient.UseOneSecFloorInTimeoutCalculationDuringLogin";
 
-        private static bool? s_legacyRowVersionNullBehavior;
-        private static bool? s_suppressInsecureTLSWarning;
-        private static bool s_makeReadAsyncBlocking;
-        private static bool s_useMinimumLoginTimeout;
+        // this field is accesses through reflection in tests and should be be renamed or have the type changed without refactoring NullRow related tests
```
jkotas, line 23:
```suggestion
        // this field is accesses through reflection in tests and should be renamed or have the type changed without refactoring NullRow related tests
```

Wraith2: fixed.

jkotas (approved):

PR #2230, opened 2023-11-19
Add | Adding net8 support

JRahnama:
Replacing Net7 with Net8.0


H-Yeo:
In CertificateTest.cs, do we want to clean up this comment as well?
// Net7.0 the new API added by dotnet runtime will check SANS and then SubjectName
to maybe
//  >=Net7.0 the new API added by dotnet runtime will check SANS and then SubjectName

H-Yeo:
Everything else looks good to me.

JRahnama:
> In CertificateTest.cs, do we want to clean up this comment as well? // Net7.0 the new API added by dotnet runtime will check SANS and then SubjectName to maybe // >=Net7.0 the new API added by dotnet runtime will check SANS and then SubjectName

Basically the API was added in net7, so it could stay the same or I can adjust it to net7 and greater version.

tools/specs/add-ons/Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider.nuspec
```diff
@@ -31,6 +31,9 @@ Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyStoreProvider.SqlColumnEncrypti
         <dependency id="Azure.Security.KeyVault.Keys" version="[4.4.0,5.0.0)" />
         <dependency id="Microsoft.Extensions.Caching.Memory" version="6.0.1" />
       </group>
+      <group targetFramework="net8.0">
+        <dependency id="Microsoft.Data.SqlClient" version="3.0.0" />
+        <dependency id="Azure.Core" version="[1.35.0,2.0.0)" />
       <group targetFramework="net7.0">
```
campersau, line 37: `</group>` missing after resolving merge conflict.

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -1443,9 +1443,13 @@ public sealed partial class SqlException : System.Data.Common.DbException
         public override string Source { get { throw null; } }
         /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlException.xml' path='docs/members[@name="SqlException"]/State/*'/>
         public byte State { get { throw null; } }
+
         /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlException.xml' path='docs/members[@name="SqlException"]/GetObjectData/*'/>
 #if !NET6_0_OR_GREATER
         [System.Security.Permissions.SecurityPermissionAttribute(System.Security.Permissions.SecurityAction.LinkDemand, Flags = System.Security.Permissions.SecurityPermissionFlag.SerializationFormatter)]
+#endif
+#if NET8_0_OR_GREATER
+        [System.Obsolete]
```
DavoudEshtehari, line 1452: Based on [.net doc](https://learn.microsoft.com/en-us/dotnet/api/system.exception.getobjectdata?view=net-8.0&devlangs=csharp&f1url=%3FappId%3DDev16IDEF1%26l%3DEN-US%26k%3Dk(System.Exception.GetObjectData)%3Bk(SolutionItemsProject)%3Bk(DevLang-csharp)%26rd%3Dtrue) suggestion.

Should this be applied on API docs as well?
cc @David-Engel

```suggestion
        [Obsolete("This API supports obsolete formatter-based serialization. It should not be called or extended by application code.", DiagnosticId="SYSLIB0051", UrlFormat="https://aka.ms/dotnet-warnings/{0}")]
```

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlException.cs
```diff
@@ -60,6 +62,9 @@ private SqlException(SerializationInfo si, StreamingContext sc) : base(si, sc)
         }
 
         /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlException.xml' path='docs/members[@name="SqlException"]/GetObjectData/*' />
+#if NET8_0_OR_GREATER
+        [Obsolete]
```
DavoudEshtehari, line 66:
```suggestion
        [Obsolete("This API supports obsolete formatter-based serialization. It should not be called or extended by application code.", DiagnosticId="SYSLIB0051", UrlFormat="https://aka.ms/dotnet-warnings/{0}")]
```

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/Batch/BatchTests.cs
```diff
@@ -55,21 +55,19 @@ public static void ConnectionCanCreateBatch()
...
-                Assert.NotNull(parameter);
-                parameter.ParameterName = "@p";
-                parameter.Value = 1;
-                batchCommand.Parameters.Add(parameter);
-
-                batch.ExecuteNonQuery();
-
-            }
+            using var connection = new SqlConnection(DataTestUtility.TCPConnectionString);
+            connection.Open();
```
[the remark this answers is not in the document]
JRahnama: I just addressed the failures on net8 and did not check the rest. 
Basically in my opinion this fix should have had a separate PR to address the test fix, but since it just showed up with net8 changes I included the fix in this PR.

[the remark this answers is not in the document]
David-Engel: > Just a comment here: In general, it appears this class uses curly braces { ... } for large code blocks, with the exception of nested usings. I'm indifferent on whether you change this or not.

@H-Yeo 
There is a lot of existing code with that style. When C# 8 eliminated the braces requirement, we started allowing it in code and using it more heavily in test code. There's no need to keep new code consistent with old code in this respect and contributors are free to convert old style to new as they see fit. In non-test code, we are more thoughtful about the using scope so we tend to keep the braces for clarity, but they aren't strictly required.

DavoudEshtehari (commented):
LGTM, waiting for the feedback from the API doc process.

robertmclaws:
Hopefully with the new downstream vulnerabilities that have been issues, this package will get updated soon.

Moving forward, this package should contain multiple builds targeting the currently supported runtimes, so that people installing this package in their apps can get the latest release for the .NET runtime they are targeting. I was SHOCKED to see one of my NuGet packages was resolving references all the way back down to .NET 6, even though I was targeting .NET 8, because of this package.

Thanks for your swift attention to getting this resolved!

DavoudEshtehari (approved):
🎉

Issue #1836, opened 2022-11-12
The certificate chain was issued by an authority that is not trusted.


Vikas8881:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

Today I update the project from core 6 to 7. and after this when i try to execute any API then I get this error 

**Microsoft.Data.SqlClient.SqlException (0x80131904): A connection was successfully established with the server, but then an error occurred during the login process. (provider: SSL Provider, error: 0 - The certificate chain was issued by an authority that is not trusted.)
 ---> System.ComponentModel.Win32Exception (0x80090325): The certificate chain was issued by an authority that is not trusted.**

But in core 6 it's working fine. 

### .NET Version

7

lcheunglci:
Thanks @ErikEJ ,

When upgrading from EF Core 6 to EF Core 7, the dependency of `Microsoft.Data.SqlClient` went from `2.1.2` to `5.0.1` and the Encrypt default value went from `false` to `true` i.e. `mandatory`, so you'll either need to set it to set it explicitly to `Encrypt=false` or `TrustServerCertificate=true` or `HostNameInCertificate=<CN Subject of your certificate>`.

https://learn.microsoft.com/en-us/ef/core/what-is-new/ef-core-7.0/breaking-changes#encrypt-true

Vikas8881:
> Seems like a SQL issue? What database are you connecting to?

I use MS SQL Server
SQL Server Management Studio						15.0.18424.0
SQL Server Management Objects (SMO)						16.100.47021.0+7eef34a564af48c5b0cf0d617a65fd77f06c3eb1
Microsoft Analysis Services Client Tools						15.0.19750.0
Microsoft Data Access Components (MDAC)						10.0.22621.819
Microsoft MSXML						3.0 6.0 
Microsoft .NET Framework						4.0.30319.42000
Operating System						10.0.22621

Vikas8881:
Thank you Everyone's Issue was Resolved By adding **TrustServerCertificate=True;** in the Connection String in AppSettings
and its work.

yogapratama12428:
it work for me, thanks @Vikas8881

SrkPrasadChangala:
> Thank you Everyone's Issue was Resolved By adding **TrustServerCertificate=True;** in the Connection String in AppSettings and its work.

It worked for me.

saman-ar:
Adding TrustServerCertificate=True to the connection string fixed this issue in entityframework 8.

[...]

Issue #1394, opened 2021-11-19
A connection was successfully established with the server, but then an error occurred during the login process.


andrew-vdb:
### Describe the bug
Unable to open trusted connection

```
A connection was successfully established with the server, but then an error occurred during the login process. (provider: SSL Provider, error: 0 - The certificate chain was issued by an authority that is not trusted.)'

```
![image](https://user-images.githubusercontent.com/12871379/142599460-3d37dd55-2ec7-4ba9-88da-dbec8c71a0d5.png)

### To reproduce
```c#
await connection.OpenAsync();
```

### Expected behavior
It should connect

### Further technical details
Microsoft.Data.SqlClient version: 4.0
.NET target: .net 6
SQL Server version: 
![image](https://user-images.githubusercontent.com/12871379/142599814-516a62ad-f2df-4077-a0d2-dbb43e062e11.png)

Operating system: Win 10

**Workaround**
Use System.Data.SqlClient, it works.... or use Microsoft.Data.SqlClient 3.0.1

cheenamalhotra:
[”Encrypt” is now true by default](https://github.com/dotnet/SqlClient/blob/main/release-notes/4.0/4.0.0.md#encrypt-default-value-set-to-true)

andrew-vdb:
Add 
Encrypt=false
to connection string will work too

RubberChickenParadise:
To anyone finding this, in my case it was due to the self signed server certificate. see https://github.com/dotnet/SqlClient/pull/1210#issuecomment-901568015:  

> Systems using SQL Server's self-generated certificate will fail after this change and the user will have to buy a certificate or create a self-signed certificate, or change their connection string. If the change was made for scenarios where the user could not modify the connection string to enable encryption, then they are left having to get new certificates.

The fixes are:
1) disable encryption on the client connection string with `Encrypt=false`
2) blindly trust the server certificate with `TrustServerCertificate=True`
3) use an external certificate on the sql server that is signed by a certificate authority in the clients trusted certificate authorities.
3b) if the sql server already has an externally signed certificate, trust the signing certificate authority on the client.

MeteAnn:
![Ekran görüntüsü 2024-06-26 123826](https://github.com/dotnet/SqlClient/assets/98713618/1addbe6f-313b-4604-a9b4-0a58456c4cd0)


I was experiencing the same issue, and it was resolved after adding Encrypt=false;.

PR #2247, opened 2023-11-28
POC: Remove Azure Identity from Microsoft.Data.SqlClient 

ErikEJ:
(to demonstrate where dependencies are used and test a build)

relates to #1108

Following linked files were removed from the .csproj:

```xml
    <Compile Include="..\..\src\Microsoft\Data\SqlClient\ActiveDirectoryAuthenticationProvider.cs">
      <Link>Microsoft\Data\SqlClient\ActiveDirectoryAuthenticationProvider.cs</Link>
    </Compile>
    <Compile Include="..\..\src\Microsoft\Data\SqlClient\ActiveDirectoryAuthenticationTimeoutRetryHelper.cs">
      <Link>Microsoft\Data\SqlClient\ActiveDirectoryAuthenticationTimeoutRetryHelper.cs</Link>
    </Compile>
 <Compile Include="..\..\src\Microsoft\Data\SqlClient\AzureAttestationBasedEnclaveProvider.cs">
      <Link>Microsoft\Data\SqlClient\AzureAttestationBasedEnclaveProvider.cs</Link>
    </Compile>

```

### DIFF
--- a/src/Microsoft.Data.SqlClient.sln
+++ b/src/Microsoft.Data.SqlClient.sln
@@ -200,6 +200,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{4F3CD363-B1E
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.SqlServer.Server", "Microsoft.SqlServer.Server\Microsoft.SqlServer.Server.csproj", "{A314812A-7820-4565-A2A8-ABBE391C11E4}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Data.SqlClient.Core", "Microsoft.Data.SqlClient\netcore\src\Microsoft.Data.SqlClient.Core.csproj", "{638A00EA-9648-45F1-B2FF-94E9A111AD73}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
@@ -472,6 +474,18 @@ Global
 		{A314812A-7820-4565-A2A8-ABBE391C11E4}.Release|x64.Build.0 = Release|Any CPU
 		{A314812A-7820-4565-A2A8-ABBE391C11E4}.Release|x86.ActiveCfg = Release|Any CPU
 		{A314812A-7820-4565-A2A8-ABBE391C11E4}.Release|x86.Build.0 = Release|Any CPU
+		{638A00EA-9648-45F1-B2FF-94E9A111AD73}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{638A00EA-9648-45F1-B2FF-94E9A111AD73}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{638A00EA-9648-45F1-B2FF-94E9A111AD73}.Debug|x64.ActiveCfg = Debug|x64
+		{638A00EA-9648-45F1-B2FF-94E9A111AD73}.Debug|x64.Build.0 = Debug|x64
+		{638A00EA-9648-45F1-B2FF-94E9A111AD73}.Debug|x86.ActiveCfg = Debug|x86
+		{638A00EA-9648-45F1-B2FF-94E9A111AD73}.Debug|x86.Build.0 = Debug|x86
+		{638A00EA-9648-45F1-B2FF-94E9A111AD73}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{638A00EA-9648-45F1-B2FF-94E9A111AD73}.Release|Any CPU.Build.0 = Release|Any CPU
+		{638A00EA-9648-45F1-B2FF-94E9A111AD73}.Release|x64.ActiveCfg = Release|x64
+		{638A00EA-9648-45F1-B2FF-94E9A111AD73}.Release|x64.Build.0 = Release|x64
+		{638A00EA-9648-45F1-B2FF-94E9A111AD73}.Release|x86.ActiveCfg = Release|x86
+		{638A00EA-9648-45F1-B2FF-94E9A111AD73}.Release|x86.Build.0 = Release|x86
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
@@ -507,6 +521,7 @@ Global
 		{B93A3149-67E8-491E-A1E5-19D65F9D9E98} = {0CC4817A-12F3-4357-912C-09315FAAD008}
 		{599A336B-2A5F-473D-8442-1223ED37C93E} = {0CC4817A-12F3-4357-912C-09315FAAD008}
 		{A314812A-7820-4565-A2A8-ABBE391C11E4} = {4F3CD363-B1E6-4D6D-9466-97D78A56BE45}
+		{638A00EA-9648-45F1-B2FF-94E9A111AD73} = {28E5EFE6-C9DD-4FF9-9FEC-532F72DFFA6E}
 	EndGlobalSection
 	GlobalSection(ExtensibilityGlobals) = postSolution
 		SolutionGuid = {01D48116-37A2-4D33-B9EC-94793C702431}
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlAuthenticationProviderManager.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlAuthenticationProviderManager.cs
@@ -20,6 +20,7 @@ namespace Microsoft.Data.SqlClient
         private const string ActiveDirectoryManagedIdentity = "active directory managed identity";
         private const string ActiveDirectoryMSI = "active directory msi";
         private const string ActiveDirectoryDefault = "active directory default";
+        private const string ActiveDirectoryWorkloadIdentity = "active directory workload identity";
 
         private readonly IReadOnlyCollection<SqlAuthenticationMethod> _authenticationsWithAppSpecifiedProvider;
         private readonly ConcurrentDictionary<SqlAuthenticationMethod, SqlAuthenticationProvider> _providers;
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlAuthenticationProviderManager.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlAuthenticationProviderManager.cs
@@ -20,7 +20,6 @@ namespace Microsoft.Data.SqlClient
         private const string ActiveDirectoryManagedIdentity = "active directory managed identity";
         private const string ActiveDirectoryMSI = "active directory msi";
         private const string ActiveDirectoryDefault = "active directory default";
-        private const string ActiveDirectoryWorkloadIdentity = "active directory workload identity";
 
         private readonly IReadOnlyCollection<SqlAuthenticationMethod> _authenticationsWithAppSpecifiedProvider;
         private readonly ConcurrentDictionary<SqlAuthenticationMethod, SqlAuthenticationProvider> _providers;
@@ -37,6 +36,7 @@ namespace Microsoft.Data.SqlClient
         {
             if (instance != null)
             {
+#if !CORE
                 var activeDirectoryAuthProvider = new ActiveDirectoryAuthenticationProvider(instance._applicationClientId);
                 instance.SetProvider(SqlAuthenticationMethod.ActiveDirectoryIntegrated, activeDirectoryAuthProvider);
                 instance.SetProvider(SqlAuthenticationMethod.ActiveDirectoryPassword, activeDirectoryAuthProvider);
@@ -46,7 +46,7 @@ namespace Microsoft.Data.SqlClient
                 instance.SetProvider(SqlAuthenticationMethod.ActiveDirectoryManagedIdentity, activeDirectoryAuthProvider);
                 instance.SetProvider(SqlAuthenticationMethod.ActiveDirectoryMSI, activeDirectoryAuthProvider);
                 instance.SetProvider(SqlAuthenticationMethod.ActiveDirectoryDefault, activeDirectoryAuthProvider);
-                instance.SetProvider(SqlAuthenticationMethod.ActiveDirectoryWorkloadIdentity, activeDirectoryAuthProvider);
+#endif
             }
         }
         /// <summary>
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
@@ -2344,6 +2344,7 @@ namespace Microsoft.Data.SqlClient
                             }
                             break;
                         }
+#if !CORE
                     case TdsEnums.SQLFEDAUTHINFO:
                         {
                             _connHandler._federatedAuthenticationInfoReceived = true;
@@ -2356,6 +2357,7 @@ namespace Microsoft.Data.SqlClient
                             _connHandler.OnFedAuthInfo(info);
                             break;
                         }
+#endif
                     case TdsEnums.SQLSESSIONSTATE:
                         {
                             if (!TryProcessSessionState(stateObj, tokenLength, _connHandler._currentSessionData))
--- a/src/Microsoft.Data.SqlClient/netcore/src/.editorconfig
+++ b/src/Microsoft.Data.SqlClient/netcore/src/.editorconfig
@@ -16,3 +16,8 @@ dotnet_diagnostic.SYSLIB0039.severity = suggestion
 
 # CA1310: Specify StringComparison for correctness
 dotnet_diagnostic.CA1310.severity = error
+
+# These are disabled for POC purposes only
+dotnet_diagnostic.CS0169.severity = none
+dotnet_diagnostic.CS0414.severity = none
+dotnet_diagnostic.CS0649.severity = none
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.Core.csproj
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.Core.csproj
@@ -1,0 +1,976 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <AssemblyName>Microsoft.Data.SqlClient.Core</AssemblyName>
+    <TargetFrameworks>net7.0;net6.0</TargetFrameworks>
+    <TargetFrameworks Condition="$(ReferenceType)=='NetStandard' AND $(TargetNetStandardVersion)=='netstandard2.1'">netstandard2.1</TargetFrameworks>
+    <GeneratePlatformNotSupportedAssemblyMessage Condition="'$(OSGroup)' == 'AnyOS'">Microsoft.Data.SqlClient is not supported on this platform.</GeneratePlatformNotSupportedAssemblyMessage>
+    <OSGroup Condition="'$(OSGroup)' == ''">$(OS)</OSGroup>
+    <TargetsWindows Condition="'$(OSGroup)'=='Windows_NT'">true</TargetsWindows>
+    <TargetsUnix Condition="'$(OSGroup)'=='Unix'">true</TargetsUnix>
+    <!-- Allow explicit addition of the Compile files instead of all project files to be included by Default -->
+    <EnableDefaultCompileItems>false</EnableDefaultCompileItems>
+    <TargetGroup Condition="'$([MSBuild]::GetTargetFrameworkIdentifier($(TargetFramework)))'=='.NETCoreApp'">netcoreapp</TargetGroup>
+    <TargetGroup Condition="'$([MSBuild]::GetTargetFrameworkIdentifier($(TargetFramework)))'=='.NETStandard'">netstandard</TargetGroup>
+    <Configurations>Debug;Release;</Configurations>
+    <Platforms>AnyCPU;x64;x86</Platforms>
+    <IntermediateOutputPath>$(ObjFolder)$(Configuration).$(Platform)\$(AssemblyName)\netcore\</IntermediateOutputPath>
+    <OutputPath>$(BinFolder)$(Configuration).$(Platform)\$(AssemblyName)\netcore\</OutputPath>
+    <DocumentationFile>$(OutputPath)\$(TargetFramework)\Microsoft.Data.SqlClient.xml</DocumentationFile>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <Product>Core $(BaseProduct)</Product>
+    <EnableTrimAnalyzer Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net6.0'))">true</EnableTrimAnalyzer>
+    <NoWarn>$(NoWarn);IL2026;IL2057;IL2072;IL2075</NoWarn>
+  </PropertyGroup>
+  <PropertyGroup>
+    <DefineConstants>CORE</DefineConstants>
+  </PropertyGroup>
+  <PropertyGroup>
+    <DebugType>portable</DebugType>
+    <DebugSymbols>true</DebugSymbols>
+  </PropertyGroup>
+  <!-- Contains common items shared between NetFx and NetCore -->
+  <ItemGroup Condition="'$(OSGroup)' != 'AnyOS'">
+    <Compile Include="..\..\src\Microsoft\Data\Common\ActivityCorrelator.cs">
+      <Link>Microsoft\Data\Common\ActivityCorrelator.cs</Link>
+    </Compile>
+    <Compile Include="..\..\src\Microsoft\Data\Common\AdapterUtil.cs">
+      <Link>Microsoft\Data\Common\AdapterUtil.cs</Link>
+    </Compile>
+    <Compile Include="..\..\src\Microsoft\Data\Common\DbConnectionOptions.Common.cs">
+      <Link>Microsoft\Data\Common\DbConnectionOptions.Common.cs</Link>
+    </Compile>
+    <Compile Include="..\..\src\Microsoft\Data\Common\DbConnectionPoolKey.cs">
+      <Link>Microsoft\Data\Common\DbConnectionPoolKey.cs</Link>
+    </Compile>
+    <Compile Include="..\..\src\Microsoft\Data\Common\DbConnectionStringCommon.cs">
+      <Link>Microsoft\Data\Common\DbConnectionStringCommon.cs</Link>
+    </Compile>
+    <Compile Include="..\..\src\Microsoft\Data\Common\MultipartIdentifier.cs">
+      <Link>Microsoft\Data\Common\MultipartIdentifier.cs</Link>
+    </Compile>
+    <Compile Include="..\..\src\Microsoft\Data\Common\NameValuePair.cs">
+      <Link>Microsoft\Data\Common\NameValuePair.cs</Link>
+    </Compile>
+    <Compile Include="..\..\src\Microsoft\Data\DataException.cs">
+      <Link>Microsoft\Data\DataException.cs</Link>
+    </Compile>
+    <Compile Include="..\..\src\Microsoft\Data\OperationAbortedException.cs">
+      <Link>Microsoft\Data\OperationAbortedException.cs</Link>
+    </Compile>
+    <Compile Include="..\..\src\Microsoft\Data\ProviderBase\DbConnectionPoolAuthenticationContext.cs">
+      <Link>Microsoft\Data\ProviderBase\DbConnectionPoolAuthenticationContext.cs</Link>
+    </Compile>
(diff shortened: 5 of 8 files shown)


src/Microsoft.Data.SqlClient.sln
```diff
@@ -206,6 +200,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{4F3CD363-B1E
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.SqlServer.Server", "Microsoft.SqlServer.Server\Microsoft.SqlServer.Server.csproj", "{A314812A-7820-4565-A2A8-ABBE391C11E4}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Data.SqlClient.Core", "Microsoft.Data.SqlClient\netcore\src\Microsoft.Data.SqlClient.Core.csproj", "{638A00EA-9648-45F1-B2FF-94E9A111AD73}"
```
AraHaan, line 203: I think a better name for the "Core" package is ``Microsoft.Data.SqlClient`` and instead document that Azure people must now reference ``Microsoft.Data.SqlClient.Azure`` for the Azure bits which would reference the "Core" package for them.

SimonCropp:
doesnt this bifurcate the ecosystem? eg if i want to write a lib that extends/uses sql, i now need two packages, one that targets each sql package

src/Microsoft.Data.SqlClient.sln
roji: Agreed that additional Azure functionality would be (far) better integrated via an additional package with Azure in its name. The problem with that, of course, is that it would be a breaking change for existing users; but it may be acceptable given that the mitigation is extremely trivial.

ErikEJ: I also agree, but that would be a huge breaking change. I also just did this to explore where the Azure dependencies are and maybe enable adding an additional package over time

AraHaan:
> doesnt this bifurcate the ecosystem? eg if i want to write a lib that extends/uses sql, i now need two packages, one that targets each sql package

Not at all, the Azure version of the package *could* depend on the base version, and the extension libs could use the base version like how it does now, just with user versions not including the azure libs. This way it all acts similar to how creating extendable/extension libs using ``Microsoft.Extensions.DependencyInjection`` is currently.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlAuthenticationProviderManager.cs
```diff
@@ -36,6 +36,7 @@ private static void SetDefaultAuthProviders(SqlAuthenticationProviderManager ins
         {
             if (instance != null)
             {
+#if !CORE
```
AraHaan, line 39: After looking further, can't this file be excluded from "Core" builds and only include this file on the Azure version of the library, then it could reference the "Core" version directly without breaking any functionality at all?

ErikEJ: I could try!

ErikEJ: Also, this is not a PR, just a demo POC. Needs design!

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -2312,6 +2312,7 @@ internal bool TryRun(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataRead
                             }
                             break;
                         }
+#if !CORE
```
AraHaan, line 2315: I feel like a generic type in the "Core" assembly that the Azure assembly could initialize values on (default null without the application "initializing" them to different values from the Azure assembly version) would be better in these internal types (``TdsParser``, ``SqlInternalConnectionTds``, ``ADP`` (AdapterUtil.cs).
The new generic type also being internal as well so only the azure assembly could use it as well.

For the case of ``TdaParer``, it can store the member information needed to ``Invoke`` the code for that case (with some null checking first to default values of said internal type to null in cases where the azure assembly is not used directly by the application. ``CreateSqlException`` can also fall under this as well, same for the other file where an enum is used to call into Active Directory code (``EnclaveDelegate`` in ``EnclaveDelegate.Crypto.cs``.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlAuthenticationProviderManager.cs
AraHaan: Yeah, I know, just offering up some ideas that could help contribute to a design.

SimonCropp:
@AraHaan 

> Not at all, the Azure version of the package could depend on the base version, and the extension libs could use the base version like how it does now, just with user versions not including the azure libs. This way it all acts similar to how creating extendable/extension libs using Microsoft.Extensions.DependencyInjection is currently.

but "creating an Azure version of the package " is not what is being proposed here. this is slicing off a new Microsoft.Data.SqlClient.Core package. 

so it means, to avoid the azure bits, the stack of libraries that currently use/extend Microsoft.Data.SqlClient would need to move to Microsoft.Data.SqlClient.Core. for example  Microsoft.EntityFrameworkCore.SqlServer. and that would be considered a breaking change for each of those libs, so would likely have a very long tail until top level apps could remove Microsoft.Data.SqlClient

while breaking out a Microsoft.Data.SqlClient.Azure  package would be a breaking change for this project, it does seem like that would have less combine fallout for consumers

unless u misunderstand what is being proposed here

ErikEJ:
> while breaking out a Microsoft.Data.SqlClient.Azure package would be a breaking change for this project, it does seem like that would have less combine fallout for consumers

Except it would be breaking for anyone using Azure auth today...

SimonCropp:
@ErikEJ 

> Except it would be breaking for anyone using Azure auth today...

yeah. no idea which group is more common

ErikEJ:
OK - once 5.2 has been released, I will try this approach (initially .NET only)

`Microsoft.Data.SqlClient.Azure` = current package, renamed (new .csproj with all current dependencies)

`Microsoft.Data.SqlClient` = new package without Azure auth dependencies

This will require anyone using Azure auth methods to use an explict reference to:

`<PackageReference Include="Microsoft.Data.SqlClient.Azure" Version="6.0.0" />`

Attempts to use Azure auth with Microsoft.Data.SqlClient should result in an actionable error message with links to a suitable docs page

Thoughts? @SimonCropp @roji @AraHaan @David-Engel

AraHaan:
> > while breaking out a Microsoft.Data.SqlClient.Azure package would be a breaking change for this project, it does seem like that would have less combine fallout for consumers
> 
> Except it would be breaking for anyone using Azure auth today...

Yeah, except they can unbreak themselves by appending ``.Azure`` to the package name and calling it a day by releasing a new patch release if they do in-fact use the azure functionality.

AraHaan:
> OK - once 5.2 has been released, I will try this approach (initially .NET only)
> 
> `Microsoft.Data.SqlClient.Azure` = current package, renamed (new .csproj with all current dependencies)
> 
> `Microsoft.Data.SqlClient` = new package without Azure auth dependencies
> 
> This will require anyone using Azure auth methods to use an explict reference to:
> 
> `<PackageReference Include="Microsoft.Data.SqlClient.Azure" Version="6.0.0" />`
> 
> Attempts to use Azure auth with Microsoft.Data.SqlClient should result in an actionable error message with links to a suitable docs page
> 
> Thoughts? @SimonCropp @roji @AraHaan @David-Engel

Sounds good, yeah.

David-Engel:
> OK - once 5.2 has been released, I will try this approach (initially .NET only)
> 
> `Microsoft.Data.SqlClient.Azure` = current package, renamed (new .csproj with all current dependencies)
> 
> `Microsoft.Data.SqlClient` = new package without Azure auth dependencies
> 
> This will require anyone using Azure auth methods to use an explict reference to:
> 
> `<PackageReference Include="Microsoft.Data.SqlClient.Azure" Version="6.0.0" />`
> 
> Attempts to use Azure auth with Microsoft.Data.SqlClient should result in an actionable error message with links to a suitable docs page
> 
> Thoughts? @SimonCropp @roji Shay Rojansky FTE @AraHaan @David-Engel David Engel (SIMBA TECHNOLOGIES INC) Vendor

We should err on the side of compatibility. The reverse would be far less disruptive. People already expect M.D.SqlClient to support Azure connectivity. Those few who object to the current long dependency tail can move to a new M.D.SqlClient.Core package, which just has the base functionality. Base functionality in my mind is SQL auth, integrated Windows auth, and token auth (AccessToken, AccessTokenCallback).

SimonCropp:
@David-Engel 

> Those few who object to the current long dependency tail can move to a new M.D.SqlClient.Core package, which just has the base functionality

to achieve that all libs using Microsoft.Data.SqlClient would need to make a breaking change. 
i think breaking change decisions should be made based on total impact to the consumers of Microsoft.Data.SqlClient.
ie SqlClient usage is an ecosystem, so decisions on braking changes should take that ecosystem into account

SimonCropp:
@ErikEJ 
```
OK - once 5.2 has been released, I will try this approach (initially .NET only)

Microsoft.Data.SqlClient.Azure = current package, renamed (new .csproj with all current dependencies)

Microsoft.Data.SqlClient = new package without Azure auth dependencies

This will require anyone using Azure auth methods to use an explict reference to:

<PackageReference Include="Microsoft.Data.SqlClient.Azure" Version="6.0.0" />
```

Yes. i think that is the best compromise of all the options discussed

David-Engel:
> > Those few who object to the current long dependency tail can move to a new M.D.SqlClient.Core package, which just has the base functionality
> 
> to achieve that all libs using Microsoft.Data.SqlClient would need to make a breaking change. i think breaking change decisions should be made based on total impact to the consumers of Microsoft.Data.SqlClient. ie SqlClient usage is an ecosystem, so decisions on braking changes should take that ecosystem into account

@SimonCropp 
How would libs need to make a breaking change? I don't think we are thinking the same thing. My goal is not to have breaking changes for any existing users of Microsoft.Data.SqlClient. Libs referencing MDS would not have to change. MDS would have a dependency on MDS.Core. Only users who want the version without Azure support would need to change their dependency from MDS to MDS.Core.

AraHaan:
> > > Those few who object to the current long dependency tail can move to a new M.D.SqlClient.Core package, which just has the base functionality
> > 
> > 
> > to achieve that all libs using Microsoft.Data.SqlClient would need to make a breaking change. i think breaking change decisions should be made based on total impact to the consumers of Microsoft.Data.SqlClient. ie SqlClient usage is an ecosystem, so decisions on braking changes should take that ecosystem into account
> 
> @SimonCropp How would libs need to make a breaking change? I don't think we are thinking the same thing. My goal is not to have breaking changes for any existing users of Microsoft.Data.SqlClient. Libs referencing MDS would not have to change. MDS would have a dependency on MDS.Core. Only users who want the version without Azure support would need to change their dependency from MDS to MDS.Core.

Basically libs the once depended on the single package would be broken for a bit because they would need to then reference the "Core" package to then remove the azure bits, vs having all the core bits pushed in a new MAJOR version of the current package and split the azure bits to a separate version with ``.Azure`` appended to it's name. Either way those who depends on the SqlClient package will be broken anyways, except it limits the breakage to those who ACTUALLY want the azure bits.

David-Engel:
> Basically libs the once depended on the single package would be broken for a bit because they would need to then reference the "Core" package to then remove the azure bits, vs having all the core bits pushed in a new MAJOR version of the current package and split the azure bits to a separate version with .Azure appended to it's name. Either way those who depends on the SqlClient package will be broken anyways, except it limits the breakage to those who ACTUALLY want the azure bits.

This still isn't making sense to me. Why do we need to break anyone? In my scenario, those that currently depend on MDS (Azure or not) should require no change. MDS currently supports Azure and will continue to. The only ones that need to change should be those that really don't want Azure (the "feature" we are removing) because they want a slimmed down dependency tree.

Think of the existing MDS package as the "all in" package. For those who don't want everything, they can make changes.

> As such I feel like from the start the azure bits should have been a separate opt-in .Azure package anyways much like how a lot of things in MSBuild and even the runtime (dotnet/runtime) has opt-in escape hatches put in place.

I'm not going to disagree with that. But it doesn't matter. We are here now and we aren't going to make a breaking change for Azure users (or any other users) that depend on MDS.

SimonCropp:
@David-Engel 
lets take a specific example Microsoft.EntityFrameworkCore.SqlServer. IMO for them to move from Microsoft.Data.SqlClient to Microsoft.Data.SqlClient.Core, that would be a breaking change. and that change would need to be done in all similar libraries.

i think we should avoid that long tail of breaking changes. and just do on in this repo

also, to fully remove the azure bits, the fact that "change to different nuget" is significantly more works and more difficult o communicate to all consuming libs. when compared to "do a nuget update" which they are used to regularly doing

also the Microsoft.Data.SqlClient + Microsoft.Data.SqlClient.Azure approach means top level apps can remove the azure bits with only a single top level ref update.

where the Microsoft.Data.SqlClient + Microsoft.Data.SqlClient.Core approach requires all intermediate nugets to be updated to remove the azure bits

SimonCropp:
let me clarify that specific example

 * an project using azure references Microsoft.EntityFrameworkCore.SqlServer and not Microsoft.Data.SqlClient.
 * we do Microsoft.Data.SqlClient + Microsoft.Data.SqlClient.Core  in a "non breaking way"
 * to get the value out of "no azure bits" Microsoft.EntityFrameworkCore.SqlServer moves from Microsoft.Data.SqlClient to Microsoft.Data.SqlClient.Core 
 * project updates Microsoft.EntityFrameworkCore.SqlServer 
 * boom because no azure bits

now multiple over all the libs that extend Microsoft.Data.SqlClient. ie it might not break when they update Microsoft.EntityFrameworkCore.SqlServer. it might break when they update ErikEJ.SqlClient.Extensions, or  Serilog.Sinks.MSSqlServer, etc. but evernually they will update the last sql client lib that removes  Microsoft.Data.SqlClient (and hence azure) and their project will break

David-Engel:
I think I see the disconnect. You are trying to make it easier for Azure to be an opt-in for everyone, including those who currently only have it via a transitive dependency. But you are proposing breaking Azure for everyone all the way down the line in order to do it.

So in your scenario, everyone (EF, for example) would only depend on/get non-Azure functionality (and all their Azure users would be broken). To get Azure authentication, apps (likely at the tail end unless libs like EF want to offer a ".Azure" variant) would simply need to add a dependency on the .Azure package.

I think you underestimate the disruption this would cause to existing Azure users. Given that Microsoft owns this library, uses it heavily internally (think about all the code that must be behind MS/Azure products and services), heavily promotes MDS in their docs (that's a lot of updating docs - if version >= 6, use M.D.S.Azure), and has an interest in keeping it simple and easy to use against Azure, I can't see how this change would be welcomed.

AraHaan:
> let me clarify that specific example
> 
>     * an project using azure references Microsoft.EntityFrameworkCore.SqlServer and not Microsoft.Data.SqlClient.
> 
>     * we do Microsoft.Data.SqlClient + Microsoft.Data.SqlClient.Core  in a "non breaking way"
> 
>     * to get the value out of "no azure bits" Microsoft.EntityFrameworkCore.SqlServer moves from Microsoft.Data.SqlClient to Microsoft.Data.SqlClient.Core
> 
>     * project updates Microsoft.EntityFrameworkCore.SqlServer
> 
>     * boom because no azure bits
> 
> 
> now multiply that over all the libs that extend Microsoft.Data.SqlClient. ie it might not break when they update Microsoft.EntityFrameworkCore.SqlServer. it might break when they update ErikEJ.SqlClient.Extensions, or Serilog.Sinks.MSSqlServer, etc. but evernually they will update the last sql client lib that removes Microsoft.Data.SqlClient (and hence azure) and their project will break

Or do like the ``Microsoft.Extensions.DependencyInjection`` approach and allow someone to register the azure bits if they want them, as a ``service``.

AraHaan:
> @AraHaan
> 
> > Or do like the Microsoft.Extensions.DependencyInjection approach and allow someone to register the azure bits if they want them, as a service. After all Azure is in a nutshell a service after all.
> 
> r u proposing only a runtime api level change ?
> 
> if so, this cant be only a run time change. we already have one significant bug cause by sqlclient nuget pulling in azure refs [AzureAD/microsoft-authentication-library-for-dotnet#4468](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/4468) i know because it broke our integration environment
> 
> it has to be nuget change



> @AraHaan
> 
> > Or do like the Microsoft.Extensions.DependencyInjection approach and allow someone to register the azure bits if they want them, as a service. After all Azure is in a nutshell a service after all.
> 
> r u proposing only a runtime api level change ?
> 
> if so, this cant be only a run time change. we already have one significant bug cause by sqlclient nuget pulling in azure refs [AzureAD/microsoft-authentication-library-for-dotnet#4468](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/4468) i know because it broke our integration environment
> 
> it has to be nuget change

Basically yes, have the azure bits after they install the azure version of the library (which has a dependency on the base version) which:
- must be explicitly initialized by the application to "register it" as a "service" for the base version to obtain and use.

SimonCropp:
@AraHaan

> Basically yes, have the azure bits after they install the azure version of the library (which has a dependency on the base version) which:    must be explicitly initialized by the application to "register it" as a "service" for the base version to obtain and use.

ahh. yes. thats what i want

SimonCropp:
if we do go with the Microsoft.Data.SqlClient + Microsoft.Data.SqlClient.Core approach. the doco will be a fun read for everyone moving forward

> use Microsoft.Data.SqlClient if you are targeting Azure. otherwise use Microsoft.Data.SqlClient.Core 

and that when people even bother to read and understand that doco. because the entire internet history of blogposts, stackoverflow, and training will continue to be pointing to "Microsoft.Data.SqlClient is the one you want"

IMO the technical debt and significant number of people who will continue to pick the incorrect package should be reason enough to rule out that approach

SimonCropp:
@David-Engel 

> I think you underestimate the disruption this would cause to existing Azure users. 

I dont think i do. We release a new major of Microsoft.Data.SqlClient . and in the release notes we include "azure users please add a ref to this other nuget and add a single line to your app startup". i think that is a very small disruption in the scheme of breaking changes

> Given that Microsoft owns this library, uses it heavily internally (think about all the code that must be behind MS/Azure products and services), heavily promotes MDS in their docs (that's a lot of updating docs - if version >= 6, use M.D.S.Azure), and has an interest in keeping it simple and easy to use against Azure, I can't see how this change would be welcomed.

to me that reads like "azure consumer trump the needs of the larger community". in my mind this is not a "Microsoft owned library" but a "community OSS library with MS as the curator". at no point should the impact on the larger community (who have helped maintain this library over the years) be disregarded for the need of the Azure consumer

virzak:
Another consideration: how fast can I get rid of Azure bits? With `.Azure` approach, I can reference the latest MDS and all the Azure related .DLLs are gone from the publish directory. With `.Core` approach, the timing of losing azure bits depends on if and when a library like `Microsoft.EntityFrameworkCore.SqlServer` decides to migrate to .Core variation. That could take years.

Otherwise I agree with all @SimonCropp's points. There is no way of avoiding a breaking change with both approaches. When I do `dotnet outdated` and see a major version update I expect a breaking change and if one doesn't occur, I consider it a bonus.

virzak:
> I think you underestimate the disruption this would cause to existing Azure users. Given that Microsoft owns this library

With `.Azure` option a user adds an extra package reference is one of the simplest to deal with. Not ideal, but straightforward.

Compare this to a list of breaking changes for a typical major EF Core release, also owned by Microsoft. [EF Core 3.1](https://learn.microsoft.com/en-us/ef/core/what-is-new/ef-core-3.x/breaking-changes) took me days to resolve. With the recent EF Core 8 the [recommendation](https://twitter.com/ErikEJ/status/1734836632740614390) was to switch to daily builds.

ErikEJ:
Thanks for the interesting discussion, everyone. I think it is obviuos that if this is going to happen (and a PR is going to be accepted),the only way to do this is via MDS.Core. There is a too long history of just having Azure auth available.

Yes, it will take years for consumers to adapt to this, and some may not. So be it.

 Base functionality in .Core will be SQL auth, integrated Windows auth, and token auth (AccessToken, AccessTokenCallback).

I will close this draft, and maybe revisit af 5.2 has been released, as next planned version seems to be 6.0 now.

0xced:
> I think it is obvious that if this is going to happen, the only way to do this is via MDS.Core.

Why?

As you said in a previous comment, _attempts to use Azure auth with Microsoft.Data.SqlClient should result in an **actionable error message** with links to a suitable docs page_.

Wouldn't that be the best solution?

I remember that Microsoft.Data.SqlClient version 4 was released with [Encrypt=true by default](https://github.com/dotnet/SqlClient/blob/507e8dd64d8c6235f81c9c1864013e7ee3f3f100/CHANGELOG.md#preview-release-400-preview1212372---2021-08-25) and this [really](https://github.com/dotnet/SqlClient/issues/1836) [caused](https://github.com/dotnet/SqlClient/issues/1394) [havoc](https://stackoverflow.com/questions/17615260/the-certificate-chain-was-issued-by-an-authority-that-is-not-trusted-when-conn/70850834#70850834). It had a good error message:

> A connection was successfully established with the server, but then an error occurred during the login process. (provider: SSL Provider, error: 0 - The certificate chain was issued by an authority that is not trusted.)

But it did not have a **actionable** error message.

For this breaking change, having the error message mentions that adding a package reference to `Microsoft.Data.SqlClient.Azure` will solve the issue would be the smoothest path forward and the easiest for all the libraries depending on `Microsoft.Data.SqlClient` package.

ErikEJ:
> Why?

Because any change needs to be approved by the responsible maintainers of this repo...

SimonCropp:
@ErikEJ so all the maintainers have chimed in?

0xced:
>> Why?

> Because any change needs to be approved by the responsible maintainers of this repo…

The why was about _the **only way** to do this is via MDS.Core_

Only @David-Engel has argued about this route, all other participants in this thread are arguing towards `Microsoft.Data.SqlClient` + `Microsoft.Data.SqlClient.Azure`. And also the 49 upvotes on https://github.com/dotnet/SqlClient/issues/1108#issuecomment-861119074

But David is the lead for SQL Server drivers so I don't have high expectations on the weight of random internet people arguments. ¯\\\_(ツ)\_/¯

roji:
A few notes/thoughts/summary from me...

* I don't have the latest figures, but IIRC a very significant number of usages of SqlClient comes through EF Core.
* If we go with the SqlClient+SqlClient.Core, EF must react to the change for it to affect its users; otherwise Microsoft.EntityFramework.SqlServer continues to reference SqlClient, and there's no way to opt out of the Azure dependencies while using EF. So at this point there are two options:
    1. Do the same at the EF level and introduce a new M.EF.SqlServer.Core, for users who don't want Azure dependencies. This means that library dependencies on top of SqlClient - EF isn't the only one - all need to bifurcate and take on added maintenance/communication burden (we'd need to explain to our users what the difference between the package is etc.). I'd really like to avoid this.
    2. Switch M.EF.SqlServer to depend on the new SqlClient.Core package; users who want Azure also add SqlClient.Azure as described above. If this happens, EF Azure users are broken in the exact same way as going the other way in SqlClient and introducing SqlClient.Azure instead (as @SimonCropp argued above). In other words, the breaking change has just been pushed up.
* In other words, I'd argue that a more holistic view of the ecosystem is needed here; if SqlClient were a single library used directly, that would lead to a different set of considerations than the current one, where SqlClient is more of an underlying library typically consumed via other 3rd-party layers (such as EF). Either you ask all layers to make a change as well, or the breakage is pushed onto them.
* Another pretty obvious point but worth saying - with SqlClient.Core we end up in a place where package naming is simply misleading; SqlClient.Azure clearly expresses what that package is about (adding Azure-specific stuff), but just having SqlClient+SqlClient.Core makes it non-obvious for users to choose the right option.
* Thinking ahead a bit, introducing SqlClient.Core blocks that name to mean "SqlClient but without Azure", forever; if we need to do a similar exercise at some point, the name is already taken. I specifically have NativeAOT in mind; if SqlClient has any functionaliy by default that's incompatible with NativeAOT, introducing a SqlClient.Core is one main way to deal with that (this has happened in other places in the ecosystem). But if SqlClient.Core is already used to mean "but without Azure", we can't do that.
* Finally, I agree with above arguments that although this a breaking change, it's the "best kind" of breaking change - it's discoverable immediately (not e.g. a subtle behavioral change discovered only after moving to production), and we hopefully can include an informative exception pointing to the exact mitigation ("install SqlClient.Azure"). We have some experience with introducing such changes in EF, and if done right, the actual impact and user dissatisfaction has been minimal.

ErikEJ:
Thanks for joining, @roji - your opinion is valuable to me. So your are in favor of [my proposal above](https://github.com/dotnet/SqlClient/pull/2247#issuecomment-1931836865) ❤️ 

Let's stand down, try to not get too emotional and think about this one more time...

I will share my proposal and collect more feedback

SimonCropp:
while i am grateful that this is be re-considered. it is annoying that it only takes 1 MS employee to re-iterate, what several  community members have already stated, to be re-considered

there is a noticeable bias on .net OSS project to give priority to who says something over what is being said.

if a community member had made exactly the same comment as @roji would we have had the same outcome?

ErikEJ:
@AraHaan > add: Microsoft.Extensions.DependencyInjection.Abstractions

That will add a new trail of dependencies for .NET Framework, so...

AraHaan:
> @AraHaan > add: Microsoft.Extensions.DependencyInjection.Abstractions
> 
> That will add a new trail of dependencies for .NET Framework, so...

But won't it be ok for .NET Framework though, even if it is used as an implementation detail internally?

edcsu:
This is great although it will be a breaking change for me. Thanks for pointing this out

ErikEJ:
@AraHaan Making this change and adding new dependencies does not seem right

roji:
Everyone, just to be clear - I work at Microsoft but I don't make any decisions where SqlClient is concerned. @David-Engel is the owner here and has the final say, and I know he has SqlClient's best interests in mind, as well as the best knowledge about who uses SqlClient, how, and what breaking changes make sense in that context. Although I have my opinion, I fully trust @David-Engel will make the right choice here etc.

@dustinmoris your comment really does not reflect the reality of the situation; there's no "corporate politics" or an employee show-down with clout. There really are just two Microsoft here engineers, discussing what to do the best for SqlClient and for the .NET ecosystem as a whole. We regularly engage with the community on important design decisions, and very frequently change our minds based on community feedback that we get. It's very common for people on the outside to assume "corporate politics" and various dark "behind the scenes" dealings at Microsoft - I deal with this all the time - but we're really just a bunch of passionate engineers discussing and trying to arrive at the best outcome given the situation.

So again, I'm just voicing my opinion here, with the added weight that I have the EF perspective (and EF is an important factor, being a major way that SqlClient is used).

@AraHaan for the record, I'm against the direction of introducing DI into SqlClient. Unlike EF, SqlClient (and other ADO.NET providers) aren't currently based on DI concepts, and rewriting SqlClient to start using DI internally is a far-reaching change that doesn't really seem related to this discussion in any way. I do agree that better integration with DI is a good idea, but I'd do that in a separate package (much like how Npgsql has [Npgsql.DependencyInjection](https://www.nuget.org/packages/Npgsql.DependencyInjection) for people who want it). Enabling Azure functionality via an external package (regardless of how we call it) is most probably doable without introducing DI.

ErikEJ:
Published payload with .NET 8 target:

image

robertmclaws:
Make this change, increment the version to 6.0, and break it.

Why would anyone running on-prem SQL Server want all of the Azure bloat in their apps?

It's super annoying that the SqlClient dragged in all these dependencies, and honestly the Microsoft team should have thought of this at the beginning.

This is a great solution. Break it, fix it, move on.

David-Engel:
Thanks for all the input folks. I've just been stating my opinion based on my own experiences. It's good to hear different opinions. I'm not against having my mind changed. After reading all the comments, I'm inclined to agree that MDS + MDS.Azure seems to make more sense. (Implementation details TBD.) I'll have to discuss this with other internal stakeholders to see what their thoughts are. I don't want to make people do a bunch of work that won't ever be merged.

SimonCropp:
@David-Engel thanks so much for re-considering

AraHaan:
Anything new about this yet? It has been a few months since the last comment on this.

cheenamalhotra:
Closing PR in favor of https://github.com/dotnet/SqlClient/pull/2247#issuecomment-1934709067 - as it will require rework and new PR would be ideal here. We've picked up the work for upcoming milestones, so it will be worked upon by our team.

AraHaan:
@cheenamalhotra mind doing an issue for this and a feature branch so people who are interested can keep track with the progress and potentially submit PR's to help land a hand if needed?

MichaelKetting:
@cheenamalhotra , I'll second @AraHaan's request. Please post this here so we know which issue to watch. Can you also confirm that this will be delivered with .NET 10? For us, this is an extremly important issue because System.Data.SqlClient is no longer officially supported on .NET 10 and Microsoft.Data.SqlClient in its current shape is a significant problem due to the current dependency tree.

AraHaan:
I would recommend delaying the System.Data.SqlClient with not being supported in .NET 10 until it is proven that Microsoft.Data.SqlClient can do this for .NET 10 before it's release so it does not break people who have STRICT security standards on dependencies.

cheenamalhotra:
Would recommend tracking the same issue https://github.com/dotnet/SqlClient/issues/1108

PR #2279, opened 2023-12-19
Tests | Fix Tests regarding Concurrency both Sync and Async versions.

H-Yeo:
On Bug Ticket #25147, we are seeing intermittent failure on the ConcurrentExecution tests.

After analysis, we have determined that as the tests are running in the pipeline, we are witnessing thread starvation as the tests are running. In addition, there is also an edge case where there an exception before the retry mechanism is encountered due to connectivity issues with the database.

In order to catch all exceptions, some refactoring was performed for thread starvation, async/sync test separation, and counting of retry exception occurrences.


H-Yeo:
All PR Check's Pass in Draft

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/RetryLogic/SqlCommandReliabilityTest.cs
```diff
@@ -524,160 +526,164 @@ public async void RetryExecuteAsyncCancel(string cnnString, SqlRetryLogicBasePro
...
-                        cmd.CommandText = query + " FOR XML AUTO";
-                        Assert.Throws<AggregateException>(() => cmd.ExecuteXmlReader());
+                        cmd.CommandText = query;
+
+                        for (int j = 0; j < delayCount; j++)
+                        {
+                            int longQueryOperation = j * j;
+                        }
+
+                        Assert.Throws<Exception>(() => commandAction(cmd));
```
DavoudEshtehari, line 609: `AggregateException` is expected in all the tests. I don't see any reason to use the `Exception` instead of it and adding one level of assertion.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/RetryLogic/SqlCommandReliabilityTest.cs
```diff
@@ -524,160 +526,164 @@ public async void RetryExecuteAsyncCancel(string cnnString, SqlRetryLogicBasePro
         [MemberData(nameof(RetryLogicTestHelper.GetConnectionAndRetryStrategyInvalidCommand), parameters: new object[] { 2 }, MemberType = typeof(RetryLogicTestHelper), DisableDiscoveryEnumeration = true)]
         public void ConcurrentExecution(string cnnString, SqlRetryLogicBaseProvider provider)
         {
-            int numberOfTries = provider.RetryLogic.NumberOfTries;
             string query = "SELECT bad command";
-            int retriesCount = 0;
             int concurrentExecution = 3;
-            provider.Retrying += (s, e) => Interlocked.Increment(ref retriesCount);
+            ProcessDataInParallel(cnnString, provider, query, concurrentExecution, cmd => cmd.ExecuteScalar());
```
DavoudEshtehari, line 531: More tidier. Nice!

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/RetryLogic/SqlCommandReliabilityTest.cs
```diff
@@ -524,160 +526,164 @@ public async void RetryExecuteAsyncCancel(string cnnString, SqlRetryLogicBasePro
...
+                        using (SqlCommand cmd = cnn.CreateCommand())
+                        {
+                            await cnn.OpenAsync();
+                            cmd.RetryLogicProvider = provider;
+                            cmd.CommandText = query;
+
+                            for (int j = 0; j < delayCount; j++)
+                            {
+                                int longQueryOperation = j * j;
+                            }
```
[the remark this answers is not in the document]
H-Yeo: Please LMK if you want this removed, but this was added to reduce/test whether we were encountering Thread Starvation with multiple simultaneous connections to the database. Adding a long query was a potential to reduce connection errors.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/RetryLogic/SqlCommandReliabilityTest.cs
H-Yeo: This was a problem with Thread Starvation. The first exception being thrown is coming out of the connection (System.Exception), which is well before we attempt our actual testing of the Retry, which is the expected Aggregate.Exception.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/RetryLogic/SqlCommandReliabilityTest.cs
H-Yeo: Thanks!

H-Yeo:
Hi @David-Engel , Would you have able to perform a PR on this when you have a moment. It was revolving around the Thread Starvation issue around connections. LMK if you have any questions. Thanks!

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/RetryLogic/SqlCommandReliabilityTest.cs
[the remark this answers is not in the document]
DavoudEshtehari: Making the CPU busy is not a safe way to add delay and you can use `Thread.Sleep` for a more stable delay. I don't suggest adding delay here to make it more possible to be in conflict for the underlaying feature implementation. If the connection status is your concern here you can take it out and pass it after making the connection to the server.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/RetryLogic/SqlCommandReliabilityTest.cs
arellegue: Just wondering if you have tried without the for-loop delay in the pipeline. I feel this may cause a high-CPU usage. Have you tried Thread.Sleep()?

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/RetryLogic/SqlCommandReliabilityTest.cs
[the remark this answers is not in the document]
David-Engel: I don't understand this block. It looks like it is just using a thread to perform a bunch of CPU operations here to make the async method execute a bit longer. How is this helping identify thread starvation?

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/RetryLogic/SqlCommandReliabilityTest.cs
```diff
@@ -524,160 +526,164 @@ public async void RetryExecuteAsyncCancel(string cnnString, SqlRetryLogicBasePro
...
+        private static void ProcessDataInParallel(string cnnString, SqlRetryLogicBaseProvider provider,
+                                                  string query, int concurrentExecution,
+                                                  Action<SqlCommand> commandAction)
+        {
+            int numberOfTries = provider.RetryLogic.NumberOfTries;
+            int delayCount = 10_000;
+            int retriesCount = 0;
+            var exceptions = new ConcurrentQueue<Exception>();
+
+            ThreadPool.SetMinThreads(concurrentExecution * 2, concurrentExecution * 2);
```
David-Engel, line 590: I don't think this is the best place to SetMinThreads. SetMinThreads affects the entire process and we need to return it to its previous value to avoid affecting subsequent tests. I'd put SetMinThreads in a constructor for the test class, make the class IDisposable, and set min threads back in the Dispose() method.

Same for line 641.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/RetryLogic/SqlCommandReliabilityTest.cs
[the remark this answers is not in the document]
H-Yeo: These long running query operations will be removed, as you're correct, they don't do anything to address the underlying problem.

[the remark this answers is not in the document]
H-Yeo: Adjusting the flow of Exception handling as recommend by Davoud.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/RetryLogic/SqlCommandReliabilityTest.cs
H-Yeo: Adjusting the exception handling as mentioned earlier. Will test shortly via Pipeline.

David-Engel (approved):

DavoudEshtehari (approved):

Issue #834, opened 2020-12-02
Improve Exception Messages


israelaece:
Note in the follow snippet that there isn't a column index 4 in the query. When we run the code, we got a `IndexOutOfRangeException`. But when the query and object initializer is more complex, is difficult to locate the problem.

My suggestion here is to include in exception message what's the column index that not exists.

`Unhandled Exception: System.IndexOutOfRangeException: Index was outside the bounds of the array.
   at Microsoft.Data.SqlClient.SqlDataReader.CheckDataIsReady(...)`

            using (var conn = new SqlConnection("Data Source=.;Initial Catalog=Test;Integrated Security=True"))
            {
                using (var cmd = new SqlCommand("SELECT Id, Name, Created, Token FROM People", conn))
                {
                    conn.Open();

                    using (var dr = cmd.ExecuteReader())
                    {
                        while (dr.Read())
                        {
                            people.Add(new Person()
                            {
                                Id = dr.GetInt32(0),
                                Name = dr.GetString(1),
                                Created = dr.GetDateTime(2),
                                Token = dr.GetGuid(3),
                                Dummy = dr.GetString(4)
                            });
                        }
                    }
                }
            }

The same approach could be used when the problem is casting `InvalidCastException`:

`Unhandled Exception: System.InvalidCastException: Unable to cast object of type 'System.Int32' to type 'System.String'.
   at Microsoft.Data.SqlClient.SqlBuffer.get_String() `

JRahnama:
@israelaece in .NETCore 3.1 you can apply the column name. for example in your code:
`Name = dr.GetString("FirstName") `, if the column name is FirstName, will get you the column value and if does not exists it will state: `Unhandled exception. System.IndexOutOfRangeException: FirstName`.

in .NETFramework or older versions of .NETCore this is not applicable. However, you can try:
`Name = dr.GetString(dr.GetOrdinal("FitstName"))` and if that does not exists you will get an error message stating the column name.

JRahnama:
For the `InvalidCastException`  the exception does states the problematic casting. Do you need column name or index number to be exactly sated on the Exception?

israelaece:
Hi @JRahnama. Yes, but I'm looking for performance, and AFAIK, `GetOrdinal` has an overhead associated to lookup names in the hashtable. `GetOrdinal` allow to use name to find the index, and I want to work in a "short-circuit" manner.

[...]

PR #2283, opened 2023-12-25
Add problematic columnIndex in exception message.

israelaece:
Improving exception message with column index that was not found. #834


David-Engel:
If you want to get this feature change in, there are more places that need updating (as noted). Please submit a new PR against the latest code along with the other places that would need to be updated for this feature.

PR #2284, opened 2023-12-30
use System.Configuration.ConfigurationManager 8.0,

### LINKED ISSUE #2232: System.Configuration.ConfigurationManager Version
Hi,

the current Microsoft.Data.SqlClient depends on the System.Configuration.ConfigurationManager Version 6.0.1 - which forces us the System.Drawing.Common Version 6.0 - which is not wanted since this is Win-Only - https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/6.0/system-drawing-common-windows-only

image

Is there any info on this? It would be at least helpful if there would be a newer release based on .net8 since System.Configuration.ConfigurationManager or the be more precise the System.Windows.Extensions does not depend on System.Drawing.Common anymore.

Any thoughts?

ErikEJ:
as it has less dependencies than previous versions

fixes #2232

### DIFF
--- a/tools/props/Versions.props
+++ b/tools/props/Versions.props
@@ -39,7 +39,7 @@
   <PropertyGroup>
     <MicrosoftWin32RegistryVersion>5.0.0</MicrosoftWin32RegistryVersion>
     <MicrosoftDataSqlClientSNIRuntimeVersion>5.2.0-preview1.23340.1</MicrosoftDataSqlClientSNIRuntimeVersion>
-    <SystemConfigurationConfigurationManagerVersion>6.0.1</SystemConfigurationConfigurationManagerVersion>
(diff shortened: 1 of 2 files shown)


David-Engel:
We are trying not to make major version dependency changes (including transitive) within non-major version bumps of MDS. So we'll merge this after the 5.2 GA. We'll bump to MDS 6.0 and include this change. It would be a nice refactor if we could have different dependency versions for the different target frameworks. I think that would allow us to make changes like this without breaking others. For example, adding future .NET 10 support with different dependency versions without a MDS major version bump and without breaking apps still targeting .NET 8.

ErikEJ:
@David-Engel Ah - so that is why the .NET 8 dependencies are staying at 6.x (for 5.2 at least)

I will see if I can find an approach to grouping the dependencies better - it already happens in the .nuspec file.

ErikEJ:
Superseeded by #2303

PR #2306, opened 2024-01-18
Perf | avoid boxing of SqlGuid in SqlBuffer (#2300)

### LINKED ISSUE #2300: High memory traffic because of boxed SqlGuid struct
When our application is starting it loads a lot of data from DB and it causes memory consumption growth and high pressure to GC because of big memory traffic.
On the third place of killed objects is boxed `System.Data.SqlTypes.SqlGuid` struct.
These boxed values are 99.9% created when `TdsParser` is putting `SqlGuid` struct to `SqlBuffer.SqlGuid` property (see the image bellow).

image

`SqlBuffer` already supports an efficient way to store `Guid` values. What I would like to do is:
- change implementation of `Microsoft.Data.SqlClient.SqlBuffer`
  -  ~~remove `SqlGuid` value from `StorageType` enum~~ (taking the second alternative approach, removing of `SqlGuid` value from enum looks like a bad idea);
  - save `Guid` representation of `SqlGuid` to `_value._guid` (similar to what `Guid` setter does, but with checking if `SqlGuid` value is null)
  - read `SqlGuid` value from `_value._guid`

1. add a field of `SqlGuid` type to `SqlBuffer`, but it seems memory inefficient
2. same as the main approach, but without removing `SqlGuid` value from `StorageType` enum, in case it is really important. The only place where it looks important to me is `s_dbTypeToStorageType` mapping array in `Microsoft.Data.SqlClient.Server.ValueUtilsSmi`, and I am not sure if it is safe to change the mapping to `SqlBuffer.StorageType.Guid`. It seems like it is not safe, because implementation of `SqlBuffer.Guid` does not support nulls.

### Additional context
- Microsoft.Data.SqlClient version is 5.1.0;
- We use Microsoft.Data.SqlClient together with NHibernate;
- Almost all our db tables contain Guid columns, so, memory traffic is really high in our case.

wilbit:
fixes #2300

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBuffer.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBuffer.cs
@@ -109,7 +109,7 @@ namespace Microsoft.Data.SqlClient
         private bool _isNull;
         private StorageType _type;
         private Storage _value;
-        private object _object;    // String, SqlBinary, SqlCachedBuffer, SqlGuid, SqlString, SqlXml
+        private object _object;    // String, SqlBinary, SqlCachedBuffer, SqlString, SqlXml
 
         internal SqlBuffer()
         {
@@ -815,14 +815,17 @@ namespace Microsoft.Data.SqlClient
                 }
                 else if (StorageType.SqlGuid == _type)
                 {
-                    return IsNull ? SqlGuid.Null : (SqlGuid)_object;
+                    return IsNull ? SqlGuid.Null : new SqlGuid(_value._guid);
                 }
                 return (SqlGuid)SqlValue; // anything else we haven't thought of goes through boxing.
             }
             set
             {
                 Debug.Assert(IsEmpty, "setting value a second time?");
-                _object = value;
+                if (!value.IsNull)
+                {
+                    _value._guid = value.Value;
+                }
                 _type = StorageType.SqlGuid;
                 _isNull = value.IsNull;
             }
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
@@ -5504,7 +5504,7 @@ namespace Microsoft.Data.SqlClient
                     break;
 
                 case SqlDbType.UniqueIdentifier:
-                    nullVal.SqlGuid = SqlGuid.Null;
+                    nullVal.SetToNullOfType(SqlBuffer.StorageType.SqlGuid);
                     break;
 
                 case SqlDbType.Bit:
--- a/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
+++ b/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
@@ -6310,7 +6310,7 @@ namespace Microsoft.Data.SqlClient
                     break;
 
                 case SqlDbType.UniqueIdentifier:
-                    nullVal.SqlGuid = SqlGuid.Null;
+                    nullVal.SetToNullOfType(SqlBuffer.StorageType.SqlGuid);
                     break;
 
                 case SqlDbType.Bit:
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/ValueUtilsSmi.netfx.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/ValueUtilsSmi.netfx.cs
@@ -152,10 +152,6 @@ namespace Microsoft.Data.SqlClient.Server
                     // special case SqlBinary, 'cause tds parser never sets SqlBuffer to null, just to empty!
                     targetBuffer.SqlBinary = SqlBinary.Null;
                 }
-                else if (SqlBuffer.StorageType.SqlGuid == stype)
-                {
-                    targetBuffer.SqlGuid = SqlGuid.Null;
-                }
                 else
                 {
                     targetBuffer.SetToNullOfType(stype);


wilbit:
@dotnet-policy-service agree [company="Intrigma"]

wilbit:
@dotnet-policy-service agree company="Intrigma"

Wraith2 (approved):
As long as the tests pass this looks good to me.

David-Engel (commented):
My only concern is that SqlBuffer.cs lines 818 and 825-828 aren't exercised by tests and I don't know enough about the changes to know if they are correct or not. Are those lines easily covered by extending one of the existing tests?

wilbit:
Honestly, I have not found any unit tests for SqlBuffer. I suppose I can add unit tests for writing/reading of values, but I not sure how to deal with boxing (over reflection and checking of object field, probably 🙃)

Wraith2:
> My only concern is that SqlBuffer.cs lines 818 and 825-828 aren't exercised by tests and I don't know enough about the changes to know if they are correct or not. Are those lines easily covered by extending one of the existing tests?

The logic to set the unmanaged `_value` buffer is not currently needed and we can't write a test for it without causing the condition this PR fixes to be possible through the library or through the use of complex reflection. The code is a safety net in case someone calls the setter in future because without this code being present (even if unused) if someone adds a call to the setter with a non-null value it'll compile and run but won't do the correct operation. An exception would require a lot of plumbing to get from here to the tds parser correctly and an assertion would be of no used because we run tests in Release mode.

David-Engel:
> I prepare another PR for the similar issue ([#2300 (comment)](https://github.com/dotnet/SqlClient/issues/2300#issuecomment-1898451114)) with boxed `SqlBinary`. Is it acceptable if I add unit tests for `SqlBuffer.Guid`, `SqlBuffer.SqlGuid` and `SqlBuffer.SqlBinary` in the upcoming PR?

If not too difficult to test, yes. Not sure if it's easy (or completely necessary) based on what Wraith2 said.

> The logic to set the unmanaged `_value` buffer is not currently needed and we can't write a test for it without causing the condition this PR fixes to be possible through the library or through the use of complex reflection. The code is a safety net in case someone calls the setter in future because without this code being present (even if unused) if someone adds a call to the setter with a non-null value it'll compile and run but won't do the correct operation. An exception would require a lot of plumbing to get from here to the tds parser correctly and an assertion would be of no used because we run tests in Release mode.

Thanks for the clarification. I'm fine with the changes, then.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBuffer.cs
```diff
@@ -815,14 +815,17 @@ internal SqlGuid SqlGuid
...
-                    return IsNull ? SqlGuid.Null : (SqlGuid)_object;
+                    return IsNull ? SqlGuid.Null : new SqlGuid(_value._guid);
                 }
                 return (SqlGuid)SqlValue; // anything else we haven't thought of goes through boxing.
             }
             set
             {
                 Debug.Assert(IsEmpty, "setting value a second time?");
-                _object = value;
+                if (!value.IsNull)
```
DavoudEshtehari, line 825: If `_value` is already set and the new `value` is set to null, there might be an inconsistency where `_value` still points to a non-null value while `_isNull` indicates that it's null. This could lead to confusion and unexpected behavior in your code. It's crucial to ensure that `_value` and `_isNull` are consistent and accurately represent the state of the variable. IMO, If the intention is to set the variable to null, both `_value` and `_isNull` should be appropriately updated to reflect this.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBuffer.cs
```diff
@@ -815,14 +815,17 @@ internal SqlGuid SqlGuid
                 }
                 else if (StorageType.SqlGuid == _type)
                 {
-                    return IsNull ? SqlGuid.Null : (SqlGuid)_object;
+                    return IsNull ? SqlGuid.Null : new SqlGuid(_value._guid);
```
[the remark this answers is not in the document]
Wraith2: The efficiency isn't gained here, This is just a knockon effect. The efficiency is gained by not assigning a struct `SqlNull.Null` to an object which causes an unavoidable box every time it happens.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBuffer.cs
Wraith2: `_value` is a value type. it is the value it doesn't point to it. If the value is non-empty and then is updated to null it doesn't matter if there is a non-empty value in the variable because it won't be read through any of the code paths.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBuffer.cs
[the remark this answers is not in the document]
wilbit: I just did perf testing and there are some perf penalties brought by this PR, so, @DavoudEshtehari highlighted a valid point 👍.
I will do more perf testing and try to find a better solution in terms of memory and performance.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBuffer.cs
wilbit: The class definitely requires its own unit tests.
I have introduced the bug in this PR in [`SqlBuffer.get_SqlValue`](https://github.com/dotnet/SqlClient/pull/2306/files/54328bdbd429029273d0f48603a800f1c0317fd2#diff-48a7ee0a7ba0aea47e0131ff3033e85fe67de5120028484929bc0a8aad8f371cR988) 😥:
```csharp
case StorageType.SqlBinary:
case StorageType.SqlGuid:
    return _object;
```

wilbit: I don't understand what is the right place to put unit tests.
I thought it is 'src\Microsoft.Data.SqlClient\tests\FunctionalTests\Microsoft.Data.SqlClient.Tests.csproj' but this project does not target all targets of Microsoft.Data.SqlClient, particularly, it misses `net80`.
May you point me to the right direction? I want to fix the bug ASAP, @Wraith2 , @David-Engel , @DavoudEshtehari

Wraith2: We don't have a net8 build currently. Write the test and use the correct attribute or preprocessor definition so that when the net8 build does exist (possibly soon-ish) it will be enabled and work.

Functional tests are run without a database connection. Anything that can be checked without running a query can go in there. Manual tests are run with a database connection, anything which can't be in functional tests belongs there.

JRahnama: > We don't have a net8 build currently

Net8 and its related changes were merged last week by #2230

wilbit: I found and fixed a couple of bugs, but faced the issue with functional tests:
- I cannot find the right way to make internal classes of Microsoft.Data.SqlClient available for Microsoft.Data.SqlClient.Tests.
`[assembly: InternalsVisibleTo("FunctionalTests")]` does not work and, maybe, it is not the right way :\
Thought, @Wraith2 ?

PS: Also, it is very inconvenient that there is no way to run a particular unit test from IDE :(

Wraith2: We should not use internals visible to. You'd need to use reflection if you want to test internals. Or, possibly include the file in the tests so it is local and then use it directly but this can be difficult because of dependencies in the code.

What is the root of the bugs and how did the get past previous review?

ErikEJ: Should the merge be reverted?

wilbit: @Wraith2 what approach is better (already used widely)?

@ErikEJ, I don't like revert commits and avoid them as many ch as possible. But it is up to you. FYI I'm preparing PR with the fixes and will submit it before Monday.

Wraith2: It depends. I'd like more details on the bug and how it was found.
If the bug is a logical bug found by looking at the code but there is no way for users to reach that piece of code and that's why there are no covering tests then it's not a huge problem. In this case i'd say work forwards from here.
If the bug is a practical but found by hitting incorrect behaviour at runtime then we're clearly missing code coverage of a reachable path and it's a more important problem. In this case i'd say revert and retry.

wilbit: These are the bugs I'm talking about: https://github.com/dotnet/SqlClient/compare/main...Intrigma:SqlClient:sqlbuffer-bugs

Also, I made alternative performance/memory improvements, which works much better than #2306: https://github.com/dotnet/SqlClient/compare/main...Intrigma:SqlClient:sqlbuffer-improvements
I will share my benchmark project and its result later with PR. For now lets decide what we do with the bugs, @Wraith2

ErikEJ: I vote for a revert given that a release is planned for 24 Jan - just my 2 cents

wilbit: Here the PR to fix that: #2310

wilbit: @Wraith2 suggested to revert this PR too. See https://github.com/dotnet/SqlClient/pull/2310#discussion_r1461497741

ErikEJ: I just did a PR to revert.

Wraith2: Now we just have to wait for the MS team to work their way through the email notifications, wonder what we've all been doing all weekend, and then approve it 😁

PR #2308, opened 2024-01-19
Perf | various performance improvements for SqlDataReader, TdsParser, SqlGuid and SqlBinary 

wilbit:
I was working on boxing of `SqlBinary` particularly, but found more kinda related issues.
The PR fixes:
- boxing of `SqlBinary` struct;
- copying of `SqlBinary`'s internal array of bytes back and forth;
- creating of empty arrays for null values in SqlDataReader;
- copying of an array of bytes for `SqlGuid`.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
@@ -5504,7 +5504,7 @@ namespace Microsoft.Data.SqlClient
                     break;
 
                 case SqlDbType.UniqueIdentifier:
-                    nullVal.SqlGuid = SqlGuid.Null;
+                    nullVal.SetToNullOfType(SqlBuffer.StorageType.SqlGuid);
                     break;
 
                 case SqlDbType.Bit:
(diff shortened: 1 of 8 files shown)


Wraith2:
For changes like avoiding the SqlBinary getter consider exceptions. For example, if that code path end up with the wrong data in the buffer, a string or something else, and you request the SqlBinary value currently it's going to throw an InvalidCastException, will your changed code path do the same? Because it's a very old library any externally visible behaviour change is going to break some code somewhere.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -5957,7 +5957,11 @@ internal bool DeserializeUnencryptedValue(SqlBuffer value, byte[] unencryptedByt
                             unencryptedBytes = bytes;
                         }
 
-                        value.SqlBinary = new SqlBinary(unencryptedBytes);   // doesn't copy the byte array
+#if NET7_0_OR_GREATER
+                        value.SqlBinary = SqlBinary.WrapBytes(unencryptedBytes);
+#else
+                        value.SqlBinary = SqlTypeWorkarounds.SqlBinaryCtor(unencryptedBytes, true);   // doesn't copy the byte array
+#endif
```
Wraith2: This may not be safe because unencryptedBytes is a parameter so you will need to audit all the callsites to make sure that ownership is always relinquished by the callers.

wilbit:
Thanks for the feedback, @Wraith2 , I will consider that.

cheenamalhotra:
_Closed PR since it's been inactive for more than a month and is not ready to be reviewed._

PR #2310, opened 2024-01-21
Fix | SqlBuffer.SqlGuid

### LINKED ISSUE #2300: High memory traffic because of boxed SqlGuid struct

wilbit:
Bug description:
`SqlBuffer.get_SqlGuid` could return `SqlGuid` with incorrect (or empty) `Guid` value inside after `SqlBuffer.SetToNullOfType(StorageType.Guid)` is invoked.

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBuffer.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBuffer.cs
@@ -811,7 +811,7 @@ namespace Microsoft.Data.SqlClient
             {
                 if (StorageType.Guid == _type)
                 {
-                    return new SqlGuid(_value._guid);
+                    return IsNull ? SqlGuid.Null : new SqlGuid(_value._guid);
                 }
                 else if (StorageType.SqlGuid == _type)
                 {
--- a/.gitignore
+++ b/.gitignore
@@ -353,6 +353,9 @@ healthchecksdb
 # Backup folder for Package Reference Convert tool in Visual Studio 2017
 MigrationBackup/
 
+# JetBrains Rider (cross platform .NET IDE) working folder
+.idea/
+
 # Ionide (cross platform F# VS Code tools) working folder
 .ionide/
 
--- a/src/Microsoft.Data.SqlClient/tests/FunctionalTests/Microsoft.Data.SqlClient.Tests.csproj
+++ b/src/Microsoft.Data.SqlClient/tests/FunctionalTests/Microsoft.Data.SqlClient.Tests.csproj
@@ -32,6 +32,7 @@
     <Compile Include="LocalizationTest.cs" />
     <Compile Include="MultipartIdentifierTests.cs" />
     <Compile Include="SqlAuthenticationProviderTest.cs" />
+    <Compile Include="SqlBufferTests.cs" />
     <Compile Include="SqlClientLoggerTest.cs" />
     <Compile Include="SqlCommandSetTest.cs" />
     <Compile Include="SqlConfigurableRetryLogicTest.cs" />
--- a/src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlBufferTests.cs
+++ b/src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlBufferTests.cs
@@ -1,0 +1,253 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Data.SqlTypes;
+using System.Linq;
+using System.Reflection;
+using Xunit;
+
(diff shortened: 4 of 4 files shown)


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBuffer.cs
```diff
@@ -109,6 +113,9 @@ internal struct Storage
         private bool _isNull;
         private StorageType _type;
         private Storage _value;
+#if !NET8_0_OR_GREATER
+        private SqlGuid _sqlGuid;
+#endif
```
Wraith2, line 118: This increases the size of the SqlBuffer type by guid+bool on all previous runtimes, this is almost doubling the size. Since SqlBuffer is used in every read this is going to cause a lot of additional memory traffic. I don't think this is a good idea.

wilbit: Yep, it is increases the size of SqlBuffer for platforms before .NET8, by 8 bytes. And that's it, no other downsides.

Wraith2: That's a big downside. Every read goes through an SqlBuffer and this almost doubles the memory required to do that.

wilbit: For GC there will be no difference, I guess, because number of references has not changed (SqlGuid is a struct).

If it is not acceptable... Well, what is an alternative? Keep the old implementation with boxing for non .NET8 platforms?

Wraith2: Larger is worse for throughput on the heap. I've put quite a lot of effort into shrinking various objects to maximize throughput. 

At this point i think it's worth reverting your other PR so we get a clean release (we've been waiting 6 months for one) then I'd like you to add your covering test cases alone and then we can investigate slowly.

wilbit: Fair point.
OK, lets revert #2306 , reopen #2300 and focus on

1. bugs in SqlBuffer first
2. removing of boxing for .NET8
3. removing of boxing for the rest platforms

Sounds good?

roji: Just to point out that an actual benchmark should easily show what's significant here and what isn't.

wilbit:
Because the revert PR #2311 is merged, I will replace this PR's commits with function tests for `SqlBuffer` and a bugfix

wilbit:
Hmm, I don't get what it actually failed in the tests =\

DavoudEshtehari:
> Hmm, I don't get what it actually failed in the tests =\

@wilbit No need to worry; this seems to be a random Azure resource issue.

wilbit:
Anything?

JRahnama:
> Anything?

This will be reviewed for the next preview release. We are on code freeze now for the GA release.

Wraith2 (approved):
Sorry for the delay, real life stuff happened.

wilbit:
May we merge this PR?
I have other changes (improvements) for `SqlBuffer`, but they conflict with this PR, so it does not make sense to submit them =\

David-Engel (approved):

DavoudEshtehari (approved):

PR #2334, opened 2024-02-01
Fix | Adding missing commitID to AKV nuspec

JRahnama:


DavoudEshtehari (approved):
Can you update Microsoft.SqlServer.Server.nuspec as well?

arellegue (approved):
LGTM

Issue #2322, opened 2024-01-26
Mismatch between versions.props and nuspec


ErikEJ:
In versions.props:

`<SystemDiagnosticsDiagnosticSourceVersion>7.0.2</SystemDiagnosticsDiagnosticSourceVersion>`

In nuspec (in 3 places, net6.0, netstandard2.0, netstandard2.1)

`<dependency id="System.Diagnostics.DiagnosticSource" version="6.0.1" exclude="Compile" />`

PR #2349, opened 2024-02-08
Fix | Mismatch System.Diagnostics.DiagnosticSource between versions.props and nuspec 

arellegue:
#2322

Changed System.Diagnostics.DiagnosticSource version from 7.0.1 to 6.0.1 in versions.props.

Removed System.Diagnostics.DiagnosticSource from Net8 and Net6 in the nuspec file.

### DIFF
--- a/tools/props/Versions.props
+++ b/tools/props/Versions.props
@@ -51,7 +51,7 @@
   <!-- NetStandard project dependencies -->
   <PropertyGroup>
     <SystemRuntimeLoaderVersion>4.3.0</SystemRuntimeLoaderVersion>
-    <SystemDiagnosticsDiagnosticSourceVersion>7.0.2</SystemDiagnosticsDiagnosticSourceVersion>
+    <SystemDiagnosticsDiagnosticSourceVersion>6.0.1</SystemDiagnosticsDiagnosticSourceVersion>
   </PropertyGroup>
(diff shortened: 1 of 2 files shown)


David-Engel (approved):

David-Engel:
> @arellegue , Can you review the test projects. Looks like they are failing due to the downgrade.
> 
> ```
> ##[error]src\Microsoft.Data.SqlClient\tests\FunctionalTests\Microsoft.Data.SqlClient.Tests.csproj(0,0): Error NU1605: Warning As Error: Detected package downgrade: System.Diagnostics.DiagnosticSource from 7.0.1 to 6.0.1. Reference the package directly from the project to select a different version. 
>  FunctionalTests -> Microsoft.Extensions.Hosting 7.0.1 -> System.Diagnostics.DiagnosticSource (>= 7.0.1) 
>  FunctionalTests -> System.Diagnostics.DiagnosticSource (>= 6.0.1)```
> ```

I wonder if we found the reason it was set to 7 in Version.props...

If the reference is only needed for the tests, it should be fine to leave it in Versions.props. Then we just need to validate that tracing on .NET 6 and .NET 8 still works without those dependencies in the resulting NuGet package. (Test against the package.)

H-Yeo (approved):

PR #2354, opened 2024-02-09
SDL | Setting ValidateAudience to true to address CodeQL flag

JRahnama:
This PR is setting the default value of ValidateAudience in  TokenValidationParameters to true.


arellegue (approved):
Looks good to me.

David-Engel:
Looks like the enclave test failures are due to this change:

`Microsoft.Data.SqlClient.SqlException : The validation of an attestation token failed. The token signature does not match the signature omputed using a public key retrieved from the attestation public key endpoint at 'IDX10208: Unable to validate audience. validationParameters.ValidAudience is null or whitespace and validationParameters.ValidAudiences is null.'. Verify the DNS apping for the endpoint - see https://go.microsoft.com/fwlink/?linkid=2157649 for more details. If correct, contact Customer Support Services.`

PR #2366, opened 2024-02-21
Build Tweaks for Onboarding

benrr101:
**Description**: Fixing a handful of issues encountered while onboarding to the project
* Fix broken links to xml docs
* Ensure all xml doc files are included in the solution, removing xml doc files that have been removed
* Fix build issue related to building solution with msbuild (the "GetTargetPath" issue)
* Adding dotnet-eng nuget feed to top nuget.config
  * This may not be the ideal solution since it allows engineering packages to be added to non-test projects. However, visual studio does not honor nuget.config on a per-project basis (see https://developercommunity.visualstudio.com/t/visual-studio-nuget-package-manager-ignores-nugetc/24733?viewtype=all), and relying on cached packages to build in VS is not a good experience.

**Testing**:
* Project builds in Visual Studio, Rider, and via build.proj
* Tests that can be expected to pass continue to pass

### DIFF
--- a/src/NuGet.config
+++ b/src/NuGet.config
@@ -3,5 +3,6 @@
   <packageSources>
     <clear />
     <add key="dotnet-public" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json" />
+    <add key="dotnet-eng" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json" />
   </packageSources>
 </configuration>
--- a/src/NuGet.config
+++ b/src/NuGet.config
@@ -2,7 +2,6 @@
 <configuration>
   <packageSources>
     <clear />
-    <add key="dotnet-public" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json" />
-    <add key="dotnet-eng" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json" />
+    <add key="sqlclient" value="https://sqlclientdrivers.pkgs.visualstudio.com/public/_packaging/sqlclient/nuget/v3/index.json" />
   </packageSources>
 </configuration>
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -95,5 +95,10 @@
   <Import Project="$(ToolsDir)props\Tools.props" />
   <!-- Provides Build properties -->
   <Import Project="$(ToolsDir)props\AssemblyInfo.props" /> 
-  <Import Project="$(ToolsDir)props\AssemblyRef.props" />  
+  <Import Project="$(ToolsDir)props\AssemblyRef.props" />
+
+  <!-- Fix GetTargetPath issues - see https://github.com/dotnet/msbuild/issues/4303#issuecomment-482345617 -->
+  <PropertyGroup>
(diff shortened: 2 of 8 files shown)


cheenamalhotra (approved):

src/NuGet.config
```diff
@@ -3,5 +3,6 @@
   <packageSources>
     <clear />
     <add key="dotnet-public" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json" />
+    <add key="dotnet-eng" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json" />
```
JRahnama, line 6: Just wondering if this is going to be flagged for component governance. We used to get flagged for having two NuGet sources, but is it okay since both are from Microsoft?

cheenamalhotra: Can we enable CG on the CI pipeline and confirm that?

benrr101:
@cheenamalhotra Any guidance on the test failures? Are these tests flaky or expected to fail? Or did I break something?

JRahnama:
> @cheenamalhotra Any guidance on the test failures? Are these tests flaky or expected to fail? Or did I break something?

Test failures seems not related to your changes. I will re run the failed jobs to confirm that.

src/NuGet.config
David-Engel: I'm sure this would get flagged if we merged the change. Long term solution is to create our own public pipeline with all required upstreams. But that come with maintenance costs defining those and managing installed package versions (if we want to do it securely). Alternatively, we can try to get off any packages that we are using from the dotnet-eng feed. Or just keep things the way they are where you have to restore the test project to cache the package before building everything.

benrr101: Did we enable CG and confirm?

It's not a great developer experience to have to restore outside of the IDE in order the build inside the IDE, though it does work. I would also be ok with removing the dependencies on dotnet-eng feed, though in my experiments so far, the conditional testing is pretty deeply ingrained in the tests.

If we can confirm the change will get flagged, I can back it out from this PR.

cheenamalhotra: I was wondering how other products like [EF Core](https://github.com/dotnet/efcore/blob/main/NuGet.config) manage multiple feeds in their configuration.. we may want to spend some time understanding the right solution that doesn't involve much dev effort every-time and provides smoother experience to all users.

ErikEJ: EF Core https://github.com/dotnet/efcore/blob/main/NuGet.config

David-Engel: Try this singular feed:

```
<add key="sqlclient" value="https://sqlclientdrivers.pkgs.visualstudio.com/public/_packaging/sqlclient/nuget/v3/index.json" />
```

And then `src/Microsoft.Data.SqlClient/tests/NuGet.config` can probably be deleted.

David-Engel: No. CG isn't enabled on branch builds. Only main/production builds.

benrr101: Well ... looks like using that single feed didn't work in the pipelines. Any suggestions?

David-Engel: The pipeline just needed a NuGet Authenticate task so that it can install upstream versions of packages to the new feed. I added that and kicked off a new run.

benrr101: @David-Engel it looks like the new builds passed. Feel confident about giving it a approval? 😃

JRahnama (approved):

David-Engel (approved):

Issue #2346, opened 2024-02-07
Docs | Undocumented API Functionality in SqlClientFactory for CreateDataSourceEnumerator


H-Yeo:
### Describe the bug
While reviewing Microsoft documentation,
SqlClientFactory.cs (implementation) and SqlClientFactory.xml (documentation) the following method:
```  C# public override DbDataSourceEnumerator CreateDataSourceEnumerator() ```

.NET Core project reference is **missing** this implementation
../netcore/ref/Microsoft.Data.SqlClient.cs

.NET FX project reference **has** this implementation
../netfx/ref/Microsoft.Data.SqlClient.cs

The original commit hash was from 2019:
d6f2e29cda769e711988c64d3cb57eb5010a3e9c

### To reproduce
Review Documentation between .NET Core and .NET FX
Documentation References:

.NET Core (missing) - Expand Methods:
https://learn.microsoft.com/en-us/dotnet/api/microsoft.data.sqlclient?view=sqlclient-dotnet-core-5.1

.NET FX (available) - Expand Methods:
https://learn.microsoft.com/en-us/dotnet/api/microsoft.data.sqlclient?view=sqlclient-dotnet-5.1

### Expected behavior
Missing method for .NET Core documentation is updated to be similar/identical to .NET FX documentation.

### Further technical details
Microsoft.Data.SqlClient version: 5.2 (Preview)
.NET target: .NET Core 3.0+ (could be earlier, going off the commit here)
SQL Server version: n/a
Operating system: n/a

**Additional context**
Fix, Add missing reference for .NET Core
MDS team generates API Documentation and verify it's published

cheenamalhotra:
This has been addressed with 6.0: https://learn.microsoft.com/en-us/dotnet/api/microsoft.data.sqlclient.sqlclientfactory.createdatasourceenumerator?view=sqlclient-dotnet-core-6.0

Issue #17, opened 2019-05-15
SqlClient implementation and API is not complete in Core


danmosemsft:
The following members are on types implemented in .NET Core. However these particular members are missing, although they are present on Desktop.

The problem with this situation is
- it defies user expectations that if a type X exists, code from other platforms that use X will work on .NET Core
- it cannot be fixed without servicing -- it is not possible to supply a standalone package with the rest of the type definition

No work is necessarily required here for 2.0. It's not part of NS2.0. I didn't look at usage data. I'm just flagging this so we're doing this consciously.

---

**Updated lists to reflect current status (5/15):**

### Missing APIs
```
T:System.Data.SqlClient.OnChangeEventHandler
T:System.Data.SqlClient.PoolBlockingPeriod
T:System.Data.SqlClient.SqlAuthenticationMethod
M:System.Data.SqlClient.SqlBulkCopyColumnMappingCollection.Clear
M:System.Data.SqlClient.SqlBulkCopyColumnMappingCollection.RemoveAt(System.Int32)
F:System.Data.SqlClient.SqlBulkCopyOptions.AllowEncryptedValueModifications
M:System.Data.SqlClient.SqlClientFactory.System#IServiceProvider#GetService(System.Type)
T:System.Data.SqlClient.SqlClientMetaDataCollectionNames
T:System.Data.SqlClient.SqlColumnEncryptionCertificateStoreProvider
T:System.Data.SqlClient.SqlColumnEncryptionCngProvider
T:System.Data.SqlClient.SqlColumnEncryptionCspProvider
T:System.Data.SqlClient.SqlColumnEncryptionKeyStoreProvider
P:System.Data.SqlClient.SqlCommand.ColumnEncryptionSetting
P:System.Data.SqlClient.SqlCommand.Notification
P:System.Data.SqlClient.SqlCommand.NotificationAutoEnlist
M:System.Data.SqlClient.SqlCommand.BeginExecuteNonQuery
M:System.Data.SqlClient.SqlCommand.BeginExecuteNonQuery(System.AsyncCallback,System.Object)
M:System.Data.SqlClient.SqlCommand.BeginExecuteReader
M:System.Data.SqlClient.SqlCommand.BeginExecuteReader(System.AsyncCallback,System.Object)
M:System.Data.SqlClient.SqlCommand.BeginExecuteReader(System.AsyncCallback,System.Object,System.Data.CommandBehavior)
[...]

PR #2369, opened 2024-02-25
Merged SqlClientFactory between .NET Core and Framework

edwardneal:
Contributes to #1261. Also relates to #17 (specifically the last comment about SqlDataSourceEnumerator) and #2346.

This merges `SqlClientFactory` between .NET Core and Framework. In the process, it also implements `CreateDataSourceEnumerator` in .NET, linking it to the newly-added `SqlDataSourceEnumerator` class.

In .NET Framework, SqlClientFactory also provides `CreatePermission` and implements the `IServiceProvider` interface. I've left the latter in place to avoid a breaking change, although IServiceProvider is only used to access `SqlProviderServices` (which is [not part of the project](https://github.com/dotnet/SqlClient/issues/17#issuecomment-301640792).)


src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -428,6 +428,16 @@ public sealed partial class SqlClientFactory : System.Data.Common.DbProviderFact
         public override System.Data.Common.DbDataAdapter CreateDataAdapter() { throw null; }
         /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlClientFactory.xml' path='docs/members[@name="SqlClientFactory"]/CreateParameter/*'/>
         public override System.Data.Common.DbParameter CreateParameter() { throw null; }
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlClientFactory.xml' path='docs/members[@name="SqlClientFactory"]/CreateDataSourceEnumerator/*'/>
+        public override System.Data.Common.DbDataSourceEnumerator CreateDataSourceEnumerator() { throw null; }
+#if NET6_0_OR_GREATER
```
DavoudEshtehari, line 433: This isn't required anymore as the lowest supported target framework version is `net6.0`.

DavoudEshtehari (approved):

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
edwardneal: Yes, understood and agreed - this PR pre-dates the removal of .NET Standard support. I'll bring it up to date for review.

David-Engel (approved):

PR #2371, opened 2024-02-26
Fix | Refine documents for GA v5.2

DavoudEshtehari:
- Added missing public APIs.
- Removed the use of `inheritdoc` due to its lack of support for cross-repository references and a few issues with the documentation engine.
- Removed redundant XML files.
- Followed the XML file pattern and standardized documentation.
- Made a few documentation improvements.

[Smoke-test](https://review.learn.microsoft.com/en-us/dotnet/api/microsoft.data.sqlclient.sqlbatch?view=sqlclient-dotnet-core-5.2&branch=pr-en-us-75)


src/Microsoft.Data.SqlClient/ref/Microsoft.Data.SqlClient.Batch.NetCoreApp.cs
```diff
@@ -1,9 +1,29 @@
...
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnection.xml' path='docs/members[@name="SqlConnection"]/CanCreateBatch/*'/>
+        public override bool CanCreateBatch { get { throw null; } }
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnection.xml' path='docs/members[@name="SqlConnection"]/CreateDbBatch/*'/>
+        protected override System.Data.Common.DbBatch CreateDbBatch() => throw null;
+    }
+
+    /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlException.xml' path='docs/members[@name="SqlException"]/SqlException/*' />
+    public partial class SqlException
+    {
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlException.xml' path='docs/members[@name="SqlException"]/DbBatchCommand/*' />
```
David-Engel, line 26: Seems like these include paths should all be similar. `../../..`?

DavoudEshtehari: The path to the ref Batch CS files is introduced alongside the SqlBatch feature with different hierarchy. Should we consider consolidating it into the same CS file?

David-Engel: If it works as-is, it's fine with me. I just assumed the paths would be relative to the file and thus would need to reference the same number of parent folders.

JRahnama (approved):

David-Engel (approved):

PR #2376, opened 2024-02-28
Merge AlwaysEncryptedHelperClasses, TdsParserHelperClasses

edwardneal:
Contributes to #1261.

This merges AlwaysEncryptedHelperClasses.cs and TdsParserHelperClasses between .NET Framework and .NET Core. There are some lingering Framework-only and Core-only components, which I've added to <file>.netfx.cs or <file>.NetCoreApp.cs respectively.

This is a large PR - sorry in advance. Almost the entire reason why this is so large is that I pulled the two files into the coding standard as part of the merge process. There were a lot of internal fields here; the coding standard calls for them to have an `_` prefix, and there are a lot of references to them.

I've tried to simplify this by splitting it out into various commits:
1. Split the classes in netcore's AlwaysEncryptedHelperClasses.cs into a corresponding file in the netfx project
2. Minor cleanup prior to merge of AEHC.cs - removing an unused field from netfx which wasn't present in the netcore projejct
3. Merge activity. Deleted the netfx AlwaysEncryptedHelperClasses.cs file, moved the netcore file into the root directory
4. First large change: all non-public fields in AEHC.cs prefixed with an underscore, internal classes marked as sealed where possible
5. Split netcore's and netfx's remaining TdsParserHelperClasses.cs files, pushing all non-common code into their corresponding TdsParserHelperClasses.NetCoreApp.cs and TdsParserHelperClasses.netfx.cs files
6. Merge activity, and syncing up a field name to make the netfx project compile (netcore had an underscore, netfx did not.) Internal classes marked as sealed where possible
7. Second large change: all non-public fields in TPHC.cs prefixed with an underscore
8. Mop-up: correcting two tests which used reflection to manually access fields

Commits 4 and 7 have the widest impact. I'm not sure whether they're the right things to do though. Most of the time, the classes and the fields are both `internal`. In that situation, I'm tempted to make the fields public (and keep the classes internal.) This would cut the PR size down, but not by much - we'd just end up updating references to `baseTI` to reference `BaseTI` rather than `_baseTI`.


DavoudEshtehari:
@edwardneal Can you address the conflicts, please?

edwardneal:
Thanks @DavoudEshtehari, I've completed the merge.

The PR touches a lot of files, and the majority of these changes come from the corrections to field names. I can strip this change out if that'll make the next PRs easier to merge.

DavoudEshtehari:
@edwardneal A smaller code change would definitely be easier to review with less risk of missing something. I'd appreciate it if you could split them meaningfully and keep them small.

edwardneal:
@DavoudEshtehari, that's definitely improved things - it's eliminated two thirds of the diff, and most of the remainder is the deletion of the .NET Framework's TdsParserHelperClasses.cs. I'll push the standardisation into a separate PR when the merge will be less painful.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AlwaysEncryptedHelperClasses.cs
```diff
@@ -14,7 +14,7 @@ namespace Microsoft.Data.SqlClient
     /// <para> Represents a single encrypted value for a CEK. It contains the encrypted CEK,
     ///  the store type, name,the key path and encryption algorithm.</para>
     /// </summary>
-    internal class SqlEncryptionKeyInfo
+    internal sealed class SqlEncryptionKeyInfo
```
JRahnama, line 17: Sealing an internal class, while it may offer some performance improvements, can make testing more difficult. Given this, we opted to use `InternalsVisibleToAttribute` for some of our classes to facilitate testing. Is there any specific reasoning behind it?

edwardneal: There's no specific rationale beyond performance - it's on the hot path, so I usually start off sealing non-inherited classes then unsealing them as appropriate. I'll revert the sealing (and one other field rename which I missed earlier.)

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserHelperClasses.cs
```diff
@@ -791,14 +812,100 @@ private static string ToFriendlyName(this SslProtocols protocol)
...
+            SP_PROT_TLS1_2 = SP_PROT_TLS1_2_SERVER | SP_PROT_TLS1_2_CLIENT,
+            SP_PROT_TLS1_3 = SP_PROT_TLS1_3_SERVER | SP_PROT_TLS1_3_CLIENT,
+            SP_PROT_NONE = 0x0
+        }
+
+        private static string ToFriendlyName(this NativeProtocols protocol)
+        {
+            string name;
+
+            if (protocol.HasFlag(NativeProtocols.SP_PROT_TLS1_3_CLIENT) || protocol.HasFlag(NativeProtocols.SP_PROT_TLS1_3_SERVER))
```
benrr101, line 859: Two things:
* Unfortunately I've done a benchmark to compare `HasFlag` with using match for comparison and `HasFlag` is about 4x slower.
* Isn't checking for `SP_PROT_TLS1_3_CLIENT` or `SP_PROT_TLS1_3_SERVER` the same as checking for `SP_PROT_TLS_1_3`?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserHelperClasses.cs
```diff
@@ -791,14 +812,100 @@ private static string ToFriendlyName(this SslProtocols protocol)
...
+            SP_PROT_SSL2 = SP_PROT_SSL2_SERVER | SP_PROT_SSL2_CLIENT,
+            SP_PROT_SSL3 = SP_PROT_SSL3_SERVER | SP_PROT_SSL3_CLIENT,
+            SP_PROT_TLS1_0 = SP_PROT_TLS1_0_SERVER | SP_PROT_TLS1_0_CLIENT,
+            SP_PROT_TLS1_1 = SP_PROT_TLS1_1_SERVER | SP_PROT_TLS1_1_CLIENT,
+            SP_PROT_TLS1_2 = SP_PROT_TLS1_2_SERVER | SP_PROT_TLS1_2_CLIENT,
+            SP_PROT_TLS1_3 = SP_PROT_TLS1_3_SERVER | SP_PROT_TLS1_3_CLIENT,
+            SP_PROT_NONE = 0x0
+        }
+
+        private static string ToFriendlyName(this NativeProtocols protocol)
```
benrr101, line 855: Looking at this big method, I'm concerned that it's very similar to the netcore implementation, and I'm worndering if there's a better way to combine them.

benrr101 (commented):
I'm most of the way on board with this and ready to take it. However, I'd like to discuss the friendly name method first.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserHelperClasses.cs
edwardneal: It's almost identical - and awkwardly, it's only used to generate a warning about outdated TLS versions in the trace logs. `SslProtocols` actually has identical values to the ones returned by the SNI, I'm pretty sure the SNI's just passing them through from the SSL security package. .NET Framework doesn't have this enum, but I'm pretty sure that I can define it and merge this fully.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserHelperClasses.cs
edwardneal: Yes - I'm pretty sure that `HasFlag` was tweaked in .NET Core to make it faster. I've eliminated that code path in the merge.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserHelperClasses.cs
benrr101: Awesome! Thanks!

benrr101:
@edwardneal thanks for your help on this - sorry it took so long to get  it addressed. I'm on the merge project, so hopefully we can move things along faster now.

benrr101 (approved):

edwardneal:
Thanks @benrr101, that's great news. I've reached a stopping point for new merge PRs until November, so I've got bandwidth to keep up with reviews as they proceed.

cheenamalhotra (approved):

Issue #1495, opened 2022-01-31
Managed Identity Access Token is refreshed very often


kevinharing:
### Describe the bug
I recently started using managed identity authentication functionality from SqlClient after using our own access token caching implementation, but I'm seeing a lot of random token refreshes and I'm wondering if this is expected. Below is a screenshot from app insights showing the refresh calls over a period of 24 hours of one app instance.

Before:
![image](https://user-images.githubusercontent.com/1201599/151762808-d7476b0c-4150-43e6-8922-767641dcead1.png)

After:
![image](https://user-images.githubusercontent.com/1201599/151761409-036d7c81-5ca5-46d1-866b-e89b49161cf1.png)

I would expect the token to be cached around 24 hours before a refresh is triggered if the token lifetime is 24 hours. What I am experiencing though is that it refreshes at a seemingly random interval.

This behavior is slowing down a bunch of our requests as you can probably imagine.

One other thing I noticed is calls to the below endpoints when the token is refreshed. We did not see these pop up before the change.

- GET crl3.digicert.com/Omniroot2025.crl
- GET mscrl.microsoft.com/pki/mscorp/crl/Microsoft%20RSA%20TLS%20CA%2001.crl

Currently I'm considering moving back to our own implementation, because this is just plain dumb behavior.

### To reproduce
Just run a application with using the below segment in the connection string.

```
Authentication=Active Directory Managed Identity
```

### Expected behavior
Refresh the token just before the token expires (5 minutes before or something).

### Further technical details
Microsoft.Data.SqlClient version: 4.0.1
.NET target: 6.0.1
SQL Server version: Azure SQL Server
Operating system: aspnet:6.0-bullseye-slim

JRahnama:
@kevinharing, a token should be refreshed in about 10 hours (5-10 minutes before expiration time) and it would be cached to make it easier to use. Is that happening sooner than that time for you? 

Even if a token is acquired manually when it goes to the pool when it is re-used (after 10 hours) server checks credentials and it may throw token the expired exception.

kevinharing:
> @kevinharing, a token should be refreshed in about 10 hours (5-10 minutes before expiration time) and it would be cached to make it easier to use. Is that happening sooner than that time for you?

A token should be refreshed 5-10 minutes before expiration time indeed (not 10 hours specifically as the token lifetime is configurable is AAD). As can be seen on the screenshots, this is happening way more often than once every 10 hours.

[...]

PR #2380, opened 2024-03-02
Add | Cache TokenCredential objects to take advantage of token caching

David-Engel:
By caching DefaultAzureCredential, ManagedIdentityCredential, ClientSecretCredential, and WorkloadIdentityCredential objects, we can take advantage of token caching in the underlying Azure.Identity implementation.

This should particularly reduce token request rates in Managed Identity environments and eliminate or reduce instances of throttling from the managed identity token endpoint.

This would address #1495.


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ActiveDirectoryAuthenticationProvider.cs
```diff
@@ -477,6 +518,30 @@ private static byte[] GetHash(string input)
...
+            }
+
+            for (int i = 0; i < a1.Length; i++)
+            {
+                if (a1[i] != a2[i])
+                    return false;
+            }
+
+            return true;
+        }
```
JRahnama, line 543: I understand this is still a WIP, but I could not help myself from looking into optimizing that for loop. I think we can benefit from more modern C# features, since we are not supporting netstandard anymore. By using [ReferenceEquals](https://learn.microsoft.com/en-us/dotnet/api/system.object.referenceequals?view=net-8.0) and `Span<T>`, we can simplify and optimize the array comparison. This is just an untested comment, but I thought it would be good if we could implement it.

We could use `ReferenceEquals` to handle cases where both arrays are `null` or having same memory reference. This will simplify the code and make it more concise. At the end, our optimized implementation would resemble the following:
```C#
private static bool AreEqual(byte[] a1, byte[] a2)
{
    if (ReferenceEquals(a1, a2))
    {
        return true;
    }

    if (a1 is null || a2 is null)
    {
        return false;
    }

    if (a1.Length != a2.Length)
    {
        return false;
    }

    return a1.AsSpan().SequenceEqual(a2.AsSpan());
}
```

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ActiveDirectoryAuthenticationProvider.cs
```diff
@@ -454,16 +442,69 @@ public Task<Uri> AcquireAuthorizationCodeAsync(Uri authorizationUri, Uri redirec
...
+                    tokenCredentialInstance = CreateTokenCredentialInstance(tokenCredentialKey, secret);
+                    s_tokenCredentialMap.TryAdd(tokenCredentialKey, tokenCredentialInstance);
+                }
+                finally
+                {
+                    s_tokenCredentialMapModifierSemaphore.Release();
+                }
+            }
+
+            return await tokenCredentialInstance._tokenCredential.GetTokenAsync(tokenRequestContext, cancellationToken).ConfigureAwait(false);
```
DavoudEshtehari, line 505: All the method calls uses `ConfigureAwait(false)` and this is not necessary, which is my preference as a helper method. Otherwise you can keep this and amend the method calls.
```suggestion
            return await tokenCredentialInstance._tokenCredential.GetTokenAsync(tokenRequestContext, cancellationToken);
```

DavoudEshtehari (approved):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ActiveDirectoryAuthenticationProvider.cs
David-Engel: Done

JRahnama (approved):

JRahnama:
@David-Engel can you address the conflict that we can merge this please?

PR #1056, opened 2021-04-28
Add multi-user key store provider registration support

johnnypham:
These changes add a public method to register providers into an instance cache at the command-level, allowing users to maintain their own set of providers in the same application. 

The current column encryption key (CEK) cache is global so a new CEK cache and signature verification cache is added at the `SqlColumnEncryptionAzureKeyVaultProvider` level. Each `SqlColumnEncryptionAzureKeyVaultProvider` is linked to a single `TokenCredential`, which serves as the user identity.

   ```csharp
public class SqlCommand
{    
          // Added API: registers into command-level key store provider cache
          public void RegisterColumnEncryptionKeyStoreProvidersOnCommand(IDictionary<string, SqlColumnEncryptionKeyStoreProvider> customProviders);
}
```

PR #2383, opened 2024-03-02
Merge AAsyncCallContext, cleanup for SqlDataReader merge

edwardneal:
The main goal here is to simplify the future merge of SqlDataReader, which is why a lot of the changes are so superficial - they're to clean up the diff between .NET Core and Framework. There are a few functionality changes in the class between the two libraries, and I want to make them a little clearer for the future.

As part of this, AAsyncCallContext moves out of .NET Core library and into the shared components. At the end of this PR, the only differences between .NET Core and Framework's SqlDataReader are:
* .NET Framework makes fairly extensive use of constrained execution regions, which I don't think .NET Core supports
* .NET Framework handles OutOfMemoryExceptions, StackOverflowExceptions and ThreadAbortExceptions by closing the underlying SqlConnection
* Certain pieces of information are cached at the connection level, rather than the SqlDataReader level
* Small changes to the way that backwards compatibility with old versions of SQL Server is handled
* One possible bug in GetValues and GetSqlValues when a column in the result set is hidden
* Another possible bug when .NET Core uses SqlDataReader to read an Always Encrypted column
* .NET Core implements IDbColumnSchemaGenerator, when .NET Framework does not

This PR just cleans up the diff so that they can be picked apart more easily.


src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlDataReader.cs
```diff
@@ -1441,6 +1438,11 @@ override public IEnumerator GetEnumerator()
         }
 
         /// <include file='../../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlDataReader.xml' path='docs/members[@name="SqlDataReader"]/GetFieldType/*' />
+#if NET6_0_OR_GREATER
```
roji, line 1441: Rather than adding #if everywhere, another technique would be to just to copy the definition of the [DynamicallyAccessedMembers] attribute into the project, doing #if over that for old TFMs. Of course it won't actually do anything there, but it saves you from having to do #if in each and every place where the attribute is used.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlDataReader.cs
```diff
@@ -162,7 +162,7 @@ override public int Depth
             {
                 if (this.IsClosed)
                 {
-                    throw ADP.DataReaderClosed("Depth");
```
Wraith2, line 165: In this and other calls to DataReaderClosed why have you removed the parameter?

edwardneal: `ADP.DataReaderClosed` has the `CallerMemberName` attribute attached to its parameter, so specifying a value isn't necessary in most cases (with exceptions for TryNextResult, TryReadInternal and TryReadColumn.)

It's one of the superficial changes which allow a cleaner diff between netfx and netcore - without these changes, SqlDataReader's diff is -1067/+630, with many of the changed areas also surrounding actual functionality changes. With the changes, that diff is -624/+165. The post-PR diff is also easier to read.

Wraith2:
> .NET Framework makes fairly extensive use of constrained execution regions, which I don't think .NET Core supports 

Ensure that they are still present in the merged version. They will not do anything on netcore and often the compiler will entirely remove the calls but keep the code structure there because it can be important on netfx even when not running in-process in sql server.

> Small changes to the way that backwards compatibility with old versions of SQL Server is handled

Where possible to not remove compatibility code. For example sql 2005 is no longer supported but if we happen to be able to connect to one of those servers and the code needed to do so it not causing a bug then keep it. Only remove extended/historic compatibility if we need to do so.

>One possible bug in GetValues and GetSqlValues when a column in the result set is hidden

Can you raise an issue with an example of this? I reworked the hidden column implementation a few years ago and I thought it was correct. If it isn't then it needs to be fixed.

> Another possible bug when .NET Core uses SqlDataReader to read an Always Encrypted column

Can you raise an issue with an example or show the code and explain why it would fail. AE is hard to understand and debug.

> .NET Core implements IDbColumnSchemaGenerator, when .NET Framework does not

It looks like that interface is supported in netfx https://apisof.net/catalog/5bfef168-0a06-2fbe-8a07-a16f198d6019 we should merge the support into shared.

cheenamalhotra:
> It looks like that interface is supported in netfx https://apisof.net/catalog/5bfef168-0a06-2fbe-8a07-a16f198d6019 we should merge the support into shared.

IDbColumnSchemaGenerator is available from dotnet/runtime for .NET Core, so this should not be added by us: [IDbColumnSchemaGenerator](https://github.com/dotnet/runtime/blob/69340eb125312dd4102ad9af332d4f445aadc25c/src/libraries/System.Data.Common/src/System/Data/Common/IDbColumnSchemaGenerator.cs#L6)

edwardneal:
Thanks. Most of these are notes which I was going to chase down after this PR is merged, but the context you've added helps. I understand your point with regard to CERs.

> > Small changes to the way that backwards compatibility with old versions of SQL Server is handled
> 
> Where possible to not remove compatibility code. For example sql 2005 is no longer supported but if we happen to be able to connect to one of those servers and the code needed to do so it not causing a bug then keep it. Only remove extended/historic compatibility if we need to do so.

That's fine - the changes relate to some criteria being gated behind a version check for SQL 2005 in .NET Framework and SQL 2008 in .NET Core. I've not yet checked to look any further than that into whether it's driven by a factual correction (i.e. SQL 2005 doesn't actually support it) or part of a drop in support.

Is it correct to say that SQL 2000 should have support actively removed, SQL 2005 is unsupported (but not worth removing) and SQL 2008+ is supported? The .NET Framework project of SqlClient still supports SQL 2000, so this determines whether we should drop that support as part of the merge, add it back to .NET Core or put it all behind conditional compilations. The third option seems a little strange to me - if it's good enough to keep in .NET Framework then it makes sense that we could open it up to .NET Core in addition; if it's old enough to remove from .NET Core, then why allow it to linger in .NET Framework.

> 
> > One possible bug in GetValues and GetSqlValues when a column in the result set is hidden
> 
> Can you raise an issue with an example of this? I reworked the hidden column implementation a few years ago and I thought it was correct. If it isn't then it needs to be fixed.

I can, but there's not much to see and no user impact. PR #521 lazily initialized the hidden column map, and PR #1443 backported it to .NET Framework. When it did so, it missed out an array index - but I don't think code ever reaches that point because the hidden columns seem to be at the end of the column set and the iterator exits before it reaches them.

The result is that .NET Core's GetSqlValues contains:
```csharp
for (int i = 0; i < copyLen; i++)
{
    values[i] = GetSqlValueInternal(_metaData.GetVisibleColumnIndex(i));
}
```
and .NET Framework's GetSqlValues contains:
```csharp
for (int i = 0; i < copyLen; i++)
{
    values[_metaData.GetVisibleColumnIndex(i)] = GetSqlValueInternal(i);
}
```

GetValues is very similar, and I think .NET Core is correct. It's a two-line change. I can raise it as a bug and make a two-line PR, or simply roll it into the full SqlDataReader merge.

> 
> > Another possible bug when .NET Core uses SqlDataReader to read an Always Encrypted column
> 
> Can you raise an issue with an example or show the code and explain why it would fail. AE is hard to understand and debug.

The AE point is probably similar, but I'm not completely sure. In TryReadColumnData, .NET Framework passes `_command` to _parser.TryReadSqlValue while .NET Core does not. This means that if something has called `SqlCommand.RegisterColumnEncryptionKeyStoreProvidersOnCommand` and not passed those into the SqlConnection too, the AE decryption process won't have access to those providers. PR #1056 implemented this, and I think it may have missed a few places (in TryReadColumnData and TryReadColumnInternal for both .NET Core and Framework.) I need to set up an AE environment to understand and test this to see if the difference actually generates a bug, which might take a while to get to grips with.

David-Engel (approved):

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlDataReader.cs
```diff
@@ -369,17 +369,18 @@ internal virtual SmiExtendedMetaData[] GetInternalSmiMetaData()
...
+                        metaDataReturn[returnIndex] =
+                            new SmiQueryMetaData(
                                                         colMetaData.type,
                                                         length,
                                                         colMetaData.precision,
                                                         colMetaData.scale,
                                                         (null != collation) ? collation.LCID : _defaultLCID,
                                                         (null != collation) ? collation.SqlCompareOptions : SqlCompareOptions.None,
                                                         colMetaData.udt?.Type,
-                                                        false,  // isMultiValued
```
benrr101, line 380: Actually instead of changing the indent, could you use the named parameters here? 
Tbh, I kinda forget the rules for using those, ie, can they be omitted in later arguments or once they start do they have to be used until the end?

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlDataReader.cs
```diff
@@ -497,6 +499,10 @@ internal void Bind(TdsParserStateObject stateObj)
         // Fills in a schema table with meta data information.  This function should only really be called by
         // UNDONE: need a way to refresh the table with more information as more data comes online for browse info like
         // table names and key information
+#if NET6_0_OR_GREATER
```
benrr101, line 502: I think these conditionals work better as `#if !NETFRAMEWORK`. My main rationale for it is that `NET6_0_OR_GREATER` implies to me that once we drop support for net6, we can remove the conditional, but that's not actually the case since it is still needed for netframework support.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlDataReader.cs
```diff
@@ -599,189 +605,189 @@ internal DataTable BuildSchemaTable()
...
+                if (!string.IsNullOrEmpty(col.baseColumn))
                 {
-                    schemaRow[BaseColumnName] = col.baseColumn;
+                    schemaRow[baseColumnName] = col.baseColumn;
                 }
-                else if (!ADP.IsEmpty(col.column))
+                else if (!string.IsNullOrEmpty(col.column))
                 {
-                    schemaRow[BaseColumnName] = col.column;
+                    schemaRow[baseColumnName] = col.column;
```
benrr101, line 790: NB: Thank you for these changes. I'd be losing my mind if I had to work on this code trying to figure out each time if it was referencing a property or a variable.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlDataReader.cs
benrr101: I personally prefer not making a no-op `DynamicallyAccessedMembersAttribute`. I agree it would make the code on this side more readable, but it becomes unclear what's actually happening. Imho, we shouldn't add no-op attributes.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlDataReader.cs
```diff
@@ -1720,7 +1728,7 @@ virtual public XmlReader GetXmlReader(int i)
                 if (_data[i].IsNull)
                 {
                     // A 'null' stream
-                    return SqlTypeWorkarounds.SqlXmlCreateSqlXmlReader(new MemoryStream(new byte[0], writable: false), closeInput: true, async: false);
+                    return SqlTypeWorkarounds.SqlXmlCreateSqlXmlReader(new MemoryStream(Array.Empty<byte>(), writable: false), closeInput: true, async: false);
```
benrr101, line 1731: Could we just use `Stream.Null` here?

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlDataReader.cs
```diff
@@ -1963,11 +1973,11 @@ private bool TryGetBytesInternal(int i, long dataIndex, byte[] buffer, int buffe
                     // note that since we are caching in an array, and arrays aren't 64 bit ready yet,
                     // we need can cast to int if the dataIndex is in range
                     if (dataIndex < 0)
```
benrr101, line 1975: `{ }` here?

benrr101: (and other places?)

benrr101 (approved):
I'm totally onboard with these changes. They look great to me.

benrr101:
@edwardneal If you can resolve the merge conflicts (probably mostly caused by me 😅) and take a look at the handful of comments I made on the PR, I'd be happy to get this merged! 🚀

edwardneal:
Thanks. I've resolved the conflicts, with good local results. I'm going to let it run through CI as-is, and then I'll work through your review comments (which I pretty much completely agree with.)

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlDataReader.cs
edwardneal: I don't think we can - `Stream.Null` doesn't do anything when written to, while the returned MemoryStream will have `CanWrite` set to `false`.

benrr101: Ah, good point!

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
```diff
@@ -1367,7 +1369,7 @@ override public string GetName(int i)
         }
 
         /// <include file='../../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlDataReader.xml' path='docs/members[@name="SqlDataReader"]/GetProviderSpecificFieldType/*' />
-#if NET8_0_OR_GREATER
+#if !NETFRAMEWORK && !NET6_0
```
benrr101, line 1372: Hmm.. I think this one should probably stay as `NET8_0_OR_GREATER`. I probably am making a mess of these suggestions here, but my thought is that if it's differentiating between versions of netcore then it's fine to leave as `NETx_OR_GREATER`, but if it's to differentiate between netfx and netcore, then I think it's better as `NETFRAMEWORK`/`!NETFRAMEWORK`

But a better question for this one might be - why was it NET**8**_0_OR_GREATER in the first place?

benrr101 (commented):
This is awesome @edwardneal - Got one more request and then we'll get it merged 😅

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
edwardneal: That makes sense, thanks - I'll use `!NETFRAMEWORK && NET8_OR_GREATER`, so it doesn't get removed by accident when we remove .NET 6.0 support.

The annotation's different because this overrides DbDataReader.GetProviderSpecificFieldType. Trimming annotations need to match between base and derived classes, and the annotation on the base class method changed between 6.0 and 8.0.

benrr101:
Thanks @edwardneal !

PR #2390, opened 2024-03-06
Merged PoolWaitHandles, DbConnectionPoolIdentity, removed DbBuffer

edwardneal:
Contributes to #1261.

This is a little more involved than the previous merges. PoolWaitHandles and DbConnectionPoolIdentity in .NET Framework currently use direct PInvoke to wait on named objects and get the Windows identity details. I've merged these with their .NET Core counterparts, which behave the same way.

Doing so also removes the last reference to DbBuffer, which was only present in the .NET Framework project. With those references gone, I've removed this too. This contributes to three quarters of the deletions - the PR's a fair bit smaller than it looks!


David-Engel (approved):

DavoudEshtehari:
@edwardneal Can you address the conflicts, please?

edwardneal:
Thanks @DavoudEshtehari. I've handled the merge conflicts, and the test failure seems to be unrelated to me.

DavoudEshtehari (approved):

DavoudEshtehari:
Given these changes, I believe the `DbConnectionPool.NetCoreApp.cs` file can be merged with the other files to make it simpler to follow. It's not a critical change request, but it would be nice to have.

edwardneal:
I agree. I've moved the method into DbConnectionPool.cs and guarded it behind `#if NET6_0_OR_GREATER` to make the later merge simpler.

wailn3-a11y (approved):

Issue #2394, opened 2024-03-07
Server principle name lookup does not include port number in the name


trexx:
### Describe the bug
Upon updating from 5.1.5 to 5.2.0, our applications running on Linux fail to connect to our databases using Integrated Security / Kerberos with an error:
` Microsoft.Data.SqlClient.SqlException (0x80131904): The target principal name is incorrect. Cannot generate SSPI context.`

This is with a connection string such as:
`"Data Source=server.domain.local;Integrated Security=true"`

`server.domain.local` here is a CNAME which resolves to `server001.domain.local`

In version 5.1.5 I can see it resolves the SPN in 2 different ways. It first resolves the SPN without the port number, and if that fails it retries with the port number in the SPN.
1st try   -> `MSSQLSvc/server001.domain.local@DOMAIN.LOCAL`
2nd try  -> `MSSQLSvc/server001.domain.local:1433@DOMAIN.LOCAL` < Lookup successful 

In version 5.2.0, it only resolves the SPN without the port number, and then performs no more searches causing connectivity to fail.
1st try   -> `MSSQLSvc/server001.domain.local@DOMAIN.LOCAL`

Our workaround has been to specify the port number in the connection string.
`"Data Source=server.domain.local,1433;Integrated Security=true"`

Is this change of behaviour intentional or a result of a bug? We're not sure if this will be fixed in SqlClient, or if we should update our SPNs, or whether we should start including port numbers in all our connection strings.

### Further technical details
Microsoft.Data.SqlClient version: 5.2.0
.NET target: 8.0
Operating system: Linux Container

JRahnama:
@trexx We are aware of this issue in v5.2.0. Currently, we are investigating the code change that caused this behavior change. Could you please test with v5.2.0-preview5 and see if it works as you expect?

ErikEJ:
@JRahnama could it be this?

trexx:
@JRahnama The behaviour and result is the same with v5.2.0-preview5; The SPN its searching for does not include the port number and connectivity fails.

David-Engel:
@trexx To narrow it down even further, can you try the earlier previews? I suspect 5.2.0 preview 4 will start working for you, but if it doesn't it will be good to know which one contains the behavior change.

trexx:
Hi @David-Engel I've been able to narrow it down to 5.2.0-preview2.23159.1 when the SPN behaviour is that same as 5.1.5. Preview versions after this fail.

[...]

PR #2240, opened 2023-11-23
Fix | Linux SPN port number using named instance and Kerberos authentication does not return port#

arellegue:
Fixes Linux SPN to return port number in managed SNI when using named instance and Kerberos authentication.

Issue #2187, opened 2023-10-17
MDS on Linux sends SPN with instance name instead of port


David-Engel:
### Describe the bug

MDS' managed SNI implementation is sending the SPN in the format MSSQLSVC/SERVERNAME.FQDN:INSTANCENAME instead of MSSQLSVC/SERVERNAME.FQDN:PORT when making TCP connections to named instances.

### To reproduce

Connect to a named instance via TCP, managed SNI, and Kerberos auth. Using a network capture, you can see MDS gets a ticket using the named instance instead of the port number.

### Expected behavior

MDS should use the MSSQLSVC/SERVERNAME.FQDN:PORT format for consistency with other drivers and aligning with the [docs](https://learn.microsoft.com/en-us/sql/database-engine/configure-windows/register-a-service-principal-name-for-kerberos-connections?view=sql-server-ver16#Client):

> User-specified SPNs are supported in client drivers. However, if an SPN isn't provided, it is generated automatically based on the type of a client connection. For a TCP connection, an SPN in the format `MSSQLSvc/FQDN:[<port>]` is used for both the named and default instances.

> For named pipes and shared memory connections, an SPN in the format `MSSQLSvc/<FQDN>:<instancename>` is used for a named instance and `MSSQLSvc/<FQDN>` is used for the default instance.

### Further technical details
Microsoft.Data.SqlClient version: latest 5.1.1
.NET target: .NET (non Framework)
SQL Server version: All
Operating system: Linux (or managed SNI on Windows)

David-Engel:
Filing this from an internal issue.

PR #2395, opened 2024-03-08
Regression | Revert PR #2281 SNIProxy code change

JRahnama:
This PR reverts back the code change from #2281 to address regression in v5.2.0

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIProxy.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIProxy.cs
@@ -654,7 +654,7 @@ namespace Microsoft.Data.SqlClient.SNI
                 Port = port;
             }
             // Instance Name Handling. 
-            if (backSlashIndex > -1)
+            else if (backSlashIndex > -1)
             {
                 // This means that there is a part separated by '\'
                 InstanceName = tokensByCommaAndSlash[1].Trim();
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/InstanceNameTest/InstanceNameTest.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/InstanceNameTest/InstanceNameTest.cs
@@ -87,6 +87,7 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests
         }
 
 #if NETCOREAPP
+        [ActiveIssue("27824")] // When specifying instance name and port number, this method call always returns false
         [ConditionalFact(nameof(IsSPNPortNumberTestForTCP))]
         public static void PortNumberInSPNTestForTCP()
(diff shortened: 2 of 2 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIProxy.cs
```diff
@@ -654,7 +654,7 @@ private bool InferConnectionDetails()
                 Port = port;
             }
             // Instance Name Handling.
```
David-Engel, line 656: I'd revert the comments, too.
```suggestion
            // Instance Name Handling. Only if we found a '\' and we did not find a port in the Data Source
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIProxy.cs
```diff
@@ -654,7 +654,7 @@ private bool InferConnectionDetails()
                 Port = port;
             }
             // Instance Name Handling. 
-            if (backSlashIndex > -1)
+            else if (backSlashIndex > -1)
             {
                 // This means that there is a part separated by '\'
```
David-Engel, line 659:
```suggestion
                // This means that there will not be any part separated by comma.
```

David-Engel (approved):

esbenbjerre (approved):

DavoudEshtehari (approved):

DavoudEshtehari:
Can you link the related issues for the record?

ErikEJ:
Wonder if it would be possible to add a test to prevent this from regressing again?

JRahnama:
> Wonder if it would be possible to add a test to prevent this from regressing again?

I was thinking the same. I compared it with SDS and will start by adding some tests. However, I need to understand the reason behind #2394 first. It seems the underlying cause of that issue might be a bit different.

arellegue (approved):

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/InstanceNameTest/InstanceNameTest.cs
```diff
@@ -87,6 +87,7 @@ public static void ConnectManagedWithInstanceNameTest(bool useMultiSubnetFailove
         }
 
 #if NETCOREAPP
+        [ActiveIssue("27824")] // When specifying instance name and port number, this method call always returns false
```
cheenamalhotra, line 90: Does that mean issue https://github.com/dotnet/SqlClient/issues/2187 should be re-opened too?

JRahnama: PR #2240 was the one that addressed #2187 and I am not sure why the code change was needed when the test was fixed.

PR #2398, opened 2024-03-09
Merged SqlFilestream

edwardneal:
Relates to issue #1261.

This merges SqlFileStream between .NET Core and Framework. In the process, I've also implicitly dropped a code path to set the process-level or thread-level exception mode. This code path would set the thread-level exception mode on Windows 7 or higher, and the process-level exception mode on Windows XP or below; I've discarded the latter path.

SqlFileStream depends on a handful of PInvokes and a PathInternal class. I've merged these too, trimming unused methods where applicable. The removal of the former instances of these accounts for 70-80% of the removals.

I've also added ReadAsync and WriteAsync methods to SqlFileStream in .NET Core. The underlying FileStream supports these, so it doesn't do any harm to pass them through.


benrr101:
Hey @edwardneal to be honest I started on merging this class before seeing your PR (I guess our list of class merges with open PRs wasn't completely up to date). Since it looks like there's a lot of conflicts in this PR now, would it be alright if we close this one and look at the one(s) I put together? I'd also really appreciate your feedback on them, since I see you made some changes to this class that I may have overlooked.

edwardneal:
Yes - no problem at all. I was hoping that the combination of our PInvoke moves/merges would tidy up a lot of the diff here, and wasn't looking forward to the merge which resulted from that. Your PRs already have that, so it makes sense to proceed with one of those.

PR #2402, opened 2024-03-12
Move Platform-Specific Files Out Of Common Folders

benrr101:
**Description**:  While looking over the codebase, I noticed there are a handful of platform-specific source files that were placed in the common source folder. This seems like a mistake and would cause confusion when trying to edit these files. This PR makes the following changes:

* Moves the following files from the common source folder to their respective platform project folders
  * `src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBatchCommand.Net8OrGreater.cs`
    * References to xml docs fixed
    * Wrapped in `#if NET8_0_OR_GREATER` to remove need for conditional inclusion with project
  * `src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SmiEventSink_Default.netfx.cs`
  * `src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SqlDataRecord.netcore.cs`
    * Reference to xml doc fixed
  * `src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SqlDataRecord.netfx.cs`
    * Reference to xml doc fixed
  * `src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/ValueUtilsSmi.netfx.cs`
* Remove conditional inclusion of the following files from netcore project
  *  `src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlTypes/SqlTypeWorkarounds.netcore.cs`
    * Wrapping file with `#if !NET8_0_OR_GREATER`
  * `src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlClientFactory.NetCoreApp.cs`
    * Wrapping file with `#if NET6_0_OR_GREATER`

**Testing**:
* Automated tests still pass as expected


cheenamalhotra:
Our goal here is to merge code-bases from netfx and netcore into a common src folder, and someday, build SqlClient from only 1 csproj file.. given that in mind, this PR is doing the opposite, I don't think we need this change IMO.

ErikEJ:
@cheenamalhotra agree, this is the wrong direction

benrr101:
@cheenamalhotra @ErikEJ 
Ok, after some offline conversation with Cheena, I understand that the files in the common src folder are considered "finalized" for what can and cannot be merged into a single file, with the end goal being all files to live in a common project. Anything still in the separate projects are unfinalized as such. Thus, moving finalized files to the separate projects is working backwards.

I'll back this PR out

Issue #2211, opened 2023-11-07
Emit OpenTelemetry metrics


roji:
OpenTelemetry has become the de-facto standard for distributed, cross-platform tracing; it is being rapidly adopted both inside the .NET ecosystem and outside. OTel defines semantic conventions for how a database client emits metrics, which are aggregated, numeric counters about the state and performance of the driver (e.g. number of commands executed, idle/busy connections in the pool, etc.).  [these specifications](https://opentelemetry.io/docs/specs/semconv/database/database-metrics/) are currently in experimental state, but stabilization is likely start happening soon.

Note that unlike with [tracing](https://github.com/dotnet/SqlClient/issues/2210), AFAIK there's no way to get data like this out of SqlClient at the moment. This sort of data can be invaluable in diagnosing performance problems (e.g. by inspecting real-time connection pool state) and tracking what's going on.

JRahnama:
Is this a duplicate of #2210 ?

gfoidl:
...but the link in the description 

> Note that unlike with [tracing](#2210),

links to this page. Should link to https://github.com/dotnet/SqlClient/issues/2210

edwardneal:
SqlClient actually already emits some metrics, but their reporting is a little spotty. On .NET Framework in Windows, every process which uses the library has a set of performance counters (listed [here](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/ProviderBase/DbConnectionPoolCounters.cs).) On .NET Core and .NET Standard 2.1 and above, Microsoft.Data.SqlClient.EventSource has a number of event counters (listed [here](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlClientEventSource.NetCoreApp.cs).) Although the counter names are slightly different, their meanings are the same:

| Performance counter | Event counter | Unit | Description |
| - | - | - | - |
| (none) | active-hard-connections | count | Actual active connections current made to servers |
| HardConnectsPerSecond | hard-connects | count / sec | The number of actual connections per second that are being made to servers |
| HardDisconnectsPerSecond | hard-disconnects | count / sec | The number of actual disconnects per second that are being made to servers |
| (none) | active-soft-connects | count | Active connections retrieved from the connection pool |
| SoftConnectsPerSecond | soft-connects | count / sec | The number of connections we get from the pool per second |
[...]

PR #2409, opened 2024-03-16
Proof of concept: OpenTelemetry metrics

edwardneal:
Contributes to #2211.

This adds an initial framework for updated telemetry using .NET's new metrics API. Leaving it in draft because I'm conscious that the issue hasn't yet had design approval from the SqlClient team, but this should be a reasonably generic and performant base to feedback on build on for the metrics design. It also provides a spot to put the eventual tracing APIs (when those are eventually standardised) and allows us to remove the project-specific connection pool counters.

I've added instrumentation for the metrics laid out in the OTel specifications, and for the existing .NET Core event counters. In most cases, this isn't wired up to anything.

From some very basic benchmarking, this has no measurable GC or CPU overhead. I don't think it has any explicit lock contention either - just a bundle of Interlocked.Increment/.Decrement calls. The OpenTelemetry SDK is another matter!

Although there's going to be some memory overhead here, it should be minimal. Since most of the tags only vary at the connection pool group level, I'm planning to cache them there and allow SqlConnection and SqlCommand to draw on that reference (unless pooling is disabled, or a specific SqlConnection wants to specify a custom tag.)

There are two trivial points, one piece of additional work, and one specification design decision which I'm not sure I agree with.
* The documentation for `Counter<T>` states that it's only supposed to increase. I'm decrementing this type of counter because we currently have to target .NET Standard and .NET 6.0, and `UpDownCounter<T>` is only available in .NET 7+.
* I've left integration points for custom tags for a connection, and for the disablement of the legacy metrics. There's space for a feature request there, but I'm leaving it alone for now.
* Measuring the connection's usage time, open time, wait time might need some rework of `SqlStatistics`. Since the collection overhead of SqlStatistics is practically nil, I'd like to always enable such and push the data into the corresponding metrics. The precision of `DateTime.UtcNow` (used by SqlStatistics) is 0.5ms-15ms, so this might need to start using stopwatches instead, particularly for local or LAN SQL Server instances. 

The specification design decision surrounds the use of the `pool.name` tag. In cases where the connection string uses Windows authentication (or, in future, perhaps some other kind of process-/thread-level identity impersonation mechanism) it's possible for two connections to have the same connection string, but to be in different connection pools because they've been opened under an impersonation context. The specification rolls both of these up by their connection string. I'm thinking about whether or not to add a `db.user` tag to handle this - it'd make it clearer that multiple pools are involved, but might break compatibility with downstream clients which expect one metric per `pool.name`.

Tagging @roji and @JRahnama for any comments on the topic, direction or methodology. There aren't that many ways to implement these metrics, but I'd prefer to have some feedback from the SqlClient team so that I'm not surprising anyone with an x000-line PR for merge.

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Telemetry/DbConnectionFactoryTelemetry.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Telemetry/DbConnectionFactoryTelemetry.cs
@@ -1,0 +1,60 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+using System.Threading;
+using Microsoft.Data.ProviderBase;
+
+namespace Microsoft.Data.SqlClient.Telemetry
+{
+    // Stores telemetry for a single DbConnectionPoolGroup. Maintained by the connection pool group itself
+    internal sealed class DbConnectionPoolTelemetry
+    {
+        private readonly DbConnectionPoolGroupOptions _options;
+        private long _poolCount;
+
+        public long TotalMaxPoolSize => _poolCount * (long)_options.MaxPoolSize;
+
+        public long TotalMinPoolSize => _poolCount * (long)_options.MinPoolSize;
+
+        public DbConnectionPoolTelemetry(DbConnectionPoolGroupOptions poolGroupOptions)
+        {
+            _options = poolGroupOptions;
+        }
+
+        public void ReportNewPool()
+        {
+            Interlocked.Increment(ref _poolCount);
+        }
+
+        public void ReportRemovedPool()
+        {
+            Interlocked.Decrement(ref _poolCount);
+        }
+
+        public void Reset()
+        {
+            Interlocked.Exchange(ref _poolCount, 0);
+        }
+    }
+
+    internal struct DbConnectionFactoryTelemetry
+    {
+        public long TotalMinPoolSize { get; private set; }
+
+        public long TotalMaxPoolSize { get; private set; }
+
+        public DbConnectionFactoryTelemetry(DbConnectionPoolTelemetry telemetry)
+        {
+            TotalMinPoolSize = telemetry.TotalMinPoolSize;
+            TotalMaxPoolSize = telemetry.TotalMaxPoolSize;
+        }
+
+        public void AddConnectionPool(DbConnectionPoolTelemetry telemetry)
+        {
+            TotalMinPoolSize += telemetry.TotalMinPoolSize;
+            TotalMaxPoolSize += telemetry.TotalMaxPoolSize;
+        }
+    }
+}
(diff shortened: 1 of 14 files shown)


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Telemetry/DbConnectionFactoryTelemetry.cs
```diff
@@ -0,0 +1,60 @@
...
+
+namespace Microsoft.Data.SqlClient.Telemetry
+{
+    // Stores telemetry for a single DbConnectionPoolGroup. Maintained by the connection pool group itself
+    internal sealed class DbConnectionPoolTelemetry
+    {
+        private readonly DbConnectionPoolGroupOptions _options;
+        private long _poolCount;
+
+        public long TotalMaxPoolSize => _poolCount * (long)_options.MaxPoolSize;
```
Wraith2, line 17: since the options can come from user input is it worth guarding against overflow here and below?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Telemetry/MetricConstants.cs
```diff
@@ -0,0 +1,167 @@
...
+
+            public const string CommitTime = LibrarySpecificPrefix + "commit_time";
+
+            public const string RollbackTime = LibrarySpecificPrefix + "rollback_time";
+
+            public const string ActiveTime = LibrarySpecificPrefix + "active_time";
+        }
+    }
+
+    internal static class MetricAttributes
```
Wraith2, line 103: `MetricsAttributesNames`? The Attributes suffix implies that it contains attributes to me.

edwardneal: Yes, thanks. I spent a few minutes dithering between describing them as attributes or tags, before deciding to use tag-based terminology for consistency. I'll rename it and its counterpart to `MetricTagNames` and `MetricTagValues` respectively.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Telemetry/DbConnectionFactoryTelemetry.cs
edwardneal: I don't think one could be triggered. MaxPoolSize ultimately comes via SqlConnectionString.MaxPoolSize, which is defined as an int and has protection against negative values.

An overflow would be triggered if there were `uint.MaxValue` connection pools in the connection pool group with a MaxPoolSize of `int.MaxValue`, but the list of connection pools is stored as a ConcurrentDictionary in DbConnectionPoolGroup (which is limited to storing `int.MaxValue` entries.)

edwardneal:
Thanks - I've just updated this to allow the use of UpDownCounter.

I've added enough code to show how I'd expect the .NET Framework version of the library to work and to ensure it compiles, but I'm going to leave this in draft until the overall design can be reviewed.

PR #2411, opened 2024-03-17
serialise scale for vartime has been explicitly set by the user to zero

EamonHetherton:
revive old pull request #1381 that got stuck in limbo due to a code freeze

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlParameter.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlParameter.cs
@@ -583,7 +583,7 @@ namespace Microsoft.Data.SqlClient
             }
         }
 
-        private bool ShouldSerializeScale() => _scale != 0; // V1.0 compat, ignore _hasScale
+        private bool ShouldSerializeScale() => _scale != 0 || (GetMetaTypeOnly().IsVarTime && HasFlag(SqlParameterFlags.HasScale));
 
         /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlParameter.xml' path='docs/members[@name="SqlParameter"]/SqlDbType/*' />
         [
@@ -1509,8 +1509,7 @@ namespace Microsoft.Data.SqlClient
                 return ScaleInternal;
             }
 
-            // issue: how could a user specify 0 as the actual scale?
-            if (GetMetaTypeOnly().IsVarTime)
+            if (GetMetaTypeOnly().IsVarTime && !HasFlag(SqlParameterFlags.HasScale))
             {
                 return TdsEnums.DEFAULT_VARTIME_SCALE;
             }


Wraith2:
It wasn't stuck in limbo it was decided against, see https://github.com/dotnet/SqlClient/issues/1380#issuecomment-1332572899

The only way you're going to get the change implemented is with an appcontext opt-in, if you include that then it might be includable. You'll need to add tests. You'll need to change the documentation.

EamonHetherton:
> It wasn't stuck in limbo it was decided against, see [#1380 (comment)](https://github.com/dotnet/SqlClient/issues/1380#issuecomment-1332572899)
> 
> The only way you're going to get the change implemented is with an appcontext opt-in, if you include that then it might be includable. You'll need to add tests. You'll need to change the documentation.

Apologies, I'd assumed that this was blocked based on this https://github.com/dotnet/SqlClient/pull/1381#issuecomment-965810818

I'm really out of options for how to mitigate the performance impact of this short of moving off MSSQL or forking this library.  I really don't want to do either :(  

I'll chip away and attempt to make this change based on an app context switch, try to get some testing around it and update the docs and then and see if I can get some traction.  

Also, it looks like the failed checks are unrelated to this pull request as best that I can tell, I'm really not sure how to get the checks to pass.

EamonHetherton:
It looks like the explicit calling out of this behaviour in the docs for the previous library System.Data.SqlClient as referred to here: https://github.com/dotnet/SqlClient/issues/1380#issuecomment-963056195 is no longer in the docs for Microsoft.Data.SqlClient.  Therefore I don't think there are any doc changes to make as part of this pull request.  Still working on tests :)

EamonHetherton:
@Wraith2 pull request updated with context switch and tests.  I don't believe there are documentation updates required for the Scale parameter itself as noted above, but the existing context switches do appear in documentation here: https://learn.microsoft.com/en-us/sql/connect/ado-net/appcontext-switches?view=sql-server-ver16 but I'm not sure where that lives. Any guidance would be appreciated.

EamonHetherton:
@Wraith2 Incidentally I noticed that the UseMinimumLoginTimeout context switch changed default behaviour in the last refactor of LocalAppContextSwitches in November 2023 and now returns a default value of "false" though the triple slash summary still says it should have a default of "true". The documentation @ https://learn.microsoft.com/en-us/sql/connect/ado-net/appcontext-switches?view=sql-server-ver16 does not give any guidance on default value but it looks to me like this default value change was unintended, but I think you would be better placed to determine that.

ErikEJ:
Looks like the team proposal was to make the new behaviour opt-in, but it looks like it is opt out now?

EamonHetherton:
no, default is to retain existing behaviour

EamonHetherton:
> Can you open a new issue with the details on that default so it can be reviewed please.

issue #2415 with fix in #2416

EamonHetherton:
Is there anything more I can do to move this along? failing that is there anywhere for me to get access to the pull request nuget package that was built "Microsoft.Data.SqlClient.6.0.0-pull.107808"?

Wraith2:
No. You just wait for the team to get to it.

EamonHetherton:
>6 months later this is still waiting to be reviewed/approved....

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlParameter.cs
```diff
@@ -583,7 +583,21 @@ internal byte ScaleInternal
...
+        {
+            if (LocalAppContextSwitches.LegacyVarTimeZeroScaleBehaviour)
+            {
+                return ShouldSerializeScale_Legacy();
+            }
+            if (_scale != 0 || (GetMetaTypeOnly().IsVarTime && HasFlag(SqlParameterFlags.HasScale)))
+            {
+                return true;
+            }
+            return false;
```
cheenamalhotra, line 598:
```suggestion
            return _scale != 0 || (GetMetaTypeOnly().IsVarTime && HasFlag(SqlParameterFlags.HasScale));
```

cheenamalhotra (approved):
Please update PR description to include the behavior change and introduction of new App Context switch for documentation purposes.

EamonHetherton:
suggested documentation for new AppContext switch to update [this](https://learn.microsoft.com/en-us/sql/connect/ado-net/appcontext-switches?view=sql-server-ver16)



## Allow explicit zero scale for VarTime parameter types

[!INCLUDE [appliesto-netfx-netcore-netst-md](https://github.com/MicrosoftDocs/sql-docs/blob/4d066044cd0aa2593cb06ffea4d851c74cc34510/docs/includes/appliesto-netfx-netcore-netst-md.md)]

(Available starting with version 6.0.0)

When using parameters with `VarTime` datatypes (DateTime2, DateTimeOffset,  Time), the library will only emit the scale if it is non-zero regardless of the user explicitly setting the Scale=0. See https://learn.microsoft.com/en-us/dotnet/api/system.data.sqlclient.sqlparameter.scale?view=netframework-4.8.1#remarks

In order to emit a zero scale, you can now set the following AppContext switch to false:

```csharp
AppContext.SetSwitch("Switch.Microsoft.Data.SqlClient.LegacyVarTimeZeroScaleBehaviour", false);
```
and then explicitly set the parameter scale to zero
```csharp
var parameter = new SqlParameter
{
    ParameterName = "@p1",
    DbType = DbType.DateTime2,
    Scale = 0
};
```
which will emit the equivalent SQL parameter as
```sql
DECLARE @p1 datetime2(0)
```

cheenamalhotra (approved):

David-Engel (approved):

Issue #2323, opened 2024-01-26
$(IsUAPAssembly)


ErikEJ:
Is this still in use/built?

David-Engel:
It's not currently in use. It's left over from the SDS fork. Plans to support UAP/UWP in MDS were in the future plans. But I don't see them getting prioritized given that MS has said that UWP is no longer the future.

https://github.com/microsoft/WindowsAppSDK/discussions/1615
https://www.techradar.com/news/microsoft-is-finally-dropping-uwp-app-support-windows-app-sdk-heres-why

I also don't know if MDS needs to make any changes for the Windows App SDK, at this point. Or if it just works with it as-is.

PR #2425, opened 2024-03-22
Change | Simplify Conditional File Inclusion

benrr101:
**Description**: This PR is some prerequisite work for removing ref projects. Although it may not be strictly required for removing ref projects, it will increase my confidence when making the changes to remove ref packages. The netcore project has a bunch of conditional inclusion of files that in some cases overlap and in some cases are cancelled out or redundant due to changes to the supported version changes over the years. In addition, the newer version of .net support more precompiler constants for easy inclusion/exclusion of code based on .net version. (eg, `NET8_0_OR_NEWER`, `NETCOREAPP`). As such, it is no longer necessary to have conditional inclusion of files based on .net version - this can be pushed to `#if` wrappers in the files. This simplifies the different groups for including files to: !AnyOS, TargetsWindows, and TargetsUnix.

IsUAPAssembly is currently unaddressed with this PR since I'm not 100% sure what that flag implies. There seems to be some discussion about removing support for UAP anyways (#2323) 

Even though there is efforts underway to move to a single project that combines netfx and netcore, this change stands on its own. Using conditional inclusion of files in a project can lead to difficult to diagnose compile-time errors, often times manifesting as "did you forget a reference or using statement?" errors. These are difficult to track down because opening the file in the IDE will usually show no error - the file is right there! why would it not build!? Requiring digging deep to realize the file is part of the project, but not included in a certain build. By unconditionally including the files in the project, and moving the conditional to `#if`, the IDE will show whether or not the code within the `#if` is included based on which version of .net is being built. 

_Ref Project Removal Background_: Ref assemblies can be generated with the implementation projects if using a certain attribute in the proj, which should remove the need for a separate ref project. The current ref projects are built for AnyOS, and stored in the nuget pkg without specifying an OS. However, the implementation projects generate empty assemblies when built for AnyOS. Thus, when generating a ref assembly, it generates a empty ref assembly. To alleviate that, I'm proposing including the non-OS-specific files in the AnyOS build and using the ProduceOnlyReferenceAssembly attribute to only generate the reference assembly. Although initial tests with simply removing the AnyOS check seemed successful, I'm less confident in them because the different conditions on including files is complex. That's why this PR exists.

**Testing**:
* Projects still build
* Tests that are expected to pass still pass
* I did a comparison of all files included in the old project and the new project and the list is the same (mostly a sanity check)


src/Microsoft.Data.SqlClient/netcore/src/Common/src/Interop/Unix/System.Net.Security.Native/Interop.GssApiException.cs
```diff
@@ -2,9 +2,11 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-using Microsoft.Data;
+#if !NET8_0_OR_GREATER
```
JRahnama, line 5: I think this part of the code` System.Net.Security` is not being used on Unix when using Net7+. The driver is using the newly added API from #2063. Basically this part of the code could be deleted when support for net6 is over.

src/Microsoft.Data.SqlClient/netcore/src/Common/src/Interop/Unix/Interop.Libraries.cs
```diff
@@ -2,6 +2,8 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+#if !NET8_0_OR_GREATER
```
cheenamalhotra, line 5: Instead of ifdefs in all files, can you manage file inclusion with conditions in NetCore csproj?

cheenamalhotra:
> Using conditional inclusion of files in a project can lead to difficult to diagnose compile-time errors, often times manifesting as "did you forget a reference or using statement?" errors. These are difficult to track down because opening the file in the IDE will usually show no error - the file is right there! why would it not build!?

You can also specify `<TargetFramework>` in the csproj that you want to verify and look at the change in project structure. Files excluded will not be visible in tree anymore.

src/Microsoft.Data.SqlClient/netcore/src/Common/src/Interop/Unix/System.Net.Security.Native/Interop.GssApiException.cs
benrr101: Good to know. There is several files that have !NET8_0_OR_GREATER in this PR, so theoretically they can be removed once net6 support is over. As for now, they should stay, right?

cheenamalhotra:
I agree with that and we will also be able to enable loading all files by default without the need to manage which files to import on what frameworks. But I wanted to provide an alternative solution with target framework while you're reviewing compilation errors..

Overall this change looks good to me, and should help us keep our project files lightweight.
cc @David-Engel for any feedback you may have!

cheenamalhotra (approved):

David-Engel:
The history here is that project file exclusions were the preferred way of doing conditional compilation. I don't know the reasons why it was preferred. I'm guessing the preference may have just been inherited from .NET Core standards (when SDS lived un dotnet runtime). I never liked it for the same reasons @benrr101 mentioned. So I'm in support of this change.

benrr101:
Assuming that this next round of tests pass (except the enclave ones b/c they never pass apparently), this PR should be ready to go.

David-Engel (approved):

PR #2433, opened 2024-03-27
Add support for SqlConnectionOverrides for OpenAsync() API

cheenamalhotra:
As the name suggests, introduces new API:

```cs
public Task SqlConnection.OpenAsync(SqlConnectionOverrides overrides, CancellationToken cancellationToken);
```

Previously, we skipped addiing overrides to OpenAsync in PR https://github.com/dotnet/SqlClient/pull/463 due to no transient fault handling, but with recent PR https://github.com/dotnet/SqlClient/pull/1983 we now enable transient fault handling by default for `OpenAsync()` too.

To support fail fast behavior in OpenAsync(), we need to allow passing `SqlConnectionOverrides` in `OpenAsync()` too.

Additional context:
https://github.com/dotnet/efcore/issues/33399
https://github.com/dotnet/SqlClient/issues/29


src/Microsoft.Data.SqlClient.sln
```diff
@@ -77,8 +77,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Microsoft.Data", "Microsoft
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Microsoft.Data.Sql", "Microsoft.Data.Sql", "{0CE216CE-8072-4985-B248-61F0D0BE9C2E}"
 	ProjectSection(SolutionItems) = preProject
-		..\doc\snippets\Microsoft.Data.Sql\SqlNotificationRequest.xml = ..\doc\snippets\Microsoft.Data.Sql\SqlNotificationRequest.xml
 		..\doc\snippets\Microsoft.Data.Sql\SqlDataSourceEnumerator.xml = ..\doc\snippets\Microsoft.Data.Sql\SqlDataSourceEnumerator.xml
+		..\doc\snippets\Microsoft.Data.Sql\SqlNotificationRequest.xml = ..\doc\snippets\Microsoft.Data.Sql\SqlNotificationRequest.xml
```
DavoudEshtehari, line 81: Could you revert the irrelevant changes in the solution file, please?

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectivityTests/ConnectivityTest.cs
```diff
@@ -382,7 +383,32 @@ public static void ConnectionOpenDisableRetry()
...
-            Assert.True(duration.Seconds > 5, $"Connection Open() with retries took less time than expected. Expect > 5 sec with transient fault handling. Took {duration.Seconds} sec.");                //    sqlConnection.Open();
+            Assert.True(duration.Seconds > 5, $"Connection Open() with retries took less time than expected. Expect > 5 sec with transient fault handling. Took {duration.Seconds} sec.");                  //    sqlConnection.Open();
+        }
+
+        [ConditionalFact(typeof(DataTestUtility), nameof(DataTestUtility.AreConnStringsSetup))]
+        public static async Task ConnectionOpenAsyncDisableRetry()
+        {
+            SqlConnectionStringBuilder connectionStringBuilder = new(DataTestUtility.TCPConnectionString)
+            {
+                InitialCatalog = "DoesNotExist0982532435423",
```
DavoudEshtehari, line 394: It's less likely to exist.
```suggestion
                InitialCatalog = DataTestUtility.GetUniqueNameForSqlServer("DoesNotExist", false),
```

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectivityTests/ConnectivityTest.cs
```diff
@@ -382,7 +383,32 @@ public static void ConnectionOpenDisableRetry()
...
+            await Assert.ThrowsAsync<SqlException>(async () => await sqlConnection.OpenAsync(SqlConnectionOverrides.OpenWithoutRetry, CancellationToken.None));
+            timer.Stop();
+            TimeSpan duration = timer.Elapsed;
+            Assert.True(duration.Seconds < 2, $"Connection OpenAsync() without retries took longer than expected. Expected < 2 sec. Took {duration.Seconds} sec.");
+
+            timer.Restart();
+            await Assert.ThrowsAsync<SqlException>(async () => await sqlConnection.OpenAsync(CancellationToken.None));
+            timer.Stop();
+            duration = timer.Elapsed;
+            Assert.True(duration.Seconds > 5, $"Connection OpenAsync() with retries took less time than expected. Expect > 5 sec with transient fault handling. Took {duration.Seconds} sec.");             //    sqlConnection.OpenAsync();
```
DavoudEshtehari, line 411:
```suggestion
            Assert.True(duration.Seconds > 5, $"Connection OpenAsync() with retries took less time than expected. Expect > 5 sec with transient fault handling. Took {duration.Seconds} sec."); 
```

mdaigle (commented):
Based on my current understanding, the existing default is to _not_ apply transient fault handling for OpenAsync. This changes the default to applying transient fault handling.

It seems like there are two options if we want to make a change now:
1. modify this change to retain the current default behavior
2. do this change, but only apply it on a new (major?) version

However, given what we've learned about the slow, serial nature of OpenAsync, part of me feels like we should wait to do this change until general async performance is improved. 

I would advocate for waiting.

DavoudEshtehari:
@cheenamalhotra Could you wrap this up for the preview 1 with the outstanding comments? Generally, this looks good to me and I'd rather not pushing it to the next preview.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnectionFactory.cs
```diff
@@ -99,6 +99,9 @@ override protected DbConnectionInternal CreateConnection(DbConnectionOptions opt
                         // NOTE: Retrieve <UserInstanceName> here. This user instance name will be used below to connect to the Sql Express User Instance.
                         instanceName = sseConnection.InstanceName;
 
+                        // Set future transient fault handling based on connection options
+                        sqlOwningConnection._applyTransientFaultHandling = opt != null && opt.ConnectRetryCount > 0;
```
cheenamalhotra, line 103: This had to be moved here as in async flow, connection attempt happens on another thread and the value gets reset very quickly to be captured again when retry is attempted. So fail fast wasn't really working. This change makes sure the value gets reset after the login flow is complete.

David-Engel (approved):

mdaigle (approved):

Issue #2178, opened 2023-10-10
ServerCertificate Connection String Parameter Not Establishing Trust in Linux (Tested: RHEL 9, Amazon Linux)


erikfarrell:
### Describe the bug
We're encountering an issue when trying to use ServerCertificate https://learn.microsoft.com/en-us/dotnet/api/microsoft.data.sqlclient.sqlconnectionstringbuilder.servercertificate?view=sqlclient-dotnet-standard-5.1

When trying to establish trust with this parameter in Linux, we get an SSL handshake error (seems to reside in the validation callback). 

Exception and stack trace details:

```
{
  "errorType": "SqlException",
  "errorMessage": "A connection was successfully established with the server, but then an error occurred during the pre-login handshake. (provider: TCP Provider, error: 35 - An internal exception was caught)",
  "stackTrace": [
    "at Microsoft.Data.SqlClient.SqlInternalConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)",
    "at Microsoft.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObject stateObj, Boolean callerHasConnectionLock, Boolean asyncClose)",
    "at Microsoft.Data.SqlClient.TdsParser.EnableSsl(UInt32 info, SqlConnectionEncryptOption encrypt, Boolean integratedSecurity, String serverCertificateFilename)",
    "at Microsoft.Data.SqlClient.TdsParser.ConsumePreLoginHandshake(SqlConnectionEncryptOption encrypt, Boolean trustServerCert, Boolean integratedSecurity, Boolean& marsCapable, Boolean& fedAuthRequired, Boolean tlsFirst, String serverCert)",
    "at Microsoft.Data.SqlClient.TdsParser.Connect(ServerInfo serverInfo, SqlInternalConnectionTds connHandler, Boolean ignoreSniOpenTimeout, Int64 timerExpire, SqlConnectionString connectionOptions, Boolean withFailover)",
    "at Microsoft.Data.SqlClient.SqlInternalConnectionTds.AttemptOneLogin(ServerInfo serverInfo, String newPassword, SecureString newSecurePassword, Boolean ignoreSniOpenTimeout, TimeoutTimer timeout, Boolean withFailover)",
[...]

PR #2439, opened 2024-04-01
Fix | Fix certificate validation

arellegue:
This PR fixes issue #2178.

Problem was with Certificate Chain Link error, which is now resolved by comparing the binary raw data of the client certificate against the server certificate. If both certificates match, we allow it to continue. Prior to this, any certificate policy error, we stopped operations and threw an exception.


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNICommon.cs
```diff
@@ -138,196 +139,154 @@ internal class SNICommon
...
-                        messageBuilder.AppendLine(Strings.SQL_RemoteCertificateNameMismatch);
+                        // if this fails, then fall back to the HostNameInCertificate or TargetServer validation.
+                        SqlClientEventSource.Log.TrySNITraceEvent(nameof(SNITCPHandle), EventType.INFO, "Connection Id {0}, Exception occurred loading specified ServerCertificate: {1}, treating it as if ServerCertificate has not been specified.", args0: connectionId, args1: e.Message);
                     }
-                    else if (targetServerName.Length == certServerName.Length)
+                }
+
+                if (validationCertificate != null)
+                {
+                    if (serverCert.GetRawCertData().SequenceEqual(validationCertificate.GetRawCertData()))
```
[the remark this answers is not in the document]
edwardneal: It might be better to use MemoryExtensions.SequenceEqual; [apisof.net](https://apisof.net/catalog/900e7aa16261db25c301df6b4c5554e0) confirms that this is part of .NET nowadays. Even if this is merged before the removal of .NET Standard support, the required System.Memory package is already a transitive dependency (due to the Azure packages and around half a dozen others.)

If so, `serverCert.GetRawCertData().AsSpan().SequenceEqual(...)` would use this. Better yet might be to get the certificate thumbprints as a Span<byte> and compare those instead - an SHA2 thumbprint is short enough to fit inside a single comparison operation (if the AVX2 intrinsic is enabled.)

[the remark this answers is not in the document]
roji: @David-Engel I'm not aware of any reason to avoid System.Linq specifically - is there some special consideration here? The System.Linq assembly may indeed have some APIs which aren't trimmable, but what matters is the APIs you actually use (e.g. SequenceEqual in this case) - anything else in System.Linq gets trimmed out if it's not used.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNINpHandle.cs
```diff
@@ -369,23 +373,20 @@ public override void DisableSsl()
...
         /// <param name="sender">Sender object</param>
-        /// <param name="cert">X.509 certificate</param>
+        /// <param name="serverCertificate">X.509 certificate</param>
         /// <param name="chain">X.509 chain</param>
         /// <param name="policyErrors">Policy errors</param>
         /// <returns>true if valid</returns>
-        private bool ValidateServerCertificate(object sender, X509Certificate cert, X509Chain chain, SslPolicyErrors policyErrors)
+        private bool ValidateServerCertificate(object sender, X509Certificate serverCertificate, X509Chain chain, SslPolicyErrors policyErrors)
         {
-            using (TrySNIEventScope.Create(nameof(SNINpHandle)))
```
JRahnama, line 378: why this is removed?

src/Microsoft.Data.SqlClient/netfx/src/Microsoft.Data.SqlClient.csproj
```diff
@@ -751,4 +752,4 @@
   <Import Project="$(NetFxSource)tools\targets\GenerateThisAssemblyCs.targets" />
   <Import Project="$(NetFxSource)tools\targets\GenerateAssemblyRef.targets" />
   <Import Project="$(NetFxSource)tools\targets\GenerateAssemblyInfo.targets" />
-</Project>
+</Project>
```
JRahnama, line 755: add extra line

JRahnama:
@David-Engel Even if two certificates have identical raw data, they will have different digital signatures if they are issued by different entities and also even if two certificates have identical raw data, they may have been issued to different subjects. The question would be is comparing two certificate raw data is enough to determine if they are equal?

DavoudEshtehari (commented):
Have you intentionally removed the Address, Circle, Shapes, and Utf8String projects?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNICommon.cs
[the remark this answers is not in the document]
David-Engel: @roji I was following @Wraith2 's efforts from #1949. Perhaps I used the wrong phrasing. Re-reading it sounds like we are just trying to allow linq to be trimmed out completely, not that it isn't trimmable itself.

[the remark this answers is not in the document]
arellegue: serverCert.GetRawCertData().Equals(validationCertificate.GetRawCertData()) did not work even when comparing to itself i.e. serverCert.GetRawCertData().Equals(serverCert.GetRawCertData()). So I tried SequenceEqual and that worked. If that is going to be trimmed, then it definitely needs changing.

David-Engel:
> Even if two certificates have identical raw data, they will have different digital signatures if they are issued by different entities and also even if two certificates have identical raw data, they may have been issued to different subjects. The question would be is comparing two certificate raw data is enough to determine if they are equal?

If that's the case, you are saying the raw data doesn't include the signature nor the subject. But that is counter to the [GetRawCertData docs](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x509certificate.getrawcertdata?view=net-8.0):

> Returns the raw data for the **entire** X.509v3 certificate as an array of bytes.

Comparing the bytes is also how other drivers do the serverCertificate validation.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNICommon.cs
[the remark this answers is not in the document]
roji: I see. Yeah, not using SequenceEquals would allow that get trimmed away, but I'm not sure how valuable that would be (that seems like quite the micro-optimization within the context of this library).

But in any case, MemoryExtensions.SequenceEqual does seems like a good, high-perf alternative that you should really feel free to use.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
```diff
@@ -46,6 +46,7 @@ public ConnectionTestParametersData()
         {
             ConnectionTestParametersList = new List<ConnectionTestParameters>
             {
+                // 1
```
JRahnama, line 49: please change them to be more meaningful comments. What are each number related to. This makes the code reading easier.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNICommon.cs
```diff
@@ -397,5 +355,35 @@ internal static uint ReportSNIError(SNIError error)
             SNILoadHandle.SingletonInstance.LastError = error;
             return TdsEnums.SNI_ERROR;
         }
+
+        internal static unsafe bool ByteArrayCompare(byte[] original, byte[] variant)
```
edwardneal, line 359: I've just re-run an older benchmark, and I don't think you need to include this. I received these results from an 8K byte array.

| Method                             | Comparison | Mean           | Error       | StdDev      |
|----------------------------------- |----------- |---------------:|------------:|------------:|
| ByteByByteComparison             | MatchingArrays | 11,499.0741 ns | 221.0199 ns | 195.9284 ns |
| ByteByByteComparison             | MismatchingArrays |  6,755.5585 ns |  43.1916 ns |  40.4015 ns |
| SpanSequenceEqualComparison                   | MatchingArrays |    232.1920 ns |   1.7383 ns |   1.4515 ns |
| SpanSequenceEqualComparison                   | MismatchingArrays |    232.1477 ns |   2.6315 ns |   2.1974 ns |
| PR2439Comparison | MatchingArrays |    623.6450 ns |  41.8087 ns | 123.2739 ns |
| PR2439Comparison | MismatchingArrays |    576.4529 ns |   4.0778 ns |   3.8143 ns |

`original.AsSpan().SequenceEqual(variant.AsSpan())` is faster, doesn't need unsafe code and removes the LINQ reference which is dragged in by `original.SequenceEqual(variant)`.

Incidentally, it's faster than comparing long-by-long, int-by-int, short-by-short and byte-by-byte because .NET can start to use intrinsics like AVX2 and AVX512 to compare 32 or 64 bytes at a time.

Wraith2: Unless this certificate validation happens extremely frequently is is incredibly slow i'm not sure the performance difference of the method used makes a difference. As long as it's the correct operation you could just use a for loop to compare the bytes if you wanted and it would be sufficient.

If we're already using the package that provides MemoryMarshal for netstandard and netfx then i'd say use MemoryMashal, otherwise just write a loop.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNICommon.cs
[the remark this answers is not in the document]
arellegue: Used AsSpan().SequenceEqual instead as suggested.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNICommon.cs
arellegue: AsSpan().SequenceEqual is now in use.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.EndPoint/TDSServerParser.cs
```diff
@@ -69,15 +69,17 @@ public void Run()
...
                             {
                                 (MessageBeingReceived[0] as TDSPreLoginToken).Encryption = TDSPreLoginTokenEncryptionType.None;
                             }
 
                             // Call into the subscriber to process the packet
                             responseMessages = Server.OnPreLoginRequest(Session, MessageBeingReceived);
-
-                            if (Session.Encryption == TDSEncryptionType.None)
+          
+                            if (Session.Encryption == TDSEncryptionType.None || Session.Encryption == TDSEncryptionType.Off)
```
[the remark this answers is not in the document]
arellegue: Applied changes so that when TDSEncryptionType.None is used TDSPreLoginTokenEncryptionType.None is set and when TDSEncryptionType.Off is used TDSPreLoginTokenEncryptionType.None is set.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
```diff
@@ -0,0 +1,236 @@
...
+                },
+                new ConnectionTestParameters
+                {
+                    TdsEncryptionType = TDSPreLoginTokenEncryptionType.On,
+                    Encrypt = SqlConnectionEncryptOption.Mandatory,
+                    TrustServerCertificate = true,
+                    Certificate = s_mismatchedcert,
+                    HostNameInCertificate = null,
+                    TestResult = true
+                },
```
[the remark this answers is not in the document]
arellegue: Added more parameters, 10 in total,  using TDSPreLoginTokenEncryptionType.Required.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParameters.cs
```diff
@@ -0,0 +1,23 @@
...
+
+namespace Microsoft.Data.SqlClient.ManualTesting.Tests.DataCommon
+{
+    public class ConnectionTestParameters
+    {
+        public TDSPreLoginTokenEncryptionType TdsEncryptionType { get; set; }
+        public SqlConnectionEncryptOption Encrypt { get; set; }
+        public bool TrustServerCertificate { get; set; }
+        public string Certificate { get; set; }
+        public string HostNameInCertificate { get; set; }
```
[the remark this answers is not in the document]
arellegue: Applied naming suggestions.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNICommon.cs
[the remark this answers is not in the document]
arellegue: System.Linq was removed.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbReferenceCollection.cs
```diff
@@ -0,0 +1,279 @@
...
+
+        private void ExitItemLockIfNeeded(bool lockObtained)
+        {
+            if (lockObtained)
+            {
+                Monitor.Exit(_itemLock);
+            }
+        }
+    }
+}
```
JRahnama, line 279: why this change is added here? another PR has merged to address this.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
```diff
@@ -0,0 +1,336 @@
...
+            yield return new object[] { Data.ConnectionTestParametersList[20] };
+            yield return new object[] { Data.ConnectionTestParametersList[21] };
+            yield return new object[] { Data.ConnectionTestParametersList[22] };
+            yield return new object[] { Data.ConnectionTestParametersList[23] };
+            yield return new object[] { Data.ConnectionTestParametersList[24] };
+            yield return new object[] { Data.ConnectionTestParametersList[25] };
+            yield return new object[] { Data.ConnectionTestParametersList[26] };
+            yield return new object[] { Data.ConnectionTestParametersList[27] };
+            yield return new object[] { Data.ConnectionTestParametersList[28] };
+            yield return new object[] { Data.ConnectionTestParametersList[29] };
```
JRahnama, line 56:
```suggestion
   for (int i = 0; i <= 29; i++)
{
    yield return new object[] { Data.ConnectionTestParametersList[i] };
}
```

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
```diff
@@ -0,0 +1,336 @@
...
+                    HostNameInCertificate = null,
+                    TestResult = true
+                },
+                new ConnectionTestParameters
+                {
+                    TdsEncryptionType = TDSPreLoginTokenEncryptionType.Off,
+                    Encrypt = SqlConnectionEncryptOption.Mandatory,
+                    TrustServerCertificate = false,
+                    Certificate = null,
+                    HostNameInCertificate = null,
```
JRahnama, line 78: same here

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
```diff
@@ -0,0 +1,336 @@
...
+                    TestResult = false
+                },
+                new ConnectionTestParameters
+                {
+                    TdsEncryptionType = TDSPreLoginTokenEncryptionType.Off,
+                    Encrypt = SqlConnectionEncryptOption.Optional,
+                    TrustServerCertificate = true,
+                    Certificate = null,
+                    HostNameInCertificate = null,
```
JRahnama, line 87: same here

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
```diff
@@ -0,0 +1,336 @@
...
+                    TestResult = true
+                },
+                new ConnectionTestParameters
+                {
+                    TdsEncryptionType = TDSPreLoginTokenEncryptionType.Off,
+                    Encrypt = SqlConnectionEncryptOption.Mandatory,
+                    TrustServerCertificate = true,
+                    Certificate = null,
+                    HostNameInCertificate = null,
```
JRahnama, line 96: same here

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
```diff
@@ -0,0 +1,336 @@
...
+                    TestResult = true
+                },
+                new ConnectionTestParameters
+                {
+                    TdsEncryptionType = TDSPreLoginTokenEncryptionType.Off,
+                    Encrypt = SqlConnectionEncryptOption.Mandatory,
+                    TrustServerCertificate = false,
+                    Certificate = s_fullPathToCer,
+                    HostNameInCertificate = null,
```
JRahnama, line 105: same here

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
```diff
@@ -0,0 +1,336 @@
...
+                    TestResult = true
+                },
+                new ConnectionTestParameters
+                {
+                    TdsEncryptionType = TDSPreLoginTokenEncryptionType.Off,
+                    Encrypt = SqlConnectionEncryptOption.Mandatory,
+                    TrustServerCertificate = true,
+                    Certificate = s_fullPathToCer,
+                    HostNameInCertificate = null,
```
JRahnama, line 114: same here

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
```diff
@@ -0,0 +1,336 @@
...
+                    HostNameInCertificate = s_hostName,
+                    TestResult = true
+                },
+                new ConnectionTestParameters
+                {
+                    TdsEncryptionType = TDSPreLoginTokenEncryptionType.On,
+                    Encrypt = SqlConnectionEncryptOption.Optional,
+                    TrustServerCertificate = false,
+                    Certificate = null,
+                    HostNameInCertificate = null,
```
JRahnama, line 141: same here

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
```diff
@@ -0,0 +1,336 @@
...
+        {
+            ConnectionTestParametersList = new List<ConnectionTestParameters>
+            {
+                new ConnectionTestParameters
+                {
+                    TdsEncryptionType = TDSPreLoginTokenEncryptionType.Off,
+                    Encrypt = SqlConnectionEncryptOption.Optional,
+                    TrustServerCertificate = false,
+                    Certificate = null,
+                    HostNameInCertificate = null,
```
JRahnama, line 69: can we use the default value which is an empty string? Why null is selected?

Also can you add a comment description of the table being used in below cases?
something like possible options are:
<and add the options int he comment here>

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
```diff
@@ -0,0 +1,336 @@
...
+                    TestResult = false
+                },
+                new ConnectionTestParameters
+                {
+                    TdsEncryptionType = TDSPreLoginTokenEncryptionType.On,
+                    Encrypt = SqlConnectionEncryptOption.Mandatory,
+                    TrustServerCertificate = false,
+                    Certificate = null,
+                    HostNameInCertificate = null,
```
JRahnama, line 150: same here

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
```diff
@@ -0,0 +1,336 @@
...
+                    TestResult = false
+                },
+                new ConnectionTestParameters
+                {
+                    TdsEncryptionType = TDSPreLoginTokenEncryptionType.On,
+                    Encrypt = SqlConnectionEncryptOption.Optional,
+                    TrustServerCertificate = true,
+                    Certificate = null,
+                    HostNameInCertificate = null,
```
JRahnama, line 159: same here

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
```diff
@@ -0,0 +1,336 @@
...
+                    TestResult = true
+                },
+                new ConnectionTestParameters
+                {
+                    TdsEncryptionType = TDSPreLoginTokenEncryptionType.On,
+                    Encrypt = SqlConnectionEncryptOption.Mandatory,
+                    TrustServerCertificate = true,
+                    Certificate = null,
+                    HostNameInCertificate = null,
```
JRahnama, line 168: same here

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
```diff
@@ -0,0 +1,336 @@
...
+                    TestResult = true
+                },
+                new ConnectionTestParameters
+                {
+                    TdsEncryptionType = TDSPreLoginTokenEncryptionType.On,
+                    Encrypt = SqlConnectionEncryptOption.Mandatory,
+                    TrustServerCertificate = true,
+                    Certificate = s_fullPathToCer,
+                    HostNameInCertificate = null,
```
JRahnama, line 186: same here

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
```diff
@@ -0,0 +1,336 @@
...
+                    HostNameInCertificate = s_hostName,
+                    TestResult = true
+                },
+                new ConnectionTestParameters
+                {
+                    TdsEncryptionType = TDSPreLoginTokenEncryptionType.Required,
+                    Encrypt = SqlConnectionEncryptOption.Optional,
+                    TrustServerCertificate = false,
+                    Certificate = null,
+                    HostNameInCertificate = null,
```
JRahnama, line 213: same here

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
```diff
@@ -0,0 +1,336 @@
...
+                    TestResult = false
+                },
+                new ConnectionTestParameters
+                {
+                    TdsEncryptionType = TDSPreLoginTokenEncryptionType.Required,
+                    Encrypt = SqlConnectionEncryptOption.Mandatory,
+                    TrustServerCertificate = false,
+                    Certificate = null,
+                    HostNameInCertificate = null,
```
JRahnama, line 222: same here

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
```diff
@@ -0,0 +1,336 @@
...
+                    TestResult = false
+                },
+                new ConnectionTestParameters
+                {
+                    TdsEncryptionType = TDSPreLoginTokenEncryptionType.Required,
+                    Encrypt = SqlConnectionEncryptOption.Optional,
+                    TrustServerCertificate = true,
+                    Certificate = null,
+                    HostNameInCertificate = null,
```
JRahnama, line 231: same here and all of the `HostNameInCertificate = null` cases

src/Microsoft.Data.SqlClient/tests/ManualTests/Microsoft.Data.SqlClient.ManualTesting.Tests.csproj
```diff
@@ -347,9 +350,18 @@
     <PackageReference Condition="'$(TargetGroup)'!='netfx'" Include="System.ServiceProcess.ServiceController" Version="$(SystemServiceProcessServiceControllerVersion)" />
   </ItemGroup>
   <ItemGroup>
-    <None Condition="'$(TargetGroup)'=='netfx' AND $(ReferenceType)=='Project'" Include="$(BinFolder)$(Configuration).AnyCPU\Microsoft.Data.SqlClient\netfx\**\*SNI*.dll" CopyToOutputDirectory="PreserveNewest" />
+   <ContentWithTargetPath Condition="'$(TargetGroup)'=='netfx' AND $(ReferenceType)=='Project'" Include="$(BinFolder)$(Configuration).AnyCPU\Microsoft.Data.SqlClient\netfx\$(TargetFramework)\*SNI*.dll">
```
JRahnama, line 353: this change is addressed in another PR why is that added here.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft.Data.SqlClient.csproj
```diff
@@ -1,10 +1,8 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="Current" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
+﻿<Project Sdk="Microsoft.Net.Sdk">
   <PropertyGroup>
     <ProjectGuid>{407890AC-9876-4FEF-A6F1-F36A876BAADE}</ProjectGuid>
     <RootNamespace>Microsoft.Data.SqlClient</RootNamespace>
-    <TargetFrameworkVersion>v4.6.2</TargetFrameworkVersion>
+    <TargetFramework>net462</TargetFramework>
```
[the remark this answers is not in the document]
arellegue: I'm not sure where that came from but it is already <TargetFramework>net462</TargetFramework>

src/Microsoft.Data.SqlClient/netfx/src/Microsoft.Data.SqlClient.csproj
```diff
@@ -746,9 +747,8 @@
       <Version>$(SystemBuffersVersion)</Version>
     </PackageReference>
   </ItemGroup>
-  <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
   <Import Project="$(CommonSourceRoot)tools\targets\GenerateResourceStringsSource.targets" />
   <Import Project="$(NetFxSource)tools\targets\GenerateThisAssemblyCs.targets" />
   <Import Project="$(NetFxSource)tools\targets\GenerateAssemblyRef.targets" />
   <Import Project="$(NetFxSource)tools\targets\GenerateAssemblyInfo.targets" />
-</Project>
+</Project>
```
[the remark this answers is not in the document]
arellegue: Extra line was added as suggested.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbReferenceCollection.cs
arellegue: Because I could not get it merged in for some reason.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
arellegue: For loop implemented as suggested.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
arellegue: Sorry could you clarify what same here means?

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
arellegue: Sorry could you clarify what same here means?

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
arellegue: Sorry could you clarify what same here means?

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
arellegue: Sorry could you clarify what same here means?

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
arellegue: Sorry could you clarify what same here means?

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
arellegue: Sorry could you clarify what same here means?

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
arellegue: Sorry could you clarify what same here means?

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
arellegue: Sorry could you clarify what same here means?

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
arellegue: Sorry could you clarify what same here means?

src/Microsoft.Data.SqlClient/tests/ManualTests/Microsoft.Data.SqlClient.ManualTesting.Tests.csproj
arellegue: Yes, why?

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
JRahnama: https://github.com/dotnet/SqlClient/pull/2439#discussion_r1571201554

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
JRahnama: https://github.com/dotnet/SqlClient/pull/2439#discussion_r1571201554

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
JRahnama: https://github.com/dotnet/SqlClient/pull/2439#discussion_r1571201554

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
JRahnama: https://github.com/dotnet/SqlClient/pull/2439#discussion_r1571201554

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
JRahnama: https://github.com/dotnet/SqlClient/pull/2439#discussion_r1571201554

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
JRahnama: https://github.com/dotnet/SqlClient/pull/2439#discussion_r1571201554

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
JRahnama: https://github.com/dotnet/SqlClient/pull/2439#discussion_r1571201554

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
JRahnama: https://github.com/dotnet/SqlClient/pull/2439#discussion_r1571201554

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
JRahnama: https://github.com/dotnet/SqlClient/pull/2439#discussion_r1571201554

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
JRahnama: https://github.com/dotnet/SqlClient/pull/2439#discussion_r1571201554https://github.com/dotnet/SqlClient/pull/2439#discussion_r1571201554

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
JRahnama: https://github.com/dotnet/SqlClient/pull/2439#discussion_r1571201554

JRahnama:
Also, can you address pipeline failures?

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
arellegue: "same here" I guess has been addressed.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
arellegue: "same here" I guess has been addressed.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
arellegue: "same here" I guess has been addressed.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
arellegue: "same here" I guess has been addressed.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
arellegue: "same here" I guess has been addressed.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
arellegue: "same here" I guess has been addressed.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
arellegue: "same here" I guess has been addressed.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
arellegue: "same here" I guess has been addressed.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
arellegue: "same here" I guess has been addressed.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
```diff
@@ -15,6 +15,8 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests.DataCommon
 {
     public class ConnectionTestParametersData
     {
+        private const int CASES = 30;
+        private string _empty = string.Empty;
```
JRahnama, line 19: If you do not specify HNIC its default value is string.Empty no need to reassign it again.

arellegue: This statement does not make sense.  The ConnectionTestParametersData fields must all be initialized for each test case.

src/Microsoft.Data.SqlClient/tests/ManualTests/makepfxcert.ps1
```diff
@@ -0,0 +1,140 @@
...
+        cp $OutDir/localhostcert.pem /usr/local/share/ca-certificates/localhostcert.crt
+
+        # Add trust to the mismatched certificate as well
+        $ openssl x509 -in $OutDir/mismatchedcert.cer -inform der -out $OutDir/mismatchedcert.pem
+        openssl x509 -trustout -addtrust "serverAuth" -in $OutDir/mismatchedcert.pem
+        cp $OutDir/mismatchedcert.pem /usr/local/share/ca-certificates/mismatchedcert.crt
+
+        # Update the certificates store
+        Write-Output "Updating the certificates store..."
+        update-ca-certificates -v
```
JRahnama, line 83: There are couple of things here, generally it is better to use bash on ubuntu, you will have a better control over the commands. 

1.  Enabling as system CA certificate is missing here. It could be done by [dpkg-reconfigure](https://learn.microsoft.com/en-us/sql/linux/sql-server-linux-encrypted-connections?view=sql-server-ver16&tabs=client#register-the-certificate-on-your-client-machine-windows-linux-or-macos) command
2. Remember since the server and client are on the same machine you need to configure them for both.
3. Removing  '!' at the beginning of certificate inside ca-certificate is missed. You can test with `sudo cat /etc/ca-certificates.conf` command to ensure '!' is removed from the certificate name. If '!' shows up Infront of the certificate name it wont be included in `update-ca-certificate` command.
4. It is better to provide chmod 755/777 to ensure sql server has access to those files.

src/Microsoft.Data.SqlClient/tests/ManualTests/makepfxcert.ps1
```diff
@@ -0,0 +1,140 @@
...
+    } else {
+        Write-Output "Creating a self-signed certificate..."
+        $params = @{
+          Type              = "SSLServerAuthentication"
+          Subject           = "CN=$fqdn"
+          KeyAlgorithm      = "RSA"
+          KeyLength         = 2048
+          HashAlgorithm     = "SHA256"
+          TextExtension     = "2.5.29.37={text}1.3.6.1.5.5.7.3.1", "2.5.29.17={text}DNS=$fqdn&DNS=$localhost&IPAddress=$LoopBackIPV4&DNS=$sqlAliasName&IPAddress=$LoopBackIPV6"
```
JRahnama, line 92: This could be simplified by adding DNSName to paramters.

src/Microsoft.Data.SqlClient/tests/ManualTests/makepfxcert.ps1
```diff
@@ -0,0 +1,140 @@
...
+        # Update the certificates store
+        Write-Output "Updating the certificates store..."
+        update-ca-certificates -v
+    } else {
+        Write-Output "Creating a self-signed certificate..."
+        $params = @{
+          Type              = "SSLServerAuthentication"
+          Subject           = "CN=$fqdn"
+          KeyAlgorithm      = "RSA"
+          KeyLength         = 2048
```
JRahnama, line 90: Why on Unix you have chosen 4096 and on windows 2048 for key length?

src/Microsoft.Data.SqlClient/tests/ManualTests/makepfxcert.ps1
```diff
@@ -0,0 +1,140 @@
...
+        Write-Output "Exporting certificate to pfx..."
+        openssl pkcs12 -export -in $OutDir/localhostcert.cer -inkey $OutDir/localhostcert.key -out $OutDir/localhostcert.pfx -password pass:nopassword
+
+        Write-Output "Converting certificate to pem..."
+        # Create pem from cer
+        cp $OutDir/localhostcert.cer $OutDir/localhostcert.pem
+
+        # Add trust to the pem certificate
+        Write-Output "Adding trust to pem certificate..."
+        openssl x509 -trustout -addtrust "serverAuth" -in $OutDir/localhostcert.pem
```
JRahnama, line 66: I think you need to add trust to certificate before copying that to root folder.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectionTestWithSSLCert/CertificateTestWithTdsServer.cs
```diff
@@ -0,0 +1,160 @@
...
+                    Encrypt = connectionTestParameters.Encrypt,
+                };
+            }
+
+            if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows) && userId == string.Empty)
+            {
+                builder.IntegratedSecurity = false;
+                builder.UserID = "user";
+                builder.Password = "password";
+            }
```
David-Engel, line 80: This seems strange. Why do we switch between integrated and un/pw here? And why use the un/pw from the config? Doesn't the auth not matter since TestTdsServer doesn't care about auth here? Seems like this could be simplified to just a few lines... No?
```suggestion
            using TestTdsServer server = TestTdsServer.StartTestServer(enableFedAuth: false, enableLog: false, connectionTimeout: 15,
                methodName: "", new X509Certificate2(s_fullPathToPfx, "nopassword", X509KeyStorageFlags.UserKeySet),
                encryptionType: connectionTestParameters.TdsEncryptionType);

            SqlConnectionStringBuilder builder = new(server.ConnectionString)
            {
                TrustServerCertificate = connectionTestParameters.TrustServerCertificate,
                Encrypt = connectionTestParameters.Encrypt,
            };
```

arellegue: This was addressed in PR #2487.

DavoudEshtehari:
Closing as the rework has been merged with #2487.

Issue #2253, opened 2023-12-01
Add hook for custom SSPI context for SqlConnection instances


twsouthwick:
# Background

There have been a couple of active issues around NTLM/kerberos authentication and customizing it for various reasons (see discussions in #305 and #31), with an attempt of enabling this in #1379. This issue is to track an API proposal and the steps needed to integrate it into SqlClient.

> NOTE:
> I'm not sure of naming in this area and very open to better names for things. I haven't spent much time within the SSPI/GSS/NTLM/Kerberos/etc space and this proposal is based on my limited understanding of these concepts. If I'm missing anything important, please let me know

# Proposal

The proposal is to add a provider that has the ability to generate a context given an incoming blob. There are currently a few ways that this is handled internally:

- A native implementation on all Windows platforms
- An internal managed implementation on .NET Standard/.NET 6
- System.Net.Security.NegotiateAuthentication on .NET 7+

As part of this change, those different implementations should be moved to follow the same pattern as an external implementer would have to ensure the abstraction is sufficient. After [attempting to do so](https://github.com/dotnet/SqlClient/compare/main...twsouthwick:SqlClient:sspi-common-base), it appears the best design would be to have a class that is used per connection (retrieved via a factory method) as there is some state that may need to be accessed between invocations:

```diff
+ /// <summary>
+ /// Defines a provider to supply and operate on an SSPI context as part of integrated authentication.
+ /// </summary>
+ public abstract class SspiClientContextProvider
+ {
+     /// <summary>
+     /// Generates a context given a (possibly empty) previously buffer.
+     /// </summary>
+     /// <param name="incomingBlob">The incoming blob (may be empty).</param>
+     /// <param name="outgoingBlobWriter">A writer that allows us to write data for the outgoing blob.</param>
+     protected abstract bool GenerateContext(ReadOnlySpan<byte> incomingBlob, IBufferWriter<byte> outgoingBlobWriter, SspiAuthenticationParameters authParams);
+ }

public class SqlConnection
{
+   /// <summary>
+   /// Gets or sets a factory to handle the SSPI context
+   /// </summary>
+   public Func<SspiClientContextProvider> SspiContextProviderFactory { get; set; }
}
```

*Alternate designs*

- We could just return a byte[] - but this would most likely requrie creating new arrays for any authentication

  ```diff
  public abstract class SspiClientContextProvider
  {
[...]

PR #2447, opened 2024-04-05
Use memory/span for SSPI generation

### LINKED ISSUE #2441: TdsParser.ProcessSSPI in NetCore uses ArrayPool.Shared.Rent which might return an non zero-initialized array
### Describe the bug
TdsParser for NetCore uses ArrayPool.Shared.Rent(...) which might return an non-zero-initialized array.
https://github.com/dotnet/SqlClient/blob/faf9b95995e6200a980bb6b1980f1d7af9f46cef/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs#L8654-L8683

The first bytes of that array are then replaced with the [negTokenResp](https://www.rfc-editor.org/rfc/rfc4178).

But as the array has some bytes set from the beginning, there might be some additional data after the actual negTokenResp, which causes the MIT-krb5 implementation to fail with a DEFECTIVE-TOKEN [here](https://github.com/krb5/krb5/blob/krb5-1.20/src/lib/gssapi/spnego/spnego_mech.c#L3731).

```
Exception message:
Cannot authenticate using Kerberos. Ensure Kerberos has been initialized on the client with 'kinit' and a Service Principal Name has been registered for the SQL Server to allow Kerberos authentication.\nErrorCode=InternalError, Exception=Interop+NetSecurityNative+GssApiException: GSSAPI operation failed with error - Invalid token was supplied (Unknown error).\n  

Stack trace:
   at System.Net.Security.NegotiateStreamPal.GssInitSecurityContext(SafeGssContextHandle& context, SafeGssCredHandle credential, Boolean isNtlm, SafeGssNameHandle targetName, GssFlags inFlags, Byte[] buffer, Byte[]& outputBuffer, UInt32& outFlags, Int32& isNtlmUsed)\n  
   at System.Net.Security.NegotiateStreamPal.EstablishSecurityContext(SafeFreeNegoCredentials credential, SafeDeleteContext& context, String targetName, ContextFlagsPal inFlags, SecurityBuffer inputBuffer, SecurityBuffer outputBuffer, ContextFlagsPal& outFlags)\n  
   at Microsoft.Data.SqlClient.SNI.SNIProxy.GenSspiClientContext(SspiClientContextStatus sspiClientContextStatus, Byte[] receivedBuff, Byte[]& sendBuff, Byte[][] serverName)\n  
   at Microsoft.Data.SqlClient.SNI.TdsParserStateObjectManaged.GenerateSspiClientContext(Byte[] receivedBuff, UInt32 receivedLength, Byte[]& sendBuff, UInt32& sendLength, Byte[][] _sniSpnBuffer)\n  
   at Microsoft.Data.SqlClient.TdsParser.SSPIData(Byte[] receivedBuff, UInt32 receivedLength, Byte[]& sendBuff, UInt32& sendLength)
```

### To reproduce
I'm using an application with multiple connection-strings which - for testing purposes - queries the database very often, afterwards it invokes `SqlConnection.ClearAllPools();`. And this is done in an endless loop -> This causes the client to establish new security-contexts heavily which causes the error to appear.

### Expected behavior
Only the negTokenResp bytes are sent to the underlying GSSAPI implementation

### Further technical details
Microsoft.Data.SqlClient version: (found on the nuget or Microsoft.Data.SqlClient.dll)
.NET target: (e.g. .NET6)
SQL Server version: Microsoft SQL Server 2019 
Operating system: aspnet:7.0-bookworm-slim

**Additional context**

I event tried the latest debian distro, but there the latest krb5 version is 1.20 which still does that check. In 1.21 they kind of revamped that check, but it's not available for debian as it stands now.

twsouthwick:
This change makes more use of memory/span APIs so that SSPI generation does not need to track arrays and indices.

- The interop layer gets some updates to use span instead of pointers
- Updates the SSPI generation to take in a ReadOnlyMemory<byte>

Part of #2253

Fixes #2441

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIProxy.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIProxy.cs
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
+using System.Buffers;
 using System.Diagnostics;
 using System.IO;
 using System.Net;
@@ -34,7 +35,21 @@ namespace Microsoft.Data.SqlClient.SNI
         /// <param name="sendBuff">Send buffer</param>
         /// <param name="serverName">Service Principal Name buffer</param>
         /// <returns>SNI error code</returns>
-        internal static void GenSspiClientContext(SspiClientContextStatus sspiClientContextStatus, byte[] receivedBuff, ref byte[] sendBuff, byte[][] serverName)
+        internal static void GenSspiClientContext(SspiClientContextStatus sspiClientContextStatus, ReadOnlyMemory<byte> receivedBuff, ref byte[] sendBuff, byte[][] serverName)
+        {
+            // TODO: this should use ReadOnlyMemory all the way through
+            byte[] array = null;
+
+            if (!receivedBuff.IsEmpty)
+            {
+                array = new byte[receivedBuff.Length];
+                receivedBuff.CopyTo(array);
+            }
+
+            GenSspiClientContext(sspiClientContextStatus, array, ref sendBuff, serverName);
+        }
+
+        private static void GenSspiClientContext(SspiClientContextStatus sspiClientContextStatus, byte[] receivedBuff, ref byte[] sendBuff, byte[][] serverName)
         {
             SafeDeleteContext securityContext = sspiClientContextStatus.SecurityContext;
             ContextFlagsPal contextFlags = sspiClientContextStatus.ContextFlags;
@@ -189,7 +204,7 @@ namespace Microsoft.Data.SqlClient.SNI
                 case DataSource.Protocol.TCP:
                     sniHandle = CreateTcpHandle(details, timeout, parallel, ipPreference, cachedFQDN, ref pendingDNSInfo,
                         tlsFirst, hostNameInCertificate, serverCertificateFilename);
-                     break;
+                    break;
                 case DataSource.Protocol.NP:
                     sniHandle = CreateNpHandle(details, timeout, parallel, tlsFirst);
                     break;
--- a/src/Microsoft.Data.SqlClient/netcore/src/Interop/SNINativeMethodWrapper.Windows.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Interop/SNINativeMethodWrapper.Windows.cs
@@ -315,7 +315,7 @@ namespace Microsoft.Data.SqlClient
         [DllImport(SNI, CallingConvention = CallingConvention.Cdecl)]
         private static extern unsafe uint SNISecGenClientContextWrapper(
             [In] SNIHandle pConn,
-            [In, Out] byte[] pIn,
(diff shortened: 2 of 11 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIProxy.cs
```diff
@@ -31,10 +32,25 @@ internal class SNIProxy
...
         /// <param name="receivedBuff">Receive buffer</param>
-        /// <param name="sendBuff">Send buffer</param>
         /// <param name="serverName">Service Principal Name buffer</param>
-        /// <returns>SNI error code</returns>
-        internal static void GenSspiClientContext(SspiClientContextStatus sspiClientContextStatus, byte[] receivedBuff, ref byte[] sendBuff, byte[][] serverName)
+        /// <returns>Memory for response</returns>
+        internal static IMemoryOwner<byte> GenSspiClientContext(SspiClientContextStatus sspiClientContextStatus, ReadOnlyMemory<byte> receivedBuff, byte[][] serverName)
+        {
+            // TODO: this should use ReadOnlyMemory all the way through
+            var array = ArrayPool<byte>.Shared.Rent(receivedBuff.Length);
```
stuka120, line 40: Please consider when switching to `ReadOnlyMemory` the problem that the array returned by `ArrayPool<byte>.Shared.Rent` might me non-null-initialized and larger than the length specified. That causes sometimes problems during runtime like mentioned in [in this issue](https://github.com/dotnet/SqlClient/issues/2441).

twsouthwick: I think this implementation will actually fix that issue as is, as it returns a Memory<byte> rather than the actual array so the extra entries wouldn't be accessible.

stuka120: But aren't we still passing an instance of `byte[]` down to the `GSSAPI.InitSecContext` which might have some wrong bytes from the `ArrayPool`?
Cause the new `MemoryStream` is created from the `outSecurityBuffer`, but the bug is caused by the `receivedBuff`, which after `receivedBuff.CopyTo(array);` might have some non-zero-bytes after the copied bytes.

twsouthwick: I see - I've update the array to be cleared out when renting

twsouthwick:
@David-Engel step 2 in the sspi journey

twsouthwick:
@David-Engel any chance to get a review here?

David-Engel (approved):

David-Engel:
Pipeline failures are due to test config changes that can be pulled in from main. I ran the Kerberos test pipeline against the branch and all tests passed. This just needs one more reviewer.

@JRahnama, @DavoudEshtehari

DavoudEshtehari (approved):

PR #2449, opened 2024-04-05
Fix | Fix ArrayPool usages in TdsParser.ProcessSSPI

### LINKED ISSUE #2441: TdsParser.ProcessSSPI in NetCore uses ArrayPool.Shared.Rent which might return an non zero-initialized array

stuka120:
Reapply the implementation of `TdsParser.ProcessSSPI` from the netfx implementation which does not utilize the `ArrayPool.Shared.Rent(...)` mechanism, as the `ArrayPool` received `byte[]` is not guaranteed to be zero-initialized as specified [in the docs](https://learn.microsoft.com/en-us/dotnet/api/system.buffers.arraypool-1.rent?view=net-8.0), and also has not the correct size, as the given `receivedLength` is only specifying the minimum-size of the returned `byte[]`.

Old netfx implementation in version 5.2.0:
https://github.com/dotnet/SqlClient/blob/faf9b95995e6200a980bb6b1980f1d7af9f46cef/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs#L9495-L9524

If the `ArrayPool.Shared.Rent(...)` array is non-zero-initialized and oversized, this further causes invalid genTokenResp to be passed to the underlying `GSSAPI#InitSecContext` implementation.

fixes #2441

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParser.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParser.cs
@@ -25,6 +25,8 @@ namespace Microsoft.Data.SqlClient
                 throw SQL.SynchronousCallMayNotPend();
             }
 
+            receivedBuff.AsSpan(receivedLength, receivedBuff.Length - receivedLength).Clear();
+
             // allocate send buffer and initialize length
             byte[] rentedSendBuff = ArrayPool<byte>.Shared.Rent((int)_authenticationProvider!.MaxSSPILength);
             byte[] sendBuff = rentedSendBuff; // need to track these separately in case someone updates the ref parameter


twsouthwick:
I'm currently reworking the sspi generation in general to enable a public api around it. One of my changes in #2447 fixes this in the direction I need for the larger item https://github.com/dotnet/SqlClient/issues/2253.

Part of the changes I'm trying to push in moves toward more memory/span operations (and helps consolidate the implementations between framework and core). There are currently 4 implementations of generating SSPI contexts and the others work fine with a `ReadOnlyMemory<byte>`. Forcing all of them to use an array because a single one requires it will cause undue memory allocations.

stuka120:
I changed the implementation to don't do another allocation but instead clear the "overflowing" bytes inside the `receivedBuff` by doing `receivedBuff.AsSpan(receivedLength, receivedBuff.Length - receivedLength).Clear();`. That should at least for the time being do the trick.
Maybe we can use that until your rework of the SSPI implementations is complete @twsouthwick?

twsouthwick:
@stuka120 my change for sspi with this fix has been merged so this can probably be closed

David-Engel:
Fixed via #2447

PR #2454, opened 2024-04-08
Expose SspiAuthenticationParameters on SspiContextProvider

twsouthwick:
This change updates the SSPI context provider to surface information to implementers via SqlAuthenticationParameters.

This also expands https://github.com/dotnet/SqlClient/pull/2559 to loop through all SPNs in the base SSPIContextProvider rather than just a specific implementation.

Part of #2253


twsouthwick:
@mdaigle next step for sspi :)

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/SSPIContextProvider.cs
```diff
@@ -26,26 +26,80 @@ private protected virtual void Initialize()
...
+            {
+                //throw new InvalidOperationException(SQLMessage.SSPIGenerateError() + Environment.NewLine + statusCode);
+                SSPIError(e.Message + Environment.NewLine + e.StackTrace, TdsEnums.GEN_CLIENT_CONTEXT);
+                return false;
             }
         }
 
+        private static SqlAuthenticationParameters CreateSqlAuthParams(SqlInternalConnectionTds connection, string serverSpn)
+        {
+            var auth = new SqlAuthenticationParameters.Builder(
```
twsouthwick, line 83: note: I'm rebuilding this, but if there's an existing one that can be reused, let me know

twsouthwick:
ping @mdaigle @cheenamalhotra

twsouthwick:
The errors appear to also be on main and are an "Access denied" for something related to certificates. It doesn't appear to be caused by my change, but let me know if you believe it is.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/NegotiateSSPIContextProvider.cs
```diff
@@ -12,33 +12,24 @@ internal sealed class NegotiateSSPIContextProvider : SSPIContextProvider
...
+            // Log session id, status code and the actual SPN used in the negotiation
+            SqlClientEventSource.Log.TryTraceEvent("{0}.{1} | Info | Session Id {2}, StatusCode={3}, SPN={4}", nameof(NegotiateSSPIContextProvider),
+                nameof(GenerateSspiClientContext), _physicalStateObj.SessionId, statusCode, _negotiateAuth.TargetName);
 
-            if (statusCode is not NegotiateAuthenticationStatusCode.Completed and not NegotiateAuthenticationStatusCode.ContinueNeeded)
+            if (statusCode == NegotiateAuthenticationStatusCode.Completed || statusCode == NegotiateAuthenticationStatusCode.ContinueNeeded)
             {
-                throw new InvalidOperationException(SQLMessage.SSPIGenerateError() + Environment.NewLine + statusCode);
+                outgoingBlobWriter.Write(sendBuff);
+                return true; // Successful case, exit the loop with current SPN.
```
mdaigle, line 29: comment no longer applies

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/SSPIContextProvider.cs
mdaigle: I don't see an instance that makes sense to reuse. They're mostly build in the context of federated auth, which isn't applicable here.

mdaigle (commented):
I think this all makes sense to me. I'll take another look tomorrow and also see why the tests are failing.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/SSPIContextProvider.cs
```diff
@@ -26,24 +26,78 @@ private protected virtual void Initialize()
...
+                SSPIError(e.Message + Environment.NewLine + e.StackTrace, TdsEnums.GEN_CLIENT_CONTEXT);
+                return false;
+            }
+        }
+
+        private static SqlAuthenticationParameters CreateSqlAuthParams(SqlInternalConnectionTds connection, string serverSpn)
+        {
+            var auth = new SqlAuthenticationParameters.Builder(
+                authenticationMethod: connection.ConnectionOptions.Authentication,
+                resource: null,
```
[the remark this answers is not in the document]
mdaigle: I'm also not sure how to feel about authority. In federated auth, it's assumed that authority will be present, but here we don't set a value.

Let me chat with the team to see if we'd prefer to create a new type to better represent this set of credentials.

mdaigle (commented):
Tests _should_ be green now. We had a build agent issue that was resolved yesterday.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/SSPIContextProvider.cs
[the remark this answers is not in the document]
twsouthwick: Sounds good. I refactored the builder so that the same pattern can be had wherever it's used if we want to use the existing `SqlAuthenticationParameters`

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/NativeSSPIContextProvider.cs
```diff
@@ -49,7 +49,7 @@ private void LoadSSPILibrary()
             }
         }
 
-        protected override void GenerateSspiClientContext(ReadOnlySpan<byte> incomingBlob, IBufferWriter<byte> outgoingBlobWriter, ReadOnlySpan<string> serverSpns)
+        protected override bool GenerateSspiClientContext(ReadOnlySpan<byte> incomingBlob, IBufferWriter<byte> outgoingBlobWriter, SqlAuthenticationParameters authParams)
```
paulmedynski, line 52: I wonder if an auth parameters object specific to SSPI would be a better choice here.  With SqlAuthenticationParameters we're exposing maybe more values than are necessary.  We're also tying ourselves to future changes to SqlAuthenticationParameters that may not be appropriate for SSPI.  We may also have to explicitly document which fields we set, and which ones we don't set.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/NegotiateSSPIContextProvider.cs
paulmedynski, line 1: I think the convention is to add `.netcore` to the filename before the `.cs` extension when the file contains code solely intended for .NET.  @benrr101 ?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlAuthenticationParameters.cs
```diff
@@ -149,11 +149,13 @@ public Builder WithConnectionTimeout(int timeout)
                 return this;
             }
 
-            internal Builder(SqlAuthenticationMethod authenticationMethod, string resource, string authority, string serverName, string databaseName)
+            internal Builder(SqlInternalConnectionTds connection, string resource, string authority)
```
paulmedynski, line 152: I think we should keep these changes, even if we make an SspiAuthenticationParameters object.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/SSPIContextProvider.cs
```diff
@@ -26,24 +26,71 @@ private protected virtual void Initialize()
...
+            {
+                auth.WithUserId(userId);
+            }
+
+            if (connection.ConnectionOptions.Password is { } password)
+            {
+                auth.WithPassword(password);
+            }
+
+            return auth;
```
paulmedynski, line 93: Implicit conversion operator from Builder to SqlAuthenticationParameters?!?  Now that's surprising...

paulmedynski (changes requested):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/NegotiateSSPIContextProvider.cs
benrr101: Ah, yes, I should have noticed that. Yes, netcore only files should end in .netcore.cs.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/NativeSSPIContextProvider.cs
David-Engel: I'm leaning this way, too... not strongly. But it would make the interface clearer.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/NegotiateSSPIContextProvider.cs
David-Engel: I thought the .netcore.cs and .unix.cs convention was for back when we were using csproj based inclusion/exclusion. Do we need the extra info in the file names? Aren't we trying to get away from separate files for netfx/netcore/windows/unix and just use compiler directives?

twsouthwick: this file has gone through a few renames for this kind of thing. Happy to rename if needed, but I'll resolve this for now

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/NativeSSPIContextProvider.cs
twsouthwick: I created a new one (`SspiAuthenticationParameters`) since there were a few comments to this effect

mdaigle (approved):

paulmedynski (approved):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/NegotiateSspiContextProvider.cs
```diff
@@ -0,0 +1,36 @@
...
+{
+    internal sealed class NegotiateSspiContextProvider : SspiContextProvider
+    {
+        private NegotiateAuthentication? _negotiateAuth = null;
+
+        protected override bool GenerateSspiClientContext(ReadOnlySpan<byte> incomingBlob, IBufferWriter<byte> outgoingBlobWriter, SspiAuthenticationParameters authParams)
+        {
+            NegotiateAuthenticationStatusCode statusCode = NegotiateAuthenticationStatusCode.UnknownCredentials;
+
+            _negotiateAuth ??= new(new NegotiateAuthenticationClientOptions { Package = "Negotiate", TargetName = authParams.Resource });
```
vonzshik, line 19: I'm probably missing something, but from my understanding `SspiContextProvider` attempts to generate sspi context by iterating over each `serverSpns` (which is passed via `authParams.Resource`). But because `_negotiateAuth` is cached here, if it fails for the first spn, it's also going to fail for the others since it's going to be the exact same spn.

mdaigle: You're right. The implementation used to reset _negotiateAuth to null if authentication failed, but that's missing now. @twsouthwick can we correct it in a follow up PR?

twsouthwick: Here's the PR to add that back: https://github.com/dotnet/SqlClient/pull/3347

PR #2459, opened 2024-04-11
Test | Adding option to clean tests output folder for manual and functional tests

JRahnama:
When debugging, we often find that we don't need to rebuild the entire solution when making changes solely to tests. This presents opportunities to tidy up the tests output directory.

### DIFF
--- a/build.proj
+++ b/build.proj
@@ -199,6 +199,11 @@
     <RemoveDir Directories='$([System.IO.Directory]::GetDirectories(".",".nuget", SearchOption.AllDirectories))' />
   </Target>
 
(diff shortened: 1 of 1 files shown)


build.proj
```diff
@@ -199,6 +199,11 @@
     <RemoveDir Directories='$([System.IO.Directory]::GetDirectories(".",".nuget", SearchOption.AllDirectories))' />
   </Target>
 
+  <Target Name="CleanTests">
```
DavoudEshtehari, line 202: I would clean up all the projects under the test folder from the build directory.

Issue #2460, opened 2024-04-11
Microsoft.Data.SqlClient 5.2.0 depends on Azure.Identity 1.10.3 with CVE-2024-29992 vulnerability


Banner-Keith:
[This CVE](https://github.com/advisories/GHSA-wvxc-855f-jvrv) just got published. 

Please publish a new patch of Microsoft.Data.SqlClient that depends on Azure.Identity 1.11.0.

JRahnama:
@Banner-Keith Thanks for bringing this up. PR #2462 is created to address this issue.

DavoudEshtehari:
Closing by PR #2462

Issue #110, opened 2019-05-20
Question: Too many Session.Reset() calls


TrayanZapryanov:
Hello,

I have a question about an internal class in 
[https://github.com/dotnet/corefx/blob/6977dddc03f8263ad82228558eb9b02e2e6f530c/src/System.Data.SqlClient/src/System/Data/SqlClient/SqlInternalConnectionTds.cs](url).
As shown in image below Session.Reset() call is one of main contributors to memory allocations:
![SessionData](https://user-images.githubusercontent.com/12857389/57999412-05ca8e80-7ade-11e9-86b6-ebc6af8ba1a2.png)
I've tried to search over internet what can cause this class , but not luck. 
Is it possible someone to say when ` public const byte FEATUREEXT_SRECOVERY  = 0x01;` was set as it obviously triggers 
`if (_deltaDirty)
            {

                _delta = new SessionStateRecord[_maxNumberOfSessionStates];
                _deltaDirty = false;
}` case.
Also is it possible to replace this array with a list as making 256 SessionStateRecord objects seems a bit too much for me.

Wraith2:
If you can work up a small isolated reproduction (including any sql server setup) then i can investigate and might be able to make improvements. I haven't seen anything related to this while profiling other cases so it might be something specific on the sql side that triggers it.

TrayanZapryanov:
Thank you for the check, but result is from a complex test and will be hard to isolate it. I hoped that somebody will remember what this "FEATUREEXT_SRECOVERY" means as it marks whole object as dirty. 
If nobody have a clue, I will say that it is normal and will close the question.

Wraith2:
I can tell you what it's for. It's a protocol feature for session recovery. 

When enabled (which i don't know how you do at the moment) the server will occasionally send session state data to the client which stores that data. If the underlying TCP connection between them is disrupted the client can attempt to reconnect to the server and include the session data in the conneciton message, if the server finds an orphaned session with the same session identifier and state as the connection request it'll reconnect the two together and you carry on as if your connection hadn't dropped.

I had a brief look through the code last night and it does seem quite wateful. In particular the new array allocation you point out might still work with just an Array.Clear() call. Without being able to enable the feture i'm a bit stuck for debugging options because i can't verify that i haven't broken anything.

TrayanZapryanov:
[...]

PR #2462, opened 2024-04-11
Change | Updating Azure.Identity version to 1.11.2

JRahnama:
Addresses the issue mentioned at #2460 for [CVE-2024-29992](https://github.com/advisories/GHSA-wvxc-855f-jvrv)

### DIFF
--- a/tools/props/Versions.props
+++ b/tools/props/Versions.props
@@ -27,10 +27,10 @@
   </PropertyGroup>
   <!-- NetFx and NetCore project dependencies -->
   <PropertyGroup>
-    <AzureIdentityVersion>[1.11.0,2.0.0)</AzureIdentityVersion>
-    <MicrosoftIdentityClientVersion>[4.60.1,5.0.0)</MicrosoftIdentityClientVersion>
-    <MicrosoftIdentityModelProtocolsOpenIdConnectVersion>[7.5.0,8.0.0)</MicrosoftIdentityModelProtocolsOpenIdConnectVersion>
-    <MicrosoftIdentityModelJsonWebTokensVersion>[7.5.0,8.0.0)</MicrosoftIdentityModelJsonWebTokensVersion>
+    <AzureIdentityVersion>1.11.0</AzureIdentityVersion>
+    <MicrosoftIdentityClientVersion>4.60.1</MicrosoftIdentityClientVersion>
+    <MicrosoftIdentityModelProtocolsOpenIdConnectVersion>7.5.0</MicrosoftIdentityModelProtocolsOpenIdConnectVersion>
+    <MicrosoftIdentityModelJsonWebTokensVersion>7.5.0</MicrosoftIdentityModelJsonWebTokensVersion>
     <SystemBuffersVersion>4.5.1</SystemBuffersVersion>
     <SystemTextEncodingsWebVersion>6.0.0</SystemTextEncodingsWebVersion>
     <MicrosoftSourceLinkGitHubVersion>8.0.0</MicrosoftSourceLinkGitHubVersion>
@@ -39,7 +39,7 @@
   <PropertyGroup>
     <MicrosoftWin32RegistryVersion>5.0.0</MicrosoftWin32RegistryVersion>
     <MicrosoftDataSqlClientSNIRuntimeVersion>5.2.0</MicrosoftDataSqlClientSNIRuntimeVersion>
-    <SystemConfigurationConfigurationManagerVersion>[6.0.1,7.0.0)</SystemConfigurationConfigurationManagerVersion>
+    <SystemConfigurationConfigurationManagerVersion>6.0.1</SystemConfigurationConfigurationManagerVersion>
     <MicrosoftSqlServerServerVersion>1.0.0</MicrosoftSqlServerServerVersion>
     <SystemDiagnosticsPerformanceCounterVersion>6.0.1</SystemDiagnosticsPerformanceCounterVersion>
     <SystemRuntimeCachingVersion>6.0.0</SystemRuntimeCachingVersion>
@@ -55,8 +55,8 @@
   </PropertyGroup>
   <!-- AKV Provider project dependencies -->
   <PropertyGroup>
-    <AzureCoreVersion>[1.38.0,2.0.0)</AzureCoreVersion>
-    <AzureSecurityKeyVaultKeysVersion>[4.4.0,5.0.0)</AzureSecurityKeyVaultKeysVersion>
+    <AzureCoreVersion>1.38.0</AzureCoreVersion>
+    <AzureSecurityKeyVaultKeysVersion>4.4.0</AzureSecurityKeyVaultKeysVersion>
     <MicrosoftExtensionsCachingMemoryVersion>6.0.1</MicrosoftExtensionsCachingMemoryVersion>
   </PropertyGroup>
   <!-- Test Project Dependencies -->
--- a/tools/props/Versions.props
+++ b/tools/props/Versions.props
@@ -27,10 +27,10 @@
   </PropertyGroup>
   <!-- NetFx and NetCore project dependencies -->
   <PropertyGroup>
-    <AzureIdentityVersion>1.11.0</AzureIdentityVersion>
-    <MicrosoftIdentityClientVersion>4.60.1</MicrosoftIdentityClientVersion>
-    <MicrosoftIdentityModelProtocolsOpenIdConnectVersion>7.5.0</MicrosoftIdentityModelProtocolsOpenIdConnectVersion>
-    <MicrosoftIdentityModelJsonWebTokensVersion>7.5.0</MicrosoftIdentityModelJsonWebTokensVersion>
+    <AzureIdentityVersion>[1.11.0,2.0.0)</AzureIdentityVersion>
+    <MicrosoftIdentityClientVersion>[4.60.1,5.0.0)</MicrosoftIdentityClientVersion>
+    <MicrosoftIdentityModelProtocolsOpenIdConnectVersion>[7.5.0,8.0.0)</MicrosoftIdentityModelProtocolsOpenIdConnectVersion>
+    <MicrosoftIdentityModelJsonWebTokensVersion>[7.5.0,8.0.0)</MicrosoftIdentityModelJsonWebTokensVersion>
     <SystemBuffersVersion>4.5.1</SystemBuffersVersion>
     <SystemTextEncodingsWebVersion>6.0.0</SystemTextEncodingsWebVersion>
     <MicrosoftSourceLinkGitHubVersion>8.0.0</MicrosoftSourceLinkGitHubVersion>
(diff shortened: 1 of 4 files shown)


arellegue (approved):
lgtm

cheenamalhotra:
Based on NuGet's docs, use * instead. See [Use Floating Versions](https://learn.microsoft.com/en-us/nuget/concepts/dependency-resolution#floating-versions)

The [comments in example](https://learn.microsoft.com/en-us/nuget/concepts/package-versioning?tabs=semver20sort#references-in-project-files-packagereference) show that smallest stable versions are resolved with version ranges.

SimonCropp:
this seems to change significantly more than "Updating Azure.Identity version to 1.11.0". it is also assigning upper version limits for many references

SimonCropp:
also. can we get a patch out for this. people should not need to wait for v6 to get this fix

JRahnama:
> also. can we get a patch out for this. people should not need to wait for v6 to get this fix

Sure, it is in our to do list for next patch release. Date yet TBD.

> this seems to change significantly more than "Updating Azure.Identity version to 1.11.0". it is also assigning upper version limits for many references

This is going to change to wild cards. something like 1.11.* or 1.* to pick the latest stable release of the dependency.

SimonCropp:
> This is going to change to floating versions. something like 1.11.* or 1.* to pick the latest stable release of the dependency.

Doesnt nuget always resolve to the lowest transitive? so, in this case, people would still need an explicit reference to Azure.Identity so as to avoid the CVE?

JRahnama:
> > This is going to change to floating versions. something like 1.11.* or 1.* to pick the latest stable release of the dependency.
> 
> Doesnt nuget always resolve to the lowest transitive? so, in this case, people would still need an explicit reference to Azure.Identity so as to avoid the CVE?

Something similar to [this](https://learn.microsoft.com/en-us/nuget/concepts/package-versioning?tabs=semver20sort#references-in-project-files-packagereference) documentation.

JRahnama:
> > > This is going to change to floating versions. something like 1.11.* or 1.* to pick the latest stable release of the dependency.
> > 
> > 
> > Doesnt nuget always resolve to the lowest transitive? so, in this case, people would still need an explicit reference to Azure.Identity so as to avoid the CVE?
> 
> Something similar to [this](https://learn.microsoft.com/en-us/nuget/concepts/package-versioning?tabs=semver20sort#references-in-project-files-packagereference) documentation.

For example if we had 1.* in our Versions.props, when 1.10.3 was marked as deprecated, the next available version would have been picked which is 1.11.0

roji:
@JRahnama @cheenamalhotra as far as I know, it's discouraged to use floating nuget dependency versions in most production software, because it leads to non-deterministic builds. That is, the moment a dependency releases a version, your next CI run automatically uses that, possibly causing failures, performance regressions or whatever. This can make sense in various prototyping or internal development scenarios, but is less recommended for stable, public software.

The alternative mechanism for keeping versions up to date is [dependabot](https://github.com/dependabot), which is an official github feature; a bot submits PRs updating dependency versions whenever new ones become available, and team members merge these as usual. This ensures a git commit for the dependency version change, allowing tracking exactly what changed and when, e.g. when looking at the project history or diagnosing some problem. I'd recommend looking into enabling dependabot - it's widely used in many .NET projects (including EF and other Microsoft .NET projects) and works very well.

Let me know if the above makes sense and if I've missed something specific to SqlClient.

JRahnama:
> @JRahnama @cheenamalhotra as far as I know, it's discouraged to use floating nuget dependency versions in most production software, because it leads to non-deterministic builds. That is, the moment a dependency releases a version, your next CI run automatically uses that, possibly causing failures, performance regressions or whatever. This can make sense in various prototyping or internal development scenarios, but is less recommended for stable, public software.
> 
> The alternative mechanism for keeping versions up to date is [dependabot](https://github.com/dependabot), which is an official github feature; a bot submits PRs updating dependency versions whenever new ones become available, and team members merge these as usual. This ensures a git commit for the dependency version change, allowing tracking exactly what changed and when, e.g. when looking at the project history or diagnosing some problem. I'd recommend looking into enabling dependabot - it's widely used in many .NET projects (including EF and other Microsoft .NET projects) and works very well.
> 
> Let me know if the above makes sense and if I've missed something specific to SqlClient.

Thanks, @roji. 

It appears that our quest to mitigate the rush for hotfix releases when a dependency becomes deprecated doesn't have a viable solution, based on your insights and we're left with no choice but to release hotfixes whenever a dependency version is deprecated.

SimonCropp:
@JRahnama 

> It appears that our quest to mitigate the rush for hotfix releases when a dependency becomes deprecated doesn't have a viable solution, based on your insights, and we're left with no choice but to release hotfixes whenever a dependency version is deprecated.

IMO the fix is to remove the unnecessarily dependency: https://github.com/dotnet/SqlClient/issues/110

roji:
@SimonCropp assuming you meant #1108, I don't think that's related. Even if the Azure stuff were split into a different package for use only by Azure users (which is what #1108 is about), that "SqlClient Azure" package would still have the exact same issue. In other words, this would stop affecting *all* SqlClient users, but it would still affect *some*.

But at the end of the day, it seems reasonable/standard that if/when a vulnerability is found in a dependency of your package, you release a new version of your package referencing a new version - IMHO that's simply how things are supposed to work. I'd definitely not want users of my package to always automatically get the latest patch version of my package's dependencies, since that would expose them to sudden bugs and behavior changes without them having done any change on their side.

SimonCropp:
@roji 

> Even if the Azure stuff were split into a different package for use only by Azure users (which is what https://github.com/dotnet/SqlClient/issues/1108 is about), that "SqlClient Azure" package would still have the exact same issue. In other words, this would stop affecting all SqlClient users, but it would still affect some.

that is true. but the split would enabled the packages to have a different release cadence. it would be more reasonable for the "SqlClient Azure" package  to do a release in response to a release of Azure.Identity. in fact the "SqlClient Azure" package  could be shipped from the Azure.Identity in lock step.

tools/props/Versions.props
```diff
@@ -27,8 +27,8 @@
   </PropertyGroup>
   <!-- NetFx and NetCore project dependencies -->
   <PropertyGroup>
-    <AzureIdentityVersion>1.10.3</AzureIdentityVersion>
-    <MicrosoftIdentityClientVersion>4.56.0</MicrosoftIdentityClientVersion>
+    <AzureIdentityVersion>1.11.0</AzureIdentityVersion>
+    <MicrosoftIdentityClientVersion>4.60.1</MicrosoftIdentityClientVersion>
```
David-Engel, line 31: Time to bump these already... 👀 

https://www.nuget.org/packages/Microsoft.Identity.Client/#versions-body-tab

AraHaan:
Or just wildcard the azure dependencies so then every build of SqlClient would pull in the latest Azure package versions.

roji:
@AraHaan please see discussion above about floating versions.

tools/specs/add-ons/Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider.nuspec
```diff
@@ -26,25 +26,25 @@ Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyStoreProvider.SqlColumnEncrypti
     <dependencies>
       <group targetFramework="net462">
         <dependency id="Microsoft.Data.SqlClient" version="5.1.5" />
-        <dependency id="Azure.Core" version="[1.35.0,2.0.0)" />
+        <dependency id="Azure.Core" version="[1.38.0,2.0.0)" />
```
DavoudEshtehari, line 29:
```suggestion
        <dependency id="Azure.Core" version="1.38.0" />
```

tools/specs/add-ons/Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider.nuspec
```diff
@@ -26,25 +26,25 @@ Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyStoreProvider.SqlColumnEncrypti
...
         <dependency id="Azure.Security.KeyVault.Keys" version="[4.5.0,5.0.0)" />
         <dependency id="Microsoft.Extensions.Caching.Memory" version="6.0.1" />
       </group>
       <group targetFramework="net8.0">
         <dependency id="Microsoft.Data.SqlClient" version="5.1.5" />
-        <dependency id="Azure.Core" version="[1.35.0,2.0.0)" />
+        <dependency id="Azure.Core" version="[1.38.0,2.0.0)" />
```
DavoudEshtehari, line 35:
```suggestion
        <dependency id="Azure.Core" version="1.38.0" />
```

tools/specs/add-ons/Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider.nuspec
```diff
@@ -26,25 +26,25 @@ Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyStoreProvider.SqlColumnEncrypti
...
         <dependency id="Azure.Security.KeyVault.Keys" version="[4.5.0,5.0.0)" />
         <dependency id="Microsoft.Extensions.Caching.Memory" version="8.0.0" />
       </group>
       <group targetFramework="net6.0">
         <dependency id="Microsoft.Data.SqlClient" version="5.1.5" />
-        <dependency id="Azure.Core" version="[1.35.0,2.0.0)" />
+        <dependency id="Azure.Core" version="[1.38.0,2.0.0)" />
```
DavoudEshtehari, line 41:
```suggestion
        <dependency id="Azure.Core" version="1.38.0" />
```

tools/specs/add-ons/Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider.nuspec
```diff
@@ -26,25 +26,25 @@ Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyStoreProvider.SqlColumnEncrypti
...
         <dependency id="Azure.Security.KeyVault.Keys" version="[4.5.0,5.0.0)" />
         <dependency id="Microsoft.Extensions.Caching.Memory" version="6.0.1" />
       </group>
       <group targetFramework="netstandard2.0">
         <dependency id="Microsoft.Data.SqlClient" version="5.1.5" />
-        <dependency id="Azure.Core" version="[1.35.0,2.0.0)" />
+        <dependency id="Azure.Core" version="[1.38.0,2.0.0)" />
```
DavoudEshtehari, line 47:
```suggestion
        <dependency id="Azure.Core" version="1.38.0" />
```

tools/specs/add-ons/Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider.nuspec
```diff
@@ -26,25 +26,25 @@ Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyStoreProvider.SqlColumnEncrypti
     <dependencies>
       <group targetFramework="net462">
         <dependency id="Microsoft.Data.SqlClient" version="5.1.5" />
-        <dependency id="Azure.Core" version="[1.35.0,2.0.0)" />
+        <dependency id="Azure.Core" version="[1.38.0,2.0.0)" />
         <dependency id="Azure.Security.KeyVault.Keys" version="[4.5.0,5.0.0)" />
```
DavoudEshtehari, line 30: This version should be compatible with Versions.props!

SimonCropp:
Azure.Identity  is now at 1.11.2

txm229 (approved):

DavoudEshtehari (approved):

BlythMeister:
when is this likely to hit the shelves?

PR #2466, opened 2024-04-17
Change | Separate tests for NetFx and NetCore - NetFx-Only Connection String Properties

benrr101:
**Description**: In an effort to make running all unit tests in vs not throw "correct" test failures, this PR creates two tests.
* `SqlConnectionStringBuilderTest.ConnectionStringTestsNetFx_OnNetCore_Throws`
  * Throws if netfx-only properties are used in the connection string
* `SqlConnectionStringBuilderTest.ConnectionStringTestsNetFx_OnNetFx_Success`
  * Same as original test, validates the property is supported

As a side effect of this, I had to add `Transparent Network IP Resolution` to the list of properties that will return a "not supported *on this platform*" exception. This might be considered a breaking change since it changes an exception type, but I see it more as a fix, since it's true - that property is not supported on netcore (yet?) and that exception should be returned.

**Tests**:
* Manually running the new tests works correctly on netcore and netfx


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionStringBuilder.cs
```diff
@@ -912,17 +911,19 @@ public override StandardValuesCollection GetStandardValues(ITypeDescriptorContex
             }
         }
 #else    
-        private static readonly string[] s_notSupportedKeywords = new string[DeprecatedKeywordsCount] {
+        private static readonly string[] s_notSupportedKeywords = {
```
benrr101, line 914: Array initialization is always the same size as the number of elements, so its unnecessary to specify it.

David-Engel: Since you are eliminating the DeprecatedKeywordsCount reference here, you should just get rid of DeprecatedKeywordsCount entirely and just reference the size of this array in the one other place it's used. Otherwise, DeprecatedKeywordsCount needs to be changed from 3 to 5.

David-Engel (approved):
Approved with the minor comment.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionStringBuilder.cs
benrr101: @David-Engel Ok, I've updated the PR, chaning `DeprecatedKeywordsCount` to 5 and removing `DeprecatedSynonymsCount` from `SqlConnectionString`. As far as I can tell, I can't access `s_notSupportedKeywords`  array from `SqlConnectionString`, so I think it makes more sense to just change the number.

As far as I can tell, the only reason for calculating the count is to determine the optimum size of the dictionary. net8 added support for a `FrozenDictionary` that has its size fixed at initialization, but it looks like we can't utilize that today. That'd be a nice option here...

Please let me know if this isn't what you had in mind.

David-Engel: I wasn't considering DeprecatedSynonymCount, just DeprecatedKeywordsCount. Just make sure the count is correct in the debug assert at line 927/928 of SqlConnectionString.cs after your changes. I'm pretty sure it's there to ensure we are initializing the Dictionary to the correct size to maximize perf. An incorrect count won't show up at runtime or in tests.
`Debug.Assert(synonyms.Count == count, "incorrect initial ParseSynonyms size");`

benrr101: Right. It sounds like we're on the same page, then.

cheenamalhotra (approved):

PR #2474, opened 2024-04-29
Test | Remove AAS - VBS tests from test pipelines

JRahnama:
VBS is not yet supported on AAS.


David-Engel (approved):

DavoudEshtehari (approved):
Should this also be considered a candidate for backporting?

PR #2487, opened 2024-05-07
Fix | Enhance certificate validation

arellegue:
This PR fixes issue #2178. This PR replaces #2439.

The issue was with Certificate Chain Link error, which is now resolved by comparing the binary raw data of the client certificate against the server certificate. If both certificates match, we allow it to continue. Prior to this, any certificate policy error, we stopped operations and threw an exception.

Unit tests were added using TestTdsServer as the SqlServer. Also, different combinations of client connection strings are used for the unit tests. The unit tests only run on Windows and Linux environment.


saurabh500:
Would it not make more sense to validate the hostname in cert, rather than comparing a binary file? In fact, validating hostname is likely more usable, since this prevents the overhead of copying over the cert binaries to the client.

David-Engel:
ServerCertificate was added as a better option to TrustServerCertificate=true. If the certificate of a server is known (self-signed, for example), the client can specify that they trust a specific certificate. When we added TDS 8 / Strict encryption, we also started ignoring TrustServerCertificate when Strict is enabled. So a better dev/test option was needed for self-signed certificate scenarios.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.EndPoint/TDSServerParser.cs
```diff
@@ -69,15 +69,19 @@ public void Run()
...
+                                    (MessageBeingReceived[0] as TDSPreLoginToken).Encryption = TDSPreLoginTokenEncryptionType.None;
+                                else
+                                    (MessageBeingReceived[0] as TDSPreLoginToken).Encryption = TDSPreLoginTokenEncryptionType.Off;
                             }
 
                             // Call into the subscriber to process the packet
                             responseMessages = Server.OnPreLoginRequest(Session, MessageBeingReceived);
 
-                            if (Session.Encryption == TDSEncryptionType.None)
+                            if (Session.Encryption == TDSEncryptionType.None || Session.Encryption == TDSEncryptionType.Off)
```
David-Engel, line 84: I'm not sure about these changes. Even the previous lines from 72-75, I think they might be unnecessary. I think any logic changes here should go in Server.OnPreLoginRequest. I haven't had time to debug this fully, so let me know if you disagree.

arellegue: When a TestTdsServer is created, in the GenericTDSServerSession constructor, the default Encryption = TDSEncryptionType.Off. However, in TDSServerParser.cs class, the Session.Encryption = TDSEncryptionType.Off value was not handled and so, MessageBeingReceived.Encryption and the transport layer encryption are both not getting disabled.

Should I change the default value of Session.Encryption from TDSEncryptionType.Off to TDSEncryptionType.None instead so no changes are necessary for TDSServerParser.cs file?

arellegue: The unnecessary changes were reverted.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectionTestWithSSLCert/CertificateTestWithTdsServer.cs
```diff
@@ -0,0 +1,146 @@
...
+        private static readonly string s_fullPathToPowershellScript = Path.Combine(Directory.GetCurrentDirectory(), "makepfxcert.ps1");
+        private static readonly string s_fullPathToPfx = Path.Combine(Directory.GetCurrentDirectory(), "localhostcert.pfx");
+        private static readonly string s_fullPathToClientCert = Path.Combine(Directory.GetCurrentDirectory(), "clientcert");
+
+        public CertificateTestWithTdsServer()
+        {
+            if (!Directory.Exists(s_fullPathToClientCert))
+            {
+                Directory.CreateDirectory(s_fullPathToClientCert);
+            }
```
JRahnama, line 28: Fail if user does not have elevated access on Windows
you can add something similar to below:
```C#
 // Confirm that user has elevated access on Windows
 if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
 {
     using WindowsIdentity identity = WindowsIdentity.GetCurrent();
     WindowsPrincipal principal = new(identity);
     if (!principal.IsInRole(WindowsBuiltInRole.Administrator))
     {
         Assert.Fail("User needs to have elevated access for these set of tests");
     }
 }
```

Also, is the created certificate disposed at the end?

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectionTestWithSSLCert/CertificateTestWithTdsServer.cs
```diff
@@ -0,0 +1,146 @@
...
+
+            builder = new(server.ConnectionString)
+            {
+                UserID = userId,
+                Password = password,
+                TrustServerCertificate = connectionTestParameters.TrustServerCertificate,
+                Encrypt = connectionTestParameters.Encrypt,
+            };
+
+            if (connectionTestParameters.Certificate != string.Empty)
```
JRahnama, line 68: Safer if you use below, this will avoid any potential NRE
```suggestion
            if (!string.IsNullOrEmpty(connectionTestParameters.Certificate))
```

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectionTestWithSSLCert/CertificateTestWithTdsServer.cs
```diff
@@ -0,0 +1,146 @@
...
+            {
+                builder.ServerCertificate = connectionTestParameters.Certificate;
+            }
+
+            if (connectionTestParameters.HostNameInCertificate != string.Empty)
```
JRahnama, line 73: Same here as above. Let's use the safer approach
```suggestion
            if (!string.IsNullOrEmpty(connectionTestParameters.HostNameInCertificate))
```

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectionTestWithSSLCert/CertificateTestWithTdsServer.cs
```diff
@@ -0,0 +1,146 @@
...
+                        CreateNoWindow = false,
+                        Verb = "runas"
+                    }
+                };
+
+                proc.EnableRaisingEvents = true;
+
+                proc.OutputDataReceived += new DataReceivedEventHandler((sender, e) =>
+                {
+                    output.AppendLine(e.Data);
```
JRahnama, line 120: NIT: e.Data could be null. To add more `DataReceivedEventHandler` is called each time the process writes a line to the redirected StandardOutput stream. When the stream is closed `DataReceivedEventHandler` will be invoked one last time with 3.Data set to null. To avoid appending null values it is better to check for null. 
Something similar to below:
```C#
if (e.Data != null)
{
    output.AppendLine(e.Data);
}
```

JRahnama:
Thanks @arellegue.

Overall looks good and tests are passing. My recent comments are regarding improvements on the tests readability and some enhancements.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
```diff
@@ -0,0 +1,321 @@
...
+{
+    public class ConnectionTestParametersData
+    {
+        private const int CASES = 30;
+        private string _empty = string.Empty;
+        // It was advised to store the client certificate in its own folder.
+        private static readonly string s_fullPathToCer = Path.Combine(Directory.GetCurrentDirectory(), "clientcert", "localhostcert.cer");
+        private static readonly string s_mismatchedcert = Path.Combine(Directory.GetCurrentDirectory(), "clientcert", "mismatchedcert.cer");
+
+        private static string s_hostName = System.Net.Dns.GetHostName();
```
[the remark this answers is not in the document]
arellegue: Readonly was added.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
```diff
@@ -0,0 +1,321 @@
...
+        public ConnectionTestParametersData()
+        {
+            // Test cases possible field values for connection parameters:
+            //     Possible values for TdsEncryptionType are Off, On, Required
+            //     Possible values for Encrypt are Optional, Mandatory
+            //     Possible values for TrustServerCertificate are true, false
+            //     Possible values for Certificate are valid path to certificate, mismatched certificate, or empty
+            //     Possible values for HostNameInCertificate are valid hostname, or empty
+            //     TestResult is the expected result of the connection test
+            // These combinations are based on the possible values of Encrypt, TrustServerCertificate, Certificate, HostNameInCertificate
```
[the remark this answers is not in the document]
arellegue: The table was added as suggessted.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParameters.cs
```diff
@@ -0,0 +1,23 @@
...
+    public class ConnectionTestParameters
+    {
+        public SqlConnectionEncryptOption Encrypt { get; set; }
+        public bool TrustServerCertificate { get; set; }
+        public string Certificate { get; set; }
+        public string HostNameInCertificate { get; set; }
+        public bool TestResult { get; set; }
+        public TDSPreLoginTokenEncryptionType TdsEncryptionType { get; set; }
+    }
+}
```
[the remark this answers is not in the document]
arellegue: Encapsulation pattern was implemented as suggessted.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
```diff
@@ -0,0 +1,321 @@
...
+                    Encrypt = SqlConnectionEncryptOption.Mandatory,
+                    TrustServerCertificate = true,
+                    Certificate = s_mismatchedcert,
+                    HostNameInCertificate = _empty,
+                    TestResult = true
+                },
+            };
+        }
+    }
+}
```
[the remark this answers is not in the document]
arellegue: Test cases constructor was refactored as suggested.

src/Microsoft.Data.SqlClient/tests/ManualTests/Microsoft.Data.SqlClient.ManualTesting.Tests.csproj
```diff
@@ -354,6 +357,12 @@
    </ContentWithTargetPath>
   </ItemGroup>
   <ItemGroup>
+    <None Update="makepfxcert.ps1">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
```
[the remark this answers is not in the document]
arellegue: PreserveNewest is used insted.

src/Microsoft.Data.SqlClient/tests/ManualTests/Microsoft.Data.SqlClient.ManualTesting.Tests.csproj
```diff
@@ -354,6 +357,12 @@
...
+    </None>
+    <None Update="mismatchedcert.cer">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
```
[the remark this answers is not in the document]
arellegue: PreserveNewest is used insted.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectionTestWithSSLCert/CertificateTestWithTdsServer.cs
arellegue: Added Assertion if not running as Administrator as suggested.

Cleaning up of certificates were implemented as suggested.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectionTestWithSSLCert/CertificateTestWithTdsServer.cs
arellegue: The suggestion to use safer method to avoid NRE was implemented.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectionTestWithSSLCert/CertificateTestWithTdsServer.cs
arellegue: The suggestion to use safer method to avoid NRE was implemented.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectionTestWithSSLCert/CertificateTestWithTdsServer.cs
arellegue: Condition to test for null value was implemented as suggested.

JRahnama (approved):

PR #2494, opened 2024-05-08
 Expose SSPI context provider as public

### LINKED ISSUE #2253: Add hook for custom SSPI context for SqlConnection instances

twsouthwick:
> [!IMPORTANT]
> [Stacked](https://stacking.dev/) against #2454. Opened as a draft to enable validation, but since this is on a separate fork, GitHub does not support stacked PRs natively AFAIK

This change:

- Adds a property to SqlConnection to allow setting a provider
- Plumbs that property into the TdsParser so that it can be used if set

Fixes #2253

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
@@ -871,6 +871,8 @@ namespace Microsoft.Data.SqlClient
         [System.ComponentModel.BrowsableAttribute(false)]
         [System.ComponentModel.DesignerSerializationVisibilityAttribute(0)]
         public Microsoft.Data.SqlClient.SqlCredential Credential { get { throw null; } set { } }
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnection.xml' path='docs/members[@name="SqlConnection"]/SSPIContextProviderFactory/*' />
+        public System.Func<SSPIContextProvider> SSPIContextProviderFactory { get { throw null; } set { } }
         /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnection.xml' path='docs/members[@name="SqlConnection"]/Database/*'/>
         [System.ComponentModel.DesignerSerializationVisibilityAttribute(0)]
         public override string Database { get { throw null; } }
@@ -1898,6 +1900,15 @@ namespace Microsoft.Data.SqlClient
         public static SqlRetryLogicBaseProvider CreateFixedRetryProvider(SqlRetryLogicOption retryLogicOption) { throw null; }
         /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConfigurableRetryFactory.xml' path='docs/members[@name="SqlConfigurableRetryFactory"]/CreateNoneRetryProvider/*' />
         public static SqlRetryLogicBaseProvider CreateNoneRetryProvider() { throw null; }
+    }
+    /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SSPIContextProvider.xml' path='docs/members[@name="SSPIContextProvider"]/SSPIContextProvider/*'/>
+    public abstract class SSPIContextProvider
+    {
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SSPIContextProvider.xml' path='docs/members[@name="SSPIContextProvider"]/AuthenticationParameters/*'/>
+        protected SqlAuthenticationParameters AuthenticationParameters { get { throw null; } }
+
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SSPIContextProvider.xml' path='docs/members[@name="SSPIContextProvider"]/GenerateSspiClientContext/*'/>
+        protected abstract void GenerateSspiClientContext(System.ReadOnlySpan<byte> incomingBlob, System.Buffers.IBufferWriter<byte> outgoingBlobWriter);
     }
 }
 namespace Microsoft.Data.SqlClient.Server
--- a/src/Microsoft.Data.SqlClient/netfx/ref/Microsoft.Data.SqlClient.cs
+++ b/src/Microsoft.Data.SqlClient/netfx/ref/Microsoft.Data.SqlClient.cs
@@ -800,6 +800,8 @@ namespace Microsoft.Data.SqlClient
         [System.ComponentModel.BrowsableAttribute(false)]
         [System.ComponentModel.DesignerSerializationVisibilityAttribute(0)]
         public Microsoft.Data.SqlClient.SqlCredential Credential { get { throw null; } set { } }
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnection.xml' path='docs/members[@name="SqlConnection"]/SSPIContextProviderFactory/*' />
+        public System.Func<SSPIContextProvider> SSPIContextProviderFactory { get { throw null; } set { } }
         /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnection.xml' path='docs/members[@name="SqlConnection"]/Database/*'/>
         [System.ComponentModel.DesignerSerializationVisibilityAttribute(0)]
         public override string Database { get { throw null; } }
@@ -1951,6 +1953,15 @@ namespace Microsoft.Data.SqlClient
         public static SqlRetryLogicBaseProvider CreateFixedRetryProvider(SqlRetryLogicOption retryLogicOption) { throw null; }
         /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConfigurableRetryFactory.xml' path='docs/members[@name="SqlConfigurableRetryFactory"]/CreateNoneRetryProvider/*' />
         public static SqlRetryLogicBaseProvider CreateNoneRetryProvider() { throw null; }
+    }
+    /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SSPIContextProvider.xml' path='docs/members[@name="SSPIContextProvider"]/SSPIContextProvider/*'/>
+    public abstract class SSPIContextProvider
+    {
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SSPIContextProvider.xml' path='docs/members[@name="SSPIContextProvider"]/AuthenticationParameters/*'/>
+        protected SqlAuthenticationParameters AuthenticationParameters { get { throw null; } }
+
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SSPIContextProvider.xml' path='docs/members[@name="SSPIContextProvider"]/GenerateSspiClientContext/*'/>
+        protected abstract void GenerateSspiClientContext(System.ReadOnlySpan<byte> incomingBlob, System.Buffers.IBufferWriter<byte> outgoingBlobWriter);
     }
 }
 namespace Microsoft.Data.SqlClient.Server
--- a/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnection.cs
+++ b/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnection.cs
@@ -73,6 +73,8 @@ namespace Microsoft.Data.SqlClient
 
         private Func<SqlAuthenticationParameters, CancellationToken, Task<SqlAuthenticationToken>> _accessTokenCallback;
 
+        private Func<SSPIContextProvider> _sspiContextProviderFactory;
+
         internal bool HasColumnEncryptionKeyStoreProvidersRegistered =>
             _customColumnEncryptionKeyStoreProviders is not null && _customColumnEncryptionKeyStoreProviders.Count > 0;
 
@@ -751,7 +753,7 @@ namespace Microsoft.Data.SqlClient
 
                 _accessToken = value;
                 // Need to call ConnectionString_Set to do proper pool group check
-                ConnectionString_Set(new SqlConnectionPoolKey(_connectionString, _credential, _accessToken, _serverCertificateValidationCallback, _clientCertificateRetrievalCallback, _originalNetworkAddressInfo, null));
+                ConnectionString_Set(new SqlConnectionPoolKey(_connectionString, _credential, _accessToken, _serverCertificateValidationCallback, _clientCertificateRetrievalCallback, _originalNetworkAddressInfo, null, _sspiContextProviderFactory));
             }
         }
 
@@ -773,8 +775,19 @@ namespace Microsoft.Data.SqlClient
                     CheckAndThrowOnInvalidCombinationOfConnectionOptionAndAccessTokenCallback((SqlConnectionString)ConnectionOptions);
                 }
 
-                ConnectionString_Set(new SqlConnectionPoolKey(_connectionString, _credential, null, _serverCertificateValidationCallback, _clientCertificateRetrievalCallback, _originalNetworkAddressInfo, value));
+                ConnectionString_Set(new SqlConnectionPoolKey(_connectionString, _credential, null, _serverCertificateValidationCallback, _clientCertificateRetrievalCallback, _originalNetworkAddressInfo, value, _sspiContextProviderFactory));
                 _accessTokenCallback = value;
+            }
+        }
+
+        /// <include file='../../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnection.xml' path='docs/members[@name="SqlConnection"]/SSPIContextProviderFactory/*' />
+        public Func<SSPIContextProvider> SSPIContextProviderFactory
+        {
+            get { return _sspiContextProviderFactory; }
+            set
+            {
+                ConnectionString_Set(new SqlConnectionPoolKey(_connectionString, _credential, null, _serverCertificateValidationCallback, _clientCertificateRetrievalCallback, _originalNetworkAddressInfo, _accessTokenCallback, value));
+                _sspiContextProviderFactory = value;
             }
         }
 
@@ -862,7 +875,7 @@ namespace Microsoft.Data.SqlClient
                         CheckAndThrowOnInvalidCombinationOfConnectionOptionAndAccessTokenCallback(connectionOptions);
                     }
                 }
-                ConnectionString_Set(new SqlConnectionPoolKey(value, _credential, _accessToken, _serverCertificateValidationCallback, _clientCertificateRetrievalCallback, _originalNetworkAddressInfo, _accessTokenCallback));
+                ConnectionString_Set(new SqlConnectionPoolKey(value, _credential, _accessToken, _serverCertificateValidationCallback, _clientCertificateRetrievalCallback, _originalNetworkAddressInfo, _accessTokenCallback, _sspiContextProviderFactory));
                 _connectionString = value;  // Change _connectionString value only after value is validated
                 CacheConnectionStringProperties();
             }
@@ -1213,7 +1226,7 @@ namespace Microsoft.Data.SqlClient
                 _credential = value;
 
                 // Need to call ConnectionString_Set to do proper pool group check
-                ConnectionString_Set(new SqlConnectionPoolKey(_connectionString, _credential, _accessToken, _serverCertificateValidationCallback, _clientCertificateRetrievalCallback, _originalNetworkAddressInfo, _accessTokenCallback));
+                ConnectionString_Set(new SqlConnectionPoolKey(_connectionString, _credential, _accessToken, _serverCertificateValidationCallback, _clientCertificateRetrievalCallback, _originalNetworkAddressInfo, _accessTokenCallback, _sspiContextProviderFactory));
             }
         }
 
@@ -2775,7 +2788,7 @@ namespace Microsoft.Data.SqlClient
                     throw ADP.InvalidArgumentLength("newPassword", TdsEnums.MAXLEN_NEWPASSWORD);
                 }
 
-                SqlConnectionPoolKey key = new SqlConnectionPoolKey(connectionString, credential: null, accessToken: null, serverCertificateValidationCallback: null, clientCertificateRetrievalCallback: null, originalNetworkAddressInfo: null, accessTokenCallback: null);
+                SqlConnectionPoolKey key = new SqlConnectionPoolKey(connectionString, credential: null, accessToken: null, serverCertificateValidationCallback: null, clientCertificateRetrievalCallback: null, originalNetworkAddressInfo: null, accessTokenCallback: null, sspiContextProviderFactory: null);
 
                 SqlConnectionString connectionOptions = SqlConnectionFactory.FindSqlConnectionOptions(key);
                 if (connectionOptions.IntegratedSecurity || connectionOptions.Authentication == SqlAuthenticationMethod.ActiveDirectoryIntegrated)
@@ -2831,7 +2844,7 @@ namespace Microsoft.Data.SqlClient
                     throw ADP.InvalidArgumentLength("newSecurePassword", TdsEnums.MAXLEN_NEWPASSWORD);
                 }
 
-                SqlConnectionPoolKey key = new SqlConnectionPoolKey(connectionString, credential, accessToken: null, serverCertificateValidationCallback: null, clientCertificateRetrievalCallback: null, originalNetworkAddressInfo: null, accessTokenCallback: null);
+                SqlConnectionPoolKey key = new SqlConnectionPoolKey(connectionString, credential, accessToken: null, serverCertificateValidationCallback: null, clientCertificateRetrievalCallback: null, originalNetworkAddressInfo: null, accessTokenCallback: null, sspiContextProviderFactory: null);
 
                 SqlConnectionString connectionOptions = SqlConnectionFactory.FindSqlConnectionOptions(key);
 
@@ -2876,7 +2889,7 @@ namespace Microsoft.Data.SqlClient
                     throw SQL.ChangePasswordRequires2005();
                 }
             }
-            SqlConnectionPoolKey key = new SqlConnectionPoolKey(connectionString, credential, accessToken: null, serverCertificateValidationCallback: null, clientCertificateRetrievalCallback: null, originalNetworkAddressInfo: null, accessTokenCallback: null);
+            SqlConnectionPoolKey key = new SqlConnectionPoolKey(connectionString, credential, accessToken: null, serverCertificateValidationCallback: null, clientCertificateRetrievalCallback: null, originalNetworkAddressInfo: null, accessTokenCallback: null, sspiContextProviderFactory: null);
 
             SqlConnectionFactory.SingletonInstance.ClearPool(key);
         }
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/NativeSSPIContextProvider.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/NativeSSPIContextProvider.cs
@@ -49,7 +49,7 @@ namespace Microsoft.Data.SqlClient
             }
         }
 
-        protected override void GenerateSspiClientContext(ReadOnlyMemory<byte> incomingBlob, IBufferWriter<byte> outgoingBlobWriter, byte[][] _sniSpnBuffer)
+        protected override void GenerateSspiClientContext(ReadOnlyMemory<byte> incomingBlob, IBufferWriter<byte> outgoingBlobWriter)
         {
 #if NETFRAMEWORK
             SNIHandle handle = _physicalStateObj.Handle;
@@ -60,7 +60,7 @@ namespace Microsoft.Data.SqlClient
 
             var outBuff = outgoingBlobWriter.GetSpan((int)s_maxSSPILength);
 
-            if (0 != SNINativeMethodWrapper.SNISecGenClientContext(handle, incomingBlob.Span, outBuff, out var sendLength, _sniSpnBuffer[0]))
+            if (0 != SNINativeMethodWrapper.SNISecGenClientContext(handle, incomingBlob.Span, outBuff, out var sendLength, AuthenticationParameters.ServerName))
             {
                 throw new InvalidOperationException(SQLMessage.SSPIGenerateError());
             }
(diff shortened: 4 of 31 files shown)


src/Microsoft.Data.SqlClient/netfx/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -1954,6 +1956,12 @@ public sealed class SqlConfigurableRetryFactory
         /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConfigurableRetryFactory.xml' path='docs/members[@name="SqlConfigurableRetryFactory"]/CreateNoneRetryProvider/*' />
         public static SqlRetryLogicBaseProvider CreateNoneRetryProvider() { throw null; }
     }
+    /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SSPIContextProvider.xml' path='docs/members[@name="SSPIContextProvider"]/SSPIContextProvider/*'/>
+    public abstract class SSPIContextProvider
+    {
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SSPIContextProvider.xml' path='docs/members[@name="SSPIContextProvider"]/GenerateSspiClientContext/*'/>
+        protected abstract void GenerateSspiClientContext(System.ReadOnlySpan<byte> incomingBlob, System.Buffers.IBufferWriter<byte> outgoingBlobWriter, SqlAuthenticationParameters authParams);
```
nhart12, line 1963:
```suggestion
        protected abstract bool GenerateSspiClientContext(System.ReadOnlySpan<byte> incomingBlob, System.Buffers.IBufferWriter<byte> outgoingBlobWriter, SqlAuthenticationParameters authParams);
```

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -1938,6 +1940,12 @@ public sealed class SqlConfigurableRetryFactory
         /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConfigurableRetryFactory.xml' path='docs/members[@name="SqlConfigurableRetryFactory"]/CreateNoneRetryProvider/*' />
         public static SqlRetryLogicBaseProvider CreateNoneRetryProvider() { throw null; }
     }
+    /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SSPIContextProvider.xml' path='docs/members[@name="SSPIContextProvider"]/SSPIContextProvider/*'/>
+    public abstract class SSPIContextProvider
+    {
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SSPIContextProvider.xml' path='docs/members[@name="SSPIContextProvider"]/GenerateSspiClientContext/*'/>
+        protected abstract void GenerateSspiClientContext(System.ReadOnlySpan<byte> incomingBlob, System.Buffers.IBufferWriter<byte> outgoingBlobWriter, SqlAuthenticationParameters authParams);
```
nhart12, line 1947:
```suggestion
        protected abstract bool GenerateSspiClientContext(System.ReadOnlySpan<byte> incomingBlob, System.Buffers.IBufferWriter<byte> outgoingBlobWriter, SqlAuthenticationParameters authParams);
```

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/SSPIContextProvider.cs
```diff
@@ -7,7 +7,8 @@
 
 namespace Microsoft.Data.SqlClient
 {
-    internal abstract class SSPIContextProvider
+    /// <include file='../../../../../../../doc/snippets/Microsoft.Data.SqlClient/SSPIContextProvider.xml' path='docs/members[@name="SSPIContextProvider"]/SSPIContextProvider/*'/>
+    public abstract class SSPIContextProvider
```
benrr101, line 11: If we're making this public, I'd really like to change this to "SspiContextProvider" to follow C# naming conventions. I won't demand it be changed all over the codebase, since that would drastically increase the scope of this PR. But since we won't be able to change it easily after we make it public, and I've been slowly fixing naming convention issues like this, we should take this opportunity to fix a long standing naming issue.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/NativeSSPIContextProvider.cs
```diff
@@ -62,9 +62,9 @@ protected override void GenerateSspiClientContext(ReadOnlySpan<byte> incomingBlo
             var sendLength = s_maxSSPILength;
             var outBuff = outgoingBlobWriter.GetSpan((int)sendLength);
 
-            if (0 != SniNativeWrapper.SNISecGenClientContext(handle, incomingBlob, outBuff, ref sendLength, serverSpns[0]))
+            if (0 != SniNativeWrapper.SNISecGenClientContext(handle, incomingBlob, outBuff, ref sendLength, authParams.ServerName))
```
benrr101, line 65: If we're changing this line, can we please flip the yoda condition? :)

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/SSPIContextProvider.cs
```diff
@@ -26,27 +27,82 @@ private protected virtual void Initialize()
...
+        private static SqlAuthenticationParameters CreateSqlAuthParams(SqlInternalConnectionTds connection, string serverSpn)
+        {
+            var auth = new SqlAuthenticationParameters.Builder(
+                authenticationMethod: connection.ConnectionOptions.Authentication,
+                resource: null,
+                authority: null,
+                serverName: serverSpn,
+                connection.ConnectionOptions.InitialCatalog);
+
+            if (connection.ConnectionOptions.UserID is { } userId)
```
benrr101, line 92: I know the pattern matching is the hip new C# language feature, but imho this really impedes readability. I'm guessing this checks for `not null` and assigns to `userId`? Can we change this to `connection.ConnectionOptions.UserID is not null` and just use the property in the following line? It would be much more readable.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionPoolKey.cs
```diff
@@ -33,12 +35,20 @@ internal override string ConnectionString
...
 
-        internal SqlConnectionPoolKey(string connectionString, SqlCredential credential, string accessToken, Func<SqlAuthenticationParameters, CancellationToken, Task<SqlAuthenticationToken>> accessTokenCallback) : base(connectionString)
+        internal SqlConnectionPoolKey(
+            string connectionString,
+            SqlCredential credential,
+            string accessToken,
+            Func<SqlAuthenticationParameters, CancellationToken,
+            Task<SqlAuthenticationToken>> accessTokenCallback,
+            Func<SSPIContextProvider> sspiContextProviderFactory
+            ) : base(connectionString)
```
benrr101, line 45: nit: I thiiiink the convention here is to put the closing paren on the same line as the last argument, then put `: base...` on the next line. Honestly kinda forget. But I'm sure there's an example somewhere else in the codebase

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionPoolKey.cs
```diff
@@ -33,12 +35,20 @@ internal override string ConnectionString
             }
         }
 
-        internal SqlConnectionPoolKey(string connectionString, SqlCredential credential, string accessToken, Func<SqlAuthenticationParameters, CancellationToken, Task<SqlAuthenticationToken>> accessTokenCallback) : base(connectionString)
+        internal SqlConnectionPoolKey(
+            string connectionString,
+            SqlCredential credential,
+            string accessToken,
+            Func<SqlAuthenticationParameters, CancellationToken,
+            Task<SqlAuthenticationToken>> accessTokenCallback,
```
benrr101, line 43: nit: please put this on the previous line - it's really unclear that this is a continuation of the previous line's func types.

benrr101 (dismissed):
Overall, really like the goal and implementation. Most of my comments are style related, since I'm sure my colleagues have tackled the bigger questions.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionPoolKey.cs
```diff
@@ -61,7 +73,8 @@ public override bool Equals(object obj)
                 && _credential == key._credential
                 && ConnectionString == key.ConnectionString
                 && _accessTokenCallback == key._accessTokenCallback
-                && string.CompareOrdinal(_accessToken, key._accessToken) == 0);
+                && string.CompareOrdinal(_accessToken, key._accessToken) == 0
+                && _sspiContextProviderFactory == key._sspiContextProviderFactory);
```
paulmedynski, line 77: Hmm... we're considering two keys equivalent if they use the same factory function, but what if that factory returns SSPIContextProvider instances that behave differently in some meaningful way?

Maybe this is a good reason to just take an SSPIContextProvider instance?

It's not foolproof of course - this whole notion of a pool key based on function/object pointers smells a bit off - but it may be less likely to misbehave.

paulmedynski (changes requested):
Made an initial pass, and came up with some questions/comments.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionPoolKey.cs
David-Engel: Just my 2 cents:

We should probably document our pool key behavior better around the access token callback and this new provider. The callback/provider should be basing its security context solely on the parameters from input parameters, which are all part of the ConnectionString or credential. The same input parameters should always produce the same security context.

I don't think function versus object instance matters for the pool key. If it's the same function or the same object, it should produce the same security context given the same input. On the other side, if all other input parameters are the same but one provider is NTLM and another is MyProviderThatSpoofsWhatever, for example, you would want those to produce two different pools because you wouldn't want MyProviderThatSpoofsWhatever connections to come from pooled connections created with the NTLM provider.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/SSPIContextProvider.cs
twsouthwick: Done in #2454 and carried through here now

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/SSPIContextProvider.cs
twsouthwick: not applicable anymore

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -905,6 +905,8 @@ public void RegisterColumnEncryptionKeyStoreProvidersOnConnection(System.Collect
         [System.ComponentModel.BrowsableAttribute(false)]
         [System.ComponentModel.DesignerSerializationVisibilityAttribute(0)]
         public Microsoft.Data.SqlClient.SqlCredential Credential { get { throw null; } set { } }
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnection.xml' path='docs/members[@name="SqlConnection"]/SSPIContextProviderFactory/*' />
+        public System.Func<SSPIContextProvider> SSPIContextProviderFactory { get { throw null; } set { } }
```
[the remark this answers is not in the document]
twsouthwick: I'm not sure. Initially, there was state that needed to be kept in the factory, but that is no longer the case. However, I'm not sure if we can always rely on that.

I've also been thinking it could just be regular callback of delegate that just take (System.ReadOnlySpan<byte> incomingBlob, System.Buffers.IBufferWriter<byte> outgoingBlobWriter, SqlAuthenticationParameters authParams).

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionPoolKey.cs
paulmedynski: Done in #3339.

paulmedynski (changes requested):
Factory vs instance in the netfx code, and a few questions about docs.

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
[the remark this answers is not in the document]
twsouthwick: I've switched it to just be an instance - lifetime can now be managed by the person who creates it, while internally, we'll just create a new one when necessary. Feels more right to be an instance :)

nhart12 (approved):
Latest package still works for our use-case

twsouthwick:
ping @mdaigle @paulmedynski @benrr101 @David-Engel

doc/snippets/Microsoft.Data.SqlClient/SspiContextProvider.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0"?>
+<docs>
+  <members name="SspiContextProvider">
+    <SspiContextProvider>
+      <summary>Provides the ability to customize SSPI context generation.</summary>
+    </SspiContextProvider>
+    <ctor>
+      <summary>Creates an instance of the SSPIContextProvider.</summary>
+    </ctor>
+    <GenerateContext>
```
mdaigle, line 10: We need some kind of commentary that guides implementers to return a stable identity per resource or they may run into issues with connection pooling. See #3339 for the details we should provide.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnection.cs
```diff
@@ -731,7 +744,7 @@ public string AccessToken
 
                 _accessToken = value;
                 // Need to call ConnectionString_Set to do proper pool group check
-                ConnectionString_Set(new SqlConnectionPoolKey(_connectionString, _credential, _accessToken, null));
+                ConnectionString_Set(new SqlConnectionPoolKey(_connectionString, _credential, _accessToken, null, _sspiContextProvider));
```
mdaigle, line 747: Can be null in this case. If we use a token, we won't use sspi.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnection.cs
```diff
@@ -753,7 +766,7 @@ public Func<SqlAuthenticationParameters, CancellationToken, Task<SqlAuthenticati
                     CheckAndThrowOnInvalidCombinationOfConnectionOptionAndAccessTokenCallback((SqlConnectionString)ConnectionOptions);
                 }
 
-                ConnectionString_Set(new SqlConnectionPoolKey(_connectionString, _credential, null, value));
+                ConnectionString_Set(new SqlConnectionPoolKey(_connectionString, _credential, null, value, _sspiContextProvider));
```
mdaigle, line 769: can also be null here.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnection.cs
```diff
@@ -1066,7 +1079,7 @@ public SqlCredential Credential
                 _credential = value;
 
                 // Need to call ConnectionString_Set to do proper pool group check
-                ConnectionString_Set(new SqlConnectionPoolKey(_connectionString, _credential, accessToken: _accessToken, accessTokenCallback: _accessTokenCallback));
+                ConnectionString_Set(new SqlConnectionPoolKey(_connectionString, _credential, _accessToken, _accessTokenCallback, _sspiContextProvider));
```
mdaigle, line 1082: I _believe_ it can be null here as well

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnection.cs
```diff
@@ -603,6 +605,17 @@ internal int ConnectRetryInterval
             get => ((SqlConnectionString)ConnectionOptions).ConnectRetryInterval;
         }
 
+        /// <include file='../../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnection.xml' path='docs/members[@name="SqlConnection"]/SspiContextProvider/*' />
+        public SspiContextProvider SspiContextProvider
+        {
+            get { return _sspiContextProvider; }
+            set
+            {
+                ConnectionString_Set(new SqlConnectionPoolKey(_connectionString, _credential, null,  _accessTokenCallback, value));
```
mdaigle, line 614: _accessTokenCallback can be null here

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/IntegratedAuthenticationTest/IntegratedAuthenticationTest.cs
```diff
@@ -56,12 +58,63 @@ public static void IntegratedAuthenticationTest_ServerSPN()
...
+            using SqlConnection conn = new(builder.ConnectionString)
+            {
+                SspiContextProvider = new TestSspiContextProvider(),
+            };
+
+            try
+            {
+                conn.Open();
+
+                Assert.Fail("Expected to use custom SSPI context provider");
```
mdaigle, line 83: It would be nice to find a way to test the positive case, so that we know we can successfully connect via a custom context provider.

mdaigle (changes requested):
A few small things I'd like to see addressed. I will also leave this blocked until we complete a security review. We need to do one soon for our upcoming releases anyway.

paulmedynski (approved):

paulmedynski:
@twsouthwick - Need to resolve conflicts here, but I know @benrr101 is also working on merging the .NET and .NET Framework instances of this class, so please coordinate with him.

paulmedynski (approved):

twsouthwick:
@paulmedynski @mdaigle the two failures look to be test failures with sql server setup. otherwise it's ready again

twsouthwick:
@mdaigle @paulmedynski any updates on timing here? Are we still waiting for a security review?

mdaigle:
@twsouthwick yes, we still need to set up a security review. We had originally planned on doing that for the 6.1 release but opted to avoid security sensitive changes. I'm hoping I'll have a chance soon to update out threat model diagrams for this feature and then schedule a feature specific review.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/SspiContextProvider.cs
```diff
@@ -16,6 +17,11 @@ internal abstract class SspiContextProvider
 
         private protected TdsParserStateObject _physicalStateObj = null!;
 
+        /// <include file='../../../../../../../doc/snippets/Microsoft.Data.SqlClient/SspiContextProvider.xml' path='docs/members[@name="SspiContextProvider"]/SspiContextProvider/*'/>
+        protected SspiContextProvider()
+        {
+        }
+
 #if NET
```
mdaigle, line 30: Note: if #3394 merges first, remove this `#if NET`

mdaigle (approved):

twsouthwick:
@mdaigle I've created a separate PR to make the non-public changes so that future merge conflicts are minimal: https://github.com/dotnet/SqlClient/pull/3511

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/IntegratedAuthenticationTest/IntegratedAuthenticationTest.cs
paulmedynski: We should be able to use the test TDS Server for this.  Implement a custom SSPI provider in the driver that just sends "Foo", have the TDS Server check for "Foo" and send back "Bar", and then have the custom provider check for "Bar" and complete the Open().

benrr101:
@twsouthwick I took the liberty of resolving the merge conflict. ~~I'd like to try and get this in by end of week, but I get that's gonna be tight.~~ nevermind we'll get it in ... someday

mdaigle:
@benrr101 we still need a threat model review before we can take this change

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -1979,6 +1981,37 @@ public sealed class SqlConfigurableRetryFactory
...
+        public string Resource { get; }
+
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SspiAuthenticationParameters.xml' path='docs/members[@name="SspiAuthenticationParameters"]/SspiAuthenticationParameters/ServerName'/>
+        public string ServerName { get; }
+
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SspiAuthenticationParameters.xml' path='docs/members[@name="SspiAuthenticationParameters"]/SspiAuthenticationParameters/UserId'/>
+        public string UserId { get; set; }
+
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SspiAuthenticationParameters.xml' path='docs/members[@name="SspiAuthenticationParameters"]/SspiAuthenticationParameters/DatabaseName'/>
+        public string DatabaseName { get; set; }
```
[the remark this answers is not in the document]
paulmedynski: The ref APIs don't use nullable context, so we can't.

doc/snippets/Microsoft.Data.SqlClient/SqlConnection.xml
```diff
@@ -2133,6 +2133,19 @@ The following sample tries to open a connection to an invalid database to simula
         Returns 0 if the connection is inactive on the client side.
       </remarks>
     </ServerProcessId>
+    <SspiContextProvider>
+      <summary>
+        Gets or sets the <see cref="SspiContextProvider"/> instance for customizing the SSPI context. If not set, the default for the platform will be used.
+      </summary>
+      <value>
+        An <see cref="T:Micorosft.Data.SqlClient.SspiContextProvider" /> instance.
```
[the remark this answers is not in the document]
cheenamalhotra: This should be fixed IMO.

mdaigle:
We have a threat model review tentatively slated for the second half of November and I've marked this feature for inclusion in that review.

paulmedynski (approved):

0xced:
@mdaigle Did the threat model review take place?

cheenamalhotra:
> @mdaigle Did the threat model review take place?

We are aiming to get it done in this month.

0xced:
I understand why stale bots exist, still, I hate them.

mdaigle:
Security review is complete. Waiting for builds to pass, then we'll get this final review and approval.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/IntegratedAuthenticationTest/IntegratedAuthenticationTest.cs
mdaigle: We have positive signals from human testing. I'll let an agent spin on this for a bit and see if it can give us a positive test.

mdaigle: Added

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/SspiTests.cs
```diff
@@ -0,0 +1,254 @@
...
+                        Arguments.ServerName);
+                    TDSDoneToken doneToken = new(
+                        TDSDoneTokenStatusType.Final | TDSDoneTokenStatusType.Error);
+                    return new TDSMessageCollection(
+                        new TDSMessage(TDSMessageType.Response, errorToken, doneToken));
+                }
+            }
+        }
+    }
+}
```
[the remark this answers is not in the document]
mdaigle: Sspi is only applicable for integrated auth. It doesn't interact with any other auth methods.

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -1876,6 +1878,39 @@ public sealed class SqlConfigurableRetryFactory
...
+    {
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SspiAuthenticationParameters.xml' path='docs/members[@name="SspiAuthenticationParameters"]/SspiAuthenticationParameters/ctor'/>
+        public SspiAuthenticationParameters(string serverName, string resource)
+        {
+            ServerName = serverName;
+            Resource = resource;
+        }
+
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SspiAuthenticationParameters.xml' path='docs/members[@name="SspiAuthenticationParameters"]/SspiAuthenticationParameters/Resource'/>
+        public string Resource { get; }
```
paulmedynski: Should these propertis all throw?

src/Microsoft.Data.SqlClient/netfx/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -1857,6 +1859,39 @@ public sealed class SqlConfigurableRetryFactory
...
+    {
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SspiAuthenticationParameters.xml' path='docs/members[@name="SspiAuthenticationParameters"]/SspiAuthenticationParameters/ctor'/>
+        public SspiAuthenticationParameters(string serverName, string resource)
+        {
+            ServerName = serverName;
+            Resource = resource;
+        }
+
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SspiAuthenticationParameters.xml' path='docs/members[@name="SspiAuthenticationParameters"]/SspiAuthenticationParameters/Resource'/>
+        public string Resource { get; }
```
paulmedynski: Throw?

PR #2497, opened 2024-05-10
Add scope trace for GenerateSspiClientContext

David-Engel:
Request originated from a question from support when analyzing a trace.

Also fixed some other *EventScope.Create usages.

This should be a pretty easy review.


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/ManagedSSPIContextProvider.cs
```diff
@@ -13,11 +13,14 @@ internal sealed class ManagedSSPIContextProvider : SSPIContextProvider
 
         internal override void GenerateSspiClientContext(ReadOnlyMemory<byte> received, ref byte[] sendBuff, ref uint sendLength, byte[][] _sniSpnBuffer)
         {
-            _sspiClientContextStatus ??= new SspiClientContextStatus();
+            using (TrySNIEventScope.Create(nameof(ManagedSSPIContextProvider)))
```
twsouthwick, line 16: if this is done in the SSPIContextProvider then all calls to GenerateSspiClientClontext would receive telemetry

twsouthwick: Just wanted to point out that this will only give telemetry for times the SSPI is generated via the managed provider. adding telemetry point like [this](https://github.com/dotnet/SqlClient/commit/2b428775612a6d924c03d9a309e8cfc6af3ffa24) would enable it for all generation of SSPI context regardless of provider.

David-Engel: Ah, right. Good call.

JRahnama (approved):

arellegue (approved):

PR #2501, opened 2024-05-15
Merge SqlColumnEncryptionEnclaveProvider, SqlColumnEncryptionCspProvider, SqlColumnEncryptionCngProvider

edwardneal:
Contributes to #1261.

Following the removal of support for .NET Standard, this merges a few classes which are identical (or close to identical) between .NET and .NET Framework.
* SqlColumnEncryptionEnclaveProvider
* SqlColumnEncryptionCspProvider
* SqlColumnEncryptionCngProvider

The only difference between the two sets of files are the SHA256 hashing methodologies. .NET uses `SHA256.Create()`, while .NET Framework explicitly constructs an `SHA256Cng` instance. I don't believe there's any breaking changes between the two, and the latter is incompatible with .NET.

I've not included SqlColumnEncryptionCertificateStoreProvider - it's got slightly different logic between .NET and .NET Framework which I want to keep away from the simple case.

In the issue, I mentioned that we could enable Unix support for the column encryption providers. This PR doesn't do that - once I have a test system and this PR is merged, I can enable that.


David-Engel (approved):

benrr101 (approved):
Looks good to me @edwardneal - if you can get the merge conflicts fixed, I'll get it merged by EOD :)

edwardneal:
Thanks @benrr101 - I've just merged and pushed, so pending any CI surprises we should be good to go.

benrr101:
Thanks again @edwardneal !

PR #2486, opened 2024-05-07
Streamline conditional compilation symbols

ErikEJ:
fixes #2484

PR #2503, opened 2024-05-16
Merge common code base for SqlUtil.cs

arellegue:
Addresses SqlUtil.cs for #1261


src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/SqlUtil.cs
```diff
@@ -271,142 +496,193 @@ internal static Exception CannotGetDTCAddress()
...
         {
             return ADP.DataAdapter(StringsHelper.GetString(Strings.SQL_InvalidSQLServerVersionUnknown));
         }
+
         internal static Exception SynchronousCallMayNotPend()
         {
             return new Exception(StringsHelper.GetString(Strings.Sql_InternalError));
         }
+
+#if NETCOREAPP
```
David-Engel, line 649: Recent PR #2486 streamlined compilation symbols. You should pull from main and then update this.
```suggestion
#if NET6_0_OR_GREATER
```

PR #2521, opened 2024-05-18
Merge SqlColumnEncryptionCertificateStoreProvider

edwardneal:
Contributes to #1261.

This sits alongside #2501, and ports the `SqlColumnEncryptionCertificateStoreProvider` class. Once again, this version uses the .NET approach of `SHA256.Create()` rather than the .NET Framework approach of constructing a new `SHA256Cng` instance.

I've separated this class from the other PR because of the slightly different implementation between .NET and .NET Framework. When the .NET Framework code performs encryption, generates signatures and validates signatures, it references the `PrivateKey` property and casts it to an `RSACryptoServiceProvider`. The .NET code uses the `GetRSAPrivateKey()` method and works with the `RSA` base class instead. This means that the unmerged .NET Framework code can't use certificates with CNG keys; the post-merge implementation will be able to do so.


JRahnama:
Thank you @edwardneal. This looks good as it is.

Jjust a minor request, while you are working on this, is it possible to address the warnings on SqlColumnEncryptionCertificateStoreProvider.Windows class?

Constant names have name rule violation and instead of `var` can you change them to use explicit type or some others such as changing `Int16` to `short` please or that needs a different PR for a different purpose?

David-Engel (approved):

saurabh500 (approved):

PR #2531, opened 2024-05-23
Add JSON Support Feature Extension

deepaksa1:
This PR aims to add a new feature extension to support JSON type in SqlClient.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
@@ -2814,6 +2814,23 @@ namespace Microsoft.Data.SqlClient
                         break;
                     }
 
+                case TdsEnums.FEATUREEXT_JSONSUPPORT:
+                    {
+                        SqlClientEventSource.Log.TryAdvancedTraceEvent("<sc.SqlInternalConnectionTds.OnFeatureExtAck|ADV> {0}, Received feature extension acknowledgement for JSONSUPPORT", ObjectID);
+                        if (data.Length < 1)
+                        {
+                            SqlClientEventSource.Log.TryTraceEvent("<sc.SqlInternalConnectionTds.OnFeatureExtAck|ERR> {0}, Unknown version number for JSONSUPPORT", ObjectID);
+                            throw SQL.ParsingError();
+                        }
+                        byte JSONSupportVersion = data[0];
+                        if (JSONSupportVersion == 0 || JSONSupportVersion > 1)
+                        {
+                            SqlClientEventSource.Log.TryTraceEvent("<sc.SqlInternalConnectionTds.OnFeatureExtAck|ERR> {0}, Invalid version number for JSONSUPPORT", ObjectID);
+                        }
+                        _parser.IsJSONSupportExist = true;
+                        break;
+                    }
+
                 default:
                     {
                         // Unknown feature ack
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
@@ -2817,7 +2817,7 @@ namespace Microsoft.Data.SqlClient
                 case TdsEnums.FEATUREEXT_JSONSUPPORT:
                     {
                         SqlClientEventSource.Log.TryAdvancedTraceEvent("<sc.SqlInternalConnectionTds.OnFeatureExtAck|ADV> {0}, Received feature extension acknowledgement for JSONSUPPORT", ObjectID);
-                        if (data.Length != 2)
+                        if (data.Length != 1)
                         {
                             SqlClientEventSource.Log.TryTraceEvent("<sc.SqlInternalConnectionTds.OnFeatureExtAck|ERR> {0}, Unknown token for JSONSUPPORT", ObjectID);
                             throw SQL.ParsingError(ParsingErrorState.CorruptedTdsStream);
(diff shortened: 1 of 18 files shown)


deepaksa1:
@saurabh500

saurabh500:
@deepaksa1 can we also add the processing of acknowledgement of this feature extension and expose a flag which says whether the connection supports this feature?
You can draw inspiration from the code for data classification

saurabh500:
You will have to make changes to the netfx part of the code too

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsEnums.cs
```diff
@@ -238,6 +238,7 @@ public enum EnvChangeType : byte
         public const byte FEATUREEXT_DATACLASSIFICATION = 0x09;
         public const byte FEATUREEXT_UTF8SUPPORT = 0x0A;
         public const byte FEATUREEXT_SQLDNSCACHING = 0x0B;
+        public const byte FEATUREEXT_JSONSUPPORT = 0x0D;
```
mdaigle, line 241: Left a question on the design doc. Curious why this is 0x0D rather than 0x0C when the last extension was 0x0B.

saurabh500: @mdaigle  usually the server teams could be working on multiple features, and someone whose feature may come out later may have reserved 0x0C already

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsEnums.cs
```diff
@@ -250,7 +251,8 @@ public enum FeatureExtension : uint
             AzureSQLSupport = 1 << (TdsEnums.FEATUREEXT_AZURESQLSUPPORT - 1),
             DataClassification = 1 << (TdsEnums.FEATUREEXT_DATACLASSIFICATION - 1),
             UTF8Support = 1 << (TdsEnums.FEATUREEXT_UTF8SUPPORT - 1),
-            SQLDNSCaching = 1 << (TdsEnums.FEATUREEXT_SQLDNSCACHING - 1)
+            SQLDNSCaching = 1 << (TdsEnums.FEATUREEXT_SQLDNSCACHING - 1),
+            JsonSupport = 1 << (TdsEnums.FEATUREEXT_JSONSUPPORT - 1)
```
apoorvdeshmukh, line 255: Can we name it as JSONSupport  to keep it consistent with other naming conventions such as UTF8Support?

apoorvdeshmukh (commented):
Is there an existing unit test which is testing these WriteXXXFeatureRequest() functions? Do we need changes there?

saurabh500:
> Is there an existing unit test which is testing these WriteXXXFeatureRequest() functions? Do we need changes there?

@apoorvdeshmukh  there are no such unit tests to target these specific functions. However @deepaksa1  perhaps it would be nice to enhance the TDS server which is a mock server to see if this feature extension is received by the server. 

Some examples of test using the mock server are https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlConnectionBasicTests.cs#L25

It might be worth looking at extending TDSFeatureExtAckSessionStateOption  in the tests to provide a response for JSON feature extension as well.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsEnums.cs
benrr101: I would suggest not doing that. Naming conventions for C# are to capitalize the first letter of a word (ie, PascalCase), regardless of whether it is an acronym or not. There could be an argument to maintain the existing precedent, but considering there are mismatched precedents all over the codebase, I think it would be better to follow the C# convention going forward rather than maintain an incorrect precedent.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -2812,6 +2814,28 @@ internal void OnFeatureExtAck(int featureId, byte[] data)
...
+                            SqlClientEventSource.Log.TryTraceEvent("<sc.SqlInternalConnectionTds.OnFeatureExtAck|ERR> {0}, Unknown token for JSONSUPPORT", ObjectID);
+                            throw SQL.ParsingError(ParsingErrorState.CorruptedTdsStream);
+                        }
+                        byte JsonSupportVersion = data[0];            
+                        if (JsonSupportVersion == 0 || JsonSupportVersion > 1)
+                        {
+                            SqlClientEventSource.Log.TryTraceEvent("<sc.SqlInternalConnectionTds.OnFeatureExtAck|ERR> {0}, Invalid version number for JSONSUPPORT", ObjectID);
+                            throw SQL.ParsingError();
+                        }
+                        byte enabled = data[1];
```
saurabh500, line 2831: This can be removed. The server will stop sending you the enabled bit. 
You could check for data length, if it is 1, dont exercise this code path. 

later with the newer build of the server, you can delete this code path.

saurabh500:
You need to make changes to https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs as well and the related files for netfx.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/GenericTDSServer.cs
```diff
@@ -544,6 +550,33 @@ protected virtual TDSMessageCollection OnAuthenticationCompleted(ITDSServerSessi
                 responseMessage.Add(featureExtActToken);
             }
 
+            // Check if Json is supported
+            if (session.IsJsonSupportEnabled)
+            {
+                // Create ack data (2 bytes: Version number, Enabled)
+                byte[] data = new byte[2];
+                data[0] = (byte)1;
+                data[1] = (byte)1;
```
saurabh500, line 559: @deepaksa1  I thought we were going to remove the Enabled byte.
Since you can now test with this TestTds server, it would make sense to change the product code to expect only 1 byte data which is the version and remove data[1]

deepaksa1: Done

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
deepaksa1: Done

PR #2553, opened 2024-06-05
Fix | TvpTest Split Phase 1

arellegue:
This is the first phase of the TvpTest split.

In this PR, the AsyncDebugScope tests were extracted from the Run function of StreamInputParam class and new unit tests were created from it. The TvpTest Main test calls StreamInputParam.Run() to run the tests.

### DIFF
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ParameterTest/StreamInputParam.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ParameterTest/StreamInputParam.cs
@@ -5,11 +5,13 @@
 using System;
 using System.Data;
 using System.Diagnostics;
+using System.Globalization;
 using System.IO;
 using System.Text;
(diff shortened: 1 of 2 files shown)


src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ParameterTest/StreamInputParam.cs
```diff
@@ -23,6 +25,371 @@ internal class CustomStreamException : Exception
         {
         }
 
+        #region AsyncDebugScopeTests
+#if DEBUG
```
paulmedynski, line 29: None of our CI runs tests in DEBUG currently, so we may get some surprises when we enable that :)

paulmedynski (approved):

paulmedynski:
Closing due to inactivity.  We may resurrect these as part of a larger test revitalization effort.

PR #2557, opened 2024-06-07
Add | Add FIC + MSI support

### LINKED ISSUE #2556: Federated Identity Credentials + Managed Identity support
Currently FIC is supported by using workload identity, however it's not compatible with (or is there any existing solutions?) using manage identity directly (FIC + MSI), we need to write our own `SqlAuthenticationProvider` everywhere and overrides existing authentication method.

Add a new authentication method to support FIC + MSI.
Basically we'll need a `SqlAuthenticationProvider` that get authentication results using `ClientAssertionCredential` and `ManagedIdentityCredential`:

```cs
internal class FICAuthProvider : SqlAuthenticationProvider
{
    private const string DefaultScopeSuffix = "/.default";

    public override async Task<SqlAuthenticationToken> AcquireTokenAsync(SqlAuthenticationParameters parameters)
    {
        // We can reuse existing environment variables or use new ones
        var msiClientId = Environment.GetEnvironmentVariable("SOME_NEW_ENV_KEY");
        var aadAppClientId = Environment.GetEnvironmentVariable("AZURE_CLIENT_ID");
        var resourceTenantId = Environment.GetEnvironmentVariable("AZURE_TENANT_ID");

        var assertion = new ClientAssertionCredential(
            resourceTenantId,
            aadAppClientId,
            async (token) => await GetManagedIdentityToken(msiClientId));

        var scopes = new[] { parameters.Resource.EndsWith(DefaultScopeSuffix) ? parameters.Resource : parameters.Resource + DefaultScopeSuffix };

        var token = await assertion.GetTokenAsync(new TokenRequestContext(scopes));
        return new SqlAuthenticationToken(token.Token, token.ExpiresOn);

        static async Task<string> GetManagedIdentityToken(string msiClientId)
        {
            return (await new ManagedIdentityCredential(msiClientId).GetTokenAsync(new Azure.Core.TokenRequestContext([$"api://AzureADTokenExchange/.default"])).ConfigureAwait(false)).Token;
        }
    }

    public override bool IsSupported(SqlAuthenticationMethod authenticationMethod)
    {
        // Add new member for SqlAuthenticationMethod
        return authenticationMethod == SqlAuthenticationMethod.FederatedIdentityCredentials;
    }
}
```

N/A

### Additional context
N/A

yhvicey:
Resolves #2556.

This PR adds support for authenticating Azure SQL Server using federated identity credentials (FIC) and managed identity (MSI). It provides similar functionality as the `ActiveDirectoryWorkloadIdentity`, but supports using managed identity directly.

It will use values from below environments to finish authentication process:

- `AZURE_CLIENT_ID`, the client id of multi-tenant Microsoft Entra App id, may be overridden by `User Id`.
- `AZURE_TENANT_ID`, the tenant id where target resource locates.
- `AZURE_MSI_CLIENT_ID`, the client id of managed identity.

Note: The first two environment variables are defined in `Azure.Identity` library, the third one is newly defined.


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ActiveDirectoryAuthenticationProvider.cs
```diff
@@ -198,6 +199,49 @@ public override async Task<SqlAuthenticationToken> AcquireTokenAsync(SqlAuthenti
                 return new SqlAuthenticationToken(accessToken.Token, accessToken.ExpiresOn);
             }
 
+            if (parameters.AuthenticationMethod == SqlAuthenticationMethod.ActiveDirectoryFederatedIdentityCredentials)
```
David-Engel, line 202: Heads up: This is going to conflict with #2380 and will need to be re-done to cache ClientAssertionCredential objects.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ActiveDirectoryAuthenticationProvider.cs
```diff
@@ -198,6 +199,49 @@ public override async Task<SqlAuthenticationToken> AcquireTokenAsync(SqlAuthenti
...
+                        {
+                            ManagedIdentityCredential miCredential = new ManagedIdentityCredential(managedIdentityClientId);
+                            var token = await miCredential.GetTokenAsync(new TokenRequestContext(new string[] { $"api://AzureADTokenExchange/.default" })).ConfigureAwait(false);
+                            return token.Token;
+                        });
+                }
+
+                if (clientAssertionCredential == null)
+                {
+                    throw new CredentialUnavailableException("FederatedIdentityCredentials authentication unavailable. The required environment variables are not fully configured.");
```
David-Engel, line 238: This string will need to be moved to string resources for localization.

David-Engel (commented):
Just a couple notes for now.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ActiveDirectoryAuthenticationProvider.cs
yhvicey: Done

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ActiveDirectoryAuthenticationProvider.cs
yhvicey: Any ETA when will #2380 will be merged?

JRahnama: > Any ETA when will #2380 be merged?

today, there was one failing test on the CI pipelines, waiting for the rerun to pass.

JRahnama: @yhvicey done. It is merged.

yhvicey: Looks like current implementation only supports credential types already defined in `Azure.Identity` lib - is this expected? As FIC + MSI is not an official credential type in `Azure.Identity` yet

David-Engel: I'm not sure what you are asking. Why can't you do something similar to the below and cache the token credential objects being created?

yhvicey: I mean, in current implementation, there's a `TokenCredentialKey` type was added as the key for caching, and one of its equality keys is `Type _tokenCredentialType`; and for all other cached token credentials, this field points to corresponding credential type in `Azure.Identity` lib, e.g. for Workload Identity:

```cs
if (parameters.AuthenticationMethod == SqlAuthenticationMethod.ActiveDirectoryWorkloadIdentity)
{
    // Cache WorkloadIdentityCredential based on authority and clientId
    TokenCredentialKey tokenCredentialKey = new(typeof(WorkloadIdentityCredential), authority, string.Empty, string.Empty, clientId);
    // If either tenant id, client id, or the token file path are not specified when fetching the token,
    // a CredentialUnavailableException will be thrown instead
    AccessToken accessToken = await GetTokenAsync(tokenCredentialKey, string.Empty, tokenRequestContext, cts.Token).ConfigureAwait(false);
    SqlClientEventSource.Log.TryTraceEvent("AcquireTokenAsync | Acquired access token for Workload Identity auth mode. Expiry Time: {0}", accessToken.ExpiresOn);
    return new SqlAuthenticationToken(accessToken.Token, accessToken.ExpiresOn);
}
```

So to follow existing code/pattern I suppose I should use a credential type representing corresponding authentication method as the cache key, right?
And if so, this PR needs to wait until we add corresponding authentication method in `Azure.Identity` lib.
So,
1. Shall we follow the same pattern (use corresponding credential type as part of the cache key) and wait until corresponding credential type have been added to `Azure.Identity`?
2. Or, use some special type reference as the cache key as a workaround?
3. Or, refactor the cache key object to support cases like this?

David-Engel: In your current code, you have:
```
                    clientAssertionCredential = new ClientAssertionCredential(
                        tenantId,
                        clientId,
                        async (cancellationToken) =>
                        {
                            ManagedIdentityCredential miCredential = new ManagedIdentityCredential(managedIdentityClientId);
                            var token = await miCredential.GetTokenAsync(new TokenRequestContext(new string[] { $"api://AzureADTokenExchange/.default" })).ConfigureAwait(false);
                            return token.Token;
                        });
```

Instead of creating a new ManagedIdentityCredential, you want to use (or create) one from the cache. Then you will also cache the ClientAssertionCredential for re-use. You just need to work out what changes are needed (if any) to the TokenCredentialKey for your new logic. The _tokenCredentialType will be ClientAssertionCredential and the other values will have to be enough information to get or create the correct ManagedIdentityCredential object.

yhvicey: The implementation of `WorkloadIdentityCredential` used `ClientAssertionCredential` as well so I was wondering if we should add a new credential type (just like `WorkloadIdentityCredential`). NVM - I'll use `ClientAssertionCredential` as part the cache key and finish the PR first

yhvicey:
Shall we use `ActiveDirectoryFederatedIdentityCredentials`, or `ActiveDirectoryFederatedIdentity`?

srxmsft:
Since it might take some time for this PR to be release to Microsoft.Data.SqlClient.
For the cross tenant FIC scenario, do you suggest to manually set AccessToken or AccessTokenCallback as of now?
Is there any example we can follow?

yhvicey:
If you're looking for a temporary workaround/solution, ping @yiwwan at Teams, I'll share workaround solution.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ActiveDirectoryAuthenticationProvider.cs
```diff
@@ -611,6 +621,50 @@ private static TokenCredentialData CreateTokenCredentialInstance(TokenCredential
...
+                    clientAssertionCredential = new ClientAssertionCredential(
+                        tenantId,
+                        clientId,
+                        async (cancellationToken) =>
+                        {
+                            ManagedIdentityCredential miCredential = new ManagedIdentityCredential(managedIdentityClientId);
+                            AccessToken token = await miCredential.GetTokenAsync(
+                                new TokenRequestContext(new string[] { $"api://AzureADTokenExchange/.default" }),
+                                cancellationToken).ConfigureAwait(false);
+                            return token.Token;
```
David-Engel, line 657: We should be able to cache and re-use a Managed Identity credential here, too. Something like this, I think (definitely double check this code):
```suggestion
                            string scope = "api://AzureADTokenExchange/.default";
                            TokenRequestContext tokenRequestContext = new TokenRequestContext(new string[] { scope });
                            TokenCredentialKey tokenCredentialKey = new(typeof(ManagedIdentityCredential), tokenCredentialKey._authority, scope, string.Empty, managedIdentityClientId);
                            AccessToken token = await GetTokenAsync(
                                tokenCredentialKey,
                                string.Empty,
                                tokenRequestContext,
                                cancellationToken).ConfigureAwait(false);
                            return token.Token;
```

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ActiveDirectoryAuthenticationProvider.cs
```diff
@@ -611,6 +621,50 @@ private static TokenCredentialData CreateTokenCredentialInstance(TokenCredential
...
+                        });
+                }
+
+                if (clientAssertionCredential == null)
+                {
+                    throw new CredentialUnavailableException(Strings.AAD_FIC_Invalid_Setup);
+                }
+
+                return new TokenCredentialData(clientAssertionCredential, GetHash(secret));
```
David-Engel, line 666: There's no secret in this flow, right?
```suggestion
                return new TokenCredentialData(clientAssertionCredential, null);
```

src/Microsoft.Data.SqlClient/src/Resources/Strings.de.resx
```diff
@@ -4737,4 +4737,7 @@
   <data name="SQL_RemoteCertificateNotAvailable" xml:space="preserve">
     <value>Das Zertifikat ist während der Überprüfung des Zertifikats nicht verfügbar.</value>
   </data>
+  <data name="AAD_FIC_Invalid_Setup" xml:space="preserve">
```
David-Engel, line 4740: There's no need to provide any non-English strings. The English strings are translated via an automated localization process after the PR is merged to main.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ActiveDirectoryAuthenticationProvider.cs
```diff
@@ -611,6 +621,50 @@ private static TokenCredentialData CreateTokenCredentialInstance(TokenCredential
...
+
+                string clientId = tokenCredentialKey._clientId;
+                if (clientId is null)
+                {
+                    clientId = Environment.GetEnvironmentVariable("AZURE_CLIENT_ID");
+                    clientId = string.IsNullOrWhiteSpace(clientId) ? null : clientId;
+                }
+
+                string managedIdentityClientId = Environment.GetEnvironmentVariable("AZURE_MSI_CLIENT_ID");
+                managedIdentityClientId = string.IsNullOrWhiteSpace(managedIdentityClientId) ? null : managedIdentityClientId;
```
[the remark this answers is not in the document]
yhvicey: Updated

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ActiveDirectoryAuthenticationProvider.cs
yhvicey: Updated

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ActiveDirectoryAuthenticationProvider.cs
yhvicey: Updated

src/Microsoft.Data.SqlClient/src/Resources/Strings.de.resx
yhvicey: Non-English strings removed

DavoudEshtehari:
@yhvicey I couldn't find the `AZURE_MSI_CLIENT_ID` environment variable in the [documentation](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/identity/Azure.Identity/README.md#environment-variables) or the [source code](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/identity/Azure.Identity/src/EnvironmentVariables.cs). I might be looking in the wrong places; could you please provide me with the correct links?

yhvicey:
@DavoudEshtehari it's not official env var used by azure.identity (see PR's description), I also worried about what if azure.identity added support for MSI + FIC in the future and didn't use same env var name. Maybe we can switch to a ADO.NET specific env var name.
Or, we can add MSI + FIC in azure.identity first, then leverage it here directly, but that'll took much longer time, I expect

DavoudEshtehari:
> @DavoudEshtehari it's not official env var used by azure.identity (see PR's description), I also worried about what if azure.identity added support for MSI + FIC in the future and didn't use same env var name. Maybe we can switch to a ADO.NET specific env var name. Or, we can add MSI + FIC in azure.identity first, then leverage it here directly, but that'll took much longer time, I expect

We usually don't accept unofficial dependencies to avoid the breaking changes.
cc @saurabh500 @David-Engel

cheenamalhotra:
> Note: The first two environment variables are defined in Azure.Identity library, the third one is newly defined.

What is the plan to document this new env variable? If this is something that will take time, I would recommend converting this PR to draft or closing if this is not in plan. But the driver only consumes documented APIs, or will not expose a feature until the behavior is documented use of the same.

emil-eklund:
> If you're looking for a temporary workaround/solution, ping @yiwwan at Teams, I'll share workaround solution.

Could you please share this in a gist? My external account does not allow me to ping you.

cheenamalhotra:
This feature inclusion is not in plan anytime soon, this PR is also out-of sync with the main branch as a lot has changed in v7.0 as we split the Azure dependencies in an extension package. 

I would recommend making a feature request in issues and tracking it instead, if this is an important business need.

PR #2559, opened 2024-06-10
Fix | Fixed GenerateSspiClientContext to retry negotiation with default port

arellegue:
This PR fixes Issue #2394 

When an application is targeting .Net 6.0, the application connects to the database successfully from Linux because the application is using System.Net.Security.NegotiateStreamPal.Unix.InitializeSecurityContext function which is retrying the negotiation with default port number 1433. However, when the application is targeting .Net8.0, the application fails to connect to the database from Linux because the newer implementation, MDS.SSPI.NegotiateSSPIContextProvider, did not retry with default port number. This issue is fixed in this PR.

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/NegotiateSSPIContextProvider.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/NegotiateSSPIContextProvider.cs
@@ -22,7 +22,7 @@ namespace Microsoft.Data.SqlClient
                 _negotiateAuth ??= new(new NegotiateAuthenticationClientOptions { Package = "Negotiate", TargetName = spnName });
                 sendBuff = _negotiateAuth.GetOutgoingBlob(received.Span, out statusCode)!;
                 // Log session id, status code and the actual SPN used in the negotiation
-                SqlClientEventSource.Log.TryTraceEvent("TdsParserStateObjectManaged.GenerateSspiClientContext | Info | Session Id {0}, StatusCode={1}, SPN={2}", _physicalStateObj.SessionId, statusCode, _negotiateAuth.TargetName);
+                SqlClientEventSource.Log.TryTraceEvent($"TdsParserStateObjectManaged.GenerateSspiClientContext | Info | Session Id {_physicalStateObj.SessionId}, StatusCode={statusCode}, SPN={_negotiateAuth.TargetName}");
 
                 if (statusCode == NegotiateAuthenticationStatusCode.Completed || statusCode == NegotiateAuthenticationStatusCode.ContinueNeeded)
                     break; // Successful case, exit the loop with current SPN.
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/NegotiateSSPIContextProvider.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/NegotiateSSPIContextProvider.cs
@@ -16,14 +16,16 @@ namespace Microsoft.Data.SqlClient
         {
             NegotiateAuthenticationStatusCode statusCode = NegotiateAuthenticationStatusCode.UnknownCredentials;
 
-            foreach (byte[] spn in _sniSpnBuffer)
+            for (int i = 0; i < _sniSpnBuffer.Length; i++)
             {
-                _negotiateAuth ??= new(new NegotiateAuthenticationClientOptions { Package = "Negotiate", TargetName = Encoding.Unicode.GetString(spn) });
+                string spnName = Encoding.Unicode.GetString(_sniSpnBuffer[i]);
+                _negotiateAuth ??= new(new NegotiateAuthenticationClientOptions { Package = "Negotiate", TargetName = spnName });
                 sendBuff = _negotiateAuth.GetOutgoingBlob(received.Span, out statusCode)!;
-                SqlClientEventSource.Log.TryTraceEvent("TdsParserStateObjectManaged.GenerateSspiClientContext | Info | Session Id {0}, StatusCode={1}", _physicalStateObj.SessionId, statusCode);
+                // Log session id, status code and the actual SPN used in the negotiation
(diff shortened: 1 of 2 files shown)


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/NegotiateSSPIContextProvider.cs
```diff
@@ -14,13 +14,25 @@ internal sealed class NegotiateSSPIContextProvider : SSPIContextProvider
 
         internal override void GenerateSspiClientContext(ReadOnlyMemory<byte> received, ref byte[] sendBuff, ref uint sendLength, byte[][] _sniSpnBuffer)
         {
-            _negotiateAuth ??= new(new NegotiateAuthenticationClientOptions { Package = "Negotiate", TargetName = Encoding.Unicode.GetString(_sniSpnBuffer[0]) });
-            sendBuff = _negotiateAuth.GetOutgoingBlob(received.Span, out NegotiateAuthenticationStatusCode statusCode)!;
-            SqlClientEventSource.Log.TryTraceEvent("TdsParserStateObjectManaged.GenerateSspiClientContext | Info | Session Id {0}, StatusCode={1}", _physicalStateObj.SessionId, statusCode);
+            NegotiateAuthenticationStatusCode statusCode = NegotiateAuthenticationStatusCode.UnknownCredentials;
+
+            foreach (byte[] spn in _sniSpnBuffer)
```
DavoudEshtehari, line 19: I'd rather not creating a new byte array variable and directly using `_sniSpnBuffer` inside the classic for-loop to avoid the extra stack operations.
```
for (int i = 0; i < _sniSpnBuffer.Length; i++)
{
    _sniSpnBuffer[i]; 
}
```

DavoudEshtehari (commented):
@David-Engel Do you see any straight forward solution to add unit test in this case?

David-Engel:
> Do you see any straight forward solution to add unit test for this case?

Straightforward, no.  It would be an edge case that would require specific environment setup and conditional checking on the test. I'd rather just get the fix in than make it wait for a test.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/NegotiateSSPIContextProvider.cs
```diff
@@ -14,13 +14,27 @@ internal sealed class NegotiateSSPIContextProvider : SSPIContextProvider
...
-            SqlClientEventSource.Log.TryTraceEvent("TdsParserStateObjectManaged.GenerateSspiClientContext | Info | Session Id {0}, StatusCode={1}", _physicalStateObj.SessionId, statusCode);
+            NegotiateAuthenticationStatusCode statusCode = NegotiateAuthenticationStatusCode.UnknownCredentials;
+
+            for (int i = 0; i < _sniSpnBuffer.Length; i++)
+            {
+                string spnName = Encoding.Unicode.GetString(_sniSpnBuffer[i]);
+                _negotiateAuth ??= new(new NegotiateAuthenticationClientOptions { Package = "Negotiate", TargetName = spnName });
+                sendBuff = _negotiateAuth.GetOutgoingBlob(received.Span, out statusCode)!;
+                // Log session id, status code and the actual SPN used in the negotiation
+                SqlClientEventSource.Log.TryTraceEvent($"TdsParserStateObjectManaged.GenerateSspiClientContext | Info | Session Id {_physicalStateObj.SessionId}, StatusCode={statusCode}, SPN={_negotiateAuth.TargetName}");
```
DavoudEshtehari, line 25: The previous style of `TryTraceEvent` has internal null checking and it's safer than this style in addition to the higher perf when trace is not enabled.
```suggestion
                SqlClientEventSource.Log.TryTraceEvent("TdsParserStateObjectManaged.GenerateSspiClientContext | Info | Session Id {0}, StatusCode={1}, SPN={2}", _physicalStateObj.SessionId, statusCode, spnName);
```

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/NegotiateSSPIContextProvider.cs
```diff
@@ -14,13 +14,27 @@ internal sealed class NegotiateSSPIContextProvider : SSPIContextProvider
...
         {
-            _negotiateAuth ??= new(new NegotiateAuthenticationClientOptions { Package = "Negotiate", TargetName = Encoding.Unicode.GetString(_sniSpnBuffer[0]) });
-            sendBuff = _negotiateAuth.GetOutgoingBlob(received.Span, out NegotiateAuthenticationStatusCode statusCode)!;
-            SqlClientEventSource.Log.TryTraceEvent("TdsParserStateObjectManaged.GenerateSspiClientContext | Info | Session Id {0}, StatusCode={1}", _physicalStateObj.SessionId, statusCode);
+            NegotiateAuthenticationStatusCode statusCode = NegotiateAuthenticationStatusCode.UnknownCredentials;
+
+            for (int i = 0; i < _sniSpnBuffer.Length; i++)
+            {
+                string spnName = Encoding.Unicode.GetString(_sniSpnBuffer[i]);
+                _negotiateAuth ??= new(new NegotiateAuthenticationClientOptions { Package = "Negotiate", TargetName = spnName });
```
DavoudEshtehari, line 22: If you don't tend to use this variable, then it can be removed.
```suggestion
                _negotiateAuth ??= new(new NegotiateAuthenticationClientOptions { Package = "Negotiate", TargetName = Encoding.Unicode.GetString(_sniSpnBuffer[i]) });
```

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/NegotiateSSPIContextProvider.cs
DavoudEshtehari: It seems you accidentally reverted this change!

DavoudEshtehari (approved):

PR #2561, opened 2024-06-10
Adding the Tds Stream APIs

saurabh500:
This PR introduces the Tds Stream, which will be consumed by the upper layers of the Tds Stream to read data, without needing to worry about the Tds header or the packet structure.
The Stream will allow reads and writes of the data. 

There are few additional method in the API which are especially interesting for the consumers of the TDS protocol.

1. PeekByteAsync : Some callsites like the Sql Data Reader would need to peek the next byte (token) before it can really decide what action to take. 
2. SkipReadBytesAsync: In some cases the Tds consumer needs to skip some bytes in the stream.
3. Reset() : This method allows resetting the buffer pointers in the stream. 

Adding only the APIs, and not the implementation, to allow the reviewers to understand the usage of this stream. 

There will be some of the APIs from the stream which will not be implemented at all.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
@@ -12,18 +12,34 @@ namespace Microsoft.Data.SqlClientX.IO
     /// A Stream abstraction over the TDS protocol.
     /// The stream can be used to read and write TDS messages on a 
     /// SQL Server physical connection.
+    /// The stream is responsible for abstracting away the calls to IO and handling the packet 
+    /// header from the consumers of TDS protocol.
+    /// The stream responds to the request of the callers, but it doesn't guarantee the correctness of the underlying TDS protocol correctness.
+    /// e.g. If the protocol states that there are N bytes that should be read, and the stream is asked to return
+    /// N + 1 bytes, then the stream will timeout trying to get N+1 bytes, or it will return N+1 bytes, if the 
+    /// N+1 byte is available.
     /// </summary>
     internal class TdsStream : Stream
     {
+        /// <inheritdoc />
         public override bool CanRead => throw new NotImplementedException();
-
+        
+        /// <inheritdoc />
         public override bool CanSeek => throw new NotImplementedException();
 
+        /// <inheritdoc />
         public override bool CanWrite => throw new NotImplementedException();
 
+        /// <inheritdoc />
         public override long Length => throw new NotImplementedException();
 
+        /// <inheritdoc />
         public override long Position { get => throw new NotImplementedException(); set => throw new NotImplementedException(); }
+
+        /// <summary>
+        /// Indicates if the cancellation is sent to the server.
+        /// </summary>
+        public virtual bool IsCancellationSent { get; internal set; }
 
         public TdsStream(Stream underLyingStream) : base()
         {
@@ -37,7 +53,18 @@ namespace Microsoft.Data.SqlClientX.IO
         /// </summary>
         /// <param name="stream"></param>
         /// <exception cref="NotImplementedException"></exception>
-        public void ReplaceUnderlyingStream(Stream stream)
+        public virtual void ReplaceUnderlyingStream(Stream stream)
+        {
+            throw new NotImplementedException();
+        }
+
+        /// <summary>
+        /// Allows manipulation of the packet size. This should only be used after login ack,
+        /// when the client and server exchange the negotiated packet size.
+        /// </summary>
+        /// <param name="packetSize">The negotiated packet size</param>
+        /// <exception cref="NotImplementedException"></exception>
+        public virtual void SetPacketSize(int packetSize)
         {
             throw new NotImplementedException();
         }
@@ -48,22 +75,25 @@ namespace Microsoft.Data.SqlClientX.IO
         /// TODO: Consider accepting an enum of packet types
         /// instead of the byte.
         /// </summary>
-        /// <param name="packetType"></param>
-        public void SetWritePacketType(TdsStreamPacketType packetType)
+        /// <param name="packetType">The type of packet to be sent out</param>
+        public virtual void SetWritePacketType(TdsStreamPacketType packetType)
         {
             throw new NotImplementedException();
         }
 
+        /// <inheritdoc />
         public override void Flush()
         {
             throw new NotImplementedException();
         }
 
+        /// <inheritdoc />
         public override int Read(Span<byte> buffer)
         {
             throw new NotImplementedException();
         }
 
+        /// <inheritdoc />
         public override int Read(
             byte[] buffer, 
             int offset, 
@@ -72,21 +102,31 @@ namespace Microsoft.Data.SqlClientX.IO
             throw new NotImplementedException();
         }
 
+        /// <inheritdoc />
         public override long Seek(long offset, SeekOrigin origin)
         {
             throw new NotImplementedException();
         }
 
+        /// <inheritdoc />
         public override void SetLength(long value)
         {
             throw new NotImplementedException();
         }
 
+        /// <inheritdoc />
+        public override void Write(ReadOnlySpan<byte> buffer)
+        {
+            throw new NotImplementedException();
+        }
+
+        /// <inheritdoc />
         public override void Write(byte[] buffer, int offset, int count)
         {
             throw new NotImplementedException();
         }
 
+        /// <inheritdoc />
         public override ValueTask WriteAsync(
             ReadOnlyMemory<byte> buffer,
             CancellationToken cancellationToken)
@@ -94,17 +134,7 @@ namespace Microsoft.Data.SqlClientX.IO
             throw new NotImplementedException();
         }
 
-        /// <summary>
-        /// We only support ValueTask based overloads for async IO.
-        /// This overload makes sure that the method throws, instead 
-        /// of using the unoptimized path.
-        /// </summary>
-        /// <param name="buffer"></param>
-        /// <param name="offset"></param>
-        /// <param name="count"></param>
-        /// <param name="cancellationToken"></param>
-        /// <returns></returns>
-        /// <exception cref="NotSupportedException"></exception>
+        /// <inheritdoc />
         public override Task WriteAsync(
             byte[] buffer,
             int offset,
@@ -127,6 +157,7 @@ namespace Microsoft.Data.SqlClientX.IO
             throw new NotImplementedException();
         }
 
+        /// <inheritdoc />
         public override ValueTask<int> ReadAsync(
             Memory<byte> buffer,
             CancellationToken cancellationToken)
@@ -143,7 +174,7 @@ namespace Microsoft.Data.SqlClientX.IO
         /// <param name="ct"></param>
         /// <returns></returns>
         /// <exception cref="NotImplementedException"></exception>
-        public ValueTask<byte> PeekByteAsync(bool isAsync,
+        public virtual ValueTask<byte> PeekByteAsync(bool isAsync,
             CancellationToken ct)
         {
             throw new NotImplementedException();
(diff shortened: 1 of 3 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
```diff
@@ -0,0 +1,168 @@
...
+using System.Threading.Tasks;
+
+namespace Microsoft.Data.SqlClientX.IO
+{
+    /// <summary>
+    /// A Stream abstraction over the TDS protocol.
+    /// The stream can be used to read and write TDS messages on a 
+    /// SQL Server physical connection.
+    /// </summary>
+    internal class TdsStream : Stream
```
saurabh500, line 16: It would make sense to encapsulate the additional APIs, which are not part of stream, in interfaces.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
```diff
@@ -0,0 +1,181 @@
...
+    {
+        public override bool CanRead => throw new NotImplementedException();
```
saurabh500, line 18: Add an API for packet size manipulation.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
```diff
@@ -0,0 +1,181 @@
...
+
+        public override bool CanSeek => throw new NotImplementedException();
```
saurabh500, line 20: An API for cancellation needs to be added.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
benrr101: This is where I was getting hung up. Whatever interface we'd create (let's just say `ITdsStream` for this example), should inherit from Stream, but Stream doesn't have an interface. If we want the consumer of this class to be able to substitute in other `Stream`s, then the consumer won't be able to rely on the additional methods (unless it constantly does a check for the interface before trying to call the additional method, then falling back to the other if it doesn't work, which is not ideal at all). If the consumer takes a dependency on `TdsStream`, then they can safely rely on the additional methods, but won't be able to use mocks (unless we make the additional methods virtual).

So, my suggestion here is:
* Make the additional methods virtual, so that they can be mocked by consumers
* *If* there will be more than once class providing these methods, we could add an interface for them, just to ensure that the classes implement them the same way.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
benrr101, line 18: I think the `Can*` calls can simply chain into the underlying stream (and add an `<inheritdoc />`

benrr101: But I suspect these will be part of a later PR?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStreamPacketType.cs
```diff
@@ -0,0 +1,16 @@
+﻿namespace Microsoft.Data.SqlClientX.IO
+{
+    internal enum TdsStreamPacketType
+    {
+        PreLogin = 0x12,
+        Login7 = 0x10,
+        SQLBatch = 0x01,
```
benrr101, line 7: I think we should be adopting the C# naming standard for this project, following `PascalCase` even for acronyms.

benrr101 (changes requested):
Holding for naming conventions
The other two comments are more general feedback.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
saurabh500: Thanks, this makes sense. 

I was going to talk to you about this, and you have given me the answer

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
saurabh500: Yeah, I will add the `<inheritdoc />` 
And the implementation is going to evolve where this TdsStream is composed of TdsReadStream and TdsWriteStream. 
The TdsStream is what the rest of the driver will use and it will delegate calls to Read and WriteStreams internally. 

Many of the Can* methods will simply return true. e.g. CanRead, CanWrite, because in TDS we dont know whether we can read or write, till we actually Read / Write. The implementor of the protocol need to make sure that they are not reading / writing when they are not supposed to.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
```diff
@@ -0,0 +1,181 @@
...
+        {
+            throw new NotImplementedException();
+        }
+
+        public override void SetLength(long value)
+        {
+            throw new NotImplementedException();
+        }
+
+        public override void Write(byte[] buffer, int offset, int count)
```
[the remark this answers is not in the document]
saurabh500: Yes. There should be. Good call.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
```diff
@@ -0,0 +1,181 @@
...
+        /// of using the unoptimized path.
+        /// </summary>
+        /// <param name="buffer"></param>
+        /// <param name="offset"></param>
+        /// <param name="count"></param>
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
+        /// <exception cref="NotSupportedException"></exception>
+        public override Task WriteAsync(
+            byte[] buffer,
```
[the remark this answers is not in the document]
saurabh500: This will never be implemented. The ReadOnlyMemory overload with ValueTask is what this stream would expose, which would call into underlying's stream's ReadOnlyMemory -> ValueTask overload.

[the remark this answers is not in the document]
saurabh500: I have explicitly called out a NotSupportedException() in this overload, so that the default overrides of this method in the base class dont allow for unoptimized behavior.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
```diff
@@ -0,0 +1,181 @@
...
+        /// A convenience method to skip the bytes in the stream,
+        /// by allowing buffer manipulation, instead of making the consumer
+        /// allocate buffers to read and discard the packets.
+        /// </summary>
+        /// <param name="skipCount"></param>
+        /// <param name="isAsync"></param>
+        /// <param name="ct"></param>
+        /// <returns></returns>
+        /// <exception cref="NotImplementedException"></exception>
+        public ValueTask SkipReadBytesAsync(int skipCount,
```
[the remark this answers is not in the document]
saurabh500: No, usecase specificity is not going to be put in the IO layer, the docs would say that this is preferred method of skipping bytes which prevents the caller from allocating buffers for reading and skipping. 
If there are other patterns in the driver which can benefit from Skipping rather than reading and skipping, then they can continue to use this method.
Calling out PLP usecase specifically, make the method unwelcome to use in other scenarios.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStreamPacketType.cs
saurabh500: I agree. I didn't pay attention to the naming. Thanks for calling this out.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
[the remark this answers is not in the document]
benrr101: > allow for unoptimized behavior.

I don't think the byte[] overload  is "unoptimized" behavior, just not necessarily the most optimal behavior. For instance, if I have a constantly defined byte array and I need to write it, it is optimal to send the byte array directly. However, if we are forcing the ReadOnlyMemory overload, that means we will always need to new up a Memory struct before writing. And worse, if we are on net48, we will need to revert it back to a byte[] when we call any underlying stream. I think offering both options is better than forcing one or the other. Especially since we are effectively saying to the consumer "here is a Stream ... it might not actually implement all the things a Stream does, tho."

The approach I took for the MarsSessionStream and MarsSessionMuxDemux was to conditionally add the ReadOnlyMemory overload if we're on net8 or greater. I think this is best because it allows passing calls directly through the stream chain. The downside to it is that any consumer needs to also switch behavior based on dotnet version (at least where either call could be used, which I'm not sure will be very many situations).

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
benrr101: Another benefit of the second suggestion ... extension methods. Eg, we could offer the stream extensions that write numbers to the stream via multiple calls to WriteByteAsync, it could be an extension method of `ITdsStream`. That way we don't have to allocate buffers for 4 bytes of int, just to copy it into the stream. We could instead do 4 separate calls to write one byte at a time.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
```diff
@@ -0,0 +1,181 @@
...
+        {
+            throw new NotImplementedException();
+        }
+
+        public override void Write(byte[] buffer, int offset, int count)
+        {
+            throw new NotImplementedException();
+        }
+
+        public override ValueTask WriteAsync(
```
cheenamalhotra, line 90: Include docs or `<inheritdoc />` on public APIs

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
```diff
@@ -0,0 +1,181 @@
...
+using System.Threading.Tasks;
+
+namespace Microsoft.Data.SqlClientX.IO
+{
+    /// <summary>
+    /// A Stream abstraction over the TDS protocol.
+    /// The stream can be used to read and write TDS messages on a 
+    /// SQL Server physical connection.
+    /// </summary>
+    internal class TdsStream : Stream
```
cheenamalhotra, line 16: Is the stream disposable? Can it implement DisposeAsync pattern?

saurabh500: Good catch. It should

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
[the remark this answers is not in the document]
saurabh500: @benrr101  and @cheenamalhotra  you are right. 
I agree with the comment of byte[] overload not being unoptimized. I am hoping to stay away from Task based overloads.
Span and Memory are castable to byte[] so if the caller has a byte[], then it can call the stream to write the byte.

For Netfx @benrr101  and I discussed, since ValueTask is available in an extension Nuget package, for .Net framework, the stream class will wrap the underlying stream's Task into the ValueTask and expose them. This would make sure that we stay as performant as possible where ValueTask is available, but at the same time, with the same APIs we can server NetFx as well. 

I will be happy to revisit this point, if we see clients which absolutely need Task based overloads. That wont be too hard to expose.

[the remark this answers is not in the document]
cheenamalhotra: Sounds good, but yes, ideally we should ensure when the code comes in, we support all APIs possible to provide complete support for all targets.

cheenamalhotra (approved):

benrr101 (approved):

PR #2564, opened 2024-06-11
Create the base classes for the chain of responsibilities for Handlers

saurabh500:
Create the base classes for the chain of responsibilities for Handlers.


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
```diff
@@ -0,0 +1,181 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Data.SqlClientX.IO
```
benrr101: I don't think this is meant to go in this PR

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStreamPacketType.cs
```diff
@@ -0,0 +1,16 @@
+﻿namespace Microsoft.Data.SqlClientX.IO
```
benrr101: Likewise, this shouldn't be in this PR?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Handler.cs
```diff
@@ -0,0 +1,61 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Data.SqlClient;
+using Microsoft.Data.SqlClient.SNI;
+using Microsoft.Data.SqlClientX.IO;
+
+
+namespace Microsoft.Data.SqlClientX.Handlers
+{
+    internal interface IHandler<TRequest> where TRequest : HandlerRequest
```
benrr101, line 16: Eventually these should be separate files, but to unblock for now, we can have them in one file

benrr101: Likewise, there's some style changes that I'd push for

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Handler.cs
```diff
@@ -0,0 +1,84 @@
...
+    {
+        /// <summary>
+        /// Stores the internal handler to be executed next.
+        /// </summary>
+        private IHandler<TRequest> _nextHandler;
+
+        /// <summary>
+        /// Property for interacting with the Next Handler.
+        /// </summary>
+        public IHandler<TRequest> NextHandler { get => _nextHandler; set => _nextHandler = value; }
```
benrr101, line 50: Shouldn't this just be an auto-property?

benrr101 (approved):
These aren't blocking changes, but I really will push to get these changes made at *some* point

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Handler.cs
saurabh500: I can take care of all this now. Tell me what I am missing ?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Handler.cs
saurabh500: Actually I think this should be virtual. The base can provide an auto-property and the subclasses can do more complex stuff .

benrr101 (approved):
As per @saurabh500's request, here's the style fix requests :)

cheenamalhotra (approved):

PR #2569, opened 2024-06-12
Initial Commit of Transport Creation Handler

benrr101:
**Description**: Handler that creates the transport for a connection request chain of responsibility. It only works for TCP connections right now. IP addresses are looked up, if there is a preference for IPv4 or IPv6, the list is sorted and each one is tried until one connects or all fail.

Exceptions still need to be properly handled.

**Testing**: Everything builds :)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
```diff
@@ -0,0 +1,153 @@
...
+        }
+
+        private ValueTask<Stream> HandleSharedMemoryRequest()
+        {
+            throw new NotImplementedException();
+        }
+
+        private async Task<Stream> HandleTcpRequest(TcpTransportCreationParams request, CancellationToken ct)
+        {
+            ct.ThrowIfCancellationRequested();
```
edwardneal, line 60: I'd recommend putting this in the IP address loop too - the `Socket.Poll` call can take a while, and in the absence of a CancellationToken overload I think it's worth giving the caller the opportunity to exit the loop early.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
```diff
@@ -0,0 +1,153 @@
...
+                catch(Exception e)
+                {
+                    socketExceptions.Add(e);
+                }
+            }
+
+            // If no socket succeeded, throw
+            if (socket is null)
+            {
+                throw new AggregateException(socketExceptions);
```
edwardneal, line 106: The current logic throws the first SocketException it can find, then an AggregateException if it can't find any. This behaviour change might be intentional though.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
```diff
@@ -0,0 +1,153 @@
...
+
+            // Enable keep-alive
+            socket.SetSocketOption(SocketOptionLevel.Socket, SocketOptionName.KeepAlive, true);
+            socket.SetSocketOption(SocketOptionLevel.Tcp, SocketOptionName.TcpKeepAliveInterval, 1);
+            socket.SetSocketOption(SocketOptionLevel.Tcp, SocketOptionName.TcpKeepAliveTime, 30);
+
+            // Open the socket
+            try
+            {
+                await socket.ConnectAsync(ipEndpoint, ct).ConfigureAwait(false);
```
edwardneal, line 125: This overload isn't supported in .NET Framework - in that situation, you might need to spool up a new Task with the synchronous Connect and wait on it with the CancellationToken. It's not ideal.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
```diff
@@ -0,0 +1,153 @@
...
+            catch (SocketException e)
+            {
+                if (e.SocketErrorCode is not SocketError.WouldBlock)
+                {
+                    throw;
+                }
+            }
+
+            // Verify the socket is open
+            // @TODO: Configure timeout?
```
edwardneal, line 136: Yes - ConnectionTimeout in the connection string. The connection timeout covers everything from TCP connection to a usable TDS connection, which will add a little complexity in subsequent stages. I'd briefly thought about converting the connection timeout to a CancellationToken and threading it through every stage, but this makes that trickier.

edwardneal (commented):
A few performance comments, but I'm looking forward to seeing how this pans out!

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
```diff
@@ -0,0 +1,153 @@
...
+            // Create the stream for the socket
+            return new NetworkStream(socket);
+        }
+
+        private async Task<Socket> OpenSocketAsync(IPEndPoint ipEndpoint, CancellationToken ct)
+        {
+            var socket = new Socket(ipEndpoint.AddressFamily, SocketType.Stream, ProtocolType.Tcp) { Blocking = false, };
+
+            // Enable keep-alive
+            socket.SetSocketOption(SocketOptionLevel.Socket, SocketOptionName.KeepAlive, true);
```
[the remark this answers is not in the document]
saurabh500: `NoDelay` should definitely be set for the protocol.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Connection/ConnectionRequest.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        /// <summary>
+        /// Stream that is created during connection.
+        /// </summary>
+        public Stream ConnectionStream { get; set; }
+
+        /// <summary>
+        /// Parameters parsed from the connection string to be used by the transport creation handler.
+        /// </summary>
+        public TransportCreationParams TransportCreationParams { get; set; }
```
saurabh500, line 23: @samsharma2700 did you and @benrr101  do a handshake on this parameter? 
We already have the DataSource internal class, which can be leveraged to exchange the details about the TransportCreationParams

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationParams.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
```
saurabh500, line 2: @samsharma2700  and @benrr101  please sync about this.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
```diff
@@ -0,0 +1,153 @@
...
+            socket.SetSocketOption(SocketOptionLevel.Tcp, SocketOptionName.TcpKeepAliveTime, 30);
+
+            // Open the socket
+            try
+            {
+                await socket.ConnectAsync(ipEndpoint, ct).ConfigureAwait(false);
+            }
+            catch (SocketException e)
+            {
+                if (e.SocketErrorCode is not SocketError.WouldBlock)
```
saurabh500, line 129: What does this mean? In what cases is `WouldBlock` error code received and why is it safe to not throw it ?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
```diff
@@ -0,0 +1,153 @@
...
+            }
+
+            // Verify the socket is open
+            // @TODO: Configure timeout?
+            #if NET8_0_OR_GREATER
+            var pollTime = TimeSpan.FromSeconds(30);
+            #else
+            const int pollTime = 30000000; // 30s in us
+            #endif
+            if (!socket.Poll(pollTime, SelectMode.SelectWrite))
```
saurabh500, line 142: Is socket.Poll not blocking? I wonder if this will cause this async method to block, before returning?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/IpAddressVersionComparer.cs
```diff
@@ -0,0 +1,53 @@
...
+            if (x is null) { throw new ArgumentNullException(nameof(x)); }
+            if (y is null) { throw new ArgumentNullException(nameof(y)); }
+
+            if (x.AddressFamily == y.AddressFamily)
+            {
+                // Versions are the same, it's a tie.
+                return 0;
+            }
+
+            return _preferV6
```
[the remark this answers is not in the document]
benrr101: Took me about 10 re-reads to get what you meant, but I agree, that's a nice improvement.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
```diff
@@ -0,0 +1,153 @@
...
+            switch (request.TransportCreationParams)
+            {
+                case SharedMemoryTransportCreationParams:
+                    throw new NotImplementedException();
+
+                case NamedPipeTransportCreationParams:
+                    throw new NotImplementedException();
+
+                case TcpTransportCreationParams tcpParams:
+                    request.ConnectionStream = await HandleTcpRequest(tcpParams, ct).ConfigureAwait(false);
```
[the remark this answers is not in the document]
benrr101: I spent sometime reading up on this and digging through the code for the calls, but I'm not sure I see the sync-over-asyn anti-pattern going on. It looks like `Dns.GetHostAddressesAsync` isn't really async, but` Socket.ConnectAsync` is definitely async. I could change the call to get the addresses to use the sync method, but I think we're justified in using async for connecting. It sounds like you might know more about this than I do, so please let me know if I'm missing something.

This might also be a time for using one of those `ValueTask`s I've heard so much about 😅

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
```diff
@@ -0,0 +1,153 @@
...
+                    break;
+            }
+
+            // Attempt to connect to one of the matching IP addresses
+            // @TODO: Handle opening in parallel
+            Socket socket = null;
+            var socketExceptions = new List<Exception>();
+            foreach (var ipAddress in ipAddresses)
+            {
+                var ipEndpoint = new IPEndPoint(ipAddress, request.Port);
```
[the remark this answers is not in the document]
benrr101: Nice. I like it

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
```diff
@@ -0,0 +1,153 @@
...
+                try
+                {
+                    socket = await OpenSocketAsync(ipEndpoint, ct).ConfigureAwait(false);
```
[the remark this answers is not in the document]
benrr101: Wow, not sure how I missed that. Thanks!

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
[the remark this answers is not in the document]
benrr101: Gotcha, will do

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
benrr101: Sounds good to me 👍

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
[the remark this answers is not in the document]
edwardneal: I'm thinking of a case where we call the method with `isAsync` set to false. Because `isAsync` is never passed down to `HandleTcpRequest`, it's effectively always asynchronous. A caller who waits on Handle is wrapping a synchronous operation around async operations occurring on the thread pool (sync-over-async.) Your points around `Dns.GetHostAddressesAsync` and `Socket.ConnectAsync` are absolutely correct, but protect against thread-level blocking within a task (async-over-sync.)

There's a minor (but immediate) performance hit because we might now allocate Tasks. There's also a more subtle one which is a pain point for the current iteration of SqlClient, which is that if the thread pool is heavily loaded it could be some time until the thread pool has enough capacity to start the work. A caller wouldn't necessarily expect that if MultiSubnetFailover is disabled. In the worst case, both sync-over-async and async-over-sync can cause deadlocks.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
benrr101: Ah, I see why you mentioned OpenSocketAsync might be sync over async. I'm wondering if using the `TaskFactory.FromAsync(BeginConnect, EndConnect)` would be a better option than `Task.Run(() => Connect())`. Obviously this would only be used if we're in netfx.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
benrr101: I think this will need to come in a later PR 😬

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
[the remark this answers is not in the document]
saurabh500: I think the Sync path would need a separate flow where the sync versions of DNS and sockets are used, rather than relying on the async overloads. Likely in this PR it would make sense to `throw` or add a TODO to say Sync is to be supported. 

For MSF in sync mode, sync over async will suffer from the problems @edwardneal  mentioned. However a potential path forward for sync mode is to open all sockets in non-blocking mode, and later check using Socket.Select to see which one is writeable for the timeout duration, and return that particular socket. 
This way the I/O is used for socket open parallelization, and we dont rely on sync over async.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
benrr101: I'm really lost on how the current error handling is implemented... For this attempt I kinda just winged it. I can change this logic, but I need a bit of guidance. Are you saying it should throw
* First SocketException in the collected exceptions (if there are any)
* AggregateException from all collected exceptions (if there are no socket exceptions in the list)

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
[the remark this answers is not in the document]
edwardneal: Either a `throw` or a TODO is fine by me, so long as the problem's raised explicitly rather than implicitly - getting the correct behaviour from various scenarios of MSF/TNIR/connection timeouts would probably be better handled separately.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
edwardneal: I've looked through `Socket.ConnectAsync` and `TaskFactory`. Unfortunately the default CancellationToken passed in the TaskFactory constructor doesn't flow through to the `FromAsync` methods, so that won't work as we expect.

The only way to cancel the async connection attempt might simply be to register a handler against the CancellationToken which disposes of the socket, similar to the below:

```csharp
ct.Register((state) => ((Socket)state!).Dispose(), socket);
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
edwardneal: Yes - unless we want to change it. In cases where there are multiple IP addresses which all fail to connect, I can see an argument in favour of returning all of these SocketExceptions.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
```diff
@@ -0,0 +1,153 @@
...
+
+            // DNS lookup
+            var ipAddresses = await Dns.GetHostAddressesAsync(request.Hostname, ct).ConfigureAwait(false);
+            if (ipAddresses.Length == 0)
+            {
+                throw new Exception("Hostname did not resolve");
+            }
+
+            // If there is an IP version preference, apply it
+            switch (request.IpAddressPreference)
```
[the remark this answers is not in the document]
benrr101: Ok, I'll definitely keep this in mind. But, I think I'll address MSF in a later PR.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
benrr101: Ok, I'll tweak the error throwing code to do this

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
```diff
@@ -0,0 +1,185 @@
...
+            Socket socket = null;
+            var socketOpenExceptions = new List<Exception>();
+
+            var ipEndpoint = new IPEndPoint(IPAddress.None, context.DataSource.Port); // Allocate once
+            foreach (var ipAddress in ipAddresses)
+            {
+                ipEndpoint.Address = ipAddress;
+                try
+                {
+                    socket = await OpenSocketAsync(ipEndpoint, ct).ConfigureAwait(false);
```
cheenamalhotra, line 119: What about when isAsync is false?

src/Microsoft.Data.SqlClient.sln
```diff
@@ -195,7 +195,6 @@ EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{4F3CD363-B1E6-4D6D-9466-97D78A56BE45}"
 	ProjectSection(SolutionItems) = preProject
 		Directory.Build.props = Directory.Build.props
-		NuGet.config = NuGet.config
```
saurabh500, line 198: Why did this get removed?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIProxy.cs
```diff
@@ -197,7 +197,7 @@ private static bool IsErrorStatus(SecurityStatusPalErrorCode errorCode)
             }
 
             SNIHandle sniHandle = null;
-            switch (details._connectionProtocol)
+            switch (details.ResolvedProtocol)
```
saurabh500, line 200: This change has already been reviewed via the other PR. Can you merge that and then rebase/merge this branch with feat/sqlclientx ?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
```diff
@@ -0,0 +1,185 @@
...
+        {
+            ct.ThrowIfCancellationRequested();
+
+            // DNS lookup
+            var ipAddresses = isAsync
+                ? await Dns.GetHostAddressesAsync(context.DataSource.ServerName, ct).ConfigureAwait(false)
+                : Dns.GetHostAddresses(context.DataSource.ServerName);
+            if (ipAddresses.Length == 0)
+            {
+                throw new Exception("Hostname did not resolve");
```
saurabh500, line 88: You need to put this exception in the Request Objects Exception property

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
```diff
@@ -0,0 +1,185 @@
...
+                    // Not sorting necessary
+                    break;
+            }
+
+            // Attempt to connect to one of the matching IP addresses
+            // @TODO: Handle opening in parallel
+            Socket socket = null;
+            var socketOpenExceptions = new List<Exception>();
+
+            var ipEndpoint = new IPEndPoint(IPAddress.None, context.DataSource.Port); // Allocate once
```
saurabh500, line 113: not sure if you want to address in this PR or later, but it would be good to check for Resolved Port, and if that is -1 then use the Port

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
saurabh500: WE definitely need a `if/else` here on `isAsync` flag. For the else, you could choose to implement the socket connect for sync, in this PR or throw an exception in this PR, and then create another PR for sync socket open.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
cheenamalhotra: The strings should be localizable, consider fetching it from the Strings.resx

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
benrr101: Resolved offline, looks like this is only necessary during sync flow

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
benrr101: Looks like this is only necessary in sync flow

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
benrr101: Ok, addressed by having separate flows for opening socket synchronously and asynchronously.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIProxy.cs
benrr101: Yep, this PR has been updated to include those merged changes now

src/Microsoft.Data.SqlClient.sln
benrr101: I suppose this should actually go against main. The file hasn't existed for a while, and it was annoying me had a warning in the solution explorer.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
benrr101: Ok, I think handling cancellation and timeouts will be held for another PR. But I will definitely keep this in mind when addressing that.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
edwardneal: Nit: Using a more specific `Exception` derived class would be nice (although the idea having some way to distinguish a timeout from a failure to resolve the DNS name is great - the lack of this distinction has annoyed me for a while.)

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
```diff
@@ -0,0 +1,192 @@
...
+            ct.ThrowIfCancellationRequested();
+
+            var socket = new Socket(ipEndPoint.AddressFamily, SocketType.Stream, ProtocolType.Tcp) { Blocking = false };
+
+            // Enable keep-alive
+            socket.SetSocketOption(SocketOptionLevel.Socket, SocketOptionName.KeepAlive, true);
+            socket.SetSocketOption(SocketOptionLevel.Tcp, SocketOptionName.TcpKeepAliveInterval, KeepAliveIntervalSeconds);
+            socket.SetSocketOption(SocketOptionLevel.Tcp, SocketOptionName.TcpKeepAliveTime, KeepAliveTimeSeconds);
+
+            if (isAsync)
```
edwardneal, line 143: If anything here throws an exception, we want to catch it, dispose of the socket, then re-throw.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
```diff
@@ -0,0 +1,192 @@
...
+                catch(Exception e)
+                {
+                    socketOpenExceptions.Add(e);
+                }
+            }
+
+            // If no socket succeeded, throw
+            if (socket is null)
+            {
+                throw socketOpenExceptions.OfType<SocketException>().FirstOrDefault()
```
edwardneal, line 124: Nit: very small reservation about adding a dependency on LINQ at this stage.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
```diff
@@ -0,0 +1,192 @@
...
+            socket.SetSocketOption(SocketOptionLevel.Socket, SocketOptionName.KeepAlive, true);
+            socket.SetSocketOption(SocketOptionLevel.Tcp, SocketOptionName.TcpKeepAliveInterval, KeepAliveIntervalSeconds);
+            socket.SetSocketOption(SocketOptionLevel.Tcp, SocketOptionName.TcpKeepAliveTime, KeepAliveTimeSeconds);
+
+            if (isAsync)
+            {
+                #if NET6_0_OR_GREATER
+                await socket.ConnectAsync(ipEndPoint, ct).ConfigureAwait(false);
+                #else
+                await new TaskFactory(ct).FromAsync(socket.BeginConnect, socket.EndConnect, ipEndpoint, null)
```
edwardneal, line 148: Same points as previous review - but we definitely don't need to construct a new TaskFactory here, we can use `Task.Factory`. The `CancellationToken` you pass in the constructor doesn't get passed to anything used by `FromAsync`.

The only safe way to fully cancel the BeginConnect/EndConnect pair is to dispose of the underlying socket.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
```diff
@@ -0,0 +1,192 @@
...
+                // ignore it. This allows us to immediately return from connect and poll it,
+                // allowing us to observe timeouts and cancellation.
+                if (e.SocketErrorCode is not SocketError.WouldBlock)
+                {
+                    throw;
+                }
+            }
+
+            // Poll the socket until it is open
+            if (!socket.Poll(DefaultPollTimeout, SelectMode.SelectWrite))
```
edwardneal, line 186: Perhaps worth reducing the default poll timeout to something aggressive like 1/2/5/10/20/50ms and looping; this'll provide you with more opportunity to call `ct.ThrowIfCancellationRequested`. This poll timeout will be a tradeoff between cancellation precision and CPU usage, I've got no preference or data to suggest what the right figure for this is.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
```diff
@@ -0,0 +1,192 @@
...
+            {
+                // @TODO: Build CoR for handling the different protocols in order
+                if (context.DataSource.ResolvedProtocol is DataSource.Protocol.TCP)
+                {
+                    context.ConnectionStream = await HandleTcpRequest(context, isAsync, ct).ConfigureAwait(false);
+                }
+            }
+            catch (Exception e)
+            {
+                context.Error = e;
```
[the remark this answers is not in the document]
benrr101: Ah, yep, that was actually a bug. Thank you for calling it out.

Regarding the general question here, I raised a similar question and the feedback I got was that it's generally better to return the exception rather than catch it. I'm not 100% sure about this, but it's what we went for so far. You'll likely see this pattern in other PRs for the connection creation code.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
```diff
@@ -0,0 +1,192 @@
...
+
+        private async ValueTask<Stream> HandleTcpRequest(ConnectionHandlerContext context, bool isAsync, CancellationToken ct)
+        {
+            ct.ThrowIfCancellationRequested();
+
+            // DNS lookup
+            var ipAddresses = isAsync
+                ? await Dns.GetHostAddressesAsync(context.DataSource.ServerName, ct).ConfigureAwait(false)
+                : Dns.GetHostAddresses(context.DataSource.ServerName);
+            if (ipAddresses.Length == 0)
```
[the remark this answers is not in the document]
benrr101: I think this is something that we will revisit in a future PR. It's definitely something that's churning in the back of my head how we'll handle cancelling if the connection timeout expires. Especially how we may or may not be able to halt synchronous operations.

Out of curiosity, what does NB mean in this context? I'm more familiar with it referring to "no bump"

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
benrr101: Good call, thanks!

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
benrr101: Fair enough, though I think the cancellation code will be revised in a later PR.
Also, fwiw, this might not even be necessary since we are also discussing the whole net462 situation and how it impacts SqlClientX... Eg, right now this code only exists in the netcore project.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
benrr101: I've thought about this as well. I'll add a todo about it to address when we really tackle cancellation.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
benrr101: Sure, we can do that now.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
benrr101: I asked ai for some help with this one and it suggested this solution that resolves both localization issue and more specific exception issue :)

@saurabh500, I have this caught now in the next higher layer, so there's only one instance where the error object needs to be set before returning. Is that suitable?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
benrr101: I feel like it's safe since it's not a fast path. But then again, this will be used in parallel for multi-subnet failover where failures will be expected. Maybe this is a fast path in that case? I just like the brevity of the code, but if the performance really is garbage, I can be convinced to rewrite as a loop.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
```diff
@@ -0,0 +1,192 @@
...
+using System.Net.Sockets;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Data.SqlClient;
+using Microsoft.Data.SqlClient.SNI;
+using Microsoft.Data.SqlClientX.Handlers.Connection;
+
+namespace Microsoft.Data.SqlClientX.Handlers.TransportCreation
+{
+    internal class TransportCreationHandler : IHandler<ConnectionHandlerContext>
```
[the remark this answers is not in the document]
edwardneal: Sounds good! I'll keep watching and check again once the entire connection flow's done. The TDS streams establish a common view of the Once the TCP/NP/etc. connection's established, the TDS stream provides a common abstraction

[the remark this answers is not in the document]
edwardneal: Sounds good! I'll keep watching and check again once the entire connection flow's done. I suggested deriving `xTransportCreationHandler` classes because although the TDS streams establish a common view of the TCP/NP/etc. connection once it's established, at this stage I think the transport creation handlers will probably take quite different paths.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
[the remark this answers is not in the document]
edwardneal: One point I'd been thinking about for the main project (and which might be useful for the future PR) was having a `TimedCancellationToken` readonly struct which combined `CancellationToken` and the time taken before it times out. Having `CancellationToken` and `RemainingTime` properties would be useful for combining the timeout handling. It's a little awkward to retrofit, but probably easier to start working with from the start.

NB is just a side note - *[nota bene](https://en.wikipedia.org/wiki/Nota_bene)*.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
edwardneal: I'm not thinking of performance - as far as I know, that's broadly fine, and I agree that it's not on a hot path, and that variations connection/network latency will probably outweigh the LINQ CPU overhead. There's potentially a point to be made around allocations, but I don't expect those to be particularly large.

My motivation's purely around avoiding unnecessary dependencies on packages. If everything else in the codebase compiles without a reference to System.Linq then I think it'd be better to use a loop (or a variable to track the first SocketException as it's raised) - the change is simple enough that it's not worth dragging in another package for. If SqlClient's already using LINQ, I've got no objections.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
```diff
@@ -0,0 +1,207 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
```
JRahnama, line 8: I am not sure how much this is going to be relevant in the new design, but we have removed linq references in the driver in PR #1949 due to some issues with trimming.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
saurabh500: I +1 this comment for LINQ removal due to @JRahnama 's comment. 

My personal Point of view: 
LINQ has great usability, but sometimes the usability comes with a perf overhead. Knowing which is a hot path and which is not, is something that the domain experts only understand. (in this case pretty much all the reviewers of the code). I am hoping to stay away from "That part of driver uses LINQ, so can this" in future. Typically when code becomes older, the ways of doing things in a code become a standard, rather than deeper reasoning. Because of this I personally would recommend against usage of LINQ.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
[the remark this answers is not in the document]
benrr101: I was definitely thinking of something similar - if we can combine both the time out and user cancelled exception together it would very much simplify the code. We probably want to be able to throw different exceptions for which token was cancelled, so some more research might be necessary, but I bet someone has done it before

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
benrr101: Affirm, will do.

(but I still love linq 🥹)

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
```diff
@@ -0,0 +1,215 @@
...
+                return;
+            }
+
+            try
+            {
+                // @TODO: Build CoR for handling the different protocols in order
+                if (context.DataSource.ResolvedProtocol is DataSource.Protocol.TCP)
+                {
+                    context.ConnectionStream = await HandleTcpRequest(context, isAsync, ct).ConfigureAwait(false);
+                }
```
saurabh500, line 47: I am assuming the TODO is to handle the "Else" of this if ?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
```diff
@@ -0,0 +1,215 @@
...
+            // Attempt to connect to one of the matching IP addresses
+            // @TODO: Handle opening in parallel
+            Socket socket = null;
+            var socketOpenExceptions = new List<Exception>();
+
+            var portToUse = context.DataSource.ResolvedPort < 0
+                ? context.DataSource.Port
+                : context.DataSource.ResolvedPort;
+            var ipEndpoint = new IPEndPoint(IPAddress.None, portToUse); // Allocate once
+            foreach (var ipAddress in ipAddresses)
```
saurabh500, line 110: I get a suggestion in IDE to change to explicit type instead of var. 
Not sure if its a IDE specific suggestion or coming from .editorconfig. 

This is the rule being followed https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0007-ide0008

Can we use var when we are declaring an object, but when the object is derived from a function/iteration over enum etc, can we use explicit types (essentially what the IDE rule at the link above says)

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
```diff
@@ -0,0 +1,215 @@
...
+        {
+            throw new NotImplementedException();
+        }
+
+        private async ValueTask<Stream> HandleTcpRequest(ConnectionHandlerContext context, bool isAsync, CancellationToken ct)
+        {
+            ct.ThrowIfCancellationRequested();
+
+            // DNS lookup
+            var ipAddresses = isAsync
```
saurabh500, line 76: Can `ipAddresses` be null if the host doesn't exist ?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
```diff
@@ -0,0 +1,215 @@
...
+                throw new AggregateException(socketOpenExceptions);
+            }
+
+            // Create the stream for the socket
+            return new NetworkStream(socket);
+        }
+
+        private async ValueTask<Socket> OpenSocket(IPEndPoint ipEndPoint, bool isAsync, CancellationToken ct)
+        {
+            ct.ThrowIfCancellationRequested();
```
saurabh500, line 146: For sync cases, the callers will provide a `default` Cancellation token. the ct.ThrowIfCancellationRequested() may not be needed in sync paths.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
```diff
@@ -0,0 +1,215 @@
...
+            }
+
+            // Connection is established
+            socket.Blocking = true;
+            socket.NoDelay = true;
+
+            return socket;
+        }
+
+        private void OpenSocketSync(Socket socket, IPEndPoint ipEndPoint, CancellationToken ct)
```
saurabh500, line 185: There are certain improvements that can be made for sync paths. Let me know what you think. 

The caller of this code is iterating over the IP addresses and calling `OpenSocketSync`

`OpenSocketSync` will block till it can connect or fail. This means that serially connecting to the sockets would take up a bunch of time before the good IP address gets its turn.

For sync opens, you could probably provide a list of IP addresses and create a socket for each IP with `Blocking = false` 
After this, call the socket.Connect for each of these sockets. Hence `socket.Connect` will execute immediately with an error message. 
If the `e.SocketErrorCode` is WouldBlock, then don't handle the exception, because this means that the socket may connect later. 

At the end of the loop call `Socket.Select` on all the sockets to see which one is open, and use that particular socket. This allows you to quickly schedule connections to each socket of the IP addresses resolved and then wait for the first one that is connected.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
```diff
@@ -0,0 +1,215 @@
...
+            socket.SetSocketOption(SocketOptionLevel.Socket, SocketOptionName.KeepAlive, true);
+            socket.SetSocketOption(SocketOptionLevel.Tcp, SocketOptionName.TcpKeepAliveInterval, KeepAliveIntervalSeconds);
+            socket.SetSocketOption(SocketOptionLevel.Tcp, SocketOptionName.TcpKeepAliveTime, KeepAliveTimeSeconds);
+
+            try
+            {
+                if (isAsync)
+                {
+                    #if NET6_0_OR_GREATER
+                    await socket.ConnectAsync(ipEndPoint, ct).ConfigureAwait(false);
```
saurabh500, line 160: How do we deal with the timeout for async ? Sync path seems to be dealing with Timeout with Socket.Poll

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
saurabh500: You may not make this change now, but this is definitely very important to MultisubnetFailover (which is not the scope of this PR) where the docs say that an endpoint Could resolve to 64 IP addresses.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
benrr101: Yes, I'll throw a NotImplementedException in the else for now.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
benrr101: As far as I can tell, that doc doesn't actually assert any preference for one versus the other, just lists what the style rules are and what the true/false values mean.

I've gone back and forth on which to use over the years. Originally I was very against var, then I got retrained to use var and started to prefer it, and then C# 9 or whatever came out and now it's basically just a thing to argue over. I can adopt the suggestion, tho.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
benrr101: I would like to utilize the cancellation token as the way to signal that either the timeout has expired or the user has requested cancellation. This will be adopted in a later PR. As such, I think it would still be valuable to keep the cancellation token in the sync code path. @edwardneal  had a good suggestion for how to cancel while the socket is opening by registering event on the cancellation token.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
benrr101: Yes, this is basically how MSF will work, however that is for a later PR.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
benrr101: This is going to be addressed in a later PR.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
benrr101: Not sure if it will be null, but I'll add the check for null in there.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
saurabh500: I am still not sure how this would look end to end. If this theory holds, then great, else we will modify.

saurabh500 (approved):

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
David-Engel: FWIW, it would be super nice to replace the native SNI implementation completely with a managed one in both netcore and netfx. There are numerous issues around getting the native DLLs to deploy during publish actions (different project types require different detections) and the locking of the native DLLs is an issue in IIS apps and online updates. Not to mention the added maintenance cost of the native code and package...

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
```diff
@@ -0,0 +1,215 @@
...
+        private const int DefaultPollTimeout = 30 * 100000; // 30 seconds as microseconds
+        #endif
+
+        /// <inheritdoc />
+        public IHandler<ConnectionHandlerContext> NextHandler { get; set; }
+
+        /// <inheritdoc />
+        public async ValueTask Handle(ConnectionHandlerContext context, bool isAsync, CancellationToken ct)
+        {
+            if (context.DataSource is null)
```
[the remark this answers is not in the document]
David-Engel: FWIW, we've inherited a bunch a debug asserts that now fire during tests but tests still pass. It's really hard to know if the assert is outdated or if there is an underlying bug in the driver somewhere and it's made it very difficult to run tests in debug mode.

benrr101:
CI build didn't kick off. It builds locally and unit tests work, so I think it's fine. Feel free to annoy me if I broke branch.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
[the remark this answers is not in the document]
edwardneal: This is a *very* rough draft, but my first thought is to have a method closely tied to the connection string which either returns `null` (for a Connection Timeout of zero) or a new instance of `CancellationTokenSource` which has had `CancelAfter(ConnectionTimeout)` called on it. The `SqlConnection.Connect` method could call this to get a CTS, then use `CancellationTokenSource.CreateLinkedTokenSource` to combine this CTS' `CancellationToken` with the user-provided one. The output of this could be threaded through all of the SqlClientX internals.

There's space for some optimisation there: if the ConnectionTimeout-based CancellationTokenSource is null, there's no need to allocate or combine CTSes; all of the CancellationTokenSources need to be disposed of. There's also the need you mentioned to figure out which CancellationToken caused the cancellation exception - I think we'd just check if the user-provided CT's `IsCancellationRequested` property is set. Finally, we might want to track how long the final operation took for telemetry purposes.

Of course, all of this logic applies just as neatly to command timeouts. Perhaps a readonly struct with a better name than `OperationCancellation` - `CancellationOrTimeoutNotifier` is too awkward!

Issue #2119, opened 2023-08-10
Implement SqlDataSource


roji:
.NET 7.0 introduced the new DbDataSource abstraction to System.Data ([issue](https://github.com/dotnet/runtime/issues/64812)), [here's a video](https://www.youtube.com/watch?v=vRUtHeUpU44&list=PLdo4fOcmZ0oX-DBuRG4u58ZTAJgBAeQ-t&index=71) discussing it in depth.

Now, the .NET 7.0 implementation provides a default implementation which can be obtained from the ADO.NET provider's DbProviderFactory without the driver needing to implement anything. So the following code already works today:

```c#
await using var dataSource = SqlClientFactory.Instance.CreateDataSource("<connection string>");
await using var connection = await dataSource.OpenConnectionAsync();
await using var command = connection.CreateCommand();
command.CommandText = "SELECT 1";
Console.WriteLine(command.ExecuteScalar());
```

However, this provides a database-agnostic DbDataSource which isn't typed to any specific provider, so it hands out DbConnections rather than SqlConnections, which may be cumbersome to work with if the user is only targeting SQL Server.

A  proper SqlDataSource implementation in SqlClient would solve this, and also provide a good configuration/setup point (e.g. for authentication/access tokens, rather than the current static methods).

JRahnama:
Thanks @roji. We will discuss this internally and will get back to you.

ErikEJ:
@roji Are you thinking of something like this?

```
builder.Services.AddSqlDataSource("Server=.\\SQLEXPRESS;Database=Northwind;Integrated Security=true;Trust Server Certificate=true", setupAction =>
   {
       setupAction.UseAccessToken(accessToken);
   });

```

roji:
Yep, something like that. Npgsql has an NpgsqlDataSourceBuilder type for configuring an NpgsqlDataSource (when you're done you call Build() on it and get your data source); that would be the type of the `setupAction` parameter in your lambda above. There's also an [Npgsql.DependencyInjection](https://www.nuget.org/packages/Npgsql.DependencyInjection#readme-body-tab) package which does precisely this kind of nice integration with DI that you're showing above.

[...]

PR #2570, opened 2024-06-12
SqlConnectionX Instantiation via SqlConnection

mdaigle:


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnection.cs
```diff
@@ -1395,6 +1412,12 @@ private bool TryOpenWithRetry(TaskCompletionSource<DbConnectionInternal> retry,
         /// <include file='../../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnection.xml' path='docs/members[@name="SqlConnection"]/OpenWithOverrides/*' />
         public void Open(SqlConnectionOverrides overrides)
         {
+            if (_isExperimental)
+            {
+                _sqlConnectionX.Open();
```
cheenamalhotra, line 1417: We would need to support `SqlConnectionOverrides` for EF Core - fail fast scenarios, so I would recommend building that in Open API from the beginning.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnectionHelper.cs
```diff
@@ -21,6 +22,8 @@ public sealed partial class SqlConnection : DbConnection
         private DbConnectionOptions _userConnectionOptions;
         private DbConnectionPoolGroup _poolGroup;
         private DbConnectionInternal _innerConnection;
+        private bool _isExperimental;
```
cheenamalhotra, line 25: How is this set to true? Please also add code comment to describe why this property exists.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlConnectionX.cs
```diff
@@ -89,6 +98,7 @@ public override void Close()
         public override Task CloseAsync()
             => throw new NotImplementedException();
 
+
```
cheenamalhotra, line 101: Nit: extra line

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnection.cs
```diff
@@ -153,15 +153,27 @@ public SqlRetryLogicBaseProvider RetryLogicProvider
         [ResDescription(StringsHelper.ResourceNames.TCE_SqlConnection_TrustedColumnMasterKeyPaths)]
         public static IDictionary<string, IList<string>> ColumnEncryptionTrustedMasterKeyPaths => _ColumnEncryptionTrustedMasterKeyPaths;
 
+        internal SqlConnection(string connectionString, bool useExperimental) : this()
+        {
```
MaceWindu, line 157:
```suggestion
        {
            _isExperimental = useExperimental;
```
?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnection.cs
```diff
@@ -153,15 +153,27 @@ public SqlRetryLogicBaseProvider RetryLogicProvider
...
-        public SqlConnection(string connectionString) : this()
+        public SqlConnection(string connectionString) : this(connectionString, false)
         {
-            ConnectionString = connectionString;    // setting connection string first so that ConnectionOption is available
         }
 
-        /// <include file='../../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnection.xml' path='docs/members[@name="SqlConnection"]/ctorConnectionStringCredential/*' />
-        public SqlConnection(string connectionString, SqlCredential credential) : this()
+        internal SqlConnection(string connectionString, SqlCredential credential, bool useExperimental) : this()
         {
```
MaceWindu, line 171:
```suggestion
        {
            _isExperimental = useExperimental;
```
?

David-Engel:
> What is SqlConnectionX and why does it exist?

@Wraith2 Part of the SqlClientX project. The goal is to address core issues in SqlClient under an experimental code path in the existing package.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnection.cs
mdaigle: Thanks! I had reworked these a bit and accidentally dropped this.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnection.cs
```diff
@@ -214,6 +228,11 @@ public SqlConnection(string connectionString, SqlCredential credential) : this()
             //      checking pool groups which is not necessary. All necessary operation is already done by calling "ConnectionString = connectionString"
         }
 
+        /// <include file='../../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnection.xml' path='docs/members[@name="SqlConnection"]/ctorConnectionStringCredential/*' />
+        public SqlConnection(string connectionString, SqlCredential credential) : this(connectionString, credential, false)
```
cheenamalhotra, line 232: Instead of setting useExperimental to false, have you thought about setting it via AppContext switch that can be configured by user?

mdaigle: I'm open to it. In practice, is AppContext only used like this article describes (as an opt-out mechanism for new behavior) or is it used as a general settings store? https://learn.microsoft.com/en-us/dotnet/fundamentals/runtime-libraries/system-appcontext

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnection.cs
```diff
@@ -1665,10 +1678,16 @@ private Task InternalOpenWithRetryAsync(CancellationToken cancellationToken)
...
         /// <include file='../../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnection.xml' path='docs/members[@name="SqlConnection"]/OpenAsync/*' />
-        public override Task OpenAsync(CancellationToken cancellationToken)
-            => IsProviderRetriable ?
+        public override Task OpenAsync(CancellationToken cancellationToken) {
+            if (_isExperimental)
+            {
+                return _sqlConnectionX.OpenAsync(cancellationToken);
+            }
+
+            return IsProviderRetriable?
```
VladimirReshetnikov, line 1700: nit: space before `?`

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnectionHelper.cs
```diff
@@ -33,6 +36,19 @@ public SqlConnection() : base()
             _innerConnection = DbConnectionClosedNeverOpened.SingletonInstance;
         }
 
+        internal SqlConnection(bool useExperimental): this()
```
VladimirReshetnikov: nit: space before `:`

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnectionHelper.cs
```diff
@@ -33,6 +36,19 @@ public SqlConnection() : base()
...
+        {
+            _isExperimental = useExperimental;
+            if (_isExperimental)
+            {
+                _sqlConnectionX = new SqlConnectionX();
+            }
+            else
+            {
+                _innerConnection = DbConnectionClosedNeverOpened.SingletonInstance;
```
VladimirReshetnikov: Hasn't this assignment already happened on line 36?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlConnectionX.cs
```diff
@@ -31,6 +32,14 @@ internal SqlConnectionX() : base()
         {
         }
 
+        internal SqlConnectionX(string connectionString) : this()
+        {
+        }
+
+        internal SqlConnectionX(string connectionString, SqlCredential credential): this()
```
VladimirReshetnikov, line 39: nit: space before `:`

VladimirReshetnikov (changes requested):

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnection.cs
roji: Another possible direction here is to use the DbDataSource abstraction as the factory for the new experimental connection objects (#2119 generally tracked introducing a SqlDataSourcei n SqlClient). In other words, you would have a SqlDataSourceBuilder which can produce a SqlDataSource with a particular configuration; one of these configurable options would be to produce "experimental" connections. The disadvantage of this method is that users must use the SqlDataSource mechanism to create such experimental connections (but people would have to modify their code in any case). @saurabh500 and I discussed these possibilities a bit.

Whatever's done, I'd recommend putting the [`[Experimental]`](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.experimentalattribute?view=net-8.0) attribute on any programmatic API here; this attribute was introduced in .NET 8.0, and causes a warning with a specific diagnostic ID to be reported when a user uses an API (the user can then decide to suppress that specific diagnostic ID). This would make it easier to remove later as an experimental, non-supported API etc.

mdaigle: @roji, thanks! We do plan to follow the lead of npgsql and use the data source builder pattern. This is a start at integrating with the existing call pattern to reduce the burden of transitioning to the new implementation. The idea being we could seamlessly default you to the new driver if you're on the right .NET version and platform. But I'm thinking I'll shelve this PR for a bit until the other pieces are in place. I'm learning more about the builder and data source pattern as we implement those pieces and I think this PR will change slightly.

roji: Sure thing! Please feel free to ping me or reach out on Teams with any questions.

I'd be maybe wary of varying the default implementation based on the .NET version targeted - it's quite unexpected for some to switch the targeted TFM and for SqlClient to suddenly be a completely different implementation. But anyway, I'm sure all this will be discussed at some point.

PR #2571, opened 2024-06-13
Adding the write stream for the TdsStream

saurabh500:
Adding the write stream for the TdsStream

The next step would be to add TdsReadStream to compose the TdsStream.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
@@ -918,11 +918,14 @@
       <AutoGen>True</AutoGen>
       <DependentUpon>Strings.resx</DependentUpon>
     </Compile>
+    <Compile Include="Microsoft\Data\SqlClientX\Handlers\DataSourceHandler.cs" />
+    <Compile Include="Microsoft\Data\SqlClientX\IO\ITdsWriteStream.cs" />
     <Compile Include="Microsoft\Data\SqlClientX\IO\TdsStreamPacketType.cs" />
     <Compile Include="Microsoft\Data\SqlClientX\IO\TdsStream.cs" />
     <Compile Include="Microsoft\Data\SqlClientX\Handlers\HandlerRequest.cs" />
     <Compile Include="Microsoft\Data\SqlClientX\Handlers\HandlerRequestType.cs" />
     <Compile Include="Microsoft\Data\SqlClientX\Handlers\IHandler.cs" />
+    <Compile Include="Microsoft\Data\SqlClientX\IO\TdsWriteStream.cs" />
     <EmbeddedResource Include="$(CommonSourceRoot)Resources\Strings.resx">
       <Link>Resources\Strings.resx</Link>
       <LogicalName>Microsoft.Data.SqlClient.Resources.Strings.resources</LogicalName>
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
@@ -918,7 +918,7 @@
       <AutoGen>True</AutoGen>
       <DependentUpon>Strings.resx</DependentUpon>
     </Compile>
-    <Compile Include="Microsoft\Data\SqlClientX\Handlers\DataSourceHandler.cs" />
+    <Compile Include="Microsoft\Data\SqlClientX\IO\ITdsStream.cs" />
     <Compile Include="Microsoft\Data\SqlClientX\IO\ITdsWriteStream.cs" />
     <Compile Include="Microsoft\Data\SqlClientX\IO\TdsStreamPacketType.cs" />
     <Compile Include="Microsoft\Data\SqlClientX\IO\TdsStream.cs" />
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
@@ -919,6 +919,8 @@
       <AutoGen>True</AutoGen>
       <DependentUpon>Strings.resx</DependentUpon>
     </Compile>
+    <Compile Include="Microsoft\Data\SqlClientX\IO\TdsStreamPacketType.cs" />
+    <Compile Include="Microsoft\Data\SqlClientX\IO\TdsStream.cs" />
     <EmbeddedResource Include="$(CommonSourceRoot)Resources\Strings.resx">
       <Link>Resources\Strings.resx</Link>
       <LogicalName>Microsoft.Data.SqlClient.Resources.Strings.resources</LogicalName>
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/ITdsStream.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/ITdsStream.cs
@@ -6,10 +6,23 @@ using System.IO;
 
 namespace Microsoft.Data.SqlClientX.IO
 {
+    /// <summary>
+    /// Represents a TDS packet stream.
+    /// </summary>
     internal interface ITdsStream
     {
+        /// <summary>
+        /// Manipulates the buffer size of the TDS stream by reallocating the buffer to the new size.
+        /// Note: should be called once during a connection lifetime after login exchange.
+        /// </summary>
+        /// <param name="bufferSize">The value of the new buffer size.</param>
         public void SetPacketSize(int bufferSize);
 
+        /// <summary>
+        /// Replaces the underlying stream for a TDS stream. 
+        /// This is useful while changing between SSL stream and non-SSL stream.
+        /// </summary>
+        /// <param name="stream"></param>
         public void ReplaceUnderlyingStream(Stream stream);
     }
 }
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/ITdsStream.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/ITdsStream.cs
@@ -1,0 +1,15 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.IO;
+
+namespace Microsoft.Data.SqlClientX.IO
+{
+    internal interface ITdsStream
+    {
+        public void SetPacketSize(int bufferSize);
+
+        public void ReplaceUnderlyingStream(Stream stream);
+    }
+}
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/ITdsWriteStream.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/ITdsWriteStream.cs
@@ -6,13 +6,15 @@ using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 
-namespace Microsoft.Data.SqlClient.Microsoft.Data.SqlClientX.IO
+namespace Microsoft.Data.SqlClientX.IO
 {
     /// <summary>
     /// An internal interface to represent the TDS specific write stream operations.
     /// </summary>
     internal interface ITdsWriteStream : ITdsStream
     {
+        TdsStreamPacketType PacketHeaderType { get; set; }
+
         public ValueTask WriteByteAsync(byte value, bool isAsync, CancellationToken ct);
 
         public void QueueCancellation();
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/ITdsWriteStream.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/ITdsWriteStream.cs
@@ -12,10 +12,23 @@ namespace Microsoft.Data.SqlClientX.IO
     /// </summary>
     internal interface ITdsWriteStream : ITdsStream
     {
-        TdsStreamPacketType PacketHeaderType { get; set; }
+        /// <summary>
+        /// Represents the type of the packet header that the stream is currently writing.
+        /// </summary>
+        TdsStreamPacketType? PacketHeaderType { get; set; }
 
+        /// <summary>
+        /// Writes a byte to the stream.
+        /// </summary>
+        /// <param name="value">The value to be written.</param>
+        /// <param name="isAsync">Whether the operation should be performed asynchronosly.</param>
+        /// <param name="ct">Cancellation token.</param>
+        /// <returns></returns>
         public ValueTask WriteByteAsync(byte value, bool isAsync, CancellationToken ct);
 
+        /// <summary>
+        /// Queues a cancellation request to the stream, which will be sent to the server.
+        /// </summary>
         public void QueueCancellation();
     }
 }
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/ITdsWriteStream.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/ITdsWriteStream.cs
@@ -1,0 +1,27 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Data.SqlClient.Microsoft.Data.SqlClientX.IO
+{
+    /// <summary>
+    /// An internal interface to represent the TDS specific write stream operations.
+    /// </summary>
+    internal interface ITdsWriteStream : ITdsStream
+    {
+        public ValueTask WriteByteAsync(byte value, bool isAsync, CancellationToken ct);
+
+        public void QueueCancellation();
+    }
+
+    internal interface ITdsStream
+    {
+        public void SetPacketSize(int bufferSize);
+
+        public void ReplaceUnderlyingStream(Stream stream);
+    }
+}
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/ITdsWriteStream.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/ITdsWriteStream.cs
@@ -1,8 +1,7 @@
-﻿using System;
-using System.Collections.Generic;
-using System.IO;
-using System.Linq;
-using System.Text;
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
 using System.Threading;
 using System.Threading.Tasks;
 
@@ -18,12 +17,5 @@ namespace Microsoft.Data.SqlClientX.IO
         public ValueTask WriteByteAsync(byte value, bool isAsync, CancellationToken ct);
 
         public void QueueCancellation();
-    }
-
-    internal interface ITdsStream
-    {
-        public void SetPacketSize(int bufferSize);
-
-        public void ReplaceUnderlyingStream(Stream stream);
     }
 }
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
@@ -1,4 +1,8 @@
-﻿using System;
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
 using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
@@ -18,7 +18,7 @@ namespace Microsoft.Data.SqlClientX.IO
     /// </summary>
     internal class TdsStream : Stream, ITdsWriteStream
     {
-        private TdsWriteStream _writeStream;
+        private readonly TdsWriteStream _writeStream;
 
         /// <inheritdoc />
         public override bool CanRead => throw new NotImplementedException();
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
@@ -12,18 +12,34 @@ namespace Microsoft.Data.SqlClientX.IO
     /// A Stream abstraction over the TDS protocol.
     /// The stream can be used to read and write TDS messages on a 
     /// SQL Server physical connection.
+    /// The stream is responsible for abstracting away the calls to IO and handling the packet 
+    /// header from the consumers of TDS protocol.
+    /// The stream responds to the request of the callers, but it doesn't guarantee the correctness of the underlying TDS protocol correctness.
+    /// e.g. If the protocol states that there are N bytes that should be read, and the stream is asked to return
+    /// N + 1 bytes, then the stream will timeout trying to get N+1 bytes, or it will return N+1 bytes, if the 
+    /// N+1 byte is available.
     /// </summary>
     internal class TdsStream : Stream
     {
+        /// <inheritdoc />
         public override bool CanRead => throw new NotImplementedException();
-
+        
+        /// <inheritdoc />
         public override bool CanSeek => throw new NotImplementedException();
 
+        /// <inheritdoc />
         public override bool CanWrite => throw new NotImplementedException();
 
+        /// <inheritdoc />
         public override long Length => throw new NotImplementedException();
 
+        /// <inheritdoc />
         public override long Position { get => throw new NotImplementedException(); set => throw new NotImplementedException(); }
+
+        /// <summary>
+        /// Indicates if the cancellation is sent to the server.
+        /// </summary>
+        public virtual bool IsCancellationSent { get; internal set; }
 
         public TdsStream(Stream underLyingStream) : base()
         {
@@ -37,7 +53,18 @@ namespace Microsoft.Data.SqlClientX.IO
         /// </summary>
         /// <param name="stream"></param>
         /// <exception cref="NotImplementedException"></exception>
-        public void ReplaceUnderlyingStream(Stream stream)
+        public virtual void ReplaceUnderlyingStream(Stream stream)
+        {
+            throw new NotImplementedException();
+        }
+
+        /// <summary>
+        /// Allows manipulation of the packet size. This should only be used after login ack,
+        /// when the client and server exchange the negotiated packet size.
+        /// </summary>
+        /// <param name="packetSize">The negotiated packet size</param>
+        /// <exception cref="NotImplementedException"></exception>
+        public virtual void SetPacketSize(int packetSize)
         {
             throw new NotImplementedException();
         }
@@ -48,22 +75,25 @@ namespace Microsoft.Data.SqlClientX.IO
         /// TODO: Consider accepting an enum of packet types
         /// instead of the byte.
         /// </summary>
-        /// <param name="packetType"></param>
-        public void SetWritePacketType(TdsStreamPacketType packetType)
+        /// <param name="packetType">The type of packet to be sent out</param>
+        public virtual void SetWritePacketType(TdsStreamPacketType packetType)
         {
             throw new NotImplementedException();
         }
 
+        /// <inheritdoc />
         public override void Flush()
         {
             throw new NotImplementedException();
         }
 
+        /// <inheritdoc />
         public override int Read(Span<byte> buffer)
         {
             throw new NotImplementedException();
         }
 
+        /// <inheritdoc />
         public override int Read(
             byte[] buffer, 
             int offset, 
@@ -72,21 +102,31 @@ namespace Microsoft.Data.SqlClientX.IO
             throw new NotImplementedException();
         }
 
+        /// <inheritdoc />
         public override long Seek(long offset, SeekOrigin origin)
         {
             throw new NotImplementedException();
         }
 
+        /// <inheritdoc />
         public override void SetLength(long value)
         {
             throw new NotImplementedException();
         }
 
+        /// <inheritdoc />
+        public override void Write(ReadOnlySpan<byte> buffer)
+        {
+            throw new NotImplementedException();
+        }
+
+        /// <inheritdoc />
         public override void Write(byte[] buffer, int offset, int count)
         {
             throw new NotImplementedException();
         }
 
+        /// <inheritdoc />
         public override ValueTask WriteAsync(
             ReadOnlyMemory<byte> buffer,
             CancellationToken cancellationToken)
@@ -94,17 +134,7 @@ namespace Microsoft.Data.SqlClientX.IO
             throw new NotImplementedException();
         }
 
-        /// <summary>
-        /// We only support ValueTask based overloads for async IO.
-        /// This overload makes sure that the method throws, instead 
-        /// of using the unoptimized path.
-        /// </summary>
-        /// <param name="buffer"></param>
-        /// <param name="offset"></param>
-        /// <param name="count"></param>
-        /// <param name="cancellationToken"></param>
-        /// <returns></returns>
-        /// <exception cref="NotSupportedException"></exception>
+        /// <inheritdoc />
         public override Task WriteAsync(
             byte[] buffer,
             int offset,
@@ -127,6 +157,7 @@ namespace Microsoft.Data.SqlClientX.IO
             throw new NotImplementedException();
         }
 
+        /// <inheritdoc />
         public override ValueTask<int> ReadAsync(
             Memory<byte> buffer,
             CancellationToken cancellationToken)
@@ -143,7 +174,7 @@ namespace Microsoft.Data.SqlClientX.IO
         /// <param name="ct"></param>
         /// <returns></returns>
         /// <exception cref="NotImplementedException"></exception>
-        public ValueTask<byte> PeekByteAsync(bool isAsync,
+        public virtual ValueTask<byte> PeekByteAsync(bool isAsync,
             CancellationToken ct)
         {
             throw new NotImplementedException();
@@ -152,14 +183,14 @@ namespace Microsoft.Data.SqlClientX.IO
         /// <summary>
         /// A convenience method to skip the bytes in the stream,
         /// by allowing buffer manipulation, instead of making the consumer
-        /// allocate buffers to read and discard the packets.
+        /// allocate buffers to read and discard the bytes.
         /// </summary>
-        /// <param name="skipCount"></param>
-        /// <param name="isAsync"></param>
-        /// <param name="ct"></param>
+        /// <param name="skipCount">Number of bytes to skip</param>
+        /// <param name="isAsync">If the method should be called Asynchronously.</param>
+        /// <param name="ct">The cancellation token.</param>
         /// <returns></returns>
         /// <exception cref="NotImplementedException"></exception>
-        public ValueTask SkipReadBytesAsync(int skipCount,
+        public virtual ValueTask SkipReadBytesAsync(int skipCount,
             bool isAsync, 
             CancellationToken ct)
         {
@@ -173,7 +204,28 @@ namespace Microsoft.Data.SqlClientX.IO
         /// want to make it available for the next set of operations.
         /// </summary>
         /// <exception cref="NotImplementedException"></exception>
-        public void Reset()
+        public virtual void Reset()
+        {
+            throw new NotImplementedException();
+        }
+
+        /// <summary>
+        /// Queues the TDS cancellation token for the stream.
+        /// </summary>
+        /// <exception cref="NotImplementedException"></exception>
+        public virtual void QueueCancellation()
+        {
+            throw new NotImplementedException();
+        }
+
+        /// <inheritdoc />
+        public override ValueTask DisposeAsync()
+        {
+            throw new NotImplementedException();
+        }
+
+        /// <inheritdoc />
+        protected override void Dispose(bool disposing)
         {
             throw new NotImplementedException();
         }
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
@@ -148,7 +148,7 @@ namespace Microsoft.Data.SqlClientX.IO
         /// <returns></returns>
         /// <exception cref="NotImplementedException"></exception>
         public override async Task FlushAsync(CancellationToken ct)
-            => await _writeStream.FlushAsync(ct);
+            => await _writeStream.FlushAsync(ct).ConfigureAwait(false);
 
         /// <inheritdoc />
         public override ValueTask<int> ReadAsync(
@@ -214,7 +214,7 @@ namespace Microsoft.Data.SqlClientX.IO
         /// <inheritdoc />
         public override async ValueTask DisposeAsync()
         {
-            await _writeStream.DisposeAsync();
+            await _writeStream.DisposeAsync().ConfigureAwait(false);
             // TODO: For Read Stream as well.
         }
 
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
@@ -30,6 +30,19 @@ namespace Microsoft.Data.SqlClientX.IO
         }
 
         /// <summary>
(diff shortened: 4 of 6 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
```diff
@@ -918,11 +918,14 @@
       <AutoGen>True</AutoGen>
       <DependentUpon>Strings.resx</DependentUpon>
     </Compile>
+    <Compile Include="Microsoft\Data\SqlClientX\Handlers\DataSourceHandler.cs" />
```
benrr101, line 921: * This file doesn't exist
* IO\ITdsStream should be added to the project

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/ITdsStream.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.IO;
+
+namespace Microsoft.Data.SqlClientX.IO
+{
+    internal interface ITdsStream
```
benrr101, line 9: Comments 😃

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/ITdsWriteStream.cs
```diff
@@ -0,0 +1,21 @@
...
+using System.Threading.Tasks;
+
+namespace Microsoft.Data.SqlClientX.IO
+{
+    /// <summary>
+    /// An internal interface to represent the TDS specific write stream operations.
+    /// </summary>
+    internal interface ITdsWriteStream : ITdsStream
+    {
+        TdsStreamPacketType PacketHeaderType { get; set; }
```
benrr101, line 15: Comments would be nice here 😄

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
```diff
@@ -0,0 +1,274 @@
...
+        internal byte PacketNumber { get; private set; } = 1; // Packets always start with 1.
+
+        public TdsWriteStream(Stream underLyingStream) : base()
+        {
+            _underlyingStream = underLyingStream;
+            _WriteBuffer = new byte[TdsEnums.DEFAULT_LOGIN_PACKET_SIZE];
+        }
+
+        /// <summary>
+        /// Should only be called after the login negotiation is done.
```
benrr101, line 39: This description is more apt to be a remark, while the summary should indicate what happens when calling it.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
```diff
@@ -0,0 +1,274 @@
...
+            await FlushAsync(ct, true).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Use hard flush when we definitely know that the end of packet is reached.
+        /// </summary>
+        /// <param name="ct">Cancellation token for async operation.</param>
+        /// <param name="isAsync">Bool to indicate async operation</param>
+        /// <param name="flushMode">Whether this is a hard flush or a softflush</param>
+        private async Task FlushAsync(CancellationToken ct, bool isAsync, FlushMode flushMode = FlushMode.SoftFlush)
```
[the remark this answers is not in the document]
benrr101: I think I understand this a bit better after reading the code more. The private FlushAsync is pulling double duty, being used for both writing the buffer when full and finishing the packet at the request of the consumer. I think in my head it make sense to have 
* `private WritePacketAsync(bool isEndOfMessage, bool isAsync, CancellationToken)` - basically this method
* `public Write*` - When buffer is full, calls WritePacketAsync(false ...)
* `public Flush*` - Calls `WritePacketAsync(true)`

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
```diff
@@ -0,0 +1,274 @@
...
+        /// An async mechanism to write a byte to the stream.
+        /// </summary>
+        /// <param name="value">The byte value to be written</param>
+        /// <param name="isAsync">Indicates if the operation is Async</param>
+        /// <param name="ct">Cancellation token to cancel the async operation.</param>
+        /// <returns></returns>
+        public virtual async ValueTask WriteByteAsync(byte value, bool isAsync, CancellationToken ct)
+        {
+            // If we are already at the end of the buffer, flush the buffer, with a softflush.
+            if (_WriteBuffer.Length - WriteBufferOffset == 0)
```
benrr101, line 165: One thought here: if we write the byte first then check if we're at the end of the buffer, we get two benefits:
* When buffer fills, it's flushed instead of waiting for next write (negligible impact)
* We can return the ValueTask from `FlushAsync` instead of awaiting it (negligible impact)

Just something to consider

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
```diff
@@ -0,0 +1,274 @@
...
+            _WriteBuffer[7] = 0;
+
+            if (isAsync)
+            {
+                await _underlyingStream.WriteAsync(_WriteBuffer, 0, WriteBufferOffset, ct).ConfigureAwait(false);
+                await _underlyingStream.FlushAsync(ct).ConfigureAwait(false);
+            }
+            else
+            {
+                _underlyingStream.Write(_WriteBuffer, 0, WriteBufferOffset);
```
benrr101, line 111: This and next method have cancellation token support

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
```diff
@@ -0,0 +1,274 @@
...
+        {
+            int len = buffer.Length;
+            // The buffer may not have enough space. Write what we can and then flush the buffer with a soft flush.
+            while (len > 0)
+            {
+                if (len > _WriteBuffer.Length - WriteBufferOffset)
+                {
+                    // Only a part of the length fits in the buffer.
+                    int bytesToWrite = _WriteBuffer.Length - WriteBufferOffset;
+                    buffer.Slice(0, bytesToWrite).CopyTo(_WriteBuffer.AsSpan(WriteBufferOffset));
```
benrr101, line 183: Resharper recommends replacing with range indexer `[..bytesToWrite]`

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
```diff
@@ -0,0 +1,274 @@
...
+                    // The whole length can be added to the buffer.
+                    buffer.CopyTo(_WriteBuffer.AsSpan(WriteBufferOffset));
+                    WriteBufferOffset += len;
+                    len = 0;
+                }
+            }
+        }
+
+        /// <inheritdoc/>
+        public async override ValueTask WriteAsync(ReadOnlyMemory<byte> buffer, CancellationToken cancellationToken = default)
```
benrr101, line 199: Resharper says inconsistent modifier order, should be `public override async`

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
```diff
@@ -0,0 +1,274 @@
...
+    /// This stream will take care of wrapping the bytes with the TDS header. 
+    /// </summary>
+    internal class TdsWriteStream : Stream, ITdsWriteStream
+    {
+        private Stream _underlyingStream;
+
+        private byte[] _WriteBuffer;
+
+        // Start at the end of the Tds header
+        internal int WriteBufferOffset { get; private set; } = TdsEnums.HEADER_LEN;
```
benrr101, line 25: I'd say this probably doesn't need to be internal and can just be a private field. While I could see the benefit of exposing it for testing, I think it can just as easily be verified by ensuring the packets written have correct byte contents.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
```diff
@@ -0,0 +1,274 @@
...
+
+        // This should be set before flushing the buffer.
+        public TdsStreamPacketType PacketHeaderType { get; set; }
+
+        internal byte PacketNumber { get; private set; } = 1; // Packets always start with 1.
```
benrr101, line 30: Likewise, I don't think this needs to be internal and can be a private field. Can be easily tested by sending a packet (check for 1), send another packet (check for 2). Maybe even send a bunch of packets and make sure it rolls over.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStreamPacketType.cs
```diff
@@ -1,7 +1,12 @@
-﻿namespace Microsoft.Data.SqlClientX.IO
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.Data.SqlClientX.IO
 {
     internal enum TdsStreamPacketType
     {
+        None = 0,
```
benrr101, line 9: Instead of None, what about making TdsStreamPacketType in the TdsWriteStream a nullable?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
```diff
@@ -41,8 +44,11 @@ internal class TdsStream : Stream
         /// </summary>
         public virtual bool IsCancellationSent { get; internal set; }
 
-        public TdsStream(Stream underLyingStream) : base()
+        public TdsStreamPacketType PacketHeaderType { get => _writeStream.PacketHeaderType; set => _writeStream.PacketHeaderType = value; }
```
benrr101, line 47: Nit: get and set on separate lines

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
```diff
@@ -117,32 +111,29 @@ public override void SetLength(long value)
...
 
         /// <inheritdoc />
-        public override ValueTask WriteAsync(
+        public override async ValueTask WriteAsync(
             ReadOnlyMemory<byte> buffer,
-            CancellationToken cancellationToken)
-        {
-            throw new NotImplementedException();
-        }
+            CancellationToken cancellationToken)
```
benrr101, line 126: Resharper is unhappy because this should have a default value (ie, add ` = default`)

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
edwardneal: To add here, if `TdsWriteStream` is used in an unsynchronised context then `PacketNumber++` has a race condition and could send two packets with the same ID. It'd probably be cheapest to set the packet number to `Interlocked.Increment(ref myReference) & 0xFF`, but this requires the ability to pass a ref; a variable is usually the simplest way to handle this.

NB: the packet ID in the MARS packet header will need similar treatment. I've not reviewed on possible header size variations for MARS because I assume that a `TdsWriteStream` will be wrapped by a future `MarsStream` or similar.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
edwardneal, line 183: How should a 1MB/5MB/10MB varbinary payload be handled, assuming a 64KB buffer size? We'd currently perform 16/80/160 `CopyTo` calls (and possibly another set of similar calls into TCP buffers at the next level down).

Is it possible to refactor `FlushAsync` such that one part of it can flush any `Span`/`Memory` into a packet and `Write`/`WriteAsync` can slice up the incoming payload and call that directly? This would skip the buffer copy unless we know there'd definitely be space for our payload in it.

If a varbinary payload won't ever be that large, this can be ignored - I'm not sure exactly what that type of large UPDATE statement would be transported on the network as.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
```diff
@@ -0,0 +1,274 @@
...
+            }
+            else
+            {
+                status = TdsEnums.ST_BATCH;
+                // We have filled up the buffer. We need to send it out, hence next packet gets an increment.
+                PacketNumber++;
+            }
+
+            _WriteBuffer[1] = status;
+            _WriteBuffer[2] = (byte)(WriteBufferOffset >> 8);   // Length upper byte
```
edwardneal, line 96: Two micro-optimisations (since this is on the hot path.) Also, one nit.

Nit: elements 4, 5 don't need to be set in `FlushAsync` - they're constant. We can set them once on buffer initialisation.

Nit: BinaryPrimitives will provide a negligible performance boost when writing the packet length.

Nit: I'm not a fan of using `WriteBufferOffset` as the length. I'd personally prefer to see it described as CurrentPacketLength (or something clearer) throughout.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
```diff
@@ -0,0 +1,274 @@
...
+        {
+            _underlyingStream = underLyingStream;
+            _WriteBuffer = new byte[TdsEnums.DEFAULT_LOGIN_PACKET_SIZE];
+        }
+
+        /// <summary>
+        /// Should only be called after the login negotiation is done.
+        /// </summary>
+        /// <param name="bufferSize">new buffer size</param>
+        public void SetPacketSize(int bufferSize) => _WriteBuffer = new byte[bufferSize];
```
edwardneal, line 42: Food for thought here: is there value in returning the old buffer and renting from ArrayPool?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
```diff
@@ -0,0 +1,274 @@
...
+            int len = buffer.Length;
+            // The buffer may not have enough space. Write what we can and then flush the buffer with a soft flush.
+            while (len > 0)
+            {
+                if (len > _WriteBuffer.Length - WriteBufferOffset)
+                {
+                    // Only a part of the length fits in the buffer.
+                    int bytesToWrite = _WriteBuffer.Length - WriteBufferOffset;
+                    buffer.Slice(0, bytesToWrite).CopyTo(_WriteBuffer.AsSpan(WriteBufferOffset));
+                    WriteBufferOffset += bytesToWrite;
```
edwardneal, line 184: If this isn't called within a synchronised context, this'll need to change to avoid a race condition. Similar story in WriteByte.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
```diff
@@ -0,0 +1,274 @@
...
+
+        // This should be set before flushing the buffer.
+        public TdsStreamPacketType PacketHeaderType { get; set; }
+
+        internal byte PacketNumber { get; private set; } = 1; // Packets always start with 1.
+
+        public TdsWriteStream(Stream underLyingStream) : base()
+        {
+            _underlyingStream = underLyingStream;
+            _WriteBuffer = new byte[TdsEnums.DEFAULT_LOGIN_PACKET_SIZE];
```
edwardneal, line 35: The connection string can specify the packet size from the outset. It might be better to account for that here, to save any future need for reallocations.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
benrr101: +1 to `BinaryPrimitives` (will need to pull in a nuget package for netfx tho)

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
benrr101: I've raised a similar concern while working on MARS stream object. The consensus so far seems to be that because we explicitly state that SqlClient isn't thread-safe (I might be overstating this), we really don't need to complicate matters by synchronizing reads/writes. Though I personally think this is a good opportunity to make it properly thread-safe without mucking it up. On the other hand, if we make it unnecessarily thread-safe it will be a lot harder to prove it is still satisfactorily thread-safe (or not) once the thread safety mechanisms are removed.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
saurabh500: My bad. I remember having cleaned this up and I see I have a commit sitting in my local branch.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/ITdsStream.cs
saurabh500: We need stylecop

saurabh500: Regardless will address.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/ITdsWriteStream.cs
saurabh500: Will address.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
saurabh500: Will address.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
```diff
@@ -0,0 +1,274 @@
...
+using System.Threading.Tasks;
+using Microsoft.Data.SqlClient;
+
+namespace Microsoft.Data.SqlClientX.IO
+{
+    /// <summary>
+    /// A stream abstraction over the TDS protocol which works to write the TDS packet
+    /// This stream will take care of wrapping the bytes with the TDS header. 
+    /// </summary>
+    internal class TdsWriteStream : Stream, ITdsWriteStream
```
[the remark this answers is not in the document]
saurabh500: We wont need the Read/Write functionality without the other one. 
The idea was to encapsulate the Read and Write of TDS stream separate, because they are sort of orthogonal, and I want to err on the side of separating these concerns in two different classes.

I could potentially merge these two in TdsStream itself.

I am not taking a hard stance on this topic. But I will leave it out of scope of this PR. 

I want to get my TdsReadStream PR in as well, and then when both exist, I will be happy to revisit and refactor, if we think that partial classes or putting both in TdsStream makes sense.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
```diff
@@ -0,0 +1,274 @@
...
+    {
+        private Stream _underlyingStream;
+
+        private byte[] _WriteBuffer;
```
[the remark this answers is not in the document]
saurabh500: Ack

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
saurabh500: Oh right. I missed it. no point exposing this.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
saurabh500: @benrr101  agree on the visibility.

@edwardneal what is the meaning of unsynchronised context? 
From the rest of your comment, I think you are hinting at concurrent execution. This stream will not be called concurrently. That's due to the way TDS protocol is. MARS is another story, and @benrr101  and I will be visiting MARS later. First we want to light up TDS.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
edwardneal: That's fair to a point - but if that's going to remain the case, I'd question whether it'd be right to support MARS at all. One of the central goals of MARS is to read multiple result sets in parallel, but if the library reading those sets isn't thread-safe then it feels like a potential source of bugs (particularly for large result sets which straddle multiple SNI packets.)

I'd suggest that at least the physical connection and MARS headers should be thread-safe, purely to allow the logical connections to be demultiplexed properly.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
```diff
@@ -0,0 +1,274 @@
...
+
+        // Start at the end of the Tds header
+        internal int WriteBufferOffset { get; private set; } = TdsEnums.HEADER_LEN;
+
+        // This should be set before flushing the buffer.
+        public TdsStreamPacketType PacketHeaderType { get; set; }
+
+        internal byte PacketNumber { get; private set; } = 1; // Packets always start with 1.
+
+        public TdsWriteStream(Stream underLyingStream) : base()
```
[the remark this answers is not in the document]
saurabh500: The coding style guide for this project doesn't really specify the above. But it doesn't specify anything :) https://github.com/dotnet/SqlClient/blob/main/coding-style.md 

I can stick to your recommendation, and thank you for it. 

I wish it could be enforceable in VS.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
saurabh500: The Protocol says that the initial packet size is always 4096. The connection string parameter is first negotiated with the server in the Login packet and response and then all the buffers need to be adjusted to follow the new negotiated packet size. 
This implementation reflects the protocol spec for TDS packets.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
saurabh500: Good point !

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
edwardneal: @saurabh500 - yes, that's correct and I agree that it eliminates the possible race condition.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
saurabh500: Well, I will err on the side of simplicity here. This method will be called only once during the connection lifecycle, after the new packet size is negotiated with the server. 
So we expect one buffer resize. The default is 8000, and in most cases the customer's dont change the default. 
however there is value in checking if the buffer size has changed from the current size, and then resize the buffer. A micro-optimization, in case the connection string was changed to negotiate a 4096 bytes packet

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
```diff
@@ -0,0 +1,274 @@
...
+        /// <inheritdoc />
+        public override bool CanRead => false;
+
+        /// <inheritdoc />
+        public override bool CanSeek => false;
+
+        /// <summary>
+        /// Always returns true since we can always write to the stream.
+        /// </summary>
+        public override bool CanWrite => true;
```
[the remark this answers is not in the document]
saurabh500: Happy to revisit later. 
Likely a disposed stream shouldn't be writeable. 
but this stream for its lifetime is always writeable. 

I will add a TODO.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
```diff
@@ -0,0 +1,274 @@
...
+        {
+            Debug.Assert(PacketHeaderType != TdsStreamPacketType.None, "PacketHeaderType is not set. Cannot flush the buffer without setting the packet header type.");
+            _WriteBuffer[0] = (byte)PacketHeaderType;
+            byte status;
+
+            // TODO: Handle cancellation queueing. If there is a cancellation queued up, then send the status with IGNORE bit set.
+
+            if (flushMode == FlushMode.HardFlush)
+            {
+                status = TdsEnums.ST_EOM;
```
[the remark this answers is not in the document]
saurabh500: I agree in spirit. But at this point I don't have a point of view of what in TdsEnums' should go where. 

I have made this effort in the past (when I was doing performance improvements of large async reads on my fork), and I realized that I needed the enums in various contexts and ended up repeating them in different logical groups.

I will keep resolution of this comment out of scope of this PR. 
TdsEnums does give me a headache.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
edwardneal: Yes, apologies - just re-checked MS-TDS 2.2.6.4 and you're correct.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
saurabh500: @edwardneal  and @benrr101 Ack. Will address them. 

@benrr101  on a separate note, this is one of the reason I am hoping to keep Read and Write Streams separate. When the read stream comes in the packetLength is the length needed from the packet header, which plays an important role. 

I am really hoping to keep the logic of read and write paths in separate classes. however there may be a better way of dealing with my conundrum, which you guys can help with, when you see my Read Stream PR

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
```diff
@@ -0,0 +1,274 @@
...
+                PacketNumber++;
+            }
+
+            _WriteBuffer[1] = status;
+            _WriteBuffer[2] = (byte)(WriteBufferOffset >> 8);   // Length upper byte
+            _WriteBuffer[3] = (byte)(WriteBufferOffset & 0xff); // Length lower byte
+            _WriteBuffer[6] = PacketNumber;
+            _WriteBuffer[4] = 0;
+            _WriteBuffer[5] = 0;
+
```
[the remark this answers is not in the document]
saurabh500: I wish I can tell, but sometimes my code author brain doesn't work as a code reviewer :D 

I will fix it.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
saurabh500: For sync writes ? The underlying will be SSL / Network stream.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
saurabh500: Thanks @benrr101

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
saurabh500: @benrr101  for this, I want to force the callers to provide a value in this case, rather than using `=default` to make sure that the consumers are forced to think about passing down the Cancellation token.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
saurabh500: I don't follow. We always need to check if the buffer has space in it before writing, else Flush and write. Otherwise we will be writing past the allocated buffer index.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
saurabh500: @edwardneal  I agree that this optimization is possible. 
At this point I am going to add a TODO: I need to test a few things with SQL server before I can make this optimization.

For cases where the incoming `Span`/`Memory` is larger than the buffer, If we write the buffer directly to the underlying stream, then first I need to write the header to the underlying stream, and then the incoming `Span`/`Memory` 
This splits the outgoing TDS packet in two parts, header and payload. I want to check how SQL server behaves in those cases. 

Your scenario is valid. A Note: the buffer size never goes above 32768 bytes (protocol spec), and the buffer size is controlled by the connection string. Hence a large varbinary will be split across across multiple TDS packets. 

Your use case is valid with SqlCommand.ExecuteNonQuery with SqlParameters where the parameter values are larger than the buffer size. 

I will do some experiments before making these changes. 

Also based on our experiments, the async interaction with SQL server can benefit a lot compared to the current implementation, esp with large data read. I haven't benchmarked large data writes yet.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStreamPacketType.cs
saurabh500: Done

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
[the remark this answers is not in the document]
saurabh500: Fixed with a check on underlying stream.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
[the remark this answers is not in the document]
benrr101: I'm not entirely sure, but I've been able to get rider/resharper to yell at me if I put things in the wrong order (well, at least if I put private methods/properties before public ones). But idk if that's a roslyn analyzers thing or what generated it.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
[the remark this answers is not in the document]
benrr101: While yes, we can't write if the underlying stream is `null`, we also can't write if the underlying stream can't write. Is there anything wrong with changing the check to `_underlyingStream?.CanWrite`?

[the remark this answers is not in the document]
benrr101: Also, comment should be updated probably to just be `<inheritdoc>`

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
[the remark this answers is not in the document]
benrr101: Still not alphabetical within the groups, but I'll accept that the groups are properly organized now :)

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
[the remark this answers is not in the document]
saurabh500: If this needs to be done right, then we need this enforced in the build pipeline. Doesn't matter if the code author is using Rider/VS/VSCode or Notepad.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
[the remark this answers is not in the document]
benrr101: I can respect that we might accidentally duplicate the enums. I think one benefit of our plan here is that we shouldn't ever need to expose the packet flags outside the tds stream. If it were me, I'd probably put the packet status flags in the IO namespace and if it turns out we need it somewhere else, we move it up a level. I'd also not remove the definitions from TdsEnums yet, and only make the enum for the packet status flags (ie, limit the scope of the change).

But we can hold for another PR.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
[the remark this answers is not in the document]
saurabh500: Not planning to enhance `CanWrite` anymore. I dont expect this being called before writes.
 The consumers can get a true from CanWrite, but actual write may fail due to network I/O failure. There are no usecases where checking for CanWrite can really account for sending of network data reliably. .

The consumers anyhow need to account for network failures while writing and handle exceptions.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
benrr101: What I'm suggesting is flipping the order of check/write. Eg, right now:
* If space in buffer == 0
  * Flush
* Write byte

My suggestion is:
* Write byte
* If space in buffer == 0 (or pointer == buffer.length)
  *  Flush

Though this would also necessitate having the write method also flip its behavior. But, it's mostly negligible impact, afaict. I'd have to run a benchmark to see if there's any impact.

saurabh500: How do you write to a buffer which doesn't have space in it? 
If I write a byte if there is no space, then where does it go ?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
benrr101: Yep, you're right. I was just going through resharper hints and didn't really think about whether it was applicable here. I think it's suggestions is "you're in an async context, why would you even want to use a synchronous method?" ... which I still don't really understand the why behind this pattern...

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
```diff
@@ -0,0 +1,308 @@
...
+        }
+
+        #endregion
+
+        #region Public Methods
+
+        /// <inheritdoc />
+        public override async ValueTask DisposeAsync()
+        {
+            await _underlyingStream.DisposeAsync();
```
cheenamalhotra, line 83: ConfigureAwait(false);

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
```diff
@@ -0,0 +1,308 @@
...
+        /// Use this in case the caller is always going to write asynchronously. 
+        /// Otherwise use the WriteByteAsync because it takes the isAsync flag.
+        /// </summary>
+        /// <param name="value">The byte value to be written to the stream.</param>
+        public override void WriteByte(byte value)
+        {
+            // If we are already at the end of the buffer, flush the buffer, with a softflush.
+            if (_writeBuffer.Length - _writeBufferEnd == 0)
+            {
+                FlushPacketAsync(false, isAsync: false, CancellationToken.None).Wait();
```
cheenamalhotra, line 157: No Waits should be used.
You should either write truly Sync API to FlushPacket or do not expose WriteByte if it cannot be made truly sync.
Only use Async APIs where needed, and GetAwaiter().GetResult() from parent caller will make the async API sync.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
```diff
@@ -0,0 +1,308 @@
...
+            while (len > 0)
+            {
+                if (len > _writeBuffer.Length - _writeBufferEnd)
+                {
+                    // Only a part of the length fits in the buffer.
+                    int bytesToWrite = _writeBuffer.Length - _writeBufferEnd;
+                    buffer[..bytesToWrite].CopyTo(_writeBuffer.AsSpan(_writeBufferEnd));
+                    _writeBufferEnd += bytesToWrite;
+                    len -= bytesToWrite;
+                    FlushPacketAsync(false, isAsync: false, CancellationToken.None).ConfigureAwait(false);
```
cheenamalhotra, line 134: Not awaited.
Same comment as above for WriteByte.

cheenamalhotra (changes requested):

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
saurabh500: No need to await here no ? The FlusPacketAsync will complete synchronously due to the flag.

saurabh500: Also I cant await in a sync method.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriteStream.cs
```diff
@@ -0,0 +1,308 @@
...
+    internal class TdsWriteStream : Stream, ITdsWriteStream
+    {
+        #region Private Fields
+
+        private Stream _underlyingStream;
+
+        private byte[] _writeBuffer;
+
+        // Start at the end of the Tds header
+        private int _writeBufferEnd { get; set; } = TdsEnums.HEADER_LEN;
```
benrr101, line 27: I think this made more sense with the name `_writeBufferOffset` since it denotes where in the buffer we are currently writing to.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
benrr101: Resharper will also yell at you if you call a method with a cancellation token arg from a method that has a cancellation token arg 😄 

eg
```
public void MyAsync(CancellationToken ct) {
  something.TakesCancellationTokenAsync(); // <-- lights up "this method has an overload with cancellation support"
}
```

Not required change for PR - just sounds like we should use Resharper 😛

benrr101 (approved):
Mostly looks good. Please consider renaming _writeBufferEnd to _writeBufferOffset

cheenamalhotra (approved):

PR #2573, opened 2024-06-13
Fix | TvpTest Split Phase 4

arellegue:
Created unit test for QueryHintsTest.


paulmedynski:
Closing due to inactivity.  We may resurrect these as part of a larger test revitalization effort.

PR #2575, opened 2024-06-15
eng | Initial YAML CI pipeline

DavoudEshtehari:
Some of the features and enhancements along with migration to YAML:
- Set all multipliers via pipeline parameters to streamline configuration.
- Initiate the build process for the project and/or package types, depending on the proposed changes.
- Activate test steps in debug mode using the "Enable debug output" parameter for detailed troubleshooting.
- Retry functional and manual tests separately.
- Enhance the multiplier for Managed SNI on Windows and bypass target frameworks that are not relevant.
- Configure stage extraction based on metadata to ensure consistency:
  - Apply the same settings for both project and package build types.
  - Utilize similar configurations across multiple images from the same pool.
  - Accommodate self-hosted and private agent pools.
  - Filter code coverage results according to the target frameworks.
  - Run x86-specific tests aligned with the target frameworks.
  - Support diverse environments including on-premises, cloud-based, and enclave SQL Servers.
  - Create steps compatible with Windows, Linux, or macOS platforms.
- Inject additional steps into each test stage using the `prebuildSteps` parameter within the `ci-run-tests-stage` template for enhanced customization.

Breaking down the review based on the commits would make it easier to track the changes.

Sample runs:
- [Project build type](https://sqlclientdrivers.visualstudio.com/public/_build/results?buildId=90979&view=results)
- [Package build type](https://sqlclientdrivers.visualstudio.com/public/_build/results?buildId=90762&view=results)


JRahnama (approved):

dauinsight (approved):

David-Engel (approved):
My suggestions are required. But I think consistently adding `@self` would be cleaner.

PR #2579, opened 2024-06-17
[5.2.x] Update Azure.Identity to 1.11.4

### LINKED ISSUE #2568: Release a new version with an update ref to 'Azure.Identity'
'Azure.Identity' 1.11.3 has a known moderate severity vulnerability, https://github.com/advisories/GHSA-m5vv-6r4h-3vj9

can you release a version that updates this reference

ErikEJ:
fixes #2568

### DIFF
--- a/tools/props/Versions.props
+++ b/tools/props/Versions.props
@@ -28,7 +28,6 @@
   <!-- NetFx and NetCore project dependencies -->
   <PropertyGroup>
     <AzureIdentityVersion>1.11.4</AzureIdentityVersion>
-    <MicrosoftIdentityClientVersion>4.61.3</MicrosoftIdentityClientVersion>
     <MicrosoftIdentityModelProtocolsOpenIdConnectVersion>6.35.0</MicrosoftIdentityModelProtocolsOpenIdConnectVersion>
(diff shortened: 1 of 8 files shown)


tools/props/Versions.props
```diff
@@ -27,8 +27,7 @@
   </PropertyGroup>
   <!-- NetFx and NetCore project dependencies -->
   <PropertyGroup>
-    <AzureIdentityVersion>1.11.3</AzureIdentityVersion>
-    <MicrosoftIdentityClientVersion>4.60.3</MicrosoftIdentityClientVersion>
+    <AzureIdentityVersion>1.11.4</AzureIdentityVersion>
```
thompson-tomo, line 30: Thoughts about bumping this 1.12.0

ErikEJ: Actually no, not in a patch

thompson-tomo: @ErikEJ what I do notice is that this is a new dependency from this batch version going forward k

ErikEJ: ?? Sure, but not a new major version... (Or I do not understand what you mean)

thompson-tomo: As it currently stands there is a dependency on Microsoft.Identity however with this pr we are removing that dependency & introducing a new dependency on Azure.Identity hence feel that we should be using the latest.

PR #2582, opened 2024-06-17
Adding Tds Read stream to read data from TDS packets

saurabh500:
This PR brings in the TdsReadStream which is used to compose the TdsStream.

The TdsReadStream is mean to reliably return the number of bytes requested by the consumer of TdsStream as long as the consumer is asking for the right number of bytes based on the TDS protocol.

The PR doesn't aim to merge the Read and Write streams together. If we feel the need to merge the 2 streams, it will be out of scope of this PR.


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReadStream.cs
```diff
@@ -0,0 +1,350 @@
...
+        private int _packetHeaderDataLength { get; set; } = 0;
+
+        private byte _packetHeaderType { get; set; } = 0;
+
+        private byte _packetStatus { get; set; }
+
+        /// <summary>
+        /// Stored internally. It will be used for tracing.
+        /// </summary>
+        private int _spid;
```
cheenamalhotra, line 50: Have you given a thought about how would this be exposed via the public API [SqlConnection.ServerProcessId](https://learn.microsoft.com/en-us/dotnet/api/microsoft.data.sqlclient.sqlconnection.serverprocessid?view=sqlclient-dotnet-standard-5.2)?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
```diff
@@ -6,6 +6,7 @@
 using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.Data.SqlClient.Microsoft.Data.SqlClientX.IO;
```
cheenamalhotra, line 9: Please fix namespace to "Microsoft.Data.SqlClientX.IO"

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReadStream.cs
```diff
@@ -0,0 +1,341 @@
...
+                // We move the header information to the beginning of the buffer.
+                Buffer.BlockCopy(_readBuffer, _readIndex, _readBuffer, 0, headerDataAvailable);
+                _readBufferDataEnd = headerDataAvailable;
+                _readIndex = 0;
+
+                while (bytesNeededToCompleteHeader > 0)
+                {
+                    int bytesRead = isAsync ? 
+                        await _underlyingStream.ReadAsync(_readBuffer.AsMemory(_readBufferDataEnd), ct).ConfigureAwait(false) 
+                            : _underlyingStream.Read(_readBuffer, _readBufferDataEnd, bytesNeededToCompleteHeader);
```
edwardneal, line 317:
```suggestion
                            : _underlyingStream.Read(_readBuffer.AsSpan(_readBufferDataEnd));
```

If more data than expected is available, this'll allow that data to be copied into the buffer, which should reduce the need for case 1.3 in `PrepareBufferAsync`.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReadStream.cs
```diff
@@ -0,0 +1,341 @@
...
+        private async ValueTask ProcessHeaderAsync(bool isAsync, CancellationToken ct)
+        {
+            int headerDataAvailable = _readBufferDataEnd - _readIndex;
+            int bytesNeededToCompleteHeader = TdsEnums.HEADER_LEN - headerDataAvailable;
+
+            // We have less than the header length available in the buffer, so we need to read more data, to atleast complete 
+            // the header.
+            if (headerDataAvailable < TdsEnums.HEADER_LEN)
+            {
+                // We move the header information to the beginning of the buffer.
```
edwardneal, line 308: Do we need to do this, or can we read to index zero in `_readBuffer` from the outset?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReadStream.cs
```diff
@@ -0,0 +1,341 @@
...
+    {
+        #region Private Fields
+        
+        private Stream _underlyingStream;
+
+        // The buffer to hold the TDS read data.
+        private byte[] _readBuffer;
+
+        // The read pointer inside the buffer.
+        private int _readIndex { get; set; } = 0;
```
edwardneal, line 27: Do you mean for these to be fields? They're properties at the moment. `TdsWriteStream` has the same pattern.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReadStream.cs
```diff
@@ -0,0 +1,341 @@
...
+        public override int Read(Span<byte> buffer)
+        {
+            int lengthToFill = buffer.Length;
+            int totalRead = 0;
+            while (lengthToFill > 0)
+            {
+                PrepareBufferIfNeeded(isAsync: false, CancellationToken.None).ConfigureAwait(false);
+
+                int lengthToCopy = MinDataAvailableBeforeRead(lengthToFill);
+                ReadOnlySpan<byte> copyFrom = _readBuffer.AsSpan(_readIndex, lengthToCopy);
```
edwardneal, line 119: Parallel comment to TdsWriteStream - I think there'll be situations where we're reading large amounts of data from the network (large varbinary blobs) and we can Read straight into the buffer rather than copying twice.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReadStream.cs
```diff
@@ -0,0 +1,341 @@
...
+
+            // There is no data left in the buffer.
+            if (_readIndex == _readBufferDataEnd)
+            {
+                // If we have left over data indicated in the packet header, then we simply need to get data from the network.
+                if (_packetDataLeft > 0)
+                {
+                    _readBufferDataEnd = isAsync ?
+                        await _underlyingStream.ReadAsync(_readBuffer, ct).ConfigureAwait(false) :
+                        _underlyingStream.Read(_readBuffer);
```
edwardneal, line 250: Speaking generally, it might be worth making the CancellationToken throw if cancelled after each Read/Write in the sync case.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
```diff
@@ -147,31 +152,19 @@ public override void Write(byte[] buffer, int offset, int count)
         /// <param name="ct"></param>
         /// <returns></returns>
         /// <exception cref="NotImplementedException"></exception>
-        public override async Task FlushAsync(CancellationToken ct)
-            => await _writeStream.FlushAsync(ct).ConfigureAwait(false);
```
edwardneal, line 151: Is this intentional? `TdsWriteStream.FlushAsync` is implemented.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
```diff
@@ -77,39 +80,41 @@ public void ReplaceUnderlyingStream(Stream stream)
...
         {
-            _writeStream.SetPacketSize(packetSize);
-            // TODO: Do this for the read stream as well.
+            throw new NotImplementedException();
+        }
+
+        /// <summary>
+        /// When writing the packet, the caller needs to 
+        /// specify the packet type. 
+        /// TODO: Consider accepting an enum of packet types
```
edwardneal, line 89: This is already done.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReadStream.cs
cheenamalhotra: This is a great point, I think generally applies to all sync end calls on I/O.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReadStream.cs
saurabh500: I hadnt realized that this field has a public surface area associated with it. 

I will make this a Property which will be gettable and defaults to 0 (to match the current SqlConnection behavior). 
Then this needs to be exposed via `TdsStream` 

The SqlConnection will use the SqlInternalConnection, and that will have a handle to TdsStream, which can then expose the SPID on the internal connection via TdsStream.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
saurabh500: Will do. Visual Studio creates this default NS based on the assembly name and the folder of the class. 
I will pay more attention to it.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
saurabh500: Sorry my bad. I will fix it .

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
saurabh500: True. Will fix in next commit

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReadStream.cs
saurabh500: Yeah, I have not done a great job with honoring the Cancellation token. 

I intend to do a sweep of the streams and handle them. 

I am hoping to write unit tests for honoring CTs. So I will punt this feedback to a later iteration when I have UT project checked in, and can then tackle cancellation token for Read/Write and TdsStream.

saurabh500: Correction: I have not done anything with cancellation token at all, except pass it downstream. I will take care of it in subsequent PRs.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReadStream.cs
saurabh500: I don't and thanks for pointing this out. I will make them fields. 

_spid will be a field which will back a property that I will introduce.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
```diff
@@ -50,9 +52,10 @@ internal class TdsStream : Stream, ITdsWriteStream
             set => _writeStream.PacketHeaderType = value; 
         }
 
-        public TdsStream(Stream underLyingStream, TdsWriteStream writeStream) : base()
+        public TdsStream(Stream underLyingStream, TdsWriteStream writeStream, TdsReadStream readStream) : base()
```
saurabh500, line 55: The underlying stream is not needed.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
saurabh500: We dont need this method at all. There is a Property `public TdsStreamPacketType? PacketHeaderType ` which does this. I will delete this method.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
saurabh500: Also @edwardneal  Thanks for catching it. Appreciate the thorough review.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReadStream.cs
```diff
@@ -0,0 +1,340 @@
...
+
+        /// <inheritdoc />
+        public virtual int Spid => _spid;
+
+        #endregion
+
+        #region Public Methods
+
+        /// <inheritdoc />
+        public override async ValueTask DisposeAsync()
```
benrr101, line 94: I'm not sure the best way to do this, but afaik if we want to use the `async using` operator, we need the class to implement `IAsyncDisposable` but that isn't available in all versions of dotnet. It's not hugely important, but something to consider.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReadStream.cs
```diff
@@ -0,0 +1,340 @@
...
+            {
+                // If we have left over data indicated in the packet header, then we simply need to get data from the network.
+                if (_packetDataLeft > 0)
+                {
+                    _readBufferDataEnd = isAsync ?
+                        await _underlyingStream.ReadAsync(_readBuffer, ct).ConfigureAwait(false) :
+                        _underlyingStream.Read(_readBuffer);
+                    _readIndex = 0;
+                }
+                // 1.2. There is no data left as indicated by packet header and the buffer is empty.
```
benrr101, line 258: If we have 1.2 and 1.3, where's 1.1? 😃

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReadStream.cs
```diff
@@ -0,0 +1,340 @@
...
+        }
+
+        /// <summary>
+        /// If the buffer is empty, or the packet data is completely read, 
+        /// then we need to prepare the buffer for the next packet.
+        /// </summary>
+        /// <param name="isAsync"></param>
+        /// <param name="ct"></param>
+        /// <returns></returns>
+        private async ValueTask PrepareBufferIfNeeded(bool isAsync, CancellationToken ct)
```
benrr101, line 288: We could just roll this into PrepareBufferAsync and invert the condition to bail out early if not needed.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReadStream.cs
```diff
@@ -0,0 +1,340 @@
...
+        {
+            // If there we have read through the packet, 
+            // or if we have reached the end of the buffer,
+            // then we need to position the buffer at the beginning of packet or buffer
+            if (_packetDataLeft == 0 || _readBufferDataEnd == _readIndex)
```
benrr101, line 293: Nit: not sure if there's a rule about this, but there should be - even if it's one line, if statements should have curly braces

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReadStream.cs
```diff
@@ -0,0 +1,340 @@
...
+        /// This is not necessarily the number of bytes available in the buffer.
+        /// </summary>
+        private int _packetDataLeft = 0;
+
+        /// <summary>
+        /// The number of bytes in the header, according to the header.
+        /// </summary>
+        private int _packetHeaderDataLength = 0;
+
+        private byte _packetHeaderType = 0;
```
benrr101, line 45: Wouldn't this need to be exposed via a property? I'm thinking whoever is consuming the packets will want to be able to verify the type of packet that's available.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
```diff
@@ -50,9 +52,13 @@ internal class TdsStream : Stream, ITdsWriteStream
             set => _writeStream.PacketHeaderType = value; 
         }
 
-        public TdsStream(Stream underLyingStream, TdsWriteStream writeStream) : base()
+        /// <inheritdoc />
+        public int Spid => _readStream.Spid;
+
+        public TdsStream(TdsWriteStream writeStream, TdsReadStream readStream) : base()
```
benrr101, line 58: Comment :)

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
```diff
@@ -215,13 +183,14 @@ public virtual void QueueCancellation()
         public override async ValueTask DisposeAsync()
         {
             await _writeStream.DisposeAsync().ConfigureAwait(false);
-            // TODO: For Read Stream as well.
+            await _readStream.DisposeAsync().ConfigureAwait(false);
         }
 
         /// <inheritdoc />
         protected override void Dispose(bool disposing)
```
benrr101, line 190: I'm not an expert on this, but from what I've seen there's a pattern that disposing should be checked for true before calling dispose, and the underlying shouldn't be disposed if the object was already disposed. May be moot if the underlying will check this, but I want to make sure we follow the best practices for disposal.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReadStream.cs
```diff
@@ -0,0 +1,340 @@
...
+        /// <summary>
+        /// The number of bytes left in the packet to be consumed.
+        /// This is not necessarily the number of bytes available in the buffer.
+        /// </summary>
+        private int _packetDataLeft = 0;
+
+        /// <summary>
+        /// The number of bytes in the header, according to the header.
+        /// </summary>
+        private int _packetHeaderDataLength = 0;
```
benrr101, line 43: This member is assigned but never used after its been assigned

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReadStream.cs
```diff
@@ -0,0 +1,340 @@
...
+
+        private byte _packetHeaderType = 0;
+
+        private byte _packetStatus;
```
benrr101, line 47: This one is also assigned but never used

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReadStream.cs
```diff
@@ -0,0 +1,340 @@
...
+        }
+
+        /// <inheritdoc />
+        public override int Read(Span<byte> buffer)
+        {
+            int lengthToFill = buffer.Length;
+            int totalRead = 0;
+            while (lengthToFill > 0)
+            {
+                PrepareBufferIfNeeded(isAsync: false, CancellationToken.None).ConfigureAwait(false);
```
benrr101, line 121: I think you meant to await this or use the result somehow? This one is giving me the following message https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2012

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
```diff
@@ -66,50 +72,40 @@ public TdsStream(Stream underLyingStream, TdsWriteStream writeStream) : base()
...
+            _readStream.ReplaceUnderlyingStream(stream);
         }
 
         /// <summary>
         /// Allows manipulation of the packet size. This should only be used after login ack,
         /// when the client and server exchange the negotiated packet size.
         /// </summary>
         /// <param name="packetSize">The negotiated packet size</param>
-        /// <exception cref="NotImplementedException"></exception>
         public void SetPacketSize(int packetSize)
```
edwardneal, line 83: Very tentative, but it's worth experimenting: when the packet size changes, perhaps the underlying socket's `ReceiveBufferSize` and `SendBufferSize` should change, accounting for TDS+MARS+SSL header sizes. I'm not sure whether this'd be an all-round performance improvement, or a CPU usage/network latency tradeoff.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReadStream.cs
```diff
@@ -0,0 +1,340 @@
...
+        public void SetPacketSize(int bufferSize)
+        {
+            _readBuffer = new byte[bufferSize];
+        }
+
+        /// <inheritdoc />
+        public async ValueTask SkipReadBytesAsync(int skipCount, bool isAsync, CancellationToken ct)
+        {
+            int lengthLeftToSkip = skipCount;
+            int totalRead = 0;
```
benrr101, line 182: This variable is being updated but its never used after being updated. So it can probabaly be safely removed

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReadStream.cs
```diff
@@ -0,0 +1,340 @@
...
+        // TODO: Handle Cancellation tokens in all async paths.
+        #region Private Fields
+
+        private Stream _underlyingStream;
+
+        // The buffer to hold the TDS read data.
+        private byte[] _readBuffer;
+
+        // The read pointer inside the buffer.
+        private int _readIndex = 0;
```
benrr101, line 29: nit: Initializing to the default value is redundant
nit: Either `//` comments, `///` comments, or no comments? I'd be ok with just `//` comments for private fields.

benrr101 (changes requested):
A handful are small potatoes, but a couple are bigger potatoes (assigned but not used, dispose async behavior, and edward's comment about copying the header to the beginning of the buffer).

I get this gut feeling that there are too many pointers/range variables being used and that it could be simplified. Or the logic for updating the buffer could be broken up a bit to make it more readable. But, it's just a feeling.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReadStream.cs
saurabh500: Well, in the current implementation of the driver, this is never read.
However, for the completeness of this Stream, I will expose a property.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReadStream.cs
saurabh500: This stream inherits from Stream.
Net 6.0 Stream had a DisposeAsync method, which was expected to be called explicitly. 
Net8.0 got the await using goodness. 
So by inheriting from Stream, we remain compatible with both versions.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReadStream.cs
saurabh500: I called the method with isAsync -> False.
When it returns, I know that I dont have to wait, and it would have modified all the states of this class when it returns.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReadStream.cs
saurabh500: yes

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReadStream.cs
saurabh500: First my brain was structured, and came up with a numbering scheme :D 
Then it got lost in the split packet madness, and the structure was lost. 

I will fix this.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReadStream.cs
saurabh500: Yes makes sense

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReadStream.cs
saurabh500: This is a special case, where a MITM is splitting the packets (inspired by the comments in the current sqlclient implementation), and we get a part of the next packet while reading an existing packet, but didnt get the full header. so we are talking about less than 8 bytes of header in the buffer.
At this point we need the full header, and allow as much space in the buffer to accomodate the full TDS packet.

saurabh500: Going to resolve this conversation. But feel free to reopen if I am missing something.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
saurabh500: Cmonnnnnn ... Its a constructor 😄 
I will add comments

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReadStream.cs
```diff
@@ -0,0 +1,340 @@
...
+        /// <inheritdoc />
+        public override bool CanWrite => false;
+
+        /// <inheritdoc />
+        public override long Length => throw new NotSupportedException();
+
+        /// <inheritdoc />
+        public override long Position 
+        {
+            get => throw new NotImplementedException();
```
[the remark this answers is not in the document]
saurabh500: Should be not supported

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReadStream.cs
saurabh500: Exposed for completeness of exposing the header.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReadStream.cs
saurabh500: I would like to keep the initialization. While dealing with so many numbers and offsets, the initialization, though redundant helps understand code better.
In the current implementation of the SqlClient, I always get lost in what is initialized, what is default and where.

Fixed the comments style

saurabh500:
Note to self: 
First add tests,
then address @edwardneal 's comment about large data perf improvement. 
And then make sure that the tests still work. 

Thanks for the comments @cheenamalhotra  @benrr101  @edwardneal 

I still have to respond to a few. I will work on this PR and respond as soon as I have the answers.

saurabh500:
@benrr101  and @cheenamalhotra  this is ready.
For @edwardneal comment on large data read, I hope I have answered the question in the comment. Added a TODO in the class to revisit this though

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReadStream.cs
benrr101: Ok, works for me.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReadStream.cs
benrr101: I've definitely done the same thing before

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReadStream.cs
edwardneal: Thanks @saurabh500, I understand and agree with your point - we want to minimise Reads. Having approximately `2 * ([requested data size] / [packet size])` Read calls will become painful if we choose a 4KB packet size and want to read a 5MB payload. This'd result in 1280 packets and 2560 Read calls; it'd probably be faster than copying 4KB of memory 1280 times, but it's still hardly ideal.

A compromise might be the best of all worlds, at the expense of 8 bytes in the buffer.
* If we guarantee that there'll always be space in the buffer for 8 bytes (i.e. when preparing the buffer at the start of a new packet, reset the current position if it's < 8 bytes away from the end of the buffer) then we simplify the parsing of the packet header significantly.
* We can fill up to the end of the buffer from the network (so we know we're always reading in large chunks), copying from the buffer to the output `Memory<byte>`/`Span<byte>`
* If we reach the end of the buffer and haven't finished writing to the output, we can Read from the underlying stream to the output Memory/Span directly
  * There might be a safeguard needed here to stop us from accidentally reading from beyond the end of the TDS packet.
  * I *think* the position in the read buffer should be trivial enough to calculate as `([offset into buffer at start of Read] + MIN([length of output Span], [buffer size] - [offset into buffer at start of Read])) MOD [buffer size]`, so we can keep that in sync.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReadStream.cs
benrr101: I see what you're saying and generally speaking I agree, but I think the code analysis rule explanation does make sense and is a better general rule to follow. If it completes synchronously (which it's supposed to), would it hurt to await it (I guess in this case via .GetResult())?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
benrr101: I have PTSD from DSCT's overly strict rules about comments. Fortunately I could just click "Add standard comment" and it would be happy. And the standard comment was something like "Constructs and initialized a new instance of <see cref="TdsStream" />". I won't complain too hard about not including such comments, and I definitely will push for turning off the "Comments must end with `.`" rule, if we ever adopt codeanalysis rules.

benrr101 (approved):
Generally speaking looks good

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
saurabh500: BTW @benrr101  my comment about "Its a constructor" was not a serious one. 

I have a strong belief that code is read a lot more than it is written. So it is always good to over-comment than not. So keep the feedback coming. There is a reason I address them as well 😄 

I think we need to think of a solution to enforce the code style/comment etc during the build of the code so that the code author can be asked to improve the code with builds. This way the reviewer doesn't have a cognitive overload of finding problems with the style of code. And this needs to be done in a way that only new code is flagged, and the old code stays.

Honestly, I wished that each field of TdsParserStateObject was documented, or documented better. Otherwise we keep reverse engineering the intentions.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReadStream.cs
```diff
@@ -0,0 +1,355 @@
...
+        public override int Read(Span<byte> buffer)
+        {
+            int lengthToFill = buffer.Length;
+            int totalRead = 0;
+            while (lengthToFill > 0)
+            {
+                PrepareBufferAsync(isAsync: false, CancellationToken.None).ConfigureAwait(false);
+
+                int lengthToCopy = MinDataAvailableBeforeRead(lengthToFill);
+                ReadOnlySpan<byte> copyFrom = _readBuffer.AsSpan(_readIndex, lengthToCopy);
```
cheenamalhotra, line 141: Suggestion: You could also consider reading via [MemoryStream](https://learn.microsoft.com/en-us/dotnet/api/system.io.memorystream?view=net-8.0) instead of readBuffer, and read as:
`int bytesRead = _memoryStream.Read(buffer.Slice(totalRead, lengthToCopy));`

saurabh500: @cheenamalhotra I didn't completely understand the comment about memory stream, and in what cases would it help?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReadStream.cs
saurabh500: I took a slightly different route. 
I consumed the VT in an assert.

PR #2583, opened 2024-06-17
Expose Protocol via DataSource Object

benrr101:
**Description**: Converting `DataSource._connectionProtocol` to an auto-property `DataSource.ResolvedProtocol`. This is to allow the transport creation handler access to the protocol resolved during parsing of data source string. Since this is an important change and has impact on non-X code, I think this needs its own PR.

**Testing**: Project builds (well ... only if the SqlConnectionX class is fixed)


saurabh500 (approved):
LGTM. 
This can go into Main as well. 
I dont see behavior changes.

cheenamalhotra (approved):
Agreed, consider pushing to main.

PR #2584, opened 2024-06-17
Make internals visible to, for Unit testing

saurabh500:
For ClientX, we need to be able to unit test the code under review.

This PR exposes the internals to the FunctionalTests assembly, which doesn't have server requirements.

An extension to this could be to have a separate project for Unit tests, and then contribute UTs there. That would have some dependency on the pipelines to be changed, to build and test ClientX. 
Till then we can contribute UTs in a different folder in FunctionalTests or Microsoft.Data.SqlClient.Tests project

Validated by modifying a test to reference SqlInternalConnectionTds, and there were no compilation errors


cheenamalhotra (approved):

saurabh500:
There is a PROBLEM with this PR. There are side-effects of using InternalsVisibleTo with SqlClient project. 
There are some types which become ambiguous when InternalsVisibleTo is made available

PR #2592, opened 2024-06-18
Adding DataSourceParser and ConnectionHandlerContext

samsharma2700:
1). Adding necessary handler and helpers for data source parsing. 
2). Updating ConnectionHandlerContext so as to be used by all other handlers.
3). Updated HandlerRequestType since we will not be using ConnectionRequest anymore.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/ConnectionHandlerContext.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/ConnectionHandlerContext.cs
@@ -1,0 +1,35 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+using System;
+using System.IO;
+using Microsoft.Data.SqlClient.SNI;
+using Microsoft.Data.SqlClientX.Handlers;
+
+namespace Microsoft.Data.SqlClient.Microsoft.Data.SqlClientX.Handlers
+{    /// <summary>
+     /// Class that contains all context data needed for various handlers.
+     /// </summary>
+    internal class ConnectionHandlerContext : HandlerRequest
+    {
+        /// <summary>
+        /// Class that contains data required to handle a connection request.
+        /// </summary>
+        public String connectionString { get; set; }
+
+        /// <summary>
+        /// Stream used by readers.
+        /// </summary>
+        public Stream ConnectionStream { get; set; }
+
+        /// <summary>
+        /// Class required by DataSourceParser and Transport layer.
+        /// </summary>
+        public DataSource dataSource { get; set; }
+
+        /// <summary>
+        /// Class used by orchestrator while chaining handlers.
+        /// </summary>
+        public Exception error { get; set; }
+    }
+}
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/ConnectionHandlerContext.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/ConnectionHandlerContext.cs
@@ -6,7 +6,7 @@ using System.IO;
 using Microsoft.Data.SqlClient.SNI;
 using Microsoft.Data.SqlClientX.Handlers;
 
-namespace Microsoft.Data.SqlClient.Microsoft.Data.SqlClientX.Handlers
+namespace Microsoft.Data.SqlClientX.Handlers
 {    /// <summary>
      /// Class that contains all context data needed for various handlers.
      /// </summary>
@@ -15,7 +15,7 @@ namespace Microsoft.Data.SqlClient.Microsoft.Data.SqlClientX.Handlers
         /// <summary>
         /// Class that contains data required to handle a connection request.
         /// </summary>
-        public String connectionString { get; set; }
+        public SqlConnectionString ConnectionString { get; set; }
 
         /// <summary>
         /// Stream used by readers.
@@ -25,11 +25,11 @@ namespace Microsoft.Data.SqlClient.Microsoft.Data.SqlClientX.Handlers
         /// <summary>
         /// Class required by DataSourceParser and Transport layer.
         /// </summary>
-        public DataSource dataSource { get; set; }
+        public DataSource DataSource { get; set; }
 
         /// <summary>
         /// Class used by orchestrator while chaining handlers.
         /// </summary>
-        public Exception error { get; set; }
+        public Exception Error { get; set; }
     }
 }
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/DataSourceParser.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/DataSourceParser.cs
@@ -9,6 +9,9 @@ using Microsoft.Data.SqlClientX.Handlers;
 
 namespace Microsoft.Data.SqlClient.Microsoft.Data.SqlClientX.Handlers
 {
+    /// <summary>
+    /// Handler to populate data source information required for transport handler
+    /// </summary>
     internal class DataSourceParser : IHandler<ConnectionHandlerContext>
     {
         public IHandler<ConnectionHandlerContext> NextHandler { get; set; }
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/DataSourceParser.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/DataSourceParser.cs
@@ -1,0 +1,74 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Data.SqlClient.SNI;
+using Microsoft.Data.SqlClientX.Handlers;
+
+namespace Microsoft.Data.SqlClient.Microsoft.Data.SqlClientX.Handlers
+{
+    internal class DataSourceParser : IHandler<ConnectionHandlerContext>
+    {
+        public IHandler<ConnectionHandlerContext> NextHandler { get; set; }
+
+        public async ValueTask Handle(ConnectionHandlerContext request, bool isAsync, CancellationToken ct)
+        {
+            SqlConnectionString result = new SqlConnectionString(request.connectionString);
+            ServerInfo serverInfo = new ServerInfo(result);
+            string fullServerName = serverInfo.ExtendedServerName;
+            string localDBDataSource = GetLocalDBDataSource(fullServerName, out bool errorWithLocalDBProcessing);
+
+            if (errorWithLocalDBProcessing)
+            {
+                request.error = new Exception(localDBDataSource);
+                return;
+            }
+            
+            // If a localDB Data source is available, we need to use it.
+            fullServerName = localDBDataSource ?? fullServerName;
+            DataSource details = DataSource.ParseServerName(fullServerName);
+            if (details == null)
+            {
+                request.error = new Exception(localDBDataSource);
+                return;
+            }
+            
+            request.dataSource = details;
+            
+            if (NextHandler is not null)
+            {
+                await NextHandler.Handle(request, isAsync, ct).ConfigureAwait(false);
+            }
+        }
+
+        private static string GetLocalDBDataSource(string fullServerName, out bool error)
+        {
+            string localDBConnectionString = null;
+            string localDBInstance = DataSource.GetLocalDBInstance(fullServerName, out bool isBadLocalDBDataSource);
+
+            if (isBadLocalDBDataSource)
+            {
+                error = true;
+                return null;
+            }
+
+            else if (!string.IsNullOrEmpty(localDBInstance))
+            {
+                // We have successfully received a localDBInstance which is valid.
+                localDBConnectionString = LocalDB.GetLocalDBConnectionString(localDBInstance);
+
+                if (fullServerName == null || string.IsNullOrEmpty(localDBConnectionString))
+                {
+                    // The Last error is set in LocalDB.GetLocalDBConnectionString. We don't need to set Last here.
+                    error = true;
+                    return null;
+                }
+            }
+            
+            error = false;
+            return localDBConnectionString;
+        }
+    }
+}
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/DataSourceParsingHandler.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/DataSourceParsingHandler.cs
@@ -6,26 +6,31 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Data.SqlClient.SNI;
 using Microsoft.Data.SqlClientX.Handlers;
+using static Microsoft.Data.SqlClient.SNINativeMethodWrapper;
 
-namespace Microsoft.Data.SqlClient.Microsoft.Data.SqlClientX.Handlers
+namespace Microsoft.Data.SqlClientX.Handlers
 {
     /// <summary>
     /// Handler to populate data source information required for transport handler
     /// </summary>
-    internal class DataSourceParser : IHandler<ConnectionHandlerContext>
+    internal class DataSourceParsingHandler : IHandler<ConnectionHandlerContext>
     {
+        /// <inheritdoc />
         public IHandler<ConnectionHandlerContext> NextHandler { get; set; }
 
+        /// <inheritdoc />
         public async ValueTask Handle(ConnectionHandlerContext request, bool isAsync, CancellationToken ct)
         {
-            SqlConnectionString result = new SqlConnectionString(request.connectionString);
-            ServerInfo serverInfo = new ServerInfo(result);
+            ServerInfo serverInfo = new ServerInfo(request.ConnectionString);
             string fullServerName = serverInfo.ExtendedServerName;
             string localDBDataSource = GetLocalDBDataSource(fullServerName, out bool errorWithLocalDBProcessing);
 
             if (errorWithLocalDBProcessing)
             {
-                request.error = new Exception(localDBDataSource);
+                //TODO: populate other details for the SQLException
+                SqlErrorCollection collection = new SqlErrorCollection();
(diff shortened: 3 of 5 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/ConnectionHandlerContext.cs
```diff
@@ -0,0 +1,35 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+using System;
+using System.IO;
+using Microsoft.Data.SqlClient.SNI;
+using Microsoft.Data.SqlClientX.Handlers;
+
+namespace Microsoft.Data.SqlClient.Microsoft.Data.SqlClientX.Handlers
```
saurabh500, line 9:
```suggestion
namespace Microsoft.Data.SqlClientX.Handlers
```

VS has generated lot of bad such namespaces for us :)

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/ConnectionHandlerContext.cs
```diff
@@ -0,0 +1,35 @@
...
+{    /// <summary>
+     /// Class that contains all context data needed for various handlers.
+     /// </summary>
+    internal class ConnectionHandlerContext : HandlerRequest
+    {
+        /// <summary>
+        /// Class that contains data required to handle a connection request.
+        /// </summary>
+        public String connectionString { get; set; }
```
saurabh500, line 18: Property names start with Caps 

```suggestion
        public String ConnectionString { get; set; }
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/ConnectionHandlerContext.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        /// <summary>
+        /// Stream used by readers.
+        /// </summary>
+        public Stream ConnectionStream { get; set; }
+
+        /// <summary>
+        /// Class required by DataSourceParser and Transport layer.
+        /// </summary>
+        public DataSource dataSource { get; set; }
```
saurabh500, line 28:
```suggestion
        public DataSource DataSource { get; set; }
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/DataSourceParser.cs
```diff
@@ -0,0 +1,77 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Data.SqlClient.SNI;
+using Microsoft.Data.SqlClientX.Handlers;
+
+namespace Microsoft.Data.SqlClient.Microsoft.Data.SqlClientX.Handlers
```
saurabh500, line 10:
```suggestion
namespace Microsoft.Data.SqlClientX.Handlers
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/DataSourceParser.cs
```diff
@@ -0,0 +1,77 @@
...
+    /// <summary>
+    /// Handler to populate data source information required for transport handler
+    /// </summary>
+    internal class DataSourceParser : IHandler<ConnectionHandlerContext>
+    {
+        public IHandler<ConnectionHandlerContext> NextHandler { get; set; }
+
+        public async ValueTask Handle(ConnectionHandlerContext request, bool isAsync, CancellationToken ct)
+        {
+            SqlConnectionString result = new SqlConnectionString(request.connectionString);
```
saurabh500, line 21: Move this object of the type `SqlConnectionString` into the `ConnectionHandlerContext` itself. 
Almost every handler will need to create it. So might as well put it in the Context. 
The existing code of SqlClient also uses a SqlConnectionString option for getting the connection string properties out.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/HandlerRequestType.cs
```diff
@@ -13,6 +13,6 @@ internal enum HandlerRequestType
         /// <summary>
         /// Request type for connection request.
```
saurabh500, line 14: Modify the comment to match the variable name

Something on the lines of `Context for Connection Request`

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/DataSourceParser.cs
```diff
@@ -0,0 +1,77 @@
...
+            ServerInfo serverInfo = new ServerInfo(result);
+            string fullServerName = serverInfo.ExtendedServerName;
+            string localDBDataSource = GetLocalDBDataSource(fullServerName, out bool errorWithLocalDBProcessing);
+
+            if (errorWithLocalDBProcessing)
+            {
+                request.error = new Exception(localDBDataSource);
```
saurabh500, line 28: Wouldnt `localDBDataSource` be null if there was an error with local db processing ?

saurabh500: If yes, the consider adding an exception message to state what really happened?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/DataSourceParser.cs
```diff
@@ -0,0 +1,77 @@
...
+using System.Threading.Tasks;
+using Microsoft.Data.SqlClient.SNI;
+using Microsoft.Data.SqlClientX.Handlers;
+
+namespace Microsoft.Data.SqlClient.Microsoft.Data.SqlClientX.Handlers
+{
+    /// <summary>
+    /// Handler to populate data source information required for transport handler
+    /// </summary>
+    internal class DataSourceParser : IHandler<ConnectionHandlerContext>
```
saurabh500, line 15: Can the class be called `DataSourceParsingHandler` ? 
I am hoping to look at the class name and know that it was a handler.

saurabh500:
There should have been changes to the `csproj` file as well. You would need to add the csproj modifications to the PR too.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/ConnectionHandlerContext.cs
cheenamalhotra, line 18: Also, use `string` instead of `String`.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/DataSourceParser.cs
cheenamalhotra: Instead of Exception, SqlClient should only be throwing `SqlException`, well formed with `SqlError` - do look at some existing usage of it's constructors, and ensure the exception message comes from `Strings.resx` so it can be localized.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/ConnectionHandlerContext.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        /// <summary>
+        /// Class used by orchestrator while chaining handlers.
+        /// </summary>
+        public Exception error { get; set; }
```
cheenamalhotra, line 33:
```suggestion
        public SqlException error { get; set; }
```
Use standard exception only.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/DataSourceParser.cs
```diff
@@ -0,0 +1,77 @@
...
+                request.error = new Exception(localDBDataSource);
+                return;
+            }
+            
+            // If a localDB Data source is available, we need to use it.
+            fullServerName = localDBDataSource ?? fullServerName;
+            DataSource details = DataSource.ParseServerName(fullServerName);
+            if (details == null)
+            {
+                request.error = new Exception(localDBDataSource);
```
cheenamalhotra, line 37: Same comment here, use SqlException.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/ConnectionHandlerContext.cs
saurabh500, line 33: The base class already has an exception being exposed isnt it?

cheenamalhotra: Should it be a SqlException? What other kinds of exceptions do we expect?

saurabh500: @cheenamalhotra  
I dont think I can tell you about the Class Types of the exceptions, but definitely a lot of transport related errors, (broken connections, timeouts), timeouts, 

Then there would errors coming from MSAL for AccessToken retrieval.

There is always OOM. 

While parsing the protocol, we definitely respond to Error tokens with SqlException messages. 

I feel we will need to stay away from SqlException in the base class since SqlException is only supposed to have the exceptions that is related to Sql Connectivity/operations etc. 

So likely we need some kind of indicator whether the Exception in the HandlerContext should be considered a SqlException or not. 

e.g. for any errors with driver implementation because of which we couldn't read a byte, we throw a generic exception 

image

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/DataSourceParser.cs
saurabh500: @cheenamalhotra  taking a dependency on SqlException may not be the right thing to do. 
I have explained in another comment above.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/ConnectionHandlerContext.cs
saurabh500: Also when we try to execute a command on the same connection, twice, I believe we get an `InvalidOperation` exception.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/ConnectionHandlerContext.cs
benrr101: I can fix that :) We basically just need to remove the default namespace from the project files. I'll send that out for a PR against main

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/ConnectionHandlerContext.cs
benrr101: Agreed (I think) - I think the type of the exception in this class should stay as `Exception`

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/DataSourceParser.cs
```diff
@@ -0,0 +1,77 @@
...
+using Microsoft.Data.SqlClientX.Handlers;
+
+namespace Microsoft.Data.SqlClient.Microsoft.Data.SqlClientX.Handlers
+{
+    /// <summary>
+    /// Handler to populate data source information required for transport handler
+    /// </summary>
+    internal class DataSourceParser : IHandler<ConnectionHandlerContext>
+    {
+        public IHandler<ConnectionHandlerContext> NextHandler { get; set; }
```
benrr101, line 17: Should add `<inheritDoc />` comment here

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/DataSourceParser.cs
```diff
@@ -0,0 +1,77 @@
...
+
+        public async ValueTask Handle(ConnectionHandlerContext request, bool isAsync, CancellationToken ct)
```
benrr101, line 19: Should also add `<inheritDoc />` comment here, too

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/DataSourceParser.cs
benrr101: I wonder is this something that will be provided before starting the handler chain?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/DataSourceParser.cs
```diff
@@ -0,0 +1,77 @@
...
+        {
+            SqlConnectionString result = new SqlConnectionString(request.connectionString);
+            ServerInfo serverInfo = new ServerInfo(result);
+            string fullServerName = serverInfo.ExtendedServerName;
+            string localDBDataSource = GetLocalDBDataSource(fullServerName, out bool errorWithLocalDBProcessing);
```
benrr101, line 24: Nit: Names should be PascalCase or camelCase, even with acronyms. So this one would be `localDbDataSource`

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/DataSourceParser.cs
```diff
@@ -0,0 +1,77 @@
...
+            
+            request.dataSource = details;
+            
+            if (NextHandler is not null)
+            {
+                await NextHandler.Handle(request, isAsync, ct).ConfigureAwait(false);
+            }
+        }
+
+        private static string GetLocalDBDataSource(string fullServerName, out bool error)
```
benrr101, line 49: Two thoughts here:
* I don't think we should be returning an error flag via an out argument. Since all the error conditions in this method return `null`, can we just have `null` indicate that there's an error? Or can we just throw from here?
* If we cannot eliminate the error flag, I think the more common pattern for methods with error flags is to return bool and have the value returned via an out param. This makes it easier to use it as the condition of an `if` statement. Eg `private static bool TryGetLocalDbDataSource(string fullServerName, out string localDbDataSource)`

benrr101 (changes requested):

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/ConnectionHandlerContext.cs
samsharma2700: Awesome! thanks Ben. I'll update the file for now.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/ConnectionHandlerContext.cs
samsharma2700: We can use SqlErrorCollection to add multiple errors to the SqlException

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/DataSourceParser.cs
samsharma2700: Done

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/DataSourceParser.cs
samsharma2700: Done, replaced the ConnectionString with SqlConnectionString

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/DataSourceParser.cs
samsharma2700: Ack. I added it as a TODO, will refactor when updating the async part.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/DataSourceParsingHandler.cs
```diff
@@ -0,0 +1,85 @@
...
+            }
+            
+            request.DataSource = details;
+            
+            if (NextHandler is not null)
+            {
+                await NextHandler.Handle(request, isAsync, ct).ConfigureAwait(false);
+            }
+        }
+        //TODO: Refactor function for better handling of error flag and return params
```
saurabh500, line 56: Nit: New line before the //todo:

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/DataSourceParsingHandler.cs
```diff
@@ -0,0 +1,85 @@
...
+        {
+            ServerInfo serverInfo = new ServerInfo(request.ConnectionString);
+            string fullServerName = serverInfo.ExtendedServerName;
+            string localDBDataSource = GetLocalDBDataSource(fullServerName, out bool errorWithLocalDBProcessing);
+
+            if (errorWithLocalDBProcessing)
+            {
+                //TODO: populate other details for the SQLException
+                SqlErrorCollection collection = new SqlErrorCollection();
+                collection.Add(new SqlError(0,0,TdsEnums.FATAL_ERROR_CLASS, null, StringsHelper.GetString("LocalDB_UnobtainableMessage"), null, 0));
```
saurabh500, line 32: NIt: This could use a bit of formatting. e.g. space after `,` in the `new SqlError` parameters

saurabh500 (approved):

PR #2595, opened 2024-06-19
SqlClientX: Implementation of TdsWriter

cheenamalhotra:
Adds TdsWriter to expose APIs for writing needs.
Maintains writer buffer, so should be instantiated with the stream working on a TDS Request.


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriter.cs
```diff
@@ -0,0 +1,160 @@
...
+
+using System;
+using System.Buffers.Binary;
+using Microsoft.Data.SqlClient;
+
+namespace Microsoft.Data.SqlClientX.IO
+{
+    internal class TdsWriter
+    {
+        private readonly TdsWriteStream _writeStream;
```
saurabh500, line 15: This and everything else which will interact with stream will only use TdsStream class. TdsWriteStream will not be used.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriter.cs
```diff
@@ -0,0 +1,187 @@
...
+
+        // Header length constants
+        private const int OutputHeaderLen = TdsEnums.HEADER_LEN;
```
saurabh500, line 18: Nothing related to TDS packets length will be tracked in any consumers of the TdsStream. 
The consumers of the TdsStream just keep writing and only flush when they know that the message is ready to be sent out.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriter.cs
```diff
@@ -0,0 +1,187 @@
...
+        private int _outBytesUsed = OutputHeaderLen; // number of bytes used in internal write buffer - initialize past header
+
+        /// <summary>
+        /// Instantiate TdsWriter with a TdsWriteStream
+        /// </summary>
+        /// <param name="stream">A Tds Writer stream to work with</param>
+        /// <exception cref="ArgumentNullException"></exception>
+        public TdsWriter(TdsWriteStream stream)
+        {
+            _writeStream = stream ?? throw new ArgumentNullException(nameof(stream));
```
saurabh500, line 31: Instead of throwing an exception, please add a Debug.Assert. This basically means that a null stream just isn't allowed. There is nothing that a caller can do with the exception. Once the caller makes sure that a non-null stream is sent, then we dont need the null check.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriter.cs
```diff
@@ -0,0 +1,187 @@
...
+
+        /// <summary>
+        /// Writes short value to out buffer
+        /// </summary>
+        /// <param name="v">Value to write</param>
+        /// <param name="isAsync">Is Async caller method</param>
+        /// <param name="ct">Cancellation token</param>
+        public async Task WriteShortAsync(short v, bool isAsync, CancellationToken ct)
+        {
+            await EnsureBufferSpaceAsync(2, isAsync, ct).ConfigureAwait(false);
```
saurabh500, line 45: Here and for everything, there is no need to ensure buffer space. The buffer is only maintained by the stream. 
The job of the caller is to write bytes to it .

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriter.cs
```diff
@@ -0,0 +1,187 @@
...
+            for (int i = 0; i < length; i++)
+            {
+                memory.Span[i] = (byte)((v >> (i * 8)) & 0xFF);
+            }
+            _outBytesUsed += length;
+        }
+        #endregion
+
+        #region Private helpers
+        private async Task EnsureBufferSpaceAsync(int requiredBytes, bool isAsync, CancellationToken ct)
```
saurabh500, line 163: Not needed. The stream takes care of these checks.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriter.cs
```diff
@@ -0,0 +1,187 @@
...
+        {
+            if ((_outBytesUsed + requiredBytes) > _outBuff.Length)
+            {
+                if (isAsync)
+                {
+                    await _writeStream.WriteAsync(_outBuff.AsMemory(0, _outBytesUsed), ct).ConfigureAwait(false);
```
saurabh500, line 169: For any writes to the stream, you dont have to split data into _outbytes etc.
You can dump an arbitrarily large array of bytes to it, and it will chunk the bytes and wrap it into packets and send it out

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriter.cs
```diff
@@ -0,0 +1,187 @@
...
+
+        #region Public APIs
+
+        /// <summary>
+        /// Writes short value to out buffer
+        /// </summary>
+        /// <param name="v">Value to write</param>
+        /// <param name="isAsync">Is Async caller method</param>
+        /// <param name="ct">Cancellation token</param>
+        public async Task WriteShortAsync(short v, bool isAsync, CancellationToken ct)
```
Tornhoof, line 43: Why not using CancellationToken cancellationToken = default. That's kinda the default naming for them.

Why v as the name of the value param? You throw exceptions with nameof, having a nicer name would be better imho.

Why Task? You have a sync path and the ``EnsureBufferSpaceAsync`` method has a sync path, so you might want to look at ValueTask, especially if at some point the Stream is replaced with something modern, like Pipeline.

saurabh500: ValueTask is the right choice because the underlying TdsStream also buffers the data before sending out to the network. So the ValueTask overloads of TdsStream can benefit here.

edwardneal: +1 for ValueTask. It'll also avoid a heap allocation in sync cases

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriter.cs
```diff
@@ -0,0 +1,187 @@
...
+            _outBytesUsed += 4;
+        }
+
+        /// <summary>
+        /// Writes long value to out buffer
+        /// </summary>
+        /// <param name="v">Value to write</param>
+        /// <param name="isAsync">Is Async caller method</param>
+        /// <param name="ct">Cancellation token</param>
+        public async Task WriteLongAsync(long v, bool isAsync, CancellationToken ct)
```
edwardneal, line 102: I'd personally think about removing these and force the callers to build up a well-formed `Span` or `Memory`. IMO we're currently encouraging upstream code to perform writes field by field.

NB: in this case, would we need the class at all? The `TdsStream` handles buffering, we'd make building/writing the on-the-wire structure the responsibility of the caller, are there any other reasons to keep it?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriter.cs
```diff
@@ -0,0 +1,187 @@
...
+{
+    internal class TdsWriter
+    {
+        private readonly TdsWriteStream _writeStream;
+
+        // Header length constants
+        private const int OutputHeaderLen = TdsEnums.HEADER_LEN;
+
+        // Out buffer variables
+        private readonly byte[] _outBuff; // internal write buffer - initialize on login
```
benrr101, line 21: Although I don't think the current implementation is necessary (as per Saurabh's comments), I do think it's a good idea to have a preinitialized buffer for storing bytes before sending them all to the stream. This is one benefit to this approach over extension methods - we don't have to share a buffer between all the writers.

However, as it's currently written, there's no thread safety within the writer. Concurrent requests to write a number could trample over the buffer. If this is expected/allowed, then no action necessary. But I think we should at least consciously make that decision.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriter.cs
benrr101: Likewise, for the buffer, we really only need to make it as large as the largest value we want to write to the stream at once (ie, long, 8 bytes)

benrr101 (changes requested):

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriter.cs
```diff
@@ -34,7 +34,7 @@ public TdsWriter(TdsStream stream)
         /// Writes short value to out buffer
         /// </summary>
         /// <param name="v">Value to write</param>
-        /// <param name="isAsync">Is Async caller method</param>
+        /// <param name="isAsync">Whether caller method is executing asynchronously.</param>
```
saurabh500, line 37: I like how you have created these comments uniformly. 
Is there some tool that you are using? Or did you manually copy paste these blurbs to make them look alike?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriter.cs
```diff
@@ -0,0 +1,187 @@
...
+
+using System;
+using System.Buffers.Binary;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Data.SqlClient;
+
+namespace Microsoft.Data.SqlClientX.IO
+{
+    internal class TdsWriter
```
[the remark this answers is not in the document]
cheenamalhotra: It's goal is to provide encapsulation to call Stream APIs, and manage buffer better by reusing allocated memory. Stream doesn't manage buffer allocations so this wraps over it.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriter.cs
cheenamalhotra: Updated

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriter.cs
cheenamalhotra: Done

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriter.cs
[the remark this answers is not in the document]
cheenamalhotra: Added

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriter.cs
cheenamalhotra: I wish there was a tool :) Manually here.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriter.cs
cheenamalhotra: We don't expect concurrent requests on 1 stream, so allocation tampering is not of concern. However, we will review this as design progresses. At the moment, our goal is to limit memory usage by reusing buffers.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriter.cs
```diff
@@ -0,0 +1,187 @@
...
+            {
+                throw new ArgumentOutOfRangeException(nameof(v), "Double value is out of range.");
+            }
+            await EnsureBufferSpaceAsync(8, isAsync, ct).ConfigureAwait(false);
+            BinaryPrimitives.WriteInt64LittleEndian(GetWritableSpan(8), BitConverter.DoubleToInt64Bits(v));
+            _outBytesUsed += 8;
+        }
+
+        /// <summary>
+        /// Writes partial long value to out buffer
```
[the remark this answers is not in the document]
cheenamalhotra: Updated

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriter.cs
```diff
@@ -0,0 +1,187 @@
...
+        /// </summary>
+        /// <param name="v">Value to write</param>
+        /// <param name="length">Length to fill</param>
+        /// <param name="isAsync">Is Async caller method</param>
+        /// <param name="ct">Cancellation token</param>
+        public async Task WritePartialLongAsync(long v, int length, bool isAsync, CancellationToken ct)
```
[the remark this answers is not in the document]
cheenamalhotra: I would say we should prefer to use limited allocations. This method is for specific cases, when writing date, time, and related types.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriter.cs
cheenamalhotra: Main reasons are to provide encapsulated APIs, and handle buffer allocations. We're keeping byte[] usage as it is at the moment, to keep simplicity, and let Stream write continuously, Something to revisit in future, we should analyse buffer requirements before enforcing that I think.

saurabh500:
@cheenamalhotra  Can you please add the changes to csproj as well. Its much easier to read the code in VS

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriter.cs
```diff
@@ -0,0 +1,174 @@
...
+
+        #region Public APIs
+
+        /// <summary>
+        /// Writes short value to out buffer, as little-endian.
+        /// </summary>
+        /// <param name="v">Value to write</param>
+        /// <param name="isAsync">Whether caller method is executing asynchronously.</param>
+        /// <param name="ct">Cancellation token</param>
+        public async ValueTask WriteShortAsync(short v, bool isAsync, CancellationToken ct)
```
saurabh500, line 39: can you make there paramter names a bit more descriptive

e.g. 
```suggestion
        public async ValueTask WriteShortAsync(short shortValue, bool isAsync, CancellationToken ct)
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriter.cs
```diff
@@ -0,0 +1,174 @@
...
+        public async ValueTask WriteBytesAsync(byte[] data, int start, int length, bool isAsync, CancellationToken ct)
+        {
+            if (isAsync)
+            {
+                await _stream.WriteAsync(data.AsMemory(start, length), ct).ConfigureAwait(false);
+            }
+            else
+            {
+                // Throw operation canceled exception before write.
+                ct.ThrowIfCancellationRequested();
```
saurabh500, line 166: Should this check be done before async call as well ?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriter.cs
```diff
@@ -0,0 +1,174 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Data.SqlClientX.IO
+{
+    /// <summary>
+    /// This class provides helper methods for writing bytes using <see cref="TdsStream"/>
+    /// It extends <see cref="TdsBufferAlloc"/> that manages allocations of bytes buffer for better memory management.
+    /// </summary>
+    internal class TdsWriter : TdsBufferAlloc
```
saurabh500, line 17: Typically, when we inherit one class from another, we expect that casting subclass to baseclass should allow base class APIs to be able to use the sub class.

In this case, it feels like TdsWrite should be composed of TdsBufferAlloc class, instead of inheriting from it.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriter.cs
saurabh500: This applies to all the methods.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriter.cs
```diff
@@ -0,0 +1,174 @@
...
+            var len = sizeof(long);
+            var buffer = GetBuffer(len);
+            for (int i = 0; i < length; i++)
+            {
+                buffer[i] = (byte)((v >> (i * 8)) & 0xFF);
+            }
+            await WriteBytesAsync(GetBuffer(len), 0, length, isAsync, ct).ConfigureAwait(false);
+        }
+
+        /// <summary>
```
saurabh500, line 148: Can TdsParser.SerializeDecimal be added to this Writer as well ?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
```diff
@@ -61,6 +63,8 @@ internal class TdsStream : Stream, ITdsWriteStream, ITdsReadStream
         /// <inheritdoc />
         public byte ReadPacketHeaderType => _readStream.ReadPacketHeaderType;
 
+        public TdsWriter GetTdsWriter() => _writer;
```
saurabh500, line 66: This can be an autoproperty with private setter

```suggestion
        public TdsWriter TdsWriter { get; private set; }
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsStream.cs
```diff
@@ -70,6 +74,7 @@ public TdsStream(TdsWriteStream writeStream, TdsReadStream readStream) : base()
         {
             _writeStream = writeStream;
             _readStream = readStream;
+            _writer = new TdsWriter(this);
```
saurabh500, line 77: If you decide to make the writer an AutoProperty then you could simply 

```suggestion
            TdsWriter = new TdsWriter(this);
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriter.cs
saurabh500, line 17: Consider making the writer Disposable and also IDisposableAsync.

This can be then disposed when the Dispose() DisposeAsync() of `TdsStream` is called.

saurabh500:
@cheenamalhotra  can you also close and then reopen this Pull request? The CI-SqlClientX was created after this Pull Request. This is a little hack to trigger CI-SqlClientX. 
Since there are no csproj changes in the PR, I dont think that the build will have any impact on this PR.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriter.cs
```diff
@@ -0,0 +1,174 @@
...
+        /// Writes partial long value to out buffer, as little-endian. This should be ideally used for writing date, 
+        /// time and related data types. The length specified should be between 0 to 8 bytes supported by long.
+        /// </summary>
+        /// <param name="v">Value to write</param>
+        /// <param name="length">Length to fill</param>
+        /// <param name="isAsync">Whether caller method is executing asynchronously.</param>
+        /// <param name="ct">Cancellation token</param>
+        public async ValueTask WritePartialLongAsync(long v, int length, bool isAsync, CancellationToken ct)
```
saurabh500, line 134: I didnt know that this even existed 😄

saurabh500 (changes requested):

saurabh500:
Also now that we have Unit tests project ready, please consider utilizing it :)

cheenamalhotra:
> @cheenamalhotra Can you please add the changes to csproj as well. Its much easier to read the code in VS

Oops, didn't notice my solution wasn't saving - added now.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriter.cs
cheenamalhotra: Oh, I didn't bring serializers yet, do we need all serializers as part of the writer class?

saurabh500: Not all. Decimal feels like the right contender. 
Then there will be sql serializers which can leverage the tdswriter 

Don't worry about all of them. Just decimal feels like it belongs here

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriter.cs
```diff
@@ -0,0 +1,202 @@
...
+        /// </summary>
+        /// <param name="stream">Tds Stream to work with</param>
+        /// <param name="v">Value to write</param>
+        /// <param name="isAsync">Whether caller method is executing asynchronously.</param>
+        /// <param name="ct">Cancellation token</param>
+        public async ValueTask WriteShortAsync(TdsStream stream, short v, bool isAsync, CancellationToken ct)
+        {
+            var len = sizeof(short);
+            BinaryPrimitives.WriteInt16LittleEndian(_alloc.GetBuffer(len), v);
+            await WriteBytesAsync(stream, _alloc.GetBuffer(len), 0, len, isAsync, ct).ConfigureAwait(false);
```
edwardneal, line 43: This applies to all of the Write*Async methods. I would personally lift the `isAsync` parameter up a level to here. If `isAsync` is true, we can use `_alloc.GetBuffer(len)` as normal. Else, we can `stackalloc byte[len]`, write into the resultant Span and write the Span down into the stream with zero heap allocations.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriter.cs
```diff
@@ -0,0 +1,202 @@
...
+            Debug.Assert(length >= 0, "Length should not be negative");
+            Debug.Assert(length <= 8, "Length specified is longer than the size of a long");
+
+            var len = sizeof(long);
+            var buffer = _alloc.GetBuffer(len);
+            for (int i = 0; i < length; i++)
+            {
+                buffer[i] = (byte)((value >> (i * 8)) & 0xFF);
+            }
+            await WriteBytesAsync(stream, _alloc.GetBuffer(len), 0, length, isAsync, ct).ConfigureAwait(false);
```
edwardneal, line 156: Nit: I'd personally write the entire long to the buffer using `BinaryPrimitives`, then simply specify the length of the Span/Memory to be `length`. It'll get rid of the slightly awkward bit-shifting, eliminating a loop.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriter.cs
```diff
@@ -0,0 +1,202 @@
...
+        /// Writes bytes directly to _stream
+        /// </summary>
+        /// <param name="stream">Tds Stream to work with</param>
+        /// <param name="data">Bytes of data</param>
+        /// <param name="start">Offset of starting position</param>
+        /// <param name="length">Length of data</param>
+        /// <param name="isAsync">Whether caller method is executing asynchronously.</param>
+        /// <param name="ct">Cancellation token</param>
+        /// <returns></returns>
+        public async static ValueTask WriteBytesAsync(TdsStream stream, byte[] data, int start, int length, bool isAsync, CancellationToken ct)
```
edwardneal, line 169: The `isAsync` flag forces us into this signature, but the signature means that we must always send byte arrays. This rules out the use of a simple `stackalloc`'d Span in an synchronous case, forcing allocations upstream.

I think it'd be better to have a pair of `WriteSpan` and `WriteMemoryAsync` methods in this case, pushing the sync/async decision upstream.

NB: there's some work in newer .NET which allows us to use `allows ref struct` in generic type constraints, which might have allowed a single `WriteBytesAsync<T>`, where `T` was either `Span<byte>` or `Memory<byte>`. This isn't permissible in .NET Framework, but might be worth considering in the far future.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsBufferAlloc.cs
```diff
@@ -0,0 +1,76 @@
...
+            _4Buffer = null;
+            _8Buffer = null;
+            _16Buffer = null;
+        }
+
+        /// <summary>
+        /// Disposes storage buffers and returns a completed <see cref="ValueTask"/>
+        /// </summary>
+        /// <returns>Completed ValueTask</returns>
+        public ValueTask DisposeAsync()
```
edwardneal, line 70: Will this ever complete asynchronously? If not, I'd discard the `IAsyncDisposable` implementation.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriter.cs
```diff
@@ -0,0 +1,202 @@
...
+        /// </summary>
+        /// <param name="stream">Tds Stream to work with</param>
+        /// <param name="v">Value to write</param>
+        /// <param name="isAsync">Whether caller method is executing asynchronously.</param>
+        /// <param name="ct">Cancellation token</param>
+        public async ValueTask WriteIntAsync(TdsStream stream, int v, bool isAsync, CancellationToken ct)
+        {
+            var len = sizeof(int);
+            BinaryPrimitives.WriteInt32LittleEndian(_alloc.GetBuffer(len), v);
+            await WriteBytesAsync(stream, _alloc.GetBuffer(len), 0, len, isAsync, ct).ConfigureAwait(false);
```
edwardneal, line 57: Nit: in situations like this (where the async method call is the last in the method; the async method being called and the calling method both return `Task`/`ValueTask`, or the calling method simply returns the awaited value) we can simply return the ValueTask wholesale as below:

```suggestion
            return WriteBytesAsync(stream, _alloc.GetBuffer(len), 0, len, isAsync, ct);
```

I think this reduces the burden on the thread pool slightly in the async case. It also means that the method doesn't necessarily need to be declared async (not strictly a problem here.)

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsBufferAlloc.cs
```diff
@@ -0,0 +1,76 @@
...
+        /// <summary>
+        /// Get buffer for reading/writing data
+        /// Accepted sizes are: 2, 4, 8, 16
+        /// </summary>
+        /// <param name="sizeInBytes">2, 4, 8 or 16 bytes of buffer size.</param>
+        /// <returns></returns>
+        internal byte[] GetBuffer(int sizeInBytes)
+        {
+            Debug.Assert(sizeInBytes == 2 || sizeInBytes == 4 || sizeInBytes == 8 || sizeInBytes == 16);
+            return sizeInBytes switch
```
edwardneal, line 45: Depending how benchmarking goes, it might be worth encouraging this method to be inlined with `[MethodImpl(MethodImplOptions.AggressiveInlining)]`.

I'd also (very tentatively) think about returning the 16-byte array in all cases. It'll shave off 14 bytes from the heap, and we already constrain the buffer size via `AsSpan` and `AsMemory` downstream. Doing so also has knock-on effects which might mean we can discard the `TdsBufferAlloc` class and move the 16-byte scratch buffer directly onto `TdsWriter`.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriter.cs
```diff
@@ -0,0 +1,202 @@
...
+        }
+
+        /// <summary>
+        /// Writes unsigned short value to out buffer, as little-endian.
+        /// </summary>
+        /// <param name="stream">Tds Stream to work with</param>
+        /// <param name="v">Value to write</param>
+        /// <param name="isAsync">Whether caller method is executing asynchronously.</param>
+        /// <param name="ct">Cancellation token</param>
+        public ValueTask WriteUnsignedShortAsync(TdsStream stream, ushort v, bool isAsync, CancellationToken ct)
```
benrr101, line 67: This feels excessive for a method name 😅 The pattern for classes like BitConverter and BinaryPrimitives is eg Uint16 or Int32. It also removes ambiguity for the size of the value to write - but I think dotnet defines the sizes of types like `short` , `long`, etc, in bits anyways (rather than words like C++ might).

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriter.cs
benrr101, line 169: I'm unsure why this is necessary. I think it'll come down to how we look at the Writer
* If we're looking at it as the entry point for *all* TDS writing to the network, I think this class would be better implemented by rolling it into the TdsStream and adding its methods to the ITdsStream interface
* If we're looking at is as convenience methods for writing numbers to the stream, then I think it's better written by not implementing this method.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriter.cs
```diff
@@ -0,0 +1,202 @@
...
+            else
+            {
+                stream.Write(data.AsSpan(start, length));
+            }
+        }
+
+        /// <summary>
+        /// Dispose Tds Writer
+        /// </summary>
+        public void Dispose()
```
benrr101, line 187: If we don't need to dispose the alloc (see other comment), we probably don't need to dispose this class

Alternatively, if this class is meant to be the entry point to all TDS writing to the network, we should probably dispose of the TdsStream from the writer.

benrr101 (changes requested):

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsBufferAlloc.cs
edwardneal: I'm not completely sure of the best practice here either. My rationale is that `sizeInBytes` will always be a constant value for every invocation of the same method (i.e. writing a ushort will always result in the value being 2.) I essentially want `GetBuffer` to be inlined into the calling method so that the JIT can see that this parameter's value is constant and eliminate the branch. If that happens the method calls on the hot path are essentially replaced by a single field access.

The two key questions for me are whether the JIT can see this already, and whether the overhead of making a method call and going through the case statement is significant enough to warrant the compiler hint. It needs a bit of benchmarking here (when the rest of SqlClientX is working end-to-end.)

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriter.cs
saurabh500: I think the existing method names would allow an easier migration from TdsParser / TdsParserStateobject callers to TdsWriter. 
I am hoping we can retain names which looks same as older analogous names in existing client.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriter.cs
saurabh500: This class wont be the only entrypoint, some operations like setting the packet header type and few others may be leveraged from the Tds Stream. TdsStream and this writer would work in tandem

cheenamalhotra: I've modified this class to not accept TdsStream in constructor to avoid ownership confusions. With new changes, TdsStream is owning the TdsWriter, and dispose it with it. Hope that helps clarify design.

edwardneal (commented):
Just one micro-optimisation from the peanut gallery here; nothing major.

I'd personally prefer it if `SerializeDecimal` wrote to a `Span<byte>` rather than returning a byte array (which needs to be copied to another buffer in order to be used.)

The more I look at the class overall though, the more I see the case for turning it into a simple shared `TdsPrimitives` class which writes common datatypes to a Span or returns them as a byte array. Now that the packet splitting behaviour has been folded into TdsStream where it belongs, the only state for `TdsWriter` to manage is the shared byte buffers. Returning those buffers implies copying their values to the correct place in a packet, but it's more efficient to write the value to the right place in the first instance. Writing the values to a Span improves that, and means that we're effectively left with a `TdsPrimitives`-style class which happens to write to a `TdsStream`. Callers of `TdsParser` can easily do that themselves.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriter.cs
```diff
@@ -0,0 +1,249 @@
...
+        /// <summary>
+        /// Writes float value to out buffer, as little-endian.
+        /// </summary>
+        /// <param name="stream">Tds Stream to work with</param>
+        /// <param name="value">Value to write</param>
+        /// <param name="isAsync">Whether caller method is executing asynchronously.</param>
+        /// <param name="ct">Cancellation token</param>
+        public async ValueTask WriteFloatAsync(TdsStream stream, float value, bool isAsync, CancellationToken ct)
+        {
+            Debug.Assert(!float.IsInfinity(value) && !float.IsNaN(value), "Float value is out of range.");
```
VladimirReshetnikov, line 113: `!float.IsInfinity(value) && !float.IsNaN(value)` can be replaced with `float.IsFinite(value)`.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriter.cs
```diff
@@ -0,0 +1,249 @@
...
+        /// <summary>
+        /// Writes double value to out buffer, as little-endian.
+        /// </summary>
+        /// <param name="stream">Tds Stream to work with</param>
+        /// <param name="value">Value to write</param>
+        /// <param name="isAsync">Whether caller method is executing asynchronously.</param>
+        /// <param name="ct">Cancellation token</param>
+        public async ValueTask WriteDoubleAsync(TdsStream stream, double value, bool isAsync, CancellationToken ct)
+        {
+            Debug.Assert(!double.IsInfinity(value) && !double.IsNaN(value), "Double value is out of range.");
```
VladimirReshetnikov, line 129: `!double.IsInfinity(value) && !double.IsNaN(value)` can be replaced with `double.IsFinite(value)`.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriter.cs
```diff
@@ -0,0 +1,249 @@
...
+        /// </summary>
+        public void Dispose()
+        {
+            _alloc.Dispose();
+        }
+
+        /// <summary>
+        /// Disposes Tds Writer and returns a completed <see cref="ValueTask"/>
+        /// </summary>
+        /// <returns>Completed ValueTask</returns>
```
VladimirReshetnikov, line 242: Why are we sure that it returns a completed `ValueTask` if the implementation uses `await` and `ConfigureAwait`? Can't it return a task that is still in progress?

src/Microsoft.Data.SqlClient/tests/UnitTests/IO/TdsWriterTests.cs
```diff
@@ -0,0 +1,47 @@
...
+
+namespace Microsoft.Data.SqlClient.UnitTests.IO
+{
+    public class TdsWriterTests
+    {
+        [Fact]
+        public void TestSerializeDecimal_SerializesSuccessfully()
+        {
+            TdsWriter writer = new();
+            var data = new decimal(8.5);
```
VladimirReshetnikov, line 18: Why not simply write `8.5m` as a constant?

VladimirReshetnikov (changes requested):

Wraith2:
What are the requirements for SqlClientX? What platforms and levels of compatibility are needed? If you want to do fast efficient network writes you should really looking at using the pipelines api as discussed in https://github.com/dotnet/datalab/issues/13 because it will handle a lot of this complexity for you, you should only need to compose the packet header layer on top of it.

saurabh500:
@Wraith2 open a discussion item and tag me there. We can discuss what led to sqlclientx and the direction  
I was hoping to have this discussion a little later. But then there is enough interest to talk about sqlclientx now.

Wraith2:
Ok, done, https://github.com/dotnet/SqlClient/discussions/2603

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriter.cs
cheenamalhotra: This is specific to design requirements to write Date and Time and DateTime to Sql Server.

cheenamalhotra:
@Wraith2 
I believe this PR may not be the right place to understand and gain full context of changes.

@saurabh500 @benrr101 could you initiate a discussion thread explaining the new Stream design approach and chain of handlers to get a wider view into new architecture?

saurabh500:
Thanks @cheenamalhotra . I will start a discussion, but it wont happen until tomorrow. 
I will enlist the TDS protocol requirements, and how we intend to use streams. It would be good to see points of view on how Pipelines will solve all the usecases. I hope we keep the comments on this PR to what is the scope of this PR. Any fundamental design related changes, should be a discussion away from the PR

roji:
@Wraith2 one thing to note about Pipelines, is that it doesn't support sync I/O. There may be possible workarounds, but that could be a source of trouble for using it in SqlClientX.

saurabh500 (approved):

Wraith2:
@cheenamalhotra consider implementing `IBufferWriter<T>` on TdsWriter. I think it would simplify some of the handling of spans and memory going on. I'm hoping/assuming that the ToArray() is temporary.

If you expose an IBufferWriter you can pass equally between sync and async contexts because it has Span and Memory functions. You request a chunk of memory to write into and it'll give you back what it can. 

Imagine the case where you're near the end of a packet and you need to write a guid, you'd ask for 16 bytes but if there isn't space you'll get back a span from the buffer writer implmentation that only only contained 8 bytes of space. You serialize to the local buffer, copy 8 bytes to the writer provided span. advance 8 bytes on the writer (telling the writer that you've written allowing it to finalize the packet and send). Then re-request a span for the remainder of the space you need and the buffer writer can handle preparing a new buffer and reserving the space for smux and packet headers before giving you back a span/memory to write into.

PR #2598, opened 2024-06-20
Adding DbDataSource Scaffolding

mdaigle:
This PR adds the SqlDataSource and SqlInternalConnectionX classes and establishes the basic relationships between them for connection creation. Next PR will establish a basic disposal flow that returns internal connections to their originating data source.

See the image below for an overview of the class architecture. Green items are new. White items are existing, from the System.Data.Common library.
SqlClient Connection Pool (8)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlInternalConnectionX.cs
```diff
@@ -0,0 +1,62 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+#if NET7_0_OR_GREATER
```
ErikEJ, line 5: Will .NET 8 not be the minimum .NET when this is released? .NET 7 is already out of support

PR #2599, opened 2024-06-21
adding the required changes for stylecop

saurabh500:


src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
```diff
@@ -969,4 +969,9 @@
   <Import Project="$(ToolsDir)targets\ResolveContract.targets" Condition="'$(OSGroup)' == 'AnyOS'" />
   <Import Project="$(ToolsDir)targets\NotSupported.targets" Condition="'$(OSGroup)' == 'AnyOS'" />
   <Import Project="$(CommonSourceRoot)tools\targets\GenerateResourceStringsSource.targets" />
+  <ItemGroup>
+  </ItemGroup>
+  <ItemGroup>
+    <AdditionalFiles Include="Microsoft\Data\SqlClientX\.editorconfig" />
```
cheenamalhotra, line 975: Interesting! Package level styling is cool.

benrr101 (commented):
There's far too many rules to really dig into which ones are good and which ones should probably be changed. I think in my past projects we used the default rule set (which we may have defined in a nuget package somewhere) and then modified them at the project level.

Overall, I'm good with this, though I have two comments:
* Good luck getting the rest of the project in line with these :)
* I think we will likely want to tweak some of the rules as we go (eg, blocking builds b/c of comments that don't end with a period or something silly like that)

saurabh500:
@benrr101 this is still wip. 
Trying to understand some nuances of stylecop rules with the builds. 
I got all the rule sets in here and made them suggestions because our csproj is designed to treat warning as errors. 
Also you can see that I am trying to enforce this for a particular folder and not the whole code

Few things to do 
1. Make the build not treat stylecop warnings as errors. 
2. Bring in only the relevant rules based on the CR comments
3. If possible make the CI enforce stylecop on the files which changed.

cheenamalhotra:
@saurabh500 
Checking into old PRs, do you intend to work on this? If not, please feel free to close and open an issue for future milestones that someone can independently work on.

PR #2600, opened 2024-06-21
Adding DbDataSource Scaffolding

mdaigle:
This PR adds the SqlDataSource and SqlInternalConnectionX classes and establishes the basic relationships between them for connection creation and disposal. I do not go into the internals of SqlInternalConnectionX. Next PR will add a basic data source implementation.

See the image below for an overview of the class architecture. Green items are new. White items are existing, from the System.Data.Common library.
SqlClient Connection Pool (8)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlConnectionX.cs
```diff
@@ -53,25 +95,48 @@ public override string Database
...
+            set
+            {
+                switch (State)
+                {
+                    case ConnectionState.Open:
+                    case ConnectionState.Connecting:
+                    case ConnectionState.Fetching:
+                    case ConnectionState.Executing:
+                        throw new InvalidOperationException("Can't change connection string on an active connection");
+                    default:
```
VladimirReshetnikov, line 117: This `default` branch doesn't do anything, you can delete it. Alternatively, you can rewrite the whole `switch` statement as `if (State is ConnectionState.Open or ConnectionState.Connecting or ...) throw ...`.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlConnectionX.cs
```diff
@@ -53,25 +95,48 @@ public override string Database
...
+                        break;
+                }
+
+                _connectionString = _connectionString = value ?? string.Empty;
```
VladimirReshetnikov, line 121: One of ` _connectionString =` seems redundant.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlConnectionX.cs
```diff
@@ -85,13 +150,46 @@ public override Task ChangeDatabaseAsync(string databaseName, CancellationToken
         public object Clone()
             => throw new NotImplementedException();
 
+        #region close
+
         /// <inheritdoc/>
         public override void Close()
-            => throw new NotImplementedException();
+            => Close(async: false).GetAwaiter().GetResult();
```
VladimirReshetnikov, line 157: Is this equivalent to `Close(async: false).Wait()`?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlConnectionX.cs
```diff
@@ -85,13 +150,46 @@ public override Task ChangeDatabaseAsync(string databaseName, CancellationToken
...
+            }
+
+            var internalConnection = _connection;
+
+            //TODO: cancel outstanding operations on connection
+
+            //TODO: support async close operation
+            internalConnection?.Close();
+
+            internalConnection = null;
```
VladimirReshetnikov, line 187: There is no point to set a local variable to `null`.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlConnectionX.cs
```diff
@@ -125,13 +223,25 @@ public override Task<DataTable> GetSchemaAsync(string collectionName, Cancellati
         public override Task<DataTable> GetSchemaAsync(string collectionName, string?[] restrictionValues, CancellationToken cancellationToken = default)
             => throw new NotImplementedException();
 
+        #region open
+
         /// <inheritdoc/>
-        public override void Open()
-            => throw new NotImplementedException();
+        public override void Open() => _Open(false, CancellationToken.None).GetAwaiter().GetResult();
```
VladimirReshetnikov, line 230: Can we replace `.GetAwaiter().GetResult()` with `.Wait()`?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlInternalConnectionX.cs
```diff
@@ -0,0 +1,58 @@
...
+
+using System;
+using System.Data;
+using System.Data.Common;
+
+namespace Microsoft.Data.SqlClientX
+{
+    internal class SqlInternalConnectionX : DbConnection
+    {
+        public SqlInternalConnectionX() { }
```
VladimirReshetnikov, line 15: This declaration is redundant, the compiler generates a default constructor if there are no constructor declarations.

VladimirReshetnikov (changes requested):

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlConnectionX.cs
```diff
@@ -53,25 +95,48 @@ public override string Database
...
+            get => _connectionString;
+            set
+            {
+                switch (State)
+                {
+                    case ConnectionState.Open:
+                    case ConnectionState.Connecting:
+                    case ConnectionState.Fetching:
+                    case ConnectionState.Executing:
+                        throw new InvalidOperationException("Can't change connection string on an active connection");
```
edwardneal, line 116: Nit: exception message should be in a resource string. I think there's already an existing method which will throw the `InvalidOperationException`.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlConnectionX.cs
```diff
@@ -85,13 +150,46 @@ public override Task ChangeDatabaseAsync(string databaseName, CancellationToken
...
+
+            switch (State)
+            {
+                case ConnectionState.Open:
+                case ConnectionState.Executing:
+                case ConnectionState.Fetching:
+                    break;
+                case ConnectionState.Connecting:
+                    //TODO: change this to match current behavior. cancels any pending async open task
+                    throw new InvalidOperationException("Can't close connection while in Connecting state");
```
edwardneal, line 175: Nit: resource strings

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlConnectionX.cs
```diff
@@ -22,19 +25,58 @@ namespace Microsoft.Data.SqlClientX
     [DesignerCategory("")]
     internal sealed class SqlConnectionX : DbConnection, ICloneable
     {
-        //TODO: reference to internal connection
+        private SqlCredential? _credential;
+        private SqlDataSource? _dataSource;
+        private SqlInternalConnectionX? _connection;
+        private string _connectionString = string.Empty;
```
edwardneal, line 31: Very tentative, but is there any point keeping a separate variable here? `_dataSource.ConnectionString` would seem to cover this, and perhaps we could plumb the properties like `ConnectionTimeout` through to it (so that we only go through the parse process once.)

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlConnectionX.cs
```diff
@@ -125,13 +223,25 @@ public override Task<DataTable> GetSchemaAsync(string collectionName, Cancellati
...
         /// <inheritdoc/>
-        public override Task OpenAsync(CancellationToken cancellationToken)
-            => throw new NotImplementedException();
+        public override Task OpenAsync(CancellationToken cancellationToken) => _Open(true, cancellationToken);
+
+        private async Task _Open(bool async, CancellationToken cancellationToken)
+        {
+            if (_dataSource == null)
+            {
+                throw new InvalidOperationException("No data source or connection string set");
```
edwardneal, line 238: Nit: resource strings

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlConnectionX.cs
mdaigle: From what I can tell, the main difference is that this preserves any exceptions and Wait() will give an AggregateException.
https://blog.stephencleary.com/2014/12/a-tour-of-task-part-6-results.html

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlConnectionX.cs
mdaigle: changed this to a TODO. we'll need to clear the internal connection's reference to the wrapper connection.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlConnectionX.cs
mdaigle: I think it's an interesting point. We need to expose the setter so that the connection string can be changed, but that could be used immediately to resolve to a new data source and avoid storing the string locally. For now, let me add a TODO and we can try it out along with the changes that will allow SqlConnectionX to create/resolve its own data source.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlConnectionX.cs
mdaigle: Thanks @edwardneal, do you have an example you can point me to? I see some competing patterns in the existing code.

VladimirReshetnikov (approved):

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlConnectionX.cs
edwardneal: The example I'm thinking of is here:
https://github.com/dotnet/SqlClient/blob/55f48c57a9b2a200c266b8739a2a477f974ab41d/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnectionHelper.cs#L74-L88

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlInternalConnectionX.cs
```diff
@@ -0,0 +1,56 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+#if NET8_0_OR_GREATER
```
cheenamalhotra, line 5: Any reason why this class is limited to .NET 8?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlConnectionX.cs
```diff
@@ -2,6 +2,8 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+#if NET8_0_OR_GREATER
```
cheenamalhotra, line 5: Why limit the entire class?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlConnectionX.cs
```diff
@@ -125,13 +224,25 @@ public override Task<DataTable> GetSchemaAsync(string collectionName, Cancellati
...
-        public override void Open()
-            => throw new NotImplementedException();
+        public override void Open() => _Open(false, CancellationToken.None).GetAwaiter().GetResult();
 
         /// <inheritdoc/>
-        public override Task OpenAsync(CancellationToken cancellationToken)
-            => throw new NotImplementedException();
+        public override Task OpenAsync(CancellationToken cancellationToken) => _Open(true, cancellationToken);
+
+        private async Task _Open(bool async, CancellationToken cancellationToken)
```
cheenamalhotra, line 235: You can call it OpenInternal and CloseInternal for the internal close method instead, to keep naming conventions use PascalCase.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlDataSource.cs
```diff
@@ -0,0 +1,36 @@
...
+#if NET8_0_OR_GREATER 
+
+using Microsoft.Data.SqlClient;
+using System.Data.Common;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Data.SqlClientX
+{
+    internal abstract class SqlDataSource : DbDataSource
```
cheenamalhotra, line 14: Should this be a public class?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlConnectionX.cs
mdaigle: Many parts of the class will rely on SqlDataSource (which extends DbDataSource), so the class will not be functional on .NET lower than 8.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlInternalConnectionX.cs
mdaigle: I'm going to remove this class' inheritance of DbConnection and rename it SqlConnector. This will bring it in line with the Npgsql naming convention. I agree that this class does not need to be limited to .NET 8+.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlDataSource.cs
mdaigle: As discussed, we'll keep this internal for now until we're ready to support a broader public API surface.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlConnectionX.cs
```diff
@@ -85,13 +153,50 @@ public override Task ChangeDatabaseAsync(string databaseName, CancellationToken
...
+
         /// <inheritdoc/>
         public override void Close()
-            => throw new NotImplementedException();
+            => Close(async: false).GetAwaiter().GetResult();
 
         /// <inheritdoc/>
         public override Task CloseAsync()
-            => throw new NotImplementedException();
+            => Close(async: true);
```
cheenamalhotra, line 164: ConfigureAwait(false)

cheenamalhotra (approved):

src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
```diff
@@ -920,6 +920,8 @@
     </Compile>
     <Compile Include="Microsoft\Data\SqlClientX\Handlers\ConnectionHandlerContext.cs" />
     <Compile Include="Microsoft\Data\SqlClientX\Handlers\DataSourceParsingHandler.cs" />
+    <Compile Include="Microsoft\Data\SqlClientX\SqlDataSource.cs" />
+    <Compile Include="Microsoft\Data\SqlClientX\SqlInternalConnectionX.cs" />
```
cheenamalhotra, line 924: Rename file here

mdaigle: Somehow, I missed committing that change

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlConnectionX.cs
mdaigle: I can only add this on lines that directly await a Task with the await keyword. I fixed one other spot where it was missing though.

PR #2602, opened 2024-06-24
Add UnpooledDataSource

mdaigle:
Adds a simple data source that always creates an internal connection from scratch.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/UnpooledDataSource.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/UnpooledDataSource.cs
@@ -1,0 +1,28 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+#if NET8_0_OR_GREATER
+
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Data.SqlClient;
+
+namespace Microsoft.Data.SqlClientX
+{
+    internal sealed class NonPooledDataSource : SqlDataSource
+    {
+        internal NonPooledDataSource(string connectionString, SqlCredential credential) : base(connectionString, credential)
+        {
+        }
+
+        internal override async ValueTask<SqlConnector> GetInternalConnection(bool async, CancellationToken cancellationToken)
+        {
+            var connection = new SqlConnector();
+            await connection.Open(async, cancellationToken).ConfigureAwait(false);
+            return connection;
+        }
+    }
+}
+
+#endif
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/UnpooledDataSource.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/UnpooledDataSource.cs
@@ -4,6 +4,7 @@
 
 #if NET8_0_OR_GREATER
 
+using System;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Data.SqlClient;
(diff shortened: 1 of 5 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/UnpooledDataSource.cs
```diff
@@ -0,0 +1,42 @@
...
+        /// <summary>
+        /// Creates and opens a new SqlConnector.
+        /// </summary>
+        /// <param name="async">Whether this method should be run asynchronously.</param>
+        /// <param name="cancellationToken">Cancels an outstanding asynchronous operation.</param>
+        /// <returns></returns>
+        internal override async ValueTask<SqlConnector> GetInternalConnection(bool async, CancellationToken cancellationToken)
+        {
+            var connection = new SqlConnector();
+            await connection.Open(async, cancellationToken).ConfigureAwait(false);
```
edwardneal, line 36: In the synchronous case, the cancellation token needs to be checked after calling `Open`. If it's cancelled then you'll probably want to close the connection and throw an exception.

The same pattern applies in quite a few places - if the operation is cancelled before completion, we need to make sure the nested resources are disposed before throwing the cancellation exception. @benrr101 I've not mentioned this part of the cancellation checks before, it also needs to be considered when we plumb it in.

mdaigle: Thanks, @edwardneal. I'm open to ideas if there's a better way to express this in the code, but if you trace the call to this method up to the "public" API surface (in SqlConnectionX), it will always pass down a `CancellationToken.None` for the sync case. My intention is to ignore cancellation tokens for the sync path. 

The public surface looks like this:
```c#
public override void Open() => Open(false, CancellationToken.None).GetAwaiter().GetResult();
```

For the async case, I'd consider this past the useful point of cancellation checking as all the work has already been done.

There is still work to be done to respect connect timeouts and I'll definitely add that in to this PR.

edwardneal: I think part of this depends on how we want to handle those connection timeouts. Some of that work is yet to be done, but it seems to me that the simplest way would be to have a datastructure which converts a user-specified connection timeout into `CancellationToken` and which provides the number of seconds remaining until the timeout fully elapses (to allow the socket's connection timeout to be set appropriately in the synchronous case.) At that point, async cancellation support is just a matter of combining the two `CancellationToken` objects and passing it down the chain. In this scenario, `CancellationToken.None` would only be passed in the synchronous case with no connection timeouts.

mdaigle: I agree. I believe the best approach will be to pass the timeout down to the network layer where this cancellation token orchestration can be owned and managed. I pushed changes that reflect this.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/UnpooledDataSource.cs
```diff
@@ -0,0 +1,42 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+#if NET8_0_OR_GREATER
+
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Data.SqlClient;
+
+namespace Microsoft.Data.SqlClientX
```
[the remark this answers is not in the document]
mdaigle: I tried the upgrade to 10 and didn't see any obvious compiler warnings or anything. So it may be straightforward? 🤞 
I'll do the change in a separate PR to keep it isolated.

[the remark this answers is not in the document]
roji: Yeah, some (most?) newer C# version features don't require targetting a new .NET TFM; for example, this feature (file-scoped namespace declarations) is purely syntactic, and can be used even if targetting netstandard2.0; so there's no reason to avoid using newer C# versions even when targetting old TFMs.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/UnpooledDataSource.cs
edwardneal: I agree, with the caveat that the connection timeout is end-to-end, so if there's some non-network activity which takes a while (such as a wait on the connection pool's availability) then it needs to encompass that too. I'll experiment to see if there's a way to make the combination of timeout + `CancellationToken` generic enough to use for the command timeout too.

mdaigle: I'll add a TODO and we can rework the API for specifying timeout as we get into the new connection pool implementation.

VladimirReshetnikov (approved):

PR #2607, opened 2024-06-25
Transport Creation Handler Improvements

benrr101:
**Description**: Here's some stuff I've been working on that has hit a reasonable stopping point.
* Splitting the "Handler" codebase into two different handler styles for different use cases
  *  `ContextHandler` base class for usage in "Chain of Handlers" pattern. Each handler is expected to do something and modify the context before handing it off to the next. Success/failure is determined by properties in the context at the end of the chain.
  * `ReturningHandler` base class for usage in "Chain of Responsibility" pattern. Only one handler is expected to fully process the input parameters and return something. 
* Reimplement the transport creation handler to have its own chain of responsibilities. One handler for each type of protocol to try. This implements trying shared memory, then tcp, then named pipes, then failure. Although the other two handlers will either pass or throw, they're mostly placeholders for now.
* Improvement to cancelling TCP via cancellation token.
* Unit tests for
  * ReturningHandler
  * IpVersionComparer
  * NamedPipeTransportCreationHandler
  * SharedMemoryTransportCreationHandler

**Testing**: See above


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/NamedPipeTransportCreationHandler.cs
```diff
@@ -0,0 +1,37 @@
...
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Data.SqlClient.SNI;
+
+namespace Microsoft.Data.SqlClientX.Handlers.TransportCreation
+{
+    internal sealed class NamedPipeTransportCreationHandler : ReturningHandler<ConnectionHandlerContext, Stream>
```
saurabh500, line 14: Comment for the class

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
```diff
@@ -3,48 +3,30 @@
...
-        #endif
-
-        /// <inheritdoc />
-        public IHandler<ConnectionHandlerContext> NextHandler { get; set; }
+        public TransportCreationHandler()
+        {
+            // Construct the chain of responsibility for handling the connections
+            _streamCreationChain = new SharedMemoryTransportCreationHandler();
+            _streamCreationChain.NextHandler = new TcpTransportCreationHandler();
+            _streamCreationChain.NextHandler.NextHandler = new SharedMemoryTransportCreationHandler();
```
saurabh500, line 21: The last one needs to be the NamedPipe creation handler

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/NamedPipeTransportCreationHandler.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        /// <inheritdoc />
+        public override async ValueTask<Stream> Handle(ConnectionHandlerContext parameters, bool isAsync, CancellationToken ct)
+        {
+            Debug.Assert(parameters is not null, "Parameters must not be null");
+            Debug.Assert(parameters.DataSource is not null, "Data source must not be null");
+
+            // This handler can only handle connection attempts if named pipes are included
+            if (parameters.DataSource.ResolvedProtocol is not (DataSource.Protocol.NP or DataSource.Protocol.None))
```
saurabh500, line 23: You could invert this condition, add a TODO for NP handling, else send the information to the HandleNext ? That way you could also get rid of the `NotImplementedException` and the TODO becomes the placeholder

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/NamedPipeTransportCreationHandler.cs
```diff
@@ -0,0 +1,37 @@
...
+            {
+                return await HandleNext(parameters, isAsync, ct);
+            }
+            
+            if (parameters.DataSource.ResolvedProtocol is DataSource.Protocol.None)
+            {
+                // @TODO: Support named pipe connections
+                return await HandleNext(parameters, isAsync, ct);
```
saurabh500, line 31: Help me understand the `ReturningHandler` The NP handler ends up as the last handler in the returning handler chain. 
In that case, there would be no `HandleNext` call right? Either it connects or throws ?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
```diff
@@ -3,48 +3,30 @@
...
-                // @TODO: Build CoR for handling the different protocols in order
-                if (context.DataSource.ResolvedProtocol is DataSource.Protocol.TCP)
-                {
-                    context.ConnectionStream = await HandleTcpRequest(context, isAsync, ct).ConfigureAwait(false);
-                }
-                else
-                {
-                    throw new NotImplementedException();
-                }
+                context.ConnectionStream = await _streamCreationChain.Handle(context, isAsync, ct);
```
saurabh500, line 29: The usual chain in connectivity is SM > TCP > NP in the absence of a protocol.

If only NP is available on the server, then SM cannot connect, neither can TCP. So what will happen ? These two will throw? and try their next handler? 
What if there is no server? Then SM/TCP/NP all will fail, but then what will NP handler do? It will swallow the exception and pass the control to HandleNext?

saurabh500 (changes requested):
Need some more clarification around the error handling of the ReturningHandlers.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/NamedPipeTransportCreationHandler.cs
saurabh500: Well, there is no `if` there is definitely the question of `when` 😄 But yes makes sense to punt this comment for when NP is implemented.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/NamedPipeTransportCreationHandler.cs
benrr101: The idea was effectively for the handlers to not know where they are in the chain or if any other handlers are in the chain. So, if it can't handle it, it passes it onto the next handler - the `HandleNext` method throws if it's at the end of the chain.

I think I made a slight mistake in the NamedPipe implementation by explicitly throwing `NoSuitableHandlerFoundException` when in that case it should be throwing `NotImplementedException`. In fixing the previous comment (about inverting the condition), I also fixed this. Hopefully it makes more sense now.

mdaigle, line 31:
```suggestion
                return await HandleNext(parameters, isAsync, ct).ConfigureAwait(false);
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TcpTransportCreationHandler.cs
```diff
@@ -0,0 +1,201 @@
...
+            // Enable keep-alive
+            socket.SetSocketOption(SocketOptionLevel.Socket, SocketOptionName.KeepAlive, true);
+            socket.SetSocketOption(SocketOptionLevel.Tcp, SocketOptionName.TcpKeepAliveInterval, KeepAliveIntervalSeconds);
+            socket.SetSocketOption(SocketOptionLevel.Tcp, SocketOptionName.TcpKeepAliveTime, KeepAliveTimeSeconds);
+
+            try
+            {
+                if (isAsync)
+                {
+                    #if NET6_0_OR_GREATER
```
mdaigle, line 123: Consider removing pathways for anything under .NET 8. The top level entry point (SqlConnectionX) will only be available in .NET >=8.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
benrr101: For question 1) Right, SM will catch exception and pass onto the TCP handler. TCP handler catches exception and passes onto the NP handler. NP handler would succeed and return the stream.

For question 2) Ok, that's a good point. It's currently implemented to try SM, fail and pass to next, try TCP, fail and pass to next, try NP, fail, then pass to next, but there isn't any next so it throws no suitable handler found. Since we have an enum here, we could presume that the getting that exception means we couldn't connect at all, but I'm not sure. What kind of exception would you want in that situation?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TcpTransportCreationHandler.cs
benrr101: In some cases this is feasible (eg, this case) - since we won't be building this project for net462, so removing the netfx option works. But in other cases (eg, above where I have a check for net8 vs net6), it's less feasible - since we build the project for net6 and net8, I still need *something* that compiles, even if it's never being used.

Alternatively we can make all these files be `#if NET8_0_OR_GREATER`...

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/NamedPipeTransportCreationHandler.cs
benrr101: With incoming changes, no longer applicable.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TcpTransportCreationHandler.cs
```diff
@@ -0,0 +1,198 @@
...
+
+            // Attempt to connect to one of the matching IP addresses
+            // @TODO: Handle opening in parallel
+            Socket socket = null;
+            var socketOpenExceptions = new List<Exception>();
+
+            int portToUse = parameters.DataSource.ResolvedPort < 0
+                ? parameters.DataSource.Port
+                : parameters.DataSource.ResolvedPort;
+            var ipEndpoint = new IPEndPoint(IPAddress.None, portToUse); // Allocate once
```
VladimirReshetnikov, line 76: if you want to be consistent with the capitalization in the class name and the parameter name on line 111, you can rename the local variable to `ipEndPoint`

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TcpTransportCreationHandler.cs
```diff
@@ -0,0 +1,198 @@
...
+        private void OpenSocketSync(Socket socket, IPEndPoint ipEndPoint, CancellationToken ct)
+        {
+            ct.ThrowIfCancellationRequested();
+
+            try
+            {
+                // Note: Although it seems logical to cancel if cancellation happens, we don't
+                //   need to do that here. Since the socket is set to be non-blocking, once we
+                //   call connect we'll throw and move onto polling where we continuously check
+                //   the cancellation token.
```
VladimirReshetnikov, line 155: I don't understand how it works. What does it mean "to cancel if cancellation happens"? What are possible scenarios when we arrive at the next line?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
```diff
@@ -3,48 +3,34 @@
...
```
VladimirReshetnikov, line 33: Do we want `.ConfigureAwait(false)` here?

src/Microsoft.Data.SqlClient/netcore/src/Properties/AssemblyInfo.cs
```diff
@@ -1,6 +1,4 @@
-﻿using System;
-using System.Runtime.CompilerServices;
-using System.Reflection;
-using System.Security;
+﻿using System.Runtime.CompilerServices;
 
 [assembly: InternalsVisibleTo("Microsoft.Data.SqlClient.NetCore.UnitTests")]
+[assembly: InternalsVisibleTo("DynamicProxyGenAssembly2")]
```
VladimirReshetnikov, line 4: What is that assembly?

saurabh500: Needed to expose internals to the dependencies of Moq.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
benrr101: Ok, I have an improved implementation (probably overengineered to heck) it will be pushed shortly.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TcpTransportCreationHandler.cs
```diff
@@ -0,0 +1,198 @@
...
+    /// </summary>
+    internal sealed class TcpTransportCreationHandler : ReturningHandler<ConnectionHandlerContext, Stream>
+    {
+        private const int KeepAliveIntervalSeconds = 1;
+        private const int KeepAliveTimeSeconds = 30;
+
+        #if NET8_0_OR_GREATER
+        private static readonly TimeSpan DefaultPollTimeout = TimeSpan.FromSeconds(30);
+        #else
+        private const int DefaultPollTimeout = 30 * 100000; // 30 seconds as microseconds
```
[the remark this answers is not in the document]
benrr101: I'm american, I can't metric system correctly 😝

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TcpTransportCreationHandler.cs
```diff
@@ -0,0 +1,198 @@
...
+                    break;
+
+                case SqlConnectionIPAddressPreference.IPv6First:
+                    Array.Sort(ipAddresses, IpAddressVersionComparer.InstanceV6);
+                    break;
+
+                case SqlConnectionIPAddressPreference.UsePlatformDefault:
+                default:
+                    // Not sorting necessary
+                    break;
```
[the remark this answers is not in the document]
benrr101: This is true, though I do get a hint from my ide that I'm not covering all possible cases. I've seen this one a lot, so I think I should probably just turn it off since it's pretty benign.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TcpTransportCreationHandler.cs
benrr101: Good catch. I'm going with the precedent dotnet set with `IPEndPoint` (but fixing the acronym capitalization).

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TcpTransportCreationHandler.cs
benrr101: Basically, if the cancellation token fires (ie, cancellation is requested), we would cancel connecting by disposing the socket. But since the socket is initialized as non-blocking, connect will immediately throw to indicate it would block (though connection still continues in the background). So there's no need to dispose the socket on cancellation until we move onto polling for a successful connection.

I've updated the comment to be a bit more clear about what's going on.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TransportCreationHandler.cs
benrr101: Yep, makes sense to me

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TcpTransportCreationHandler.cs
```diff
@@ -0,0 +1,193 @@
...
+    /// </summary>
+    internal sealed class TcpTransportCreationHandler : IReturningHandler<ConnectionHandlerContext, Stream>
+    {
+        private const int KeepAliveIntervalSeconds = 1;
+        private const int KeepAliveTimeSeconds = 30;
+
+        #if NET8_0_OR_GREATER
+        private static readonly TimeSpan DefaultPollTimeout = TimeSpan.FromSeconds(30);
+        #else
+        private const int DefaultPollTimeout = 30 * 1000000; // 30 seconds as microseconds
```
saurabh500, line 28:
```suggestion
        private const int DefaultPollTimeout = 30 * 1_000_000; // 30 seconds as microseconds
```

VladimirReshetnikov (approved):

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TcpTransportCreationHandler.cs
```diff
@@ -0,0 +1,193 @@
...
+            // If no socket succeeded, throw
+            if (socket is null)
+            {
+                // If there are any socket exceptions in the collected exceptions, throw the first
+                // one. If there are not, collect all exceptions and throw them as an aggregate.
+                foreach (Exception exception in socketOpenExceptions)
+                {
+                    if (exception is SocketException)
+                    {
+                        throw exception;
```
saurabh500, line 95: We spoke offline about aligning with the way exceptions are surfaced from here to orchestrator. Please take a look at the error codes used by SNITcpHandle and how errors are created for them.
There are helpers now in SqlClientX added with the Prelogin PR which can help to go from a TCP provider to the SqlError and then that can be added to the error collection in the Context, which can then be used to create a SqlException.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/NamedPipeTransportCreationHandler.cs
```diff
@@ -0,0 +1,35 @@
...
+
+namespace Microsoft.Data.SqlClientX.Handlers.TransportCreation
+{
+    /// <summary>
+    /// Handler for connecting via named pipe.
+    /// </summary>
+    internal sealed class NamedPipeTransportCreationHandler : IReturningHandler<ConnectionHandlerContext, Stream>
+    {
+        /// <inheritdoc />
+        public ValueTask<Stream> Handle(ConnectionHandlerContext parameters, bool isAsync, CancellationToken ct)
```
saurabh500, line 20: Nit: Name can be made in sync with what the class name
```suggestion
        public ValueTask<Stream> Handle(ConnectionHandlerContext context, bool isAsync, CancellationToken ct)
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TcpTransportCreationHandler.cs
saurabh500: You will need to revisit all the exceptions being thrown out of here to align with what the current driver is throwing. The pattern is similar for all of them .

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/TcpTransportCreationHandler.cs
```diff
@@ -0,0 +1,193 @@
...
+                    }
+                }
+
+                throw new AggregateException(socketOpenExceptions);
+            }
+
+            // Create the stream for the socket
+            return new NetworkStream(socket);
```
saurabh500, line 103: Does disposing NetworkStream dispose socket also? If not, then can you add it to the context, so that it can be disposed, when the connection is closed ?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/ReturningHandlerChainExceptionBehavior.cs
```diff
@@ -0,0 +1,39 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.Data.SqlClientX.Handlers
+{
+    /// <summary>
+    /// Options for how exceptions should be handled in the <see cref="ReturningHandlerChain{TParameters,TOutput}"/>.
+    /// </summary>
+    internal enum ReturningHandlerChainExceptionBehavior
```
saurabh500, line 10: Do you have a few usecases in mind for SqlClient, where the ReturningHandler concept can be leveraged?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIProxy.cs
```diff
@@ -484,6 +484,16 @@ internal enum Protocol { TCP, NP, None, Admin };
 
         internal bool IsSsrpRequired { get; private set; } = false;
 
+        /// <summary>
+        /// Constructs and initializes a data source <b>for testing purposes</b> that
+        /// is of the provided protocol.
+        /// </summary>
+        /// <param name="protocol">Protocol to set the instance to use.</param>
+        internal DataSource(Protocol protocol)
```
saurabh500, line 492: Was this changed purely for testing? 
If yes, it might benefit to keep this change out of the clientX change and leverage the appropriate connection string. like tcp:server,port or np:localhost  to get the various protocols out. 
I understand that this is away from the principles of unit testing, but something to consider to avoid changing the rest of the driver, to allow an easier merge to `main`

saurabh500 (changes requested):

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/TransportCreation/NamedPipeTransportCreationHandler.cs
benrr101: I'll avoid taking this one since I think it's better for argument names to come from the arguments of the interface it is implementing.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/ReturningHandlerChainExceptionBehavior.cs
benrr101: At least a couple -
* Feature extension parsing - feed a feature extension token in as the parameters to a chain of responsibility with a separate handler for each supported feature extension, returning an appropriate feature extension object (or class that inherits from feature extension)
* Row reading - read a token off the stream, pass it into a chain of responsibility with a separate handler for each object type, the one that can handle the object will return the appropriately parsed object.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIProxy.cs
benrr101: Yes, this change was made for testing.
I think for data source we're going to need to make a separate class for SqlClientX development. We will need a shared memory enum value for Protocol, but making a change like that in main would be very bad. The best way to do that would be to duplicate the data source class for SqlClientX usage. This also allows for moving parsing logic out of the data source class and into the data source parsing handler, allowing for data source class to be a mostly data-only class, allowing for better testing of it.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Connection/TransportCreationSubHandlers/IpAddressVersionComparer.cs
```diff
@@ -7,33 +7,33 @@
 using System.Net;
 using System.Net.Sockets;
 
-namespace Microsoft.Data.SqlClientX.Handlers.TransportCreation
+namespace Microsoft.Data.SqlClientX.Handlers.Connection.TransportCreationSubHandlers
```
cheenamalhotra, line 10: Handlers need not come twice in namespace, would recommend naming as: "Microsoft.Data.SqlClientX.Handlers.Connection.TransportCreation"

And that it also contains the TransportCreationHandler.cs class.

Similar should be done for PreLoginHandler.cs, can be done separately.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/ConnectionHandlerContext.cs
```diff
@@ -14,7 +14,7 @@ namespace Microsoft.Data.SqlClientX.Handlers
 {    /// <summary>
      /// Class that contains all context data needed for various handlers.
      /// </summary>
-    internal class ConnectionHandlerContext : HandlerRequest
+    internal class ConnectionHandlerContext
```
cheenamalhotra, line 17: This type should be in Microsoft.Data.SqlClientX.Handlers.Connection namespace.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/DataSourceParsingHandler.cs
```diff
@@ -12,13 +12,10 @@ namespace Microsoft.Data.SqlClientX.Handlers
     /// <summary>
     /// Handler to populate data source information required for transport handler
     /// </summary>
-    internal class DataSourceParsingHandler : IHandler<ConnectionHandlerContext>
```
cheenamalhotra, line 15: This should be in Microsoft.Data.SqlClientX.Handlers.Connection namespace, considering data source parsing is only applicable to Connection flow.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Connection/TransportCreation/TcpTransportCreationHandler.cs
```diff
@@ -13,93 +12,51 @@
...
-        {
-            throw new NotImplementedException();
-        }
-
-        private async ValueTask<Stream> HandleTcpRequest(ConnectionHandlerContext context, bool isAsync, CancellationToken ct)
-        {
             ct.ThrowIfCancellationRequested();
 
             // DNS lookup
             IPAddress[] ipAddresses = isAsync
```
cheenamalhotra, line 43: Is this the only one time Dns lookup will happen for connection attempt(s)? If yes, no change is needed.
If no, you may consider this Dns Lookup result be stored somewhere. 

One of the important things to consider is that Dns calls often timeout, i.e. they take 5 secs+ to return, I would recommend this result be cached in context by Data Source parser for the connection event if this will lead to timeout multiple times for the same connection open request.

cheenamalhotra (approved):
Overall looks good to me, just one comment - something to think and can be addressed in a separate PR :)

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/ReturningHandlerChain.cs
```diff
@@ -0,0 +1,164 @@
...
+                TOutput result = await handler.Handle(parameters, isAsync, ct);
+                if (result is not null)
+                {
+                    return result;
+                }
+            }
+
+            // If we make it to here, none of the handlers successfully handled the message. Throw
+            // that no handlers found.
+            throw new NoSuitableHandlerFoundException();
```
cheenamalhotra, line 83: Note that this should not lead to public exception, should be caught and handled where expected.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Connection/TransportCreation/TcpTransportCreationHandler.cs
benrr101: Yeah, this is something that will likely be addressed in a later PR.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/ReturningHandlerChain.cs
benrr101: Yes, this is something that will likely be addressed in a later PR.

PR #2616, opened 2024-06-27
Add HandlerOrchestrator and update ConnectionHandlerContext

samsharma2700:
- Added the new HandlerOrchestrator, capable of creating chain of handlers, add ConnectionHandlerContext of various Handlers into the routingHistory, maintain total number of reroutes, initiate the chain. 
- Updated the ConnectionHandlerContext so that it can be cloned in case of reroutes.


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/ConnectionHandlerContext.cs
```diff
@@ -31,5 +31,40 @@ internal class ConnectionHandlerContext : HandlerRequest
...
+        {
+            ConnectionHandlerContext clonedContext = new ConnectionHandlerContext();
+
+            if (this.ConnectionStream != null)
+            {
+                MemoryStream clonedStream = new MemoryStream();
+                this.ConnectionStream.CopyTo(clonedStream);
+                clonedStream.Position = 0;
+                clonedContext.ConnectionStream = clonedStream;
+            }
```
edwardneal, line 48: Is this correct? I expect `ConnectionStream` would ultimately be based on a `NetworkStream`. Particularly for read-only routing, I would have thought we'd only want to clone the connection details and let the handlers re-establish the connection.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/HandlerOrchestrator.cs
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        /// <summary>
+        /// Intiates the pre-defined chain of handlers
+        /// </summary>
+        /// <param name="isAsync"></param>
+        /// <param name="ct"></param>
+        /// <returns></returns>
+        /// <exception cref="HandlerNotFoundException"></exception>
+        public async Task ProcessRequestAsync(bool isAsync, CancellationToken ct)
```
edwardneal, line 40:
```suggestion
        public async ValueTask ProcessRequestAsync(bool isAsync, CancellationToken ct) 
```

Nit: switched to ValueTask to avoid a heap allocation when `isAsync` is false.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/ConnectionHandlerContext.cs
```diff
@@ -31,5 +31,40 @@ internal class ConnectionHandlerContext : HandlerRequest
         /// Class used by orchestrator while chaining handlers.
         /// </summary>
         public Exception Error { get; set; }
+
+        /// <summary>
+        /// Method needed to clone ConnectionHandlerContext as part of history along the chain.
```
VladimirReshetnikov, line 36: Remove `Method needed to` and write simply to `Clone <see cref="ConnectionHandlerContext"/> as part of history along the chain.`

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/ConnectionHandlerContext.cs
```diff
@@ -31,5 +31,40 @@ internal class ConnectionHandlerContext : HandlerRequest
...
+        /// </summary>
+        public object Clone() 
+        {
+            ConnectionHandlerContext clonedContext = new ConnectionHandlerContext();
+
+            if (this.ConnectionStream != null)
+            {
+                MemoryStream clonedStream = new MemoryStream();
```
VladimirReshetnikov, line 44: you can use `new()`

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/ConnectionHandlerContext.cs
```diff
@@ -31,5 +31,40 @@ internal class ConnectionHandlerContext : HandlerRequest
...
+                this.ConnectionStream.CopyTo(clonedStream);
```
VladimirReshetnikov, line 45: This will copy content of `this.ConnectionStream` starting from its current `Position` and move its position to its end. Is it what you intended?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/ConnectionHandlerContext.cs
```diff
@@ -31,5 +31,40 @@ internal class ConnectionHandlerContext : HandlerRequest
...
+
+            if (this.ConnectionString != null)
+            {
+                clonedContext.ConnectionString = new SqlConnectionString(this.ConnectionString.ToString());
+            }
+
+            if (this.DataSource != null)
+            {
+                //TODO: Move DataSource class into SqlClientX and implement clone
+                clonedContext.DataSource = (DataSource)this.DataSource;
```
VladimirReshetnikov, line 58: Why do you need the explicit cast?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/HandlerOrchestrator.cs
```diff
@@ -0,0 +1,95 @@
...
+        /// Class responsible to create the chain of handlers, initiate chain, retain handler-context, maintain 
+        /// routing count and history of all handler contexts for all handlers in the chain
+        /// </summary>
+        private  IHandler<HandlerRequest> _initialHandler;
+        private ConnectionHandlerContext _initialConnectionHandlerContext;
+        private InternalConnectionContext _internalConnectionContext;
+        private int _routingCount;
+        private List<ConnectionHandlerContext> _routingHistory;
+
+        public HandlerOrchestrator(IHandler<HandlerRequest> initialHandler, ConnectionHandlerContext initialConnectionHandlerContext, InternalConnectionContext internalConnectionContext, List<ConnectionHandlerContext> history)
```
edwardneal, line 24: Are the `initialHandler` and `history` parameters needed?
If the `history` parameter is needed, `_routingCount` might need to be set to `_routingHistory.Count`.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/ConnectionHandlerContext.cs
```diff
@@ -31,5 +31,40 @@ internal class ConnectionHandlerContext : HandlerRequest
...
+            }
+
+            if (this.Error != null)
+            {
+                clonedContext.Error = new Exception(this.Error.Message, this.Error.InnerException);
```
VladimirReshetnikov, line 63: Why do you need to create a new exception object? Do you care about the exception type? Usually, exceptions are immutable objects. And you do not clone the `InnerException`, it will be shared between the two outer exceptions anyways.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/HandlerOrchestrator.cs
```diff
@@ -0,0 +1,95 @@
...
+        /// <summary>
+        /// Intiates the pre-defined chain of handlers
+        /// </summary>
+        /// <param name="isAsync"></param>
+        /// <param name="ct"></param>
+        /// <returns></returns>
+        /// <exception cref="HandlerNotFoundException"></exception>
+        public async Task ProcessRequestAsync(bool isAsync, CancellationToken ct) 
+        {
+            CreateHandlerChain();
```
edwardneal, line 42: Do we need to create the handler chain on every request? These are currently stateless, so I'd think that we could turn this into a singleton and avoid some allocations.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/HandlerOrchestrator.cs
```diff
@@ -0,0 +1,95 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Data.SqlClientX.Handlers.TransportCreation;
+
+namespace Microsoft.Data.SqlClientX.Handlers
+{
+    internal class HandlerOrchestrator
+    {
+        /// <summary>
+        /// Class responsible to create the chain of handlers, initiate chain, retain handler-context, maintain
```
VladimirReshetnikov, line 15: "responsible to create" doesn't sound right, maybe "responsible for creation of"?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/HandlerOrchestrator.cs
```diff
@@ -0,0 +1,95 @@
...
+
+namespace Microsoft.Data.SqlClientX.Handlers
+{
+    internal class HandlerOrchestrator
+    {
+        /// <summary>
+        /// Class responsible to create the chain of handlers, initiate chain, retain handler-context, maintain 
+        /// routing count and history of all handler contexts for all handlers in the chain
+        /// </summary>
+        private  IHandler<HandlerRequest> _initialHandler;
```
VladimirReshetnikov, line 18: remove extra space

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/HandlerOrchestrator.cs
```diff
@@ -0,0 +1,95 @@
...
+        /// Intiates the pre-defined chain of handlers
+        /// </summary>
+        /// <param name="isAsync"></param>
+        /// <param name="ct"></param>
+        /// <returns></returns>
+        /// <exception cref="HandlerNotFoundException"></exception>
+        public async Task ProcessRequestAsync(bool isAsync, CancellationToken ct) 
+        {
+            CreateHandlerChain();
+            if (_initialHandler != null)
```
VladimirReshetnikov, line 43: `is not null`, see my comment above

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/HandlerOrchestrator.cs
```diff
@@ -0,0 +1,95 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Creates the chain of handlers in the order definition
+        /// </summary>
+        private void CreateHandlerChain()
+        {
+            // List of handlers to be added to the chain, ex : DSH -> TCH -> Prelogin -> Login -> null
+            List<IHandler<ConnectionHandlerContext>> handlers = new List<IHandler<ConnectionHandlerContext>>
```
VladimirReshetnikov, line 60: `new()`

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/HandlerOrchestrator.cs
```diff
@@ -0,0 +1,95 @@
...
+
+            for (int i = 0; i < handlers.Count - 1; i++)
+            {
+                handlers[i].NextHandler = handlers[i + 1];
+            }
+
+            // Set the first handler in the chain
+            if (handlers.Count > 0)
+            {
+                _initialHandler = (IHandler<HandlerRequest>)handlers[0];
```
VladimirReshetnikov, line 75: why do you need explicit cast?

VladimirReshetnikov (dismissed):

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/HandlerOrchestrator.cs
```diff
@@ -0,0 +1,95 @@
...
+        /// Adds the context of previous handler into the routingHistory object in case
+        /// of a reroute, exception etc.
+        /// </summary>
+        /// <param name="context"></param>
+        /// <exception cref="ArgumentNullException"></exception>
+        private void AddToHistory(ConnectionHandlerContext context)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException(nameof(context), "Context cannot be null");
```
edwardneal, line 89: Nit: resource string

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/ConnectionHandlerContext.cs
```diff
@@ -10,7 +10,7 @@ namespace Microsoft.Data.SqlClientX.Handlers
 {    /// <summary>
      /// Class that contains all context data needed for various handlers.
      /// </summary>
-    internal class ConnectionHandlerContext : HandlerRequest
+    internal class ConnectionHandlerContext : HandlerRequest, ICloneable
```
imasud00, line 13: The use of a default constructor in the ConnectionHandlerContext class allows the creation of an instance where all properties are null. This implies that an object with all its properties set to null is considered valid, which is likely not the case. Allowing such states can lead to runtime errors and logical inconsistencies within the application.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/HandlerOrchestrator.cs
saurabh500: Well this seems more like a Debug assert than an exception. If this exception is every thrown, then the driver imlpementation is wrong

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/HandlerOrchestrator.cs
```diff
@@ -0,0 +1,95 @@
...
+// See the LICENSE file in the project root for more information.
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Data.SqlClientX.Handlers.TransportCreation;
+
+namespace Microsoft.Data.SqlClientX.Handlers
+{
+    internal class HandlerOrchestrator
```
benrr101, line 12: I think it would be beneficial:
1) Rename to ConnectionHandlerOrchestrator (since it is specific to connection handling)
2) Move to the recent Handlers/Connection folder

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/HandlerOrchestrator.cs
benrr101, line 43: This can be simplified to
```
if (is null)
{
  throw
}

handle
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/HandlerOrchestrator.cs
benrr101: Yeah I really think we should create it as a singleton and generate the chain on construction

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/HandlerOrchestrator.cs
benrr101, line 60: I'm gonna be honest here, this seems vastly overcomplicated for what its doing. Rather than create a list, iterate over it, etc, etc ... just create them and wire them up as you go.

```
_initialHandler = new DataSourceParsingHandler();
var  transportHandler = new TransportCreationHandler();
_initialHandler.NextHandler = transportHandler;
var preLoginHandler = new PreLoginHandler();
_transportHandler.Next = preLoginHandler;
```

or something like that

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/HandlerNotFoundException.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
```
benrr101, line 8: Please make sure to optimize the using statements before checking in

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/ConnectionHandlerContext.cs
```diff
@@ -31,5 +31,40 @@ internal class ConnectionHandlerContext : HandlerRequest
...
+            }
+
+            if (this.ConnectionString != null)
+            {
+                clonedContext.ConnectionString = new SqlConnectionString(this.ConnectionString.ToString());
+            }
+
+            if (this.DataSource != null)
+            {
+                //TODO: Move DataSource class into SqlClientX and implement clone
```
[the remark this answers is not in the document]
benrr101: We will need to get on the same page for this. It was my idea to create a new data source class for SqlClientX for the purposes of having shared memory in the protocol enum. This also allows for making the data source object a data-only object by moving logic for parsing the data source into the data source parser handler. Making these kinds of changes to the existing DataSource object won't be feasible due to widespread changes it'll cause to the codebase, but not adding the enum value won't be feasible either.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/ConnectionHandlerContext.cs
```diff
@@ -31,5 +31,40 @@ internal class ConnectionHandlerContext : HandlerRequest
         /// Class used by orchestrator while chaining handlers.
         /// </summary>
         public Exception Error { get; set; }
+
+        /// <summary>
+        /// Method needed to clone ConnectionHandlerContext as part of history along the chain.
+        /// </summary>
+        public object Clone()
```
benrr101, line 38: I don't understand why we need to clone the context. Or why its history along the chain needs to be maintained. What is the reason for this?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/ConnectionHandlerContext.cs
benrr101: I think this is expected for the handler context. It is going to be a mostly data-only object that is passed between handlers, so having all properties set to null or default is likely a valid initial state. Additionally allowing all properties to be set without using a non-default constructor, allows for easier testing.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/InternalConnectionContext.cs
```diff
@@ -0,0 +1,19 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Data.SqlClientX.Handlers
+{
+    /// <summary>
+    /// Placeholder context class needed for SqlConnector
+    /// </summary>
+    internal class InternalConnectionContext
```
benrr101, line 15: If this is needed for SqlConnection, shouldn't it be added by the SqlConnector?
I'm confused why this is necessary.

benrr101 (changes requested):

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/ConnectionHandlerContext.cs
imasud00: I really doubt that null object is nulll because it is not able to do anything. Unless you think context is just a dictionary and does not have any logic. 
Context should ensure that it never gets in an invalid state and hence I forsee it has business logic otherwise there is a risk of duplicating the business across different handlers.

imasud00: At least it needs to have a Connection String.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/ConnectionHandlerContext.cs
imasud00, line 38: What is the purpose of cloning a context?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/HandlerNotFoundException.cs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace Microsoft.Data.SqlClientX.Handlers
+{
+    /// <summary>
+    /// Exception when a handler is not defined
+    /// </summary>
+    internal class HandlerNotFoundException : Exception
```
imasud00, line 15: I will suggest if we are creating our own exception than have a two base class may PermenantException or TransitentException and let us custom exception inherit from one. This also helps differentiate our own exception vs catch system level exception.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/HandlerOrchestrator.cs
imasud00, line 12: What is it orchestrating.? Is it specific for a particular TDS Message such as SqlBulk or something. Do we create different Orchestrator for different Messages for example BulkCopy message or is it generic orchestrator.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/HandlerNotFoundException.cs
imasud00: WHy not just throw new Excpetion ("HandlerNotFoundException")

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/ConnectionHandlerContext.cs
cheenamalhotra: I think this shouldn't be relevant anymore, based on last discussions, as we don't aim to keep clones of context, but only history of redirected endpoints in a specialized data object.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/InternalConnectionContext.cs
cheenamalhotra: Not needed IMO too.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/ConnectionHandlerContext.cs
samsharma2700: Yes, this isn't required anymore.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/HandlerOrchestrator.cs
samsharma2700: Good point. Updated.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/ConnectionHandlerContext.cs
samsharma2700: I added a to-do for it. Want to make sure how Orchestrator/chain needs to be initialized by the SqlConnector, will add necessary constructors with it.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/HandlerOrchestrator.cs
samsharma2700: Good point, will we have other kinds of orchestrators as well? Will move with the next PR.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/HandlerOrchestrator.cs
samsharma2700: Agreed on the overcomplication. My idea was that there will be multiple handlers in the chain DSPHandler->SSRP->TransPort->PreLogin->Login.. etc. so 10 statements just creating the chain but as a list we can have one single point of definition.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/HandlerNotFoundException.cs
samsharma2700: Will do

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/InternalConnectionContext.cs
samsharma2700: Would not be needed after the next PR.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/HandlerNotFoundException.cs
samsharma2700: Transient exceptions are handled by their respective handlers (maybe there is retry involved). Permanent exceptions are immediately thrown. 
The reason for creating HandlerNotFoundException is to have a new constructor which takes the inner exceptions and implements all 4 constructors https://learn.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2015/code-quality/ca1032-implement-standard-exception-constructors?view=vs-2015&redirectedfrom=MSDN

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/HandlerOrchestrator.cs
samsharma2700: Yes will do so on the next PR

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Connection/ConnectionHandlerContext.cs
```diff
@@ -91,5 +99,46 @@ internal class ConnectionHandlerContext
...
+        /// Adds the context of previous handler into the routingHistory object in case
+        /// of a reroute
+        /// </summary>
+        /// <param name="context"></param>
+        /// <exception cref="ArgumentNullException"></exception>
+        private void AddToRoutingHistory(ConnectionHandlerContext context)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException(nameof(context), "Context cannot be null");
```
edwardneal, line 138: Nit: resource strings.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/HandlerOrchestrator.cs
benrr101: I'm not entirely sure we *will* have other orchestrators, but I think this one at least needs to indicate it is for connection handling. No problem, moving stuff is likely going to happen at least several times.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/HandlerNotFoundException.cs
benrr101: I don't think these have actually been changed - ie, you only need `using System` here.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/HandlerOrchestrator.cs
```diff
@@ -0,0 +1,74 @@
...
+    {
+        /// <summary>
+        /// Class responsible for creation of chain of handlers, initiate chain, retain handler-context, maintain 
+        /// routing count and history of all handler contexts for all handlers in the chain
+        /// </summary>
+        private  IHandler<HandlerRequest> _initialHandler;
+        private ConnectionHandlerContext _initialConnectionHandlerContext;
+        private InternalConnectionContext _internalConnectionContext;
+
+        public HandlerOrchestrator(IHandler<HandlerRequest> initialHandler, ConnectionHandlerContext initialConnectionHandlerContext, InternalConnectionContext internalConnectionContext, List<ConnectionHandlerContext> history)
```
benrr101, line 22: Nit: When lines get this long, it becomes more readable to "chop" them into:
```
public ReturnType Method(
  Type1 parameter1,
  Type2 parameter2,
  Type3 parameter3)
{
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/HandlerOrchestrator.cs
```diff
@@ -0,0 +1,74 @@
...
+                throw new HandlerNotFoundException();
+                
+            }
+            await _initialHandler.Handle(_initialConnectionHandlerContext, isAsync, ct).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Creates the chain of handlers in the order definition
+        /// </summary>
+        private void CreateHandlerChain()
```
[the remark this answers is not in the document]
benrr101: I saw the comment to do this on the next PR, but I would encourage doing it as part of this PR :)

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/HandlerOrchestrator.cs
benrr101: I think this is a situation where readability wins out over rule-of-3. If the chain were to be dynamic, then I would agree that this approach is better. But since it's static, I think it makes more sense to just show x connects to y connects to z.

It's ok if you want to take this as part of the next PR that makes chain creation happen once.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/HandlerOrchestrator.cs
benrr101: Ok I see it was resolved, but that doesn't answer why we need explicit cast here :)

samsharma2700: Thought I left a note here. Compiler cannot infer that `IHandler<ConnectionHandlerContext>` can be cast to `IHandler<HandlerRequest>` without an explicit cast. I created _initialHandler to be generic but with recent changes this won't matter much.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/HandlerOrchestrator.cs
samsharma2700: Will make the change.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/HandlerOrchestrator.cs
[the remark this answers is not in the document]
samsharma2700: Will update with this PR

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Connection/ConnectionHandlerContext.cs
samsharma2700: Will look into it in the next PR

benrr101 (approved):

PR #2618, opened 2024-06-27
Fix | Fix decrypt failure to drain data

arellegue:
This fix is to fix the side effect of PR #1968.

When the SqlDataReader fails to decrypt an Encrypted column, a "Failed to decrypt" error is thrown. When that connection is returned to the pool and is used again, another error follows, "Internal error". This fix will ensure that when the "Failed to decrypt" error is thrown, it will drain that connection's buffer first before returning it to the connection pool.


JRahnama (approved):

DavoudEshtehari (approved):

mdaigle:
Hi @arellegue, I'm looking at this change in regard to #2683.
Was any testing added to avoid regressions in this area?

arellegue:
@mdaigle We discussed and decided to add a unit test for PR #2618.

PR #2635, opened 2024-07-01
Add a pooling data source

mdaigle:
Adds the skeleton of a concrete, pooling SqlDataSource class.
Introduces a "return" flow so that data sources can received an internal connection for recycling or closing.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/PoolingDataSource.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/PoolingDataSource.cs
@@ -1,0 +1,54 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+#if NET8_0_OR_GREATER
+
+using System.Collections.Concurrent;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Data.ProviderBase;
+using Microsoft.Data.SqlClient;
+
+namespace Microsoft.Data.SqlClientX
+{
+    internal sealed class PoolingDataSource : SqlDataSource
+    {
+        private SemaphoreSlim poolSemaphore;
+        private ConcurrentQueue<SqlConnector> _sqlConnectors;
+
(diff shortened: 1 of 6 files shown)


VladimirReshetnikov (approved):
Looks good

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlConnectionX.cs
```diff
@@ -184,13 +184,20 @@ internal Task Close(bool async)
...
+                var internalConnection = _internalConnection;
 
-                //TODO: remove the internal connection's reference to this wrapper connection
+                if (internalConnection != null)
+                {
+                    //TODO: cancel outstanding operations on connection before close
+                    //TODO: if pooling, reset the connector
+
+                    await internalConnection.Return(async).ConfigureAwait(false);
+                    internalConnection.OwningConnection = null;
```
vonzshik, line 198: Probably should be done before `Return` as otherwise there's a race condition with other thread renting connection and overwriting `OwningConnection`

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/PoolingDataSource.cs
```diff
@@ -0,0 +1,81 @@
...
+        internal int MinPoolSize => _connectionPoolGroupOptions.MinPoolSize;
+        internal int MaxPoolSize => _connectionPoolGroupOptions.MaxPoolSize;
+
+        /// <summary>
+        /// Initializes a new PoolingDataSource.
+        /// </summary>
+        //TODO: support auth contexts and provider info
+        PoolingDataSource(SqlConnectionStringBuilder connectionStringBuilder,
+            SqlCredential credential,
+            RemoteCertificateValidationCallback userCertificateValidationCallback,
```
saurabh500, line 35: Why do these callbacks exist?

mdaigle: I think this is a holdover from npgsql inspiration. I can clean them up and come back around later as the auth flows are fleshed out.

PR #2651, opened 2024-07-05
Fix - Don't error when using infinte connect timeout and Entra auth

### LINKED ISSUE #2563: ActiveDirectoryAuthenticationProvider.AcquireTokenAsync method fails when ConnectTimeout set to 0 or >= int.MaxValue/1000
Well that was fun to debug!

https://github.com/dotnet/SqlClient/blob/3f0c4b1a706136e417efa3a49bc168342470513e/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ActiveDirectoryAuthenticationProvider.cs#L117

As the `ConnectionTimeout` is used to set the cancellation token's `CancelAfter` * 1000...

- If `ConnectionTimeout = 0` then `0 * 1000 = 0` - this causes the cancellation token to immediately expire, so all operations such as gather the tenant metadata fail with `A task was canceled` exception message (which is what led me down this rabbit hole).
- If `ConnectionTimeout = int.MaxValue` (or any value greater than int.MaxValue / 1000) then `int.MaxValue * 1000` throws as the result is larger than int.

Ideally this needs more defensive coding, eg:
- When `ConnectionTimeout == 0` we do not require a `CancelAfter`?
- When `ConnectionTimeout >= int.MaxValue / 1000` set the `CancelAfter` to `int.MaxValue`?

David-Engel:
Fixes #2563

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ActiveDirectoryAuthenticationProvider.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ActiveDirectoryAuthenticationProvider.cs
@@ -120,7 +120,11 @@ namespace Microsoft.Data.SqlClient
             using CancellationTokenSource cts = new();
 
             // Use Connection timeout value to cancel token acquire request after certain period of time.
-            cts.CancelAfter(parameters.ConnectionTimeout * 1000); // Convert to milliseconds
+            int timeout = parameters.ConnectionTimeout * 1000;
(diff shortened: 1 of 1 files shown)


DavoudEshtehari (approved):

David-Engel:
@DavoudEshtehari @dauinsight @arellegue - CI is clean! 🎉

dauinsight (approved):

JRahnama (approved):

PR #2654, opened 2024-07-08
Access Token Support for SqlConnection in Manual Tests

dauinsight:
In our manual tests we are changing the way we authenticate to Azure SQL DB by using managed identity.

In order to support this, when we make a connection from a macOS container, we must assign an access token to the connection. This authorizes the instance with access to the managed identity and hence the DB.

A few tests have been skipped (when running macOS) as we're unable to assign the access token to the connection from the test.


David-Engel:
This seems like it would make it difficult to change test connection strings due to validation on the AccessToken property. What problem is this PR solving?

dauinsight:
> This seems like it would make it difficult to change test connection strings due to validation on the AccessToken property. What problem is this PR solving?

Since assigning a UAMI directly to an agent pool is not recommended + our Mac agent pool, we are using a service connection in ADO and have federated credentials to it on a UAMI which has access to the database. As a result, we have an Azure Cli step in ADO to retrieve the token and assign it to the connection to authenticate.

arellegue (approved):
Looks good to me.

src/Microsoft.Data.SqlClient/tests/ManualTests/ProviderAgnostic/ReaderTest/ReaderTest.cs
```diff
@@ -24,8 +24,13 @@ public static void TestMain()
...
             try
             {
-                using (DbConnection con = provider.CreateConnection())
+                using (SqlConnection con = (SqlConnection) provider.CreateConnection())
                 {
+                    // Testing in macOS Azure SQL container requires access token authentication
+                    if (!DataTestUtility.AuthenticatingWithoutAccessToken)
+                    {
+                        con.AccessToken = DataTestUtility.AADAccessToken;
+                    }
```
DavoudEshtehari, line 33: If you introduce an extension method your change will be simpler without repeating yourself.

```suggestion
                using (DbConnection con = provider.CreateConnectionEx())
```

```c#
static DbConnection CreateConnectionEx(this DbProviderFactory provider)
{
    DbConnection conn = provider.CreateConnection();

    // Testing in macOS Azure SQL container requires access token authentication
    if (!DataTestUtility.AuthenticatingWithoutAccessToken && conn is SqlConnection sqlConn)
    {
        sqlConn.AccessToken = DataTestUtility.AADAccessToken;
        return sqlConn;
    }
    return conn;
}
```

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
```diff
@@ -114,7 +114,16 @@ public static bool TcpConnectionStringDoesNotUseAadAuth
...
+                return builder.Authentication == SqlAuthenticationMethod.SqlPassword || (builder.Authentication == SqlAuthenticationMethod.NotSpecified && AuthenticatingWithoutAccessToken);
+            }
+        }
+
+        // Testing in macOS Azure SQL container requires access token authentication
+        public static bool AuthenticatingWithoutAccessToken
+        {
+            get
+            {
+                return !RuntimeInformation.IsOSPlatform(OSPlatform.OSX);
```
DavoudEshtehari, line 126: This more sense with positive naming. Feel free to modify the naming as you like.
```suggestion
        public static bool ForceAuthenticationWithAccessToken
        {
            get
            {
                return RuntimeInformation.IsOSPlatform(OSPlatform.OSX);
```

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/SqlSchemaInfoTest/SqlSchemaInfoTest.cs
```diff
@@ -76,16 +76,18 @@ public static void TestCommandBuilder()
...
         public static void TestInitialCatalogStandardValues()
         {
-            using (SqlConnection connection = new SqlConnection(DataTestUtility.TCPConnectionString))
+            Console.WriteLine($"TCPConnectionString = {DataTestUtility.TCPConnectionString}");
+
+            using (SqlConnection connection = DataTestUtility.GetSqlConnection(DataTestUtility.TCPConnectionString))
             {
                 string currentDb = connection.Database;
-                SqlConnectionStringBuilder builder = new SqlConnectionStringBuilder(connection.ConnectionString);
+                SqlConnectionStringBuilder builder = new SqlConnectionStringBuilder(DataTestUtility.TCPConnectionString);
```
DavoudEshtehari, line 88: I believe the point of this test is building and reading the connection string properties from a `SqlConnection.ConnectionString`. So, instead of skipping these kind of tests you can force the using of the `DataTestUtility.TCPConnectionString` through the helper method by adding a new argument and bypassing the internal logic.
```suggestion
                SqlConnectionStringBuilder builder = new SqlConnectionStringBuilder(connection.ConnectionString);
```

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/SqlBulkCopyTest/Helpers.cs
```diff
@@ -15,10 +17,16 @@ internal static void ProcessCommandBatch(Type connType, string constr, string[]
             {
                 object[] activatorArgs = new object[1];
                 activatorArgs[0] = constr;
-                using (DbConnection conn = (DbConnection)Activator.CreateInstance(connType, activatorArgs))
+                using (SqlConnection conn = (SqlConnection)Activator.CreateInstance(connType, activatorArgs))
                 {
+                    // Testing in macOS Azure SQL container requires access token authentication
+                    if (!DataTestUtility.AuthenticatingWithoutAccessToken)
```
DavoudEshtehari, line 23: You can add another property in `DataTestUtility` and keep it positive, instead of multiplying a negative property by adding a negative sign, which is a bit tricky to follow.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/SqlBulkCopyTest/AdjustPrecScaleForBulkCopy.cs
```diff
@@ -11,7 +11,8 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests
 {
     public static class AdjustPrecScaleForBulkCopy
     {
-        [ConditionalFact(typeof(DataTestUtility), nameof(DataTestUtility.AreConnStringsSetup))]
+        // Skip: We cannot assign an access token to the connection in this test (#31362)
+        [ConditionalFact(typeof(DataTestUtility), nameof(DataTestUtility.AreConnStringsSetup), nameof(DataTestUtility.AuthenticatingWithoutAccessToken))]
```
DavoudEshtehari, line 15: It doesn't make sense to me that authentication could affect this test. Could you double-check it?

David-Engel:
> > This seems like it would make it difficult to change test connection strings due to validation on the AccessToken property. What problem is this PR solving?
> 
> Since assigning a UAMI directly to an agent pool is not recommended + our Mac agent pool, we are using a service connection in ADO and have federated credentials to it on a UAMI which has access to the database. As a result, we have an Azure Cli step in ADO to retrieve the token and assign it to the connection to authenticate.

@dauinsight 

Can we simply use Authentication=ActiveDirectoryDefault for the connection string on the mac jobs and take advantage of the fact that AzureCliCredential is part of the credential chain?

I feel like requiring all tests to wrap their connection strings in DataTestUtility.GetSqlConnection is adding unnecessary complexity to test configuration.

If AD Default doesn't work, another alternative that would be much simpler/cleaner is to create a custom authentication provider that overrode the ActiveDirectoryMSI option to return the access token configured on the job.

dauinsight:
> > > This seems like it would make it difficult to change test connection strings due to validation on the AccessToken property. What problem is this PR solving?
> > 
> > 
> > Since assigning a UAMI directly to an agent pool is not recommended + our Mac agent pool, we are using a service connection in ADO and have federated credentials to it on a UAMI which has access to the database. As a result, we have an Azure Cli step in ADO to retrieve the token and assign it to the connection to authenticate.
> 
> @dauinsight
> 
> Can we simply use Authentication=ActiveDirectoryDefault for the connection string on the mac jobs and take advantage of the fact that AzureCliCredential is part of the credential chain?
> 
> I feel like requiring all tests to wrap their connection strings in DataTestUtility.GetSqlConnection is adding unnecessary complexity to test configuration.
> 
> If AD Default doesn't work, another alternative that would be much simpler/cleaner is to create a custom authentication provider that overrode the ActiveDirectoryMSI option to return the access token configured on the job.

@David-Engel I agree that this adds a little bit of complexity, but i haven't found another way to use our service connection in ADO to grant MI to the agent using an access token. 

ActiveDirectoryDefault works locally if we login to our machine via `az login` but not in the pipeline.

For more context into the implementation of this PR: https://msazure.visualstudio.com/One/_wiki/wikis/Microsoft%20Graph%20Engineering/675144/SFI-MI?anchor=mi-for-e2es-in-dev-environment-and-azure-devops(ado)

David-Engel:
@dauinsight
 > For more context into the implementation of this PR: https://msazure.visualstudio.com/One/_wiki/wikis/Microsoft%20Graph%20Engineering/675144/SFI-MI?anchor=mi-for-e2es-in-dev-environment-and-azure-devops(ado)

After reading that, in addition to the custom provider option I mentioned, prioritizing #2557 could also provide a better solution that could be used. As it is, I would NOT merge this PR. Dynamically modifying the connection string like that across all tests is just asking for maintenance headaches and added cost.

Additionally, I noticed the AZURE_DB_SP_TCP_CONN_STRING was removed from the test config variable group. That could have been used for macOS, too. We have an exception for that for continuing to test AD SP auth.

CC: @saurabh500 @cheenamalhotra

saurabh500:
I strongly agree with @David-Engel 
Unintentional modification of the connection string puts too many hinderances in test enhancements.

dauinsight:
Okay that makes sense, for now i'll look to have the service principal set up again.

PR #2657, opened 2024-07-09
Duplicate DataSource class and add SharedMemory to protocol enum

benrr101:
**Description**: This PR does two things
* Duplicates the DataSource class from SNIProxy into SqlClientX
  * The main reason for doing this is because the DataSource.Protocol enum doesn't include shared memory. Adding shared memory to the existing data source doesn't seem safe, so duplicating it seems to be the best approach for now
  * The second reason for doing so is to allow for making DataSource a dumb object, and move parsing logic into the data source parsing handler (this is not implemented as part of this PR since it is non-trivial).
  * Applying some stylistic changes
    * Sorting methods/properties
    * Cleaning up names as per naming rules
    * Using `x is y` style comparisons when possible (see inline PR comments)
    * Flipping `constant.Equals(variable)` to `variable == constant`
* Add SharedMemory to DataSource.Protocol enum
  * Enum is moved to `DataSourceProtocol` enum, although this is up for debate.

**Testing**:
* Existing test cases pass


cheenamalhotra:
> The main reason for doing this is because the DataSource.Protocol enum doesn't include shared memory

What's the reason for adding SharedMemory protocol to Data Source class? Can you explain a bit about how you intend to use this protocol later?

mdaigle (commented):
Everything makes sense to me, and I agree with the clarity of having an explicit value for shared memory rather than defaulting to it if we don't match any other protocol. 

But, in the interest of focus on the happy path, will duplicating DataSource lead us to duplicating a bunch of other logic as well?

cheenamalhotra:
Given that Shared Memory provider wasn't supported in Managed SNI, it makes it less important to be supported anytime soon for the scope of ClientX, so you may proceed accordingly. Yes there will be duplication needed, but it feels like this can be pushed to a later stage, when ClientX can eventually replace Managed SNI implementation.

cheenamalhotra:
Closing ClientX PRs for time being.

ErikEJ:
@cheenamalhotra can you share any updated plans or roadmap for SqlClientX ??

cheenamalhotra:
> @cheenamalhotra can you share any updated plans or roadmap for SqlClientX ??

Still in baking, we will share more on it soon.. Currently we're focusing on shipping 6.0 with critical changes.

PR #2663, opened 2024-07-10
TDS Reader implementation

cheenamalhotra:
Addresses https://github.com/dotnet/SqlClient/issues/2624

Using ArrayPool to rent and return bytes, instead of rewriting on the same byte array as old design, as this provides a threadsafe read/write mechanism.


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsBufferAlloc.cs
```diff
@@ -2,37 +2,104 @@
...
+        public ValueTask DoWithRentedBuffer(int size, OperationAsync operation)
+        {
+            byte[] buffer = RentBuffer(size);
+            try
+            {
+                return operation(buffer);
+            }
+            finally
+            {
+                ReturnBuffer(buffer);
```
Tornhoof, line 60: Note: in runtime, the general pattern is usually not to return the buffer on exception, but let the GC claim it. There are multiple reasons for that, e.g. no try/finally and after an exception the state might not be well understood/defined  anymore.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsBufferAlloc.cs
```diff
@@ -2,37 +2,104 @@
...
     /// </summary>
-    internal class TdsBufferAlloc 
+    internal class TdsBufferManager
     {
-        private readonly byte[] _buffer;
+        private readonly ArrayPool<byte> _bufferPool = ArrayPool<byte>.Shared;
+
+        private byte[] RentBuffer(int size) => _bufferPool.Rent(size);
+
+        private void ReturnBuffer(byte[] buffer) => _bufferPool.Return(buffer, clearArray: true);
```
Tornhoof, line 19: Are you sure you want to always clear the buffer, it's usually unnecessary overhead for byte buffers.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReader.cs
```diff
@@ -0,0 +1,227 @@
...
+                : (byte)_tdsStream.ReadByte(); // UNSAFE - TODO Expose ReadByte from TdsStream that returns 'byte'.
+        }
+        /// <summary>
+        /// Reads next char from TDS stream asynchronously.
+        /// </summary>
+        /// <param name="isAsync">Whether caller method is executing asynchronously.</param>
+        /// <param name="ct">Cancellation token.</param>
+        /// <returns>Async <see cref="ValueTask"/> that returns a <see cref="char"/> value.</returns>
+        public ValueTask<char> ReadCharAsync(bool isAsync, CancellationToken ct)
+            => DoWithRentedBuffer(sizeof(char), async buffer =>
```
Tornhoof, line 52: You might want to benchmark, how much this extra functor indirection affects performance, besides readability. I would personally not do it but use the pool directly in here

roji: I think that even more important point, is the use of asynchronous lambdas or functions; each aync function adds an additional state machine, which can add up to a lot of overhead. So the general tendency to refactor common code out to smaller functions doesn't work well for async.

As a very general guidance, I'd recommend going over where async is used, and trying to minimize the instances of that, generally concentrating it in a few few big functions rather than spreading it all around in a lot of small helpers.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReader.cs
```diff
@@ -0,0 +1,227 @@
...
+            return result;
+        }
+
+        /// <summary>
+        /// Reads short value from Tds Stream asynchronously.
+        /// </summary>
+        /// <param name="isAsync">Whether caller method is executing asynchronously.</param>
+        /// <param name="ct">Cancellation token.</param>
+        /// <returns>Async <see cref="ValueTask"/> that returns a <see cref="short"/> value.</returns>
+        public ValueTask<short> ReadInt16Async(bool isAsync, CancellationToken ct)
```
edwardneal, line 85: Tying to roji and Tornhoof's earlier comments, the rough pseudocode I'd suggest for the strongly-typed read logic is:
```csharp
public ValueTask<X> ReadXAsync(int lengthOfX, bool isAsync, CancellationToken ct)
{
    const int StackAllocationThreshold = 512;

    byte[] rentedBuffer = isAsync || lengthOfX > StackAllocationThreshold
        ? ArrayPool<byte>.Shared.Rent(lengthOfX)
        : null;
    Span<byte> buffer = rentedBuffer == null
        ? stackalloc byte[StackAllocationThreshold].Slice(0, lengthOfX)
        : rentedBuffer.AsSpan(0, lengthOfX);

    try
    {
        ct.ThrowIfCancellationRequested();

        int bytesRead = isAsync
            ? await _tdsStream.ReadAsync(rentedBuffer.AsMemory(0, lengthOfX), ct).ConfigureAwait(false)
            : _tdsStream.Read(buffer);

        if (bytesRead != lengthOfX)
        {
            throw new InvalidOperationException(...);
        }

        return BinaryPrimitives.ReadInt16LittleEndian(buffer);
    }
    finally
    {
        if (rentedBuffer != null)
        {
            ArrayPool<byte>.Shared.Return(rentedBuffer);
        }
    }
}
```

This simplifies when the type `X` and `lengthOfX` is constant, and it's possible that the comparison against `bytesRead` is unnecessary. The main objective is to avoid unnecessary heap allocations, but it makes the use of `TdsBufferManager` more difficult. It also doesn't account for streaming data, or long strings/byte arrays.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReader.cs
```diff
@@ -0,0 +1,227 @@
...
+        /// <param name="length">Length of array to read.</param>
+        /// <param name="isAsync">Whether caller method is executing asynchronously.</param>
+        /// <param name="ct">Cancellation token.</param>
+        /// <returns>Async <see cref="ValueTask"/> that returns a <see cref="char"/> array.</returns>
+        public async ValueTask<char[]> ReadCharArrayAsync(int length, bool isAsync, CancellationToken ct)
+        {
+            char[] result = new char[length];
+            for (int i = 0; i < length; i++)
+            {
+                result[i] = await ReadCharAsync(isAsync, ct).ConfigureAwait(false);
```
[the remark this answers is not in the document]
roji: Agree - I'd definitely avoid doing char-by-char reading. If you read a bunch of bytes, you should be able to slice that Span/Memory and do endianness (or whatever) without any unsafe code, I thnk.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReader.cs
roji: I have very little context and knowledge of the overall design, but this seems all seems like it would do a ReadAsync for 2 bytes (Int16); at the very least this would mean an async state machine (even on sync completion, where the two bytes are buffered somewhere inside `_tdsStream`), for each and every little read.

For efficient async I/O, it's usually a good idea to split apart the I/O code path - which is the async part where actual I/O can happen, and which is only ever invoked when you know you need to do I/O (not enough bytes cached), and processing/decoding, which operates purely on memory that's known to conain cached data, and is purely synchronous code. For example, this allows you to first make sure that you've read X bytes which you know you're going to need (for a big/complex protocol message or similar), and then you can process that 100-byte chunk efficiently without ever doing async etc (this is how Npgsql works).

Basically the game is how to reduce the amount of needless async/await by lifting that code up the stack, and creating big "chunks" of I/O, rather than letting async permeate everywhere, causing tons of state machines everywhere where you need ro read a byte or two.

One last thing - putting too much pressure on the shared array pool can end up being a source of perf problems too, as contention rises (because too many threads are constantly doing rent/return for tiny buffers). Npgsql has a per-connection private buffer into which we read (so zero contention by design) - you could also have a private ArrayPool (as opposed to the global shared one) on your physical connections. But of course, if/when you know you have a small, fixed-size payload, stackalloc is indeed best.

Tornhoof: > I have very little context and knowledge of the overall design

This is a very valid point, it is hard to suggest improvements, if we don't really know the invariants,   e.g. which .NET version is minimum etc.

There quite a few high performance concepts to work with buffers in recent .NET versions, incl. PipeReader, ReadOnlySequence etc, some are designed to keep the async path as small as possible, a problem here as shay notes. In one of the other discussions, quite a few nice diagrams were shown to better describe the lay of the land, so to speak.
Maybe this is helpful here too.

edwardneal: In most cases for the TDS protocol, I agree - I'd prefer to be able to read the first X bytes, read the token type, read the remaining bytes in the token and interpret the Span/Memory. For most TDS tokens in a token stream (each subsection of [section 2.2.7](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-tds/67b6113c-d722-42d1-902c-3f6e8de09173) in the spec is one type of token) we have a length field, so this is pretty practical.

Reading row data is harder though. As I understand it, SQL Server sends a COLMETADATA token describing the column data types (and in some cases, lengths) in each result set. It then sends a ROW token, followed by a continuous stream of data (the spec refers to this as a tokenless stream). We don't know the length of this ROW token, and can't guarantee that we can calculate it from COLMETADATA; in the case of nullable or dynamic-length data, there's no choice but to read from the TDS stream field-by-field. Once all rows in the result set have been sent, it sends a DONE token which contains the number of rows.

A field is naturally its own type of data - it's essentially the same basic principles and problems...

If COLMETADATA indicates that X consecutive fields are non-nullable and fixed-length then we can optimise the reads a little, and we might even know the minimum (and/or total) length of each row which follows. I think we should read byte Spans and interpret rows outside of the IO completion in this case.

The TDS protocol thus hedges us towards streaming reads. This definitely makes life easier for the SQL Server instance (which can now stream rows and field values as it reads them) but it can make it difficult to separate the IO and interpretation layers. There's a degree of separation here because `TdsReadStream` reads TDS-packet-by-TDS-packet, but it doesn't eliminate the problems which come from allocating the state machines for synchronous completions on the async path.

I'd definitely like to understand how the TDS reader/parser is going to be designed though.

(side note: are you sure that sync completion of the async path allocates? I haven't benchmarked to check, but thought that .NET pooled ValueTask.CompletedTask and that from .NET 5.0 this allocation [didn't occur](https://devblogs.microsoft.com/dotnet/async-valuetask-pooling-in-net-5/#:~:text=When%20used%20as,ValueTask%3CTResult%3E.).)

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriter.cs
```diff
@@ -176,41 +180,45 @@ public async ValueTask WriteBytesAsync(byte[] data, bool isAsync, CancellationTo
         /// <returns>Span of bytes for the TDS decimal type.</returns>
         public Span<byte> SerializeDecimal(decimal value)
```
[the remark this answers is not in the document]
Wraith2: The function signature is incorrect for span usage. In general for low level apis you do not want to return a span you should provide one as an input parameter and then have the return value indicate if the operation succeeded. You also need an out int parameter for variable length writes, tuples are useful here. See the functions in the BCL that return OperationResult as an example of this. 

My suggested version of the SerializeDecimal function would be this:
```csharp
        public OperationStatus SerializeDecimal(decimal value, Span<byte> buffer)
        {
            if (buffer.Length < 17)
            {
                return OperationStatus.DestinationTooSmall;
            }

            Span<int> decimalBits =
#if !NET5_0_OR_GREATER
                stackalloc int[4];
            decimal.TryGetBits(value, decimalBits, out int valuesWritten);
#else
            decimal.GetBits(value);
#endif

            /*
                Returns a binary representation of a Decimal. The return value is an integer
                array with four elements. Elements 0, 1, and 2 contain the low, middle, and
                high 32 bits of the 96-bit integer part of the Decimal. Element 3 contains
                the scale factor and sign of the Decimal: bits 0-15 (the lower word) are
                unused; bits 16-23 contain a value between 0 and 28, indicating the power of
                10 to divide the 96-bit integer part by to produce the Decimal value; bits 24-
                30 are unused; and finally bit 31 indicates the sign of the Decimal value, 0
                meaning positive and 1 meaning negative.
                SQLDECIMAL/SQLNUMERIC has a byte stream of:
                struct {
                    BYTE sign; // 1 if positive, 0 if negative
                    BYTE data[]; // 16 bits = int [4]
                }
                For TDS 7.0 and above, there are always 17 bytes of data
            */
            // write the sign (note that COM and SQL are opposite)
            buffer[0] = (uint)decimalBits[3] != 0u ? (byte)1 : (byte)0;

            BinaryPrimitives.TryWriteInt32LittleEndian(buffer.Slice(1, 4), decimalBits[0]);
            BinaryPrimitives.TryWriteInt32LittleEndian(buffer.Slice(5, 4), decimalBits[1]);
            BinaryPrimitives.TryWriteInt32LittleEndian(buffer.Slice(9, 4), decimalBits[2]);
            BinaryPrimitives.TryWriteInt32LittleEndian(buffer.Slice(13, 4), 0);

            return OperationStatus.Done;
        }
```

- It uses the newest verison of TryGetBits if it can so we can avoid a heap allocation.
- It avoids using the `current` variable to keep track of the position because every offset is statically known.
- it simplifies the sign check slightly, see [sharplab](https://sharplab.io/#v2:EYLgtghglgdgNAFxBAzmAPgAQEwAYCwAUDgIxHmFHACeCApgAQCCAFAEp0QAmA8jADbUAygAcIMADywEAPgZc6AYyiR+AISgIUASiIBvIgyMNMAdgYsa9bS1ssArtO0LlqjVoDaAZgC62hgB6DLgAHgAcuJFRuP4ycl4k2gDcRAC+FFaMauycvALCYpLSci4qEOqaOvqGxmYWjjAIzkplFZ6+DACEALzB9gwA/BaZ2iQMIMO0dNq4KYTplITGDKLiUo0lLW6VDL0oCBCKANbl/AD2igzSHgAsPnPL9jCKABZKR3RcevJb5e4o3h8uwsTlCEWikSSC2W0gYEGBrFK2y0yRqRlhwGB2SRf0qqMWQA=) for the reason.
- It takes the buffer to write the decimal into as a parameter so the caller manages the renting and pooling. 

If you try return rented buffers you're going to have nasty lifetime issues with memory and you're going to spend a lot of cpu time renting returning and checking rentals etc. Don't think about it that way. Use span and memory everywhere and manage them as far up as possible.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReader.cs
Wraith2: You also need to know that any value longer than a single byte can span packets and you have to deal with it correctly. This sounds easy but forces you to code all read steps knowing that the value may be incomplete and that the minimum number of bytes you will get is 1 and that you may need a network read to get more. This makes life very complicated in the existing codebase. One of the reasons I wanted to use pipelines was that they use ReadOnlySequence which encapsulates this logic very nicely.

The same is true for writes. You are expected to fill every packet apart from the last one in a stream so you must be able to write a minimum of 1 byte of every fixed length value to the output buffer and then write the rest somewhere else. Using the OperationResult pattern you can implement this using retry on too-short buffers rather than using exceptions or having all leaf methods be able to re-request buffers.

benrr101 (changes requested):
Thought I finished this review a while ago, but looks like I didn't

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsBufferAlloc.cs
cheenamalhotra: Simplified back to dedicated buffer, to avoid complications of array pool, also noticed array pool allocates a lot more than requested, so may not be the best solution applying to us.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReader.cs
cheenamalhotra: Simplified

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReader.cs
```diff
@@ -0,0 +1,227 @@
...
+            });
+
+        /// <summary>
+        /// Reads string value from Tds Stream asynchronously.
+        /// </summary>
+        /// <param name="length">Length of string to read.</param>
+        /// <param name="isAsync">Whether caller method is executing asynchronously.</param>
+        /// <param name="ct">Cancellation token.</param>
+        /// <returns>Async <see cref="ValueTask"/> that returns a <see cref="string"/> value.</returns>
+        public ValueTask<string> ReadStringAsync(int length, bool isAsync, CancellationToken ct)
```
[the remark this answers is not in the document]
cheenamalhotra: I've implement chunk reading in ReadString, so that should help us dynamically adjust buffer size based on length of strings to be read.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsWriter.cs
```diff
@@ -5,16 +5,17 @@
 using System;
 using System.Buffers.Binary;
 using System.Diagnostics;
+using System.Linq;
```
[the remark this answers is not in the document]
cheenamalhotra: Not needed too..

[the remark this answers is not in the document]
cheenamalhotra: No, I don't need it... it came from thin air.. 🛩️

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsBufferAlloc.cs
```diff
@@ -2,37 +2,104 @@
...
+        /// <summary>
+        /// Performs <paramref name="operation"/> asynchronously with rented buffer.
+        /// </summary>
+        /// <typeparam name="TResult"></typeparam>
+        /// <param name="size"></param>
+        /// <param name="operation"></param>
+        /// <returns></returns>
+        public ValueTask<TResult> DoWithRentedBuffer<TResult>(int size, OperationAsync<TResult> operation)
+        {
+            byte[] buffer = RentBuffer(size);
```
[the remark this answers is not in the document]
cheenamalhotra: Still weighing in to whether we need renting abilities.. reverted back to dedicated array for now.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsBufferAlloc.cs
```diff
@@ -2,37 +2,104 @@
...
+using System.Buffers;
+using System.Threading.Tasks;
 
 namespace Microsoft.Data.SqlClientX.IO
 {
     /// <summary>
     /// Provides buffer management for reader/writer implementations
     /// </summary>
-    internal class TdsBufferAlloc 
+    internal class TdsBufferManager
```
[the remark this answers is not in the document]
cheenamalhotra: Simplified to basic design

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReader.cs
cheenamalhotra: Addressed majority of the Async-await feedback and simplified back to dedicated buffer for now.
However few follow ups to do:
* Improve buffer allocations (keeping it out of scope here as it may need some refactoring from Streams)
* Redesign reading Char Array and String

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReader.cs
```diff
@@ -0,0 +1,350 @@
...
+            }
+
+            // Calculate byte length for char array
+            int byteLength = sizeof(char) * length;
+
+            // Allocate byte array to store the result
+            byte[] bytes = new byte[byteLength];
+
+            // Read bytes asynchronously
+            await ReadBytesAsync(new ArraySegment<byte>(bytes), isAsync, ct);
```
saurabh500, line 102: ConfigureAwait?

saurabh500 (approved):

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/TdsReader.cs
cheenamalhotra:
```suggestion
            await ReadBytesAsync(new ArraySegment<byte>(bytes), isAsync, ct).ConfigureAwait(false);
```

cheenamalhotra:
Will proceed to merge by EOD if there is no additional feedback.

PR #2664, opened 2024-07-10
Bring 'feat/sqlclientx' branch up-to-speed with 'main'

cheenamalhotra:
Bringing all latest updates from 'main' to 'feat/sqlclientx' to avoid future conflicts.


benrr101 (approved):
Tbh, I'm not going to analyze all these changes since I don't think they make much impact for sqlclientx. If it builds, I say :shipit:

Though, I wonder if this change should be merged via --no-ff rather than a squash and merge. I'm not 100% sure how this works, but I think if we preserve the commit history like that, it will keep future merging from reporting a boatload of conflicts?

cheenamalhotra:
Used magic powers to retain history. :)

PR #2667, opened 2024-07-11
Add scaffolding for pool throttling

mdaigle:
This adds a mechanism for throttling enforcement within the connection pool. Basic passthrough and concurrency limiters are included. Focusing on the high throughput case (no rate limit), so I will add a backlog work item to implement the pool blocking period rate limiter.


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlDataSource.cs
```diff
@@ -40,14 +36,10 @@ protected override SqlConnectionX CreateDbConnection()
 
         internal SqlDataSource(
             SqlConnectionStringBuilder connectionStringBuilder,
-            SqlCredential credential,
-            RemoteCertificateValidationCallback userCertificateValidationCallback,
```
mdaigle, line 44: removing these for now until we fully implement additional auth pathways

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/RateLimiters/BlockingPeriodRateLimiter.cs
```diff
@@ -0,0 +1,33 @@
...
+
+#nullable enable
+
+namespace Microsoft.Data.SqlClientX.RateLimiters
+{
+    /// <summary>
+    /// A rate limiter that enforces a backoff (blocking) period upon error. 
+    /// Each subsequent error increases the blocking duration, up to a maximum, until a success occurs.
+    /// </summary>
+    internal class BlockingPeriodRateLimiter : IRateLimiter
```
edwardneal, line 17: Nit, but the same also applies to the other rate limiters.

```suggestion
    internal sealed class BlockingPeriodRateLimiter : IRateLimiter
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/RateLimiters/ConcurrencyRateLimiter.cs
```diff
@@ -0,0 +1,66 @@
...
+
+namespace Microsoft.Data.SqlClientX.RateLimiters
+{
+    /// <summary>
+    /// A rate limiter that enforces a concurrency limit. 
+    /// When the limit is reached, new requests must wait until a spot is freed upon completion of an existing request.
+    /// </summary>
+    internal class ConcurrencyRateLimiter : IRateLimiter
+    {
+        private SemaphoreSlim _concurrencyLimitSemaphore;
```
edwardneal, line 17: `SemaphoreSlim` is disposable, so we'll need some plumbing to dispose of any rate limiters with the connection pool.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/PoolingDataSource.cs
```diff
@@ -49,7 +51,12 @@ internal override ValueTask<SqlConnector> GetInternalConnection(SqlConnectionX o
         /// <inheritdoc/>
         internal override ValueTask<SqlConnector> OpenNewInternalConnection(SqlConnectionX owningConnection, TimeSpan timeout, bool async, CancellationToken cancellationToken)
         {
-            throw new NotImplementedException();
+            return _rateLimiter.Execute<SqlConnector>(() => RateLimitedOpen(owningConnection, timeout, async, cancellationToken), async, cancellationToken);
```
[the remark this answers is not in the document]
mdaigle: I was trying to balance these additional/duplicated closed params with keeping the rate limiter agnostic to the functionality provided by the callback. I added a delegate that includes a bool and CancellationToken as in parameters, and gives a hint to the rate limiter for how to populate these appropriately by exposing param names.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlDataSourceBuilder.cs
```diff
@@ -42,11 +46,67 @@ public SqlDataSourceBuilder(string connectionString = null, SqlCredential creden
...
+                DbConnectionPoolGroupOptions poolGroupOptions = new DbConnectionPoolGroupOptions(
+                    ConnectionStringBuilder.IntegratedSecurity,
+                    ConnectionStringBuilder.MinPoolSize,
+                    ConnectionStringBuilder.MaxPoolSize,
+                    //TODO: carry over connect timeout conversion logic from SqlConnectionFactory? if not, don't need an extra allocation for this object, just use connection string builder
+                    ConnectionStringBuilder.ConnectTimeout,
+                    ConnectionStringBuilder.LoadBalanceTimeout,
+                    ConnectionStringBuilder.Enlist);
+
+                //TODO: evaluate app context switch for concurrency limit
```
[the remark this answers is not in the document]
mdaigle: We haven't taken a hard stance on this yet, but in general we want to enable new functionality by default and provide a mechanism for opt-out (using appropriate semver). Otherwise it will be practically impossible to drive adoption.

One motivation behind the SqlDataSourceBuilder pattern we're introducing is that we can also expose an ExperimentalSqlDataSourceBuilder which may break contract often and grants access to preview experimental features. As features are proven, they can be ported over to the base SqlDataSourceBuilder and default enabled with option to opt out.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/RateLimiters/IRateLimiter.cs
```diff
@@ -0,0 +1,37 @@
...
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Data.SqlClientX.RateLimiters
+{
+    /// <summary>
+    /// An interface for rate limiters that execute arbitraty code. Intended to be small and self contained and chained together to achieve more complex behavior.
+    /// </summary>
+    internal abstract class IRateLimiter
```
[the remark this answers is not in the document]
mdaigle: 100% the long term idea behind this pattern. For now, I'm biasing towards leaving new functionality as internal. We need to be careful with the public API we expose because we're required to update docs and other things appropriately, so just want to cross that bridge when we get to it.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlDataSourceBuilder.cs
```diff
@@ -42,11 +46,67 @@ public SqlDataSourceBuilder(string connectionString = null, SqlCredential creden
...
+                IRateLimiter rateLimiter = IsBlockingPeriodEnabled() ? new BlockingPeriodRateLimiter() : new PassthroughRateLimiter();
```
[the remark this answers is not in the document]
mdaigle: See my comments above: https://github.com/dotnet/SqlClient/pull/2667#discussion_r1676551848

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/RateLimiters/IRateLimiter.cs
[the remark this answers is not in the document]
mdaigle: Of note, ResiliencePipelineT in polly understandably requires access to the return type of the pipeline.
https://github.com/App-vNext/Polly/blob/main/src/Polly.Core/ResiliencePipelineT.Async.cs#L5
So to integrate with Polly, we'd need to expose the SqlConnector type which should be an implementation detail of the driver. There may be some solution with an added layer of indirection? But it's not my top focus right now. Most customers are looking for fewer/no rate limits, not the ability to customize their rate limiting.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/RateLimiters/ConcurrencyRateLimiter.cs
mdaigle: Good spot. I'll need to think about this a little bit.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlDataSourceBuilder.cs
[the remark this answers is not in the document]
edwardneal: That makes sense. So the app context switch would essentially be the maximum "functionality version" available for use, with the default value being to enable all functionality in the SqlClient version? There's certainly prior art for this - Azure has its API versions, and Cloudflare do something similar with [compatibility dates](https://developers.cloudflare.com/workers/configuration/compatibility-dates). I don't think they enable functionality by default, but they're otherwise-versionless cloud services and SqlClient's a versioned library, which gives us that latitude; I've not got any objections to doing that.

I'm in two minds on whether the version should actually be the master, or whether we should have a collection of feature flags which can be independently enabled/disabled and enabled in bulk by setting the version. It'll probably depend on how many/which features we put behind the switch - testing those would scale factorially...

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/PoolingDataSource.cs
```diff
@@ -46,10 +48,32 @@ internal override ValueTask<SqlConnector> GetInternalConnection(SqlConnectionX o
             throw new NotImplementedException();
         }
 
+        internal struct OpenInternalConnectionState
```
edwardneal, line 51: Nit: readonly.
```suggestion
        internal readonly struct OpenInternalConnectionState
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/RateLimiters/IRateLimiter.cs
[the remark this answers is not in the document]
edwardneal: Your point around `public` makes sense, and I agree that integration with a third-party rate limiter is a little distant right now.

I'll wait to see the way the `BlockingPeriodRateLimiter` works before making any firm suggestion, but I agree with not exposing `SqlConnector`. Maybe we could make this work by instantiating a new instance of a `RateLimiterBase`-derived class which calls a third party rate limiter, passing a delegate internal to the instance of the derived class which creates a SqlConnector and assigns it to a field - something vaguely similar to the below, if we can make the generics work (or remove them.)

```csharp
public interface IResiliencyHook
{
    Task ExecuteAsync(Func<ValueTask> onConnect, SqlConnectionStringBuilder connectionString, CancellationToken cancellationToken);

    void Execute(Action onConnect, SqlConnectionStringBuilder connectionString, CancellationToken cancellationToken);
}

internal sealed class ExternalRateLimiter<TResiliencyHook> : RateLimiterBase
    where TResiliencyHook : IResiliencyHook
{
    private SqlConnector _resultantConnector;
    private TResiliencyHook _hook;

    public ExternalRateLimiter (TResiliencyHook hook)
    {
        _hook = hook;
    }

    internal override ValueTask<TResult> Execute<State, TResult>(AsyncFlagFunc<State, ValueTask<TResult>> callback, State state, bool async, CancellationToken cancellationToken = default)
    {
        if (async)
        {
            await _hook.ExecuteAsync(...);
        }
        else
        {
            _hook.Execute(...);
        }

        return _resultantConnector;
    }
}
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/PoolingDataSource.cs
```diff
@@ -30,13 +32,13 @@ internal sealed class PoolingDataSource : SqlDataSource
...
         /// </summary>
         //TODO: support auth contexts and provider info
-        PoolingDataSource(SqlConnectionStringBuilder connectionStringBuilder,
+        internal PoolingDataSource(SqlConnectionStringBuilder connectionStringBuilder,
             SqlCredential credential,
-            RemoteCertificateValidationCallback userCertificateValidationCallback,
-            Action<X509CertificateCollection> clientCertificatesCallback,
-            DbConnectionPoolGroupOptions options) : base(connectionStringBuilder, credential, userCertificateValidationCallback, clientCertificatesCallback)
+            DbConnectionPoolGroupOptions options,
+            RateLimiterBase connectionRateLimiter) : base(connectionStringBuilder, credential)
```
benrr101, line 38: Nit: Idk, I've seen some rules that say `: base ...` should go on the next line.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/RateLimiters/AsyncFlagFunc.cs
```diff
@@ -0,0 +1,20 @@
...
+    /// A function that operates asynchronously based on a flag. If isAsync is true, the function operates asynchronously.
+    /// If isAsync is false, the function operates synchronously.
+    /// </summary>
+    /// <typeparam name="State">The type accepted by the callback as input.</typeparam>
+    /// <typeparam name="TResult">The returned by the callback.</typeparam>
+    /// <param name="state">An instance of State to be passed to the callback.</param>
+    /// <param name="isAsync">Indicates whether the function should operate asynchronously.</param>
+    /// <param name="cancellationToken">Allows cancellation of the operation.</param>
+    /// <returns>Returns the result of the callback.</returns>
+    internal delegate TResult AsyncFlagFunc<in State, out TResult>(State state, bool isAsync, CancellationToken cancellationToken);
```
benrr101, line 19: Nit: `State` should be `TState`

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/RateLimiters/BlockingPeriodRateLimiter.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    /// <summary>
+    /// A rate limiter that enforces a backoff (blocking) period upon error. 
+    /// Each subsequent error increases the blocking duration, up to a maximum, until a success occurs.
+    /// </summary>
+    internal sealed class BlockingPeriodRateLimiter : RateLimiterBase
+    {
+
+        /// <inheritdoc/>
+        internal override ValueTask<TResult> Execute<State, TResult>(AsyncFlagFunc<State, ValueTask<TResult>> callback, State state, bool async, CancellationToken cancellationToken = default)
```
benrr101, line 21: Nit: for lines this long, please chunk them

ie
```
Execute<TState, TResult> (
    param1,
    param2)

File: src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/RateLimiters/ConcurrencyRateLimiter.cs (hunk 3 above)
Comment (benrr101) [reply src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/RateLimiters/ConcurrencyRateLimiter.cs:17]: It can also be made `readonly` :)

File: src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/RateLimiters/ConcurrencyRateLimiter.cs [hunk 12]
```diff
@@ -0,0 +1,61 @@
...
+            {
+                _concurrencyLimitSemaphore.Wait(cancellationToken);
+            }
+
+            try
+            {
+                cancellationToken.ThrowIfCancellationRequested();
+                if (Next != null)
+                {
+                    return await Next.Execute<State, TResult>(callback, state, async, cancellationToken).ConfigureAwait(false);
```
benrr101, line 48: Nit: In this instance, you don't need to include the generics, they can be inferred by the compiler

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/RateLimiters/PassthroughRateLimiter.cs
```diff
@@ -0,0 +1,31 @@
...
+    internal sealed class PassthroughRateLimiter : RateLimiterBase
+    {
+        //TODO: no state, add static instance
+
+        /// <inheritdoc/>
+        internal override ValueTask<TResult> Execute<State, TResult>(AsyncFlagFunc<State, ValueTask<TResult>> callback, State state, bool async, CancellationToken cancellationToken = default)
+        {
+            if (Next != null)
+            {
+                return Next.Execute<State, TResult>(callback, state, async, cancellationToken);
```
benrr101, line 23: Nit, same applies here, generics don't need to be specified

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/RateLimiters/PassthroughRateLimiter.cs
```diff
@@ -0,0 +1,31 @@
...
+{
+    /// <summary>
+    /// A no-op rate limiter that simply executes the callback or passes through to the next rate limiter.
+    /// </summary>
+    internal sealed class PassthroughRateLimiter : RateLimiterBase
+    {
+        //TODO: no state, add static instance
+
+        /// <inheritdoc/>
+        internal override ValueTask<TResult> Execute<State, TResult>(AsyncFlagFunc<State, ValueTask<TResult>> callback, State state, bool async, CancellationToken cancellationToken = default)
```
benrr101, line 19: Same comment here, long lines should be chunked

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/RateLimiters/RateLimiterBase.cs
```diff
@@ -0,0 +1,39 @@
...
+    /// An interface for rate limiters that execute arbitraty code. Intended to be small and self contained and chained together to achieve more complex behavior.
+    /// </summary>
+    internal abstract class RateLimiterBase
+    {
+        private RateLimiterBase _next;
+
+        /// <summary>
+        /// The next rate limiter that should be executed within the context of this rate limiter.
+        /// </summary>
+        internal virtual RateLimiterBase Next
```
benrr101, line 21: This can safely be converted to an auto-property (ie, no need for an explicit field backing it, just use `get; set;`)

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/RateLimiters/RateLimiterBase.cs
```diff
@@ -0,0 +1,39 @@
...
+        /// Execute the provided callback within the context of the rate limit, or pass the responsibility along to the next rate limiter.
+        /// </summary>
+        /// <typeparam name="State">The type accepted by the callback as input.</typeparam>
+        /// <typeparam name="TResult">The type of the result returned by the callback.</typeparam>
+        /// <param name="callback">The callback function to execute.</param>
+        /// <param name="state">An instance of State to be passed to the callback.</param>
+        /// <param name="async">Whether this method should run asynchronously.</param>
+        /// <param name="cancellationToken">Cancels outstanding requests.</param>
+        /// <returns>Returns the result of the callback or the next rate limiter.</returns>
+        internal abstract ValueTask<TResult> Execute<State, TResult>(AsyncFlagFunc<State, ValueTask<TResult>> callback, State state, bool async, CancellationToken cancellationToken = default);
```
benrr101, line 37: I'm wondering - using a delegate here might introduce some overhead we can avoid. I'm going to try a quick benchmark test to verify what I'm thinking, let me get back to you.

benrr101 (commented):
Mostly nit changes, but let me get back to you about the delegate stuff. I'm not convinced we can't do that with less overhead.

VladimirReshetnikov (approved):

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlDataSourceBuilder.cs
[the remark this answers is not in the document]
mdaigle: Added an issue for this: #2678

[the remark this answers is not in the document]
mdaigle: Agreed, we'll need to limit the number of flags we use and plan to phase them out over time.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/RateLimiters/IRateLimiter.cs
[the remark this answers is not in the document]
mdaigle: The BlockingPeriodRateLimiter implementation will be covered by #2677

benrr101 (approved):
Sounds pretty good to me 👍

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/RateLimiters/RateLimiterBase.cs
mdaigle: Based on @benrr101 's benchmarking, delegates are the slowest (compared to lambdas and concrete functions). I'll give this setup some thought as I implement the logic of the callback to see if this setup still makes sense or if we can switch to something more performant while still preserving the intent of the interface we expose to the rate limiters.

PR #2674, opened 2024-07-15
Update to .NET 9

### LINKED ISSUE #2669: Update to .Net 9
The SqlClient build system needs to be updated to dotnet 9 to be able to run the CI and test SqlDbType.Json added in dotnet 9.

ErikEJ:
fixes #2669

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -2,6 +2,6 @@
 <configuration>
   <packageSources>
     <clear />
-    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
+    <add key="sqlclient" value="https://sqlclientdrivers.pkgs.visualstudio.com/public/_packaging/sqlclient/nuget/v3/index.json" />
   </packageSources>
 </configuration>
--- a/NuGet.config
+++ b/NuGet.config
@@ -2,6 +2,6 @@
 <configuration>
   <packageSources>
     <clear />
-    <add key="sqlclient" value="https://sqlclientdrivers.pkgs.visualstudio.com/public/_packaging/sqlclient/nuget/v3/index.json" />
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
   </packageSources>
 </configuration>
--- a/eng/pipelines/libraries/mds-validation-variables.yml
+++ b/eng/pipelines/libraries/mds-validation-variables.yml
@@ -15,7 +15,7 @@ variables:
   - name: expectedFolderNames
     value: lib,ref,runtimes
   - name: expectedDotnetVersions
-    value: net462,net6.0,net8.0
+    value: net462,net6.0,net8.0,net9.0
   - name: Database
     value: Northwind
   - name: platform
@@ -23,7 +23,7 @@ variables:
   - name: TargetNetFxVersion
     value: net481
   - name: TargetNetCoreVersion
-    value: net8.0
+    value: net9.0
   - name: SQLTarget
     value: localhost
   - name: encrypt
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SqlNormalizer.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SqlNormalizer.cs
@@ -152,7 +152,11 @@ namespace Microsoft.Data.SqlClient.Server
                 if (nullByte == 0)
                 {
                     result = _nullInstance;
-                    s.Read(_padBuffer, 0, _padBuffer.Length);
+#if NET8_0_OR_GREATER
+                    s.ReadExactly(_padBuffer, 0, _padBuffer.Length);
+#else
+                   s.Read(_padBuffer, 0, _padBuffer.Length);
+#endif
                     return result;
                 }
             }
@@ -288,7 +292,7 @@ namespace Microsoft.Data.SqlClient.Server
         protected object GetValue(FieldInfo fi, object obj) => fi.GetValue(obj);
 #if NETFRAMEWORK
         [System.Security.Permissions.ReflectionPermission(System.Security.Permissions.SecurityAction.Assert, MemberAccess = true)]
-#endif 
+#endif
         protected void SetValue(FieldInfo fi, object recvr, object value) => fi.SetValue(recvr, value);
 
         internal abstract int Size { get; }
@@ -379,7 +383,11 @@ namespace Microsoft.Data.SqlClient.Server
         internal override void DeNormalize(FieldInfo fi, object recvr, Stream s)
         {
             byte[] b = new byte[2];
+#if NET8_0_OR_GREATER
+            s.ReadExactly(b, 0, b.Length);
+#else
             s.Read(b, 0, b.Length);
+#endif
             if (!_skipNormalize)
             {
                 b[0] ^= 0x80;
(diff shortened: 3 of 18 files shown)


NuGet.config
```diff
@@ -2,6 +2,6 @@
 <configuration>
   <packageSources>
     <clear />
-    <add key="sqlclient" value="https://sqlclientdrivers.pkgs.visualstudio.com/public/_packaging/sqlclient/nuget/v3/index.json" />
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
```
ErikEJ, line 5: Need the new packages in the sqlclient feed or a new feed

eng/pipelines/libraries/mds-validation-variables.yml
```diff
@@ -15,15 +15,15 @@ variables:
...
+    value: net462,net6.0,net8.0,net9.0
   - name: Database
     value: Northwind
   - name: platform
     value: AnyCPU
   - name: TargetNetFxVersion
     value: net481
   - name: TargetNetCoreVersion
-    value: net8.0
+    value: net9.0
```
ErikEJ, line 26: Not sure what the exact purpose of this value is

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SqlNormalizer.cs
```diff
@@ -152,7 +152,11 @@ internal sealed class BinaryOrderedUdtNormalizer : Normalizer
                 if (nullByte == 0)
                 {
                     result = _nullInstance;
-                    s.Read(_padBuffer, 0, _padBuffer.Length);
+#if NET8_0_OR_GREATER
```
ErikEJ, line 155: Fixes the new CA2022

src/Microsoft.Data.SqlClient/tests/Directory.Build.props
```diff
@@ -12,6 +12,7 @@
     <Configurations>Debug;Release;</Configurations>
     <Platforms>AnyCPU;x86;x64</Platforms>
     <ReferenceType Condition="'$(ReferenceType)'==''">Project</ReferenceType>
+    <NuGetAudit>false</NuGetAudit>
```
ErikEJ, line 15: To prevent NuGet audit on test projects giving warnings as errors on old packages

DavoudEshtehari:
FYI, the pipeline is expected to fail because lack of `.NET` 9 on running agents.

ErikEJ:
@DavoudEshtehari Yes, that was expected, but it did not even fail ??

ErikEJ:
@JRahnama Thanks, much better error this time 😄 

`The current .NET SDK does not support targeting .NET 9.0.`

saurabh500:
@cheenamalhotra @DavoudEshtehari @JRahnama @David-Engel  What do we need to do to move this along? 
Our build agents require .Net runtime 9 installed. Is the runtime supposed to be baked into the image itself? Or is there a script which downloads and installs the runtime?

JRahnama:
> @cheenamalhotra @DavoudEshtehari @JRahnama @David-Engel What do we need to do to move this along? Our build agents require .Net runtime 9 installed. Is the runtime supposed to be baked into the image itself? Or is there a script which downloads and installs the runtime?

1ES images do not have Net9.0 added. You can see included SDK in 1ES at [here](https://github.com/actions/runner-images/blob/win22/20240707.1/images/windows/Windows2022-Readme.md#net-core-tools).

DavoudEshtehari:
> @cheenamalhotra @DavoudEshtehari @JRahnama @David-Engel What do we need to do to move this along? Our build agents require .Net runtime 9 installed. Is the runtime supposed to be baked into the image itself? Or is there a script which downloads and installs the runtime?

To streamline the process and follow the pattern, a new artifact should be included in the images to reduce installation time for each run. But the compatible VS would be required though.

ErikEJ:
@JRahnama  I just use this setting in VS Options:

image

JRahnama:
> @JRahnama I just use this setting in VS Options:
> 
> image

but if you look into VS Installer you probably will see something similar as below. 
However, I will try running them with a test pipeline. We may end up getting good results with C# 13 and not having VS 2022-preview.
image

JRahnama:
@ErikEJ , hope you don't mind me pushing the yaml changes into your PR.

NuGet.config
David-Engel: The sqlclientdrivers feed upstreams from NuGet.org. If you revert this and push, I believe the CI will have permissions to pull from upstream and install the new versions to the feed.

ErikEJ: @David-Engel It did not work for me locally, but maybe things have changed?

ErikEJ: Got it! I will revert

David-Engel:
I think this will need a change in ci-prebuild-step.yml like:

```
- task: UseDotNet@2
  inputs:
    packageType: 'sdk'
    version: '9.x'
    includePreviewVersions: true
```

arellegue:
When tested on a local machine having .Net9.0 preview, Code [SYSLIB0057](https://learn.microsoft.com/en-us/dotnet/fundamentals/syslib-diagnostics/syslib0057) was thrown by the compiler for the following files:

- SNICommon,cs
- VirtualSecureModeEnclaveProviderBase.cs

JRahnama:
> When tested on a local machine having .Net9.0 preview, Code [SYSLIB0057](https://learn.microsoft.com/en-us/dotnet/fundamentals/syslib-diagnostics/syslib0057) was thrown by the compiler for the following files:
> 
> * SNICommon,cs
> * VirtualSecureModeEnclaveProviderBase.cs

Basically you need to change the code to something similar to below:
in SNICommon:

```C#
#if !NET9_0_OR_GREATER
                        validationCertificate = new X509Certificate(validationCertFileName);
#else
                        validationCertificate = X509CertificateLoader.LoadCertificateFromFile(validationCertFileName);
#endif
```

in VirtualSecureModeEnclaveProviderBase.cs:
```C#
#if !NET9_0_OR_GREATER
                    certificateCollection.Import(data);
#else
                   _ =  X509CertificateLoader.LoadCertificate(data);
#endif
```

and in VirtualSecureModeEnclaveProvider.cs:
```C#
#if !NET9_0_OR_GREATER
            Certificate = new X509Certificate2(payload);
#else
            _ = X509CertificateLoader.LoadCertificate(payload);
#endif
```

After that you will get 5 more errors such as below:
```MSBuild
SqlClient\artifacts\Project\obj\Debug.AnyCPU\Microsoft.Data.SqlClient\netcore\net9.0\Microsoft.Data.SqlClie
nt.notsupported.cs(798,37): error IL2093: 'DynamicallyAccessedMemberTypes' in 'DynamicallyAccessedMembersAttribute' on
the return value of method 'Microsoft.Data.SqlClient.SqlDataReader.GetFieldType(Int32)' don't match overridden return v
alue of method 'System.Data.IDataRecord.GetFieldType(Int32)'. All overridden members must have the same 'DynamicallyAcc
essedMembersAttribute'
```

Now we have to address those.

arellegue:
For the new CA2022 error, in addition to #if NET8_0_OR_GREATER for ReadExactly, since the compiler is throwing error for TargetFramework=net6.0, a #pragma warning disable/restore CA2022 was added for the #else block.

arellegue:
@ErikEJ, can I push minor fixes into this branch for issues found when compiling MDS locally?

JRahnama:
> For the new CA2022 error, in addition to #if NET8_0_OR_GREATER for ReadExactly, since the compiler is throwing error for TargetFramework=net6.0, a #pragma warning disable/restore CA2022 was added for the #else block.

Preferably, add them the .editorconfig file inside netcore folder

ErikEJ:
@arellegue go ahead...

arellegue:
A "dotnet_diagnostic.CA2022.severity = error" was added to .editorconfig instead since it is the preferred method.

cheenamalhotra:
Hi @ErikEJ 

I will proceed to close the PR, as PR https://github.com/dotnet/SqlClient/pull/2946 is actively being worked upon and will replace this PR once images are available with latest .NET 9 SDK

PR #2686, opened 2024-07-18
Core pooling logic

mdaigle:
This PR covers core pooling logic including SqlConnector creation, retrieval, return, and statistics. Integrates with rate limiter scaffolding added in #2667. Also covers small fixes for create, close, and dispose flows exposed by unit tests.

Not included in this PR are:
- warmup
- pruning
- clearing the pool
- data source disposal
- connection lifetime evaluation
- metrics and reporting
- exception patterns in line with the rest of the driver


edwardneal (commented):
I think this matches our discussion about pooling (pending the missing parts you've already noted.) There are a few points of code formatting/indentation/resource strings, but nothing worth a specific comment until the PR's out of draft. I'm happy with the base methodology (besides one or two nits in review comments.)

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/PoolingDataSource.cs
```diff
@@ -45,19 +83,224 @@ internal sealed class PoolingDataSource : SqlDataSource
...
+            {
+                connector.Close();
+            }
+            catch
+            {
+                //TODO: LogMessages.ExceptionWhenClosingPhysicalConnection(_logger, connector.Id, exception);
+            }
+
+            var i = 0;
+            for (; i < MaxPoolSize; i++)
```
benrr101, line 273: Please use `{}`

In my opinion, seeing `for (; ...)` invokes a 'wat' response. Though I see the reason for doing it, you want to access `i` outside of the `for` block. Here's a couple options for less 'wat' response:
* `int i; for (i = 0; ...)`
* `int i = 0; while (i < MaxPoolSize) { ... i++; }`

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/PoolingDataSource.cs
```diff
@@ -45,19 +83,224 @@ internal sealed class PoolingDataSource : SqlDataSource
...
+                if (Interlocked.CompareExchange(ref _connectors[i], null, connector) == connector)
+                    break;
+
+            // If CloseConnector is being called from within OpenNewConnector (e.g. an error happened during a connection initializer which
+            // causes the connector to Break, and therefore return the connector), then we haven't yet added the connector to Connectors.
+            // In this case, there's no state to revert here (that's all taken care of in OpenNewConnector), skip it.
+            if (i == MaxPoolSize)
```
benrr101, line 280: Please use `{}`

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/PoolingDataSource.cs
```diff
@@ -67,32 +310,110 @@ internal OpenInternalConnectionState(SqlConnectionX owningConnection, TimeSpan t
...
+                        //TODO: set clear counter on connector
+
+                        //TODO: actually open the connector
+                        //await connector.Open(timeout, async, cancellationToken).ConfigureAwait(false);
+#if NET7_0_OR_GREATER
+                        //TODO: MetricsReporter.ReportConnectionCreateTime(Stopwatch.GetElapsedTime(startTime));
+#endif
+
+                        var i = 0;
+                        for (; i < MaxPoolSize; i++)
```
benrr101, line 347: Same comment as above regarding 'wat' applies
Also please use `{}` :)

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/PoolingDataSource.cs
```diff
@@ -67,32 +310,110 @@ internal OpenInternalConnectionState(SqlConnectionX owningConnection, TimeSpan t
...
+                            if (Interlocked.CompareExchange(ref _connectors[i], connector, null) == null)
+                                break;
+
+                        Debug.Assert(i < MaxPoolSize, $"Could not find free slot in {_connectors} when opening.");
+                        if (i == MaxPoolSize)
```
benrr101, line 352: Please use `{}`

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/PoolingDataSource.cs
```diff
@@ -67,32 +310,110 @@ internal OpenInternalConnectionState(SqlConnectionX owningConnection, TimeSpan t
...
+            var written = IdleConnectorWriter.TryWrite(connector);
+            Debug.Assert(written);
         }
 
-        /// <summary>
-        /// Closes extra idle connections.
-        /// </summary>
-        /// <exception cref="NotImplementedException"></exception>
-        internal void PruneIdleConnections()
+            /// <summary>
```
benrr101, line 412: I think these indents shouldn't have been added?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlConnectionX.cs
```diff
@@ -32,8 +32,15 @@ internal sealed class SqlConnectionX : DbConnection, ICloneable
         private SqlDataSource? _dataSource;
         private SqlConnector? _internalConnection;
 
+        bool _disposed;
+
+        internal SqlConnector? InternalConnection { get => _internalConnection; }
```
benrr101, line 37: I honestly forget, is this the same as `InternalConnection => _internalConnection;`? I forget if that does set and get or if it just does get.

benrr101: Also, properties should go below constructors

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlConnectionX.cs
```diff
@@ -32,8 +32,15 @@ internal sealed class SqlConnectionX : DbConnection, ICloneable
...
+
         //TODO: Investigate if we can just use dataSource.ConnectionString. Do this when this class can resolve its own data source.
         private string _connectionString = string.Empty;
+        internal SqlConnectionStringBuilder Settings { get; private set; } = DefaultSettings;
+        static readonly SqlConnectionStringBuilder DefaultSettings = new();
```
benrr101, line 42: Readonly members should go at the top of the class, properties should go after constructors

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlConnectionX.cs
```diff
@@ -195,20 +202,40 @@ async Task CloseAsync(bool async)
...
         #endregion
 
+        /// <summary>
+        /// Releases all resources used by the <see cref="SqlConnectionX"/>.
+        /// </summary>
+        /// <param name="disposing"><see langword="true"/> when called from <see cref="Dispose"/>;
+        /// <see langword="false"/> when being called from the finalizer.</param>
+        protected override void Dispose(bool disposing)
+        {
+            if (_disposed)
```
benrr101, line 223: Please use `{}`

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlConnectionX.cs
```diff
@@ -195,20 +202,40 @@ async Task CloseAsync(bool async)
...
+                Close();
+            _disposed = true;
+        }
+
         /// <inheritdoc/>
-        public override ValueTask DisposeAsync()
-            => throw new NotImplementedException();
+        public override async ValueTask DisposeAsync()
+        {
+            if (_disposed)
```
benrr101, line 233: Please use `{}`

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlConnector.cs
```diff
@@ -34,21 +33,31 @@ internal class SqlConnector
         /// <summary>
         /// Represents the current state of this connection.
         /// </summary>
-        internal ConnectionState State => throw new NotImplementedException();
+        /// TODO: set and change state appropriately
+        internal ConnectionState State = ConnectionState.Open;
+
+        internal bool IsOpen => State == ConnectionState.Open;
```
benrr101, line 39: Please move properties to after constructors

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlDataSource.cs
```diff
@@ -49,32 +52,31 @@ protected override SqlConnectionX CreateDbConnection()
...
         {
-            return (SqlConnection)await base.OpenConnectionAsync(cancellationToken).ConfigureAwait(false);
+            return (SqlConnectionX)await base.OpenConnectionAsync(cancellationToken).ConfigureAwait(false);
         }
 
         /// <summary>
         /// Creates a new <see cref="SqlCommand"/> object.
         /// </summary>
-        public new SqlCommand CreateCommand(string commandText = null)
+        public new SqlCommand CreateCommand(string? commandText)
```
benrr101, line 79: Do we still want it to be an optional argument that defaults to `null`?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlDataSource.cs
```diff
@@ -113,7 +114,14 @@ public new SqlBatch CreateBatch()
         /// <param name="async">Whether this method should be run asynchronously.</param>
         /// <param name="cancellationToken">Cancels an outstanding asynchronous operation.</param>
         /// <returns></returns>
-        internal abstract ValueTask<SqlConnector> OpenNewInternalConnection(SqlConnectionX owningConnection, TimeSpan timeout, bool async, CancellationToken cancellationToken);
+        internal abstract ValueTask<SqlConnector?> OpenNewInternalConnection(SqlConnectionX owningConnection, TimeSpan timeout, bool async, CancellationToken cancellationToken);
+
+        private protected void CheckDisposed()
+        {
+            //TODO: implement disposal
+            if (_isDisposed == 1)
```
benrr101, line 122: Please make sure to use `{}`

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlDataSourceBuilder.cs
```diff
@@ -35,10 +35,14 @@ internal sealed class SqlDataSourceBuilder
         /// <summary>
         /// Constructs a new <see cref="SqlDataSourceBuilder" />, optionally starting out from the given <paramref name="connectionString"/>.
         /// </summary>
-        public SqlDataSourceBuilder(string connectionString = null, SqlCredential credential = null)
+        public SqlDataSourceBuilder(string connectionString = null, SqlCredential credential = null) 
+            : this(new SqlConnectionStringBuilder(connectionString), credential)
+        {}
+
+        public SqlDataSourceBuilder(SqlConnectionStringBuilder connectionStringBuilder, SqlCredential sqlCredential = null)
```
benrr101, line 42: Probably needs a comment block here

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/ThrowHelper.cs
```diff
@@ -0,0 +1,91 @@
...
+
+using System;
+using System.Diagnostics.CodeAnalysis;
+
+#nullable enable
+
+namespace Microsoft.Data.SqlClientX
+{
+
+    static class ThrowHelper
```
benrr101, line 13: I don't really understand the why behind this class. To me it seems equally unverbose to use a `throw new` versus call a method like this. 
Some of the exception classes also add `ThrowIf` static methods to them, too

src/Microsoft.Data.SqlClient/tests/UnitTests/PoolTests.cs
```diff
@@ -0,0 +1,539 @@
...
+{
+    public class PoolTests
+    {
+        private TestBase testBase;
+
+        public PoolTests()
+        {
+            testBase = new TestBase();
+        }
+
```
benrr101, line 26: nit: extra line here

src/Microsoft.Data.SqlClient/tests/UnitTests/PoolTests.cs
```diff
@@ -0,0 +1,539 @@
...
+
+        [Fact]
+        public async Task MinPoolSize_equals_MaxPoolSize()
```
benrr101, line 29: Just going to suggest my philosophy of unit test organization - organizing them into Arrange, Act, Assert is a fairly standard practice that helps make it clear what's being tested.

src/Microsoft.Data.SqlClient/tests/UnitTests/PoolTests.cs
```diff
@@ -0,0 +1,539 @@
...
+            await using var dataSource = testBase.CreateDataSource(csb =>
+            {
+                csb.MinPoolSize = 30;
+                csb.MaxPoolSize = 30;
+            });
+            await using var conn = await dataSource.OpenConnectionAsync();
+        }
+
+        [Fact]
+        public void MinPoolSize_bigger_than_MaxPoolSize_throws()
```
benrr101, line 40: Also usually naming the tests like `MethodName_OptionalCondition_Outcome` is a bit more readable
Eg, for this one I'd rename it `MinPoolSize_BiggerThanMaxPoolSize_Throws`

src/Microsoft.Data.SqlClient/tests/UnitTests/PoolTests.cs
```diff
@@ -0,0 +1,539 @@
...
+            Assert.Equal(conn.InternalConnection!.ServerProcessId, backendId);
+        }
+
+        public static readonly object[][] AsyncParams = new object[][] { 
+            new object[] { true }, 
+            new object[] { false } 
+        };
+
+        [Theory]
+        [MemberData(nameof(AsyncParams))]
```
benrr101, line 68: For situations like this (compile-time constant parameters) you can use `[InlineData(true)]` and `[InlineData(false)]` rather than set up the async params.

However, I'd personally recommend creating two different tests for async and sync. Since they're two different methods being testes, it makes sense to me to split into two tests rather than switch between the two different methods in the same test.

src/Microsoft.Data.SqlClient/tests/UnitTests/PoolTests.cs
```diff
@@ -0,0 +1,539 @@
...
+                {
+                    await conn2.OpenAsync();
+                    AssertPoolState(dataSource, expectedOpen: 1, expectedIdle: 0);
+                }
+                AssertPoolState(dataSource, expectedOpen: 1, expectedIdle: 1);
+                return Environment.CurrentManagedThreadId;
+            };
+
+            // Start an async open which will not complete as the pool is exhausted.
+            var asyncOpenerTask = asyncOpener();
```
benrr101, line 293: I'm a bit confused on this one... Would this be more simply expressed using `Task.Run`?

src/Microsoft.Data.SqlClient/tests/UnitTests/PoolTests.cs
```diff
@@ -0,0 +1,539 @@
...
+                    .With.InnerException.TypeOf<SocketException>());
+            AssertPoolState(dataSource, open: 0, idle: 0);
+        }*/
+
+        //[Test, Explicit]
+        //[TestCase(10, 10, 30, true)]
+        //[TestCase(10, 10, 30, false)]
+        //[TestCase(10, 20, 30, true)]
+        //[TestCase(10, 20, 30, false)]
+        public static object[][] PoolExerciseCases = new object[][]
```
benrr101, line 420: For this, I think you can use `[InlineData]` over the `MemberData`

benrr101 (changes requested):

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/PoolingDataSource.cs
```diff
@@ -33,6 +35,42 @@ internal sealed class PoolingDataSource : SqlDataSource
         private static int _objectTypeCount; // EventSource counter
         private readonly int _objectID = Interlocked.Increment(ref _objectTypeCount);
 
+        //TODO: readonly TimeSpan _connectionLifetime;
+
+        // Counts the total number of open connectors tracked by the pool.
+        volatile int _numConnectors;
```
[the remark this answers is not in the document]
mdaigle: _numConnectors is modifiable by multiple threads. My understanding is that we need this to prevent compiler optimizations that may affect correctness (for example, using a stale value). https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/volatile

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/PoolingDataSource.cs
```diff
@@ -33,6 +35,42 @@ internal sealed class PoolingDataSource : SqlDataSource
...
+
+        // Counts the number of connectors currently sitting idle in the pool.
+        volatile int _idleCount;
+
+        // Prevents synchronous operations from blocking on all available threads,
+        // which would stop async tasks from being scheduled and cause deadlocks.
+        private static SemaphoreSlim SyncOverAsyncSemaphore { get; } = new(Math.Max(1, Environment.ProcessorCount / 2));
```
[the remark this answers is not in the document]
mdaigle: This semaphore is used to guard a bit of sync over async code. It needs to be something between **1** and **processor count - 1** to allow async tasks to progress. But I don't know specifically why 2 was picked. I'm guessing it approximates the optimal case where each thread that is allowed to block can also have its corresponding async task running at the same time.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/PoolingDataSource.cs
```diff
@@ -33,6 +35,42 @@ internal sealed class PoolingDataSource : SqlDataSource
...
+
+        /// <summary>
+        /// Tracks all connectors currently managed by this pool, whether idle or busy.
+        /// Only updated rarely - when physical connections are opened/closed - but is read in perf-sensitive contexts.
+        /// </summary>
+        private readonly SqlConnector?[] _connectors;
```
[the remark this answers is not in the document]
mdaigle: Individual array elements may be set to null as they are closed and removed from the pool.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/PoolingDataSource.cs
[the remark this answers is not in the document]
edwardneal: The relevant PR from Npgsql is npgsql/npgsql#5196 - it looks like it was a ballpark estimate which never needed to be adjusted.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/PoolingDataSource.cs
```diff
@@ -45,19 +83,224 @@ internal sealed class PoolingDataSource : SqlDataSource
...
         internal override ValueTask<SqlConnector> GetInternalConnection(SqlConnectionX owningConnection, TimeSpan timeout, bool async, CancellationToken cancellationToken)
         {
-            throw new NotImplementedException();
+            CheckDisposed();
+
+            return TryGetIdleConnector(out SqlConnector? connector)
+                ? ValueTask.FromResult(connector)
+                : RentAsync(owningConnection, timeout, async, cancellationToken);
+
+            async ValueTask<SqlConnector> RentAsync(
```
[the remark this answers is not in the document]
mdaigle: I think I'll just remove it. It doesn't provide much value to have it broken out and the name is confusing.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/PoolingDataSource.cs
```diff
@@ -45,19 +83,224 @@ internal sealed class PoolingDataSource : SqlDataSource
...
+                }
+            }
         }
 
+        /// <summary>
+        /// Tries to read a connector from the idle connector channel.
+        /// </summary>
+        /// <param name="connector">Out parameter to store the connector result.</param>
+        /// <returns>Returns true if a valid idles connector is found, otherwise returns false.</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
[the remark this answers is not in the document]
mdaigle: I'll add a todo to profile this. Won't really know if it's helping or hurting without that.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlConnectionX.cs
mdaigle: Looks like it's equivalent: https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/using-properties#the-get-accessor

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/PoolingDataSource.cs
```diff
@@ -33,6 +35,42 @@ internal sealed class PoolingDataSource : SqlDataSource
...
+
+
+        /// <summary>
+        /// Reader side for the idle connector channel. Contains nulls in order to release waiting attempts after
+        /// a connector has been physically closed/broken.
+        /// </summary>
+        readonly ChannelReader<SqlConnector?> _idleConnectorReader;
+        internal ChannelWriter<SqlConnector?> IdleConnectorWriter { get; }
+
+        internal sealed override (int Total, int Idle, int Busy) Statistics
```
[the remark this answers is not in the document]
edwardneal: I agree, and it also broaches the topic of OTel metrics support (#2211.) We need to be able to summarise multiple PoolingDataSource statistics into a single set of "per-connection-string" metric values, so these statistics are needed.

Small nit: I'd prefer to read both the total and idle connection figures atomically, since they're so closely related. If both fields are in a class, we can use `FieldOffset` to position them at 0 and 4 bytes, have a `ulong` positioned at 0 bytes, then use `Interlocked.Read` on the `ulong` to perform an atomic read of both values at once.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/ThrowHelper.cs
mdaigle: Forgot to clean this one up. I don't think we have a need for it.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/PoolingDataSource.cs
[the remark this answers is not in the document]
benrr101: Hmm, I see, so it mostly works the same as it would in the realm I'm familiar with it being used in. My follow up question is whether this is better, worse, or the same as accessing it via Interlocked methods?

But also - we should still add `private` to it.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/PoolingDataSource.cs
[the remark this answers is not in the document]
benrr101: Ok, makes sense. And there isn't a warning about not assigning `_connectors` at construction (I swear I saw warnings before from stuff being set in constructor only, even though that doesn't make a ton of sense to me).

benrr101 (approved):
A few low-tier commends, but nothing that I'd hold up checkin in on. I think.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/PoolingDataSource.cs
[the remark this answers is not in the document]
mdaigle: Done!

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/PoolingDataSource.cs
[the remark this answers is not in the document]
mdaigle: Let me capture this in a new issue. I think this deserves focus but I don't want it to block this PR at the moment.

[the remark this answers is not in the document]
mdaigle: #2708

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/PoolingDataSource.cs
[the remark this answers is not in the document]
saurabh500: I am curious about the guidance used here for Aggressive Inlining.

Tagging @roji to see if he has any comments. How do you decide to use `AggressiveInlining` annotation on the methods?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/PoolingDataSource.cs
```diff
@@ -45,19 +90,246 @@ internal sealed class PoolingDataSource : SqlDataSource
...
         }
 
         /// <inheritdoc/>
-        internal override ValueTask<SqlConnector> GetInternalConnection(SqlConnectionX owningConnection, TimeSpan timeout, bool async, CancellationToken cancellationToken)
+        internal override async ValueTask<SqlConnector> GetInternalConnection(SqlConnectionX owningConnection, TimeSpan timeout, bool async, CancellationToken cancellationToken)
         {
-            throw new NotImplementedException();
+            CheckDisposed();
+
+            if (TryGetIdleConnector(out SqlConnector? connector))
```
saurabh500, line 110: I am curious, why is TryGet a better pattern than using `SqlConnector? connector = GetIdleConnector() ` ?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/PoolingDataSource.cs
```diff
@@ -45,19 +90,246 @@ internal sealed class PoolingDataSource : SqlDataSource
...
+            }
+
+            // We're at max capacity. Block on the idle channel with a timeout.
+            // Note that Channels guarantee fair FIFO behavior to callers of ReadAsync (first-come first-
+            // served), which is crucial to us.
+            using CancellationTokenSource linkedSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+            CancellationToken finalToken = linkedSource.Token;
+            linkedSource.CancelAfter(timeout);
+            //TODO: respect remaining time, linkedSource.CancelAfter(timeout.CheckAndGetTimeLeft());
+            //TODO: MetricsReporter.ReportPendingConnectionRequestStart();
```
saurabh500, line 125: What is `MetricsReporter` ?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/PoolingDataSource.cs
```diff
@@ -45,19 +90,246 @@ internal sealed class PoolingDataSource : SqlDataSource
...
+                            return connector;
+                        }
+                    }
+                    catch (OperationCanceledException)
+                    {
+                        cancellationToken.ThrowIfCancellationRequested();
+                        Debug.Assert(finalToken.IsCancellationRequested);
+
+                        //TODO: MetricsReporter.ReportConnectionPoolTimeout();
+                        /*TODO: throw new NpgsqlException(
```
saurabh500, line 168: Remove references to NpgSqlException from comments.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/PoolingDataSource.cs
saurabh500: I believe this may have been copied over from NpgSql driver. Clean up the commented out code to say something to the effect that this needs to be plumbed into the existing connection pool counters in SqlClient.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlConnectionX.cs
```diff
@@ -139,6 +146,10 @@ public override bool CanCreateBatch
         protected override DbProviderFactory? DbProviderFactory
             => throw new NotImplementedException();
 
+        internal SqlConnector? InternalConnection => _internalConnection;
+
+        internal SqlConnectionStringBuilder Settings { get; private set; } = DefaultSettings;
```
saurabh500, line 151: Use the `SqlConnectionString` object not the builder.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/PoolingDataSource.cs
mdaigle: I actually do prefer that. It brings the signature in line with the other methods.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlConnectionX.cs
saurabh500: The `SqlConnectionString` can be created from the `string connectionString`

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlConnectionX.cs
```diff
@@ -195,20 +206,47 @@ async Task CloseAsync(bool async)
...
+        public override async ValueTask DisposeAsync()
+        {
+            if (_disposed)
+            {
+                return;
+            }
 
+            await CloseAsync().ConfigureAwait(false);
+            _disposed = true;
+        }
```
saurabh500, line 249: Nit: New line after }

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/ThrowHelper.cs
```diff
@@ -0,0 +1,91 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Diagnostics.CodeAnalysis;
+
+#nullable enable
+
+namespace Microsoft.Data.SqlClientX
```
saurabh500, line 10: Doesn't ADP aka AdapterUtil class already provide these convenience throw helpers ?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlConnector.cs
```diff
@@ -19,6 +18,16 @@ namespace Microsoft.Data.SqlClientX
     /// </summary>
     internal class SqlConnector
     {
+        private static int SpoofedServerProcessId = 1;
+
+        internal SqlConnector(SqlConnectionX owningConnection, SqlDataSource dataSource)
+        {
+            OwningConnection = owningConnection;
+            DataSource = dataSource;
+            ServerProcessId = Interlocked.Increment(ref SpoofedServerProcessId);
```
saurabh500, line 27: Why do we need to increment the Server process Id in client code ?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/PoolingDataSource.cs
```diff
@@ -67,25 +336,111 @@ internal OpenInternalConnectionState(SqlConnectionX owningConnection, TimeSpan t
...
+                    {
+                        // We've managed to increase the open counter, open a physical connections.
+#if NET7_0_OR_GREATER
+                        var startTime = Stopwatch.GetTimestamp();
+#endif
+                        SqlConnector? connector = new SqlConnector(state._owningConnection, this);
+                        //TODO: set clear counter on connector
+
+                        //TODO: actually open the connector
+                        //await connector.Open(timeout, async, cancellationToken).ConfigureAwait(false);
```
saurabh500, line 369: Should we not uncomment this, and add a `Task.Delay` in the connector.Open to simulate a delay ? I see your tests could benefit from this, instead of adding a sleep in your tests.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/PoolingDataSource.cs
mdaigle: cleaned up

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlConnector.cs
mdaigle: This is just to simulate different server process ids. I added a TODO to clarify and to set it appropriately based on the info reported back during login.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/ThrowHelper.cs
mdaigle: yes, deleted this file. Ben had a similar comment.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/PoolingDataSource.cs
mdaigle: I pushed a change to try this out, let's see how it goes in the pipeline 🤞

mdaigle: This worked

PR #2690, opened 2024-07-20
Change internal read functions to return TdsOperationResult

Wraith2:
Split out from https://github.com/dotnet/SqlClient/pull/2608 per discussion detailed in https://github.com/dotnet/SqlClient/pull/2608#issuecomment-2239803005

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
@@ -24,6 +24,13 @@ namespace Microsoft.Data.SqlClient
         internal long _value;
     }
 
+    internal enum TdsOperationStatus : int
+    {
+        Done = 0,
+        NeedMoreData = 1,
+        InvalidData = 2
+    }
+
     partial class TdsParserStateObject
     {
         private static int s_objectTypeCount; // EventSource counter
@@ -466,7 +473,7 @@ namespace Microsoft.Data.SqlClient
             return remaining;
         }
 
-        internal bool TryStartNewRow(bool isNullCompressed, int nullBitmapColumnsCount = 0)
+        internal TdsOperationStatus TryStartNewRow(bool isNullCompressed, int nullBitmapColumnsCount = 0)
         {
             Debug.Assert(!isNullCompressed || nullBitmapColumnsCount > 0, "Null-Compressed row requires columns count");
 
@@ -478,9 +485,10 @@ namespace Microsoft.Data.SqlClient
                 // assert that NBCROW is not in use by 2005 or before
                 Debug.Assert(_parser.Is2008OrNewer, "NBCROW is sent by pre-2008 server");
 
-                if (!_nullBitmapInfo.TryInitialize(this, nullBitmapColumnsCount))
+                TdsOperationStatus result = _nullBitmapInfo.TryInitialize(this, nullBitmapColumnsCount);
+                if (result != TdsOperationStatus.Done)
                 {
-                    return false;
+                    return result;
                 }
             }
             else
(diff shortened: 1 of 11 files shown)


saurabh500:
@Wraith2 please add me, @David-Engel  and @cheenamalhotra  as reviewers once you publish this PR and move it out of the draft stage.

Wraith2:
Ready for review @saurabh500 @David-Engel @cheenamalhotra

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
```diff
@@ -1263,15 +1287,17 @@ internal bool TryReadByte(out byte value)
                 {
                     _networkPacketTaskSource.TrySetResult(null);
                 }
-                return false;
+                return TdsOperationStatus.InvalidData;
+                ;
```
saurabh500, line 1291: Nit: Looks like an unintentional `;`

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
```diff
@@ -1985,12 +2032,13 @@ internal bool TryReadNetworkPacket()
 
                 // previous buffer is in snapshot
                 _inBuff = new byte[_inBuff.Length];
+                result = TdsOperationStatus.NeedMoreData;
```
saurabh500, line 2035: Is this unintentional? `result` is being assigned  `NeedMoreData` which will be treated as `!= Done` by the callers of this function, which would lead to an exception, which shouldn't be the case.

saurabh500 (changes requested):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
saurabh500: I re-read code. It looks like we return `false` when we have the previous buffer in snapshot. 
Hence your change of setting result to `TdsOperationStatus.NeedMoreData` should not cause any impact from what the code is doing right now. This could be fine. 
My previous comment about "an exception being thrown" is incorrect. 

However for the spirit of this PR, can you remove this line, "2035" or comment it out, so that this looks similar to what the current code is doing, except replace of bool with tri-state enum?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
Wraith2: Removed.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
Wraith2: Removed.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
```diff
@@ -1383,7 +1383,7 @@ internal TdsOperationStatus TryReadInt32(out int value)
             {
                 // If the int isn't fully in the buffer, or if it isn't fully in the packet,
                 // then use ReadByteArray since the logic is there to take care of that.
-                TdsOperationStatus result = TryReadByteArray(buffer, 2);
+                TdsOperationStatus result = TryReadByteArray(buffer, 4);
```
saurabh500, line 1385: Sometimes I wonder how this has been working fine for so long ! 😄 
Thanks for catching and fixing this.

saurabh500 (approved):

saurabh500:
LGTM ! @David-Engel  and @cheenamalhotra . No changes to the code in this PR. This brings in the "noisy" but important part of the Multiplexer PR. 
Please help move this along.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
Wraith2: This was a mistake by me when re-implementing the changes. It's correct in main. It caused some tests to fail so i could track it down and fix it.

cheenamalhotra (approved):

David-Engel (changes requested):
Just found one issue that needs to be addressed. Really appreciate the work you put in here. Thank you!

David-Engel (approved):
LGTM. Thanks, again!

Wraith2:
Excellent. I'll get started on part 2

PR #2715, opened 2024-07-25
Read Support for JSON column

apoorvdeshmukh:
**Summary:**
This commit contains changes necessary for reading a JSON column using `SqlDataReader.GetString(i)`.
Tests will be added for read and write when we finalize our approach for testing this against custom build of SqlServer that has JSON support.


apoorvdeshmukh:

PR #2722, opened 2024-07-26
Implement read and write functionality for JSON col

apoorvdeshmukh:
**Summary:**
This commit contains changes necessary for reading from and writing to a JSON column.
Tests will be added for read and write when we finalize our approach for testing this against custom build of SqlServer that has JSON support.


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -9905,7 +9908,7 @@ private Task TDSExecuteRPCAddParameter(TdsParserStateObject stateObj, SqlParamet
                     stateObj.WriteByte(0);       // No schema
                 }
             }
-            else if (mt.IsCharType)
```
saurabh500, line 9908: This is needed for data writes, isnt it ?

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlJson.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.Data.SqlClient
+{
+    internal class SqlJson
```
cheenamalhotra, line 7: Can this file be moved to src (shared) folder? Look for files in path: `src\Microsoft.Data.SqlClient\src\...`

cheenamalhotra, line 7: Can you add TODO comments or something that sets the expectation of what's going to come in this type?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
apoorvdeshmukh: Yes. I included this change as it was needed as a fix for reads to work on top of write changes.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlJson.cs
apoorvdeshmukh: Will add it later under SqlTypes as part of implmenting SqlJson class. I have removed it for now.

apoorvdeshmukh: File removed. To be added later as part of SqlJson implementation.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -13652,7 +13674,7 @@ internal void WriteParameterVarLen(MetaType type, int size, bool isNull, TdsPars
                         WriteInt(unchecked((int)TdsEnums.VARLONGNULL), stateObj);
                     }
                 }
-                else if (type.NullableType == TdsEnums.SQLXMLTYPE || unknownLength)
+                else if (type.NullableType is TdsEnums.SQLXMLTYPE or TdsEnums.SQLJSON || unknownLength)
```
cheenamalhotra, line 13677:
```suggestion
                else if (type.NullableType is TdsEnums.SQLXMLTYPE or TdsEnums.SQLJSON or unknownLength)
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -12662,7 +12684,7 @@ internal void WriteParameterVarLen(MetaType type, int size, bool isNull, TdsPars
                         WriteInt(unchecked((int)TdsEnums.VARLONGNULL), stateObj);
                     }
                 }
-                else if (type.NullableType == TdsEnums.SQLXMLTYPE || unknownLength)
+                else if (type.NullableType is TdsEnums.SQLXMLTYPE or TdsEnums.SQLJSON || unknownLength)
```
cheenamalhotra, line 12687:
```suggestion
                else if (type.NullableType is TdsEnums.SQLXMLTYPE or TdsEnums.SQLJSON or unknownLength)
```

cheenamalhotra (approved):
Just some fancy comments.. overall LGTM!

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -11485,10 +11491,18 @@ private Task WriteUnterminatedSqlValue(object value, MetaType type, int actualLe
...
                 case TdsEnums.SQLNTEXT:
                 case TdsEnums.SQLXMLTYPE:
+                case TdsEnums.SQLJSON:
 
                     if (type.IsPlp)
                     {
-                        if (IsBOMNeeded(type, value))
+                        if (type.NullableType == TdsEnums.SQLJSON)
+                        {
+                            // TODO : Performance and BOM check. Saurabh
```
Wraith2, line 11500: What is the work that is needed under the TODO?

saurabh500: @Wraith2  
Problem: A string, small or large, is being converted to a byte array causing allocation. 
That byte array is then written to network. 

2 ways of fixing this:
1. Send this out as Plp chunks. Say some chunk, smaller than the packetsize, write the Plp chunk length, write the string bytes, rinse and repeat till all the bytes can be written.
2. Do not allocate the byte[] to send out the string. Get the length, write it, and then based on the packet size, write out the UTF8 encoded bytes of the string using the Encoding/Text buffer APIs, so that the string is directly translated to the _outBuff, instead of the intermediate allocation. 

These are the solutions in my mind. However I have not looked at the encoding / text buffer APIs in detail. 

Any insights? This is the problem I had discussed on the call with you as well.

Wraith2:
At some point, perhaps not in this PR, you will need to update the conversion matrix used by smi and make sure that this warning in the code is satisfied:
https://github.com/dotnet/SqlClient/blob/0c4519871c9b69f7e05fe90f59d224d5f9ab3e3e/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/ValueUtilsSmi.cs#L2698-L2704

The CI runs on Release builds so the CI will not save you from violating assertions in this project.

You will probably need the sql server team to define the extensions to the conversion operations tables that start at https://github.com/dotnet/SqlClient/blob/0c4519871c9b69f7e05fe90f59d224d5f9ab3e3e/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/ValueUtilsSmi.cs#L2920
I suspect that json to binary might be the only conversion but someone needs to make that decision.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
Wraith2: Definitely option 2.
Getting the length to write it and then getting the bytes again is a bit wasteful but it's better than assuming we can buffer the input string and the bytes in memory at the same time. Consider a containerized scenario which has low memory working with large strings of json.

If we're working in netcore then we probably want to try and use https://learn.microsoft.com/en-gb/dotnet/api/system.text.unicode.utf8.fromutf16?view=net-8.0 which will allow us to use a ROS<char> from the string to get incremental chunks of bytes from it and write to the packets directly.

If we're only using ns2.0 level apis so we have netfx compatibility then sadly we're stuck with a UTF8 Encoder instance https://learn.microsoft.com/en-gb/dotnet/api/system.text.encoder.getbytes?view=net-8.0 which can be a little complicated to work with correctly.

roji: > Definitely option 2.
> Getting the length to write it and then getting the bytes again is a bit wasteful but it's better than assuming we can buffer the input string and the bytes in memory at the same time. Consider a containerized scenario which has low memory working with large strings of json.

I'm probably missing context/TDS knowledge here, but PLP seems ideal for large string data, no? You can have a single byte[] buffer, and repeatedly use Encoder.GetBytes (or Utf8.FromUtf16) to write out a chunk of the string into the buffer, prefixing it with the chunk length and sending that chunk out, no?

PLP is one thing in TDS which (if I understand it correctly) is missing in the PG protocol: everything has to be fully length-prefixed, so when sending a long string, Npgsql has to either do a full pass just to calculate the exact byte length in advance, or continuously allocate more memory as we're encoding, which means arbitrarily-large memory requirements (not good). The ability to send a huge value in multiple chunks - each length-prefixed and without having to prefix the whole thing with a length - seems very useful here.

Wraith2: Yes, PLP is really useful for string data but the packetization requirements of tds also make it a but tricky. Any UTF16 codepoint can encode to 1..3 output bytes and typically if a 3 byte output is encountered and doesn't fit entirely in the output buffer it isn't output and the function would return only complete encodings. This is a problem because TDS packets must be filled to the last byte so you need to be able to get partial utf sequences.

I was talking to @GrabYourPitchforks about this on discord yesterday because I wanted to make sure my recommendations were correct. The best idea that was suggested was to use the Convert function and pass it an output buffer specifically of (space in packet + 3) bytes, this ensures that if any multibyte sequences are present they will be fully present in the output and we can then handle taking the bytes we need and temporarily storing the remaining parts of the encoded output until the next packet.

roji: Thanks for the context @Wraith2, that helps! But doesn't Encoder.GetBytes() exist exactly for this kind of thing? In other words, if there are only two bytes left in the packet buffer and the next character needs 3, wouldn't it populate 2 bytes and allow you to write the 3rd byte in the next one, after having sent the packet?

Wraith2: Yes. That's why I suggested using an Encoder instance. It was pointed out to me that using Convert may not need an instance though which saves an allocation.

roji: Makes sense, thanks. Worst case, keeping an Encoder instance per physical connection and reusing it should also be fine (as I'm assuming no need for concurrent writing).

saurabh500: Looks like we have a path forward.

@apoorvdeshmukh can you open a followup issue to tackle this performance improvement and add the link in the comment. We can tackle it in the next wave of change for JSON.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsEnums.cs
```diff
@@ -490,6 +490,9 @@ public enum ActiveDirectoryWorkflow : byte
         public const int SQLDATETIME2 = 0x2a;
         public const int SQLDATETIMEOFFSET = 0x2b;
 
+        // New datatypes
```
saurabh500, line 493: Remove the comment which says `//New datatypes`

saurabh500 (approved):
Approved with comments.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
apoorvdeshmukh: Raised issue #2732 to track this.

PR #2723, opened 2024-07-26
Fix the prelogin handler inheritance and remove the Error object

saurabh500:
The changes do the following:

1. Change the base type of Prelogin handler to ContextHandler.
2. Add a test to construct the chain of handlers, so that any base class changes dont break the sub-handler.
3. Removed the error from the context based on the earlier discussions where we will throw exception rather than saving it. 
4. Added test to create a chain of handlers to test this behavior to make sure that the chain works. In this test the prelogin handler is not being tested since there is no server to test with ATM.

I will be sending LoginHandler based on these changes, but these changes make sure that the code before prelogin handler is working OK.

### DIFF
--- a/src/Microsoft.Data.SqlClient/tests/UnitTests/Handlers/LoginChainTest.cs
+++ b/src/Microsoft.Data.SqlClient/tests/UnitTests/Handlers/LoginChainTest.cs
@@ -1,0 +1,54 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net.Sockets;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Data.SqlClientX.Handlers.Connection;
+using Xunit;
+
+namespace Microsoft.Data.SqlClient.NetCore.UnitTests.Handlers
+{
+    /// <summary>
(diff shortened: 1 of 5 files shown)


src/Microsoft.Data.SqlClient/tests/UnitTests/Handlers/LoginChainTest.cs
```diff
@@ -0,0 +1,54 @@
...
+    ///  without the orchestrator.
+    /// </summary>
+    public sealed class LoginChainTest
+    {
+
+        /// <summary>
+        ///  Datasource to use for connection test. This may be changed locally to test against a different server.
+        ///  TODO: Migrate to config when we enable CI.
+        /// </summary>
+        private string _dataSource = "tcp:localhost,1446";
```
cheenamalhotra, line 28: Is it required to be a working host or this could be any random server name/port combination?

cheenamalhotra (approved):

src/Microsoft.Data.SqlClient/tests/UnitTests/Handlers/LoginChainTest.cs
saurabh500: Expected to be random right now. The test is expecting a socket exception on this endpoint.

saurabh500: I intend to evolve this test into a valid e2e test. Since we don't have the infrastructure and the login is not complete, I have expected a socket exception. But when we light up e2e then this test moves into integration tests project and this will come from config.

PR #2724, opened 2024-07-26
Minor Touch-ups + consistent naming

cheenamalhotra:


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/InternalConnectionContext.cs
```diff
@@ -1,19 +0,0 @@
...
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
-
-namespace Microsoft.Data.SqlClientX.Handlers
-{
-    /// <summary>
-    /// Placeholder context class needed for SqlConnector
-    /// </summary>
-    internal class InternalConnectionContext
```
cheenamalhotra, line 15: Removing this context for now, as I will create one for TdsParser as needed.

saurabh500:
Can you please hold off the name of type changes like name of PreloginHandler casing. I want to get the login handler and dealing with merge conflicts is for these name changes can be avoided till some more code goes in. 
I am ok with comment changes and removal of async keyword

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Connection/ConnectionHandlerOrchestrator.cs
```diff
@@ -0,0 +1,32 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Data.SqlClientX.Handlers.Connection
+{
+    internal static class ConnectionHandlerOrchestrator
```
cheenamalhotra, line 10: Renamed to Connection Handler orchestrator, and made static. I am looking forward to SqlConnector to own the ConnHandlerContext and pass on to Orchestrator to process connection request.

cheenamalhotra:
> Can you please hold off the name of type changes like name of PreloginHandler casing. I want to get the login handler and dealing with merge conflicts is for these name changes can be avoided till some more code goes in. I am ok with comment changes and removal of async keyword

Sure, I can park it for later

saurabh500:
@cheenamalhotra  according to https://learn.microsoft.com/en-us/openspecs/sql_server_protocols/ms-sstds/75e62f67-f057-4d46-82b3-6920fe0ebada prelogin is a keyword. So it should continue to be Prelogin instead of PreLogin

cheenamalhotra:
It looks like we have some PreLogin terms in use as well.. I'm fine with either way, as long as it's consistent and Visual Studio can accept the spelling :P

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Connection/ConnectionHandlerContext.cs
```diff
@@ -58,12 +59,24 @@ internal class ConnectionHandlerContext : HandlerRequest, ICloneable
...
         /// <summary>
         /// Indicates if fed auth needed for this connection.
         /// </summary>
-        public bool FedAuthRequired { get; internal set; }
+        public bool IsFedAuthRequired { get; internal set; }
+
+        /// <summary>
+        /// Get if data classification is enabled by the server.
+        /// </summary>
+        public bool IsDataClassificationEnabled =>
```
saurabh500, line 72: DataClassification enabled is 2 things
1. That the feature extension is acknowledged,
2. The FE is acknowledged with the valid version.
This change only brings in the version check. i.e. point 2. Point 1 needs to be accounted for as well, unless you mean that invalid version is also set for servers which dont acknowledge the FE. 

Typically the 2 checks above must be performed. The client will send the FE bit with the data signifying the highest Version it supports. Hence the server will always respond with the Min of the highest version it supports and the client's requested version.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Connection/ConnectionHandlerContext.cs
```diff
@@ -58,12 +59,24 @@ internal class ConnectionHandlerContext : HandlerRequest, ICloneable
...
+                (DataClassificationVersion != TdsEnums.DATA_CLASSIFICATION_NOT_ENABLED);
+
+        /// <summary>
+        /// Get or set data classification version.
+        /// A value of 0 means that sensitivity classification is not enabled.
+        /// </summary>
+        public int DataClassificationVersion { get; set; }
```
saurabh500, line 79: I am hoping that we have a type called `ConnectionFeatures` in the context, and the ConnectionFeatures type has the details of each of the FE. Each of the feature would contain 
1. Whether it was negotiated.
2. Whether the server acknowledged it
3. The version (if any) of the feature acknowledged by the server.
This way the responsibilities of all the features end up in a different object, instead of rooting them at the connection handler context.

saurabh500: I have something like this in mind https://github.com/saurabh500/SqlClient/blob/dev/loginhandler2/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Connection/Login/FeatureExtensions.cs

This will come in with LoginHandler

PR #2726, opened 2024-07-26
Replace unused out arguments with discards

benrr101:
**Description**: C# 7.0 added support for discards, such that if an out variable is not used after the call to a method that sets it, it can be replaced with `_` to discard to value. There are around 90 instances of this in the codebase, and they are very easy to remove. As far as I can tell this makes no semantic difference to the code, but does clean up the code a bit and bring it up to a more modern standard.

The only real argument I could see against making this quick cleanup change is that it reduces the understanding of what the code aims to do. I feel this change is fairly neutral, but please feel free to disagree as necessary.

**Testing**: Projects still build so it should be good 👍

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Common/src/Interop/Windows/sspicli/SSPIWrapper.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Common/src/Interop/Windows/sspicli/SSPIWrapper.cs
@@ -406,7 +406,7 @@ namespace System.Net
 
         public static object QueryContextAttributes(SSPIInterface secModule, SafeDeleteContext securityContext, Interop.SspiCli.ContextAttribute contextAttribute)
         {
-            return QueryContextAttributes(secModule, securityContext, contextAttribute, out int _);
+            return QueryContextAttributes(secModule, securityContext, contextAttribute, out _);
         }
 
         public static object QueryContextAttributes(SSPIInterface secModule, SafeDeleteContext securityContext, Interop.SspiCli.ContextAttribute contextAttribute, out int errorCode)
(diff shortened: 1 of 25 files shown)


David-Engel (approved):

src/Microsoft.Data.SqlClient/netcore/src/Common/src/Interop/Windows/sspicli/SSPIWrapper.cs
```diff
@@ -406,8 +406,7 @@ public static SafeFreeContextBufferChannelBinding QueryContextChannelBinding(SSP
 
         public static object QueryContextAttributes(SSPIInterface secModule, SafeDeleteContext securityContext, Interop.SspiCli.ContextAttribute contextAttribute)
         {
-            int errorCode;
-            return QueryContextAttributes(secModule, securityContext, contextAttribute, out errorCode);
+            return QueryContextAttributes(secModule, securityContext, contextAttribute, out int _);
```
cheenamalhotra, line 409: Can you not avoid types?

```suggestion
            return QueryContextAttributes(secModule, securityContext, contextAttribute, out _);
```

roji, line 409: Should also be able to leave out the type when doing discards:

```suggestion
            return QueryContextAttributes(secModule, securityContext, contextAttribute, out _);
```

DavoudEshtehari (approved):
@benrr101 Thank you for cleaning them up. I believe discarding the redundant arguments makes the code clearer and easier to follow by eliminating distractions.😉

src/Microsoft.Data.SqlClient/netcore/src/Common/src/Interop/Windows/sspicli/SSPIWrapper.cs
benrr101: This is true. I don't know why the tool I was using preferred to add the type when it wasn't necessary 🤔

benrr101: Yep, I agree. The tool I was using put the types in for some reason.

benrr101:
@roji @cheenamalhotra I did a sweep of all the typed discards and removed all the ones I could. This likely touched some files that were untouched before this, but it's verified at compilation. There was one locations that couldn't be changed and that was due to it being ambiguous call if the type was omitted.

PR #2727, opened 2024-07-26
Fix IP Comparison

saurabh500:
When comparing IpV6 and IPV4 based on IP preferences, the first IP address that should be used is the one that matches the preference.

This means that the Version Comparer should expect the Matching Address family to be higher in precedence, or lower in sort order.

This change brings in the fix and adjusts the tests to honor this fix.


cheenamalhotra (approved):

benrr101:
😑

benrr101 (approved):

PR #2728, opened 2024-07-26
[5.1] Fix | Dotnet build command

dauinsight:


paulmedynski (approved):

benrr101 (approved):
🤦‍♂️ For pete's sake... We let this sit for the better part of a year and it was a 1 character change? I'm more surprised why this wasn't a showstopper bug...

mdaigle (approved):

cheenamalhotra (approved):

PR #2730, opened 2024-07-29
Login handler for login

saurabh500:
The PR brings in a login handler.

1. LoginHandler has the behavior for constructing and sending the login packet.
2. The LoginHandlerContext is a POCO which provides all the information needed for login. It takes the information needed from ConnectionHandlerContext, and then creates new properties for the handler to work on.

Testing?
1. I have an E2E test, which can be run by pointing it to a SQL server. After the data is sent to the server, I inspected if the server logs contain any information about a malformed packet. The login is working fine with SQL Auth. The test demonstrates how the testing can be done. However the validation is manual. 

2. The small writes need to be tackled more effectively. I have added a TODO and we need a consensus on the strategy. Right now the intention is to have a working login path, so that the functionality can be added on top of this.


saurabh500:
cc @edwardneal

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Connection/LoginHandler.cs
```diff
@@ -0,0 +1,636 @@
...
+                features.FedAuthFeatureExtensionData =
+                    new FederatedAuthenticationFeatureExtensionData
+                    {
+                        libraryType = TdsEnums.FedAuthLibrary.MSAL,
+                        authentication = context.ConnectionOptions.Authentication,
+                        fedAuthRequiredPreLoginResponse = context.FedAuthNegotiatedInPrelogin
+                    };
+            }
+
+            if (context.AccessTokenInBytes != null)
```
edwardneal, line 83: Style point here: I'd integrate `context.AccessTokenInBytes` into the return value of `ShouldRequestFedAuth`, then just set `FedAuthFeatureExtensionData` differently based on `context.AccessTokenInBytes`. I'd also consider having two different constructors for `FederatedAuthenticationFeatureExtensionData`.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Connection/LoginHandler.cs
```diff
@@ -0,0 +1,636 @@
...
+                await writer.WriteIntAsync(0, isAsync, ct).ConfigureAwait(false);  // ClientTimeZone is not used
+                await writer.WriteIntAsync(0, isAsync, ct).ConfigureAwait(false);  // LCID is unused by server
+
+                // Start writing offset and length of variable length portions
+                int offset = TdsEnums.SQL2005_LOG_REC_FIXED_LEN;
+
+                // write offset/length pairs
+
+                // note that you must always set ibHostName since it indicates the beginning of the variable length section of the login record
+                await writer.WriteShortAsync(offset, isAsync, ct).ConfigureAwait(false); // host name offset
```
edwardneal, line 329: I'd be surprised if this made a difference in the real world, but many of these offsets are flagged as unsigned shorts rather than signed ones.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Connection/LoginHandler.cs
```diff
@@ -0,0 +1,636 @@
...
+                                        SessionData recoverySessionData,
+                                        FederatedAuthenticationFeatureExtensionData fedAuthFeatureExtensionData,
+                                        bool isAsync,
+                                        CancellationToken ct)
+        {
+            if (context.UseFeatureExt)
+            {
+                if (_sessionRecoveryFeature.ShouldUseFeature(requestedFeatures))
+                {
+                    await _sessionRecoveryFeature.WriteFeatureData(context, isAsync, ct).ConfigureAwait(false);
```
cheenamalhotra, line 529: Since you're already passing context, can the WriteFeatureData not determine itself whether this feature was requested? I wonder if ShouldUseFeature could be possibly removed and made internal to WriteFeatureData?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Connection/LoginHandler.cs
```diff
@@ -0,0 +1,636 @@
...
+                }
+                if (_fedAuthFeature.ShouldUseFeature(requestedFeatures))
+                {
+                    SqlClientEventSource.Log.TryTraceEvent("<sc.TdsParser.TdsLogin|SEC> Sending federated authentication feature request");
```
cheenamalhotra, line 533: Trace events can be made internal to WriteFeatureData too based on whether feature is requested and is being written.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Connection/LoginHandler.cs
```diff
@@ -0,0 +1,636 @@
...
+                await writer.WriteShortAsync(offset, isAsync, ct).ConfigureAwait(false); // language name offset
+                await writer.WriteShortAsync(context.Language.Length, isAsync, ct).ConfigureAwait(false);
+                offset += context.Language.Length * 2;
+
+                await writer.WriteShortAsync(offset, isAsync, ct).ConfigureAwait(false); // database name offset
+                await writer.WriteShortAsync(context.Database.Length, isAsync, ct).ConfigureAwait(false);
+                offset += context.Database.Length * 2;
+
+                if (null == s_nicAddress)
+                    s_nicAddress = TdsParserStaticMethods.GetNetworkPhysicalAddressForTdsLoginOnly();
```
edwardneal, line 387: Nit: I'd personally want to keep this method dedicated to only explicitly showing the network IO, so I'd reference a NicAddress property similar to:
```csharp
public static byte[] NicAddress
    => s_nicAddress ??= TdsParserStaticMethods.GetNetworkPhysicalAddressForTdsLoginOnly();
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Connection/LoginHandler.cs
```diff
@@ -0,0 +1,636 @@
...
+                    await _utf8SupportFeature.WriteFeatureData(context, isAsync, ct).ConfigureAwait(false);
+                }
+
+                if (_sqlDnsCachingFeature.ShouldUseFeature(requestedFeatures))
+                {
+                    await _sqlDnsCachingFeature.WriteFeatureData(context, isAsync, ct).ConfigureAwait(false);
+                }
+
+                // terminator
+                await context.TdsStream.WriteByteAsync(0xFF, isAsync, ct).ConfigureAwait(false);
```
cheenamalhotra, line 560: Please use constant to place feature ext terminator.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Connection/LoginHandler.cs
```diff
@@ -0,0 +1,636 @@
...
+        /// </summary>
+        /// <param name="context"></param>
+        /// <returns></returns>
+        private int CalculateFeatureExtensionLength(LoginHandlerContext context)
+        {
+            int feLength = 0;
+            TdsEnums.FeatureExtension requestedFeatures = context.Features.RequestedFeatures;
+            if (context.UseFeatureExt)
+            {
+                if (_sessionRecoveryFeature.ShouldUseFeature(requestedFeatures))
```
cheenamalhotra, line 233: Same feedback for this block, ShouldUseFeature could be internally handled by GetLengthInBytes,

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Connection/LoginHandler.cs
```diff
@@ -0,0 +1,636 @@
...
+                                                || context.ConnectionOptions.Authentication == SqlAuthenticationMethod.ActiveDirectoryDeviceCodeFlow
+                                                || context.ConnectionOptions.Authentication == SqlAuthenticationMethod.ActiveDirectoryServicePrincipal
+                                                || context.ConnectionOptions.Authentication == SqlAuthenticationMethod.ActiveDirectoryManagedIdentity
+                                                || context.ConnectionOptions.Authentication == SqlAuthenticationMethod.ActiveDirectoryMSI
+                                                || context.ConnectionOptions.Authentication == SqlAuthenticationMethod.ActiveDirectoryDefault
+                                                || context.ConnectionOptions.Authentication == SqlAuthenticationMethod.ActiveDirectoryWorkloadIdentity
+                                                // Since AD Integrated may be acting like Windows integrated, additionally check _fedAuthRequired
+                                                || (context.ConnectionOptions.Authentication == SqlAuthenticationMethod.ActiveDirectoryIntegrated);
```
cheenamalhotra, line 120: You should be including a check for _fedAuthRequired with AD Integrated?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Connection/LoginHandler.cs
```diff
@@ -0,0 +1,636 @@
...
+                };
+                // No need any further info from the server for token based authentication. So set _federatedAuthenticationRequested to true
+                features.FederatedAuthenticationRequested = true;
+            }
+
+            // The GLOBALTRANSACTIONS, DATACLASSIFICATION, TCE, and UTF8 support features are implicitly requested
+            requestedFeatures |= TdsEnums.FeatureExtension.GlobalTransactions | TdsEnums.FeatureExtension.DataClassification | TdsEnums.FeatureExtension.Tce | TdsEnums.FeatureExtension.UTF8Support;
+
+            // The SQLDNSCaching feature is implicitly set
+            requestedFeatures |= TdsEnums.FeatureExtension.SQLDNSCaching;
```
cheenamalhotra, line 100: Combine them?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/PasswordChangeRequest.cs
```diff
@@ -0,0 +1,61 @@
...
+
+namespace Microsoft.Data.SqlClientX.Handlers
+{
+    internal class PasswordChangeRequest
+    {
+        public string NewPassword { get; internal set; }
+
+        public SqlCredential Credential { get; internal set; }
+
+        public SecureString NewSecurePassword { get; internal set; }
```
edwardneal, line 16: This corresponds to a public API surface, but SecureString [isn't actively recommended](https://learn.microsoft.com/en-us/dotnet/fundamentals/runtime-libraries/system-security-securestring) any more. We'll have to decrypt the SecureString and (merely) obfuscate it in the packet buffer anyway, so it'll be stored in memory whatever happens. Depending on its security implications, we might be able to remove this field, perform the transformation from SecureString to string in PasswordChangeRequest and thus remove the need to translate this in the login handler.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Connection/LoginHandler.cs
```diff
@@ -0,0 +1,636 @@
...
+
+            features.RequestedFeatures = requestedFeatures;
+
+            await TdsLogin(context, isAsync, ct).ConfigureAwait(false);
+
+            // If the workflow being used is Active Directory Authentication and server's prelogin response
+            // for FEDAUTHREQUIRED option indicates Federated Authentication is required, we have to insert FedAuth Feature Extension
+            // in Login7, indicating the intent to use Active Directory Authentication for SQL Server.
+            static bool ShouldRequestFedAuth(LoginHandlerContext context)
```
cheenamalhotra, line 109: Can you push this method also to FeatExtension types? 
e.g. FeatureExtensionFeatures.IsRequested()
Since context is already available to the FeatureExtensionFeatures, each feature can own this decision.

You can then call "_fedAuthFeature.IsRequested()" before appending the flag.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Connection/LoginHandler.cs
```diff
@@ -0,0 +1,636 @@
...
+        private readonly Utf8Feature _utf8SupportFeature;
+        private readonly SqlDnsCachingFeature _sqlDnsCachingFeature;
+        private readonly SessionRecoveryFeature _sessionRecoveryFeature;
+        private readonly FedAuthFeature _fedAuthFeature;
+        private readonly TceFeature _tceFeature;
+
+        /// <summary>
+        /// Parameterless constructor for the login handler.
+        /// </summary>
+        public LoginHandler()
```
cheenamalhotra, line 35: Can LoginHandler be initialized with context instead of passing context in the handler?
I'm wondering if these features can be initialized with context instead?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Connection/Login/FeatureExtensions.cs
```diff
@@ -0,0 +1,513 @@
...
+            int len = 1;
+            SessionData reconnectData = context.Features.ReconnectData;
+            if (context.Features.ReconnectData == null)
+            {
+                // Length of the data size int.
+                len += 4;
+            }
+            else
+            {
+                Debug.Assert(reconnectData._unrecoverableStatesCount == 0, "Unrecoverable state count should be 0");
```
edwardneal, line 162: I'd replicate the GetLengthInBytes pattern to the SessionData instance.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Connection/LoginHandler.cs
```diff
@@ -0,0 +1,636 @@
...
+            {
+                requestedFeatures |= TdsEnums.FeatureExtension.FedAuth;
+                features.FedAuthFeatureExtensionData = new FederatedAuthenticationFeatureExtensionData
+                {
+                    libraryType = TdsEnums.FedAuthLibrary.SecurityToken,
+                    fedAuthRequiredPreLoginResponse = context.FedAuthNegotiatedInPrelogin,
+                    accessToken = context.AccessTokenInBytes
+                };
+                // No need any further info from the server for token based authentication. So set _federatedAuthenticationRequested to true
+                features.FederatedAuthenticationRequested = true;
```
cheenamalhotra, line 93: Can you not combine both logics to create 1 method to determine whether this feature is being requested?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Connection/LoginHandler.cs
```diff
@@ -0,0 +1,636 @@
...
+        }
+
+        private async ValueTask SendLogin(LoginHandlerContext context, bool isAsync, CancellationToken ct)
+        {
+
+            PasswordChangeRequest passwordChangeRequest = context.PasswordChangeRequest;
+
+            TdsEnums.FeatureExtension requestedFeatures = TdsEnums.FeatureExtension.None;
+            FeatureExtensions features = context.Features;
+            if (context.ConnectionOptions.ConnectRetryCount > 0)
```
cheenamalhotra, line 63: Push this logical decision to "SessionRecoveryFeature.IsRequested()" check.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Connection/LoginHandler.cs
```diff
@@ -0,0 +1,636 @@
...
+
+            Debug.Assert(TdsEnums.MAXLEN_CLIENTINTERFACE >= context.ClientInterfaceName.Length, "cchCltIntName can specify at most 128 unicode characters. See Tds spec");
+
+            // Calculate the fixed length
+            checked
+            {
+                length += context.CalculateLoginRecordLength() * 2;
+                if (context.UseFeatureExt)
+                {
+                    length += 4;
```
cheenamalhotra, line 144: Can you push this logic block also in CalculateLoginRecordLength() ?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Connection/Login/FeatureExtensions.cs
```diff
@@ -0,0 +1,513 @@
...
+
+        public override async ValueTask WriteFeatureData(LoginHandlerContext context, bool isAsync, CancellationToken ct)
+        {
+            Debug.Assert(context.Features.SessionRecoveryRequested == true, "SessionRecoveryRequested should be true");
+
+            TdsStream stream = context.TdsStream;
+
+            await stream.WriteByteAsync(TdsEnums.FEATUREEXT_SRECOVERY, isAsync, ct).ConfigureAwait(false);
+            SessionData reconnectData = context.Features.ReconnectData;
+            if (reconnectData == null)
```
edwardneal, line 217: This seems similar to the logic behind `GetLengthInBytes(context) - 4`...

edwardneal (commented):
This all seems to make sense, thanks for CCing me. I think there's some overlap between my feedback and cheenamalhotra's, but it's pretty minimal.

#2689 now looks a lot more important though - invoking that many awaits seems like it would be expensive, even in the (potentially non-allocating) sync case.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Connection/LoginHandler.cs
saurabh500: It doesn't fit into the context-less design of handlers, today the orchestrator assumes that the handlers are stateless, and states are passed during Handle. 

The drawback is to instantiate a new handler for every connection request, which is unnecessary.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Connection/LoginHandler.cs
saurabh500: In my mind these feature classes contain the logic for the work done on the connection, rather than contain information like "Should Use feature" which is contained with the connection. This follows the handler paradigm where the state is maintained in the context, and the word is done in these feature classes.
The state is essentially flowing with the context, which the handler decides on, to invoke the right actions using the feature classes. 

Based on this feedback, I am going to move the features to the context, and then these features can call "Should Use" and then when your Parser is ready, it can add information to these feature classes to add the relevant information to be saved with these features.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Connection/Login/FeatureExtensions.cs
```diff
@@ -0,0 +1,513 @@
...
+            }
+
+            // length of feature id (1 byte), data length field (4 bytes), and feature data (dataLen)
+            int totalLen = dataLen + 5;
+            return totalLen;
+        }
+
+        public override async ValueTask WriteFeatureData(LoginHandlerContext context, bool isAsync, CancellationToken ct)
+        {
+            FederatedAuthenticationFeatureExtensionData fedAuthFeatureData = context.Features.FedAuthFeatureExtensionData;
```
mdaigle, line 398: great to have all of this encapsulated

mdaigle (commented):
The structure makes sense to me and I agree that I'd like to preserve this separation of responsibility if possible when we refactor for write performance.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Connection/LoginHandler.cs
cheenamalhotra: That seems fine then.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Connection/Login/FeatureExtensions.cs
```diff
@@ -0,0 +1,513 @@
...
+                    await stream.TdsWriter.WriteIntAsync(fedAuthFeatureData.accessToken.Length, isAsync, ct).ConfigureAwait(false);
+                    ct.ThrowIfCancellationRequested();
+                    if (isAsync)
+                    {
+                        await stream.WriteAsync(fedAuthFeatureData.accessToken.AsMemory(), ct).ConfigureAwait(false);
+                    }
+                    else
+                    {
+                        stream.Write(fedAuthFeatureData.accessToken);
+                    }
```
cheenamalhotra, line 504: Why not use only TdsWriter to write?
You need not reference stream and perform this logic here, TdsWriter already does that.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Connection/Login/FeatureExtensions.cs
```diff
@@ -0,0 +1,513 @@
...
+                                workflow = TdsEnums.MSALWORKFLOW_ACTIVEDIRECTORYTOKENCREDENTIAL;
+                            }
+                            else
+                            {
+                                Debug.Assert(false, "Unrecognized Authentication type for fedauth MSAL request");
+                            }
+                            break;
+                    }
+
+                    await stream.WriteByteAsync(workflow, isAsync, ct);
```
cheenamalhotra, line 492: You should only use TdsWriter to write IMO, if any API is needed, expose it directly from the TdsWriter.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Connection/Login/FeatureExtensions.cs
saurabh500: Because I was sitting on this PR before the TdsWriter came along. I can use TdsWriter.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Connection/LoginHandler.cs
saurabh500: In the current stateless design, there is anyhow going to be one only one login handler.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Connection/Login/FeatureExtensions.cs
saurabh500: I had migrated a bunch of calls to TdsWriter, this one was clearly left out

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Connection/LoginHandler.cs
saurabh500: The reason I want to stay away from the classic singletons is because it doesn't lend itself well to unit testing.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Connection/LoginHandler.cs
saurabh500: This is the check we do today while login. 
I didn't quite understand the suggestion. Are you asking whether we should continue checking for AD integrated? Or are you saying that we could remove it ?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Connection/LoginHandler.cs
saurabh500: I want to keep them separate.
I agree that GetLengthInBytes can check for "ShouldUseFeature"
But then that overloads the return value of Getting the length. 
A feature can have 0 byte length by design if there is no data exchange in the feature extension. 

I want to make the intention clear with a step which checks if feature should be used and it is a 0 byte feature vs if it is 0 byte because it was not requested.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Connection/LoginHandler.cs
saurabh500: Removing traceevents. That requires additional thinking.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Connection/LoginHandler.cs
cheenamalhotra: This is how it's compared in old code:

(ConnectionOptions.Authentication == SqlAuthenticationMethod.ActiveDirectoryIntegrated && _fedAuthRequired)

saurabh500: 1. I should have deleted that comment.
2. The _fedAuthRequired has been renamed to `FedAuthNegotiatedInPrelogin` in the context object and is being honored in the next line of code.. The _fedAuthRequired was set in ConsumePrelogin in SqlInternalConnectionTds. In this code, the PReloginhandler sets `FedAuthNegotiatedInPrelogin` with proper naming assigned to it. 

This function faithfully represents what was being done earlier with a slight change of checking for access token based on your PR comments about putting the check for AcceessTokens inside the FedAuth feature request check. 


The check for `Authentication` options has been first stored in `IsEntraIdAuthInConnectionString` to say that we are first evaluating the conditions in Connection string, after this we evaluate the conditions with Prelogin handshake as well as the other checks like callback and token bytes.

src/Microsoft.Data.SqlClient/tests/UnitTests/Handlers/LoginChainTest.cs
```diff
@@ -38,17 +34,23 @@ public async void TestConnectivity()
             {
                 DataSource = _dataSource,
                 Encrypt = SqlConnectionEncryptOption.Mandatory,
-                TrustServerCertificate = true
+                TrustServerCertificate = true,
+                UserID = "sa",
+                Password = Environment.GetEnvironmentVariable("PASSWORD"),
```
cheenamalhotra, line 39: To ensure we can read configuration from config.json, I've opened https://github.com/dotnet/SqlClient/issues/2736, can be picked up by anyone in next iteration.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Connection/Login/FeatureExtensions.cs
saurabh500: Changed all the instances I could find

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/PasswordChangeRequest.cs
saurabh500: Yeah, but out of scope of this PR. We will need to change the public API and then we can remove all this occurrences. 

I have brought in the consideration of SecureString, but haven't implemented it anywhere yet. 

There are some clever techniques that are used in the current SqlClient implementation to convert SecureString as late as possible to I/O buffer, but yes its not foolproof.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Connection/LoginHandler.cs
saurabh500: done

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Handlers/Connection/Login/FeatureExtensions.cs
```diff
@@ -0,0 +1,631 @@
...
+        {
+            TdsStream stream = context.TdsStream;
+            if (null != s)
+            {
+                await stream.TdsWriter.WriteByteAsync(checked((byte)s.Length), isAsync, ct).ConfigureAwait(false);
+                await stream.WriteStringAsync(s, isAsync, ct).ConfigureAwait(false);
+            }
+            else
+            {
+                await stream.WriteByteAsync((byte)0, isAsync, ct).ConfigureAwait(false);
```
cheenamalhotra, line 456: A few places still need TdsWriter to be used, not critical but something to address maybe later :)

cheenamalhotra (approved):

saurabh500:
@edwardneal  there was a sense of urgency to get this in, and I know that there is a bunch of changes that would be needed as pointed out in your review comments. I wanted to acknowledge your review comments..
@cheenamalhotra  is working to light up the "Read" login response, and needed this PR merged. 
We will have this working E2E and then have a benchmark ready. And we will be able to test the perf improvement hypothesis with benchmarks

cheenamalhotra:
Thanks @saurabh500 !

@edwardneal 
Please feel free to open a GH issue and tag it with `ClientX` for any pending feedback and we will triage it :)

edwardneal:
Thanks for the update - looking forward to seeing the login response code land! I'll wait for that to come through, look at performance benchmarking and see if any of my comments still hold in light of that. The thought behind SecureString usage sits separately to that... I'll give it a little more thought and submit it separately.

Issue #2630, opened 2024-06-28
Implement pruning in connection pool


mdaigle:

mdaigle:
Relates to https://github.com/dotnet/SqlClient/issues/343

PR #2739, opened 2024-07-30
Pool Warmup

mdaigle:
Performs initial pool warmup upon instantiation. Does not include logic for maintaining min pool size over time, that will follow in #2630 along with pruning logic.

It's important to feed "pool initiated" opens through the same rate limiter as user initiated opens. This ensures that the pool as a whole obeys the desired rate limiting behavior.

Warm up connections one at a time to maintain today's behavior. Users who need a faster warmup can just start using the pool without rate limitation and create connections as quickly as a parallel warmup process could.


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/PoolingDataSource.cs
```diff
@@ -424,13 +428,46 @@ internal void PruneIdleConnections()
...
+            // Make sure we're at quota by posting a callback to the threadpool.
+            ThreadPool.QueueUserWorkItem(_warmupCallback);
+        }
+
         /// <summary>
         /// Warms up the pool to bring it up to min pool size.
         /// </summary>
         /// <exception cref="NotImplementedException"></exception>
-        internal void WarmUp()
+        private async void WarmupCallback(object? state)
```
roji, line 441: It's generally a good idea to avoid async void methods whenever possible ([article](https://learn.microsoft.com/en-us/archive/msdn-magazine/2013/march/async-await-best-practices-in-asynchronous-programming#avoid-async-void)); if you have to use them, it's worth paying very close attention to what happens with exceptions - any exception that bubbles out of an async void method causes AppDomain.UnhandledException to get fired.

I'd recommend making this async Task and assigning the Task to some instance variable on the pool; your could even make sure that Task is complete when the pool is shuw down/disposed (pretty contrived edge case, but at least for completeness).

Otherwise if you keep async void, maybe consider wrapping the method body with a try/catch that doesn't bubble any exceptions out (but rather logs).

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/PoolingDataSource.cs
```diff
@@ -65,11 +67,11 @@ internal sealed class PoolingDataSource : SqlDataSource
         /// Initializes a new PoolingDataSource.
         /// </summary>
         //TODO: support auth contexts and provider info
-        internal PoolingDataSource(SqlConnectionStringBuilder connectionStringBuilder,
+        internal PoolingDataSource(SqlConnectionString connectionString,
```
benrr101, line 70: nit: first argument when chopped should go on its own line

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/PoolingDataSource.cs
```diff
@@ -307,18 +311,18 @@ private void CloseConnector(SqlConnector connector)
         /// </summary>
         internal readonly struct OpenInternalConnectionState
         {
-            internal readonly SqlConnectionX _owningConnection;
+            internal readonly SqlConnectionX? _owningConnection;
```
benrr101, line 314: My gut says these should be auto-properties... I think the general pattern is that member variables should always be private.

This might also change the requirements for having a constructor, such that you can use the object initializer syntax to create the instance instead.

It's six of one, half dozen of the other, but I think in general having member variables be private is good.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/PoolingDataSource.cs
benrr101: I'm just going to push it over the fence and say "this should be `async Task` or `async ValueTask`" :)

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/PoolingDataSource.cs
```diff
@@ -424,13 +428,46 @@ internal void PruneIdleConnections()
...
+            while (_numConnectors < MinPoolSize)
+            {
+                // Obey the same rate limit as user-initiated opens.
+                // Ensures that pool-initiated opens are queued properly alongside user requests.
+                SqlConnector? connector = await OpenNewInternalConnection(
+                    null,
+                    TimeSpan.FromSeconds(Settings.ConnectTimeout),
+                    true,
+                    CancellationToken.None
+                ).ConfigureAwait(false);
```
benrr101, line 458: I personally like this style of putting the closing paren on a new line, but MSFT C# style seems to be to leave it on the same line as the last parameter. The `.ConfigureAwait` would go on the next line, I think...

```
var connector = await OpenNewInternalConnection(
    ...
    CancellationToken.None)
    .ConfigureAwait(false);
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlDataSource.cs
```diff
@@ -22,29 +21,30 @@ namespace Microsoft.Data.SqlClientX
...
             Credential = credential;
         }
         #endregion
 
         #region properties
         /// <inheritdoc />
-        public override string ConnectionString => _connectionStringBuilder.ConnectionString;
+        public override string ConnectionString { get; }
+
+        internal SqlConnectionString Settings { get; }
```
benrr101, line 47: I'd recommend we add comments for any public _or internal_ properties/methods/constructors as we go. Good documentation will help us in the future!

src/Microsoft.Data.SqlClient/tests/UnitTests/PoolTests.cs
```diff
@@ -459,6 +459,19 @@ public async Task ConnectionLifetime()
...
         */
+
+        [Theory]
+        [InlineData(0)]
+        [InlineData(1)]
+        [InlineData(5)]
+        public async Task Warmup_OpenedConnectionsEqualsMinPoolSize(int minPoolSize)
+        {
+            await using var dataSource = testBase.CreateDataSource(csb => csb.MinPoolSize = minPoolSize);
+            Thread.Sleep(500*minPoolSize);
```
benrr101, line 470: Three things:
* Organizing tests into a Arrange / Act / Assert helps make them more readable what is setup tasks, what is the test action, and what needs to be verified.
* Since this test is `async`, you can use `await Task.Delay()` to sleep without blocking a thread
* However, to me having delays in a test is a code smell... imho, having arbitrary sleep times either cause tests to become flaky if the timeouts are too short or cause the test to take longer than necessary, slowing down the test run.
  * This is a complicated thing to try and avoid. My suggestion is to consider exposing a task that can be awaited. It would likely be an internal, test only, exposed object. The idea here is that the code needs to be built to be tested in a efficient manner. If you want to discuss some ideas, we can chat offline.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/PoolingDataSource.cs
edwardneal: Since we're thinking about having this as a `Task`, then I suggest using `TaskCreationOptions.LongRunning`, so that the pool maintenance task doesn't sit in the normal thread pool (and won't generate/suffer from the effects of thread pool exhaustion.)

We also definitely want to be able to cancel the maintenance task using a CancellationToken - I think `SqlConnection.ClearAllPools()` will need to pull that lever.

mdaigle: The way I'm reading the `TaskCreationOptions.LongRunning` option is that it makes sense if you're doing a lot of synchronous work on that Task. This warmup process and the broader pool maintenance tasks will all be async, so I believe it won't make sense to set that option.

src/Microsoft.Data.SqlClient/tests/UnitTests/PoolTests.cs
mdaigle: I was able to rework things a bit to expose the task. This allows a direct await on the warmup task in the tests.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/PoolingDataSource.cs
mdaigle: I got this working using a task stored in an instance variable. It also improved the unit testing as this task can be awaited directly to determine when warmup has completed.

edwardneal: Yes, you're right - not sure how I missed that, sorry! I can see the CancellationToken is in place, so I'm happy with the methodology.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/PoolingDataSource.cs
```diff
@@ -425,12 +438,55 @@ internal void PruneIdleConnections()
...
-        /// Warms up the pool to bring it up to min pool size.
+        /// Warms up the pool by bringing it up to min pool size.
         /// </summary>
-        /// <exception cref="NotImplementedException"></exception>
-        internal void WarmUp()
+        /// <returns>A ValueTask representing the warmup process.</returns>
+        internal ValueTask WarmUp()
         {
-            throw new NotImplementedException();
+            lock (_warmupLock)
```
saurabh500, line 446: Locks can be cheap until they are not. 
Consider moving away from the usage of locks in a method which is meant to be run asynchronously.

saurabh500:
Few comments:
1. Rethink the warmup scheduling. Recommendation: schedule warm up on the first connection open.
2. Warmup should be cancelled on Data Source dispose. 
3. Remove usage of locks in Methods that return ValueTask/Task. SemaphoreSlim with WaitAsync would be a better choice in the absence of something else more useful. 
4. In case of failure to warmup, warmup task should have enough resiliency to schedule itself later. 
5. Consider scenarios where Pool is cleared and then it would need to be warmed up again, to meet the Min Pool Size criterion. 

Please tackle these in later PRs. Approving this since it brings in a baseline implementation to improve upon.

saurabh500 (approved):
Approved with comments

PR #2744, opened 2024-08-01
Json read and write tests

deepaksa1:
This PR aims to add read and write tests for json data type.

### DIFF
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/JsonTest/JsonTest.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/JsonTest/JsonTest.cs
@@ -1,0 +1,78 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Data;
+using System.Data.Common;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.Data.SqlClient.ManualTesting.Tests
+{
+    public class JsonTest
+    {
+        private readonly ITestOutputHelper _output;
+
+        public JsonTest(ITestOutputHelper output)
+        {
+            _output = output;
+        }
+
+        [ConditionalFact(typeof(DataTestUtility), nameof(DataTestUtility.IsJsonSupported))]
+        public void TestJsonWrite()
+        {
+            AppContext.SetSwitch("Switch.Microsoft.Data.SqlClient.UseManagedNetworkingOnWindows", true);
+            
+            string jsonDataString = "[\r\n    {\r\n        \"name\": \"Dave\",\r\n        \"skills\": [ \"Python\" ]\r\n    },\r\n    {\r\n        \"name\": \"Ron\",\r\n        \"surname\": \"Peter\"\r\n    }\r\n]";
+
+            using (SqlConnection connection = new SqlConnection(DataTestUtility.TCPConnectionString))
+            {
+                connection.Open();
+
+                string query = "INSERT INTO dbo.JsonTable VALUES (@jsonData)";
+                using (SqlCommand command = new SqlCommand(query, connection))
+                {
+                    // Add the parameter and set its value
+                    var parameter = new SqlParameter("@jsonData", jsonDataString);
+                    parameter.SqlDbType = SqlDbTypeExtensions.Json;
+                    command.Parameters.Add(parameter);
+
+                    for (int i = 0; i < 10; i++)
+                    {
+                        int rowsAffected = command.ExecuteNonQuery();
+                        _output.WriteLine($"Rows affected: {rowsAffected}");
+                    }
+                }
+            }
+        }
+
+        [ConditionalFact(typeof(DataTestUtility), nameof(DataTestUtility.IsJsonSupported))]
+        public void TestJsonRead()
+        {
+            AppContext.SetSwitch("Switch.Microsoft.Data.SqlClient.UseManagedNetworkingOnWindows", true);
+            
+            SqlConnection connection = new SqlConnection(DataTestUtility.TCPConnectionString);
+            connection.Open();
+            SqlCommand command = connection.CreateCommand();
+            string commandText = "SELECT * FROM dbo.JsonTable";
+            command.CommandText = commandText;
+            SqlDataReader reader = command.ExecuteReader();
+            while (reader.Read())
+            {
+                string jsonData = reader.GetString(0);
+                _output.WriteLine(jsonData);
+                Assert.NotNull(jsonData);
+            }
+            System.Collections.ObjectModel.ReadOnlyCollection<DbColumn> schema = reader.GetColumnSchema();
+            foreach (DbColumn column in schema)
+            {
+                _output.WriteLine("Column Name is " + column.ColumnName);
+                _output.WriteLine("Column DataType is " + column?.DataType.ToString());
+                _output.WriteLine("Column DataTypeName is " + column.DataTypeName);
+            }
+            reader.Close();
+            connection.Close();
+        }
+    }
+}
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/JsonTest/JsonTest.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/JsonTest/JsonTest.cs
@@ -6,8 +6,7 @@ using System;
 using System.Data;
 using System.Data.Common;
 using System.Data.SqlTypes;
-using System.Xml.Serialization;
-using Microsoft.Data.SqlClient.ManualTesting.Tests.AlwaysEncrypted;
+using System.Threading.Tasks;
 using Xunit;
 using Xunit.Abstractions;
 
@@ -33,6 +32,16 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests
         private void ValidateRows(SqlDataReader reader)
         {
             while (reader.Read())
+            {
+                string jsonData = reader.GetString(0);
+                _output.WriteLine(jsonData);
+                Assert.Equal(jsonDataString, jsonData);
+            }
+        }
+
+        private async Task ValidateRowsAsync(SqlDataReader reader)
+        {
+            while (await reader.ReadAsync())
             {
                 string jsonData = reader.GetString(0);
                 _output.WriteLine(jsonData);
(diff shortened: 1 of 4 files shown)


src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/JsonTest/JsonTest.cs
```diff
@@ -0,0 +1,78 @@
...
+                string jsonData = reader.GetString(0);
+                _output.WriteLine(jsonData);
+                Assert.NotNull(jsonData);
+            }
+            System.Collections.ObjectModel.ReadOnlyCollection<DbColumn> schema = reader.GetColumnSchema();
+            foreach (DbColumn column in schema)
+            {
+                _output.WriteLine("Column Name is " + column.ColumnName);
+                _output.WriteLine("Column DataType is " + column?.DataType.ToString());
+                _output.WriteLine("Column DataTypeName is " + column.DataTypeName);
```
saurabh500, line 72: You may want to assert that the datatype is json

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/JsonTest/JsonTest.cs
```diff
@@ -0,0 +1,78 @@
...
+    public class JsonTest
+    {
+        private readonly ITestOutputHelper _output;
+
+        public JsonTest(ITestOutputHelper output)
+        {
+            _output = output;
+        }
+
+        [ConditionalFact(typeof(DataTestUtility), nameof(DataTestUtility.IsJsonSupported))]
```
deepaksa1, line 22: @David-Engel @cheenamalhotra 

Need some help on the usage of `ConditionalFact` 
I am trying to set this test to run only if JSON is supported. However, this test also runs even if `IsJsonSupported` is set as `false`

JRahnama: @deepaksa1Are you using Console or Visual Studio? `xunit.runner.visualstudio` and `xunit.runner.console` work differently. In Visual Studio, you might not be able to run platform-specific tests or certain xUnit extensions as expected. It's better to test using the Console to see the accurate outcome. (This is by my experience. @David-Engel or @cheenamalhotra may have better inputs to help you on this)

David-Engel: JRahnama is correct. I've seen VS run tests even when the ConditionalFact says it shouldn't. Console results should behave correctly.

deepaksa1: Thanks @JRahnama and @David-Engel for pointing it out. I ran the tests from console and the results were as expected.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/JsonTest/JsonTest.cs
deepaksa1: added

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/JsonTest/JsonTest.cs
```diff
@@ -0,0 +1,161 @@
...
+        public void TestJsonWrite()
+        {
+            AppContext.SetSwitch("Switch.Microsoft.Data.SqlClient.UseManagedNetworkingOnWindows", true);
```
saurabh500, line 26: Why ? 
Don't use this AppContext switch.

ErikEJ:
How about reader.GetFieldValue<JsonDocument> (which together  with SqlDdType.Json for SqlParameter) is waht EF Core needs. I see no tests for that? (But I could be looking in the wrong place)

saurabh500:
@ErikEJ the support hasn't been added yet. The API is on the list of things to add.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/JsonTest/JsonTest.cs
```diff
@@ -0,0 +1,161 @@
...
+                    //Write json value using SP
+                    using (SqlCommand command2 = connection.CreateCommand())
+                    {
+                        command2.CommandText = spName;
+                        command2.CommandType = CommandType.StoredProcedure;
+                        var parameter2 = new SqlParameter("@jsonData", SqlDbTypeExtensions.Json);
+                        parameter2.Value = jsonDataString;
+                        command2.Parameters.Add(parameter2);
+                        int rowsAffected3 = command2.ExecuteNonQuery();
+                        _output.WriteLine($"Rows affected: {rowsAffected3}");
```
saurabh500, line 77: No asserts?

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/JsonTest/JsonTest.cs
```diff
@@ -0,0 +1,161 @@
...
+                    }
+                    reader.Close();
+
+                    //Test 3
+                    //Read json value using SP
+                    using (SqlCommand command2 = connection.CreateCommand())
+                    {
+                        command2.CommandText = spName;
+                        command2.CommandType = CommandType.StoredProcedure;
+                        SqlDataReader reader2 = command2.ExecuteReader();
```
saurabh500, line 148: You could try wrapping this in using block

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/JsonTest/JsonTest.cs
```diff
@@ -0,0 +1,161 @@
...
+
+                    //Test 2
+                    //Read the column metadata
+                    System.Collections.ObjectModel.ReadOnlyCollection<DbColumn> schema = reader.GetColumnSchema();
+                    foreach (DbColumn column in schema)
+                    {
+                        _output.WriteLine("Column Name is " + column.ColumnName);
+                        _output.WriteLine("Column DataType is " + column?.DataType.ToString());
+                        _output.WriteLine("Column DataTypeName is " + column.DataTypeName);
+                        Assert.Equal("json",column.DataTypeName);
```
saurabh500, line 138: Space after ,

saurabh500: Can you format this code? Looks like a few locations could use space after comma

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/JsonTest/JsonTest.cs
deepaksa1: removed

src/Microsoft.Data.SqlClient/tests/tools/Microsoft.Data.SqlClient.TestUtilities/Config.cs
```diff
@@ -42,6 +42,7 @@ public class Config
         public string KerberosDomainUser = null;
         public bool IsManagedInstance = false;
         public string AliasName = null;
+        public bool IsJsonSupported = false;
```
cheenamalhotra, line 45: Is this something you can detect after making connection?

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/JsonTest/JsonTest.cs
```diff
@@ -0,0 +1,157 @@
...
+    {
+        private readonly ITestOutputHelper _output;
+
+        public JsonTest(ITestOutputHelper output)
+        {
+            _output = output;
+        }
+
+        [ConditionalFact(typeof(DataTestUtility), nameof(DataTestUtility.IsJsonSupported))]
+        public void TestJsonWrite()
```
cheenamalhotra, line 24: Please add async API testing as well.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/JsonTest/JsonTest.cs
```diff
@@ -0,0 +1,157 @@
...
+                    using (SqlCommand command2 = connection.CreateCommand())
+                    {
+                        command2.CommandText = spName;
+                        command2.CommandType = CommandType.StoredProcedure;
+                        SqlDataReader reader2 = command2.ExecuteReader();
+                        while (reader2.Read())
+                        {
+                            string jsonData = reader2.GetString(0);
+                            _output.WriteLine(jsonData);
+                            Assert.NotNull(jsonData);
```
cheenamalhotra, line 149: I don't think this is a very good way to validate, can you compare jsonData with source data that was written to DB?

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/JsonTest/JsonTest.cs
```diff
@@ -0,0 +1,157 @@
...
+                    SqlDataReader reader = command.ExecuteReader();
+                    while (reader.Read())
+                    {
+                        string jsonData = reader.GetString(0);
+                        _output.WriteLine(jsonData);
+                        Assert.Equal(jsonDataString,jsonData);
+                    }
+
+                    //Test 2
+                    //Read the column metadata
```
cheenamalhotra, line 127: I'd recommend adding multiple tests instead of running all tests in 1 big one. You can create tables in setup and drop them in tear down.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/JsonTest/JsonTest.cs
deepaksa1: Added asserts for write tests

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/JsonTest/JsonTest.cs
deepaksa1: Added `using` block

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/JsonTest/JsonTest.cs
deepaksa1: Done.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/JsonTest/JsonTest.cs
deepaksa1: Modified the assert statement

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/JsonTest/JsonTest.cs
deepaksa1: These are quite small tests (single operation) and uses common sql command. Should I break them into individual tests?

src/Microsoft.Data.SqlClient/tests/tools/Microsoft.Data.SqlClient.TestUtilities/Config.cs
saurabh500: @cheenamalhotra 
Can't be done without turning on Feature Switches on Debug builds of Sql server today. 
We will need a hardcoded config value at this moment unfortunately :( 

also question for my education @cheenamalhotra . In ConditionalFact check, do we make server calls to check for features. Because that may lead to transient fault handling in test execution condition. I like the idea of dynamically checking the server for the feature, as long as it is not done in the body of the test, since the test would show as executed but it really wont be executed. So adding a server query in ConditionalFact condition would make sense. But wanted to check if there is prior art.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/JsonTest/JsonTest.cs
saurabh500: @deepaksa1  this is great feedback.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/JsonTest/JsonTest.cs
saurabh500: I think it makes sense to keep this test as is. Data is being read and various aspects of the data are being verified, i.e. Data / Metadata.

However it would be nicer to see these refactored into various methods.
e.g. You could have a reader being executed and data read/asserted in a different function, and then from the same reader extract the Schema and send it to another function to validate. 

Having multiple tests also tend to increase runtime, for setup which, could otherwise have been done in a single test. There is no golden rule. But for this test, it makes sense to have a single `ConditionalFact` and then have multiple verification methods being called from the same test.

Basically 
```
var reader = command.ExecuteReader()
ValidateRows(reader);
ValidateScehma(reader.GetColumnSchema());
reader.Close();
```

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/JsonTest/JsonTest.cs
deepaksa1: Added async tests

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/JsonTest/JsonTest.cs
deepaksa1: Refactored test cases.

PR #2749, opened 2024-08-02
Code Cleanup | `null == x` to `x == null`

benrr101:
**Description**: This PR is part of a series that aims to remove "yoda conditions" from the codebase. In my opinion, these make the code harder to read due to the natural tendency to read an if condition like "x equals y". In this case "null equals x" is unnatural compared to "x equals null". In most if not all cases, there is no semantic difference between `x == null` and `null == x`, meaning these changes can be safely made without impact to the behavior.

**AI Analysis**: Since this PR is large and boring for people to review, I ran the patch through a large language model to ask it to assess the safety of it. Here's is its analysis:

image

**Testing**: The code builds, I will wait for CI to verify it more thoroughly.

_Note, this is only part 1 of several parts_

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
@@ -615,7 +615,7 @@ namespace Microsoft.Data.SqlClient
             get
             {
                 // if the transaction object has been zombied, just return null
-                if ((null != _transaction) && (null == _transaction.Connection))
+                if ((null != _transaction) && _transaction.Connection == null)
                 {
                     _transaction = null;
                 }
@@ -783,7 +783,7 @@ namespace Microsoft.Data.SqlClient
         {
             get
             {
-                if (null == _parameters)
+                if (_parameters == null)
                 {
                     // delay the creation of the SqlParameterCollection
                     // until user actually uses the Parameters property
@@ -1028,12 +1028,12 @@ namespace Microsoft.Data.SqlClient
                     // if we have pending data, but it is not a result of this command, then we don't cancel either.  Note that
                     // this model is implementable because we only allow one active command at any one time.  This code
                     // will have to change we allow multiple outstanding batches
-                    if (null == _activeConnection)
+                    if (_activeConnection == null)
                     {
                         return;
                     }
                     SqlInternalConnectionTds connection = (_activeConnection.InnerConnection as SqlInternalConnectionTds);
-                    if (null == connection)
+                    if (connection == null)
                     {  // Fail with out locking
                         return;
                     }
@@ -1050,11 +1050,10 @@ namespace Microsoft.Data.SqlClient
                         }
 
                         TdsParser parser = connection.Parser;
-                        if (null == parser)
+                        if (parser == null)
                         {
                             return;
                         }
-
 
                         if (!_pendingCancel)
                         { // Do nothing if already pending.
@@ -1644,7 +1643,7 @@ namespace Microsoft.Data.SqlClient
                 }
             }
 
-            Debug.Assert(null == _stateObj, "non-null state object in EndExecuteNonQuery");
+            Debug.Assert(_stateObj == null, "non-null state object in EndExecuteNonQuery");
             return _rowsAffected;
         }
 
@@ -1716,7 +1715,7 @@ namespace Microsoft.Data.SqlClient
                 }
             }
 
-            Debug.Assert(isAsync || null == _stateObj, "non-null state object in InternalExecuteNonQuery");
+            Debug.Assert(isAsync || _stateObj == null, "non-null state object in InternalExecuteNonQuery");
             return task;
         }
 
@@ -2564,7 +2563,7 @@ namespace Microsoft.Data.SqlClient
             CheckThrowSNIException();
 
             SqlDataReader reader = CompleteAsyncExecuteReader(isInternal);
-            Debug.Assert(null == _stateObj, "non-null state object in InternalEndExecuteReader");
+            Debug.Assert(_stateObj == null, "non-null state object in InternalEndExecuteReader");
             return reader;
         }
 
@@ -3297,7 +3296,7 @@ namespace Microsoft.Data.SqlClient
 
             // Use common parser for SqlClient and OleDb - parse into 4 parts - Server, Catalog, Schema, ProcedureName
             string[] parsedSProc = MultipartIdentifier.ParseMultipartIdentifier(CommandText, "[\"", "]\"", Strings.SQL_SqlCommandCommandText, false);
-            if (null == parsedSProc[3] || string.IsNullOrEmpty(parsedSProc[3]))
+            if (string.IsNullOrEmpty(parsedSProc[3]))
             {
                 throw ADP.NoStoredProcedureExists(CommandText);
             }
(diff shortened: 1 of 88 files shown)


DavoudEshtehari (commented):
Thanks for cleaning them up; I always want to fix these conditions.

The `!=` conditions can be addressed similarly. I started suggesting the code change, but it would be better to be done on your side. I'll push my suggestions so far. Also, using `is` or `is not` instead of `==` or `!=` can be slightly more efficient because it avoids the potential overhead of an overloaded operator. The approach you're taken is the safest, but feel free to choose the other one if you're ready for the potential headache of verifying possible overloads.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
```diff
@@ -615,7 +615,7 @@ internal SqlStatistics Statistics
             get
             {
                 // if the transaction object has been zombied, just return null
-                if ((null != _transaction) && (null == _transaction.Connection))
+                if ((null != _transaction) && _transaction.Connection == null)
```
David-Engel, line 618:
```suggestion
                if ((_transaction != null) && _transaction.Connection == null)
```

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/ProviderBase/DbConnectionInternal.cs
```diff
@@ -92,7 +92,7 @@ protected internal Transaction EnlistedTransaction
             set
             {
                 Transaction currentEnlistedTransaction = _enlistedTransaction;
-                if (((null == currentEnlistedTransaction) && (null != value))
+                if ((currentEnlistedTransaction == null && (null != value))
```
David-Engel, line 95:
```suggestion
                if ((currentEnlistedTransaction == null && (value != null))
```

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlClientPermission.cs
```diff
@@ -198,7 +198,7 @@ override public bool IsSubsetOf(IPermission target)
             {
                 if (!IsUnrestricted() &&
                     (!AllowBlankPassword || superset.AllowBlankPassword) &&
-                    ((null == _keyvalues) || (null != superset._keyvaluetree)))
+                    (_keyvalues == null || (null != superset._keyvaluetree)))
```
David-Engel, line 201:
```suggestion
                    (_keyvalues == null || (superset._keyvaluetree != null)))
```

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlCommand.cs
```diff
@@ -711,7 +711,7 @@ internal SqlStatistics Statistics
             get
             {
                 // if the transaction object has been zombied, just return null
-                if ((null != _transaction) && (null == _transaction.Connection))
+                if ((null != _transaction) && _transaction.Connection == null)
```
David-Engel, line 714:
```suggestion
                if ((_transaction != null) && _transaction.Connection == null)
```

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnectionFactory.cs
```diff
@@ -92,8 +92,7 @@ override protected DbConnectionInternal CreateConnection(DbConnectionOptions opt
                     redirectedUserInstance = true;
                     string instanceName;
 
-                    if ((null == pool) ||
-                         (null != pool && pool.Count <= 0))
+                    if (pool == null || (null != pool && pool.Count <= 0))
```
David-Engel, line 95:
```suggestion
                    if (pool == null || (pool != null && pool.Count <= 0))
```

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -347,16 +347,16 @@ internal SqlInternalTransaction CurrentTransaction
...
                           || (null != _currentTransaction && !_currentTransaction.IsLocal), "attempting to change current transaction?");
 
                 // If there is currently a transaction active, we don't want to
                 // change it; this can occur when there is a delegated transaction
                 // and the user attempts to do an API begin transaction; in these
                 // cases, it's safe to ignore the set.
-                if ((null == _currentTransaction && null != value)
-                  || (null != _currentTransaction && null == value))
+                if ((_currentTransaction == null && null != value) ||
+                    (null != _currentTransaction && value == null))
```
David-Engel, line 359:
```suggestion
                if ((_currentTransaction == null && value != null) ||
                    (_currentTransaction != null && value == null))
```

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -13874,12 +13880,11 @@ internal int ReadPlpAnsiChars(ref char[] buff, int offst, int len, SqlMetaDataPr
             {
                 Encoding enc = metadata.encoding;
 
-                if (enc == null)
+                if (null == enc)
```
David-Engel, line 13883:
```suggestion
                if (enc == null)
```

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/NameValuePair.cs
```diff
@@ -42,7 +42,7 @@ internal NameValuePair Next
             get => _next;
             set
             {
-                if ((null != _next) || (null == value))
+                if ((null != _next) || value == null)
```
David-Engel, line 45:
```suggestion
                if (_next != null || value == null)
```

David-Engel (commented):
I also prefer this direction. 👍

benrr101:
@DavoudEshtehari @David-Engel Once this PR goes through, I have all the code changes lined up for doing `null != x` to `x != null` as well as several others for numbers and other constants. I originally started with making the change to `x is null` but since there is that potential for it to call an overloaded `==` operator, I wanted to err on the side of caution for safest change rather than risk breaking something in a very hard to detect way.

David-Engel (approved):

roji:
Just noting that it's becoming increasingly common/standard to do `x is null` and `x is not null` rather than `x == null` and `x != null`; with the latter, operator overloading can change the meaning of the expression, whereas with the former you know you're always checking for null.

benrr101:
@roji Acknowledged - I think I'm being overly cautious here just because it wouldn't surprise me if buried somewhere in the code there's an `==` overload that treats `null` as `1` or something ridiculous. (I suppose it's possible too that the there could be an overload that makes `==` not commutative, but I hope that didn't happen) I didn't want to dive into each of the operators to see if they have unexpected behavior, so this is the fastest, safest option.

We can convert to `is`/`is not` in another PR.

Wraith2:
Did you make the changes manually or did you write an analyzer+codefixer to do it for you?

benrr101:
@Wraith2 definitely not the latter, but not entirely the former. I had some automated finding and making the changes, but I assessed each one (very quickly, that is) as I made the change.

PR #2750, opened 2024-08-02
Switch macOS to test against local server

David-Engel:
This PR switches macOS jobs to install Docker and a local SQL Server container instance instead of connecting to an Azure SQL DB instance for testing. This will allow us to get rid of the network permissions required for the mac agents to access Azure SQL DB and we can also get rid of the AADServicePrincipal connection info.

The ConfigurableIpPreferenceTest had to be adjusted to not run on macOS when targeting localhost as I couldn't get IPv6 working. It appears IPv6 in Docker is only supported on Linux:

https://docs.docker.com/engine/daemon/ipv6/

> IPv6 is only supported on Docker daemons running on Linux hosts.

It might be coming soon for macOS, though: https://github.com/docker/for-mac/issues/1432

Additions: (these should make #3094 obsolete)

**Update: This also fixes CI for PRs from forks.**
PRs from forks won't have access to any secrets (tightened security policy) so they won't be able to run the AE enclave jobs and a few tests that exercised AAD Service Principal Secret methods. Any PRs that break those from forks may only be caught by the nightly scheduled CI runs against main.

You can see fewer jobs ran in this screenshot for this PR compared to #3108:
image

- Change tests against local SQL Server to always generate a random password for local accounts.
- Change jobs against local SQL Server to not include AAD connection tests.
- Update some configs to not pass secret variables to jobs if they aren't available to the pipeline. See #3108 for a comparison of CI pipeline output from a branch in this repo versus a fork (this PR).

Code coverage difference shows 66% from this fork PR and 72% from #3108.


eng/pipelines/common/templates/steps/configure-sql-server-macos-step.yml
```diff
@@ -0,0 +1,90 @@
...
+    sqlcmd -S 127.0.0.1 -No -U sa -P $MSSQL_SA_PW -Q "SELECT @@VERSION" -l 2
+    echo ::1
+    sqlcmd -S ::1 -No -U sa -P $MSSQL_SA_PW -Q "SELECT @@VERSION" -l 2
+    echo localhost
+    sqlcmd -S localhost -No -U sa -P $MSSQL_SA_PW -Q "SELECT @@VERSION" -l 2
+    echo "(sqlcmd default / not specified)"
+    sqlcmd -No -U sa -P $MSSQL_SA_PW -Q "SELECT @@VERSION" -l 2
+
+    echo "Configuring Dedicated Administer Connections to allow remote connections..."
+    sqlcmd -S 0.0.0.0 -No -U sa -P $MSSQL_SA_PW -Q "sp_configure 'remote admin connections', 1; RECONFIGURE;"
```
mdaigle, line 80: Are there tests that require DAC access?

David-Engel: Yes. `Microsoft.Data.SqlClient.ManualTesting.Tests.ConnectivityTest.DacConnectionTest`

eng/pipelines/dotnet-sqlclient-ci-core.yml
```diff
@@ -310,9 +292,11 @@ stages:
             TCPConnectionString: $(AZURE_DB_TCP_CONN_STRING)
             NPConnectionString: $(AZURE_DB_NP_CONN_STRING)
             AADAuthorityURL: $(AADAuthorityURL)
-            AADPasswordConnectionString: $(AAD_PASSWORD_CONN_STR)
+            ${{ if eq(variables['system.pullRequest.isFork'], 'False') }}:
```
David-Engel, line 295: Note: I did NOT like having to use the isFork variable to determine if secrets are available. But everything else I tried (trying to reference secret variables various ways to detect if they were present) would not run consistently across forks and non-forks.

mdaigle (approved):

cheenamalhotra (approved):

benrr101 (approved):
This all looks pretty decent to me, especially if it unblocks fork builds.
Only thing I'm not sure about - this removes tests for macos against azure sql?

eng/pipelines/dotnet-sqlclient-ci-core.yml
samsharma2700, line 295:
```suggestion
            #Note: I did NOT like having to use the isFork variable to determine if secrets are available. But everything else I tried (trying to reference secret variables various ways to detect if they were present) would not run consistently across forks and non-forks.
            ${{ if eq(variables['system.pullRequest.isFork'], 'False') }}:
```

samsharma2700: We can leave a note in the file for posterity

samsharma2700 (approved):

David-Engel:
> thoughts on adding back a mac-azure run only for non-forked builds? You mentioned above that we can remove some network permissions. Is that still a benefit given the forked build fix included here.

@mdaigle The original reason to eliminate macOS+Azure was to get rid of usage of the AAD SP Secret (the "eliminate secrets" compliance task). We can't do AD Managed Identity from the Azure-hosted mac pools, so we don't have a password-less auth option for mac. Moving Mac to test against localhost honestly may get us more code coverage than testing Mac against Azure SQL DB. There are a lot of tests that are skipped against Azure SQL. If we want to do Mac + Azure in the future (honestly not a very high priority, in my mind since Linux + Azure is so close to Mac + Azure), we may be able to leverage a new Mac hosting service in MS that we are looking at for ODBC, which might eventually support AAD Managed Identity.

PR #2751, opened 2024-08-03
Code Cleanup | `null != x` => `x != null`

benrr101:
**Description**: In the second of a series of code cleanup PRs, this PR removes the Yoda conditions `null != x` and replaces it with `x != null`. This is the most requested PR from my second PR, so hopefully it will be appreciated :)

As with the previous PR, I have opted to only flip the operands and not replace the condition with `x is not null`. This is to make the changes as safe as possible, since `is not null` will potentially behave differently than `!= null`.

**AI Analysis**: I ran the patch file through a large language model to assess safety ~~but uhhhh, it didn't like me.~~ and it seems to approve of this PR
image

**Testing**: projecs build andif CI goes through we should be good

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Common/src/Microsoft/Data/ProviderBase/DbConnectionFactory.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Common/src/Microsoft/Data/ProviderBase/DbConnectionFactory.cs
@@ -55,7 +55,7 @@ namespace Microsoft.Data.ProviderBase
                 foreach (KeyValuePair<DbConnectionPoolKey, DbConnectionPoolGroup> entry in connectionPoolGroups)
                 {
                     DbConnectionPoolGroup poolGroup = entry.Value;
-                    if (null != poolGroup)
+                    if (poolGroup != null)
                     {
                         poolGroup.Clear();
                     }
@@ -69,7 +69,7 @@ namespace Microsoft.Data.ProviderBase
             using (TryEventScope.Create("<prov.DbConnectionFactory.ClearPool|API> {0}", GetObjectId(connection)))
             {
                 DbConnectionPoolGroup poolGroup = GetConnectionPoolGroup(connection);
-                if (null != poolGroup)
+                if (poolGroup != null)
                 {
                     poolGroup.Clear();
                 }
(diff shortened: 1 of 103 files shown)


Tornhoof:
> AI Analysis: I ran the patch file through a large language model to assess safety but uhhhh, it didn't like me. and it seems to approve of this PR

Your AI Analysis is not really correct afaik, technically they are different:

```x != null``` calls the inequality operator != with x on the left hand and ```null != x``` with x on the right hand.
If there are maniacs who implement != **that** wrong, i.e. with different null behavior for left hand vs. right hand the result will be different. You can run https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEUCuA7AHwAEAmARgFgAoIgZgAJT6Bhegb2vq8ZJPc+6CAsgE9mAGwCGAZ2n0E9ALz08MAO71REmdLYA1SeJwwl9MgF8A3AMFcAlgDMAFHhzjx9AITKEAShu2HFS2IYxkAJxOAESu7lG+1sGh5gGCjk4K3ipu4v5JIUGhtkQR0bHi8YlFKfm2qdzpCnZyeBAY2e55RYVF3CWRMTmV9Vw1gjU11HQ8mmJSstQ9XNN2eO0GRiZsAOYwGJbSe5Zj3AHTJUj0wBAQHhAADjBQkhjQSopOWvNyCGiz2rJ6CIutwlsUAOzyAB0G2M7yBMMMxiq3BOywYFyuNzuj2erygXg+Xx08j+xMBwICYMEI3ojTpLRyIIKtME/WiTUZcQStLRdVqaWcIgZHVytOpoXZUWFzVFwwFqNptKIkPkAH5EZtCUCNbCYCjRtRzEA=

for an example, to see the call difference.

That's why the whole ``is not null`` syntax is superior, it bypasses the inequality operator alltogether.

Having said that, the only way I know of, how ``x != null`` or ``null != x`` behaves differently than ``x is not null`` is, if someone overloads the inequality operator ``!=`` in the source code, a short look through the code via GitHub code search only shows one instance in a [test ](https://github.com/dotnet/SqlClient/blob/6fbbb787cd7ea35c06e292bb83d01ae94900b0bf/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/UdtTest/UDTs/Utf8String/Utf8String.cs#L148) and that one looks fine. I assume here that types in external libraries, e.g. BCL are fine.

Personally I'd recommend normalizing ``x != null`` `` null != x``  to ``is not null`` as it is either equally _wrong_ as your changes or equally _correct_, but infinitely more readable.

benrr101:
@Tornhoof I won't disagree too hard - you're right that it is possible someone wrote a `==`/`!=` overload that treats left and right nulls differently, and you're right that a person doing that would be a maniac. I slightly disagree that replacing with `x is not null` would be equally wrong vs `x != null`. The way I see it, there is three scenarios

* `!=` operator was overloaded to treat left and right `null` differently - maniacal, and would be broken with both replacements
* `!=` operator was overloaded to treat null differently - not ideal, but `is not null` would break it, `!= null` would not
* `!=` operator is not overloaded - `!= null` and `is not null` behave the same

It all likelihood, the last case is probably 90+% of the cases in the code, and in that case I would indeed advocate for using `is not null`. But, I want to take these changes one step at a time and try to break as little in the process as possible. I have a slew more of these cleanup PRs ready to go, and I plan to do one that addresses `== null` to `is null` at the same time as using conditional access and pattern matching (to some extent, not the `x is { foo: bar }` kind)

DavoudEshtehari (approved):

JRahnama:
I think, we can also have [RoslynAnalyzer](https://learn.microsoft.com/en-us/dotnet/csharp/roslyn-sdk/#enforce-team-coding-standards) to enforce some coding standards to prevent this in future as this cannot be done in editorconfig file.

PR #2758, opened 2024-08-07
Introducing type based Token parsing and processing

cheenamalhotra:
This PR adds the type based TDS token parsing workflow.

Note: Perf improvements are subject to benchmark testing, but there are design options available to us that we will revisit in the async callbacks to improve state machine handling.

Keeping that work out of scope from this PR.

**How will token parsing work?**
The new Parser uses TokenStreamHandler to read token type and all related data for each token, which will result in customized 'Token' object to be returned to the parser. This customized token object will then be handled back in the parser's state information using 'TokenProcessors'. This way we are separating token parsing from processing into TDS Context, keeping SRP intact.

Further perf improvements can be targeted in token Parsers, keeping processors untouched.


mdaigle (commented):
LGTM, just curious to see how all of the different state objects come together.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlConnectionX.cs
```diff
@@ -313,6 +333,35 @@ protected override DbBatch CreateDbBatch()
...
+            // SqlClientEventSource.Log.TryTraceEvent("SqlConnection.OnInfoMessage | API | Info | Object Id {0}, Message '{1}'", ObjectID, imevent.Message);
+            SqlInfoMessageEventHandler? handler = InfoMessage;
+            if (null != handler)
+            {
+                notified = true;
+                try
+                {
+                    handler(this, imevent);
+                }
+                catch (Exception e)
```
Tornhoof, line 351: Are exception filters allowed? They might make it more readable.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Tds/Tokens/LoginAck/LoginAckTokenParser.cs
```diff
@@ -0,0 +1,61 @@
...
+
+            byte type = await tdsStream.TdsReader.ReadByteAsync(isAsync, ct).ConfigureAwait(false);
+            SqlInterfaceType interfaceType = LoginAckTokenParser.GetSqlInterfaceType(type);
+
+            uint version = await tdsStream.TdsReader.ReadUInt32BEAsync(isAsync, ct).ConfigureAwait(false);
+            TdsVersion tdsVersion = LoginAckTokenParser.GetTdsVersion(version);
+            tdsStream.TdsVersion = tdsVersion;
+
+            string progName = await tdsStream.TdsReader.ReadBVarCharAsync(isAsync, ct).ConfigureAwait(false);
+            byte major = await tdsStream.TdsReader.ReadByteAsync(isAsync, ct).ConfigureAwait(false);
```
Tornhoof, line 26: I've read the disclaimer in the PR regarding benchmarks, I still have a few questions (this is generic, I'll just take this token as an example):
1. How many of the tokens (any token) exist per typical sql operation, I'm a bit concerned about the amount of objects created
2. Are there any tokens with only a single Read operation? because the whole ValueTask as return type is kinda moot, if there is always an await in there, the same applies to the ValueTasks in TdsReader.

Honestly, the amount of awaits here is frightening, we're talking about a protocol parser after all, there should be 0 (maybe 1) awaits in here besides a possible await for  VarCharAsync read as you know the minimum bytes required to build this token.

In a different PR I was told that the C# pipeline concept does not really apply to the usage of TDS, but with the token parser concept, can that previous explanation be revisited?

cheenamalhotra: Delegating this discussion to https://github.com/dotnet/SqlClient/discussions/2689, I'd request posting further comments on this topic there. We are considerate of the amount of awaits, and this is something we need to incrementally rework in the I/O design, using ByteBuffer or a similar solution by working with Buffer directly.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlConnectionX.cs
cheenamalhotra: The SqlInfoMessageEventHandler is implemented by user, so we capture generic exception here.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Tds/Tokens/LoginAck/LoginAckTokenParser.cs
edwardneal: I was going to wait for the PR to merge before I started to look at benchmarks, but to keep performance comments together, [this](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-tds/490e563d-cc6e-4c86-bb95-ef0186b98032) is the relevant section of the TDS specification for a LOGINACK token.

My first instinct would be to look at the token's Length field, then to fill a buffer with a sensible number of bytes up to that length, and allow the token to interpret it. I'm not sure whether there's much value in this though: the TDS reader already stores a 4KB TDS packet in memory, and it'll read synchronously from this packet if it can. An extra buffer on top of this seems like overkill. There might be value in providing a way to instruct the TDS reader to guarantee that a certain number of bytes are available in its buffer though, as this'd render the token parsing synchronous.

I agree with your point about creating one object per token; it needs benchmarking. We're not yet at the point of executing SQL statements, but I'm conscious that each row in a result set is also a ROW token...

cheenamalhotra: We're on the same page!
I agree with not creating any extra buffer and reading off the stream directly. The solution we need is certainly being able to ensure buffer can be prefilled, which is what we need to achieve somewhere outside the token stream handling in I/O layer. Because it doesn't exist as of now, these are async calls, I would imagine us to convert the 'ParseAsync' to a sync API in future, ensuring buffer is prefilled, and reading off it synchronously.

So please don't take these PRs as final designs as of now, these are incremental changes :)

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Tds/Tokens/EnvChange/EnvChangeTokenParser.cs
```diff
@@ -0,0 +1,124 @@
...
+            }
+
+            EnvChangeTokenSubType tokenSubType = (EnvChangeTokenSubType)subType;
+
+            if (!subTypeParsers.ContainsKey(tokenSubType))
+            {
+                throw new InvalidOperationException($"Unsupported EnvChange Token type: {tokenSubType}");
+            }
+
+            return await subTypeParsers[tokenSubType](tdsStream, isAsync, ct).ConfigureAwait(false);
```
benrr101, line 121: I'm a bit wary of using a dictionary of Funcs, defined with lambdas here. I think we could just as easily have a switch statement for the different sub types that calls the appropriate constructors rather than all the overhead of lambdas and lookups.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Tds/Tokens/EnvChange/EnvChangeTokenSubType.cs
```diff
@@ -0,0 +1,107 @@
...
+        /// </summary>
+        ResetConnection = 18,
+
+        /// <summary>
+        /// User instance name.
+        /// </summary>
+        UserInstanceName = 19,
+
+        /// <summary>
+        /// Routing.
```
benrr101, line 103: Personally I'd prefer it if there was a short explanation of what the token actually means rather than just restating the enum name with spaces.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Tds/Tokens/Error/ErrorTokenParser.cs
```diff
@@ -0,0 +1,36 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Data.SqlClientX.IO;
+
+namespace Microsoft.Data.SqlClientX.Tds.Tokens.Error
+{
+    internal class ErrorTokenParser : TokenParser
```
benrr101, line 11: comments :)

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Tds/Tokens/Error/ErrorTokenParser.cs
```diff
@@ -0,0 +1,36 @@
...
+            uint number = await tdsStream.TdsReader.ReadUInt32Async(isAsync, ct).ConfigureAwait(false);
+            byte state = await tdsStream.TdsReader.ReadByteAsync(isAsync, ct).ConfigureAwait(false);
+            byte severity = await tdsStream.TdsReader.ReadByteAsync(isAsync, ct).ConfigureAwait(false);
+
+            string message = await tdsStream.TdsReader.ReadUsVarCharAsync(isAsync, ct).ConfigureAwait(false);
+            string serverName = await tdsStream.TdsReader.ReadBVarCharAsync(isAsync, ct).ConfigureAwait(false);
+            string procName = await tdsStream.TdsReader.ReadBVarCharAsync(isAsync, ct).ConfigureAwait(false);
+
+            uint lineNumber;
+            if (tdsStream.TdsVersion < TdsVersion.V7_2)
```
benrr101, line 25: nit: This could be done as a ternary assignment

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Tds/Tokens/FeatureExtAck/FeatureExtAckTokenParser.cs
```diff
@@ -0,0 +1,34 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Data.SqlClientX.IO;
+
+namespace Microsoft.Data.SqlClientX.Tds.Tokens.FeatureExtAck
+{
+    internal sealed class FeatureExtAckTokenParser : TokenParser
```
benrr101, line 12: Comments :)

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Tds/Tokens/FedAuthInfo/FedAuthInfoToken.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// <summary>
+        /// Token type.
+        /// </summary>
+        public override TokenType Type => TokenType.FedAuthInfo;
+
+        /// <summary>
+        /// Service principal name. 
+        /// Can be null.
+        /// </summary>
+        public string SPN { get; }
```
benrr101, line 21: Properties should go after constructors

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Tds/Tokens/FedAuthInfo/FedAuthInfoTokenParser.cs
```diff
@@ -0,0 +1,50 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Data.SqlClientX.IO;
+
+namespace Microsoft.Data.SqlClientX.Tds.Tokens.FedAuthInfo
+{
+    internal class FedAuthInfoTokenParser : TokenParser
```
benrr101, line 12: Comments :)

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Tds/Tokens/Info/InfoTokenParser.cs
```diff
@@ -0,0 +1,38 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Data.SqlClientX.IO;
+
+namespace Microsoft.Data.SqlClientX.Tds.Tokens.Info
+{
+    internal class InfoTokenParser : TokenParser
```
benrr101, line 11: Comments :)

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Tds/Tokens/LoginAck/LoginAckTokenParser.cs
```diff
@@ -0,0 +1,61 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Data.SqlClientX.IO;
+
+namespace Microsoft.Data.SqlClientX.Tds.Tokens.LoginAck
+{
+    internal class LoginAckTokenParser : TokenParser
```
benrr101, line 12: Comments

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Tds/Tokens/LoginAck/LoginAckTokenParser.cs
```diff
@@ -0,0 +1,61 @@
...
+        public override async ValueTask<Token> ParseAsync(TokenType tokenType, TdsStream tdsStream, bool isAsync, CancellationToken ct)
+        {
+            _ = await tdsStream.TdsReader.ReadUInt16Async(isAsync, ct).ConfigureAwait(false); // length
+
+            byte type = await tdsStream.TdsReader.ReadByteAsync(isAsync, ct).ConfigureAwait(false);
+            SqlInterfaceType interfaceType = LoginAckTokenParser.GetSqlInterfaceType(type);
+
+            uint version = await tdsStream.TdsReader.ReadUInt32BEAsync(isAsync, ct).ConfigureAwait(false);
+            TdsVersion tdsVersion = LoginAckTokenParser.GetTdsVersion(version);
+            tdsStream.TdsVersion = tdsVersion;
```
benrr101, line 23: Why not directly assign?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Tds/Tokens/ReturnStatus/ReturnStatusTokenParser.cs
```diff
@@ -0,0 +1,16 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Data.SqlClientX.IO;
+
+namespace Microsoft.Data.SqlClientX.Tds.Tokens.ReturnStatus
+{
+    internal class ReturnStatusTokenParser : TokenParser
```
benrr101, line 11: Comments

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Tds/Tokens/TokenStreamHandler.cs
```diff
@@ -0,0 +1,82 @@
...
+using Microsoft.Data.SqlClientX.Tds.Tokens.Error;
+using Microsoft.Data.SqlClientX.Tds.Tokens.FeatureExtAck;
+using Microsoft.Data.SqlClientX.Tds.Tokens.FedAuthInfo;
+using Microsoft.Data.SqlClientX.Tds.Tokens.Info;
+using Microsoft.Data.SqlClientX.Tds.Tokens.LoginAck;
+using Microsoft.Data.SqlClientX.Tds.Tokens.ReturnStatus;
+
+namespace Microsoft.Data.SqlClientX.Tds.Tokens
+{
+    internal sealed class TokenStreamHandler
```
benrr101, line 28: Comments

benrr101 (changes requested):

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Tds/Tokens/DataClassification/DataClassificationTokenParser.cs
```diff
@@ -0,0 +1,115 @@
...
+        {
+            if (tdsStream.DataClassificationVersion == 0)
+            {
+                throw SQL.ParsingError(ParsingErrorState.DataClassificationNotExpected);
+            }
+
+            // get the labels
+            ushort numLabels = await tdsStream.TdsReader.ReadUInt16Async(isAsync, ct).ConfigureAwait(false);
+
+            List<Label> labels = new List<Label>(numLabels);
```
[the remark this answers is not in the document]
saurabh500: LIst translate well to the ReadOnlyCollection<Label> on SqlDataReader's API which exposes classification information. This is OK to maintain as a list.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Tds/Tokens/DoneProc/DoneProcTokenParser.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        public override async ValueTask<Token> ParseAsync(TokenType tokenType, TdsStream tdsStream, bool isAsync, CancellationToken ct)
+        {
+            ushort status = await tdsStream.TdsReader.ReadUInt16Async(isAsync, ct).ConfigureAwait(false);
+            DoneStatus doneStatus = (DoneStatus)status;
+
+            ushort currentCommand = await tdsStream.TdsReader.ReadUInt16Async(isAsync, ct).ConfigureAwait(false);
+
+            ulong rowCount;
+            if (tdsStream.TdsVersion > TdsVersion.V7_2)
```
saurabh500, line 22: Why is TdsVersion stored with the stream? Should it not be closer to the connection/internal connection? 
Ideally streams should not hold the meaning that is associated with the connection like version information or other connection capabilities.

saurabh500: Also in login we send a Tds version of 7.4 are we ever going to get into the `else` block?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Tds/Tokens/EnvChange/EnvChangeTokenParser.cs
saurabh500: I agree with Ben, however if you must use a dictionary then use the TryGet method. This way you can avoid 2 lookups at 116 and 121.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Tds/Tokens/EnvChange/EnvChangeTokenSubType.cs
```diff
@@ -0,0 +1,107 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.Data.SqlClientX.Tds.Tokens.EnvChange
+{
+    /// <summary>
+    /// Environment change token sub type.
+    /// </summary>
+    internal enum EnvChangeTokenSubType : byte
```
saurabh500, line 10: This could be called `EnvChangeType` for simplicity.

saurabh500:
For env change types, do we need a new type for every env change with old and new value? Since the `EnvChangeToken<T>` could itself be enough to solve the purpose. This can be dereferenced using the right variable name, and ultimate discarded. 

For each token, there seems to be a parser class. Can't the Token class itself have a `static Self Decode(stream, isAsync ...)` method, so that it can decode itself from the stream and provide the token.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Tds/State/TdsCommandContext.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.Data.SqlClientX.Tds.State
+{
+    internal class TdsCommandContext
```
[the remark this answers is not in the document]
cheenamalhotra: This came because I was trying to separate context info for command/connection.. This is not used by parser as of now, because I'm only writing it for login flow, but in future this can be extended and used by parser. Will remove it as it has no purpose as of now.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Tds/State/TdsContext.cs
```diff
@@ -0,0 +1,48 @@
...
+using System.Collections.Generic;
+using Microsoft.Data.SqlClient;
+using Microsoft.Data.SqlClientX.IO;
+
+namespace Microsoft.Data.SqlClientX.Tds.State
+{
+    /// <summary>
+    /// Captures context information required to run TDS parser operations.
+    /// </summary>
+    internal class TdsContext
```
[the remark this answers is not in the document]
cheenamalhotra: It's going to be used to capture login response data from server that parser is reading off stream, and send it back to SqlConnector who will own the TdsContext.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Tds/State/TdsErrorWarningsState.cs
```diff
@@ -0,0 +1,106 @@
...
+
+        /// <summary>
+        /// Adds an error to the error collection
+        /// </summary>
+        /// <param name="error"></param>
+        internal void AddError(SqlError error)
+        {
+            Debug.Assert(error != null, "Trying to add a null error");
+
+            lock (_errorAndWarningsLock)
```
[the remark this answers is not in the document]
cheenamalhotra: This came from existing design, we can certainly remove the locks now and make concurrent collections.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Tds/Tokens/DataClassification/DataClassificationTokenParser.cs
```diff
@@ -0,0 +1,115 @@
...
+                var label = await tdsStream.TdsReader.ReadStringAsync(labelLen, isAsync, ct).ConfigureAwait(false);
+                var idLen = await tdsStream.TdsReader.ReadByteAsync(isAsync, ct).ConfigureAwait(false);
+                var id = await tdsStream.TdsReader.ReadStringAsync(idLen, isAsync, ct).ConfigureAwait(false);
+                labels.Add(new Label(label, id));
+            }
+
+            // get the information types
+            ushort numInformationTypes = await tdsStream.TdsReader.ReadUInt16Async(isAsync, ct);
+
+            List<InformationType> informationTypes = new List<InformationType>(numInformationTypes);
```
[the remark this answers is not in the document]
cheenamalhotra: These are publicly exposed as list too, so this will be retained.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Tds/Tokens/DataClassification/DataClassificationTokenParser.cs
```diff
@@ -0,0 +1,115 @@
...
+                var rank = await tdsStream.TdsReader.ReadInt32Async(isAsync, ct).ConfigureAwait(false);
+                if (!Enum.IsDefined(typeof(SensitivityRank), rank))
+                {
+                    return null;
+                }
+            }
+
+            // get the per column classification data (corresponds to order of output columns for query)
+            var numResultColumns = await tdsStream.TdsReader.ReadUInt16Async(isAsync, ct).ConfigureAwait(false);
+            List<ColumnSensitivity> columnSensitivities = new List<ColumnSensitivity>(numResultColumns);
```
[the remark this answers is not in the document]
cheenamalhotra: We're not changing public signature, so this will stay.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/IO/ByteBuffer.cs
```diff
@@ -0,0 +1,934 @@
...
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Microsoft.Data.SqlClientX.IO
+{
+    /// <summary>
+    /// Buffer to handle low-level byte operations.
+    /// </summary>
+    internal class ByteBuffer : IEnumerable<byte>, IEquatable<ByteBuffer>
```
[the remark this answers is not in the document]
cheenamalhotra: I needed a way to read Bytes off the stream in a buffer, therefore this came in here. This is temporary, and should not be needed later once we can support direct TdsStream buffer reading.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Tds/Tokens/DataClassification/DataClassificationToken.cs
```diff
@@ -0,0 +1,45 @@
...
+{
+    /// <summary>
+    /// Token indicating data classification information received for classified columns.
+    /// </summary>
+    internal class DataClassificationToken : Token
+    {
+        /// <summary>
+        /// Token type.
+        /// </summary>
+        public override TokenType Type => TokenType.DataClassification;
```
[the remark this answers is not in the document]
cheenamalhotra: I am storing mappings of types to their parser implementations to fire them, therefore these types are needed.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Tds/Tokens/DataClassification/DataClassificationTokenParser.cs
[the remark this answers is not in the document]
benrr101: What public signatures? The token and token parser classes are all internal and implementation details of how tokens are parsed.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Tds/State/TdsContext.cs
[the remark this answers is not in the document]
benrr101: Is your plan still to have the tds parser be a static/stateless class? If not, then the the data from the server can be stored in the parser, no need for context objects. If it is ... then I think we really need to rediscuss why we're taking this approach.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Tds/Tokens/DataClassification/DataClassificationToken.cs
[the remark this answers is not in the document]
benrr101: Although I have other comments about having mappings between token types and parsers (see below), I think that's a different situation. What I'm suggesting here is that we don't need to have a token type stored in each instance of the token. The token has already been parsed, and any other switching that we need to do in the parser, etc, could be just as easily done by checking class type.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Tds/Tokens/DataClassification/DataClassificationTokenParser.cs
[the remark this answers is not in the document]
benrr101: Ok, I see what you mean, but the constructor for ReadOnlyCollection takes `IList`, which `Label[]` implements. So using an array would be a-ok, and even if we couldn't use an array, we should still clamp the size of the list down to the number of elements.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Tds/Tokens/DataClassification/DataClassificationTokenParser.cs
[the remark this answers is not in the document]
benrr101: See previous comment response

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Tds/Tokens/DataClassification/DataClassificationTokenParser.cs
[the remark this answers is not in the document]
benrr101: Ok, I see what you mean by having ReadOnlyCollection as part of the interface, but ReadOnlyCollection can be initialized from an array (since it implements IList)

saurabh500:
@cheenamalhotra comments aside, I want to test it to parse out the Login response at the end of the login handler. 

Can you provide the sample snippet that I can try out with the LoginHandler to try out this change ?

cheenamalhotra:
> @cheenamalhotra comments aside, I want to test it to parse out the Login response at the end of the login handler.
> 
> Can you provide the sample snippet that I can try out with the LoginHandler to try out this change ?

@saurabh500 LoginChainTest now works with Password provided. Do remember all the server response is being read, but context may not be updated perfectly post reading server response. This is somewhere we can play with managing Context information better, if any rework is needed. As of now what I've added is proposal for organizing context information, but feel free to suggest any reorganization/renaming as needed.

edwardneal (commented):
Setting aside the discussion going on about I/O and performance, a quick sense check. In most cases, I've just checked the enum values and semantics against the TDS specification. It looks good to me though - looking forward to testing it!

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/Tds/Tokens/DataClassification/DataClassificationTokenParser.cs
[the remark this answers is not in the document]
JRahnama: @benrr101, @cheenamalhotra, I spent sometime on this part and I was thinking if we could optimized the loop by starting the string reads immediately after reading the byte lengths for both label and ID.
This reduces the need for separate await calls and allows the label and ID reading operations to be performed in parallel,
potentially improving performance, especially if these operations are I/O-bound (which they likely are since they involve reading from a stream).

Have not tested, just as an idea, something similar to below, again I might be totally wrong, but thought it might worth sharing that with you as well.
```C#
for (ushort i = 0; i < numLabels; i++)
{
    byte labelLen = await tdsStream.TdsReader.ReadByteAsync(isAsync, ct).ConfigureAwait(false);
    ValueTask<string> labelTask = tdsStream.TdsReader.ReadStringAsync(labelLen, isAsync, ct);

    byte idLen = await tdsStream.TdsReader.ReadByteAsync(isAsync, ct).ConfigureAwait(false);
    ValueTask<string> idTask = tdsStream.TdsReader.ReadStringAsync(idLen, isAsync, ct);

    labels.Add(new Label(await labelTask.ConfigureAwait(false), await idTask.ConfigureAwait(false)));
}
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClientX/SqlConnector.cs
```diff
@@ -16,17 +18,34 @@ namespace Microsoft.Data.SqlClientX
...
             OwningConnection = owningConnection;
             DataSource = dataSource;
+
             //TODO: Set this based on the real server process id.
             //We only set this in client code right now to simulate different processes and to differentiate internal connections.
-            ServerProcessId = Interlocked.Increment(ref SpoofedServerProcessId);
+            ServerProcessId = Interlocked.Increment(ref s_spoofedServerProcessId);
+
+            // TODO enable parser registration with Parser introduction.
+            var connString = new SqlConnectionString(dataSource.ConnectionString);
```
JRahnama, line 37: Can we use explicit types instead of var? Occurred in several places

Issue #2378, opened 2024-03-01
Errors after upgrading to 5.2.0 from 5.1.5 on Linux


alex-jitbit:
After upgrading 5.1.5 to 5.2.0 on Linux (Ubuntu) .NET 8 I'm getting thousands of errors:
```
Exception message:

Microsoft.Data.SqlClient.SqlException (0x80131904): A network-related or instance-specific error occurred while establishing a connection to SQL Server. The server was not found or was not accessible. Verify that the instance name is correct and that SQL Server is configured to allow remote connections. (provider: TCP Provider, error: 35 - An internal exception was caught)
 ---> System.TimeoutException: The socket couldn't connect during the expected 14965 remaining time.

Stack trace:

Microsoft.Data.SqlClient.SqlException (0x80131904): A network-related or instance-specific error occurred while establishing a connection to SQL Server. The server was not found or was not accessible. Verify that the instance name is correct and that SQL Server is configured to allow remote connections. (provider: TCP Provider, error: 35 - An internal exception was caught)
 ---> System.TimeoutException: The socket couldn't connect during the expected 14965 remaining time.
   at Microsoft.Data.SqlClient.SNI.SNITCPHandle.Connect(String serverName, Int32 port, TimeoutTimer timeout, SqlConnectionIPAddressPreference ipPreference, String cachedFQDN, SQLDNSInfo& pendingDNSInfo)
   at Microsoft.Data.SqlClient.SNI.SNITCPHandle..ctor(String serverName, Int32 port, TimeoutTimer timeout, Boolean parallel, SqlConnectionIPAddressPreference ipPreference, String cachedFQDN, SQLDNSInfo& pendingDNSInfo, Boolean tlsFirst, String hostNameInCertificate, String serverCertificateFilename)
   at Microsoft.Data.ProviderBase.DbConnectionPool.TryGetConnection(DbConnection owningObject, UInt32 waitForMultipleObjectsTimeout, Boolean allowCreate, Boolean onlyOneCheckConnection, DbConnectionOptions userOptions, DbConnectionInternal& connection)
   at Microsoft.Data.ProviderBase.DbConnectionPool.TryGetConnection(DbConnection owningObject, TaskCompletionSource`1 retry, DbConnectionOptions userOptions, DbConnectionInternal& connection)
[...]

PR #2098, opened 2023-07-21
Dev | Improved connection timeout

DavoudEshtehari:
- Improved the connection timeout accuracy, by involving the IP address resolution and socket connection retry process on Managed SNI.
- Expanded using the timeout timer on managed code.
- Improved connection timeout tests.

PR #2777, opened 2024-08-15
Fix | Fix the issue with Socke.Connect in managed SNI

JRahnama:
Fixes issue #2378 

PR #2098 improved the handling of timeouts, but there were a couple of asynchronous tasks in a synchronous flow that were intended to respect these timeouts. The change in `SNITCPHandle.Connect` introduced a regression. It seems that the entire asynchronous task is unnecessary. The socket is already set to non-blocking mode, so `Socket.Connect` will not block, and we are subsequently looping with `socket.Select` to check if the connection is established within the timeout period.

If the fix is approved and merged it will need to be backported to 5.2 servicing branch.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNICommon.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNICommon.cs
@@ -292,20 +292,14 @@ namespace Microsoft.Data.SqlClient.SNI
 
         internal static IPAddress[] GetDnsIpAddresses(string serverName, TimeoutTimer timeout)
         {
-            using (TrySNIEventScope.Create(nameof(SNICommon)))
+            IPAddress[] ipAddresses = GetDnsIpAddresses(serverName);
+
+            // We cannot timeout accurately in sync code above, so throe TimeoutException if we've now exceeded the timeout.
+            if (timeout.IsExpired)
             {
-                int remainingTimeout = timeout.MillisecondsRemainingInt;
-                SqlClientEventSource.Log.TrySNITraceEvent(nameof(SNICommon), EventType.INFO,
-                                                          "Getting DNS host entries for serverName {0} within {1} milliseconds.",
-                                                          args0: serverName,
-                                                          args1: remainingTimeout);
-                using CancellationTokenSource cts = new CancellationTokenSource(remainingTimeout);
-                // using this overload to support netstandard
-                Task<IPAddress[]> task = Dns.GetHostAddressesAsync(serverName);
-                task.ConfigureAwait(false);
-                task.Wait(cts.Token);
-                return task.Result;
+                throw new TimeoutException();
             }
+            return ipAddresses;
         }
 
         internal static IPAddress[] GetDnsIpAddresses(string serverName)
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNICommon.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNICommon.cs
@@ -294,7 +294,7 @@ namespace Microsoft.Data.SqlClient.SNI
         {
             IPAddress[] ipAddresses = GetDnsIpAddresses(serverName);
 
-            // We cannot timeout accurately in sync code above, so throe TimeoutException if we've now exceeded the timeout.
+            // We cannot timeout accurately in sync code above, so throw TimeoutException if we've now exceeded the timeout.
             if (timeout.IsExpired)
             {
                 throw new TimeoutException();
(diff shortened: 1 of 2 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNICommon.cs
```diff
@@ -292,20 +292,14 @@ internal static bool ValidateSslServerCertificate(Guid connectionId, string targ
 
         internal static IPAddress[] GetDnsIpAddresses(string serverName, TimeoutTimer timeout)
         {
-            using (TrySNIEventScope.Create(nameof(SNICommon)))
+            IPAddress[] ipAddresses = GetDnsIpAddresses(serverName);
+
+            // We cannot timeout accurately in sync code above, so throe TimeoutException if we've now exceeded the timeout.
```
saurabh500, line 297:
```suggestion
            // We cannot timeout accurately in sync code above, so throw TimeoutException if we've now exceeded the timeout.
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
```diff
@@ -509,11 +492,11 @@ private static Socket Connect(string serverName, int port, TimeoutTimer timeout,
                         return socket;
                     }
                 }
-                catch (AggregateException aggregateException) when (aggregateException.InnerException is SocketException socketException)
+                catch (SocketException e)
```
saurabh500, line 495: Now that you have removed the capability of timing out when we fail to get an open socket at the end of the timeout duration (which was enforced by a Task timeout), would it make sense to check if the socket is not connected and we have exceeded the timeout. Hence throw an exception?

saurabh500 (commented):
The changes look OK to me.
However my

saurabh500:
Overall, I think the fix makes sense. It takes away the timeout capability on failing to get a socket at the end of timeout duration remaining. Is it possible to check for timeout after the `do` `while` loop to see if we dont have a connected socket, and past the timeout value, and throw an exception? (unless it is already happening and I am missing something 😄 )

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
David-Engel: I had a hard time groking the logic here. But it does time out with an error. When socket.Blocking is set to true (line 413), socket.Connect (line 429) will throw a SocketException immediately and socketException.SocketErrorCode  will be SocketError.WouldBlock. We loop inside the catch where socket.Select (line 469) does accept a timeout. socket.Select will proceed either at a successful connect (something selected) or timeout. If timed out, the selected list will be empty and the loop repeats. If the timeout is expired, we return a null Socket (line 455), which is handled (report SNI error) in the calling methods.

David-Engel:
> Overall, I think the fix makes sense. It takes away the timeout capability on failing to get a socket at the end of timeout duration remaining. Is it possible to check for timeout after the `do` `while` loop to see if we dont have a connected socket, and past the timeout value, and throw an exception? (unless it is already happening and I am missing something 😄 )

It's not an obvious flow, but timeout still works in the logic. See my other comment. 😄

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
JRahnama: @saurabh500 when we have  `!isInfiniteTimeout` no matter what we throw an exception of type SocketException with error code wouldBlock then inside catch block Socket.Select will check for the timeout expiration. This was suggested by @roji [here](https://github.com/dotnet/SqlClient/issues/826#issuecomment-736224118) since Socket.Connect (sync) does not support timeout.

saurabh500: Ah got it. Then why did we need the original change to timeout with Tasks? Is it to have more high fidelity timeout? 

> When socket.Blocking is set to true (line 413), socket.Connect (line 429) will throw a SocketException immediately

Did you mean this to be the other way around? socket.Blocking = false should lead to an immediate exception. Or that's what I understand. This means that "hey socket do not block any calls, but keep trying and I will come back and check later using Select", and the socket responds by throwing an exception immediately with the error code specifying the reason, so that the caller can omit the exception, and move on to the select.

saurabh500: @JRahnama  I understand the reasoning behind teh SocketException with WouldBlock and that makes sense to me. My question was different

David-Engel: > Ah got it. Then why did we need the original change to timeout with Tasks? Is it to have more high fidelity timeout?

My assumption is the previous changes were made because of a misunderstanding of the logic, thinking it wasn't timing out. The original PR included many other changes improving the timeout logic in SNI.

> 
> > When socket.Blocking is set to true (line 413), socket.Connect (line 429) will throw a SocketException immediately
> 
> Did you mean this to be the other way around? socket.Blocking = false should lead to an immediate exception. Or that's what I understand. This means that "hey socket do not block any calls, but keep trying and I will come back and check later using Select", and the socket responds by throwing an exception immediately with the error code specifying the reason, so that the caller can omit the exception, and move on to the select.

Oops, yes. Fixed.

David-Engel (approved):

saurabh500 (approved):

PR #2801, opened 2024-08-21
Add streaming support for netcore and netfx

apoorvdeshmukh:
This PR adds support for streaming JSON datatype to and from SQL Server.

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlEnums.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlEnums.cs
@@ -124,6 +124,7 @@ namespace Microsoft.Data.SqlClient
             type == SqlDbType.Char ||
             type == SqlDbType.VarChar ||
             type == SqlDbType.Text ||
+            type == SqlDbTypeExtensions.Json ||
             type == SqlDbType.Xml;
 
         private static bool _IsNCharType(SqlDbType type) =>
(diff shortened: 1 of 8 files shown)


src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/JsonTest/JsonStreamTest.cs
```diff
@@ -0,0 +1,146 @@
...
+            {
+                using (StreamReader jsonFile = File.OpenText(jsonFilename))
+                {
+                    cmd.Parameters.Add("@jsondata", Microsoft.Data.SqlDbTypeExtensions.Json, -1).Value = jsonFile;
+                    cmd.ExecuteNonQuery();
+                }
+            }
+        }
+
+        private void deleteFile(string filename)
```
saurabh500, line 111: All C# functions start with Upper case letters

deepaksa1:
Does the test cases cover `SqlDataReader.GetStream()` ?

apoorvdeshmukh:
> Does the test cases cover `SqlDataReader.GetStream()` ?

As per the [docs](https://learn.microsoft.com/en-us/dotnet/api/microsoft.data.sqlclient.sqldatareader.getstream?view=sqlclient-dotnet-standard-5.2) GetStream() API is only supported for binary, image, varbinary and udt datatypes hence not required as part of the current testing.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlEnums.cs
```diff
@@ -124,6 +124,7 @@ private static bool _IsCharType(SqlDbType type) =>
             type == SqlDbType.Char ||
             type == SqlDbType.VarChar ||
             type == SqlDbType.Text ||
+            type == SqlDbTypeExtensions.Json ||
```
deepaksa1, line 127: Is the usual read operation working as expected after adding this? (We had a problem earlier in read which was fixed by removing this)

apoorvdeshmukh: Yes. The read issue was caused due to the existing code path trying to find the encoding and collation for char datatypes. The fix for the same is already in place for read/write changes.

deepaksa1 (approved):
LGTM

saurabh500 (approved):

PR #2577, opened 2024-06-17
Change | Remove some unneeded references and update Azure.Identity

ErikEJ:
fixes #2048
fixes #2574
fixes #2568 

removes explicit ref to Microsoft.Identity.Client

PR #2808, opened 2024-08-22
Update Azure dependencies to latest stable versions

SimonCropp:
* Azure.Identity 1.12.0
 * Azure.Core 1.42.0
 * Azure.Security.KeyVault.Keys 4.6.0


JRahnama:
I was wondering if we could apply the same approach to Azure.Core that @ErikEJ used for Microsoft.Identity.Client in PR #2577.

cheenamalhotra:
@DavoudEshtehari Since the version bumps are minor versions - can this be tagged as a non-breaking change? Unless you want to elaborate your understanding here?

DavoudEshtehari:
Thanks @cheenamalhotra, there was confusion due to a major version bump of a transient dependency affecting versions 4.0 and earlier, which does not apply here.

cheenamalhotra:
@SimonCropp 

Please update as below (latest):
* Azure.Identity v1.12.1
* Azure.Core v1.44.0

PR #2812, opened 2024-08-23
Merge DbConnectionPool

edwardneal:
Contributes to #1261.

DbConnectionPool is almost identical between .NET Core and Framework. When combined with #2410, this should mean that everything in the ProviderBase folders besides DbConnectionPoolCounters can be merged.

Core differences here:
* In DbConnectionPool.CleanupCallback, Semaphore.WaitOne used an overload with an exitContext parameter in Framework but not in Core. This parameter seems to be ignored in .NET Core, but I've included the change for the sake of .NET Framework.
* DbConnectionPool.CreateCleanupTimer returned ADP.UnsafeCreateTimer in .NET Core but created a Timer instance directly in .NET Framework. I've adopted the .NET Core methodology - the comment for UnsafeCreateTimer indicates that it's used to prevent the current ExecutionContext from flowing into a global timer and not being GCed, which seems reasonable.

More generally, there are a lot of conditionally-compiled blocks to enable the differing metrics implementations. Hopefully these will cease to exist when the OTel metrics are implemented.


src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbConnectionPool.cs
```diff
@@ -453,10 +459,7 @@ internal DbConnectionPool(
             _pooledDbAuthenticationContexts = new ConcurrentDictionary<DbConnectionPoolAuthenticationContextKey, DbConnectionPoolAuthenticationContext>(concurrencyLevel: 4 * Environment.ProcessorCount /* default value in ConcurrentDictionary*/,
                                                                                                                                                         capacity: 2);
 
-            if (ADP.s_isPlatformNT5)
-            {
-                _transactedConnectionPool = new TransactedConnectionPool(this);
-            }
+            _transactedConnectionPool = new TransactedConnectionPool(this);
```
benrr101, line 462: I guess this is pretty much given nowadays, right?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbConnectionPool.cs
```diff
@@ -1292,13 +1316,16 @@ private bool TryGetConnection(DbConnection owningObject, uint waitForMultipleObj
...
                     RuntimeHelpers.PrepareConstrainedRegions();
+#endif
                     try
                     {
+#if NETFRAMEWORK
                         // We absolutely must have the value of waitResult set, 
                         // or we may leak the mutex in async abort cases.
                         RuntimeHelpers.PrepareConstrainedRegions();
+#endif
                         try
```
benrr101, line 1329: What is the point of an empty try block?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -50,6 +50,16 @@ internal struct ReliabilitySection
             internal static void Assert(string message)
             {
             }
+
+            [Conditional("NETFRAMEWORK")]
```
benrr101, line 54: I'm a bit confused why something related to netfx is being added to the netcore project...

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/AdapterUtil.cs
```diff
@@ -1330,12 +1330,6 @@ internal static Task<T> CreatedTaskWithCancellation<T>()
             return completion.Task;
         }
 
-        internal static void TraceExceptionForCapture(Exception e)
```
benrr101, line 1333: What prompted removal of this one?

benrr101 (changes requested):
Overall looks really good! I just want to know what's up with the Start/Stop methods before signing off.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbConnectionPool.cs
edwardneal: Exactly, yes - every supported version of .NET Framework / .NET Core requires NT 5.0+.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbConnectionPool.cs
edwardneal: In .NET Core, it's a no-op. In .NET Framework, it's a Constrained Execution Region.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/AdapterUtil.cs
edwardneal: Purely neatness - it's no longer referenced in either project.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
edwardneal: The .NET Framework code's `WaitForPendingOpen` method uses the ReliabilitySection struct, and calls Start and Stop. The ReliabilitySection struct already exists in both [netcore](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs#L44) and [netfx](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs#L89), so I added two methods to it. The `[Conditional]` attribute on the method means that its call will always be compiled out.

ReliabilitySection in general will always no-op in .NET Core - ReliabilitySection sets up Constrained Execution Regions, which only do anything in .NET Framework.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbConnectionPool.cs
benrr101: Interesting 🤔 I'll have to give this a deeper read.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/AdapterUtil.cs
benrr101: That's the best reason 👍

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
benrr101: Ok - I've got merging TdsParser on my stack ... and it should be "fun". No problem with this for now.

benrr101 (approved):

benrr101:
Looks good, once the merge conflicts are fixed, I'll get it merged

edwardneal:
Thanks @benrr101 - merge conflicts are now addressed and CI's running.

cheenamalhotra (approved):

benrr101:
Thanks again @edwardneal this one unblocked me to work on DbConnectionFactory and DbConnectionInternal

edwardneal:
Great! If it saves any effort, I think DbConnectionFactory is already merged. I did take a quick look at DbConnectionInternal, but found it slightly frustrating - I didn't see any deal-breaking differences, but didn't get an easily reviewable PR because the .NET Core version was split between two files.

Issue #2671, opened 2024-07-12
SNI package does not respect `RuntimeIdentifier` property


julealgon:
### Describe the bug
When assigning a specific `RuntimeIdentifier` during `publish`, NuGet packages are supposed to only copy the required DLLs for the affected platforms/runtimes.

For example, if I specify `win-x64` as the RID, I should not expect to see `x86` and `arm` versions of DLLs in the output.

However, the `Microsoft.Data.SqlClient.SNI` package does not respect that, and always copies all of the native versions to the target output folder, namely:
- `Microsoft.Data.SqlClient.SNI.x86.dll`
- `Microsoft.Data.SqlClient.SNI.x64.dll`
- `Microsoft.Data.SqlClient.SNI.arm64.dll`

This creates unnecessary clutter in my applications, increasing the deployment package size.

### To reproduce
1. Create a brand new console application.
2. Add a dependency to either `Microsoft.Data.SqlClient` or `Microsoft.Data.SqlClient.SNI` packages.
3. `publish` the project using RID `win-x64`
4. Observe `x86` and `amd64` native dependencies in the output

### Expected behavior
Only the DLLs that match the runtime/architecture of the project should be present in the output if the setting is explicitly defined. It should only contain all possibilities if publishing without any specific architecture in mind.

### Further technical details
N/A

**Additional context**
I believe this issue exists because the package does not declare its native dependencies using the proper RID folder structure, and instead follows what seems to be a completely custom approach to copying the native dependencies.

It looks to me that moving to the intended `runtime` folder with the runtime identifiers inside would be the way to go. Here is another dependency we use that uses that format:

![image](https://github.com/user-attachments/assets/e7723b31-90ad-47fb-8b0e-bfd9787e877e)
https://www.nuget.org/packages/Grpc.Core

JRahnama:
@julealgon thank you for bringing this up. We will take a closer look to SNI.targets file and will get back to you.

Zippy22226:
0101101010000010011/0101001//0010101001
000101_001010010110-_01001010101_/01011

cheenamalhotra:
Hi @julealgon 

The SNI package is designed to load all DLLs during compilation and loads the applicable DLL for respective architecture at runtime in running process.

Closing this as by-design.

julealgon:
I don't understand why you still can't leverage the `RuntimeIdentifier` mechanism though @cheenamalhotra ...

[...]

PR #2828, opened 2024-09-01
Remove unused PInvokes

edwardneal:
This is the first step of a process to merge PInvokes between the .NET and .NET Framework projects. I've started by removing most of the unused PInvokes from both projects. There shouldn't be any surprises in CI.

I've also changed SqlUtil's `SocketDidNotThrow` method to throw an Exception rather than an InternalException. This was the only reference to InternalException which for the .NET 8.0 target, so this change means that when .NET 6.0 support is dropped, everything in the netcore/src/Common/src/System folder except for NotImplemented.cs can be removed.

The next step of the PInvoke merge process will be more involved. What's the coding standard for this? dotnet/runtime uses one folder per OS, one subfolder per DLL, one file per method, and there are quite a few files which fit this pattern. Other parts of the .NET project have a single file containing every PInvoke for a single DLL (the SNI DLL.)

Step 2 will be to merge the OS PInvokes which are needed for .NET Framework and .NET 8.0, and step 3 is currently to address the native SNI PInvokes. At the point of step 3, I'd also like to change the .NET Framework project to refer to Microsoft.Data.SqlClient.SNI.runtime, so it matches the .NET project. This might depend on #2671.


src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/Interop/SNINativeMethodWrapper.cs
```diff
@@ -397,9 +397,6 @@ internal struct SNI_Error
         #endregion
 
         #region DLL Imports
-        [DllImport("kernel32.dll", SetLastError = true)]
-        public static extern IntPtr LoadLibrary(string dllToLoad);
-
         [DllImport("secur32.dll", ExactSpelling = true, SetLastError = true)]
         internal static extern uint QueryContextAttributes(ref CredHandle contextHandle, [In] ContextAttribute attribute, [In] IntPtr buffer);
```
JRahnama, line 401: This method, along with all its references, can be deleted as it is a leftover from the old CertAuthentication implementation, which is no longer in use. The entire section in TDSParser that validates the server certificate can also be removed.
```C#
// Validate server certificate
if (serverCallback != null)
{
    X509Certificate2 serverCert = null;

    error = SNINativeMethodWrapper.SNISecGetServerCertificate(_physicalStateObj.Handle, ref serverCert);
    if (error != TdsEnums.SNI_SUCCESS)
    {
        _physicalStateObj.AddError(ProcessSNIError(_physicalStateObj));
        ThrowExceptionAndWarning(_physicalStateObj);
    }

    bool valid = serverCallback(serverCert);
    if (!valid)
    {
        throw SQL.InvalidServerCertificate();
    }
}
```

edwardneal: Are there definitely no internal builds with ADONET_CERT_AUTH enabled? If there aren't, I'll remove that implementation in a separate PR so that this one's easier to review.

Does the same apply to ADONET_ORIGINAL_CLIENT_ADDRESS and CTAIP support?

JRahnama: > Are there definitely no internal builds with ADONET_CERT_AUTH enabled? If there aren't, I'll remove that implementation in a separate PR so that this one's easier to review.
> 
> Does the same apply to ADONET_ORIGINAL_CLIENT_ADDRESS and CTAIP support?

All ADONET_CERT_AUTH related items can be safely removed. Thanks, that would be a big help—this has been on our to-do list for a long time.

JRahnama: > Does the same apply to ADONET_ORIGINAL_CLIENT_ADDRESS and CTAIP support?

Correct. Those could be also removed.

edwardneal: I've submitted that as #2831.

cheenamalhotra:
Please resolve conflicts to help us move forward with reviews.
Thanks!

edwardneal:
Thanks @cheenamalhotra - now resolved, and being CI'd.

mdaigle:
I'm working on dropping net6 support here #2927

benrr101 (approved):
In general I'm happy to take this change! However, we've kinda got a three-headed race between you, me, an @mdaigle that'll be challenging to sort out. So here's what I'll propose to make it go more smoothly:

* Let's take this PR
  * Let's leave the files in netcore alone for now, though
* I'll get my chain of PRs for moving interop files into the common project done
  * These will now delete a lot of it (net6 code), but also address most of the concerns you had for how to move them.
* @edwardneal can you send out a PR after that that removes InternalException and NetEventSource.Common?
* I'm going to continue the merging process with SNI and LocalDBAPI stuff. I selfishly want to do the safe/nativemethods migration, but if I won't stop you if you want to do it :)

edwardneal:
That sounds sensible to me, thanks. I've noticed that we're working in similar areas!

I've reverted the changes in netcore as suggested, but have left `SqlUtil.SocketDidNotThrow` continuing to throw an Exception rather than an InternalException. I can revert this too if it causes merge conflicts.

> I'm going to continue the merging process with SNI and LocalDBAPI stuff. I selfishly want to do the safe/nativemethods migration, but if I won't stop you if you want to do it :)

I've got no problem stepping away from the safe/nativemethods migration there. A complete merge of the SNI PInvokes will likely need changes to the native SNI package, so I think that work would have to come to you and the rest of the team whatever happens.

I'm interested to see how we merge LocalDBAPI. There were enough differences between netcore and netfx that I'd kicked it into the long grass while I tried to deal with the simple merges.

benrr101:
I think that all sounds good! I just don't want to steal your thunder 😛 We really appreciate your contributions and feedback!

The main reason I got into the interop/pinvoke stuff was because I started on LocalDBAPI, then found it had too many dependencies on SNINativeWrapper. So I started merging SNINativeWrapper, and found it had too many dependencies on Interop. So I started merging the Interop stuff... The LocalDBAPI and SNINativeWrapper stuff is going to be challenging to merge since the code isn't aligned the same in both versions (in at least one case, split between files in one version, and in another all in the same file).

Either way, let's get one more review on this one 👍

src/Microsoft.Data.SqlClient/src/Interop/Windows/Interop.Libraries.cs
```diff
@@ -6,11 +6,14 @@ internal static partial class Interop
 {
     internal static partial class Libraries
     {
-        internal const string Crypt32 = "crypt32.dll";
         internal const string Kernel32 = "kernel32.dll";
         internal const string NtDll = "ntdll.dll";
 #if !NET8_0_OR_GREATER
+        internal const string Crypt32 = "crypt32.dll";
         internal const string SspiCli = "sspicli.dll";
+#endif
```
mdaigle, line 14: Can we consolidate this into the `#if NETFRAMEWORK` below? I think it reads better given we've dropped .NET 6 and 7.

mdaigle (changes requested):

src/Microsoft.Data.SqlClient/src/Interop/Windows/Interop.Libraries.cs
benrr101: Don't worry too much, this file will be dropped shortly.

mdaigle (approved):

src/Microsoft.Data.SqlClient/src/Interop/Windows/Interop.Libraries.cs
mdaigle: Having trouble keeping track 😆

PR #2831, opened 2024-09-02
Removed CTAIP, certificate authentication

edwardneal:
Picking up from a comment on #2828.

This removes anything guarded behind ADONET_CERT_AUTH and ADONET_ORIGINAL_CLIENT_ADDRESS symbols. It then cleans up the knock-on effects of this: an unused Certificate option in the connection string, the PInvokes and their now-unreferenced structs, a delegate type.

The PR touches quite a few files, but it's necessary to start approaching any merge of SqlConnection.

/cc @JRahnama


saurabh500 (commented):
@edwardneal we will not be merging this PR anytime soon. There are certain decisions to be taken internally for CTAIP. 

We have some internal builds of the driver which rely on this capability and is important for us. 
The capabilities are not shipped with any public SKU of sql server and hence the capability doesn't ship in the public nuget driver.

saurabh500:
@edwardneal  scratch what I said. I just had a hallway conversation with @David-Engel 
We should be OK to take this forward.

edwardneal:
That's great, thanks @saurabh500. If that changes, I can split the PR so that we remove CTAIP support and the .NET Framework-specific SSL callbacks separately.

David-Engel (approved):

PR #2601, opened 2024-06-22
Prelogin handlers for connectivity

saurabh500:
Adding Prelogin handler

There is a Handler called PreloginHandler, which is composed of two subhandlers depending on the Encryption option of the Connection String.

Encrypt = Strict signifies TDS8 and anything else means TDS7

For TDS8, the chain in prelogin looks like Tds8TlsHandler -> PreloginPacketHandler
Else PreloginPacketHandler -> Tds74TlsHandler. 

The creation of SslStreams, and their ordering is done in the PreloginHandler itself. 

Tds8TlsHandler and Tds74TlsHandlers, take care of TLS before or after prelogin exchange depending on the protocol.

The code in this PR hasn't changed anything from the existing code, except removing checks for unsupported SQL Server versions. 

Tests have been added.

I also tested end to end (not checkedin) by creating the DataSourceParsingHandler>TransportCreationHandler > PreloginHandler in a chain.

To test TDS8, I used Azure SQL DB and for TDS74, I used a local server. 

1. DataSourceParsingHandler has been changed to look for ServerInfo instead of ConnectionOptions. 
2. TransportCreationHandler has been changed to add a GUID which is the connection Id.
3. For Prelogin handler, I chose to use a different context object, this is to add some derived items and to specifically reference some prelogin related fields.

## How to review

Start from the `PreloginHandler` class which is the handler entrypoint for Prelogin. All the Tds8 specific handling is in `Tds8TlsHandler` and Tds74 handling is in `Tds74TlsHandler`.
The transmission and reading Prelogin packet is available in `PreloginPacketHandler` 

## Tests

Unit tests have been added. 

For sanity integration test I used the following in a Unit test to validate connectivity. However this snippet is not checked in

```C#
DataSourceParsingHandler dspHandler = new DataSourceParsingHandler();
TransportCreationHandler tcHandler = new TransportCreationHandler();
PreloginHandler plHandler = new PreloginHandler();
ConnectionHandlerContext chc = new ConnectionHandlerContext();
SqlConnectionStringBuilder csb = new SqlConnectionStringBuilder();
csb.DataSource = "tcp:localhost,1444";
csb.Encrypt = SqlConnectionEncryptOption.Mandatory;

csb.TrustServerCertificate = true;

SqlConnectionString scs = new SqlConnectionString(csb.ConnectionString);
chc.ConnectionString = scs;
var serverInfo = new ServerInfo(scs);
serverInfo.SetDerivedNames(null, serverInfo.UserServerName);
[...]

PR #2839, opened 2024-09-08
Prevent connections to SQL 7.0 & 2000

edwardneal:
This prevents connections from the .NET Framework build of Microsoft.Data.SqlClient to SQL Server 7.0 and SQL Server 2000 (RTM & SP1.) Doing so aligns the .NET Framework build with the .NET Core build on all OSes: the library will no longer be able to connect to any version of SQL Server prior to 2005.

I've also included a test to verify failed/successful connections to different simulated server versions.

There's a bit of SQL Server 2000-specific code in the .NET Framework project, and this hasn't been removed yet. Since this could be considered a breaking change, I'm aiming simply to implement the functional block for the v6 release, ready for the then-redundant paths to be removed later.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
+++ b/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
@@ -4315,25 +4315,6 @@ namespace Microsoft.Data.SqlClient
             // information provided by S. Ashwin
             switch (majorMinor)
             {
-                case TdsEnums.SQL70OR2000_MAJOR << 24 | TdsEnums.DEFAULT_MINOR:    // 7.0 & 2000 RTM
-                    // note that 7.0 and 2000 can only be distinguished by the increment
-                    switch (increment)
-                    {
-                        case TdsEnums.SQL2000_INCREMENT:
-                            _is2000 = true;
-                            break;
-                        case TdsEnums.SQL70_INCREMENT:
-                            // no flag will be set
-                            break;
-                        default:
-                            throw SQL.InvalidTDSVersion();
-                    }
-                    break;
-                case TdsEnums.SQL2000SP1_MAJOR << 24 | TdsEnums.SQL2000SP1_MINOR: // 2000 SP1
-                    if (increment != TdsEnums.SQL2000SP1_INCREMENT)
-                    { throw SQL.InvalidTDSVersion(); }
-                    _is2000SP1 = true;
-                    break;
                 case TdsEnums.SQL2005_MAJOR << 24 | TdsEnums.SQL2005_RTM_MINOR:     // 2005
                     if (increment != TdsEnums.SQL2005_INCREMENT)
                     { throw SQL.InvalidTDSVersion(); }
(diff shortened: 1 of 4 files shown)


Wraith2:
While those systems aren't supported is there a reason to remove the ability to use it? I've encountered a live sql 2000 server recently and while i wasn't programming against it others may be, for example sql management studio will lose the ability to connect.

edwardneal:
That's a completely fair point, and part of the reason I raised the PR is to talk that through. As we start to merge the more sensitive classes relating to connections and transactions, there are code paths which we handle differently for downlevel clients. The options for this are:
1. Implement the same code paths in the .NET Core library, and test them with a SQL 2000 system. I'm not opposed to doing this if it's technically practical for the CI to run those tests, but at that point we've effectively relaxed the minimum version from SQL 2014 to SQL 2000
2. Lock them behind conditional compilation - which isn't the worst thing in the world, but I'd prefer to only do this for cases where .NET Core/Framework have different APIs for the same operation, rather than to maintain a codebase which has different capabilities based on how it's targeted
3. Remove SQL 2000 & SQL 7.0 support from the .NET Framework library

The [initial feedback](https://github.com/dotnet/SqlClient/pull/2601#discussion_r1651696242) was:
> There's no reason that main needs to work on anything older than 2014. Scratch that. 2016, since 2014 EOLs July 9. Users who are running EOL servers can use old clients.

I've seen enough SQL 2008/R2/2012 servers recently to err on the side of caution when thinking about enforcing a minimum version of SQL 2014 across the board, but a baseline of SQL 2005 seems fair enough. I'm only proposing the removal of SQL 2000 support from the .NET Framework version, which still has the supported alternative of System.Data.SqlClient; the .NET Core library has never supported it.

As far as SSMS support goes: I expect that people who still need to connect to downlevel versions of SQL Server would need an older version of SSMS.

To put some context to the downstream changes this'll enable, I've got a second PR after this which covers the actual code changes. The diff of this is [here](https://github.com/edwardneal/SqlClient/compare/sql2000-support-removal...edwardneal:SqlClient:sql2000-complete-removal).

David-Engel:
We aren't going to test EOL SQL Server versions in CI. Security compliance rules that out.

I'm all for cleaning up old code. We should not connect to anything prior to TDS 7.4.

CC: @saurabh500 and @cheenamalhotra as an FYI.

MichelZ:
Should these be removed then and the usages cleaned up as well?

https://github.com/dotnet/SqlClient/blob/aef195460d0a55e23efa91a1d8d7187d33e632e7/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs#L212-L214

MichelZ:
Removing SQL 7.0/2000 support from TdsParser would remove quite a few differences between netcore and netfx and make merging the classes much easier. I hope this gets in soon :)

benrr101 (approved):
I'm sort of on the fence about this one. I can totally sympathize with @Wraith2 desire to leave support for old versions intact (eg, I work on a side project that targets a totally out-of-date version of nodejs so I can have the widest support I can). Yet, I also acknowledge that with corporate-sponsored open-source projects, we have to follow the compliance rules, and that generally means dropping support for EOL products. If there's no easy/approved way to test against EOL products, the code is going to wither, we'll get complaints about it not working, and won't be able to address the issues. So ultimately, I think we'll have to drop support for EOL SQL Server versions. As long as we don't pull old versions of MDS from Nuget, there'll still be an option for users of old versions of SQL Server.

mdaigle (approved):

SimonCropp:
just a heads up to people that this is now in the 6.1 milestone. 
so i raised a discussion to talk about if this should be considered "breaking enough" to not be in a minor release

https://github.com/dotnet/SqlClient/discussions/3276

PR #2841, opened 2024-09-09
Fix | Improve async pattern in TCPHandle for ParallelConnect for MultiSubnetFailover scenarios

JRahnama:
This PR tries to address below:

1.  the fire-and-forget scenario in ParallelConnectHelper, as `async Task void` should be avoided in non-event-handler methods. We never see any exception that is thrown from here.
2. Additionally, it removes `if (!(connectTask.Wait(isInfiniteTimeOut ? -1 : timeout.MillisecondsRemainingInt)))`, which blocks the asynchronous pattern and may lead to deadlocks.

This might be a potential fix for issue #2400 

Any suggestion to make this better is appreciated.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
@@ -17,6 +17,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Data.Common;
 using Microsoft.Data.ProviderBase;
+using Microsoft.Identity.Client;
 
 namespace Microsoft.Data.SqlClient.SNI
 {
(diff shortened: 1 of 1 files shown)


David-Engel:
I don't think you'll fix the issue without eliminating async from the logic. Had a chat with Saurabh and he pointed out that we should be able to leverage the same Blocking=false logic from the Connect method in the parallel method. On the same thread, create and connect all the sockets. Then Socket.Select, given lists of all the sockets, should return when the first one is available.

> [Select](https://learn.microsoft.com/en-us/dotnet/api/system.net.sockets.socket.select?view=net-8.0) returns when at least one of the sockets of interest (the sockets in the checkRead, checkWrite, and checkError lists) meets its specified criteria, or the microSeconds parameter is exceeded, whichever comes first.

Just find the first one that has connected.

JRahnama:
> I don't think you'll fix the issue without eliminating async from the logic. Had a chat with Saurabh and he pointed out that we should be able to leverage the same Blocking=false logic from the Connect method in the parallel method. On the same thread, create and connect all the sockets. Then Socket.Select, given lists of all the sockets, should return when the first one is available.
> 
> > [Select](https://learn.microsoft.com/en-us/dotnet/api/system.net.sockets.socket.select?view=net-8.0) returns when at least one of the sockets of interest (the sockets in the checkRead, checkWrite, and checkError lists) meets its specified criteria, or the microSeconds parameter is exceeded, whichever comes first.
> 
> Just find the first one that has connected.

Thank you, @David-Engel. I have removed the async logic from TCPHandle. Could you please review the changes again?

The use of Parallel.ForEach is intended to test different failover scenarios more quickly. The advantage is that the Task Scheduler will determine the optimal number of threads based on the OS's capabilities, which may even result in using just a single thread if that’s deemed most efficient.

PR #2842, opened 2024-09-09
[5.2.3] | Advance column index to avoid double clean

mdaigle:


cheenamalhotra:
I don't think we should need to patch all previous stable releases for this bug, due to the nature of bug and it's impact and to prevent risk of regression. Unless someone asks for it specifically, users should be able to upgrade to newer version of SqlClient to get the fix.

cc @David-Engel

PR #2844, opened 2024-09-09
Test | Unit Test for decrypt failure to drain data fix

arellegue:
This PR is for unit test of PR #2618.

### DIFF
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/AKVTests.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/AKVTests.cs
@@ -103,11 +103,12 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests.AlwaysEncrypted
                     customerFirstParam.ForceColumnEncryption = true;
 
                     using SqlDataReader sqlDataReader = sqlCommand.ExecuteReader();
-                    var error = Assert.Throws<SqlException>(() => sqlDataReader.Read());
-                    if (error != null)
+                    while (sqlDataReader.Read())
+                    {
+                        var error = Assert.Throws<SqlException>(() => DatabaseHelper.CompareResults(sqlDataReader, new string[] { @"string" }, 1));
+                        Console.WriteLine($"Error = {error.Message}");
                         Assert.Contains("Failed to decrypt column", error.Message);
-                    else
-                        Assert.Fail("Decryption unexpectedly succeeded.");
+                    }
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/AKVTests.cs
```diff
@@ -58,11 +58,74 @@ public void TestEncryptDecryptWithAKV()
             DatabaseHelper.ValidateResultSet(sqlDataReader);
         }
 
+        /*
+         This unit test is designed for PR #2618, which addresses an issue where a failed decryption leaves a connection in a bad state
```
DavoudEshtehari, line 62:
```suggestion
         This unit test is going to assess an issue where a failed decryption leaves a connection in a bad state  
```

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/AKVTests.cs
```diff
@@ -58,11 +58,74 @@ public void TestEncryptDecryptWithAKV()
...
+
+            // Setup Empty key store provider
+            Dictionary<String, SqlColumnEncryptionKeyStoreProvider> emptyKeyStoreProviders = new()
+            {
+                { "AZURE_KEY_VAULT", new EmptyKeyStoreProvider() }
+            };
+
+            // Three consecutive connections should fail with "Failed to decrypt column" error. This proves that an error in decryption
+            // does not leave the connection in a bad state.
+            for (int i = 0; i < 3; i++)
```
DavoudEshtehari, line 101: Could you explain what happens internally during each retry and why it specifically requires 3 attempts, and include the explanation in the code?

arellegue: Before the fix was implemented, this is what happens internally:

1.  A connection is used from a connection in pool.
2. That connection is used to query a table with a column that is encrypted.
3. When the decryption failed for some reason, a "Failed to decrypt column." error is raised. The connection is returned to the connection pool with the error still in its buffer.
4. When the client connects again, a connection is used from the pool again and grabbing the same connection recently returned back to the pool.
5. This time when the connection is opened it will raise an "Internal connection fatal error." because that connection still has an existing error in it.  Thus, the driver dooms that connection and not returned it back to the connection pool.
6. When the client connects the third time, it will grab an unused connection from the pool. 
7. That new connection is used to query a table with a column that is encrypted.
8. When the decryption failed for some reason, a "Failed to decrypt column." error is raised. The connection is returned to the connection pool with the error still in its buffer.
9. And the cycle continues.

So, this unit test will try to connect 3 times and do the same query to ensure that the expected error messages are identical each time, "Failed to decrypt column". This proves that the fix worked, and the connection is returned to the pool in a good state.

DavoudEshtehari (approved):

cheenamalhotra (approved):

PR #2845, opened 2024-09-09
[5.2] | Fix strong name in AKV provider

dauinsight:


src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider.csproj
```diff
@@ -15,6 +15,14 @@
     <BuildProjectReferences>false</BuildProjectReferences>
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
     <PackageRequireLicenseAcceptance>true</PackageRequireLicenseAcceptance>
+  </PropertyGroup>
+   <!--Generating Strong Name-->
+  <PropertyGroup Condition="$(CDP_BUILD_TYPE)==Official">
```
campersau, line 20: Normally string comparisons in msbuild should use quotes, or is this something else?

```suggestion
  <PropertyGroup Condition="'$(CDP_BUILD_TYPE)'=='Official'">
```

DavoudEshtehari:
> Why is this issue patched in all branches? Are you building AKV provider from all servicing branches?

Currently, we have two different versions of AKV, requiring at least two branches for each version. Since its source code resides within MDS, we manage it under the MDS servicing branches. When releasing a potential AKV hotfix, we can decide which version to use, but it can't be solely from the main branch. Additionally, OneBranch pipeline runs won’t function properly without this change.

cheenamalhotra:
Thanks for the clarification @DavoudEshtehari

arellegue (approved):

Issue #2838, opened 2024-09-08
.NET Framework SQLCLR support


edwardneal:
This partially comes from the netfx/netcore project merge, but has opened into a more general question: is SqlClient's .NET Framebuild build supported for use in SQLCLR scenarios? My experience so far has been that it isn't working, but some review of the code suggests to me that it can't work.

### Background context
I'm trying to test the context connection. When Context Connection is true, SqlClient starts to use a lot of the `*Smi` and `Smi*` classes which are in the .NET Framework project, plus various branches within other classes. My library is the foundation of a testbed to prove that the functionality works: first with .NET Framework's intrinsic System.Data.SqlClient, second with Microsoft.Data.SqlClient as-is, and third after any code merges.

### System.Data.SqlClient
The sample library which I've used as a testbed is pretty simple. It contains one class, as below:
```csharp
public class NetFxStoredProcedures
{
    [Microsoft.SqlServer.Server.SqlProcedure]
    public static void uspEcho(System.Data.SqlTypes.SqlString spToEcho)
    {
        var record = new Microsoft.SqlServer.Server.SqlDataRecord(
            new Microsoft.SqlServer.Server.SqlMetaData("Echo", System.Data.SqlDbType.NVarChar, 200)
        );

        record.SetSqlString(0, spToEcho);

        Microsoft.SqlServer.Server.SqlContext.Pipe.Send(record);
    }

    [Microsoft.SqlServer.Server.SqlProcedure]
    public static void uspPrintDate()
    {
        using (var cConn = new System.Data.SqlClient.SqlConnection("Context Connection=true"))
        {
            cConn.Open();

            var cmd = new System.Data.SqlClient.SqlCommand("select getdate() as MyDate", cConn);

            DateTime dt = (DateTime)cmd.ExecuteScalar();

            Microsoft.SqlServer.Server.SqlContext.Pipe.Send("The current date is: " + dt.ToString("O"));
        }
    }
}
```

I then load it into SQL Server with the script:
```sql
use [master]

create database SqlClrTestDb

alter database SqlClrTestDb
    set trustworthy on

go

use [SqlClrTestDb]

create assembly NetFx from '<path>\SqlClrTestbed.dll'
    with permission_set = unsafe
go

create procedure usp_NetFxEcho
    @spToEcho nvarchar(200)
as external name NetFx.[SqlClrTestbed.NetFxStoredProcedures].uspEcho
go

create procedure usp_NetFxPrint
[...]

PR #2862, opened 2024-09-15
Merge ContextConnection handling

edwardneal:
Relates to #2838, #1261.

This is partially to fix the documentation and the obsolescence message on `SqlConnectionStringBuilder.ContextConnection`. It goes a little further though: setting ContextConnection to true will now always throw an ArgumentException, and manually specifying it in the connection string will now throw an explicit error highlighting that Microsoft.Data.SqlClient doesn't support it.

Doing this also meant merging the handling of it between .NET Core and Framework. Once that was done, I've reintroduced a number of places where references to SqlConnectionString.ContextConnection had been removed from the .NET Core code. This is a prerequisite to merge SqlDataReaderSmi et al.

One point of note here is that SqlConnectionString.ContextConnection (line 763 of SqlConnectionString.cs) is hardcoded to return false. This is so that the .NET JIT can optimise away any evaluation - example [here](https://sharplab.io/#v2:C4LghgzgtgPgAgJgIwFgBQcDMACR2DKAjgDYDCA9gHaUCmAxsAJZX7ABOjlA5tgN7rZBuHACNy5YtgqVgNAB7BptBs0rYAvAD5sAMzDEINANwChpwVlwAWbAEEGACgCUfc0MGMd2B9NkKl9ExUTm7u/GjukUJwSACcDgBEvvKKVMpBaowQ2OwArjQJTiYRUdgAvqFCNAY0lYLhpZEx8UlUfqnUgarYWbr6hoXFjRUlgiNlQA).

This changes the public API surface slightly: it introduces the SqlConnectionStringBuilder.ContextConnection property to .NET Core, and changes the message on its `Obsolete` attribute on both platforms.


cheenamalhotra:
@edwardneal Please resolve conflicts so we can continue with reviews.
Thanks!

edwardneal:
Thanks @cheenamalhotra - now resolved, pending CI.

cheenamalhotra:
I would vote for removing the property from .NET Framework too instead of adding and making it obsolete in .NET Core.. This property was first marked obsolete in v3.0.Does that sound acceptable @David-Engel ?

edwardneal:
Thanks @cheenamalhotra. Part of the reason I picked this approach is to make a specific statement that SqlClient doesn't support connecting to SQL Server's context connection, so if the documentation is updated to make that lack of support clear then I don't have any objections to removing the property from .NET Framework (besides the API compatibility concerns - but a v6.0 release is probably our last opportunity to make this kind of breaking change for a while.)

The `SqlConnectionStringBuilder.ContextConnection` property also backs a `Context Connection` key in the connection string. If we decide to remove the property, I'd prefer to keep treating this as a valid key and to throw a specific exception if its value parses to `true`, since it's a behavioural change from S.D.S.

David-Engel:
> I would vote for removing the property from .NET Framework too instead of adding and making it obsolete in .NET Core.. This property was first marked obsolete in v3.0.
> 
> Does that sound acceptable @David-Engel ?

@saurabh500 - Do you think MDS will ever get into the CLR? If not, I agree we should just remove the ContextConnection property.

saurabh500:
@David-Engel @cheenamalhotra dont know about forever. 😀
But this hasn't been on the radar or been tested in the past 5 years. 
I will vote for removal from netfx.

cheenamalhotra:
@edwardneal Could you please update the PR as per comments from David and Saurabh?

edwardneal:
Thanks all - I've just rolled back the related additions to .NET Core's codebase and removed the ContextConnection property from both codebases. There's one piece of new functionality: specifying `Context Connection=true` in the connection string will now throw a new exception which explicitly states that this behaviour isn't supported.

SQLCLR casts quite a long shadow here. I've handled the API surface in this PR, and will remove the various Smi classes and code paths once 6.0 is released. Once those are gone, I'll close my original issue 2838.

benrr101 (approved):
I'm ok with these changes, but I guess I want to know why we don't go further with them - can we strip out all of the contextConnection stuff from the codebase? I guess I don't understand why we just stub it all out to returning false or throwing when we could just remove it from the API altogether.

edwardneal:
It's in the works, but there's a lot to review. An [early draft](https://github.com/edwardneal/SqlClient/compare/main...edwardneal:SqlClient:remove-contextconnection-paths?expand=1) comes to +1152, -8142 lines and I don't want to drop it in a single PR a month before v6.0. I'm thinking of introducing a set of four PRs post-release:
1. Remove SqlCommand/SqlDataReader Smi code paths and files
2. Remove SqlConnection Smi code paths and files
3. With all referencing code now removed, the SmiEventSink references are potentially no longer relevant. Remove these code paths and files if not
4. Remove any references within SqlUtil and the resource strings

cheenamalhotra:
@edwardneal 
I will proceed to merge this as we're in a major version bump for 6.0, so this change is acceptable. Remaining cleanup I hope doesn't impact public surface area any further, so we can continue to track with #2838.

edwardneal:
Thanks @cheenamalhotra. I've kept the 6.0 milestone in mind, and have front-loaded all of the surface area changes that my cleanup needs. The rest of the cleanup can wait until later.

I'm conscious that it's pretty unusual to completely delete part of the API surface, and if that's too disruptive for the release, I can easily submit changes which add it back.

Issue #2013, opened 2023-04-20
CDATA Block is not formatter correctly in VS


YuryKomar:
The function descriprion includes a comment CDATA block is not completely formatted in Visual Studio. I mean the XREF hyperlink, it stil visible in popup...

Will it be fixed? 

see attached image
![Снимок](https://user-images.githubusercontent.com/12536828/233348792-8b2ca5ff-f54f-48f7-9644-83673945d894.PNG)
![SqlClient Info](https://user-images.githubusercontent.com/12536828/233348806-547bc107-50ce-46ac-b066-fefa9a8c988c.png)

JRahnama:
We will check this with VS team and will get back to you.

arkalyanms:
Moved to the right repo via https://github.com/dotnet/roslyn/issues/68134. Feel free to close this one.

JRahnama:
Thank you @arkalyanms. @YuryKomar you can follow this on [roslyn repo](https://github.com/dotnet/roslyn/issues/68134). Closing this as an external issue.

sharwell:
The rendering of this comment is an accurate reflection of the content provided here:

https://github.com/dotnet/SqlClient/blob/f478be5bf3597f5c0f0940ac471c6b481c38416f/doc/snippets/Microsoft.Data.SqlClient/SqlCommand.xml#L7-L47

Note that `<format type="text/markdown">` is not a recognized content for documentation comments, so the contents of this element will be rendered the same way as if the content was written inline (i.e. not surrounded by the `<format>` element).

JRahnama:
This is to inform you that work is currently in progress to resolve the issue. The first [PR](https://github.com/dotnet/SqlClient/pull/2822) has been merged, and the remaining updates will follow shortly.

benrr101:
At long last, the batches have been completed, and the docs have been restructured into XML format. There are a couple follow up tasks, but the bulk of the work is complete.

benrr101:
(also I guess I messed up by linking the issue to the PR and it auto-closed it...)

JRahnama:
> (also I guess I messed up by linking the issue to the PR and it auto-closed it...)

I think this is fine/correct., as the issue is addressed/being addressed in several PRs.

PR #2864, opened 2024-09-16
XML Documentation Fixes - Batch 5

benrr101:
**Description**: Batch 5 for the documentation restructuring changes to chip away at #2013. These changes shouldn't interfere with each other, so I think it's ok to just blast all the PRs at once. 

**Testing**: I checked that the ref projects still build and skimmed the docs as they appear in the IDE.


src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -1453,7 +1459,7 @@ internal SqlException() { }
         /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlException.xml' path='docs/members[@name="SqlException"]/State/*'/>
         public byte State { get { throw null; } }
 
-        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlException.xml' path='docs/members[@name="SqlException"]/GetObjectData/*'/>
+        /// <inheritdoc cref="System.Exception.GetObjectData" />
```
David-Engel, line 1462: I've heard inhericdoc doesn't work across doc repositories. Does this appear correctly in the IDE?

doc/snippets/Microsoft.Data.SqlClient/SqlConnectionStringBuilder.xml
```diff
@@ -1,1129 +1,1921 @@
...
+              }
+          }
+        </code>
+        <para>
+          The example displays the following output in the console window:
+        </para>
+        <code>
+          Connection string = Data Source=(local);Initial Catalog=AdventureWorks;Integrated Security=True
+          True
+          True
```
mdaigle, line 583: looks like something went wrong here

doc/snippets/Microsoft.Data.SqlClient/SqlException.xml
```diff
@@ -1,390 +1,889 @@
...
+        </para>
+        <para>
+          Bad input can cause a .NET Framework data provider type to raise an exception such as System.ArgumentException or System.IndexOutOfRangeException. Calling a method at the wrong time can raise System.InvalidOperationException.
+        </para>
+        <para>
+          So, in general, write an exception handler that catches any provider specific exceptions as well as exceptions from the common language runtime. These can be layered as follows:
+        </para>
+        <code language="c#">
+          try
+          {
```
mdaigle, line 37: remove these?

doc/snippets/Microsoft.Data.SqlClient/SqlDataAdapter.xml
```diff
@@ -1,501 +1,1374 @@
...
+          
+                  return adapter;
+              }
+          }
+        </code>
+      </example>
+    </ctorSelectCommandTextSelectConnectionString>
+    <AddToBatch>
+      <param name="command">
+        To be added.
```
mdaigle, line 527: do we plan to keep these? I know others were removed

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
benrr101: Hmm. It shows up correctly in the IDE, but if @JRahnama is correct about how it gets generated into doc websites, then we should probably not use this.

benrr101: Also for the record, the xml link was to an inheritdoc tag, which for whatever reason doesn't show up correctly in the IDE. I can replace it with a "to be added" summary if that's ok?

doc/snippets/Microsoft.Data.SqlClient/SqlConnectionStringBuilder.xml
benrr101: Nope, that's about right. That is supposed to show the output the example code and it outputs the connection string then outputs whether it contains a handful of keys. So the expected output is a bunch of trues. But, it's always good to double check stuff like that :+1:

doc/snippets/Microsoft.Data.SqlClient/SqlException.xml
benrr101: Honestly these examples seem totally pointless to me because it's basically "here's how to use a try/catch and catch SqlException separately". But I'm not really trying to change the contents here.

Was your comment just about the extra space?

doc/snippets/Microsoft.Data.SqlClient/SqlDataAdapter.xml
benrr101: So, I removed the remarks with "To be added" under the rationale that if we haven't added them now, we probably don't need them. But in those scenarios we already had other comment blocks, so there would still be comments in the output.

On the other hand, if I remove these "to be added" comments, the corresponding declaration no longer has any comments. That might cause the build to fail, but if it doesn't we don't have any easy way of searching for missing comments. So I'm leaving them here for now so we can easily go back through and add them at a later date (not as part of the work for #2013!)

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
David-Engel: It looks like inheritdoc support may have actually been added to the API doc generation tool in the last few years. I vote leave it and we'll see how it turns out.

JRahnama: We can add them and request a build from the documentation team on a test branch to review the results and determine whether they should remain.

I remember that for GetObjectData, everything worked fine with no issues. However, in the SqlBatch area and related classes, we encountered many cases where we ended up duplicating code.

doc/snippets/Microsoft.Data.SqlClient/SqlException.xml
mdaigle: It was about the contents of the example. I don't think they're adding any value. I'm fine with leaving it if it's out of scope for your effort.

mdaigle (approved):

Issue #2872, opened 2024-09-19
Different behavior for Routing in netcore and netfx


JRahnama:
There are two different behavior for Routing in SqlClient

in .Net inside [SqlInternalConnectionTds](https://github.com/dotnet/SqlClient/blob/3416ad92b9107d2618bc7ffdf1d595a86cf0e7d0/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs#L1561) we have 
```C#
if (routingAttempts > 0)
                        {
                            throw SQL.ROR_RecursiveRoutingNotSupported(this);
                        }
```
Where as the same place in [.NetFramework](https://github.com/dotnet/SqlClient/blob/3416ad92b9107d2618bc7ffdf1d595a86cf0e7d0/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs#L1886) we have

```C#
if (routingAttempts > _maxNumberOfRedirectRoute)
                        {
                            throw SQL.ROR_RecursiveRoutingNotSupported(this);
                        }
```

`_maxNumberOfRedirectRout` is 10

David-Engel:
@JRahnama 

I noticed this a little while ago and looked into the history a bit. The max routing attempts changes was included as part of the Azure read only routing support changes in SDS netfx. I saved a branch of what I thought were the relevant changes required from bring everything from the netfx to the netcore MDS code: https://github.com/dotnet/SqlClient/compare/main...David-Engel:SqlClient:port?expand=1

JRahnama:
> @JRahnama
> 
> I noticed this a little while ago and looked into the history a bit. The max routing attempts changes was included as part of the Azure read only routing support changes in SDS netfx. I saved a branch of what I thought were the relevant changes required from bring everything from the netfx to the netcore MDS code: https://github.com/dotnet/SqlClient/compare/main...David-Engel:SqlClient:port?expand=1

Thanks David, I have updated the PR #2873  with some extra changes and addressed some issues in your changes.

PR #2873, opened 2024-09-19
Fix | Increase routing attempt to 10 in netcore for LoginNoFailover and adding routing support to LoginWithFailover

JRahnama:
fixes issue #2872

### DIFF
--- a/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
+++ b/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
@@ -139,7 +139,6 @@ namespace Microsoft.Data.SqlClient
         internal readonly Func<SqlAuthenticationParameters, CancellationToken,Task<SqlAuthenticationToken>> _accessTokenCallback;
 
         private readonly ActiveDirectoryAuthenticationTimeoutRetryHelper _activeDirectoryAuthTimeoutRetryHelper;
-        private readonly SqlAuthenticationProviderManager _sqlAuthenticationProviderManager;
 
         // Certificate auth calbacks.
         ServerCertificateValidationCallback _serverCallback;
@@ -288,7 +287,7 @@ namespace Microsoft.Data.SqlClient
         private int _asyncCommandCount; // number of async Begins minus number of async Ends.
 
         // FOR SSE
-        private string _instanceName = String.Empty;
+        private string _instanceName = string.Empty;
 
         // FOR NOTIFICATIONS
         private DbConnectionPoolIdentity _identity; // Used to lookup info for notification matching Start().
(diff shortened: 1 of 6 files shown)


JRahnama:
It seems like the Azure triggers are not working to validate this PR.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -2795,7 +2802,7 @@ internal SqlFedAuthToken GetFedAuthToken(SqlFedAuthInfo fedAuthInfo)
             // Username to use in error messages.
             string username = null;
 
-            var authProvider = _sqlAuthenticationProviderManager.GetProvider(ConnectionOptions.Authentication);
+            SqlAuthenticationProvider authProvider = SqlAuthenticationProvider.GetProvider(ConnectionOptions.Authentication);
```
David-Engel, line 2805: With this change, you can now remove the local variable _sqlAuthenticationProviderManager (lines 142 and 499). This change in the netcore code was to eliminate loading of the manager if AAD auth isn't even used and this syncs it to netfx. Even though technically unrelated to the PR description, my thought was to try to sync netfx and netcore and bit more with future code merging in mind.

JRahnama: I noticed that. I also made a few other changes to align .NET Core and .NET Framework more closely. Should I create a separate PR for those changes, or are you okay with including them in this one?

David-Engel: I'd just include it here.

edwardneal (commented):
Looks good! Between this and the other PRs, I think the only remaining behaviour difference between .NET Core and Framework for this class is TransparentNetworkIPResolution, which is pretty minor.

I'm not sure when exactly this behaviour is triggered - I imagine it'll be an Azure SQL database with geo-replication and a readable secondary replica configured, but haven't tested.

There might be some extra handling in OnEnvChange to adjust. On [lines 2081-2084](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs#L2081-L2084) in the .NET Core version, we throw an exception if we receive the ENV_LOGSHIPNODE EnvChange. This exception isn't thrown in the .NET Framework.

Looking more generally: while there's a RoutingTDSServer object in the TDS.Servers project, I can't see any test coverage for routing. I think the three critical scenarios to test would be zero routing (with ApplicationIntent of ReadOnly), one re-route to another TDSServer, and a routing loop (to verify that the routingAttemps comparison takes effect and doesn't hang the driver) I'm not certain about a fourth scenario, where a TDS server tries to route to itself - I've not tested whether a replica's read-only routing URLs can actually be misconfigured like that.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
JRahnama: done

David-Engel (approved):

cheenamalhotra (approved):

PR #2877, opened 2024-09-20
Fix | Check Connected property of sockets during parallel connect

David-Engel:
Address the failure in the DNN scenario described in #2400.

In the MSF parallel connect path, socket connections can fail quickly (as opposed to timing out), resulting in the first socket connect task being the "success" task that closes all other sockets. This happens because in a VNN, there is no listener at the other end of inactive IP addresses and socket connects take a while before they time out. In a DNN, the gateway responds with an RST on the socket for inactive IP addresses and the RST may come in before the connection completes on the active IP address. This means a "success" socket connect task (success just meaning ConnectAsync finished) will disconnect all other attempted sockets.

The change here is to define success as the ConnectAsync task finished **_and_** the socket.Connected property is true.

If this addresses the DNN issue, this simple fix would be a good candidate for backport. There are other code changes that could be made to eliminate tasks on this sync path. I have some drafted up, but the changes would require much more testing.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
@@ -564,7 +564,7 @@ namespace Microsoft.Data.SqlClient.SNI
                 // Try to connect.  If we're successful, store this task into the result task.
                 await connectTask.ConfigureAwait(false);
                 // A connection attempt could return quickly if a server refuses connection or there is some other error
-                success = tcs.TrySetResult(socket) && socket.Connected;
+                success = socket.Connected && tcs.TrySetResult(socket);
                 if (success)
                 {
                     // Whichever connection completes the return task is responsible for disposing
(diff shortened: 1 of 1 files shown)


JRahnama (approved):

JRahnama:
I can ask the client to test with this set of changes. Hope they do not need it to be signed.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
```diff
@@ -563,7 +563,8 @@ private static async void ParallelConnectHelper(
             {
                 // Try to connect.  If we're successful, store this task into the result task.
                 await connectTask.ConfigureAwait(false);
-                success = tcs.TrySetResult(socket);
+                // A connection attempt could return quickly if a server refuses connection or there is some other error
+                success = tcs.TrySetResult(socket) && socket.Connected;
```
edwardneal, line 567:
```suggestion
                success = socket.Connected && tcs.TrySetResult(socket);
```

We'll want to make sure that the socket is connected before setting the TCS result. We can only set the TCS result once, so if we set it before checking if the socket is connected then we might set it to a disconnected socket. In that situation, `success` would be false for every socket, even the ones which successfully connected - so connection establishment would fail.

cheenamalhotra:
@JRahnama can you confirm if this has been tested by client? Thanks.

David-Engel:
I don't think this fix will actually work. Testing #2915 with a mock server that sends a RST after connect, MDS moves on before the RST even registers on the socket. The RST might never even register on the socket without using a socket.Poll (or other read/write). Even adding socket.Poll in the parallel logic only gets me ~30ms grace between connect and RST before MDS will pass the connected socket on up the stack. I'm very hesitant to add a wait here, watching for the endpoint to send a delayed RST.

Issue #2541, opened 2024-05-29
SqlClient's reference assemblies should declare all dependencies


benjamin-hodgson:
### Describe the bug
`SqlClient` was compiled with `Azure.Core` version 1.**35**.0.0. When building a project using version 1.**38**.0.0 (or indeed any other version), the appropriate binding redirect is not generated. This results in `FileLoadException`s in the AD authentication codepath - example:

```
Unhandled Exception: System.IO.FileLoadException: Could not load file or assembly 'Azure.Core, Version=1.35.0.0, Culture=neutral, PublicKeyToken=92742159e12e44c8' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)
   at Microsoft.Data.SqlClient.ActiveDirectoryAuthenticationProvider.AcquireTokenAsync(SqlAuthenticationParameters parameters)
   at Program.Main() in D:\test\Program.cs:line 7
```

(My minimal repro, below, trips on `Azure.Core` but I've also seen errors mentioning `ValueTuple` and `System.Threading.Tasks.Extensions` in other contexts. I believe they're all due to the same issue.)

The problem here is that the `ResolveAssemblyReferences` target consumes **reference** assemblies from Nuget (when they exist), and `SqlClient`'s reference assemblies do not declare the library's runtime dependency on `Azure.Core`.

The reference assembly in ILSpy (`ref\net462\Microsoft.Data.SqlClient.dll`):
![ref\net462\Microsoft.Data.SqlClient.dll](https://github.com/dotnet/SqlClient/assets/3857587/dd6474cc-9f5a-4bba-b43e-aba70894770f)

The code assembly (`lib\net462\Microsoft.Data.SqlClient.dll`):
![lib\net462\Microsoft.Data.SqlClient.dll](https://github.com/dotnet/SqlClient/assets/3857587/e4d29de0-59ca-430c-85cc-7bc3a8039a83)

This means `ResolveAssemblyReferences` doesn't realise there's a conflict (it thinks `SqlClient` doesn't require `Azure.Core` at all) so fails to generate the correct binding redirects. When `SqlClient`'s **code** assembly is deployed into the bin folder and run, it requires version 1.35, and, absent the required binding redirect, it dies because it can only find 1.38.

### To reproduce

[...]

Issue #2082, opened 2023-07-05
Could not load file or assembly 'System.Memory, Version=4.0.1.1, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51' or one of its dependencies.


chandrashekharjoshi-abb:
### Describe the bug
When i add nuget package Microsoft.SqlServer.SqlManagementObjects 170.18.0 for SMO APIs(it also adds lot of dependecy packages), One of the dependency is Microsoft.Data.SqlClient, which is using old version of System.Memory 4.0.1.1 while package downloads 4.0.1.2.
![image](https://github.com/dotnet/SqlClient/assets/137038435/2310ae95-b159-4618-b7b3-c5765369c1eb)

we always get exception on SqlConnection.Open() function call. below is the fusion log:
> Pre-bind state information 
> LOG: DisplayName = System.Memory, Version=4.0.1.1, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51
>  (Fully-specified)
> LOG: Appbase = file:///C:/ABB.SDS.SDK/host/Server/
> LOG: Initial PrivatePath = NULL
> Calling assembly : Microsoft.Data.SqlClient, Version=5.0.0.0, Culture=neutral, PublicKeyToken=23ec7fc2d6eaa4a5.
> 
> LOG: This bind starts in LoadFrom load context.
> WRN: Native image will not be probed in LoadFrom context. Native image will only be probed in default load context, like with Assembly.Load().
> LOG: Using application configuration file: C:\ABB.SDS.SDK\host\Server\MOMServices.exe.Config
> LOG: Using host configuration file: 
> LOG: Using machine configuration file from C:\Windows\Microsoft.NET\Framework64\v4.0.30319\config\machine.config.
> LOG: Post-policy reference: System.Memory, Version=4.0.1.1, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51
> LOG: Attempting download of new URL file:///C:/ABB.SDS.SDK/host/Server/System.Memory.DLL.
> LOG: Attempting download of new URL file:///C:/ABB.SDS.SDK/host/Server/System.Memory/System.Memory.DLL.
> LOG: Attempting download of new URL file:///C:/ABB.SDS.SDK/host/Server/Apps/Configurator/desktop/System.Memory.DLL.
[...]

PR #2878, opened 2024-09-23
Sync ref project references with runtime projects

edwardneal:
This PR syncs the references of the reference assemblies with those of their respective runtime assemblies. Doing so enables MSBuild's ResolveAssemblyReferences and GenerateBindingRedirects targets to generate the correct binding redirects - so if a downstream library references a later version of Azure.Core or Azure.Identity, VS will generate binding redirects so that SqlClient will use that later version.

Early testing looks promising and I think it'll resolve the issue - but for now, it just relates to #2541 and to #2082.


benrr101 (approved):
This change seems pretty reasonable to me. Let's get the merge conflicts fixed and make sure the test runs pass and we'll get it merged. Thanks for looking into this!

edwardneal:
Thanks @benrr101 - just merged!

mdaigle (approved):

cheenamalhotra (approved):

edwardneal:
@cheenamalhotra can I port this back to the next releases of 5.1 and 5.2? This should let us close #2541 for the remaining supported versions.

cheenamalhotra:
Sure, that makes sense.

PR #2882, opened 2024-09-23
Add streaming support for JSON

apoorvdeshmukh:
This commit adds support for streaming JSON data to and from Sql Server

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
@@ -1994,7 +1994,7 @@ namespace Microsoft.Data.SqlClient
                 {
                     encoding = _metaData[i].encoding;
                 }
-
+                encoding = mt.SqlDbType == SqlDbTypeExtensions.Json ? System.Text.Encoding.UTF8 : encoding;
                 _currentTextReader = new SqlSequentialTextReader(this, i, encoding);
                 _lastColumnWithDataChunkRead = i;
                 return _currentTextReader;
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
@@ -14,6 +14,7 @@ using System.Globalization;
 using System.IO;
 using System.Reflection;
 using System.Runtime.CompilerServices;
+using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using System.Xml;
@@ -1985,7 +1986,11 @@ namespace Microsoft.Data.SqlClient
                 }
 
                 System.Text.Encoding encoding;
-                if (mt.IsNCharType)
+                if (mt.SqlDbType == SqlDbTypeExtensions.Json)
+                {
+                    encoding = new UTF8Encoding();
+                }
+                else if (mt.IsNCharType)
                 {
                     // NChar types always use unicode
                     encoding = SqlUnicodeEncoding.SqlUnicodeEncodingInstance;
@@ -1994,7 +1999,6 @@ namespace Microsoft.Data.SqlClient
                 {
                     encoding = _metaData[i].encoding;
                 }
-                encoding = mt.SqlDbType == SqlDbTypeExtensions.Json ? System.Text.Encoding.UTF8 : encoding;
                 _currentTextReader = new SqlSequentialTextReader(this, i, encoding);
                 _lastColumnWithDataChunkRead = i;
                 return _currentTextReader;
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
@@ -11913,7 +11913,7 @@ namespace Microsoft.Data.SqlClient
 
             encoding = encoding ?? TextDataFeed.DefaultEncoding;
 
-            using (ConstrainedTextWriter writer = encoding == Encoding.UTF8 ? new ConstrainedTextWriter(new StreamWriter(new TdsOutputStream(this, stateObj, null)), size) : new ConstrainedTextWriter(new StreamWriter(new TdsOutputStream(this, stateObj, null), encoding), size))
+            using (ConstrainedTextWriter writer = new ConstrainedTextWriter(new StreamWriter(new TdsOutputStream(this, stateObj, null), encoding), size))
             {
                 if (needBom)
                 {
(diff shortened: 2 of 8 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
```diff
@@ -1994,7 +1994,7 @@ override public TextReader GetTextReader(int i)
                 {
                     encoding = _metaData[i].encoding;
                 }
-
+                encoding = mt.SqlDbType == SqlDbTypeExtensions.Json ? System.Text.Encoding.UTF8 : encoding;
```
mdaigle, line 1997: ✏️ nit, but I usually find it easier to read if this kind of behavior is expressed as mutually exclusive cases. This way it's clear that there's not something special about the relationship between NCharType and Json that would further change the outcome. Even better would be a switch statement, but then we'd have to unravel the IsNCharType determination.

Something like:
```c#
System.Text.Encoding encoding;
if (mt.SqlDbType == SqlDbTypeExtensions.Json) {
    encoding = System.Text.Encoding.UTF8
}
else if (mt.IsNCharType)
{
    // NChar types always use unicode
    encoding = SqlUnicodeEncoding.SqlUnicodeEncodingInstance;
}
else
{
    encoding = _metaData[i].encoding;
}
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -11913,7 +11913,7 @@ private async Task WriteTextFeed(TextDataFeed feed, Encoding encoding, bool need
 
             encoding = encoding ?? TextDataFeed.DefaultEncoding;
 
-            using (ConstrainedTextWriter writer = new ConstrainedTextWriter(new StreamWriter(new TdsOutputStream(this, stateObj, null), encoding), size))
+            using (ConstrainedTextWriter writer = encoding == Encoding.UTF8 ? new ConstrainedTextWriter(new StreamWriter(new TdsOutputStream(this, stateObj, null)), size) : new ConstrainedTextWriter(new StreamWriter(new TdsOutputStream(this, stateObj, null), encoding), size))
```
mdaigle, line 11916: Is this change needed?
These should already be equivalent:
```
new StreamWriter(new TdsOutputStream(this, stateObj, null))
```
```
new StreamWriter(new TdsOutputStream(this, stateObj, null), Encoding.UTF8)
```

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -12910,7 +12910,7 @@ private async Task WriteTextFeed(TextDataFeed feed, Encoding encoding, bool need
             char[] inBuff = new char[constTextBufferSize];
 
             encoding = encoding ?? new UnicodeEncoding(false, false);
-            ConstrainedTextWriter writer = new ConstrainedTextWriter(new StreamWriter(new TdsOutputStream(this, stateObj, null), encoding), size);
+            ConstrainedTextWriter writer = encoding == Encoding.UTF8 ? new ConstrainedTextWriter(new StreamWriter(new TdsOutputStream(this, stateObj, null)), size) : new ConstrainedTextWriter(new StreamWriter(new TdsOutputStream(this, stateObj, null), encoding), size);
```
mdaigle, line 12913: See comment above, I don't think this is needed?

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlDataReader.cs
```diff
@@ -2282,7 +2282,7 @@ override public TextReader GetTextReader(int i)
                 {
                     encoding = _metaData[i].encoding;
                 }
-
+                encoding = mt.SqlDbType == SqlDbTypeExtensions.Json ? System.Text.Encoding.UTF8 : encoding;
```
mdaigle, line 2285: See comments above

mdaigle (changes requested):

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
apoorvdeshmukh: [StreamWriter ](https://learn.microsoft.com/en-us/dotnet/api/system.io.streamwriter.-ctor?view=net-8.0#system-io-streamwriter-ctor(system-io-stream)) without the encoding parameter uses UTF8 encoding with no BOM by default which is what the server also expects for json. In case the encoding is specified explicitly, we need to remove the BOM explicitly before sending the data to the server.
```csharp
        public StreamWriter(Stream stream)
            : this(stream, UTF8NoBOM, DefaultBufferSize, false)
        {
        }

        public StreamWriter(Stream stream, Encoding encoding)
            : this(stream, encoding, DefaultBufferSize, false)
        {
        }
```

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/JsonTest/JsonStreamTest.cs
```diff
@@ -0,0 +1,197 @@
...
+            }
+        }
+
+        private void DeleteFile(string filename)
+        {
+            if (File.Exists(filename))
+            {
+                File.Delete(filename);
+            }
+
```
cheenamalhotra, line 158: Please cleanup extra lines

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
mdaigle: I see. So why not pass in UTF8Encoding.UTF8Encoding() https://learn.microsoft.com/en-us/dotnet/api/system.text.utf8encoding.-ctor?view=net-8.0#system-text-utf8encoding-ctor when we call WriteTextFeed?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
apoorvdeshmukh: Updated with suggested changes.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
apoorvdeshmukh: Done

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
apoorvdeshmukh: Yes. This is not needed anymore.

mdaigle (approved):

cheenamalhotra (approved):

deepaksa1 (approved):

PR #2887, opened 2024-09-25
Project for Common Files

benrr101:
**Description**: As part of the plans for merging projects, I believe it is beneficial to add a project for the common files. There are two big reasons for this: 1) we will eventually want a single project for netfx and netcore, and 2) it allows moving files from projects to common folder without leaving the IDE. 

This new project is not built in any configuration. It should have no impact on the current build procedure. For now, it is sort of a playground for merging the projects.

**Testing**: Local builds work just fine without any problems.

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft.Data.SqlClient.csproj
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft.Data.SqlClient.csproj
@@ -2,6 +2,7 @@
 	<PropertyGroup>
 		<OutputType>ClassLibrary</OutputType>
 		<TargetFrameworks>net6.0;net8.0;net462</TargetFrameworks>
+		<RootNamespace />
 	</PropertyGroup>
   
 </Project>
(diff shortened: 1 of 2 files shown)


src/Microsoft.Data.SqlClient/src/Microsoft.Data.SqlClient.csproj
```diff
@@ -0,0 +1,7 @@
+<Project Sdk="Microsoft.NET.Sdk">
+	<PropertyGroup>
+		<OutputType>ClassLibrary</OutputType>
+		<TargetFrameworks>net6.0;net8.0;net462</TargetFrameworks>
```
cheenamalhotra, line 4: I'm pretty sure there are netfx/netcore specific files - do you intend to disable default compile items and include specific files accordingly?

benrr101: Since the project isn't being built, it doesn't cause much issue to have files included.

If/when we eventually want to make this project the single build project, we will want to turn files on/off based on OS as well as netfx vs netcore. To me, conditional inclusion of the files in the project is not ideal since it makes it difficult to know at-a-glance what files are in the project. As such, I'd suggest using `#if` to wrap the files to include based on .net (already built-in), or OS (will require some consideration).

mdaigle (approved):

PR #2891, opened 2024-09-30
Enhancement tasks for json support

deepaksa1:
This PR eliminates the debug flag from the feature extension that supports JSON, and adjusts some of the test code. It ensures that any pre-existing table is removed before a new one with a similar name is created, utilizing `DataTestUtility.CreateTable()`.

### DIFF
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/JsonTest/JsonTest.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/JsonTest/JsonTest.cs
@@ -64,9 +64,9 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests
 
         [ConditionalFact(typeof(DataTestUtility), nameof(DataTestUtility.IsJsonSupported))]
         public void TestJsonWrite()
-        {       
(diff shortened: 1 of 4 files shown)


src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/JsonTest/JsonTest.cs
```diff
@@ -67,8 +67,7 @@ public void TestJsonWrite()
         {       
             string tableName = DataTestUtility.GetUniqueNameForSqlServer("Json_Test");
```
apoorvdeshmukh, line 68: Is there a scenario for which we may need unique name for `Json_test` and `spJson_WriteTest`? If not, we can avoid it. In case of test failure, we may end up with a leftover table and stored proc with unique name.

apoorvdeshmukh (changes requested):

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/JsonTest/JsonTest.cs
deepaksa1: Done

apoorvdeshmukh (approved):

saurabh500 (approved):

PR #2893, opened 2024-10-01
Merge | Bring netcore Interop to Common Project

benrr101:
**Description**: This is some prerequisite work for merging SqlFileStream. SqlFileStream uses a bunch of interop calls - in netfx they are stored in Safe/UnsafeNativeMethods classes, while in netcore they are stored in the Interop class. Since the netcore interop classes seem newer and "better", I plan to use those as the "merged" versions. But in order to keep merges/moves reasonably separated, I'm presenting this as a separate PR.

The move is pretty low effort here, I just moved the files over, changed their locations in the project netcore project. No changes were made to the netfx project since it does not references these files. (though I intend to clean up these interop files at some point).

**Testing**: Project builds, and since there were no changes to the contents of the files, it should be good to go.


src/Microsoft.Data.SqlClient/src/Interop/Windows/Interop.Errors.cs
```diff
@@ -0,0 +1,15 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+internal partial class Interop
+{
+    // https://msdn.microsoft.com/en-us/library/windows/desktop/ms681382.aspx
+    internal partial class Errors
```
mdaigle, line 8: I see all the other interop file changes as moves, but why were this file and GetFileType_SafeHandle.cs created new?

edwardneal: I'm pretty sure this is because the original `netcore/src/Common/src/Interop/Windows/Interop.Errors.cs` and `netcore/src/Common/src/CoreLib/Interop/Windows/Kernel32/Interop.GetFileType_SafeHandle.cs` need to be deleted.

benrr101: Yep, it looks like either I forgot to remove these files or I didn't pick up the deletion in the commit. Easy fix :+1:

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.EndPoint/TDS.EndPoint.csproj
```diff
@@ -41,8 +41,8 @@
     <Compile Include="TDSServerParser.cs" />
   </ItemGroup>
   <ItemGroup Condition="'$(TargetsWindows)' == 'true'">
-    <Compile Include="$(CommonPath)\Interop\Windows\Interop.Libraries.cs">
-      <Link>Common\Interop\Windows\Interop.Libraries.cs</Link>
+    <Compile Include="$(CommonSourceRoot)\Interop\Windows\Interop.Libraries.cs">
+      <Link>Interop\Windows\Interop.Libraries.cs</Link>
```
David-Engel, line 45: Nit: Does this need to change and be different from the other projects?
```suggestion
      <Link>Common\Interop\Windows\Interop.Libraries.cs</Link>
```

David-Engel (approved):

cheenamalhotra (approved):

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.EndPoint/TDS.EndPoint.csproj
benrr101: I guess the way I look at it, the interop code isn't particularly "common". And, generally speaking I think we should be following a pattern where the file paths match their namespaces. So to me, the common folder seems redundant.

PR #2894, opened 2024-10-02
Implements SqlException.IsTransient based on SqlConfigurableRetryFactory known codes

### LINKED ISSUE #649: Implement SqlException.IsTransient
https://github.com/dotnet/runtime/issues/34817 introduces DbException.IsTransient for .NET 5.0, database-agnostic transient error detection. Note that this is orthogonal to SqlClient's own resilient connection mechanism - IsTransient is only about surfacing error transience to external retry strategies/policies such as Polly.

Note that EF Core's SQL Server provider has [a transient exception detector](https://github.com/dotnet/efcore/blob/bde2b140d6f4cf94d6d1285d402941e20193ec60/src/EFCore.SqlServer/Storage/Internal/SqlServerTransientExceptionDetector.cs), this could be a starting point for implementing this within SqlClient. The goal would eventually be for EF Core to rely on the new IsTransient property rather than contain specific knowledge on SqlClient error codes.

tbolon:
Fix #649,

This new PR is based on feedback on #2127 and uses the already known codes used in the retry logic.

### DIFF
--- a/doc/snippets/Microsoft.Data.SqlClient/SqlException.xml
+++ b/doc/snippets/Microsoft.Data.SqlClient/SqlException.xml
@@ -310,6 +310,14 @@
     <GetObjectData>
       <summary>To be added</summary>
     </GetObjectData>
+    <IsTransient>
+      <summary>
+        Indicates whether the error represented by this <see cref="T:Microsoft.Data.SqlClient.SqlException" /> could be a transient error, i.e. if retrying the triggering operation may succeed without any other change. Examples of transient errors include failure to acquire a database lock, networking issues. This allows automatic retry execution strategies to be developed without knowledge of specific database error codes.
+      </summary>
+      <returns>
+        <see langword="true"/> if the error represented could be a transient error; <see langword="false"/> otherwise.
+      </returns>
+    </IsTransient>
     <LineNumber>
       <summary>
         Gets the line number within the Transact-SQL command batch or stored procedure that generated the error.
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlException.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlException.cs
@@ -96,6 +96,28 @@ namespace Microsoft.Data.SqlClient
         /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlException.xml' path='docs/members[@name="SqlException"]/Class/*' />
         public byte Class => Errors.Count > 0 ? Errors[0].Class : default;
 
+#if NET6_0_OR_GREATER
+        /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlException.xml' path='docs/members[@name="IsTransient"]/IsTransient/*' />
+        public override bool IsTransient
+        {
+            get
+            {
+                if (Errors.Count == 0)
+                    return false;
+
+                for (int i = 0; i < Errors.Count; i++)
+                {
+                    if (!SqlConfigurableRetryFactory.IsKnownTransientErrorCode(Errors[i].Number))
+                    {
+                        return false;
+                    }
+                }
+
+                return true;
+            }
+        }
+#endif
+
         /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlException.xml' path='docs/members[@name="SqlException"]/LineNumber/*' />
         public int LineNumber => Errors.Count > 0 ? Errors[0].LineNumber : default;
 
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Reliability/SqlConfigurableRetryFactory.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Reliability/SqlConfigurableRetryFactory.cs
@@ -77,6 +77,8 @@ namespace Microsoft.Data.SqlClient
             => InternalCreateRetryProvider(retryLogicOption,
                                             retryLogicOption != null ? new SqlFixedIntervalEnumerator(retryLogicOption.DeltaTime, retryLogicOption.MaxTimeInterval, retryLogicOption.MinTimeInterval) : null);
 
+        internal static bool IsKnownTransientErrorCode(int errorCode) => s_defaultTransientErrors.Contains(errorCode);
+
         private static SqlRetryLogicBaseProvider InternalCreateRetryProvider(SqlRetryLogicOption retryLogicOption, SqlRetryIntervalBaseEnumerator enumerator)
         {
             if (retryLogicOption == null)


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlException.cs
```diff
@@ -96,6 +96,28 @@ public override void GetObjectData(SerializationInfo si, StreamingContext contex
...
+        public override bool IsTransient
+        {
+            get
+            {
+                if (Errors.Count == 0)
+                    return false;
+
+                for (int i = 0; i < Errors.Count; i++)
+                {
+                    if (!SqlConfigurableRetryFactory.IsKnownTransientErrorCode(Errors[i].Number))
```
edwardneal, line 110: Thank you for this - it'll work for the general case. A user can also [configure SqlClient](https://learn.microsoft.com/en-us/sql/connect/ado-net/configurable-retry-logic-config-file-sqlclient?view=sql-server-ver16) to pass a custom set of error numbers to a built-in retry logic method, and can write their own SqlRetryLogicBase derivative with a custom TransientPredicate implementation (or a completely custom SqlRetryLogicBaseProvider derivative.) If this retry logic fails, it'll throw an AggregateException containing the various SqlException instances.

In a situation where the user's configured a custom list of transient errors (or defined their own TransientPredicate implementation), could they potentially receive an AggregateException containing a set of SqlExceptions where IsTransient is false?

If they could and this is a problem, one way to handle these custom sets of error numbers and the SqlRetryLogicBase derivative might be with some work in SqlRetryLogicProvider. I'm not sure how a SqlRetryLogicBaseProvider derivative could be handled though.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlException.cs
```diff
@@ -96,6 +96,28 @@ public override void GetObjectData(SerializationInfo si, StreamingContext contex
         /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlException.xml' path='docs/members[@name="SqlException"]/Class/*' />
         public byte Class => Errors.Count > 0 ? Errors[0].Class : default;
 
+#if NET6_0_OR_GREATER
+        /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlException.xml' path='docs/members[@name="IsTransient"]/IsTransient/*' />
+        public override bool IsTransient
```
cheenamalhotra, line 101: I'd recommend making a shared utility of `IsTransientError(SqlException ex)` in a separate file that is currently defined in `SqlInternalConnectionTds` class and calling that from here, as well as in the `SqlInternalConnectionTds` ctor (as it is now).

cheenamalhotra: You can extract the `transientErrors` to be defined in common type and use the same here too, so driver implementation that is used internally is the same as what this API exposes.

doc/snippets/Microsoft.Data.SqlClient/SqlException.xml
```diff
@@ -310,6 +310,14 @@
     <GetObjectData>
       <summary>To be added</summary>
     </GetObjectData>
+    <IsTransient>
+      <summary>
+        Indicates whether the error represented by this <see cref="T:Microsoft.Data.SqlClient.SqlException" /> could be a transient error, i.e. if retrying the triggering operation may succeed without any other change. Examples of transient errors include failure to acquire a database lock, networking issues. This allows automatic retry execution strategies to be developed without knowledge of specific database error codes.
```
paulmedynski, line 315: Could be transient?  I think we need to be certain it's transient :)  Retrying may or may not succeed, but when IsTransient is true, the driver is 100% confident the action may be retried.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlException.cs
```diff
@@ -96,6 +96,28 @@ public override void GetObjectData(SerializationInfo si, StreamingContext contex
         /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlException.xml' path='docs/members[@name="SqlException"]/Class/*' />
         public byte Class => Errors.Count > 0 ? Errors[0].Class : default;
 
+#if NET6_0_OR_GREATER
```
paulmedynski, line 99: We just use `#if NET` now.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlException.cs
```diff
@@ -96,6 +96,28 @@ public override void GetObjectData(SerializationInfo si, StreamingContext contex
...
+        /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlException.xml' path='docs/members[@name="IsTransient"]/IsTransient/*' />
+        public override bool IsTransient
+        {
+            get
```
paulmedynski, line 103: This needs to be an immutable property of the exception.  The transient-ness must be determined at construction time.  As written now, the exception may be transient when constructed, but later on, when someone inspects this property, it may no longer be transient.  The transient error codes may change independently from any exception instances.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlException.cs
```diff
@@ -96,6 +96,28 @@ public override void GetObjectData(SerializationInfo si, StreamingContext contex
...
+            {
+                if (Errors.Count == 0)
```
paulmedynski, line 105: Braces for all if-statement blocks, please.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlException.cs
```diff
@@ -96,6 +96,28 @@ public override void GetObjectData(SerializationInfo si, StreamingContext contex
...
+                    return false;
+
+                for (int i = 0; i < Errors.Count; i++)
```
paulmedynski, line 108: Prefer foreach:
```
foreach (var number in Errors)
{
  if (! SqlConfigurableRetryFactory.IsKnownTransientErrorCode(number))
  {
    ...
  }
  ...
}
```

paulmedynski (changes requested):
Thank you for this contribution!  We have a few comments and suggestions.

paulmedynski:
Closing due to inactivity from author.

tbolon:
Sorry for not giving news, we are still using System.Data.SqlClient in our projects and I did not took the time to reopen this PR.
I will create a new PR from scratch based on your feedbacks when we migrate to MS.SqlClient.

PR #2899, opened 2024-10-07
Merge | SqlFileStream (Opt 2)

benrr101:
**Description**: This PR aims to merge `SqlFileStream` from the netfx and netcore projects. The strategy for this merge is to use the Interop.* libraries from the netcore project to enable merging the classes. Therefore, the code for this class is mostly based on the netcore implementation with a handful of `#if NETFRAMEWORK` blocks to add overrides for netfx behavior.

The commits are broken up into fairly digestible chunks that explain each step along the process. I'd recommend reviewing the PR by stepping through the commits and seeing if you agree with the changes. This version of the merge does not include any extra cleanup at the end of the commit chain. It should have less changes going on overall. If this PR is accepted first, the code cleanup commits will be sent out as a separate PR. If you'd rather just get it all out of the way, take a look at the Opt 1 PR.

**Testing**: The existing manual tests for SqlFileStream were used to validate behavior.


benrr101:
No longer needed as option 1 was accepted

PR #2907, opened 2024-10-09
Make Sure Common Project Doesn't Built

benrr101:
**Description**: I'm not entirely sure what was causing the project to be included in solution-based builds (ie, VS, etc), but I've verified that it is not being included any longer. Building the solution in VS works as expected without any errors from the common project.

**Testing**: I've verified the solution builds correctly in VS 2022 and Rider. It never built in CI pipelines, and there aren't any changes to make it build in pipelines.

Please verify it works on your system. It may be necessary to run a `git clean -xfd` on your solution root folder if you continue to encounter build issues.

### DIFF
--- a/src/Microsoft.Data.SqlClient.sln
+++ b/src/Microsoft.Data.SqlClient.sln
@@ -2,7 +2,7 @@ Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio Version 17
 VisualStudioVersion = 17.0.31912.275
 MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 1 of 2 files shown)


David-Engel (approved):
> It may be necessary to run a `git clean -xfd` on your solution root folder if you continue to encounter build issues.

Even though a build "succeeded", I did have to run git clean to get rid of all the errors reported in VS. Even a rebuild (clean+build) in VS didn't get rid of them.

benrr101:
@David-Engel Yeah, a "clean" doesn't really do much of anything. It leaves behind loads of files and the best way to properly clean up a solution is to `git -xfd`. But thanks for confirming that worked.

cheenamalhotra (approved):
Worked for me too! Thanks for addressing this!

PR #2909, opened 2024-10-10
Merge | DbConnectionInternal (w/Cleaning)

benrr101:
**Description**: This PR aims to merge `DbConnectionInternal` from the netfx and netcore projects. The strategy for this merge is to move the netfx version of the class over (since it has the entire class in one file) and do a two-step merge of the two netcore files. Once combined, the bulk of the code has no differences between the netfx and netcore implementations. The only major difference is that netfx has a collection of performance counters that are updated in certain scenarios.

The commits are broken up into fairly digestible chunks that explain each step along the process. I'd recommend reviewing the PR by stepping through the commits and seeing if you agree with the changes. Nevertheless, this PR has a lot of code cleanup changes at the end that make the overall PR fairly large. In anticipation of pushback on this, I will offer a separate PR that only contains the merge changes. If that one is accepted, I'll still send out the cleanup change on a separate PR. But, I think this option would be faster overall. Ideally there is some other cleanup tasks that could be done (public -> internal, protected -> private, etc), but those changes require larger picture understanding of the class hierarchy. 

**Behavioral Changes**: The method `DbConnectionInternal.DoomThisConnection()` logs a trace event, but `DbConnectionInternal.UnDoomThisConnection()` does not. I added a trace event of the same format to `UnDoomThisConnection` to make it match. 

**Testing**: Most of the functional tests run locally, though I'm having some difficulty with getting all tests to work. These issues happen main branch as well, so I assume it is unrelated to the code changes.


benrr101:
Moving back to draft to since the merge-only PR got merged.

PR #2910, opened 2024-10-10
Merge | DbConnectionInternal (Merge Only)

benrr101:
**Description**: This PR aims to merge `DbConnectionInternal` from the netfx and netcore projects. The strategy for this merge is to move the netfx version of the class over (since it has the entire class in one file) and do a two-step merge of the two netcore files. Once combined, the bulk of the code has no differences between the netfx and netcore implementations. The only major difference is that netfx has a collection of performance counters that are updated in certain scenarios.

The commits are broken up into fairly digestible chunks that explain each step along the process. I'd recommend reviewing the PR by stepping through the commits and seeing if you agree with the changes. This version of the PR only contains the merge, no cleanup changes were made.

**Testing**: Most of the functional tests run locally, though I'm having some difficulty with getting all tests to work. These issues happen main branch as well, so I assume it is unrelated to the code changes.


src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbConnectionInternal.cs
```diff
@@ -268,25 +273,29 @@ internal bool IsEmancipated
...
                 //
                 // Now, our job is to return TRUE when the connection is out
                 // of the pool and it's owning object is no longer around to
                 // return it.
 
+                #if NETFRAMEWORK
                 return !IsTxRootWaitingForTxEnd && (_pooledCount < 1) && !_owningObject.TryGetTarget(out _);
+                #else
+                return (_pooledCount < 1) && !_owningObject.TryGetTarget(out _);
+                #endif
```
edwardneal, line 298: This behavioural difference pre-dates the initial commit of SqlClient, so it's not just a matter of a PR making a change to netfx but not to netcore.

I think this is actually a bugfix in netfx which wasn't cross-ported to netcore. `IsTxRootWaitingForEnd` just returns whether or not this is currently in stasis, and the "is in stasis" flag is set by `SetInStasis`. That method is called within DbConnectionPool, and has this comment:
> // SQLHotfix# 50003503 - connections that are affiliated with a 
> //   root transaction and that also happen to be in a connection 
> //   pool that is being shutdown need to be put in stasis so that 
> //   the root transaction isn't effectively orphaned with no 
> //   means to promote itself to a full delegated transaction or
> //   Commit or Rollback

Would you mind digging into that hotfix to see whether or not this is correct? If so, we should make .NET Core behave in the same way - whether as part of this PR, or as a separate one-line PR.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbConnectionPoolCounters.netfx.cs
```diff
@@ -2,6 +2,8 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+#if NETFRAMEWORK
+
```
edwardneal, line 6: The performance counters and event counters are a little irritating to merge, and aren't always incremented or decremented in precisely the same places. Rather than moving DbConnectionPoolCounters (and the event counters equivalent) into the shared project, I'd rather we look at the OpenTelemetry metrics issue and replace them both with a single shared implementation.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlInternalConnectionSmi.stub.cs
```diff
@@ -0,0 +1,16 @@
...
+
+#if NETFRAMEWORK
+
+namespace Microsoft.Data.SqlClient
+{
+    // DO NOT USE THIS FILE IN ANY PROJECT!
+    // This is a temporary stub to enable migrating DbConnectionInternal to the common project.
+    internal class SqlInternalConnectionSmi
+    {
+    }
```
edwardneal, line 13: For awareness: I'll de-stub this once the context connections PR is merged.

edwardneal (commented):
I'm leaving my comments here, but they should all apply to #2909 too.

The merge looks correct to me - I didn't see any logic change between the files, there are no missing/extra members. There's one logical difference between netfx and netcore though, which might be the result of a bugfix not being ported from netfx.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbConnectionPoolCounters.netfx.cs
benrr101: Although we've got someone exploring the OpenTelemetry implementation, I think it will be a ways off compared to bringing together the codebases. Can we just move the performance counters over for now and when otel is going, we can remove it?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlInternalConnectionSmi.stub.cs
benrr101: Works for me! I'm adding this for now just so I don't get red squigglies when working in the common project. (at least for the files in this PR - I know a lot of other ones in the project are missing dependencies in the common project)

src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbConnectionPoolCounters.netfx.cs
edwardneal: If OpenTelemetry's not going to be available for a while, I've got no objections. In that case, can you add the file to both projects to keep the approach consistent please?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbConnectionInternal.cs
benrr101: So, I think the older ticket systems (buvsts, sqlhotfix, etc) don't exist anymore, and as such I don't think I can pull up anymore information on this. I'm checking with the team to see if there are archived copies of the tickets somewhere, but I suspect we're on our own now.

Personally I think it makes sense that if the there is a tx that the connection is waiting on, then the connection isn't emancipated. But I genuinely don't understand the intricacies of what it means to be emancipated, in stasis, etc...

edwardneal: The .NET Framework behaviour definitely makes sense, and it's very old behaviour. The oldest version of the reference source is .NET 4.5.1, and the netfx logic's present in that version; it also exists in the .NET 3.5 SP1 assemblies.

I've spent more time investigating, and I think there's another possibility. System.Transactions wasn't implemented until .NET Core 2.0, and System.Data.SqlClient predates that. Maybe System.Data.SqlClient was ported to .NET Core without this logic, (because transactions weren't supported by the underlying runtime) and when support for transactions was added it just wasn't added back to DbConnectionInternal. If that's the case, we should be fine to use the .NET Framework behaviour.

David-Engel: > I've spent more time investigating, and I think there's another possibility. System.Transactions wasn't implemented until .NET Core 2.0, and System.Data.SqlClient predates that. Maybe System.Data.SqlClient was ported to .NET Core without this logic, (because transactions weren't supported by the underlying runtime) and when support for transactions was added it just wasn't added back to DbConnectionInternal. If that's the case, we should be fine to use the .NET Framework behaviour.

That is the case. Distributed transactions came in .NET 7 so the original SDS Core code was likely missing some bits.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbConnectionPoolCounters.netfx.cs
benrr101: You mean add DbConnectionPoolCounters to both netfx and netcore projects? Netcore didn't have it at the time of doing this merge, and I don't really want to make big functional changes like that during these merges. Unless Unless I'm missing something...

edwardneal: Not exactly - I'm asking for the .cs file to be included in both projects (so the approach for framework-specific classes is consistent with the .NET 6.0-only types) and for the contents of the file to be wrapped in `#if NETFRAMEWORK` as they already are in this PR.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbConnectionInternal.cs
benrr101: Ok, I adopted the netfx implementation of the return statement. Let me know if there's any issues with that.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbConnectionPoolCounters.netfx.cs
benrr101: 🤔 I don't see a precedent for including netfx/netcore files in the opposite project... at least not by searching for netfx in the netcore csproj, or vice versa. DbConnectionPoolCounters is in the common project now and is wrapped with `#if NETFRAMEWORK`

src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbConnectionInternal.cs
edwardneal: Thanks - that explanation makes sense, and I agree that netfx has the better implementation to use.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbConnectionPoolCounters.netfx.cs
edwardneal: > DbConnectionPoolCounters is in the common project now and is wrapped with #if NETFRAMEWORK

Thanks, I'd forgotten about the common project. If it's in there, we probably don't need it in the netcore project.

It's academic with the common project in place, but the precedent I had in mind was the set of files containing the [`NetSecurityNative`](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/netcore/src/Common/Interop/Unix/System.Net.Security.Native/Interop.NetSecurityNative.cs) PInvoke and associated types; these are included in both target frameworks for the netcore project, even though they're only used in one of them.

David-Engel (approved):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbConnectionPoolCounters.netfx.cs
benrr101: Forgotten about the common project 😲 I figured that's the whole point of this work 😄

cheenamalhotra (approved):

PR #2915, opened 2024-10-11
Remove async from SNITcpHandle

David-Engel:
_*For easier code review, hide whitespace changes in this PR._

In our quest to remove async use on sync paths, I've refactored parallel connections in SNITcpHandle.cs. The new code leverages non-blocking sockets and Socket.Select in a single thread to connect in parallel.

To test, I created multiple semi-random entries in my local hosts file like:

```
127.0.0.1	test-msf
::2		test-msf
193.168.176.1	test-msf
11.13.99.44     test-msf
```

And used the following TCPConnectionString in config.json:
`Data Source=tcp:test-msf;Database=Northwind;User=sa;Password=...;Encrypt=false;MultiSubnetFailover=true;`

Here is a snippet of traces from SNITcpHandle for a failed run (invalid port: `Data Source=tcp:test-msf,1444`):
```
SNICommon.GetDnsIpAddresses | INFO | Getting DNS host entries for serverName test-msf.
SNITCPHandle.ParallelConnect | INFO | Connecting to IP address 128.0.0.1 and port 1444 using InterNetwork address family. Is infinite timeout: False
SNITCPHandle.ParallelConnect | INFO | Connecting to IP address 193.168.176.1 and port 1444 using InterNetwork address family. Is infinite timeout: False
SNITCPHandle.ParallelConnect | INFO | Connecting to IP address 11.13.99.44 and port 1444 using InterNetwork address family. Is infinite timeout: False
SNITCPHandle.ParallelConnect | INFO | Connecting to IP address 10.12.0.1 and port 1444 using InterNetwork address family. Is infinite timeout: False
SNITCPHandle.ParallelConnect | INFO | Connecting to IP address ::2 and port 1444 using InterNetworkV6 address family. Is infinite timeout: False
SNITCPHandle.ParallelConnect | INFO | Determining the status of sockets during the remaining timeout of 47710000 microseconds.
SNITCPHandle.ParallelConnect | INFO | Failed to connect to endpoint: ::2. Error: null
SNITCPHandle.ParallelConnect | INFO | Determining the status of sockets during the remaining timeout of 47709000 microseconds.
SNITCPHandle.ParallelConnect | INFO | Failed to connect to endpoint: 10.12.0.1. Error: null
SNITCPHandle.ParallelConnect | INFO | Determining the status of sockets during the remaining timeout of 34270000 microseconds.
SNITCPHandle.ParallelConnect | INFO | Failed to connect to endpoint: 128.0.0.1. Error: null
SNITCPHandle.ParallelConnect | INFO | Determining the status of sockets during the remaining timeout of 26904000 microseconds.
SNITCPHandle.ParallelConnect | INFO | Failed to connect to endpoint: 193.168.176.1. Error: null
SNITCPHandle.ParallelConnect | INFO | Determining the status of sockets during the remaining timeout of 26858000 microseconds.
SNITCPHandle.ParallelConnect | INFO | Failed to connect to endpoint: 11.13.99.44. Error: null
SNITCPHandle.ParallelConnect | INFO | Disposing non-selected socket for endpoint: 128.0.0.1
SNITCPHandle.ParallelConnect | INFO | Disposing non-selected socket for endpoint: 193.168.176.1
SNITCPHandle.ParallelConnect | INFO | Disposing non-selected socket for endpoint: 11.13.99.44
SNITCPHandle.ParallelConnect | INFO | Disposing non-selected socket for endpoint: 10.12.0.1
SNITCPHandle.ParallelConnect | INFO | Disposing non-selected socket for endpoint: ::2
SNITCPHandle.ParallelConnect | ERR | No connections succeeded. Last error: null
```

Here is a snippet of traces from SNITcpHandle for a successful run (one valid IP):
```
SNICommon.GetDnsIpAddresses | INFO | Getting DNS host entries for serverName test-msf.
SNITCPHandle.ParallelConnect | INFO | Connecting to IP address 193.168.176.1 and port 1433 using InterNetwork address family. Is infinite timeout: False
SNITCPHandle.ParallelConnect | INFO | Connecting to IP address 11.13.99.44 and port 1433 using InterNetwork address family. Is infinite timeout: False
SNITCPHandle.ParallelConnect | INFO | Connecting to IP address 10.12.0.1 and port 1433 using InterNetwork address family. Is infinite timeout: False
SNITCPHandle.ParallelConnect | INFO | Connecting to IP address 127.0.0.1 and port 1433 using InterNetwork address family. Is infinite timeout: False
SNITCPHandle.ParallelConnect | INFO | Connecting to IP address ::2 and port 1433 using InterNetworkV6 address family. Is infinite timeout: False
SNITCPHandle.ParallelConnect | INFO | Watching pending sockets during the remaining timeout of 47724000 microseconds.
SNITCPHandle.ParallelConnect | INFO | Connected to endpoint: 127.0.0.1:1433
SNITCPHandle.ParallelConnect | INFO | Disposing non-selected socket for endpoint: 193.168.176.1
SNITCPHandle.ParallelConnect | INFO | Disposing non-selected socket for endpoint: 11.13.99.44
SNITCPHandle.ParallelConnect | INFO | Disposing non-selected socket for endpoint: 10.12.0.1
SNITCPHandle.ParallelConnect | INFO | Disposing non-selected socket for endpoint: ::2
```


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
```diff
@@ -396,217 +366,299 @@ private static IEnumerable<IPAddress> GetHostAddressesSortedByPreference(string
...
+                    {
+                        SqlClientEventSource.Log.TryAdvancedTraceEvent(
+                            "{0}.{1}{2}ParallelConnect timeout expired.", nameof(SNITCPHandle), nameof(ParallelConnect), EventType.INFO);
+                        break;
                     }
 
-                    if (isConnected)
+                    // See if one of the sockets in the Selected list is connected without errors (exists in read list or write list but not error list)
+                    // Concat read/write lists since we treat them the same from here
+                    checkWriteLst.AddRange(checkReadLst);
```
edwardneal, line 593: Should we definitely treat the lists of readable and writable sockets in the same way?

The documentation for [Socket.Select](https://learn.microsoft.com/en-us/dotnet/api/system.net.sockets.socket.select?view=net-8.0) says:
> If you are in a listening state, readability means that a call to Accept will succeed without blocking. If you have already accepted the connection, readability means that data is available for reading. In these cases, all receive operations will succeed without blocking. **Readability can also indicate whether the remote Socket has shut down the connection**; in that case a call to Receive will return immediately, with zero bytes returned.

Both handshakes start with a client-initiated write, (the TDS handshake sends a PRELOGIN packet, the TLS handshake sends a Client Hello) so I wouldn't expect the client to be able to read data from the socket immediately. This leaves the scenario I've bolded above - a remote socket accepts the connection, but then immediately shuts it down after the TCP handshake is complete.

I haven't tested this, but it seems that a socket in `checkReadLst` should actually be treated the same way as a socket in `checkErrorLst` - we should skip any sockets contained in either list.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
```diff
@@ -396,217 +366,299 @@ private static IEnumerable<IPAddress> GetHostAddressesSortedByPreference(string
...
+                        { // timeout loop for when select timed out but no sockets have changed state
+                          // Socket.Select timeout is in microseconds and can only handle up to int.MaxValue
+                          // This means our connection timeout can be greater than the max select timeout
+                            socketSelectTimeout =
                                 checked((int)(Math.Min(timeout.MillisecondsRemainingInt, int.MaxValue / 1000) * 1000));
 
-                            checkReadLst = new List<Socket>(1) { socket };
-                            checkWriteLst = new List<Socket>(1) { socket };
-                            checkErrorLst = new List<Socket>(1) { socket };
+                            checkReadLst = new List<Socket>(socketsInFlight.Count);
```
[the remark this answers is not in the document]
edwardneal: They need to be reset every iteration - Socket.Select will mutate these lists when we pass them to it.

We can avoid three allocations per iteration of the loop though; we can clear the lists and re-add `socketsInFlight` to them. If we do this, then the variable declarations for all three lists should change from:
`List<Socket> checkReadLst = new()`
to
`List<Socket> checkReadLst = new(socketsInFlight.Count);`

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
David-Engel: I was unsure about this scenario. In the event a server shuts down the socket immediately after the handshake is complete, would this update the Connected property? I think this is the DNN case that is fixed in #2877. Unfortunately, I don't have an environment to test this in. It might be good to write a simple test server to be able to test socket behavior. Anyone have the time / desire and want to drive this forward? If not, I can try to get to it later.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
```diff
@@ -396,217 +366,299 @@ private static IEnumerable<IPAddress> GetHostAddressesSortedByPreference(string
...
-                        }
-
-                        isConnected = true;
+                        socket.Connect(address, port);
+                        throw SQL.SocketDidNotThrow();
                     }
-                    catch (SocketException socketException) when (!isInfiniteTimeout && 
-                                                                  socketException.SocketErrorCode == SocketError.WouldBlock)
+                    catch (SocketException socketException) when (socketException.SocketErrorCode == SocketError.WouldBlock)
                     {
```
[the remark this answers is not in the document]
David-Engel: We are specifically catching the expected exception that is thrown when calling Connect on a non-blocking socket. (We throw on line 521 if Connect() didn't throw.) Logging this exception would be pointless.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
edwardneal: I think the test server would be a good idea; it'd need to tackle two situations:
1. Accept, then close a TCP connection (the DNN case, along with Istio/Envoy)
2. Accept a TCP connection, then immediately start sending bytes to it (so SqlClient ignores sockets which don't comply with the TDS spec.)
I'll make a note to pick it up later, once we're happy with SNITcpHandle's behaviour.

Without testing, I'd expect you to be correct about the Connected property; it'd be strange for Socket.Select to reflect a different state to the top-level property on Socket. If we're just interested in testing whether or not the socket is connected, (i.e. #2877, situation 1) I don't think we'd need to include any reference to `checkReadLst` - whether to include it in the list of sockets to check, or to ensure that none of our sockets are in the list.

This just leaves situation 2, where another service is running on that port and immediately starts sending data. We can exclude that from the MultiSubnetFailover socket selection pretty trivially at this stage by making sure that none of the connected sockets are in `checkReadLst` - a server which immediately sends something to every connected client is violating the TDS protocol from the outset.

As a slight aside, while #2877 should fix DNNs, its associated issue opened up an interesting topic: TCP proxies such as Envoy (and by extension, Istio, which was the original service mentioned and which seems to encapsulate Envoy.) These proxies decouple the front-end acceptance of inbound client connections from the establishment of connections to the back-end SQL Servers. I'm fairly sure this wouldn't be compatible with the current MSF implementation, but haven't gotten around to testing that. I'm not actually sure they can/should be made compatible, either: the design of bypassing MSF/read-only routing and putting a TCP-level load balancer in front of X stateful SQL Server containers (which aren't guaranteed to be in sync in the way that read-only routing is supposed to guarantee) strikes me as a bad idea.

David-Engel: I couldn't resist trying #1, which led me down a rabbit hole testing with a local mock server. My #2877 fix is unlikely to do anything for the DNN situation. (I added a comment there.) I think we can ignore/discard the checkReadList, as you said.

Need more time to add details here. Will try to add some tomorrow. There are a lot of potential negative scenarios to test here that could lead to a big time suck.

edwardneal: Thanks for checking this. The more I think about the edge cases here, the more I think that MultiSubnetFailover should work in terms of application-level connection state. Perhaps we shouldn't just pick the first socket to establish a TCP connection, but should pick the first IP address to respond to a PRELOGIN request. That eliminates the need for the pause to allow connections to stabilise and lets us fold the various failure states into a single code path, but brings a host of other implications which need a lot more testing.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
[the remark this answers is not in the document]
David-Engel: Optimized

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
```diff
@@ -396,217 +366,299 @@ private static IEnumerable<IPAddress> GetHostAddressesSortedByPreference(string
...
-                        List<Socket> checkReadLst;
-                        List<Socket> checkWriteLst;
-                        List<Socket> checkErrorLst;
+                    }
+                }
 
+                // https://github.com/dotnet/SqlClient/issues/826#issuecomment-736224118
+                // Socket.Select is used because it supports timeouts, while Socket.Connect does not
+                // Socket.Select also allows us to wait for multiple sockets using a single thread
+                // Socket.Select will return as soon any any socket in the list meets read/write/error
```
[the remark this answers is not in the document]
David-Engel: I updated the comment

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
David-Engel: After validating, I changed the logic to only consider the write list. As for the DNN issue, after talking with CSS, it seems their scenario is very specific. They are trying to replicate DNNs on-prem and their networking stack is behaving differently than Azure setups. This is unlikely to be common to anyone else so I'm leaving out any changes specific to that issue.

cheenamalhotra (approved):

PR #2927, opened 2024-10-25
Drop net6 support

mdaigle:
This PR includes the minimum necessary to drop net6 support.

I have two follow up PRs planned:
- update documentation and build guides
- update net6 symbol checks to net8 in test and common directories and remove net6 symbol checks from the netcore directory


ErikEJ:
LGTM, I think you forgot to remove some more lines from the .nuspec file.

ErikEJ:
Looks like this file can be deleted now: 

`src\Microsoft.Data.SqlClient\src\Microsoft\Data\SqlClient\SSPI\ManagedSSPIContextProvider.cs`

edwardneal:
It can. Almost everything beneath netcore/src/Common/System can join it, as can everything beneath netcore/src/Common/Interop/Unix, and quite a few files from netcore/src/Common/Interop/Windows. It'll also close #303 by default.

benrr101 (approved):
While this is definitely a minimal set of changes required for removing net6 support, there will be a lot of code removal required as well. If you can create an item in ado for that (and optionally assign to me), that'd be great. Any opportunity to remove derelect code is appreciated!

eng/pipelines/dotnet-sqlclient-ci-core.yml
```diff
@@ -12,12 +12,17 @@ parameters: # parameters are shown up in ADO UI in a build queue time
...
 - name: targetFrameworksLinux
   displayName: 'Target Frameworks on Non-Windows'
   type: object
-  default: [net6.0, net8.0]
+  default: [net8.0]
+
+- name: netcoreVersionTestUtils
+  displayName: 'Netcore Version for Test Utilities'
+  type: object
+  default: net8.0
```
mdaigle, line 25: This version is set here, but passed down through to the test util project compilation. Trying to keep this at the highest level possible so that it's easy to bump in the future without digging down in the individual jobs.

cheenamalhotra (approved):

PR #2930, opened 2024-10-26
remove unnecessary System.Text.Json dependency on modern .NET

mus65:
Partial revert of #2921 . System.Text.Json is part of the framework on modern .NET, so this dependency is unnecessary and wil just cause false positives in NuGet audit.


mus65:
@dotnet-policy-service agree

saurabh500 (approved):

David-Engel (approved):

tools/props/VersionsNet8OrLater.props
```diff
@@ -5,7 +5,6 @@
     <SystemConfigurationConfigurationManagerVersion>8.0.1</SystemConfigurationConfigurationManagerVersion>
     <MicrosoftExtensionsCachingMemoryVersion>8.0.1</MicrosoftExtensionsCachingMemoryVersion>
     <MicrosoftDotNetRemoteExecutorVersion>9.0.0-beta.24157.1</MicrosoftDotNetRemoteExecutorVersion>
-    <SystemTextJsonVersion>8.0.5</SystemTextJsonVersion>
```
cheenamalhotra, line 8: This is needed in order to override System.Text.Json included version with non-vulnerable version.

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.csproj
```diff
@@ -28,6 +28,5 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="Azure.Identity" Version="$(AzureIdentityVersion)" />
-    <PackageReference Include="System.Text.Json" Version="$(SystemTextJsonVersion)" />
```
cheenamalhotra, line 31: This is required as System.Text.Json is exposed by SqlJson API

cheenamalhotra (changes requested):
I don't agree with the changes, as this is added purposely. What false positives are you concerned about?

mus65:
> I don't agree with the changes, as this is added purposely. What false positives are you concerned about?

This dependency is simply not necessary since the .NET runtime itself already contains System.Text.Json . This causes false positives in NuGet Audit because if a library like Microsoft.Data.SqlClient e.g. depends on the vulnerable System.Text.Json 8.0.4, but I'm actually using the latest .NET runtime which contains the fixed 8.0.5, the SDK will use the latest version. Meaning I'm not actually affected by the vulnerability, but NuGet Audit will still claim that I am.

NuGet is actually working on fixing this, so it considers the runtime dependencies (see [Supplied by Platform](https://devblogs.microsoft.com/nuget/nugetaudit-2-0-elevating-security-and-trust-in-package-management/#supplied-by-platform) ). But even when they fix this, not having this unnecessary dependency in the first place is imho the better option.

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.csproj
benrr101: @apoorvdeshmukh @deepaksa1 can you confirm this?

cheenamalhotra:
Based on comment https://github.com/NuGet/Home/issues/7344#issuecomment-2311075251, I believe Supplied by Platform is not yet available in .NET 8 or 9. Can you confirm my understanding?

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.csproj
apoorvdeshmukh: Yes, this is correct.

ErikEJ:
@cheenamalhotra that is correct, but only affects potential nuget audit warnings. Adding this as an explicit reference with .NET 8 is wrong.

cheenamalhotra (approved):

mungojam:
I hope this can be backported unless v6 is coming very soon and in ef SQL by default. Every time a system.text.json vuln comes up we're having to update it explicitly just because of SQL client lib, other Ms libs don't show the same explicit dependency

ErikEJ:
@mungojam Why backported? This was introduced in 6.x

mungojam:
> Partial revert of #2921 . System.Text.Json is part of the framework on modern .NET, so this dependency is unnecessary and wil just cause false positives in NuGet audit.

It will still be needed for .net framework consumers so you might need a conditional package reference instead

tools/specs/Microsoft.Data.SqlClient.nuspec
```diff
@@ -46,7 +46,6 @@ When using NuGet 3.x this package requires at least version 3.4.</description>
         <dependency id="Microsoft.IdentityModel.Protocols.OpenIdConnect" version="7.5.0" />
         <dependency id="Microsoft.SqlServer.Server" version="1.0.0"/>
         <dependency id="System.Configuration.ConfigurationManager" version="8.0.1" exclude="Compile" />
-        <dependency id="System.Text.Json" version="8.0.5" />
       </group>
       <group targetFramework="net6.0">
```
mdaigle, line 50: @mus65 FYI, .NET 6 support was just dropped from the main branch today.

mus65: I merged main and fixed the conflict.

mus65:
@mungojam 
> It will still be needed for .net framework consumers so you might need a conditional package reference instead

The .NET Framework reference is still there, it was already conditional because .NET Framework/.NET have different csproj files.

mdaigle (approved):

mungojam:
> @mungojam Why backported? This was introduced in 6.x

@ErikEJ my mistake, it comes in with sub-dependencies in v5. Hopefully these don't bring it in in v6:

image

PR #2933, opened 2024-10-28
Merge | Pull in the rest of the interop files

benrr101:
**Description**: This PR is pretty straightforward, it brings in the rest of the files used for "interop". This is the first stage of a series of PRs that I've been working on that migrate the interop _class_ to an interop _namespace_. This is a prerequisite for a prerequisite for merging LocalDBAPI. It is not strictly required, but it will make my life easier for merging the LocalDBAPI, and ideally it is work that should be done eventually anyhow.

Keep in mind if #2927 goes through, much of this code can be deleted since it is no longer needed in net8 or above.

This PR only moves the files, it does not make any code changes.

**Testing**: Projects still build, so there is no issue.


src/Microsoft.Data.SqlClient/src/Interop/Windows/Handles/DebugSafeHandle.netcore.cs
```diff
@@ -2,16 +2,13 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#if DEBUG && !NET8_0_OR_GREATER
+#if DEBUG && !NETFRAMEWORK && !NET8_0_OR_GREATER
```
mdaigle, line 5:
```suggestion
#if DEBUG && NET6_0
```

mdaigle: I think it communicates the intent a bit better, but up to you. We'll delete them soon anyway.

src/Microsoft.Data.SqlClient/src/Interop/Windows/Kernel32/Interop.FreeLibrary.netcore.cs
```diff
@@ -2,6 +2,8 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+#if !NETFRAMEWORK
```
mdaigle, line 5:
```suggestion
#if NET
```

src/Microsoft.Data.SqlClient/src/Interop/Windows/Kernel32/Interop.GetProcAddress.netcore.cs
```diff
@@ -2,10 +2,11 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+#if !NETFRAMEWORK
```
mdaigle, line 5:
```suggestion
#if NET
```

src/Microsoft.Data.SqlClient/src/Interop/Windows/Kernel32/Interop.LoadLibraryEx.netcore.cs
```diff
@@ -2,10 +2,11 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+#if !NETFRAMEWORK
```
mdaigle, line 5:
```suggestion
#if NET
```

src/Microsoft.Data.SqlClient/src/Interop/Windows/SChannel/Interop.SecPkgContext_ApplicationProtocol.netcore.cs
```diff
@@ -2,6 +2,8 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+#if !NETFRAMEWORK
```
mdaigle, line 5:
```suggestion
#if NET
```

mdaigle (changes requested):

src/Microsoft.Data.SqlClient/src/Interop/Windows/Handles/DebugSafeHandle.netcore.cs
benrr101: My rationale for using !NET8_0_OR_GREATER is because covers all versions below net8. I know *we* only build net6 and net8, and I guess it would be silly for someone to pull down the source and try to build it for eg, net7, but this is a fairly benign way to cover more bases. 

I think of it similar to how we could do `string.IndexOf("foo") == -1` or we could do `string.IndexOf("foo") < 0`, and how the latter is generally thought to be safer.

src/Microsoft.Data.SqlClient/src/Interop/Windows/Kernel32/Interop.FreeLibrary.netcore.cs
benrr101: Potato potato, I suppose... I thought I had some kind of rationale for it. Maybe that if we ever drop support for netfx, it'll be easier to find instances of NETFRAMEWORK vs NET (since that's used in a lot more?). Eh. I'll do a sweep and make the changes.

mdaigle (approved):

edwardneal:
This looks good to me.

On a more general note, it looks like we're moving away from the various NativeMethods, SafeNativeMethods and UnsafeNativeMethods classes? If so my PR #2828 removes the currently-unused PInvokes from these code files, so there's no need to bother migrating those. There are quite a few other PInvokes in those files which are only used within the [SQLDebugging](https://learn.microsoft.com/en-us/dotnet/api/microsoft.data.sqlclient.sqldebugging?view=sqlclient-dotnet-5.2&viewFallbackFrom=sqlclient-dotnet-core-5.2) class (and associated debugging infrastructure) though. I'd suggest ignoring those too - SQLDebugging is only used when debugging a client application connected to a local SQL 2000 instance, and when this is no longer permitted I'm planning to submit a PR which will eliminate that functionality and remove these PInvokes anyway.

PR #2942, opened 2024-10-31
Update dependency version

cheenamalhotra:
Addresses dependabot alerts


ErikEJ:
@cheenamalhotra  Given that .NET 6 support will be removed, I think it should version 8.0.1

cheenamalhotra:
Discarding with dropping of .NET 6

PR #2944, opened 2024-10-31
Clean up build guide and docs after net6 removal

mdaigle:
Restructures the buildguide document. Adds a reference section and a workflows section. Previously, the doc was structured as a collection of example commands, but it didn't clearly call out dependencies (e.g. building tests before running them and how to do so for that type of test run) and was a poor way to discover and understand the MSBuild targets and parameters.

Also includes small doc and test updates to remove .NET6 references.


cheenamalhotra (approved):

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataStreamTest/DataStreamTest.cs
```diff
@@ -1556,14 +1556,10 @@ private static void ReadStream(string connectionString)
                             DataTestUtility.AssertThrowsWrapper<ArgumentOutOfRangeException>(() => stream.Read(buffer, -1, 2));
                             DataTestUtility.AssertThrowsWrapper<ArgumentOutOfRangeException>(() => stream.Read(buffer, 2, -1));
 
-                            // Prior to net6 comment:ArgumentException is thrown in net5 and earlier. ArgumentOutOfRangeException in net6 and later
```
cheenamalhotra, line 1559: Prior to .NET 6 includes .NET Framework as well right?

mdaigle: I think you're right, I reverted this.

samsharma2700 (approved):

PR #2946, opened 2024-10-31
Add support for .NET 9

mdaigle:
Enhancements:
- Adds .NET 9 to target frameworks

Fixes:
- Adds missing DynamicallyAccessedMembers attribute to `GetProviderSpecificFieldType `and `GetFieldType` in netcore ref project.
  - Removes suppressions due to missing annotations in net6.0
- Switches X509Certificate construction to use X509CertificateLoader.
- Switches PKCS decoding to use SignedCMS.
- Adds internal ReadExactly extension to System.IO.Stream in netfx to provide API parity between netcore and netfx and to address [CA2022](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2022).
- Guards ServicePointManager usage to netfx only, deprecated in .NET 9
  - https://learn.microsoft.com/en-us/dotnet/fundamentals/syslib-diagnostics/syslib0014
  - https://learn.microsoft.com/en-us/dotnet/api/system.net.servicepointmanager?view=net-9.0
 
Package Versions:
- Adds System.Security.Cryptography.Pkcs:9.0.0 to address [SYSLIB0057](https://learn.microsoft.com/en-us/dotnet/fundamentals/syslib-diagnostics/syslib0057)
- Adds Microsoft.Bcl.Cryptography:9.0.0, dependency of the pkcs package
- Bumps Microsoft.Extensions.Caching.Memory to 9.0.0
- Bumps System.Configuration.ConfigurationManager to 9.0.0

Test Dependencies:
- Bumps Microsoft.DotNet.RemoteExecutor to 10.0.0-beta.24564.1
- Bumps Xunit to 2.6.3 for all frameworks


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SqlNormalizer.cs
```diff
@@ -433,7 +445,11 @@ internal override void Normalize(FieldInfo fi, object obj, Stream s)
         internal override void DeNormalize(FieldInfo fi, object recvr, Stream s)
         {
             byte[] b = new byte[4];
+#if NET8_0_OR_GREATER
+            s.ReadExactly(b, 0, b.Length);
```
roji, line 449: Am curious - ReadExactly and Read are very different things... If what you want here is ReadExactly (i.e. fill up b.Length bytes), doesn't that mean that pre-net8.0 has a bug? In other words, either in this point you really do want to loop and read repeatedly until b.Length bytes have been read (ReadExactly behavior), or you don't need that...

roji (commented):
Am curious - is there any net9.0 API that's actually needed/used? If not, is there a specific reason to add the net9.0 TFM?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SqlNormalizer.cs
edwardneal: If #2674 is any indication, this was added to fix the new warning CA2022. I think Read and ReadExactly will work identically in this context though: this is used to deserialize user-defined types from the database. The stream it's reading from will thus either be a MemoryStream (wrapped around the full byte value) or a custom stream which reads an entire field value at a time. In either event, the read should be able to completely populate the buffer.

It'd probably be a good idea to add a comment or some assertions on the return value of `s.Read` in netfx to be clearer here though...

David-Engel:
> Am curious - is there any net9.0 API that's actually needed/used? If not, is there a specific reason to add the net9.0 TFM?

@roji JSON support in SqlDbType: https://github.com/dotnet/runtime/issues/103925

roji:
@David-Engel of course, I forgot about that, thanks :)

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SqlNormalizer.cs
roji: @edwardneal OK. I'm assuming that there are other places in SqlClient where  data is read from the network, and a loop is used to recreate ReadExactly behavior - you could also use that here, just in case (e.g. if somehow this code in the future runs against a stream that may not read fully).

In fact, I'd recommend simply having a ReadExactly extension method which does this ([see the Npgsql code](https://github.com/npgsql/npgsql/blob/main/src/Npgsql/Shims/StreamExtensions.cs)), for older TFMs. This way, you can simply use ReadExactly everywhere regardless of which TFM you're targeting, and don't need #if around your codebase.

eng/pipelines/common/templates/steps/ci-project-build-step.yml
```diff
@@ -34,6 +34,30 @@ parameters:
       - all
 
 steps:
+- task: UseDotNet@2
+  inputs:
+    packageType: sdk
+    version: '8.x'
```
MichelZ, line 40: Did you mean to use 9.x here?

mdaigle: Nope! This is to make sure that the 8.x sdk is installed in the same temp directory on the build agent as the 9.x sdk. This may go away if I can figure out how to make dotnet look in multiple places for sdk versions. It's documented that it should work, but doesn't seem to in this instance. I let it go for the moment while I address some APIs that were marked as obsolete in 9, but I'm getting back to it after that.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
```diff
@@ -180,7 +180,11 @@ static DataTestUtility()
             AliasName = c.AliasName;
             IsJsonSupported = c.IsJsonSupported;
 
+#if NET9_0_OR_GREATER
+//TODO: set protocol here?
```
mdaigle, line 184: This is an outstanding question. If tests pass without setting a default protocol, nothing to do here.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/StreamExtensions.cs
```diff
@@ -0,0 +1,22 @@
+﻿#if NETFRAMEWORK
```
benrr101, line 1: Needs copyright info

benrr101, line 1: Also if this file is only relevant for netfx, it should be named `xyz.netfx.cs`

src/Microsoft.Data.SqlClient/src/Microsoft/Data/StreamExtensions.cs
```diff
@@ -0,0 +1,22 @@
...
+
+namespace System.IO
```
benrr101, line 3: Namespaces should match the location of the file. Also, I generally believe adding code to System namespaces should be avoided since it leads to assumptions that the code is publicly document when it isn't.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/StreamExtensions.cs
```diff
@@ -0,0 +1,22 @@
...
+    internal static class StreamExtensions
+    {
+        internal static void ReadExactly(this Stream stream, byte[] buffer, int offset, int count)
+        {
+            var totalRead = 0;
+            do
+            {
+                var read = stream.Read(buffer, offset + totalRead, count - totalRead);
+                if (read is 0)
+                    throw new EndOfStreamException();
```
benrr101, line 15: Stop it! I don't want to see if statements without braces in new code!

https://www.imperialviolet.org/2014/02/22/applebug.html

src/Microsoft.Data.SqlClient/tests/ManualTests/Extensions/StreamExtensions.cs
```diff
@@ -0,0 +1,22 @@
+﻿#if NETFRAMEWORK
+
+namespace System.IO
+{
+    // Helpers to read/write Span/Memory<byte> to Stream before netstandard 2.1
+    internal static class StreamExtensions
```
benrr101, line 6: This class shouldn't be necessary - the internals of the netfx project should be available to the test projects. If they aren't, we need to fix that rather than duplicate this code.

benrr101 (changes requested):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/VirtualSecureModeEnclaveProviderBase.cs
```diff
@@ -200,7 +204,13 @@ private X509Certificate2Collection GetSigningCertificate(string attestationUrl,
 
                 try
                 {
+#if NET9_0_OR_GREATER
```
MichelZ, line 207: All these conditionals because of cert handling could be removed when System.Security.Cryptography.Pkcs is added to all target frameworks. It seems supported for net462/net8+ and would make it unified

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataStreamTest/DataStreamTest.cs
```diff
@@ -1566,7 +1572,7 @@ private static void ReadStream(string connectionString)
                         }
 
                         // Once Reader is closed
-                        action = (() => stream.Read(buffer, 0, buffer.Length));
+                        action = (() => stream.ReadExactly(buffer, 0, buffer.Length));
```
MichelZ, line 1575: Been having issues with this in my test branch:
https://dev.azure.com/sqlclientdrivers/public/_build/results?buildId=102097&view=logs&j=12f36a2b-3298-5283-ba0d-570258d2371e&t=5c7d99dc-33ce-5a1c-b601-d5438d2e70c8&l=217

mdaigle:
> You should be able to add:
> 
> ```
> - task: UseDotNet@2
>   displayName: 'Use .NET 9.x sdk'
>   inputs:
>     packageType: sdk
>     version: '9.x'
>     installationPath: $(Agent.ToolsDirectory)/dotnet
> 
> - task: UseDotNet@2
>   displayName: 'Install .NET 8.x runtime'
>   inputs:
>     packageType: runtime
>     version: '8.x'
>     installationPath: $(Agent.ToolsDirectory)/dotnet
> ```
> 
> Before this line, as a first step:
> 
> https://github.com/dotnet/SqlClient/blob/8d500a3c63c6f2d57edb67a31f21aa6346581741/eng/pipelines/common/templates/steps/ci-project-build-step.yml#L37
> 
> That should get the build going. You probably need to cache a few more packages in the "official" feed.
> 
> Have a look at [my attempt](https://github.com/dotnet/SqlClient/pull/3030) on getting it to work

Thanks @MichelZ, I'll give that a shot. I had issues previously with the sdk version resolver picking up the .NET 9 rc version when installing that way, but it may be working better now with the GA version available.

src/Microsoft.Data.SqlClient/tests/ManualTests/Extensions/StreamExtensions.cs
mdaigle: How would you do that, typically? Using this attribute: https://learn.microsoft.com/en-us/dotnet/fundamentals/runtime-libraries/system-runtime-compilerservices-internalsvisibletoattribute ?

tools/props/VersionsNet9OrLater.props
edwardneal, line 1: We might need this file in the future, but all three of the package versions listed here are compatible with .NET 8.0 and .NET Standard 2.0. I *think* we could safely change the dependency in Versions.props.

tools/props/Versions.props
```diff
@@ -35,6 +35,7 @@
     <MicrosoftIdentityModelJsonWebTokensVersion>7.5.0</MicrosoftIdentityModelJsonWebTokensVersion>
     <MicrosoftIdentityModelProtocolsOpenIdConnectVersion>7.5.0</MicrosoftIdentityModelProtocolsOpenIdConnectVersion>
     <SystemTextJsonVersion>6.0.10</SystemTextJsonVersion>
```
edwardneal, line 37: This version of System.Text.Json is part of .NET 6.0. Same point with `SystemTextEncodingsWebVersion` on line 29.

eng/pipelines/common/templates/steps/ci-project-build-step.yml
```diff
@@ -78,6 +90,24 @@ steps:
         msbuildArguments: '-t:BuildAKVNetCoreAllOS -p:BuildNumber=${{ parameters.buildNumber }}'
 
 - ${{ if or(eq(parameters.operatingSystem, 'Linux'), eq(parameters.operatingSystem, 'MacOS'), eq(parameters.operatingSystem, 'deferedToRuntime')) }}:
+  - task: UseDotNet@2
```
MichelZ, line 93: These new tasks can be removed

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/VirtualSecureModeEnclaveProvider.cs
```diff
@@ -168,7 +168,8 @@ internal class HealthReport
         public HealthReport(byte[] payload)
         {
             Size = payload.Length;
-            Certificate = new X509Certificate2(payload);
+
```
MichelZ, line 171: What's with the space? :)

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataStreamTest/DataStreamTest.cs
```diff
@@ -1515,8 +1515,8 @@ private static void ReadStream(string connectionString)
                             // Basic case
                             using (stream = reader.GetStream(0))
                             {
-                                stream.Read(smallBuffer, 0, smallBuffer.Length);
-                                stream.Read(buffer, 2, 2);
+                                int _ = stream.Read(smallBuffer, 0, smallBuffer.Length);
```
MichelZ, line 1518: is the type needed here for discard?

tools/props/Versions.props
MichelZ: They are both not referenced in the netcore project.
I think the variable here should be moved up to the "netfx dependencies" section

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionString.cs
```diff
@@ -858,7 +858,7 @@ internal static Dictionary<string, string> GetParseSynonyms()
 #endif // NETFRAMEWORK
                     { SYNONYM.IPADDRESSPREFERENCE, KEY.IPAddressPreference }
                 };
-                Debug.Assert(synonyms.Count == count, "incorrect initial ParseSynonyms size");
+                //Debug.Assert(synonyms.Count == count, "incorrect initial ParseSynonyms size");
```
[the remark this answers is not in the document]
mdaigle: Thanks, didn't mean to commit that.

src/Microsoft.Data.SqlClient/src/Microsoft.Data.SqlClient.csproj
```diff
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 	<PropertyGroup>
 		<TargetFrameworks>net8.0;net462</TargetFrameworks>
```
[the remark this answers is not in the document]
mdaigle: This project isn't compiled directly yet, so it shouldn't complain if it's not there. But I'll add it in to avoid confusion.

tools/props/VersionsNet8OrLater.props
```diff
@@ -2,9 +2,6 @@
 <Project ToolsVersion="Current" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <!-- Net8 project dependencies -->
   <PropertyGroup>
-    <SystemConfigurationConfigurationManagerVersion>8.0.1</SystemConfigurationConfigurationManagerVersion>
-    <MicrosoftExtensionsCachingMemoryVersion>8.0.1</MicrosoftExtensionsCachingMemoryVersion>
-    <MicrosoftDotNetRemoteExecutorVersion>9.0.0-beta.24157.1</MicrosoftDotNetRemoteExecutorVersion>
     <XunitVersion>2.6.3</XunitVersion>
```
cheenamalhotra, line 5: Please move this to versions.props too

src/Microsoft.Data.SqlClient/tests/tools/Microsoft.Data.SqlClient.ExtUtilities/Microsoft.Data.SqlClient.ExtUtilities.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net8.0</TargetFramework>
+    <TargetFrameworks>net8.0;net9.0</TargetFrameworks>
```
cheenamalhotra, line 4: Targeting multiple frameworks is not necessary for projects like these since its an internal project.

tools/props/Versions.props
```diff
@@ -27,20 +27,23 @@
...
+    <SystemSecurityCryptographyPkcsVersion>9.0.0</SystemSecurityCryptographyPkcsVersion>
+    <MicrosoftBclCryptographyVersion>9.0.0</MicrosoftBclCryptographyVersion>
   </PropertyGroup>
   <!-- NetCore project dependencies -->
   <PropertyGroup>
     <MicrosoftDataSqlClientSNIRuntimeVersion>6.0.0-preview1.24226.4</MicrosoftDataSqlClientSNIRuntimeVersion>
-    <SystemConfigurationConfigurationManagerVersion>6.0.1</SystemConfigurationConfigurationManagerVersion>
+    <SystemConfigurationConfigurationManagerVersion>9.0.0</SystemConfigurationConfigurationManagerVersion>
     <MicrosoftSqlServerServerVersion>1.0.0</MicrosoftSqlServerServerVersion>
+    <MicrosoftDotNetRemoteExecutorVersion>10.0.0-beta.24564.1</MicrosoftDotNetRemoteExecutorVersion>
```
cheenamalhotra, line 46: Isn't this a test dependency? Its mentioned twice in this file it seems. Please fix it.

src/Microsoft.Data.SqlClient/tests/ManualTests/Extensions/StreamExtensions.netfx.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+#if NETFRAMEWORK
+
+namespace System.IO
+{
+    // Helpers to read/write Span/Memory<byte> to Stream before netstandard 2.1
+    internal static class StreamExtensions
```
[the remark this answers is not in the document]
cheenamalhotra: Ideally, I would create a Common/Shared project that both test and src projects depend on, but this class is too small for that. If we have more cases like this, we can consider coming up with a common project.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
```diff
@@ -180,7 +180,9 @@ static DataTestUtility()
             AliasName = c.AliasName;
             IsJsonSupported = c.IsJsonSupported;
 
+#if NETFRAMEWORK
```
cheenamalhotra, line 183: Can you elaborate on this change?

mdaigle: ServicePointManager is marked as obsolete in Net 9.
https://learn.microsoft.com/en-us/dotnet/fundamentals/syslib-diagnostics/syslib0014
I removed the call to no effect in the tests, so it didn't seem to be changing anything. But if there's any concern, I can add a suppression instead.

benrr101: @mdaigle Are you sure it's the ServicePointManager? I had seen similar warnings in the code when using Tls12 security type.

src/Microsoft.Data.SqlClient/tests/ManualTests/Extensions/StreamExtensions.netfx.cs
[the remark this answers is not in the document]
benrr101: @mdaigle Like I said in the previous PR, internals of the MDS projects should be visible to the test projects. If they aren't, then we should make them so.

@cheenamalhotra I personally don't think we need to add another project for this purpose, if we leverage internalsvisibleto. We can discuss more if we need to, tho.

benrr101 (approved):
Looks good to me in general. I think we should make InternalsVisibleTo test projects (I think that change was only made in the clientx branch) so we can avoid duplication of the stream extensions. But I won't hold up approval on that.

cheenamalhotra (approved):

PR #2948, opened 2024-11-01
Replace public async void with public async Task for XUnit

### LINKED ISSUE #2947: XUnit does not support public async void in next version
The next version of XUnit does not support `public async void` in test methods.
This should be changed to `public async Task`

https://dotnet.social/@xunit/112265523775265162

More recent versions of XUnit 2.x also issue a Warning - which gets treated as an Error in this project.

MichelZ:
Fixes #2947

### DIFF
--- a/src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlConfigurableRetryLogicTest.cs
+++ b/src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlConfigurableRetryLogicTest.cs
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
+using System.Threading.Tasks;
 using Xunit;
 
 namespace Microsoft.Data.SqlClient.Tests
@@ -10,7 +11,7 @@ namespace Microsoft.Data.SqlClient.Tests
     public class SqlConfigurableRetryLogicTest
     {
         [Fact]
-        public async void InvalidExecute()
+        public async Task InvalidExecute()
         {
             SqlRetryLogicOption option = new SqlRetryLogicOption()
             {
--- a/src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlConnectionBasicTests.cs
+++ b/src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlConnectionBasicTests.cs
@@ -342,7 +342,7 @@ namespace Microsoft.Data.SqlClient.Tests
         [InlineData(10)]
         [InlineData(5)]
         [InlineData(1)]
-        public async void ConnectionTimeoutTestAsync(int timeout)
+        public async Task ConnectionTimeoutTestAsync(int timeout)
         {
             // Start a server with connection timeout from the inline data.
             using TestTdsServer server = TestTdsServer.StartTestServer(false, false, timeout);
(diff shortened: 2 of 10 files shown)


MichelZ:
@dotnet-policy-service agree

mdaigle (approved):

benrr101 (approved):
Thanks for both reporting the issue and fixing the issue :D I think the likelihood that we'll upgrade xunit versions anytime soon is sorta slim - iirc there's a dependency issue with a library provided by the dotnet team, that won't work with a newer version. It's been like 7 months since I looked at that, so it might not be a concern anymore. But, regardless, this is the proper way to do async tests, so we should be doing this.

I'll kick off a CI build, and assuming it passes without issue (no reason it shouldn't), we'll get it merged 🚢

edwardneal (commented):
Thanks for this - I didn't realise we were using `async void` anywhere.

I can see a bundle of other async void tests when I search for it in the solution. I can submit a PR to your branch with them if you want, unless they were excluded deliberately?

MichelZ:
Hmm. I upgraded XUnit and it gave Warnings about these. I‘m not sure why it wouldn‘t warn about the others, but of course I/we can change them as well

MichelZ:
I think the reason the others are not flagged is that they don't use the [Fact] or [Theory] attributes, but the [ConditionalTheory]. The analyzer probably ignores those. I'll update them as well

MichelZ:
Done.

There are also async void's in the BenchmarkRunners for Benchmark.NET, I'm not too familiar with Benchmark.NET, so this might be on purpose? If not, I can also fix them in a separate PR.

Additionally, there is one async void in SNIPacket.cs here:
https://github.com/dotnet/SqlClient/blob/9d5ca32e666b3da1cc27b4ab145367308998b147/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIPacket.cs#L330

Not sure if that needs fixing, too (in another PR)

MichelZ:
> Thanks for both reporting the issue and fixing the issue :D I think the likelihood that we'll upgrade xunit versions anytime soon is sorta slim - iirc there's a dependency issue with a library provided by the dotnet team, that won't work with a newer version. It's been like 7 months since I looked at that, so it might not be a concern anymore. But, regardless, this is the proper way to do async tests, so we should be doing this.
> 
> I'll kick off a CI build, and assuming it passes without issue (no reason it shouldn't), we'll get it merged 🚢

You mean this comment?
https://github.com/dotnet/SqlClient/pull/2379#issuecomment-1979544650

Wasn't this fixed by upgrading to .NET8 now?

MichelZ:
Let's see how TdsParser goes... 
I'm curious about SqlClientX and what the roadmap/timeframe looks for that

benrr101:
> Do you think I can add a PR to upgrade to the latest versions?

Yknow what, I think you're right. Since we dropped net6 we should be able to take the latest version of Microsoft.DotNet.XUnitExtensions and probably take the latest of XUnit (might even be a requirement for the aforementioned). I'm always a bit hesitant with this codebase, but once this one is checked in, I would support a PR that upgrades XUnit/extensions

> I'm curious about SqlClientX and what the roadmap/timeframe looks for that
... same 😅

PR #2951, opened 2024-11-02
Fix SystemMemoryVersion variable missing

### LINKED ISSUE #2950: Project dependency System.Memory does not contain in inclusive lower bound
My Visual Studio complains about this:

Error (active)	NU1604	Warning As Error: Project dependency System.Memory does not contain an inclusive lower bound. Include a lower bound in the dependency version to ensure consistent restore results.	Microsoft.Data.SqlClient (net462), Microsoft.Data.SqlClient (net6.0), Microsoft.Data.SqlClient (net8.0)	

And sure enough, here is there reference to System.Memory:
https://github.com/dotnet/SqlClient/blob/9d5ca32e666b3da1cc27b4ab145367308998b147/src/Microsoft.Data.SqlClient/src/Microsoft.Data.SqlClient.csproj#L12

But there is no corresponding version variable in Versions.props:
https://github.com/dotnet/SqlClient/blob/main/tools/props/Versions.props

I'm not 100% sure if the reference is needed or not, but a reference without a version is just wrong.

MichelZ:
Fixes #2950

### DIFF
--- a/tools/props/Versions.props
+++ b/tools/props/Versions.props
@@ -25,6 +25,7 @@
   <PropertyGroup>
     <MicrosoftDataSqlClientSniVersion>6.0.0-preview1.24226.4</MicrosoftDataSqlClientSniVersion>
     <SystemBuffersVersion>4.5.1</SystemBuffersVersion>
+    <SystemMemoryVersion>4.5.5</SystemMemoryVersion>
     <SystemTextEncodingsWebVersion>6.0.0</SystemTextEncodingsWebVersion>
   </PropertyGroup>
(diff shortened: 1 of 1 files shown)


benrr101 (approved):
I don't see any issues with fixing the issue this way, thanks for tackling it 👍 

For others reviewing this, this change doesn't have any bearing on today's release builds. This fixes an error that shows up during nuget restore of the common project.

MichelZ:
It was giving me errors in VS, I hate that :)

cheenamalhotra (approved):

samsharma2700 (approved):

Issue #2953, opened 2024-11-02
Work towards merging TdsParser


MichelZ:
This is for tracking purposes connected to #1261 
I would like to merge TdsParser to a single code base netfx/netcore

The approach I'd like to take is this:
- Do small PR's first which only change code style to align netfx/netcore codebases
- Do small PR's that target specific things (e.g. change from Types to type aliases e.g. Int32 to int)
- Do more PR's that align code but do need code changes

This way (I hope?) changes are small-ish and/or low-riskish, so that they can be reviewed fairly fast (much easier than "big bang" PR's IMHO)

Hope that makes sense, if not just tell me please :)

edwardneal:
Thanks for this @MichelZ - these bundles of changes look good to me.

Another pattern of changes will also need to be made at some point, across TdsParser and other core parts of SqlClient: Constrained Execution Region support exists in netfx, but was removed in netcore. This is (roughly) a no-op, but each instance adds ~45 lines to the diff and it's usually inside methods which are otherwise identical. It's probably not necessary as part of this cleanup, but it'd be a good idea to port the CER support to netcore before starting to merge identical methods.

On a wider point: I think it makes sense to use the TdsParser approach as a template for the core parts of SqlClient.

benrr101:
Thanks for taking a stab at merging this behemoth of a class 👍 I'll do my best to look at these PRs in a timely manner, but keep in mind we need two signoffs on them to get them merged. I'll also do my best to encourage my teammates to look at your PRs as well 😅 

That being said, I think for the smaller changes, it might be more beneficial to combine a couple together - eg, comments and comment typos could be rolled into a single comment PR. I've been trying to structure my PRs such that each commit is a complete item. So, eg, if you had a PR for comments, one commit could be getting the blocks aligned, the second could be to fix the typos. You don't need to go back recall the PRs that are currently out, but it might be beneficial going forward. It's a delicate balance between too much overhead per PR and too much to review in a single PR.

My strategy for merging classes has been to create a "merge file" in the commit history of the PR. Basically indicate what blocks of code are from only one side of the merge. Then, merge work through each chunk of conflicts to either resolve as one or the other (eg, Int32 vs int) or leave with an `#if NETFRAMEWORK` sort of thing.

edwardneal:
At the moment, TdsParser has a diff of -550/+430. The tasks I'd identified to finish this merge are:

#### Trivial ####
* Sync tracing - netcore takes precedence
* Sync exception messages
[...]

PR #2957, opened 2024-11-02
Merge | Align Type aliases between netfx and netcore

MichelZ:
This PR solely aligns type aliases between netfx and netcore versions so they can be merged easier later.

Part of #2953


benrr101 (approved):
Looks good, but going forward, I think we can combine a couple of PRs like this together. I'm starting to get fatigue from reviewing these all independently 😅

MichelZ:
> Looks good, but going forward, I think we can combine a couple of PRs like this together. I'm starting to get fatigue from reviewing these all independently 😅

Apologies, I really thought it makes it easier for review to have smaller, more targeted PR's... Guess I was wrong :)
Will combine them in the future

benrr101:
There's definitely a balance :) I think we'll still take stuff when it's bite-sized like this, but it just might be slow with all the overhead. On the other hand, gigantic PRs are a lot more likely to just be outright rejected

mdaigle (approved):
Looks good after merge conflicts are fixed

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -10384,7 +10384,7 @@ internal Task TdsExecuteRPC(SqlCommand cmd, IList<_SqlRPC> rpcArray, int timeout
                                         else
                                         {
                                             Debug.Assert(value is SqlChars, "Unknown value for Ansi datatype");
-                                            s = new String(((SqlChars)value).Value);
+                                            s = new string(((SqlChars)value).Value)
```
cheenamalhotra, line 10387:
```suggestion
                                            s = new string(((SqlChars)value).Value);
```

cheenamalhotra (changes requested):

MichelZ:
AZP needs a smack on the head

David-Engel (approved):

MichelZ:
@cheenamalhotra Anything I can do to move this along? Thx

PR #2960, opened 2024-11-02
Remove redundant Namespace to align netfx/netcore

MichelZ:
An easy to review change that removes redundant namespace definitions to align netfx/netcore

Part of #2953


benrr101 (approved):
I'm surprised there aren't more instances in this file...

cheenamalhotra (approved):

PR #2962, opened 2024-11-02
Mark InternalException, NetEventSource & SecPkgContext_ApplicationProtocol as .NET 6.0-only

edwardneal:
This picks up from the comments on #2828, to resolve the merge race between @benrr101, @mdaigle and I.

InternalException, NetEventSource and the types associated with SecPkgContext_ApplicationProtocol are only used by the old .NET 6.0 target. This PR guards these behind `#if NET && !NET8_0_OR_GREATER`, ready to join the rest of the .NET 6.0-only files and types when they're removed.

Is the removal of the .NET 6.0-only code already in motion, or will this happen after the SqlClient 6.0 release?

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Common/System/Net/InternalException.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Common/System/Net/InternalException.cs
@@ -2,6 +2,8 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+#if NET && !NET8_0_OR_GREATER
+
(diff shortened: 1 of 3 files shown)


benrr101 (approved):
Looks good to me 🚢 
While this touches a file I've got queued up in a later PR, it won't cause many issues. I can just update that PR to delete the file with prejudice 😈

cheenamalhotra:
> Is the removal of the .NET 6.0-only code already in motion, or will this happen after the SqlClient 6.0 release?

We'll take in changes as much as we can get accomplished.

cheenamalhotra (approved):

Issue #2965, opened 2024-11-02
Work towards merging SqlDataReader


MichelZ:
Same as #2953 but for SqlDataReader

- Comments #2966 
- Port IDBColumnSchemaGenerator to netf #2967 
- #2969

edwardneal:
Besides the coding style/comments/CERs, there's other work:
* #2964 
* Port #499 to netfx
* Consider implementing IDbColumnSchemaGenerator. I've noticed this absence and noticed that the interface exists in .NET Standard 2.0 and .NET Framework 4.7.1. We're building for .NET Framework 4.6.2, and I've not checked how the dependencies will shake out here

edwardneal:
After merging #2967, the diff between netfx and netcore's SqlDataReader implementation is now -407/+159. Most of this is now taken up by two points:
* Porting of #499
* Reinstate the exception handling for OutOfMemoryException, StackOverflowException and ThreadAbortException. None of these have any effect on netcore

#3042 is a notable style change though. Once that's merged, I can port the exception handling across as time permits (unless someone else wants to pick this up sooner.) I'll submit a port of PR 499 (and any remaining dependencies/post-PR changes.)

The rest of the differences are just differences in logging text and debug assertions, along with ~10 lines of behavioural changes in TrySetMetaData and IsDBNull.

PR #2966, opened 2024-11-02
Align comments between netfx and netcore codebases

MichelZ:
Hopefully an easy to review PR that deals with differences in comments and some whitespace between netfx and netcore for easier future merging

Part of #2965

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
@@ -5,6 +5,7 @@
 using System;
 using System.Collections;
 using System.Collections.ObjectModel;
+using System.ComponentModel;
 using System.Data;
 using System.Data.Common;
 using System.Data.SqlTypes;
@@ -917,7 +918,7 @@ namespace Microsoft.Data.SqlClient
                         {
                             if (_stateObj != null)
                             {  // reader not closed while we waited for the lock
-                               // TryCloseInternal will clear out the snapshot when it is done
+                                // TryCloseInternal will clear out the snapshot when it is done
                                 if (_snapshot != null)
                                 {
 #if DEBUG
@@ -2045,6 +2046,7 @@ namespace Microsoft.Data.SqlClient
         }
 
         /// <include file='../../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlDataReader.xml' path='docs/members[@name="SqlDataReader"]/GetChar/*' />
+        [EditorBrowsableAttribute(EditorBrowsableState.Never)]
         override public char GetChar(int i)
         {
             throw ADP.NotSupported();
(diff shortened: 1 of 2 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
```diff
@@ -2045,6 +2046,7 @@ override public TextReader GetTextReader(int i)
         }
 
         /// <include file='../../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlDataReader.xml' path='docs/members[@name="SqlDataReader"]/GetChar/*' />
+        [EditorBrowsableAttribute(EditorBrowsableState.Never)]
```
benrr101, line 2049: I'm not sure that this should be moved across to netcore.

benrr101 (changes requested):
Please rollback the change for `EditorBrowsableAttribute` if we don't have a strong reason to move it over.

Also, in the future, please just combine all the comment cleanup changes into a single PR.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
MichelZ: For alignment purpuses, is it OK to remove it from netfx then?

edwardneal: I've just checked the ref projects; this is present in both [netfx](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/netfx/ref/Microsoft.Data.SqlClient.cs#L1294) and [netcore](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs#L1282). I personally think it's more likely that someone missed a spot when adding the attribute before the source code was uploaded. In light of the ref projects, perhaps we should add the EditorBrowsableAttribute to netcore?

MichelZ: > I think this is the sort of thing that for merging purposes we should leave alone. In the merged version, wrap it in a `#if NETFRAMEWORK`. While it's not the cleanest solution, we have precedent for doing this in other merges, and it is the safest change.

What about a combination, wrapping it in a #IF NETFRAMEWORK in both versions?

benrr101: I think if it's present in both ref projects, we should probably have it in both netfx and netcore implementation projects. I officially rescind my comment :)

benrr101 (approved):

mdaigle (approved):

PR #2967, opened 2024-11-02
Merge | Add IDBColumnSchemaGenerator interface to netfx SqlDataReader

MichelZ:
Bring IDBColumnSchemaGenerator to netfx for later code base merging

I made sure to enable the respective test for netfx

Part of #2965


MichelZ:
@edwardneal Would you mind running the pipeline for me on this one? :)

edwardneal:
Thanks for this MichelZ. The changes look good to me; would you mind feeding the extra package reference through to the [nuspec file](https://github.com/dotnet/SqlClient/blob/main/tools/specs/Microsoft.Data.SqlClient.nuspec) and the .NET Framework [reference csproj](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/netfx/ref/Microsoft.Data.SqlClient.csproj) please?

I don't have access to the pipelines, but hopefully the SqlClient team will be able to look at it in a few days. Something's definitely odd there - your PRs didn't run the CI builds, and my commit ran the CI build but encountered a lot more timeouts than normal in the tests.

MichelZ:
Will do. I'm not a contributor, that's probably why the pipelines don't run for me (yet)

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserHelperClasses.cs
```diff
@@ -327,11 +325,9 @@ private _SqlMetaDataSet(_SqlMetaDataSet original)
             id = original.id;
             _hiddenColumnCount = original._hiddenColumnCount;
             _visibleColumnMap = original._visibleColumnMap;
-#if !NETFRAMEWORK
             dbColumnSchema = original.dbColumnSchema;
-#else
             schemaTable = original.schemaTable;
```
MichelZ, line 329: Not 100% sure about this one. Was the intention here to exclude schemaTable for netcore?

MichelZ:
Thanks for trying :)

benrr101:
@ErikEJ @edwardneal @MichelZ We have changed security rules recently such that only contributors can kick off pipeline runs. This is due to the potential for contributors to run code in PRs that could be hazardous to our build agents or cause a DoS.

MichelZ:
This package needs to be added to the sqlclientdriver nuget feed for this build to succeed:
System.Data.Common: 4.3.0

benrr101:
@MichelZ I might've mentioned it before but yep, we've got security on the internal nuget feed such that only contributors can pull upstream packages from the feed. I've gone ahead and added System.Data.Common 4.3.0 to the feed, so it should be good to go.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserHelperClasses.cs
benrr101: Yeah, this kind of code doesn't make a ton of sense to me (the original, that is). It'd be nice if it came with a comment that explains what's the rationale for whats going on but ... whatever. 

As for your change, I'm also not 100% about it either. It basically says we assign the dbColumnSchema in netcore, and assign the schemaTable in netfx. So, I wonder if assigning both in both cases is the right behavior...

src/Microsoft.Data.SqlClient/src/Microsoft.Data.SqlClient.csproj
```diff
@@ -10,6 +10,7 @@
 	<ItemGroup>
 	  <PackageReference Include="System.Buffers" Version="$(SystemBuffersVersion)" />
 	  <PackageReference Include="System.Memory" Version="$(SystemMemoryVersion)" />
+    <PackageReference Include="System.Data.Common" Version="$(SystemDataCommonVersion)" Condition="'$(TargetFramework)' == 'net462'" />
```
benrr101, line 13: Is this package not compatible with netcore? I know a lot of packages are still compatible with netcore but just don't do anything in scenarios where they're not needed (eg, System.Memory being part of the netcore).

src/Microsoft.Data.SqlClient/netfx/ref/Microsoft.Data.SqlClient.csproj
```diff
@@ -27,5 +27,6 @@
   <ItemGroup>
     <PackageReference Include="Azure.Identity" Version="$(AzureIdentityVersion)" />
     <PackageReference Include="System.Text.Json" Version="$(SystemTextJsonVersion)" />
+    <PackageReference Include="System.Data.Common" Version="$(SystemDataCommonVersion)" />
```
benrr101, line 30: I'm kinda confused about this one - it seems like we've needed to build against System.Data.Common for a long time. So I'm concerned why we're only just now adding it. @cheenamalhotra do you know the history behind how we relate to System.Data.Common?

src/Microsoft.Data.SqlClient/src/Microsoft.Data.SqlClient.csproj
MichelZ: It's a netstandard1.2 package, there are no explicit .NET Core versions of it, so I thought better not mess with the .NET Core side

src/Microsoft.Data.SqlClient/netfx/ref/Microsoft.Data.SqlClient.csproj
edwardneal: It [looks like](https://apisof.net/catalog/5bfef1680a062fbe8a07a16f198d6019?fx=net462) System.Data.Common is packaged in-box with .NET Framework 4.7.1 and upwards, but since we're targeting .NET Framework 4.6.2 we need the package reference. .NET Core doesn't have this problem - all the versions we target package it in-box.

cheenamalhotra: I think this should work and shouldn't be a problem, if we also make sure this dependency is added to nuspec. Being OOB allows us to reference types available in NuGet packages for target frameworks where in-box types are missing, like this one.

cheenamalhotra (approved):
LGTM overall, please resolve conflicts when possible.

MichelZ:
Done. I'd need another /azp run  I guess :)

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserHelperClasses.cs
MichelZ: I think I'm leaving this in for now

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/UdtTest/SqlServerTypesTest.cs
```diff
@@ -252,7 +252,6 @@ private static void TestUdtSqlDataReaderGetStream(CommandBehavior behavior)
         }
 
         // Synapse: Parse error at line: 1, column: 41: Incorrect syntax near 'hierarchyid'.
-        [ActiveIssue("25421", TargetFrameworkMonikers.NetFramework)]
         [ConditionalFact(typeof(DataTestUtility), nameof(DataTestUtility.AreConnStringsSetup), nameof(DataTestUtility.IsNotAzureSynapse))]
         public static void TestUdtSchemaMetadata()
```
edwardneal, line 256: Would you also mind removing `using System.Data.Common;" from this file please? There's also an extension method named `GetColumnSchema`, and I just want to sanity-check that we're testing `SqlDataReader.GetColumnSchema` rather than that.

MichelZ: Done

MichelZ:
I'm currently unsure what causes this issue....
```

"G:\MIZE\DEV\GitHub\SqlClient\build.proj" (BuildTestsNetFx target) (1) ->
"G:\MIZE\DEV\GitHub\SqlClient\src\Microsoft.Data.SqlClient\tests\ManualTests\Microsoft.Data.SqlClient.ManualTesting.Tests.csproj" (default target) (11:11) ->
"G:\MIZE\DEV\GitHub\SqlClient\src\Microsoft.Data.SqlClient\tests\ManualTests\Microsoft.Data.SqlClient.ManualTesting.Tests.csproj" (Build target) (11:12) ->
(CoreCompile target) ->
  G:\MIZE\DEV\GitHub\SqlClient\src\Microsoft.Data.SqlClient\tests\ManualTests\SQL\UdtTest\SqlServerTypesTest.cs(264,86): error CS1061: 'SqlDataReader' does not contain a definition for 'GetColumnSchema' and no accessib
le extension method 'GetColumnSchema' accepting a first argument of type 'SqlDataReader' could be found (are you missing a using directive or an assembly reference?) [G:\MIZE\DEV\GitHub\SqlClient\src\Microsoft.Data.Sql
Client\tests\ManualTests\Microsoft.Data.SqlClient.ManualTesting.Tests.csproj::TargetFramework=net462]
```

The method is clearly available

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlDataReader.cs
```diff
@@ -27,7 +28,7 @@
 namespace Microsoft.Data.SqlClient
 {
     /// <include file='../../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlDataReader.xml' path='docs/members[@name="SqlDataReader"]/SqlDataReader/*' />
-    public class SqlDataReader : DbDataReader, IDataReader
+    public class SqlDataReader : DbDataReader, IDataReader, IDbColumnSchemaGenerator
```
edwardneal, line 31: This change (as well as the additional method) to the public API surface needs to be replicated into the netfx ref project. Once you've done this, I think CI will compile. Visual Studio compiles using a project reference, CI might use a package reference (which introduces the ref project as a layer of indirection.)

MichelZ: Thank you!

edwardneal: I haven't seen it compile on the NuGet stage, but the other netfx targets seem to compile their tests properly. Good catch on the netcore ref project though, I didn't realise that this was missing from there too.

benrr101 (approved):

mdaigle (approved):

cheenamalhotra:
@MichelZ could you please resolve conflicts? Thx.

MichelZ:
@cheenamalhotra Done! :)

PR #2968, opened 2024-11-03
Reduce automated test crashes

edwardneal:
This is an attempt to stabilise the automated tests slightly. They sometimes crash and need to be restarted; I've recently been able to reproduce the same behaviour in my local testing.

Several tests instantiate a number of threads, and these threads don't handle exceptions (such as failed test assertions.) If these assertions fail, the testhost exits.

I've converted these threads to tasks, specifying TaskCreationOptions.LongRunning. This will force the task scheduler to give the task its own thread, and it'll mean that the exceptions are propagated back to the caller. The test will thus fail, but the testhost will stay running.

This should hopefully mean that ManualTests takes about 15-20 minutes to run, rather than 30-40 minutes.

### DIFF
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/ApiShould.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/ApiShould.cs
@@ -1983,29 +1983,34 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests.AlwaysEncrypted
 
                     CommandHelper.s_sleepDuringTryFetchInputParameterEncryptionInfo?.SetValue(null, true);
 
-                    Thread[] threads = new Thread[2];
+                    Task[] tasks = new Task[2];
 
                     // Invoke ExecuteReader or ExecuteNonQuery in another thread.
+                    // Use long-running tasks to create the thread. This enables any failed assertions to propagate, rather than
+                    // allowing the exception to kill the thread and the process.
                     if (executeMethod == @"ExecuteReader")
                     {
-                        threads[0] = new Thread(new ParameterizedThreadStart(Thread_ExecuteReader));
+                        tasks[0] = new Task(Thread_ExecuteReader,
+                            new TestCommandCancelParams(sqlCommand, _tableName, numberOfCancelCalls),
+                            TaskCreationOptions.LongRunning);
                     }
                     else
                     {
-                        threads[0] = new Thread(new ParameterizedThreadStart(Thread_ExecuteNonQuery));
+                        tasks[0] = new Task(Thread_ExecuteNonQuery,
+                            new TestCommandCancelParams(sqlCommand, _tableName, numberOfCancelCalls),
+                            TaskCreationOptions.LongRunning);
                     }
-
-                    threads[1] = new Thread(new ParameterizedThreadStart(Thread_Cancel));
+                    tasks[1] = new Task(Thread_Cancel, TaskCreationOptions.LongRunning);
 
                     // Start the execute thread.
-                    threads[0].Start(new TestCommandCancelParams(sqlCommand, _tableName, numberOfCancelCalls));
+                    tasks[0].Start();
 
                     // Start the thread which cancels the above command started by the execute thread.
-                    threads[1].Start(new TestCommandCancelParams(sqlCommand, _tableName, numberOfCancelCalls));
+                    tasks[1].Start();
 
                     // Wait for the threads to finish.
-                    threads[0].Join();
-                    threads[1].Join();
+                    tasks[0].Wait();
+                    tasks[1].Wait();
 
                     CommandHelper.s_sleepDuringTryFetchInputParameterEncryptionInfo?.SetValue(null, false);
 
@@ -2033,26 +2038,30 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests.AlwaysEncrypted
                     CommandHelper.s_sleepDuringRunExecuteReaderTdsForSpDescribeParameterEncryption?.SetValue(null, true);
 
                     // Invoke ExecuteReader or ExecuteNonQuery in another thread.
-                    threads = new Thread[2];
+                    tasks = new Task[2];
                     if (executeMethod == @"ExecuteReader")
                     {
-                        threads[0] = new Thread(new ParameterizedThreadStart(Thread_ExecuteReader));
+                        tasks[0] = new Task(Thread_ExecuteReader,
+                            new TestCommandCancelParams(sqlCommand, _tableName, numberOfCancelCalls),
+                            TaskCreationOptions.LongRunning);
                     }
                     else
                     {
-                        threads[0] = new Thread(new ParameterizedThreadStart(Thread_ExecuteNonQuery));
+                        tasks[0] = new Task(Thread_ExecuteNonQuery,
+                            new TestCommandCancelParams(sqlCommand, _tableName, numberOfCancelCalls),
+                            TaskCreationOptions.LongRunning);
                     }
-                    threads[1] = new Thread(new ParameterizedThreadStart(Thread_Cancel));
+                    tasks[1] = new Task(Thread_Cancel, TaskCreationOptions.LongRunning);
 
                     // Start the execute thread.
-                    threads[0].Start(new TestCommandCancelParams(sqlCommand, _tableName, numberOfCancelCalls));
+                    tasks[0].Start();
 
                     // Start the thread which cancels the above command started by the execute thread.
-                    threads[1].Start(new TestCommandCancelParams(sqlCommand, _tableName, numberOfCancelCalls));
+                    tasks[1].Start();
 
                     // Wait for the threads to finish.
-                    threads[0].Join();
-                    threads[1].Join();
+                    tasks[0].Wait();
+                    tasks[1].Wait();
 
                     CommandHelper.s_sleepDuringRunExecuteReaderTdsForSpDescribeParameterEncryption?.SetValue(null, false);
 
@@ -2079,26 +2088,30 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests.AlwaysEncrypted
 
                     CommandHelper.s_sleepAfterReadDescribeEncryptionParameterResults?.SetValue(null, true);
 
-                    threads = new Thread[2];
+                    tasks = new Task[2];
                     if (executeMethod == @"ExecuteReader")
                     {
-                        threads[0] = new Thread(new ParameterizedThreadStart(Thread_ExecuteReader));
+                        tasks[0] = new Task(Thread_ExecuteReader,
+                            new TestCommandCancelParams(sqlCommand, _tableName, numberOfCancelCalls),
+                            TaskCreationOptions.LongRunning);
                     }
                     else
                     {
-                        threads[0] = new Thread(new ParameterizedThreadStart(Thread_ExecuteNonQuery));
+                        tasks[0] = new Task(Thread_ExecuteNonQuery,
+                            new TestCommandCancelParams(sqlCommand, _tableName, numberOfCancelCalls),
+                            TaskCreationOptions.LongRunning);
                     }
-                    threads[1] = new Thread(new ParameterizedThreadStart(Thread_Cancel));
+                    tasks[1] = new Task(Thread_Cancel, TaskCreationOptions.LongRunning);
 
                     // Start the execute thread.
-                    threads[0].Start(new TestCommandCancelParams(sqlCommand, _tableName, numberOfCancelCalls));
+                    tasks[0].Start();
 
                     // Start the thread which cancels the above command started by the execute thread.
-                    threads[1].Start(new TestCommandCancelParams(sqlCommand, _tableName, numberOfCancelCalls));
+                    tasks[1].Start();
 
                     // Wait for the threads to finish.
-                    threads[0].Join();
-                    threads[1].Join();
+                    tasks[0].Wait();
+                    tasks[1].Wait();
 
                     CommandHelper.s_sleepAfterReadDescribeEncryptionParameterResults?.SetValue(null, false);
 
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/MirroringTest/ConnectionOnMirroringTest.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/MirroringTest/ConnectionOnMirroringTest.cs
@@ -6,6 +6,7 @@ using System;
 using System.Collections.Generic;
 using System.Data;
 using System.Threading;
+using System.Threading.Tasks;
 using Xunit;
 
 namespace Microsoft.Data.SqlClient.ManualTesting.Tests
@@ -28,17 +29,14 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests
                 builder.ConnectTimeout = 0;
 
                 TestWorker worker = new TestWorker(builder.ConnectionString);
-                Thread childThread = new Thread(() => worker.TestMultipleConnection());
-                childThread.Start();
+                Task childTask = Task.Factory.StartNew(() => worker.TestMultipleConnection(), TaskCreationOptions.LongRunning);
 
                 if (workerCompletedEvent.WaitOne(10000))
                 {
-                    childThread.Join();
+                    childTask.Wait();
                 }
                 else
                 {
-                    // currently Thread.Abort() throws PlatformNotSupportedException in CoreFx.
-                    childThread.Interrupt();
                     throw new Exception("SqlConnection could not open and close successfully in timely manner. Possibly connection hangs.");
                 }
             }
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ParameterTest/ParametersTest.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ParameterTest/ParametersTest.cs
@@ -9,6 +9,8 @@ using System.Data;
 using System.Data.SqlTypes;
 using System.Threading;
 using Xunit;
+using System.Threading.Tasks;
+
 #if NET6_0_OR_GREATER
 using Microsoft.SqlServer.Types;
 using Microsoft.Data.SqlClient.Server;
@@ -886,10 +888,10 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests
         [ConditionalFact(typeof(DataTestUtility), nameof(DataTestUtility.AreConnStringsSetup))]
         public static void ClosedConnection_SqlParameterValueTest()
         {
-            var threads = new List<Thread>();
+            var tasks = new List<Task>();
             for (int i = 0; i < 100; i++)
             {
-                var t = new Thread(() =>
+                var t = Task.Factory.StartNew(() =>
                 {
                     for (int j = 0; j < 1000; j++)
                     {
@@ -902,13 +904,12 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests
                             Assert.Fail($"Unexpected exception occurred: {e.Message}");
                         }
                     }
-                });
-                t.Start();
-                threads.Add(t);
+                }, TaskCreationOptions.LongRunning);
+                tasks.Add(t);
             }
-            for (int i = 0; i < threads.Count; i++)
+            for (int i = 0; i < tasks.Count; i++)
             {
-                threads[i].Join();
+                tasks[i].Wait();
             }
         }
 
@@ -927,7 +928,7 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests
             cm.Parameters.Add(new SqlParameter("@id2", SqlDbType.UniqueIdentifier) { Direction = ParameterDirection.Output });
             try
             {
-                System.Threading.Tasks.Task<int> task = cm.ExecuteNonQueryAsync(cancellationToken.Token);
+                Task<int> task = cm.ExecuteNonQueryAsync(cancellationToken.Token);
                 task.Wait();
             }
             catch (Exception)
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectivityTests/ConnectivityTest.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectivityTests/ConnectivityTest.cs
@@ -7,6 +7,7 @@ using System.Collections.Generic;
 using System.Data;
 using System.Diagnostics;
 using System.Threading;
+using System.Threading.Tasks;
 using Xunit;
 
 namespace Microsoft.Data.SqlClient.ManualTesting.Tests
@@ -170,7 +171,7 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests
             private static List<ConnectionWorker> s_workerList = new();
             private ManualResetEventSlim _doneEvent = new(false);
             private double _timeElapsed;
-            private Thread _thread;
+            private Task _task;
             private string _connectionString;
             private int _numOfTry;
 
@@ -179,7 +180,7 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests
                 s_workerList.Add(this);
                 _connectionString = connectionString;
                 _numOfTry = numOfTry;
-                _thread = new Thread(new ThreadStart(SqlConnectionOpen));
+                _task = new Task(SqlConnectionOpen, TaskCreationOptions.LongRunning);
             }
 
             public static List<ConnectionWorker> WorkerList => s_workerList;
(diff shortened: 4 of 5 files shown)


edwardneal:
While not strictly related to this, two comments on the `ApiShould.TestSqlCommandCancel` test: this is always going to be a somewhat flaky test.

The flaky part of the test methodology is as follows:
* Create a SqlCommand
* Start Thread A, which calls SqlCommand.ExecuteNonQuery/.ExecuteReader on it. Expect it to return an exception indicating that it's been cancelled
* Start Thread B, which calls SqlCommand.Cancel

Besides using threads directly, the current code enables Thread A to run immediately, while Thread B sleeps for 0.5s. This introduces a race condition where sometimes the command could have finished executing before the cancellation runs. In such a case, the original Assert.Throws call would fail (because ExecuteNonQuery/ExecuteReader completes without throwing an InvalidOperationException.) There was also an edge case where ExecuteNonQuery/ExecuteReader might throw a SqlException rather than an InvalidOperationException. In both of these cases, the exceptions went unhandled, the thread crashed and the testhost restarted.

I've tried to rework this coordination slightly with an interlock and a tweak to the approach:
1. Thread A now waits for a ManualResetEvent controlled by Thread B. Thread B will only set this MRE directly before calling SqlCommand.Cancel
2. Thread B now hammers calls to SqlCommand.Cancel until it receives a signal on an MRE controlled by Thread A. Thread A sets this signal when the ExecuteNonQuery/ExecuteReader call has finished

This reduces the race condition, but doesn't eliminate it: calling ExecuteNonQuery and ExecuteReader implicitly resets the SqlCommand's cancellation status, so it's possible that thread B might call Cancel, then thread A might reset this. We can never guarantee that a single Cancel call (or X sequential Cancel calls) will be enough to actually cancel the command; this is why Thread B now calls Cancel in a loop.

A second comment on the same TestSqlCommandCancel test: following my changes, there's another couple of errors which look worrying:
<details>
<summary>Managed SNI, on Linux</summary>

```
  [xUnit.net 00:01:29.38]     Microsoft.Data.SqlClient.ManualTesting.Tests.AlwaysEncrypted.ApiShould.TestSqlCommandCancel(connection: "Data Source=tcp:10.0.0.4;Database=NORTHWIND;UID=sa"···, value: "ExecuteNonQuery") [FAIL]
  [xUnit.net 00:01:29.39]       System.AggregateException : One or more errors occurred. (Assert.Equal() Failure: Strings differ
  [xUnit.net 00:01:29.39]                  ↓ (pos 0)
  [xUnit.net 00:01:29.39]       Expected: "Operation cancelled by user."
  [xUnit.net 00:01:29.39]       Actual:   "The request failed to run because the bat"···
  [xUnit.net 00:01:29.39]                  ↑ (pos 0))
  [xUnit.net 00:01:29.39]       ---- Assert.Equal() Failure: Strings differ
  [xUnit.net 00:01:29.39]                  ↓ (pos 0)
  [xUnit.net 00:01:29.39]       Expected: "Operation cancelled by user."
  [xUnit.net 00:01:29.39]       Actual:   "The request failed to run because the bat"···
  [xUnit.net 00:01:29.39]                  ↑ (pos 0)
  [xUnit.net 00:01:29.39]       Stack Trace:
  [xUnit.net 00:01:29.39]            at System.Threading.Tasks.Task.ThrowIfExceptional(Boolean includeTaskCanceledExceptions)
  [xUnit.net 00:01:29.39]            at System.Threading.Tasks.Task.Wait(Int32 millisecondsTimeout, CancellationToken cancellationToken)
  [xUnit.net 00:01:29.39]            at System.Threading.Tasks.Task.Wait()
  [xUnit.net 00:01:29.39]         /_/src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/ApiShould.cs(2012,0): at Microsoft.Data.SqlClient.ManualTesting.Tests.AlwaysEncrypted.ApiShould.TestSqlCommandCancel(String connection, String value)
  [xUnit.net 00:01:29.39]            at InvokeStub_ApiShould.TestSqlCommandCancel(Object, Span`1)
  [xUnit.net 00:01:29.39]            at System.Reflection.MethodBaseInvoker.InvokeWithFewArgs(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
  [xUnit.net 00:01:29.39]         ----- Inner Stack Trace -----
  [xUnit.net 00:01:29.39]         /_/src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/ApiShould.cs(3160,0): at Microsoft.Data.SqlClient.ManualTesting.Tests.AlwaysEncrypted.ApiShould.Thread_ExecuteNonQuery(Object state)
  [xUnit.net 00:01:29.39]            at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
  [xUnit.net 00:01:29.39]         --- End of stack trace from previous location ---
  [xUnit.net 00:01:29.39]            at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
  [xUnit.net 00:01:29.39]            at System.Threading.Tasks.Task.ExecuteWithThreadLocal(Task& currentTaskSlot, Thread threadPoolThread)
```

</details>

<details>
<summary>Native SNI, on Windows</summary>

```
  [xUnit.net 00:01:29.12]     Microsoft.Data.SqlClient.ManualTesting.Tests.AlwaysEncrypted.ApiShould.TestSqlCommandCancel(connection: "Data Source=tcp:10.0.0.4;Database=NORTHWIND;UID=sa"..., value: "ExecuteReader") [FAIL]
  [xUnit.net 00:01:29.12]       System.AggregateException : One or more errors occurred.
  [xUnit.net 00:01:29.12]       ---- Assert.Equal() Failure
  [xUnit.net 00:01:29.12]                 Γåô (pos 0)
  [xUnit.net 00:01:29.12]       Expected: Operation cancelled by user.
  [xUnit.net 00:01:29.12]       Actual:   A severe error occurred on the current co┬╖┬╖┬╖
  [xUnit.net 00:01:29.12]                 Γåæ (pos 0)
  [xUnit.net 00:01:29.13]       Stack Trace:
  [xUnit.net 00:01:29.13]            at System.Threading.Tasks.Task.ThrowIfExceptional(Boolean includeTaskCanceledExceptions)
  [xUnit.net 00:01:29.13]            at System.Threading.Tasks.Task.Wait(Int32 millisecondsTimeout, CancellationToken cancellationToken)
  [xUnit.net 00:01:29.13]            at System.Threading.Tasks.Task.Wait()
  [xUnit.net 00:01:29.13]         /_/src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/ApiShould.cs(2013,0): at Microsoft.Data.SqlClient.ManualTesting.Tests.AlwaysEncrypted.ApiShould.TestSqlCommandCancel(String connection, String value)
  [xUnit.net 00:01:29.13]         ----- Inner Stack Trace -----
  [xUnit.net 00:01:29.13]         /_/src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/ApiShould.cs(3138,0): at Microsoft.Data.SqlClient.ManualTesting.Tests.AlwaysEncrypted.ApiShould.Thread_ExecuteReader(Object state)
  [xUnit.net 00:01:29.13]            at System.Threading.Tasks.Task.Execute()
```

</summary>

I think the managed SNI exception message is probably "The request failed to run because the batch is aborted, this can be caused by abort signal sent from client, or another request is running in the same session, which makes the session busy." If so, it's another instance of #26. The native SNI exception message looks like a bug though.

---

While this test can fail in two ways, the failure now shows up as a test failure rather than a crash & restart of testhost. I can revert my third commit (the changes to TestSqlCommandCancel) if it's better discussed in an issue.

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/TestFixtures/DatabaseHelper.cs
```diff
@@ -299,10 +299,10 @@ public IEnumerator<object[]> GetEnumerator()
         {
             foreach (string connStrAE in DataTestUtility.AEConnStrings)
             {
-                yield return new object[] { connStrAE, @"ExecuteReader", 1 };
-                yield return new object[] { connStrAE, @"ExecuteReader", 3 };
-                yield return new object[] { connStrAE, @"ExecuteNonQuery", 1 };
-                yield return new object[] { connStrAE, @"ExecuteNonQuery", 3 };
+                yield return new object[] { connStrAE, @"ExecuteReader" };
```
benrr101, line 302: If we remove the `1` and `3` parameters, aren't we basically saying run the test twice for `ExecuteReader` and `ExecuteNonQuery`?

... actually these tests are really a mess. I want to clean them up so bad but I just don't have the bandwidth to do it :(

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/MirroringTest/ConnectionOnMirroringTest.cs
```diff
@@ -28,17 +29,14 @@ public static void TestMultipleConnectionToMirroredServer()
                 builder.ConnectTimeout = 0;
 
                 TestWorker worker = new TestWorker(builder.ConnectionString);
-                Thread childThread = new Thread(() => worker.TestMultipleConnection());
-                childThread.Start();
+                Task childTask = Task.Factory.StartNew(() => worker.TestMultipleConnection(), TaskCreationOptions.LongRunning);
```
benrr101, line 32: I'm curious if we really need to be doing these as long running tasks or if this code is just so old that it predates the TPL and so we had no choice but use threads for everything. If it doesn't need to be long running, I'd say we drop that and use `Task.Run` or better yet make `TestMultipleConnection` be async.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ParameterTest/ParametersTest.cs
```diff
@@ -902,13 +904,12 @@ public static void ClosedConnection_SqlParameterValueTest()
...
                         }
                     }
-                });
-                t.Start();
-                threads.Add(t);
+                }, TaskCreationOptions.LongRunning);
+                tasks.Add(t);
             }
-            for (int i = 0; i < threads.Count; i++)
+            for (int i = 0; i < tasks.Count; i++)
```
benrr101, line 910: This is basically the same as saying Task.WaitAll, right?

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ParameterTest/ParametersTest.cs
```diff
@@ -902,13 +904,12 @@ public static void ClosedConnection_SqlParameterValueTest()
                             Assert.Fail($"Unexpected exception occurred: {e.Message}");
```
benrr101, line 904: This is super redundant ... the test will fail if an unexpected exception is thrown, so there's no need for this. *sigh* I can't expect you to fix all the issues with this code but it really needs help

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/ApiShould.cs
```diff
@@ -1983,29 +1979,38 @@ public void TestSqlCommandCancel(string connection, string value, int number)
 
                     CommandHelper.s_sleepDuringTryFetchInputParameterEncryptionInfo?.SetValue(null, true);
 
-                    Thread[] threads = new Thread[2];
+                    Task[] tasks = new Task[2];
+                    ManualResetEventSlim entryLock = new ManualResetEventSlim(false);
```
benrr101, line 1983: Although this is clever, I feel like this does complicate this test significantly. I'm wondering if it makes more sense to do something simple like making the command be a WAITFOR statement that guarantees we'll have the execute task running before attempting to cancel it.

benrr101 (changes requested):
@edwardneal thanks so much for helping to make our automated testing more robust!! This is one of my biggest complaints in my day to day work, so anything that reduces the number of times I have to restart a build b/c of flaky tests is super appreciated.

I do have some concerns regarding the interlocking change, and a few other comments I'd like to see addressed before checking it in. But, I'll approve it asap once they are addressed

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/MirroringTest/ConnectionOnMirroringTest.cs
edwardneal: I think it's the age of the code. My starting point was to make sure that exceptions wouldn't crash testhost, so I was just thinking about how to make sure the second thread's exception propagated into the main thread. The TPL guarantees that, and I used LongRunning to guarantee that we would create a new thread rather than use something on the thread pool.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ParameterTest/ParametersTest.cs
edwardneal: Yes - agreed!

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/ApiShould.cs
edwardneal: While a WAITFOR makes perfect sense when dealing with an ExecuteScalar call, we need to test what happens if someone cancels a SqlCommand while another thread is iterating through the ExecuteReader results - we need to put that WAITFOR "in the middle of" the SELECT statement. I can't put it in a scalar-valued UDF and call that, it has a side effect.

A simple `Thread.Sleep` at the start of the cancellation thread would probably work... but that's naturally going to vary based on latency and command execution times, even on the CI legs (because these test both on local SQL Server instances and Azure SQL instances.) Too low, and thread B's cancellation logic will set the SqlCommand's internal cancellation state, just in time for thread A to call ExecuteScalar and reset that state without ever respecting the cancellation request. Too high, and thread A will have connected and completed execution before thread B cancels the command.

You're not alone in disliking the test! I used the combination of entry/exit locks and hammering on SqlCommand.Cancel as a way to make it pass most of the time, only to be met with an error in the native SNI library which looks suspiciously like memory corruption...

The operative part of this PR is purely to make sure that exceptions propagate across threads cleanly rather than terminate the process. If we're opening a can of worms, I'm perfectly happy to revert the [third commit](https://github.com/dotnet/SqlClient/pull/2968/commits/7c766b44dceee221d5c6b6eef002eb2865649fd7) and open an issue so we can look at it later.

benrr101: Hmm. Maybe I'm thinking a bit too hard about this... Looking at the original code, Thread_ExecuteReader is expecting to get a reader back, but all attempts at reading it are expected to fail. That's an even bigger race condition where the cancellation must happen between returning the metadata for the first result set and reading the first record. 

(I'm kinda struggling to fully understand the changes from the diff alone, and don't really want to load it all up in my ide right now, so forgive me if I'm restating the obvious) My thought would be to basically have this sequence

|Execute Thread|Cancel Thread|
|-----------------|---------------|
|Start thread|-|
|Execute reader|StartThread|
|-|Wait for Executed|
|Set Executed|-|
|Wait for Cancelled|-|
|-|Cancel|
|-|Set Cancelled|
|Read (and fail)|-|
|-End-|-End-|

Maybe that's exactly what you're trying to do and I'm just not seeing it. In my defense "entry" and "exit" locks I guess don't explain it well enough to me. Idk, what do you think?

edwardneal: The original test isn't an ideal point of reference; I'm not actually convinced it's a valid test case at all. It essentially opens a data reader, then iterates over every row in the result set and tries to open another data reader from the same command, expecting that to throw an InvalidOperationException.

The description in my previous comment ("we need to test what happens if someone cancels a SqlCommand while another thread is iterating through the ExecuteReader results") is the behaviour I want to test, but the current code doesn't do that. I'll correct it.

Your expected cross-thread behaviour is pretty much there. The only difference is that I've swapped the order that the threads start in: the "workload" thread (whether this is ExecuteScalar or ExecuteReader) will wait until the cancellation thread signals it, rather than vice versa. I don't think it makes a real difference in this case, I just thought about the execution flow in terms of "that workload needs to stay where it is until I'm ready to start hammering the Cancel invocation".

The entry and exit locks refer to the workload thread's activity: they wait for the entry lock to be signalled before starting, then signal the exit lock to indicate that they've finished. The cancellation thread is the inverse: it signals the entry lock to indicate that it's ready for the workload thread to start, and calls SqlCommand.Cancel until the exit lock is signalled. The naming on these probably needs some work: what about calling them "StartWorkloadSignal" and "WorkloadCompleteSignal"?

benrr101: > ... I'm not actually convinced it's a valid test case at all. ...

I guess I also have no idea these tests are special to always encrypted - you'd think encryption has no bearing on cancelling a query...

> ... The only difference is that I've swapped the order that the threads start in: the "workload" thread ...

Ok, maybe we can give a little comment similar to the table I proposed just so we can have some documentation of what the expected behavior is. Otherwise I could see myself getting very lost in this code.

> ... what about calling them "StartWorkloadSignal" and "WorkloadCompleteSignal"?

Sure, I think that's a better naming scheme 👍

benrr101 (dismissed):

benrr101:
@edwardneal Thanks again for working on this. I really *really* appreciate any help with improving our build process.

Once we get another review, I'll be happy to merge this 👍🚀

cheenamalhotra (dismissed):

cheenamalhotra:
It seems there are test failures:

```
Failed Microsoft.Data.SqlClient.ManualTesting.Tests.AlwaysEncrypted.ApiShould.TestSqlCommandCancel(connection: "Data Source=tcp:10.0.0.5;Database=NORTHWIND;UID=sa"···, value: "ExecuteNonQuery") [1 s]
     System.AggregateException : One or more errors occurred. (Assert.Equal() Failure: Strings differ
             ↓ (pos 0)
  Expected: "Operation cancelled by user."
  Actual:   "The request failed to run because the bat"···
             ↑ (pos 0))
  ---- Assert.Equal() Failure: Strings differ
             ↓ (pos 0)
  Expected: "Operation cancelled by user."
  Actual:   "The request failed to run because the bat"···
             ↑ (pos 0)
    Stack Trace:
       at System.Threading.Tasks.Task.WaitAllCore(Task[] tasks, Int32 millisecondsTimeout, CancellationToken cancellationToken)
     at System.Threading.Tasks.Task.WaitAll(Task[] tasks)
     at Microsoft.Data.SqlClient.ManualTesting.Tests.AlwaysEncrypted.ApiShould.TestSqlCommandCancel(String connection, String value) in /_/src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/ApiShould.cs:line 2134
     at InvokeStub_ApiShould.TestSqlCommandCancel(Object, Span`1)
     at System.Reflection.MethodBaseInvoker.InvokeWithFewArgs(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
...

Failed Microsoft.Data.SqlClient.ManualTesting.Tests.AlwaysEncrypted.ApiShould.TestSqlCommandCancel(connection: "Data Source=tcp:10.0.0.4;Database=NORTHWIND;UID=sa"···, value: "ExecuteReader") [1 s]
     System.AggregateException : One or more errors occurred. (Assert.Equal() Failure: Strings differ
             ↓ (pos 0)
  Expected: "Operation cancelled by user."
  Actual:   "The request failed to run because the bat"···
             ↑ (pos 0))
  ---- Assert.Equal() Failure: Strings differ
             ↓ (pos 0)
  Expected: "Operation cancelled by user."
  Actual:   "The request failed to run because the bat"···
             ↑ (pos 0)
    Stack Trace:
       at System.Threading.Tasks.Task.WaitAllCore(Task[] tasks, Int32 millisecondsTimeout, CancellationToken cancellationToken)
     at System.Threading.Tasks.Task.WaitAll(Task[] tasks)
     at Microsoft.Data.SqlClient.ManualTesting.Tests.AlwaysEncrypted.ApiShould.TestSqlCommandCancel(String connection, String value) in /_/src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/ApiShould.cs:line 2080
     at InvokeStub_ApiShould.TestSqlCommandCancel(Object, Span`1)
     at System.Reflection.MethodBaseInvoker.InvokeWithFewArgs(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
...
```

cheenamalhotra:
It also appears this test failed on Windows too in first round and passed on rerunning, so I'm a little skeptic of this change, though they look promising..

```
Failed Microsoft.Data.SqlClient.ManualTesting.Tests.AlwaysEncrypted.ApiShould.TestSqlCommandCancel(connection: "Data Source=tcp:10.0.0.4;Database=NORTHWIND;UID=sa"..., value: "ExecuteReader") [976 ms]
##[error]EXEC(0,0): Error Message: 
EXEC : error Message:  [D:\a\_work\1\s\build.proj]
     System.AggregateException : One or more errors occurred.
  ---- Assert.Equal() Failure
            Γåô (pos 0)
  Expected: Operation cancelled by user.
  Actual:   A severe error occurred on the current co┬╖┬╖┬╖
            Γåæ (pos 0)
    Stack Trace:
       at System.Threading.Tasks.Task.WaitAll(Task[] tasks, Int32 millisecondsTimeout, CancellationToken cancellationToken)
     at System.Threading.Tasks.Task.WaitAll(Task[] tasks, Int32 millisecondsTimeout)
     at System.Threading.Tasks.Task.WaitAll(Task[] tasks)
     at Microsoft.Data.SqlClient.ManualTesting.Tests.AlwaysEncrypted.ApiShould.TestSqlCommandCancel(String connection, String value) in /_/src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/ApiShould.cs:line 2027
```

edwardneal:
Thanks @cheenamalhotra, that's fair. The main objective of this PR is to prevent the exceptions thrown by the secondary thread from killing the testhost process. At the moment, a flaky test kills the process and CI restarts it; this adds 10-15 minutes to the build time. With this change, the flaky test fails, the process returns a non-zero exit code and CI restarts it.

I hadn't spotted the second restart criteria earlier, so this PR doesn't reduce the build time as I first thought. It does make some previously-failing tests visible though, and hopefully some future work can start moving progressively more tests into a test run which doesn't have that restart criteria, allowing us to shift the remaining ones towards restarting the individual test rather than the entire testhost.

The exception message which starts with "The request failed to run because the bat" is a simple enough fix, I'm pretty sure it's just another instance of [efcore#29861](https://github.com/dotnet/efcore/issues/29861) and will change the test to account for that. I'm more concerned about the exception with the corrupted message string though. That looks a little like the native SNI's `SNIGetLastError` method is overwriting or freeing the memory behind the `SNI_Error.errorMessage` member which it returns to TdsParser.

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/ApiShould.cs
```diff
@@ -1949,19 +1957,15 @@ public void TestBeginAndEndExecuteReaderWithAsyncCallback(string connection, Com
             }
         }
 
-        [SkipOnTargetFramework(TargetFrameworkMonikers.Netcoreapp)]
         [ConditionalTheory(typeof(DataTestUtility), nameof(DataTestUtility.AreConnStringSetupForAE))]
         [ClassData(typeof(AEConnectionStringProviderWithExecutionMethod))]
-        public void TestSqlCommandCancel(string connection, string value, int number)
+        public void TestSqlCommandCancel(string connection, string value)
```
paulmedynski, line 1962: I think we're missing some mocking/injection layers here.  I don't see why testing an outstanding command being cancelled requires any networking or SQL Servers or elaborate thread interaction.  Do we not have a way to mock at, say, the SNI layer?

edwardneal: We technically have a TdsServer project which can act as a basic mock of a functioning TDS server and run queries. I'd prefer to revisit whether a test should be in the Tests or ManualTesting project in a separate PR though - the places we use InternalsVisibleTo will change a number of plans anyway.

paulmedynski: For sure - not suggesting any changes for this PR.  I was just thinking out loud.  These tests are getting pretty complicated, so wondering about a longer term approach to simplify things.

edwardneal: Definitely. In the long run, I'd like to be able to force TdsServer to cancel while halfway through sending metadata, or at exactly the same time as a CancellationToken fires. These cases are difficult to trigger organically, and make it harder to prove our work with TdsParser.

PR #2972, opened 2024-11-04
Merge | Cleanup Interop Handles

benrr101:
**Description**: This PR is more prerequisite work to help my sanity while working on merging the netfx and netcore projects. This particular PR wrangles handles that are being used for interop stuff. Here are the highlights:

* Deleting most of the handles classes since they were only being used in net6 (which is no longer supported)
* Moving the remaining `SafeLibraryHandle` to a temporary namespace
  * Temporary namespace is created because as long as the Interop class in the root namespace exists, we cannot have an Interop namespace
  * Temporary namespace will continue to be used for the next few PRs until it is cleaned up in the final PR

**Testing**: Projects continue to build, this PR makes no functional changes to the code, just deleted dead code and changes a couple namespaces.

@edwardneal your review on this would be appreciated :)


edwardneal (commented):
Thanks for the ping. I've not got anything to add and CI is compiling, so this looks good to me. Onwards and upwards!

src/Microsoft.Data.SqlClient/src/Interop/Windows/Handles/SafeLibraryHandle.netcore.cs
```diff
@@ -2,9 +2,11 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using Microsoft.Win32.SafeHandles;
+
 #if NET
 
-namespace Microsoft.Win32.SafeHandles
+namespace Interop_TEMP.Windows.Handles
```
samsharma2700, line 9: Should we leave a TODO or comment here, till it is cleaned up in the last PR.

samsharma2700 (approved):

mdaigle (approved):

PR #2981, opened 2024-11-06
Remove SQL 2000 client-side debugging support

edwardneal:
Follows up #2940, removing the client-side support for SQL 2000's debugging mechanisms from netfx and aligning with netcore.

The SqlConnection diff will be easier to parse with whitespace changes disabled. I've removed the methods which enable the now-defunct debugging, the `SqlDebugContext` class and `MEMMAP` struct which they use and the PInvokes they reference. These PInvokes were the only thing in `netfx\src\Microsoft\Data\Common\NativeMethods.cs`, so I've deleted the file.

This isn't critical for preview3.

/cc @benrr101 - there's no need to port PInvokes from NativeMethods.cs anywhere. It's the same story for the SafeNativeMethods.cs file in the same folder.


benrr101:
Uhhhhhhhhhhhhhhhhhhhhhhhhhhh 😅 
Can I get my PR in first and you can delete whatever from it that is no longer needed?

edwardneal:
No worries - I'll just merge back into my branch once #2997 is done.

benrr101 (approved):
Let's gooooooooo delete that code!

MichelZ:
#2839  next! :)

David-Engel (approved):

PR #2982, opened 2024-11-06
Merge | Align Task usage / ArrayPool / IsColumnEncryptionSupported netcore/netfx

MichelZ:
Part of #2953 

Aligning Task return, bringing over ArrayPool usage and aligning IsColumnEncryptionSupported changes between netcore and netfx


src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -10512,7 +10512,9 @@ internal Task TdsExecuteRPC(SqlCommand cmd, IList<_SqlRPC> rpcArray, int timeout
                                         stateObj.WriteByte(TdsEnums.SQL70_DEFAULT_NUMERIC_PRECISION);
                                 }
                                 else
+                {
```
benrr101, line 10531: indentation looks wrong here 🤔

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -6710,19 +6710,35 @@ private TdsOperationStatus TryReadSqlStringValue(SqlBuffer value, byte type, int
...
+                            if (result == TdsOperationStatus.Done)
                             {
-                                return result;
+                                if (length > 0)
+                                {
+                                    s = new string(cc, 0, length);
+                                }
+                                else
+                                {
+                                    s = "";
```
benrr101, line 6724: string.Empty?

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -13693,32 +13710,42 @@ internal int ReadPlpUnicodeChars(ref char[] buff, int offst, int len, TdsParserS
         // requested length is -1 or larger than the actual length of data. First call to this method
         //  should be preceeded by a call to ReadPlpLength or ReadDataLength.
         // Returns the actual chars read.
-        internal TdsOperationStatus TryReadPlpUnicodeChars(ref char[] buff, int offst, int len, TdsParserStateObject stateObj, out int totalCharsRead)
+        internal TdsOperationStatus TryReadPlpUnicodeChars(ref char[] buff, int offst, int len, TdsParserStateObject stateObj, out int totalCharsRead, bool supportRentedBuff, ref bool rentedBuff)
```
benrr101, line 13721: I think once it gets to be this long, we should probably chop the parameters like
```
methodName(
  arg1,
  arg2,
  ...
  argn)
```

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -13693,32 +13710,42 @@ internal int ReadPlpUnicodeChars(ref char[] buff, int offst, int len, TdsParserS
...
             if (stateObj._longlen == 0)
             {
                 Debug.Assert(stateObj._longlenleft == 0);
                 totalCharsRead = 0;
                 return TdsOperationStatus.Done;       // No data
             }
 
-            Debug.Assert(((ulong)stateObj._longlen != TdsEnums.SQL_PLP_NULL),
-                    "Out of sync plp read request");
+            Debug.Assert(((ulong)stateObj._longlen != TdsEnums.SQL_PLP_NULL), "Out of sync plp read request");
```
benrr101, line 13734: Remove redundant parentheses, please :)

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -12807,119 +12807,127 @@ private async Task WriteXmlFeed(XmlDataFeed feed, TdsParserStateObject stateObj,
         private async Task WriteTextFeed(TextDataFeed feed, Encoding encoding, bool needBom, TdsParserStateObject stateObj, int size, bool useReadBlock)
         {
             Debug.Assert(encoding == null || !needBom);
-            char[] inBuff = new char[constTextBufferSize];
+            char[] inBuff = ArrayPool<char>.Shared.Rent(constTextBufferSize);
```
benrr101, line 12810: For something like this, can we just do a stackalloc?

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -12807,119 +12807,127 @@ private async Task WriteXmlFeed(XmlDataFeed feed, TdsParserStateObject stateObj,
...
+            ArrayPool<char>.Shared.Return(inBuff, clearArray: true);
         }
 
         private async Task WriteStreamFeed(StreamDataFeed feed, TdsParserStateObject stateObj, int len)
         {
-            TdsOutputStream output = new TdsOutputStream(this, stateObj, null);
-            byte[] buff = new byte[constBinBufferSize];
-            int nWritten = 0;
-            do
+            byte[] buff = ArrayPool<byte>.Shared.Rent(constBinBufferSize);
```
benrr101, line 12887: Same question here, can we use a stackalloc?

benrr101 (changes requested):
In general, I think this is a decent change. I'm hesitant to start touching things in the TdsParser, but these changes seem to be pretty constrained. Only a few comments. Not sure about the feasibility of using stackalloc instead of array pool in the last commit.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
MichelZ: That's unsafe territory... I don't think I'm the right person for that :)

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
MichelZ: That's unsafe territory... I don't think I'm the right person for that :)

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
benrr101: Understandable. I think we *could* consider changing this to stackalloc, but it would also require changing the type of the variable, and any usages of the variable, and we'd need to consider the size we're allocating... etc etc. So, yes, I see how it could be unsafe while not technically `unsafe` :)

benrr101 (approved):

PR #2983, opened 2024-11-06
Merge | Interop NtDll Namespace

benrr101:
**Description**: This PR is the next phase of the interop namespace migration. This PR targets files that are related to the ntdll dlls. It's the same strategy as before - each commit should build by itself, they move a single file at a time. Files that has multiple classes/structs/etc in it are split into one file per class/struct/etc.

**Testing**: No functional changes made, just moving files around. Should be good if the build passes


mdaigle (approved):
LGTM

cheenamalhotra (approved):

samsharma2700 (approved):

PR #2984, opened 2024-11-07
Merge | Move ReliabilitySection / BestEffortCleanup to framework specific files

MichelZ:
I don't think we want to wrap those in #IF NETFRAMEWORK / NETCORE conditional compiles, it makes sense to have them in their own files, so I've moved them out of the "future-common-project" file(s)

Part of #2953


benrr101 (approved):
Overall I think this makes sense and is probably a good idea. Still not a huge fan of the conditional attribute, but since it's already there, I won't raise a fuss :)

cheenamalhotra (approved):

benrr101:
@MichelZ if you can resolve the conflict, I can get this merged 🚢

MichelZ:
@benrr101 Done! Thx

PR #2988, opened 2024-11-07
Enable testing with Ubuntu 22

cheenamalhotra:


cheenamalhotra:
We don't seem to have .NET 9 available in Ubuntu 22 runner images just yet. [Latest image](https://github.com/actions/runner-images/blob/releases/ubuntu22/20241201/images/ubuntu/Ubuntu2204-Readme.md)
Neither can we install them as artifacts, due to .NET 9 missing from packages.microsoft.com, ref: https://github.com/dotnet/sdk/issues/44869

MichelZ:
@cheenamalhotra Can you update the branch from main and see if it works?
I think the pipeline should take care of all .NET needs now

MichelZ:
Also from the latest build the issue seems to be that it fails to configure SQL Server:
https://sqlclientdrivers.visualstudio.com/public/_build/results?buildId=101168&view=logs&j=73c774a9-2c54-5dc8-eea7-24a41fc100ce&t=9c8fa434-0e21-5143-7879-db8639331619
```

/usr/bin/bash /mnt/vss/_work/_temp/346a7ca7-f41d-40a6-89d8-7f847b3fb24f.sh
Failed to stop mssql-server.service: Unit mssql-server.service not loaded.
Running mssql-conf setup...
sudo: a terminal is required to read the password; either use the -S option to read from standard input or configure an askpass helper
sudo: a password is required
Waiting for SQL Server to start...
Waiting for SQL Server to start...
Waiting for SQL Server to start...
Waiting for SQL Server to start...
Waiting for SQL Server to start...
Cannot connect to SQL Server, installation aborted

##[error]Bash exited with code '1'.
Finishing: Configure SQL Server [Linux]

```

PR #2990, opened 2024-11-08
Code health - preprocessor directives

MichelZ:
Unifying preprocessor directives to NET8_0_OR_GREATER

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/ManagedSSPIContextProvider.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/ManagedSSPIContextProvider.cs
@@ -1,4 +1,4 @@
-﻿#if NET6_0
+﻿#if NET8_0_OR_GREATER
 
 using System;
 using Microsoft.Data.SqlClient.SNI;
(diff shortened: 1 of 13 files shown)


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/ManagedSSPIContextProvider.cs
```diff
@@ -1,4 +1,4 @@
-﻿#if NET6_0
```
MichelZ, line 1: This is probably a bug, or does this class need removal in NET8?

ErikEJ: It should be removed in NET 8

ErikEJ: That's why I marked it

MichelZ: I think I'm going to remove all these other !NET8_0_OR_GREATER in another PR

ErikEJ:
I have seen someone mention using !NETFRAMEWORK instead, maybe that is better, avoids doing this for each new .NET?

MichelZ:
> I have seen someone mention using !NETFRAMEWORK instead, maybe that is better, avoids doing this for each new .NET?

True, not the only 8+ is supported anyways, that probably makes sense

benrr101 (approved):
No issues here 🚢

David-Engel (approved):

cheenamalhotra (approved):

PR #2992, opened 2024-11-08
WIP - See if it's worth speeding up the build a bit

MichelZ:
Some of the stages take much longer than others. I'm testing out if it's maybe worth to split the X64 and X86 tests out into separate steps/stages


benrr101:
Hmmm... I like the idea, but I think I'd want some feedback from @cheenamalhotra to see if it fits in with today's build/test process.

cheenamalhotra (approved):
LGTM

PR #2994, opened 2024-11-08
Merge | Interop - SChannel, SspiCli, Unix, Move Interop_TEMP to Interop

benrr101:
**Description**: This is the second to last PR for the interop code. It was originally going to be much larger, but with the removal of net6 code, it became a lot smaller. As with the previous PRs in this series, the commits have been made into bite-sized, fully functional changes. Here's a rundown of the changes:

* Removing SChannel code since it was only used in net6
* Removing code from SspiCli that was only used in net6
* Moving the remaining SspiCli code into Interop_TEMP namespace
* Remove the Unix interop code since it was only being used in net6
* Remove Interop.Libraries
* Move Interop_TEMP to Interop

There will be one more interop change after this, which migrates the [safe]nativemethods classes.

**Testing**: Once again, this just moves files around but does not change the functionality. If the test build passes, we should be good to go.


mdaigle (approved):

mdaigle (commented):
Overall, looks good. Looks like the build is complaining about the libraries file.

samsharma2700 (approved):

PR #2996, opened 2024-11-09
Remove SqlCommand code paths for context connections

edwardneal:
Contributes to #2838. PR 2/7.

This doesn't need to be merged before 6.0. It removes the core `SqlDataReaderSmi` type, and the code paths in `SqlCommand.ExecuteNonQuery` and `SqlCommand.ExecuteReader` which reference it. There are also a handful of Stream and TextReader derivatives which are only used in those contexts.

One point of note is in SqlDataRecord. InOutOfProcHelper.InProc will always be false, so I've trimmed the dead code in the if condition. As a result, `_recordContext` will always be null, and null will always propagate to the `context` parameter in ValueUtilsSmi. I plan to clean this up in the next PR.

If this is too large to review, I can partially revert some of these removals.


src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlCommand.cs
```diff
@@ -2067,20 +2041,9 @@ private Task InternalExecuteNonQuery(TaskCompletionSource<object> completion, st
...
-                            statistics.SafeIncrement(ref statistics._unpreparedExecs);
-                        }
-
-                        RunExecuteNonQuerySmi(sendToPipe);
-                    }
-
                     //Always Encrypted generally operates only on parameterized queries. However enclave based Always encrypted also supports unparameterized queries
                     //We skip this block for enclave based always encrypted so that we can make a call to SQL Server to get the encryption information
-                    else if (!ShouldUseEnclaveBasedWorkflow && !_batchRPCMode && (System.Data.CommandType.Text == this.CommandType) && (0 == GetParameterCount(_parameters)))
+                    if (!ShouldUseEnclaveBasedWorkflow && !_batchRPCMode && (System.Data.CommandType.Text == this.CommandType) && (0 == GetParameterCount(_parameters)))
```
benrr101, line 2046: While you're here, can you remove the redundant parentheses and flip the 0 == condition? :)

benrr101 (dismissed):
I am 10000% ok with these changes. My only complain its I'm kinda jealous I'm not the one getting to delete all this code 💣

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlCommand.cs
edwardneal: Yes - done!

edwardneal:
I've addressed the merge conflict on this, can someone re-run CI please?

edwardneal:
The new merge conflict is now resolved. I'd appreciate a CI run and review.

edwardneal:
I've dealt with the merge conflict - could this be run through CI and reviewed please?

benrr101:
@edwardneal How's this one looking now that we've got most of the SMI code removed? Can we close it out or is there other stuff that we haven't covered yet?

edwardneal:
We're almost done. The only things remaining are the work submitted in #3438, the removal of [`SmiRecordBuffer`](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SmiRecordBuffer.cs) and some updates to `SqlDataRecord` to remove the (now nearly identical) [netcore](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SqlDataRecord.netcore.cs) and [netfx](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SqlDataRecord.netfx.cs) versions.

I don't think we need this PR for that - once we're happy with #3438, I'll clean the two other points up separately.

PR #2997, opened 2024-11-11
Merge | NativeMethods/SafeNativeMethods

benrr101:
**Description**: This is the last major change for the Interop namespace. There was a handful of interop calls tucked away in the native safe and unsafe classes as well as a super secret one in the SNI wrapper class. For this PR they were moved into the Interop namespace in the classes that correspond to the DLLs they were referencing.

_One semi-important note_: Everything in SafeNativeMethods was moved into a Kernel32Safe class - this was because the "Safe" methods were given an attribute that indicates that the security checks when calling across the managed/unmanaged boundary should not be performed. Rather than potentially change perf, I've elected to leave it as-is but call out when the calls are "safe".

**Testing**: For the most part, this PR just moves stuff around without changing functionality. It should be good if the tests pass.


mdaigle (approved):

src/Microsoft.Data.SqlClient/src/Interop/Windows/Kernel32/Kernel32Safe.netfx.cs
```diff
@@ -1,41 +1,43 @@
...
+namespace Interop.Windows.Kernel32
 {
-
-    [SuppressUnmanagedCodeSecurityAttribute()]
-    internal static class SafeNativeMethods
+    /// <remarks>
+    /// Be insanely careful when using methods from this class. They are configured to skip
+    /// runtime security checks.
+    /// </remarks>
+    [SuppressUnmanagedCodeSecurity]
```
mdaigle, line 19: Are there few enough spots where these methods are called that we could add the suppression at each call site and remove it here without it being a huge burden? Then you would get a compiler error if you add a new unsuppressed call.

samsharma2700 (approved):

mdaigle (approved):

src/Microsoft.Data.SqlClient/src/Interop/Windows/Kernel32/Kernel32Safe.netfx.cs
benrr101: I don't think we could move it to the call site, since the attribute only applies to classes, methods, delegates. Unless you mean to add it to the method signature for a method that calls the "safe" methods... then I'm not really sure what would happen. Like, I don't know if it applies to all things down the stack or if it only applies to immediate transitions from managed to unmanaged. I think this is likely the easiest method.

src/Microsoft.Data.SqlClient/src/Interop/Windows/Kernel32/Kernel32Safe.netfx.cs
```diff
@@ -1,41 +1,43 @@
...
         [ResourceExposure(ResourceScope.Process)]
-        static internal extern IntPtr GetModuleHandle([MarshalAs(UnmanagedType.LPTStr), In] string moduleName/*lpctstr*/);
+        internal static extern IntPtr GetModuleHandle([MarshalAs(UnmanagedType.LPTStr), In] string moduleName);
 
         [DllImport(Kernel32, CharSet = CharSet.Ansi, BestFitMapping = false, ThrowOnUnmappableChar = true, SetLastError = true)]
-        //        [DllImport(ExternDll.Kernel32, CharSet=CharSet.Ansi)]
         [ResourceExposure(ResourceScope.None)]
-        static internal extern IntPtr GetProcAddress(IntPtr HModule, [MarshalAs(UnmanagedType.LPStr), In] string funcName/*lpcstr*/);
-
+        internal static extern IntPtr GetProcAddress(IntPtr HModule, [MarshalAs(UnmanagedType.LPStr), In] string funcName);
```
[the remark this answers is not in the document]
benrr101: Ohhhhhh... yeah. I've got most of those staged up, not sure if it's ready yet.

cheenamalhotra (approved):

cheenamalhotra:
Little concerned about why the Windows Enclave 'functional' tests are hung up.. any idea what's going on?

Issue #646, opened 2020-07-12
Add async System.Data resultset and database schema APIs 


roji:
https://github.com/dotnet/runtime/issues/38028 adds async schema APIs to System.Data, on both DbConnection and DbDataReader; these automatically fall back to calling the sync versions if not overridden. However, SqlClient will now be able to override them, and provide truly async support for these operations.

Note: this would require cross-targeting to net5.0.

cheenamalhotra:
Thanks @roji

Acknowledged for .NET 5.0 new APIs.

PR #3005, opened 2024-11-14
Feature | Implement SqlConnection.GetSchemaAsync

edwardneal:
Contributes to #646, adding `SqlConnection.GetSchemaAsync`.

I can't add the IDbColumnSchemaGenerator method because the .NET Framework version of SqlDataReader doesn't yet implement that interface. #2967 tracks this. `SqlDataReader.GetSchemaTableAsync` requires some extra work with the state machine to guarantee that the column metadata is available, and I'll submit a separate PR for this.

I've added some basic documentation, but this effectively follows the `SqlCommand.[Something]Async` pattern of "An asynchronous version of [Something], which <etc.>"

I've also moved the netfx GetSchema methods from SqlConnectionHelper.cs to SqlConnection.cs, to align that part of it with netcore. This contributes to the merge slightly.


src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbMetaDataFactory.cs
```diff
@@ -175,13 +177,18 @@ private DataTable ExecuteCommand(DataRow requestedCollectionRow, string[] restri
                     Locale = CultureInfo.InvariantCulture
                 };
 
+                cancellationToken.ThrowIfCancellationRequested();
+#if NET
+                DataTable schemaTable = isAsync ? await reader.GetSchemaTableAsync(cancellationToken).ConfigureAwait(false) : reader.GetSchemaTable();
+#else
```
roji, line 183: Would it make sense to have the .NET Framework path also use an async API here? Unless I'm missing something, the exact same methods/code could be define for both, with the sole extension that in .NET Framework it wouldn't override?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbMetaDataFactory.cs
```diff
@@ -107,7 +109,7 @@ protected virtual void Dispose(bool disposing)
             }
         }
 
-        private DataTable ExecuteCommand(DataRow requestedCollectionRow, string[] restrictions, DbConnection connection)
+        private async ValueTask<DataTable> ExecuteCommandAsync(DataRow requestedCollectionRow, string[] restrictions, DbConnection connection, bool isAsync, CancellationToken cancellationToken)
```
roji, line 112: Really nice to be seeing proper async/await in SqlClient!

src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbMetaDataFactory.cs
edwardneal: I can do that; I thought it might look a little clumsy for the .NET Framework version of Microsoft.Data.SqlClient to have a dedicated async method which isn't present on System.Data.SqlClient and which doesn't actually run asynchronously.

(GetSchemaTable just constructs a DataTable from the cached metadata - there's nothing which we can make async.)

edwardneal: Running this through CI now...

The IDbColumnSchemaGenerator method will need to be done after (or possibly as part of) #2967.

roji: > GetSchemaTable just constructs a DataTable from the cached metadata - there's nothing which we can make async

Oh I must be misunderstanding what this is about then - I thought the idea here is to actually get the schema information (asynchronously).

In any case, the idea is to have the least amount of separate code paths here for .NET Framework vs. modern .NET...

edwardneal: I implemented this, but on re-checking, we access a property within SqlDataReader which performs a read from the network. I'm going to revert the SqlDataReader changes tomorrow so that I don't introduce async-over-sync within the new API. I'll submit the new SqlDataReader API in a subsequent PR, which will need more review since it'll involve touching the state machine.

SqlConnection.GetSchemaAsync is much simpler, since it just layers atop SqlCommand. This PR can still introduce that safely.

paulmedynski:
@edwardneal - Can you resolve the conflicts, and then I will kick off another CI run?

edwardneal:
Thanks @paulmedynski, I've just brought the branch up to date.

doc/snippets/Microsoft.Data.SqlClient/SqlConnection.xml
```diff
@@ -1363,6 +1363,22 @@ For more information on working with events, see [Connection Events](https://lea
         A <see cref="T:System.Data.DataTable" /> that contains schema information.
       </returns>
     </GetSchema2>
+    <GetSchemaAsync>
+      <param name="cancellationToken">
+        The cancellation token.
+      </param>
+      <summary>
+        An asynchronous version of <see cref="M:Microsoft.Data.SqlClient.SqlConnection.GetSchema()" />, which returns schema information for the data source of this <see cref="T:Microsoft.Data.SqlClient.SqlConnection" />. For more information about scheme, see <see href="https://learn.microsoft.com/sql/connect/ado-net/sql-server-schema-collections">SQL Server Schema Collections</see>.
```
paulmedynski: scheme -> schemas ?

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -993,6 +993,21 @@ public event Microsoft.Data.SqlClient.SqlInfoMessageEventHandler InfoMessage { a
         public override System.Data.DataTable GetSchema(string collectionName) { throw null; }
         /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnection.xml' path='docs/members[@name="SqlConnection"]/GetSchemaCollectionNameRestrictionValues/*'/>
         public override System.Data.DataTable GetSchema(string collectionName, string[] restrictionValues) { throw null; }
+#if NET
```
paulmedynski: This whole file is NET only, so is this \#if necessary?

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -993,6 +993,21 @@ public event Microsoft.Data.SqlClient.SqlInfoMessageEventHandler InfoMessage { a
...
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnection.xml' path='docs/members[@name="SqlConnection"]/GetSchemaAsync/*'/>
+        public override System.Threading.Tasks.Task<System.Data.DataTable> GetSchemaAsync(System.Threading.CancellationToken cancellationToken = default) { throw null; }
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnection.xml' path='docs/members[@name="SqlConnection"]/GetSchemaCollectionNameAsync/*'/>
+        public override System.Threading.Tasks.Task<System.Data.DataTable> GetSchemaAsync(string collectionName, System.Threading.CancellationToken cancellationToken = default) { throw null; }
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnection.xml' path='docs/members[@name="SqlConnection"]/GetSchemaCollectionNameRestrictionValuesAsync/*'/>
+        public override System.Threading.Tasks.Task<System.Data.DataTable> GetSchemaAsync(string collectionName, string[] restrictionValues, System.Threading.CancellationToken cancellationToken = default) { throw null; }
+#else
```
paulmedynski: Are we missing `GetSchemaTableAsync` ?

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/SqlSchemaInfoTest/SqlSchemaInfoTest.cs
```diff
@@ -56,6 +57,47 @@ public static void TestGetSchema(bool openTransaction)
...
+                    }
+                }
+                else
+                {
+                    dataBases = await conn.GetSchemaAsync("DATABASES");
+                }
+
+                Assert.True(dataBases.Rows.Count > 0, "At least one database is expected");
+
+                DataTable metaDataCollections = await conn.GetSchemaAsync(DbMetaDataCollectionNames.MetaDataCollections);
```
paulmedynski, line 109: Why are we only testing one of the GetSchemaAsync() methods within a transaction, but not the others?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlDataReader.cs
```diff
@@ -1638,6 +1638,31 @@ public override DataTable GetSchemaTable()
             }
         }
 
+        /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlDataReader.xml' path='docs/members[@name="SqlDataReader"]/GetSchemaTableAsync/*' />
+#if NETFRAMEWORK
+        internal Task<DataTable> GetSchemaTableAsync(CancellationToken cancellationToken = default)
```
paulmedynski: I think we're missing a test for this new method.

paulmedynski (changes requested):
A few inconsistencies and questions.

doc/snippets/Microsoft.Data.SqlClient/SqlConnection.xml
edwardneal: Thanks - corrected in GetSchemaAsync and GetSchema.

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
edwardneal: The file is also used to build the netstandard2.0 ref assemblies, which doesn't provide this member in the base class.

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
edwardneal: I've removed this method from the implementation and the ref assemblies. At the moment, we get the metadata for a result set by getting the `SqlDataReader.MetaData` property, which synchronously waits on the network. A future PR can perform the work needed to unpick the async-over-sync here and provide a fully async implementation.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/SqlSchemaInfoTest/SqlSchemaInfoTest.cs
edwardneal: It was an artifact of the existing synchronous tests. I've added that coverage to both the async and the sync tests.

paulmedynski (dismissed):

edwardneal:
I've updated this branch after the SqlConnection merges were applied, could I have a CI run please?

paulmedynski:
@edwardneal - Sigh, more conflicts.

edwardneal:
Yes - merged, thanks.

paulmedynski (dismissed):

edwardneal:
Updating to hold this open for now.

edwardneal:
This PR is not stale. I've merged main and resolved conflicts.

paulmedynski (dismissed):

edwardneal:
This PR is not stale.

edwardneal:
This PR is not stale.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlMetaDataFactory.cs
```diff
@@ -159,13 +161,20 @@ private DataRow FindMetaDataCollectionRow(string collectionName)
...
         }
 
-        public DataTable GetSchema(DbConnection connection, string collectionName, string[] restrictions)
+        public DataTable GetSchema(DbConnection connection, string collectionName, string[] restrictions) =>
+            GetSchemaCore(connection, collectionName, restrictions, isAsync: false, default).Result;
+
+        public async Task<DataTable> GetSchemaAsync(DbConnection connection, string collectionName, string[] restrictions, CancellationToken cancellationToken) =>
+            await GetSchemaCore(connection, collectionName, restrictions, isAsync: true, cancellationToken).ConfigureAwait(false);
+
+        public async ValueTask<DataTable> GetSchemaCore(DbConnection connection, string collectionName, string[] restrictions, bool isAsync, CancellationToken cancellationToken)
```
mdaigle, line 170: It may be simpler to keep these all as Task. Most of the operations are unlikely to complete synchronously and they're all bubbling up to the that top await at L168. It's possible we could get some performance benefit from ValueTask here, but I think it would also depend on the type of work that Saransh is doing in #3939 where we cache the metadata.

Then you can make this sync and directly return the tasks from L220 and L224. L168 can then directly return the task as well which will let us avoid a few state machines.

edwardneal: Stripping away the `async` keyword on `GetSchemaCore` at L170 makes sense, but this method is also called by the existing synchronous `GetSchema` method. If I make GetSchemaCore return `Task<DataTable>` directly, it'll introduce a new Task allocation on the synchronous path.

mdaigle: Fair enough. We can come back with a performance lens later when we have better benchmarking set up.

mdaigle (dismissed):

mdaigle (approved):

paulmedynski (approved):

paulmedynski:
@edwardneal - Can you rebase/merge from main?  The macOS failure is strange, and I know @benrr101 has made many build changes lately that are passing elsewhere.

edwardneal:
I've just merged - hopefully that fixes the failure.

PR #3015, opened 2024-11-18
Merge | SniNativeWrapper Interface

benrr101:
**Description**: This is part 2 of 2 for merging SNI native method wrappers. It's a little more involved than before and requires a bit of explanation. 

In the netfx version of the SNI native wrapper, the SniNativeWrapper class is truly a wrapper around three native libraries (x86, x64, and arm64). The prior code for this had a switch on each case that would check the architecture of the system and call the appropriate native library. This isn't ideal since it requires a switch statement each time a native SNI call is made. A better approach would be to have the wrapper call into a native library that is picked at static construction of the wrapper.

However, there are limitations to this approach:
1) The only difference between the three native libraries is the name of the `[DllImport]` dll. However, this must be a compile time constant. This means there must be a class for each of the native libraries.
2) In order to use `[DllImport]`, the method must be `extern`, which must be `static`. Interfaces cannot have static methods (at least not prior to net7, iirc), so in order for the library classes to implement an interface, they have to have an instance method that calls into the static, extern method.
3) The easiest way to integrate the netcore implementation into this pattern is to have a implementation of the interface that uses the netcore dll. This introduces a bit of unnecessary indirection, but I think it will be negligible.

Thus, here is the class hierarchy:
* `SNINativeMethodWrapper` [netfx], `SNINativeMethodWrapper.Windows` [netcore]
  * `ISniNativeWrapper` [netfx, netcore]
    * `SniNativeMethods` [netcore]
    * `SniNativeMethodsArm64` [netfx]
    * `SniNativeMethodsNotSupported` [netfx] - used when architecture is not supported
    * `SniNativeMethodsX64` [netfx]
    * `SniNativeMethodsX86` [netfx]

Although I can't speak in specifics, when comparing the performance of a switch per call and calling an interface which calls the method, the interface approach is significantly more performant. (sorry @David-Engel )

**Testing**: For the most part, this just moves code around and doesn't really change the functionality. Everything *should* still work, but the native SNI CI tests will confirm.


src/Microsoft.Data.SqlClient/netcore/src/Interop/SNINativeMethodWrapper.Windows.cs
```diff
@@ -34,100 +36,106 @@ internal static int SniMaxComposedSpnLength
...
+            NativeMethods.SniTerminate();
+        
+        internal static uint SNIWaitForSSLHandshakeToComplete([In] SNIHandle pConn, int dwMilliseconds, out uint pProtocolVersion) =>
+            NativeMethods.SniWaitForSslHandshakeToComplete(pConn, dwMilliseconds, out pProtocolVersion);
+        
+        internal static uint UnmanagedIsTokenRestricted([In] IntPtr token, [MarshalAs(UnmanagedType.Bool)] out bool isRestricted) =>
+            NativeMethods.SniIsTokenRestricted(token, out isRestricted);
+        
+        private static uint GetSniMaxComposedSpnLength() =>
+            NativeMethods.SniGetMaxComposedSpnLength();
```
edwardneal, line 92: This kind of situation has around a dozen `private static` methods which are only each called 1-3 times within SNINativeMethodWrapper, and which do nothing but call their corresponding methods on the `NativeMethods` variable. We could just as easily discard them and update the call sites to call `NativeMethods` directly.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/Interop/SNINativeMethodWrapper.cs
```diff
@@ -17,8 +17,15 @@ namespace Microsoft.Data.SqlClient
 {
     internal static class SNINativeMethodWrapper
     {
+        private static readonly ISniNativeMethods NativeMethods = RuntimeInformation.ProcessArchitecture switch
+        {
+            Architecture.Arm64 => new SniNativeMethodsArm64(),
+            Architecture.X64 => new SniNativeMethodsX64(),
+            Architecture.X86 => new SniNativeMethodsX86(),
+            _ => new SniNativeMethodsNotSupported(RuntimeInformation.ProcessArchitecture)
+        };
```
edwardneal, line 26: With this being the only major variation between netcore and netfx versions of the class, do you think it'd be viable to merge them in this PR?

src/Microsoft.Data.SqlClient/src/Interop/Windows/Sni/SniNativeMethodsArm64.netfx.cs
```diff
@@ -0,0 +1,319 @@
...
+#if NETFRAMEWORK
+
+using System;
+using System.Runtime.InteropServices;
+using System.Text;
+using Microsoft.Data.SqlClient;
+
+namespace Interop.Windows.Sni
+{
+    internal class SniNativeMethodsArm64 : ISniNativeMethods
```
edwardneal, line 14:
```suggestion
    internal sealed class SniNativeMethodsArm64 : ISniNativeMethods
```

A few other cases of this too.

edwardneal (commented):
I was actually wondering how we were going to approach this merge. I've only got a few comments on the way the PR handles the approach, but are there any wider plans to merge the references to the two native SNI packages?

---

> Although I can't speak in specifics, when comparing the performance of a switch per call and calling an interface which calls the method, the interface approach is significantly more performant.

That's true in this case, but the JIT will also make its presence felt. `RuntimeInformation.ProcessArchitecture` is a constant expression at runtime, so the JIT will see a switch with a constant expression and optimise it away. On .NET Core, you might not actually have a switch expression at all.

src/Microsoft.Data.SqlClient/netcore/src/Interop/SNINativeMethodWrapper.Windows.cs
benrr101: 100% agree - these are all removed in a cleanup PR. I wanted to limit the amount of changes going into this one so the later PR(s) make sense https://github.com/benrr101/SqlClient/commit/1cc5f16d8fe7018494b7db31593804427f49dc83

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/Interop/SNINativeMethodWrapper.cs
benrr101: There's a few more differences between... I could send it all out at once, I just didn't want anyone to get overwhelmed

src/Microsoft.Data.SqlClient/src/Interop/Windows/Sni/SniNativeMethodsArm64.netfx.cs
benrr101: Yep, good call. I did it later, but it makes sense to do it now.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/Interop/SNINativeMethodWrapper.cs
benrr101: Ok, I've rolled the branch forward to take the rest of the SniNativeWrapper merge (no cleanup) changes

mdaigle (approved):
LGTM

PR #3019, opened 2024-11-18
Preserve distributed transactions on pooled connection reset

mdaigle:
The current behavior only preserves transaction state for the root transaction. Enlisted connections that are reset will be placed in the transacted connection pool, but in some situations, will no longer participate in the transaction. The next connection to request the connection will expect it to be participating in the transaction, but actions on the connection will auto-commit.

This changes the reset logic to preserve transaction state for enlisted connections (part of a distributed transaction, but not root).


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -950,7 +950,7 @@ private void ResetConnection()
                 // distributed transaction - otherwise don't reset!
                 // Prepare the parser for the connection reset - the next time a trip
                 // to the server is made.
-                _parser.PrepareResetConnection(IsTransactionRoot && !IsNonPoolableTransactionRoot);
+                _parser.PrepareResetConnection(EnlistedTransaction != null && Pool != null);
```
benrr101, line 953: I don't particularly care, but is this a scenario for using the fancy pants new `is not null` ?

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/TransactionTest/DistributedTransactionTest.Windows.cs
```diff
@@ -0,0 +1,46 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Runtime.InteropServices;
+using System.Threading.Tasks;
+using System.Transactions;
+using Xunit;
+
+#if NET8_0_OR_GREATER
```
benrr101, line 11: `#if NET`?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -950,7 +950,7 @@ private void ResetConnection()
                 // distributed transaction - otherwise don't reset!
                 // Prepare the parser for the connection reset - the next time a trip
```
benrr101, line 951: I'm thinking these comments should be updated to explain the logic below?

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/TransactionTest/DistributedTransactionTest.Windows.cs
ErikEJ: Do you not settle on` !NETFRAMEWORK` ?

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -1075,9 +1075,11 @@ private void ResetConnection()
                 // distributed transaction - otherwise don't reset!
                 if (Is2000)
```
mdaigle, line 1076: Leaving Is2000 logic in place for now until we remove support for it. The fix will not apply to versions before 2000.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/TransactionTest/DistributedTransactionTest.Windows.cs
```diff
@@ -0,0 +1,174 @@
...
+            conn2.EnlistTransaction(transaction);
+
+            // Possible deadlock
+            transaction.Commit();
+        }
+#endif
+
+        [ConditionalFact(typeof(DataTestUtility), nameof(DataTestUtility.AreConnStringsSetup), Timeout = 10000)]
+
+        public void Test_EnlistedTransactionPreservedWhilePooled()
```
mdaigle, line 50: Verified that these tests currently fail when run on main against an azure db.

paulmedynski (approved):

cheenamalhotra (approved):

PR #3027, opened 2024-11-20
Update vulnerable packages for v5.2

MichelZ:
5.2 has the following vulnerable packages:
System.Formats.Asn1
System.Private.Uri
Microsoft.Extensions.Caching.Memory

This PR updates already existing packages or takes direct dependencies on packages that can't be updated otherwise (at least I wasn't able to)


MichelZ:
hmm... the 5.2 build seems to be broken and it seems it's not based entirely on YAML, so I'm not sure I can fix that.
e.g. the `ADO-UB20-SQL19` needs to change to `ADO-UB20-SQL22` it seems:
https://sqlclientdrivers.visualstudio.com/public/_build/results?buildId=101979&view=logs&j=3ad9154c-b66f-5e48-f703-44743de98f45

```
##[error]Failed to request agent. Exception Image ADO-UB20-SQL19 doesn't exist in pool ADO-CI-1ES-Pool
,##[error]The remote provider was unable to process the request.
```

@cheenamalhotra

cheenamalhotra:
We're aware of the issue, and have been working with engg teams offline to fix the `ADO-UB20-SQL19` image so it can be used and we can continue testing with SQL 19.

PR #3014, opened 2024-11-18
Cross-platform SqlColumnEncryptionCertificateStoreProvider support

edwardneal:
Wider idea: #3048.

`SqlColumnEncryptionCertificateStoreProvider` is the only built-in `SqlColumnEncryptionKeyStoreProvider` implementation which needs no significant code changes to enable cross-platform support out of the box. I've removed the Unix-specific guards on this and widened the test coverage.

In the process, I've switched the Linux/macOS tests to use SqlColumnEncryptionCertificateStoreProvider rather than the Azure Key Vault provider. This doesn't seem to have had an impact on the code coverage statistics, but it's notable nonetheless.

PR #3034, opened 2024-11-24
Tests | Remove hardcoded certificates from functional tests

edwardneal:
This is a pre-requisite for #3014. I'm splitting this prereq in half though, it's already a little unwieldy to review.

There are a variety of hardcoded certificates (most with private keys, one without) in the test code. This PR lays out some simple enough infrastructure which helps to replace them, then replaces all the references I can see in the functional tests.

Key features here:
* TDS.Servers and TDS.EndPoint have had some minor rework. I removed the hardcoded default certificate, it'll now throw an exception if one is missing and we try to use SSL connections.
* Added a `CertificateFixtureBase` class to the TestUtilities project. This base class provides the infrastructure needed to generate an ad-hoc certificate. It'll use the CertificateRequest class in .NET Core, and it has to run PowerShell in .NET Framework (as per the current manual tests.) It also cleans up the certificates it leaves behind.
* Added a `ColumnEncryptionCertificateFixture` class to TestUtilities. This is derived from the above, and it generates a set of basic certificates, adding them to the appropriate store.
* Reworked the `SqlColumnEncryptionCertificateStoreProviderShould` tests, removing three hardcoded certificates (with private keys.) Switched this over to the above fixture class.
* Started to work on `ExceptionsCertStore` to remove its references to Utility.CreateCertificate, then realised that both of its tests were special cases of the above test class and removed them. In the process, I was also able to enable the `CertificateWithNoPrivateKey` test on .NET Core.
* Switched `ExceptionsAlgorithmErrors` to the ColumnEncryptionCertificateFixture fixture class, removing the remaining reference to Utility.CreateCertificate.
* Removed Utility.CreateCertificate (which in reality, just returned the same certificate every time.)

There are other hardcoded credentials in the manual tests, but this is already a large PR at +428/-407.


cheenamalhotra (approved):

benrr101 (approved):
Let's goooo 🚢 🚀 🌕

PR #3042, opened 2024-11-26
Cleanup/Merge | Remove Reliability Section

benrr101:
**Description**: After some discussion with @saurabh500 we came to the conclusion that it's just not beneficial to have the ReliabilitySection code in the netfx project. It is only included in debug builds (which we don't even use for CI builds anyhow), and it makes massive clutter in the code base. Although I've experimented with some ideas to leave it in place but cleanup the code, simply removing it is probably the easiest solution.

**Testing**: Project still builds without it and since it provides no functionality, it won't break anything to remove it.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlDataReader.cs
+++ b/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlDataReader.cs
@@ -5640,7 +5640,6 @@ namespace Microsoft.Data.SqlClient
             {
                 if (reader.IsCommandBehavior(CommandBehavior.SequentialAccess) && reader._sharedState._dataReady)
                 {
-                    RuntimeHelpers.PrepareConstrainedRegions();
                     bool internalReadSuccess = reader.TryReadColumnInternal(context._columnIndex, readHeaderOnly: true, forStreaming: false) == TdsOperationStatus.Done;
 
                     if (internalReadSuccess)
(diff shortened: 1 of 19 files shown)


edwardneal:
This looks good to me. Are the TdsParser.ReliabilitySection struct definitions due to be removed in a future PR?

MichelZ:
These can be deleted as well in SqlDataReader:

https://github.com/benrr101/SqlClient/blob/431494fb30d153d4b7c7690a9b726d60e9dbbea9/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs#L264

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlDataReader.cs
```diff
@@ -5824,18 +5640,8 @@ private static Task<T> GetFieldValueAsyncExecute<T>(Task task, object state)
             {
                 if (reader.IsCommandBehavior(CommandBehavior.SequentialAccess) && reader._sharedState._dataReady)
                 {
-                    bool internalReadSuccess = false;
-                    TdsParser.ReliabilitySection tdsReliabilitySection = new TdsParser.ReliabilitySection();
                     RuntimeHelpers.PrepareConstrainedRegions();
```
edwardneal, line 5643: This is only useful if it's immediately followed by a try block (see the Remarks section of the [PrepareConstrainedRegions](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.runtimehelpers.prepareconstrainedregions?view=netframework-4.6.2#remarks) documentation). It can be removed.

benrr101: Good catch 👍

MichelZ:
@cheenamalhotra @mdaigle Can this be merged? That would help with #2953

paulmedynski (approved):

mdaigle (approved):

PR #3056, opened 2024-12-03
Cleanup | SNI Native Wrapper

benrr101:
**Description**: This is a followup to the SNI Native Wrapper code merge. Each commit in this PR is an atomic change, so if the PR gets too cluttered to review all up, it can be stepped through one commit at a time. Most of these changes are just renaming and moving code around. One larger change is removing a bunch of private methods that were no longer necessary after extracting the DLL imports (as mentioned by @edwardneal). 

**Testing**: There aren't really any functional changes here, so CI pass should be sufficient validation.

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlDependency.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlDependency.cs
@@ -13,6 +13,7 @@ using System.Runtime.Remoting;
 using System.Runtime.Serialization;
 using System.Runtime.Versioning;
 using System.Security.Permissions;
+using Interop.Windows.Sni;
 #endif
 using System.Text;
 using System.Threading;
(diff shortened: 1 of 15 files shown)


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlDependency.cs
```diff
@@ -470,47 +470,38 @@ private static void ObtainProcessDispatcher()
...
+                            GetSerializedObject(objRef, serializer, stream);
+                            SniNativeWrapper.SetData(stream.ToArray()); // Native will be forced to synchronize and not overwrite.
                         }
                     }
                     else
                     {
-                        SqlClientEventSource.Log.TryNotificationTraceEvent("<sc.SqlDependency.ObtainProcessDispatcher|DEP|ERR> ERROR - AppDomain.CreateInstance returned null!");
-                        throw ADP.InternalError(ADP.InternalErrorCode.SqlDependencyProcessDispatcherFailureCreateInstance);
+                        SqlClientEventSource.Log.TryNotificationTraceEvent("<sc.SqlDependency.ObtainProcessDispatcher|DEP|ERR> ERROR - ObjectHandle.Unwrap returned null!");
+                        throw ADP.InternalError(ADP.InternalErrorCode.SqlDependencyObtainProcessDispatcherFailureObjectHandle);
```
mdaigle, line 501: The net change here will result in a different error code (and trace message), but I think the change makes sense

mdaigle (approved):

benrr101:
Closing so branch can be moved under dotnet/sqlclient repo. Please see replacement #3156

PR #3060, opened 2024-12-04
Tests | Split MultipleResultsTest, clean up console logging

edwardneal:
This starts to adjust the way the SqlClient tests use the console stdout. It contains two changes.

#### MultipleResultsTest rework ####
This was one large method. It would temporarily replace the console stdout, log to it, then manually perform a diff between the stdout and a file bundled with the project.

I've modified this, splitting the large method into three tests (ExecuteNonQuery, ExecuteReader, ExecuteScalar.) These tests are designed to prove the way that SqlClient handles informational messages, exceptions and result sets when MARS is enabled.

#### Console stdout rework ####
I've noticed a few instances where standard output from the TDS servers get mixed in with the test results logged in the CI pipeline. This output isn't useful, even when the tests fail; I've disabled it in a few areas.


mdaigle (approved):

src/Microsoft.Data.SqlClient/tests/FunctionalTests/AADAuthenticationTests.cs
```diff
@@ -76,7 +76,6 @@ public void CustomActiveDirectoryProviderTest_AppClientId_DeviceFlowCallback()
 
         private Task CustomDeviceFlowCallback(DeviceCodeResult result)
```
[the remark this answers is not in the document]
edwardneal: Thanks, I've removed it. Can you trigger another CI run please?

src/Microsoft.Data.SqlClient/tests/ManualTests/ProviderAgnostic/MultipleResultsTest/MultipleResultsTest.cs
```diff
@@ -2,366 +2,154 @@
...
-                    PrintException(e.InnerException.GetType(), e.InnerException);
-                }
-                Console.Out.Flush();
-            }
-            catch (Exception f)
-            {
-                Console.WriteLine(f);
-            }
+            // ExecuteNonQuery will drain every result set, info message and exception, collating these into a single exception.
+            SqlException exNonQuery = Assert.Throws<SqlException>(() => command.ExecuteNonQuery());
```
benrr101, line 62: I'm probably going to sound super pedantic here, but it would be really nice to use this as an opportunity to clean up the tests a wee bit more. I try to write my tests in the Arrange/Act/Assert style, when possible. In this case, we're doing assert at the same time as acting, which makes it a bit difficult to see where the test action is taking place. If possible, can we extract the `command.ExecuteNonQuery()` to a variable (in nUnit, it'd be a `TestDelegate` type, but I think xUnit allows for `Action`)?

src/Microsoft.Data.SqlClient/tests/ManualTests/ProviderAgnostic/MultipleResultsTest/MultipleResultsTest.cs
```diff
@@ -2,366 +2,154 @@
...
+            Assert.Equal($"{ResultSet2_Error}\n{ResultSet4_Error}\n{ResultSet6_Error}\n{ResultSet8_Error}\n{ResultSet10_Error}\n" +
```
benrr101, line 63: Three things:
* Just to make this more readable, could we store this as a variable before the assert?
* Can we use `Environment.NewLine` instead of replacing `\r\n` with `\n`
* When defining the variable, can we put each message/error on a new line?

combine it all, it'd look like:
```
var expectedMessage =
    $"{ResultSet2_Error}${Environment.NewLine}" +
    $"{ResultSet4_Error}${Environment.NewLine}" +
    ...
```

src/Microsoft.Data.SqlClient/tests/ManualTests/ProviderAgnostic/MultipleResultsTest/MultipleResultsTest.cs
```diff
@@ -2,366 +2,154 @@
...
+            // One message following the final result set
+            Assert.True(messages.TryDequeue(out string lastMessage));
+            Assert.Empty(messages);
+            Assert.Equal(ResultSet11_Message, lastMessage);
         }
 
-        /// <summary>
-        ///  special wrapper for the text writer to replace single "\n" with "\n"
-        /// </summary>
-        private sealed class CarriageReturnLineFeedReplacer : TextWriter
```
benrr101, line 303: HOLY SMOKES what the heck is this 🗑️ ?????? 😂

benrr101 (approved):
This is awesome, I love anything that breaks up complicated tests and cleans up the console output on the tests. 🚀 

Would love to see the few comments I made addressed, but I also won't hold things up if you don't want to address them.

src/Microsoft.Data.SqlClient/tests/ManualTests/ProviderAgnostic/MultipleResultsTest/MultipleResultsTest.cs
edwardneal: I've factored it out into a separate variable, and adjusted it to use Environment.NewLine. Taking a second look at the code though, I'm not sure whether it's more readable to have the original idea of `$"{ResultSet2_Error}\n{...}"` or a `string.Join` call. What do you think?

src/Microsoft.Data.SqlClient/tests/ManualTests/ProviderAgnostic/MultipleResultsTest/MultipleResultsTest.cs
edwardneal: In the original test, I think someone was trying to account for Linux/Windows differences by replacing CRLF with LF. It's odd though - and unnecessary, since SqlException.CreateException uses `Environment.NewLine` anyway.

src/Microsoft.Data.SqlClient/tests/ManualTests/ProviderAgnostic/MultipleResultsTest/MultipleResultsTest.cs
edwardneal: I think I see what you mean, `Assert.Throws` blurs the line a little. I've separated this out.

src/Microsoft.Data.SqlClient/tests/ManualTests/ProviderAgnostic/MultipleResultsTest/MultipleResultsTest.cs
benrr101: Hmmm - hadn't thought about a String.Join(). I'll take either as long as it's using `Environment.NewLine` :) Feel free to pick whichever approach looks best to you!

edwardneal: Sounds good thanks - I'll leave it as it is now

PR #3063, opened 2024-12-05
v6.0.0 Release Notes

benrr101:
Adding v6.0.0 release notes, updating related files. The strategy employed here is to combine the release notes from preview 1, 2, and 3, combining changes that go together/cancel each other out. I feel this is important because not everyone pays attention to preview releases and should still see breaking/major changes.


ErikEJ:
Is "since preview x" adding any value? Seems not needed.

benrr101:
@ErikEJ To me it seems marginally useful to indicate which preview it came in or if it's part of none of the previews.

CHANGELOG.md
```diff
@@ -6,111 +6,86 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 # Release Notes
 
-## [Preview Release 6.0.0-preview3.24332.3] - 2024-11-27
```
mdaigle, line 9: I think we should keep the preview versions documented in here.

release-notes/6.0/6.0.0.md
```diff
@@ -0,0 +1,141 @@
...
+- [mus65](https://github.com/mus65)
+- [SimonCropp](https://github.com/SimonCropp)
+- [TrayanZapryanov](https://github.com/TrayanZapryanov)
+- [twsouthwick](https://github.com/twsouthwick)
+- [wilbit](https://github.com/wilbit)
+- [Wraith2](https://github.com/Wraith2)
+
+### Breaking Changes
+
+- Removed support for .NET Standard. [#2386](https://github.com/dotnet/SqlClient/pull/2386), since Preview 1
```
cheenamalhotra, line 26: I am also aligned with @ErikEJ that the 'since preview...' are not really needed for end users. The release notes for past previews should continue to stay in Changelog if we want to find their references. I haven't seen any other .NET product/library publish this way.

CHANGELOG.md
cheenamalhotra: The changelog doesn't need to duplicate the list of changes, if nothing's changed, it can simply say so. Only 6.0.0.md needs to provide summarized view of all changes, which will be published on GitHub release.

release-notes/6.0/6.0.0.md
```diff
@@ -0,0 +1,141 @@
...
+- Fixed `GetSchema` by excluding unsupported engines due to lack of support for `ASSEMBLYPROPERTY` function. [#2593](https://github.com/dotnet/SqlClient/pull/2593), since Preview 1
+- Fixed SSPI retry negotiation with default port in .NET. [#2559](https://github.com/dotnet/SqlClient/pull/2559), since Preview 1
+- Fixed assembly path in .NET 8.0 and `.AssemblyAttributes`. [#2550](https://github.com/dotnet/SqlClient/pull/2550), since Preview 1
+- Fixed certificate chain validation. [#2487](https://github.com/dotnet/SqlClient/pull/2487), since Preview 1
+- Fixed clone of `SqlConnection` to include `AccessTokenCallback`. [#2525](https://github.com/dotnet/SqlClient/pull/2525), since Preview 1
+- Fixed issue with `DateTimeOffset` in table-valued parameters, which was introduced in 5.2. [#2453](https://github.com/dotnet/SqlClient/pull/2453), since Preview 1
+- Fixed `ArgumentNullException` on `SqlDataRecord.GetValue` when using user-defined data type on .NET. [#2448](https://github.com/dotnet/SqlClient/pull/2448), since Preview 1
+- Fixed `SqlBuffer` and `SqlGuild` when it's null. [#2310](https://github.com/dotnet/SqlClient/pull/2310), since Preview 1
```
MichelZ, line 57: Type in SqlGuild

README.md
```diff
@@ -11,7 +11,7 @@ Microsoft.Data.SqlClient is a .NET data provider for [Microsoft SQL Server]([url
 The Microsoft.Data.SqlClient package supports the following environments:
 
 - .NET Framework 4.6.2+
-- .NET 8.0
+- .NET 8.0+
```
paulmedynski, line 14: Should we call out .NET 9.0 here explicitly?  When I see "8.0+" I think "Any future 8.x version".

release-notes/6.0/6.0.0.md
paulmedynski: Agreed.  The phrase "This update brings the below changes over the previous stable release" already makes it clear that the change summary is a delta from the previous stable release.  Knowing which preview a change happened to appear first in doesn't seem helpful, and users could easily find it by looking at the preview release notes.  Additionally, some changes in this new stable release may not have appeared in any preview.

README.md
MichelZ: You're not calling out .NET Framework 4.7 and 4.8, right?
I think 8.0+ is fair

paulmedynski (commented):
Please review my comments.

README.md
ErikEJ: .NET 8.0+ seems perfectly clear to me (any future .NETs)

benrr101: Sure we can call out 9.0 explicitly here. It's a bit confusing since right above we say 4.6.2+ which, to me, says anything greater than 4.6.2 (4.8, etc).

The other thing is even if we explicitly build for .NET 4.6.2, eg, it'll totally work with later versions. So, to me, saying "+" to everything seems a bit redundant 🤷‍♂️

release-notes/6.0/6.0.0.md
```diff
@@ -0,0 +1,141 @@
+# Release Notes
+
+## [Stable release 6.0.0] - 2024-12-09
+
+This update brings the below changes over the previous stable release:
+
+### Contributors
```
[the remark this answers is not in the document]
benrr101: For this one I was just following the pattern from previous releases. Though, I agree, I think the contributors section should go below the changes.

release-notes/6.0/6.0.0.md
```diff
@@ -0,0 +1,139 @@
...
+
+- Azure.Identity 1.11.4
+- Microsoft.Bcl.Cryptography 9.0.0
+- Microsoft.Data.SqlClient.SNI.runtime 6.0.0-preview1.24226.4
+- Microsoft.Extensions.Caching.Memory 9.0.0
+- Microsoft.IdentityModel.JsonWebTokens 7.5.0
+- Microsoft.IdentityModel.Protocols.OpenIdConnect 7.5.0
+- Microsoft.SqlServer.Server 1.0.0
+- System.Configuration.ConfigurationManager 9.0.0
+- System.Security.Cryptography.Pkcs 9.0.0
```
David-Engel, line 110: I'm concerned that we are targeting the .NET 9 version of dependencies in our .NET 8 target. Since .NET 9 is not an LTS release and will go out of support before .NET 8, we could end up caught in a situation where we can't update a dependency version to address a vulnerability because .NET 9 won't be serviced for that vulnerability. (We also want to avoid bumping major versions of dependencies within a major version of MDS.) We should try to target LTS versions of dependencies from .NET, when possible.

I didn't catch the PR that bumped these, or I would have said something earlier.

ErikEJ: No need for a .net 9 version props, but the external packages should be LTS versions, imo

benrr101: @David-Engel I agree, we should be on a stable release for SNI. The whole thing kinda depends on it :) @cheenamalhotra is there anything I need to do for that?

ErikEJ:
Is Windows x86 supported?

David-Engel:
> Is Windows x86 supported?

Yes but I would think of it more like x86 on Windows. Running x86 apps on Windows will live longer than any Windows x86-only offering. Windows 10 support is ending in 2025 and there is no 32-bit only Windows 11...

release-notes/6.0/6.0.0.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Target Platform Support
+- .NET Framework 4.6.2+ (Windows ARM64, Windows x64, Windows x86)
+- .NET 8.0, 9.0 (Linux, macOS, Windows ARM64, Windows x64, Windows x86)
```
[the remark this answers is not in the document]
benrr101: @paulmedynski Please feel free to argue here.

release-notes/6.0/6.0.0.md
```diff
@@ -0,0 +1,228 @@
...
+        /// Gets the string representation of the Json content of this <see cref="SqlJson" /> instance.
+        /// </summary>
+        public string Value { get ; }
+    }
+}
+```
+
+JSON datatype is supported by the driver for reading, writing, streaming and performing bulk copy operations.
+
+#### Introducing SqlClientDiagnostics [#2226](https://github.com/dotnet/SqlClient/pull/2226)
```
cheenamalhotra, line 68: Please update this based on latest changes from https://github.com/MicrosoftDocs/sql-docs-pr/pull/32771

release-notes/6.0/6.0.0.md
```diff
@@ -0,0 +1,228 @@
...
+### Breaking Changes
+
+- Removed support for .NET Standard. [#2386](https://github.com/dotnet/SqlClient/pull/2386)
+- Removed support for .NET 6 [#2927](https://github.com/dotnet/SqlClient/pull/2927)
+- Removed UWP (uap) references. [#2483](https://github.com/dotnet/SqlClient/pull/2483)
+- Removed SQL 2000 client-side debugging support for .NET Framework [#2981](https://github.com/dotnet/SqlClient/pull/2981), [#2940](https://github.com/dotnet/SqlClient/pull/2940)
+
+### Added
+
+#### JSON Support [#2916](https://github.com/dotnet/SqlClient/pull/2916), [#2892](https://github.com/dotnet/SqlClient/pull/2892), [#2891](https://github.com/dotnet/SqlClient/pull/2891), [#2880](https://github.com/dotnet/SqlClient/pull/2880), [#2882](https://github.com/dotnet/SqlClient/pull/2882), [#2829](https://github.com/dotnet/SqlClient/pull/2829), [#2830](https://github.com/dotnet/SqlClient/pull/2830)
```
cheenamalhotra, line 16: I would suggest we retain these features with their PR reference in list form along with other additions, before providing detailed overview..

The links to PRs makes the heading more complex to look at and lack of full list of changes in list form is not ideal for anyone landing on this page. You will also need the list for GitHub Release publishing as there we do not publish feature details in depth.

benrr101: Ok, I'll split the difference here. I removed the PR numbers from the header, but I don't want to add the items to the list of msc additions. Having the additions called out specially and then also included in the catch-all bucket of msc additions doesn't make a ton of sense. In this way, the important features are called out, the PR numbers are available, and they don't clutter up the header. 

It's fine if we don't publish the feature details in the github release, we just put an entry in the release for the called out items. 30s tops. Besides, I'm gonna be the one authoring it anyhow ...

release-notes/6.0/6.0.0.md
```diff
@@ -13,17 +13,113 @@ This update brings the below changes over the previous stable release:
...
+```csharp
+using(SqlConnection sqlConnection = new SqlConnection("Data Source=(local);Integrated Security=true;Initial Catalog=AdventureWorks;"))
+{
+    await sqlConnection.OpenAsync(SqlConnectionOverrides.OpenWithoutRetry, cancellationToken);
+}
+```
+
+Implemented in PR: [#2433](https://github.com/dotnet/SqlClient/pull/2433)
+
+#### Msc Additions
```
David-Engel, line 113:
```suggestion
#### Other Additions
```

cheenamalhotra (approved):

PR #3073, opened 2024-12-10
Include net8.0 in ExtUtilities and TestUtilities

cheenamalhotra:
Fixes Kerberos pipeline errors (tracked internally)


cheenamalhotra:
Not needed anymore

Issue #2995, opened 2024-11-09
SqlBulkCopy.WriteToServer is not working with Fabric Data Warehouse (Azure)


geiraron:
When I try to write a DataTable to the user with the WriteToServer, I receive the the following exception:

```
Unhandled exception. Microsoft.Data.SqlClient.SqlException (0x80131904): INSERT is not a supported statement type.
   at Microsoft.Data.SqlClient.SqlConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   at Microsoft.Data.SqlClient.SqlInternalConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   at Microsoft.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObject stateObj, SqlCommand command, Boolean callerHasConnectionLock, Boolean asyncClose)
   at Microsoft.Data.SqlClient.TdsParser.TryRun(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj, Boolean& dataReady)
   at Microsoft.Data.SqlClient.TdsParser.Run(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj)
   at Microsoft.Data.SqlClient.SqlBulkCopy.RunParser(BulkCopySimpleResultSet bulkCopyHandler)
   at Microsoft.Data.SqlClient.SqlBulkCopy.SubmitUpdateBulkCommand(String TDSCommand)
   at Microsoft.Data.SqlClient.SqlBulkCopy.CopyBatchesAsync(BulkCopySimpleResultSet internalResults, String updateBulkCommandText, CancellationToken cts, TaskCompletionSource`1 source)
   at Microsoft.Data.SqlClient.SqlBulkCopy.WriteToServerInternalRestContinuedAsync(BulkCopySimpleResultSet internalResults, CancellationToken cts, TaskCompletionSource`1 source)
   at Microsoft.Data.SqlClient.SqlBulkCopy.WriteToServerInternalRestAsync(CancellationToken cts, TaskCompletionSource`1 source)
   at Microsoft.Data.SqlClient.SqlBulkCopy.WriteToServerInternalAsync(CancellationToken ctoken)
[...]

PR #3084, opened 2024-12-17
Support special casing with Fabric endpoints

ErikEJ:
Support the special  casing for Fabric endpoints

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/AdapterUtil.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/AdapterUtil.cs
@@ -750,16 +750,20 @@ namespace Microsoft.Data.Common
 
         private const string ONDEMAND_PREFIX = "-ondemand";
         private const string AZURE_SYNAPSE = "-ondemand.sql.azuresynapse.";
+        private const string FABRIC_DATAWAREHOUSE = ".datawarehouse.fabric.microsoft.com";
 
         internal static bool IsAzureSynapseOnDemandEndpoint(string dataSource)
         {
-            return IsEndpoint(dataSource, ONDEMAND_PREFIX) || dataSource.Contains(AZURE_SYNAPSE);
+            return IsEndpoint(dataSource, ONDEMAND_PREFIX) 
+                || dataSource.Contains(AZURE_SYNAPSE)
+                || dataSource.Contains(FABRIC_DATAWAREHOUSE);
         }
 
         internal static readonly string[] s_azureSqlServerEndpoints = { StringsHelper.GetString(Strings.AZURESQL_GenericEndpoint),
                                                                         StringsHelper.GetString(Strings.AZURESQL_GermanEndpoint),
                                                                         StringsHelper.GetString(Strings.AZURESQL_UsGovEndpoint),
-                                                                        StringsHelper.GetString(Strings.AZURESQL_ChinaEndpoint)};
+                                                                        StringsHelper.GetString(Strings.AZURESQL_ChinaEndpoint),
+                                                                        ".database.fabric.microsoft.com"};
 
         internal static bool IsAzureSqlServerEndpoint(string dataSource)
         {
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/AdapterUtil.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/AdapterUtil.cs
@@ -754,6 +754,11 @@ namespace Microsoft.Data.Common
         private const string PBI_DATAWAREHOUSE = "datawarehouse.pbidedicated.microsoft.com";
         private const string PBI_DATAWAREHOUSE2 = ".pbidedicated.microsoft.com";
         private const string PBI_DATAWAREHOUSE3 = ".pbidedicated.windows.net";
+        private const string AZURE_SQL = ".database.windows.net";
+        private const string AZURE_SQL_GERMANY = ".database.cloudapi.de";
+        private const string AZURE_SQL_USGOV = ".database.usgovcloudapi.net";
+        private const string AZURE_SQL_CHINA = ".database.chinacloudapi.cn";
+        private const string AZURE_SQL_FABRIC = ".database.fabric.microsoft.com";
 
         internal static bool IsAzureSynapseOnDemandEndpoint(string dataSource)
         {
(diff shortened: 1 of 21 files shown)


src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/AdapterUtil.cs
```diff
@@ -750,16 +750,20 @@ internal static Version GetAssemblyVersion()
...
+                || dataSource.Contains(AZURE_SYNAPSE)
+                || dataSource.Contains(FABRIC_DATAWAREHOUSE);
         }
 
         internal static readonly string[] s_azureSqlServerEndpoints = { StringsHelper.GetString(Strings.AZURESQL_GenericEndpoint),
                                                                         StringsHelper.GetString(Strings.AZURESQL_GermanEndpoint),
                                                                         StringsHelper.GetString(Strings.AZURESQL_UsGovEndpoint),
-                                                                        StringsHelper.GetString(Strings.AZURESQL_ChinaEndpoint)};
+                                                                        StringsHelper.GetString(Strings.AZURESQL_ChinaEndpoint),
+                                                                        ".database.fabric.microsoft.com"};
```
edwardneal, line 766: Perhaps this should be a resource string?

edwardneal (commented):
Only one comment on the code style, but the checks themselves are correct.

I think there are a handful of [feature differences](https://learn.microsoft.com/en-us/fabric/database/sql/feature-comparison-sql-database-fabric) between Azure SQL and Fabric SQL which merit a test suite and documentation updates though. At a glance, the driver would be impacted by the lack of support for Always Encrypted and the BULK INSERT statement.

ErikEJ:
@edwardneal There are a number of feature differences between the many engine variations this driver supports or happen to work with - is it really the responsibility of the driver to document this, or is the page you are linking to not enough?

edwardneal:
I think the page I linked to would cover most cases, I'm thinking primarily about a once-over to make sure that the SqlClient documentation doesn't contradict it and is clear enough for people to understand what functionality the feature depends upon. The example I had in mind was `SqlBulkCopy` - it might not be obvious that the lack of BULK INSERT support blocks this feature. I think that's probably the root cause of #2995.

ErikEJ:
#2995 refers to Fabric warehouse,  which is something completely different from Fabric SQL Database

edwardneal:
Sorry, slightly delayed here!

You're right, but the root cause is similar: some of SqlClient's functionality can't be supported on a particular engine variation, because that functionality relies upon a feature the engine doesn't support (in this case, SqlBulkCopy relies upon the BULK INSERT statement, which neither Fabric Warehouse nor Fabric SQL Database support.)

My broader point is that we should make sure that the SqlClient documentation for functionality which depends on BULK INSERT or Always Encrypted doesn't claim to be compatible across all engine variations, and that specifically for SqlBulkCopy we should make it clear that the feature depends upon server-side BULK INSERT support (since that's hard to discover in the documentation, and without that insight it's hard to know what won't support the feature.) If you want to keep this PR scoped specifically to the endpoint detection, I'm happy enough to check that documentation in a separate PR.

ErikEJ:
@mdaigle any thoughts on this?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/AdapterUtil.cs
```diff
@@ -750,16 +750,20 @@ internal static Version GetAssemblyVersion()
...
 
         internal static bool IsAzureSynapseOnDemandEndpoint(string dataSource)
         {
-            return IsEndpoint(dataSource, ONDEMAND_PREFIX) || dataSource.Contains(AZURE_SYNAPSE);
+            return IsEndpoint(dataSource, ONDEMAND_PREFIX) 
+                || dataSource.Contains(AZURE_SYNAPSE)
+                || dataSource.Contains(FABRIC_DATAWAREHOUSE);
         }
 
         internal static readonly string[] s_azureSqlServerEndpoints = { StringsHelper.GetString(Strings.AZURESQL_GenericEndpoint),
```
cheenamalhotra, line 762: There are quite a few endpoints we could add here, e.g. https://learn.microsoft.com/en-us/fabric/security/fabric-allow-list-urls#data-warehouse and yes, they need not be localized.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/AdapterUtil.cs
cheenamalhotra: They are included in the resource files, but not localized, so seems unnecessary to add them to resource files.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/AdapterUtil.cs
ErikEJ: I will look into that.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/AdapterUtil.cs
ErikEJ: Ok, so no need to localize any of them??

cheenamalhotra: Yes, please move them to constants.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/AdapterUtil.cs
ErikEJ: @cheenamalhotra Are all the 7 endpoint listed in your link similar to Synapse serverless??

mdaigle (changes requested):
Thanks for cleaning up the resource files!
I just have two small requests.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/AdapterUtil.cs
cheenamalhotra: They're Fabric DW endpoints - so yes, they're serverless and synapse (data warehouse).

cheenamalhotra:
@ErikEJ do review and update docs where the current endpoints are mentioned. Thanks!

ErikEJ:
@cheenamalhotra @mdaigle  Docs updated and PR feedback addressed

ErikEJ:
@cheenamalhotra @mdaigle CI never kicked in??

ErikEJ:
I used the new resource designer in VS for the removal, please make sure to review the Designer file changes.

paulmedynski (approved):

mdaigle (approved):

ErikEJ:
I will revert the Designer changes..

ErikEJ:
@mdaigle Designer changes reverted

paulmedynski:
> @mdaigle Designer changes reverted

I liked the Designer changes - mostly fixing typos from what I could see.  Did this cause problems elsewhere?

Also, you seem to have rebased _and_ pushed 1 new commit to revert the Designer changes, but GitHub still has me marked as Approved.  I haven't reviewed the rebase nor the 1 new commit, so that's weird.

ErikEJ:
@paulmedynski The designer changes caused the build to fail. 

Lets try to avoid scope creep, I am happy to submit a new PR with the designer changes next. (And yes, it looked like mostly typos)

ErikEJ:
@paulmedynski The PR pipeline was not run 😢 - hope I did not break main 🙏

paulmedynski:
Sigh, I will monitor.  Looks good so far though.

PR #3090, opened 2024-12-30
Tests | Remove hardcoded credentials from ManualTests

edwardneal:
This PR mops up almost all of the remaining hardcoded credentials, and does three things:
* Removes references to `CertificateUtility.CreateCertificate`, which always used a hardcoded certificate.
* Forces the tests to dynamically generate keys in Azure Key Vault.
* Removes references to `CertificateUtilityWin` in `CspProviderExt`, then the class itself.

#### Removal of hardcoded credentials ####

The removal of CertificateUtility.CreateCertificate is pretty uncontroversial. However, the [`AKVTest.TestRoundTripWithAKVAndCertStoreProvider`](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/AKVTests.cs#L132-L146) test tries to round-trip encrypted column data, encrypting it with the hardcoded certificate and decrypting it with the Azure Key Vault key. I've thus treated this key as another static credential and removed it.

It's worth noting that the [`CoreCryptoTests`](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/CoreCryptoTests.cs) class loads hardcoded credentials from TCECryptoNativeBaselineRsa.txt. I've not touched this because I don't know what to do with it. It isn't testing any SqlClient-specific functionality, just .NET's ability to decrypt text which is encrypted by native code. I'm not sure whether this is necessary any more though: we already test this implicitly with the end-to-end AE tests. There's no modification required here - just a choice on whether to keep or delete the test entirely.

#### CspProviderExt tests ####

One partially-related change was to remove CertificateUtilityWin. This was only ever used by the [`CspProviderExt`](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/CspProviderExt.cs) tests, which used RSACryptoServiceProvider to generate a key in a specific CSP and encrypt/decrypt data with it. This consisted of three tests:
1. TestKeysFromCertificatesCreatedWithMultipleCryptoProviders, which looked for all CSPs with a name containing "RSA and AES", then generated a key in that CSP and tested the Always Encrypted provider with it.
2. TestRoundTripWithCSPAndCertStoreProvider, which generated a certificate and tried to round-trip between the certificate-based AE provider and the CSP-based AE provider. XUnit was already being told to skip this test.
3. TestEncryptDecryptWithCSP, which did exactly the same thing as test 1, but with a hard-coded CSP of "Microsoft Enhanced RSA and AES Cryptographic Provider".

Tests 1 and 3 are actually identical. "Microsoft Enhanced RSA and AES Cryptographic Provider" is the only CSP which fits the criteria for Test 1. I've thus eliminated test 1, and modified test 3 to run against all matching CSPs.

Of the remaining two tests, TestEncryptDecryptWithCSP would run a PowerShell script to generate a certificate, then extract the private key and use this as the CSP. It didn't need to do this, it could simply instantiate an RSACryptoServiceProvider directly and use that. I've thus eliminated the certificate generation. The round-trip test did need the certificate though, so I've just switched it to use the same certificate generation fixture as the rest of the tests. It still fails if I force it to run, but it fails with the same error as before. Once these changes were done, CertificateUtilityWin was no longer referenced and could be deleted wholesale.

The CspProviderExt/CertificateUtilityWin changes comprise around half of the changes. If it's easier to review, I can split them into a separate PR.


src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/TestFixtures/SQLSetupStrategyCspProvider.cs
```diff
@@ -0,0 +1,77 @@
...
+        public SQLSetupStrategyCspProvider(CspParameters cspParameters)
+            : base(cspParameters.ProviderName + "/" + cspParameters.KeyContainerName)
+        {
+            // Create a new instance of RSACryptoServiceProvider to generate 
+            // a new key pair.  Pass the CspParameters class to persist the  
+            // key in the container.
+            using RSACryptoServiceProvider rsaAlg = new RSACryptoServiceProvider(KeySize, cspParameters);
+            rsaAlg.PersistKeyInCsp = true;
+
+            _cspKeyParameters.Add(cspParameters);
```
benrr101, line 29: Am I right that there's only one parameter going in here? Do we need this to be a list?

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/TestTrustedMasterKeyPaths.cs
```diff
@@ -152,8 +152,7 @@ public void TestTrustedColumnEncryptionMasterKeyPathsWithMultipleServers(string
             // Add some random key paths
             foreach (char c in new char[] { 'A', 'B' })
             {
-                string tempThumbprint = new string('F', CertificateUtility.CreateCertificate().Thumbprint.Length);
-                string invalidKeyPath = string.Format(@"{0}/my/{1}", StoreLocation.CurrentUser.ToString(), tempThumbprint);
+                string invalidKeyPath = string.Format(@"{0}/my/{1}", StoreLocation.CurrentUser.ToString(), dummyThumbprint);
```
benrr101, line 155: I don't see where `c` is being used int this (or the original) - can we either have the values be useful or just make it be a for loop? (or remove it if it's not useful)

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/SqlClientCustomTokenCredential.cs
```diff
@@ -31,11 +34,11 @@ private async Task<AccessToken> AcquireTokenAsync()
         {
             // Added to reduce HttpClient calls.
             // For multi-user support, a better design can be implemented as needed.
-            if (_akvUrl != DataTestUtility.AKVUrl)
+            if (_akvUrl != AKVUrl)
```
benrr101, line 37: as far as I can tell this variable isn't being used to do anything other than indicate that the token has been acquired. If that's true, I'd recommend either making it a bool or checking `_authority`/`_resource` for `null` as the check. 

But I'm also wondering, can we store the access token itself? Or does that need to be re-acquired each time the call to this method is made? If we can store the access token, we could use `_accessToken` as the value to check here (plus we could get rid of `_authority` and `_resource`.

src/Microsoft.Data.SqlClient/tests/tools/Microsoft.Data.SqlClient.TestUtilities/Fixtures/CspCertificateFixture.cs
```diff
@@ -0,0 +1,47 @@
...
+
+using System;
+using System.Security.Cryptography;
+using System.Security.Cryptography.X509Certificates;
+
+namespace Microsoft.Data.SqlClient.TestUtilities.Fixtures
+{
+    public class CspCertificateFixture : CertificateFixtureBase
+    {
+        public X509Certificate2 CspCertificate { get; }
```
benrr101, line 13: Properties after constructor, please :)

src/Microsoft.Data.SqlClient/tests/tools/Microsoft.Data.SqlClient.TestUtilities/Fixtures/ColumnMasterKeyCertificateFixture.cs
```diff
@@ -0,0 +1,29 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Security.Cryptography.X509Certificates;
+
+namespace Microsoft.Data.SqlClient.TestUtilities.Fixtures
+{
+    public class ColumnMasterKeyCertificateFixture : CertificateFixtureBase
+    {
+        public X509Certificate2 ColumnMasterKeyCertificate { get; }
```
benrr101, line 12: Properties after constructor, please :)

benrr101 (changes requested):
Absolutely fantastic! Just a few changes I'd like to see made, but otherwise I fully support these changes!

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/SqlClientCustomTokenCredential.cs
edwardneal: I agree. AKVUrl was used all over the place for a static AKV credential, this was the last referenced location. It's now gone!

I'm not sure about the impact that caching the access token could have on test coverage - only the first test to run would ever be able to exercise the case where the AKV access token can't be retrieved. I think it'd be better to explicitly add a TokenCredential derivative which throws and run some of the AKV tests based on that, then revisit any caching.

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/TestTrustedMasterKeyPaths.cs
edwardneal: It's not - slightly odd code. I've replaced it with a collection initializer.

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/TestFixtures/SQLSetupStrategyCspProvider.cs
edwardneal: It's not used right now, but I've done this so that it's consistent with CertificateFixtureBase

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/TestFixtures/SQLSetupStrategy.cs
```diff
@@ -59,15 +59,16 @@ public class SQLSetupStrategy : IDisposable
...
 
         public SQLSetupStrategy()
+            : base(true)
         {
-            if (certificate == null)
-            {
-                certificate = CertificateUtility.CreateCertificate();
-            }
-            keyPath = string.Concat(StoreLocation.CurrentUser.ToString(), "/", StoreName.My.ToString(), "/", certificate.Thumbprint);
+            ColumnMasterKeyPath = string.Concat(StoreLocation.CurrentUser.ToString(), "/", StoreName.My.ToString(), "/", ColumnMasterKeyCertificate.Thumbprint);
```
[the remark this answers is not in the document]
edwardneal: I just renamed `keyPath` to something more understandable! I've made the column path formatting consistent across the tests now though.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/SqlClientCustomTokenCredential.cs
```diff
@@ -34,10 +31,11 @@ private async Task<AccessToken> AcquireTokenAsync()
         {
             // Added to reduce HttpClient calls.
             // For multi-user support, a better design can be implemented as needed.
-            if (_akvUrl != AKVUrl)
+            if (string.IsNullOrEmpty(_authority) || string.IsNullOrEmpty(_resource))
             {
                 using (HttpClient httpClient = new HttpClient())
                 {
+                    string AKVUrl = (new Uri(DataTestUtility.AKVBaseUri, $"/keys/{AKVKeyName}")).AbsoluteUri;
```
benrr101, line 38: There's a level of redundant parenthesis here, right?

src/Microsoft.Data.SqlClient/tests/ManualTests/Microsoft.Data.SqlClient.ManualTesting.Tests.csproj
```diff
@@ -49,6 +49,7 @@
     <Compile Include="AlwaysEncrypted\SqlBulkCopyTruncation.cs" />
     <Compile Include="AlwaysEncrypted\SqlNullValues.cs" />
     <Compile Include="AlwaysEncrypted\TestFixtures\DatabaseHelper.cs" />
+    <Compile Include="AlwaysEncrypted\TestFixtures\AEConnectionStringProviderWithCspParameters.cs" />
```
benrr101, line 52: Nit: can this be moved one line up so the files are alphabetical?

edwardneal:
Thanks - I've just resolved both comments.

Won't the AE tests only run against a branch in this repo?

benrr101:
@edwardneal ... uhhh good question. I don't think it will... :/
How about if the CI I just kicked off is successful, I can push your branch to dotnet/SqlClient and kick off a build from there to do a one-off verification.

I'm quickly getting tired of these rules that prevent us from running all our tests against external branches 😠

edwardneal:
Thanks @benrr101, that's fine by me

benrr101:
Consider this my approval ... closing and rebuilding as #3204

Rereading my original plan I have no idea what I meant... So let's just follow the pattern that was working before.

PR #3091, opened 2024-12-31
Reduce AOT binary size

edwardneal:
Contributes to #1942.

For win-x64, the pre-PR binary size is 21MB. Sizoscope indicates that these changes trim 0.9MB (~4.25%) from the final binary.

This has two small changes and one which requires explanation:
1. Replace a Hashtable in the connection string parsing with a generic Dictionary.
2. Remove a reference to Process.GetCurrentProcess, replacing it with the new Environment.ProcessId property.

In both cases, I'm just removing references to types which aren't used frequently. The Azure dependency drags the Process class in, but when that dependency's refactored out there'll be nothing else using it in the core library.

The third change removes the only reference to `DataContractJsonSerializer`, which is used to get the signing certificates for the attestation authority. I've built a standalone server running the Host Guardian Services role, and confirmed that the API endpoint returns the certificate data in text format as a JSON array. Example output from this API endpoint is thus the text string `[ 128, 0, 35, 0, ... ]`.

JsonSerializer only returns a byte[] when passed a base64-encoded string, so I've returned a List<byte> instead.

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionString.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionString.cs
@@ -3,7 +3,6 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
-using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Globalization;
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionString.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionString.cs
@@ -396,12 +396,12 @@ namespace Microsoft.Data.SqlClient
             if (_networkLibrary != null)
             { // MDAC 83525
                 string networkLibrary = _networkLibrary.Trim().ToLower(CultureInfo.InvariantCulture);
-                Hashtable netlib = NetlibMapping();
+                Dictionary<string, string> netlib = NetlibMapping();
                 if (!netlib.ContainsKey(networkLibrary))
                 {
                     throw ADP.InvalidConnectionOptionValue(KEY.Network_Library);
                 }
-                _networkLibrary = (string)netlib[networkLibrary];
+                _networkLibrary = netlib[networkLibrary];
             }
             else
             {
@@ -1101,14 +1101,14 @@ namespace Microsoft.Data.SqlClient
             }
         }
 
-        static internal Hashtable NetlibMapping()
+        static internal Dictionary<string, string> NetlibMapping()
         {
             const int NetLibCount = 8;
 
-            Hashtable hash = s_netlibMapping;
+            Dictionary<string, string> hash = s_netlibMapping;
             if (hash == null)
             {
-                hash = new Hashtable(NetLibCount)
+                hash = new Dictionary<string, string>(NetLibCount)
                 {
                     { NETLIB.TCPIP, TdsEnums.TCP },
                     { NETLIB.NamedPipes, TdsEnums.NP },
@@ -1150,7 +1150,7 @@ namespace Microsoft.Data.SqlClient
             internal const string VIA = "dbmsgnet";
         }
 
-        private static Hashtable s_netlibMapping;
+        private static Dictionary<string, string> s_netlibMapping;
 
 #if NETFRAMEWORK
         protected internal override PermissionSet CreatePermissionSet()
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/VirtualSecureModeEnclaveProvider.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/VirtualSecureModeEnclaveProvider.cs
@@ -7,8 +7,8 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Net.Http;
-using System.Runtime.Serialization.Json;
 using System.Security.Cryptography.X509Certificates;
+using System.Text.Json;
 using System.Threading;
 
 namespace Microsoft.Data.SqlClient
@@ -57,7 +57,7 @@ namespace Microsoft.Data.SqlClient
         }
 
         // Makes a web request to the provided url and returns the response as a byte[]
-        protected override byte[] MakeRequest(string url)
+        protected override List<byte> MakeRequest(string url)
         {
             Exception exception = null;
 
@@ -72,8 +72,7 @@ namespace Microsoft.Data.SqlClient
 
                     using (Stream stream = s_client.GetStreamAsync(url).ConfigureAwait(false).GetAwaiter().GetResult())
                     {
-                        var deserializer = new DataContractJsonSerializer(typeof(byte[]));
-                        return (byte[])deserializer.ReadObject(stream);
+                        return JsonSerializer.Deserialize<List<byte>>(stream);
                     }
                 }
                 catch (Exception e)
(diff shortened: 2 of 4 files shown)


Wraith2:
Is System.Text.Json an unconditional dependency on all platforms already? Can it not deseiralize to a byte[] and thus requires the change from `byte[]` to `List<byte>` ?

edwardneal:
System.Text.Json is already required by the SqlJson type, although the Azure libraries also require it. I think it's also built in to newer versions of .NET, so we're not dragging a new dependency in for this. 

I'm not completely happy with the switch to List<byte> either. S.T.J only successfully deserialises a base64 string to a byte array, a JSON array of numbers fails. I tested a couple of different approaches:
1. Deserialising to an array of ints/shorts and copying to a new byte array.
2. Write a custom JsonConverter.

A single certificate could be 2-4kb and there could be multiple of these, so I'm trying to avoid any data copies. This (along with the associated memory usage) rules approach 1 out. Approach 2 turned into something pretty similar: at the start of the JSON array we don't know how many elements it has, so we'd have to add to a List<byte> (and perform the associated copy in ToArray when returning the final value.)

Wraith2:
Try this. 

```csharp
                Utf8JsonReader reader = new Utf8JsonReader(bytes);
                using (MemoryStream stream = new MemoryStream())
                {
                    bool writing = false;
                    while (reader.Read())
                    {
                        switch (reader.TokenType)
                        {
                            case JsonTokenType.StartArray:
                                writing = true;
                                break;
                            case JsonTokenType.EndArray:
                                writing = false;
                                break;
                            case JsonTokenType.Number:
                                if (reader.TryGetByte(out byte value))
                                {
                                    stream.WriteByte(value);
                                }
                                else
                                {
                                    throw new Exception();
                                }
                                break;
                            default:
                                break;
                        }
                    }

                    if (!writing)
                    {
                        stjBytes = stream.ToArray();
                    }
                }
```

edwardneal:
It'd definitely be more idiomatic, but the closing `ToArray` will always return a copy of the buffer, which is what I'm trying to avoid. I could call GetBuffer and convert the resultant ArraySegment to a Span, but this implies returning a MemoryStream (which is a little neater than returning a List<byte>, but still not ideal because we're tied to a single implementation of Stream. At a functional level, we're still returning a wrapper for a dynamically-sized array.)

IMO, part of the reason this feels messy is because `MakeRequest` isn't a natural boundary between the base class and the child implementation. What do you think about renaming it to `GetSigningCertificateCore` and having it return the X509Certificate2Collection instead? The base class is internal, so we've got some freedom of movement for that type of change.

This would make the child implementation responsible for converting a user-provided attestation URL into a "sensible" value, and for getting the set of signing certificates for a given attestation URL. The base class would just orchestrate those two implementations and cache the result.

Wraith2:
Does this happen often enough and waste enough memory that it's worth doing crazy things to avoid an array allocation of a few KB? As far as I can see the json parser affects a very uncommon scenario around login security so it isn't running frequently (in comparison to tds packet allocations for example) and it isn't something that people will hit multiple times per second (like reading from tds packets). 

I'd say live with the relatively small and uncommon allocation, save your performance effort for places where it will affect the majority unless someone produces a specific use-case where this path is a problem.

Wraith2 (approved):

edwardneal:
Thanks, and I agree. The performance comparison sparked a bit of curiosity though, so I benchmarked it.

* Custom deserialization logic yields about a 12-15% reduction in execution time (when compared to the simple JsonSerializer method call) but triples memory usage. I imagine that's because we're reading the stream into memory first, and there's probably space to deal with that
* Pre-sizing the MemoryStream or the List<byte> used in custom deserialization logic to the normal size of a certificate reduces execution time by a further 5%
* None of the above matters. DataContractJsonSerializer allocates 20x more memory than JsonSerializer, and takes around 10x the execution time

DataContractJsonSerializer can take 1-2ms to execute, and is called directly after a network request. We can safely stick with JsonSerializer and perform the extra array copy, it'll still be faster and allocate less memory!

Wraith2 (approved):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/VirtualSecureModeEnclaveProvider.cs
```diff
@@ -72,8 +72,7 @@ protected override byte[] MakeRequest(string url)
 
                     using (Stream stream = s_client.GetStreamAsync(url).ConfigureAwait(false).GetAwaiter().GetResult())
                     {
-                        var deserializer = new DataContractJsonSerializer(typeof(byte[]));
-                        return (byte[])deserializer.ReadObject(stream);
+                        return JsonSerializer.Deserialize<List<byte>>(stream)?.ToArray();
```
paulmedynski, line 75: Interesting - the docs for JsonSerializer claim it isn't available for .NET Framework 4.6.2.  I wrote a little test program targeting 4.6.2 and it seems to exist though!

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionString.cs
```diff
@@ -396,12 +396,12 @@ internal SqlConnectionString(string connectionString) : base(connectionString, G
             if (_networkLibrary != null)
             { // MDAC 83525
                 string networkLibrary = _networkLibrary.Trim().ToLower(CultureInfo.InvariantCulture);
-                Hashtable netlib = NetlibMapping();
+                Dictionary<string, string> netlib = NetlibMapping();
```
paulmedynski, line 399: I don't think we need the `using System.Collections;` any more.

(Sadly, GitHub doesn't let me put a comment on unchanged lines.)

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/VirtualSecureModeEnclaveProviderBase.cs
```diff
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
+using System.Collections.Generic;
```
paulmedynski, line 6: This appears to be unused.

paulmedynski (changes requested):
Please see my comments in-file.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/VirtualSecureModeEnclaveProvider.cs
edwardneal: Agreed. It comes from the [System.Text.Json package](https://www.nuget.org/packages/system.text.json#supportedframeworks-body-tab), which is marked as being compatible with .NET Framework 4.6.2. The JsonSerializer.Deserialize method appears in [apisof.net](https://apisof.net/catalog/3b8078ce2208228ed09a70307d7e961d), and confirms compatibility there too.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionString.cs
edwardneal: Yes, agreed - removed

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/VirtualSecureModeEnclaveProviderBase.cs
edwardneal: Yes, agreed - removed

edwardneal:
Thanks @paulmedynski, I've removed the unused usings as noted

paulmedynski (approved):

mdaigle (approved):
Tentatively approving. I'd like to see the tests pass on #3160 before we merge. Secure enclave tests are only run on branches.

PR #3099, opened 2025-01-08
Construct client interface name with additional info

cheenamalhotra:
Appends additional information to the client interface name in the `TdsLogin` method that is sent to the server.

The Client inferface name format with this change includes:
"Microsoft SqlClient - {OS Name} {OS Version}, {Framework Description} - {Architecture}"

e.g. on Windows , "Microsoft SqlClient - Windows 10.0.26100.0, .NET 8.0.11 - X64"

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParser.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParser.cs
@@ -10,6 +10,65 @@ namespace Microsoft.Data.SqlClient
 {
     internal partial class TdsParser
     {
+        private static string? s_clientInterfaceName;
+
+        internal void InitClientInterfaceName()
+        {
+            if (string.IsNullOrEmpty(s_clientInterfaceName))
+            {
+                // Construct client interface name in format: Microsoft SqlClient - {OS}, {Platform} - {architecture}
+                // e.g. "Microsoft SqlClient - Windows 10.0.26100, .NET 8.0.11 - X64"
+                // The client interface name is limited to 128 characters
+                s_clientInterfaceName = new StringBuilder(TdsEnums.SQL_PROVIDER_NAME)
+                    .Append(" - ").Append(GetOSInfo())
+                    .Append(", ").Append(RuntimeInformation.FrameworkDescription ?? TdsEnums.UNKNOWN)
+#if TARGET_X86 || TARGET_AMD64 || TARGET_ARM || TARGET_ARM64 || TARGET_WASM || TARGET_S390X || TARGET_LOONGARCH64 || TARGET_POWERPC64 || TARGET_RISCV64
+                    // Process Architecture is different than OS Architecture,
+                    // and indicates the architecture of the process that is running the application.
+                    .Append(" - ").Append(RuntimeInformation.ProcessArchitecture)
+#else
+                    .Append(" - ").Append(TdsEnums.UNKNOWN) // Architecture is unknown
+#endif
+                    .ToString();
+
+                // Do a final check to ensure the length of the client interface name is within the limit
+                if (s_clientInterfaceName.Length >= TdsEnums.MAXLEN_CLIENTINTERFACE)
+                {
+                    s_clientInterfaceName = s_clientInterfaceName.Substring(0, TdsEnums.MAXLEN_CLIENTINTERFACE);
+                }
+            }
+        }
+
+        /// <summary>
+        /// Returns the OS information in the format: {OS Name} {OS Version}
+        /// </summary>
+        /// <returns></returns>
+        private static string GetOSInfo()
+        {
+            string osName = TdsEnums.UNKNOWN;
+            Version osVersion = Environment.OSVersion.Version;
+
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                osName = $"{nameof(OSPlatform.Windows)} {osVersion}";
+            }
+            else if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                osName = $"macOS {osVersion}";
+            }
+            else if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+            {
+                osName = $"{nameof(OSPlatform.Linux)} {osVersion}";
+            }
+#if NET
+            else if (RuntimeInformation.IsOSPlatform(OSPlatform.FreeBSD))
+            {
+                osName = $"{nameof(OSPlatform.FreeBSD)} {RuntimeEnvironment.GetSystemVersion()}";
+            }
+#endif
+            return osName;
+        }
+
         internal void ProcessSSPI(int receivedLength)
         {
             Debug.Assert(_authenticationProvider is not null);
@@ -58,6 +117,9 @@ namespace Microsoft.Data.SqlClient
             FederatedAuthenticationFeatureExtensionData fedAuthFeatureExtensionData,
             SqlConnectionEncryptOption encrypt)
         {
+            // TODO: Initialization can be moved to class ctor in future after classes are merged.
+            InitClientInterfaceName();
+
             _physicalStateObj.SetTimeoutSeconds(rec.timeout);
 
             Debug.Assert(recoverySessionData == null || (requestedFeatures & TdsEnums.FeatureExtension.SessionRecovery) != 0, "Recovery session data without session recovery feature request");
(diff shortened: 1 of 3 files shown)


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -121,11 +123,17 @@ internal void TdsLogin(
             // length in bytes
             int length = TdsEnums.SQL2005_LOG_REC_FIXED_LEN;
 
-            string clientInterfaceName = TdsEnums.SQL_PROVIDER_NAME;
+            // Construct client interface name in format: Microsoft SqlClient - {OS}, {Platform} - {architecture}
+            // e.g. "Microsoft SqlClient - Microsoft Windows 10.0.26100, .NET 8.0.11 - X64"
+            string clientInterfaceName = new StringBuilder(TdsEnums.SQL_PROVIDER_NAME)
```
saurabh500, line 128: This can be static class property. This wont need to be recomputed. It can be computed once and reused.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -121,11 +123,17 @@ internal void TdsLogin(
...
+                .Append(" - ").Append(GetOSInfo())
+                .Append(", ").Append(RuntimeInformation.FrameworkDescription ?? TdsEnums.UNKNOWN)
+                .Append(" - ").Append(RuntimeInformation.ProcessArchitecture)
+                .ToString();
+
             Debug.Assert(TdsEnums.MAXLEN_CLIENTINTERFACE >= clientInterfaceName.Length, "cchCltIntName can specify at most 128 unicode characters. See Tds spec");
```
saurabh500, line 134: @cheenamalhotra  do you think its a good idea to have any additional safeguards to avoid even accidentally spilling over `MAXLEN_CLIENTINTERFACE` ?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -212,5 +220,24 @@ internal void TdsLogin(
...
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                return $"Windows {Environment.OSVersion.Version}";
+            }
+            else if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                return $"macOS {Environment.OSVersion.Version}";
+            }
+            else if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
```
saurabh500, line 235: OSPlatform has the 4th enum of `FreeBSD` 
Should we simply add that for completeness, its another bit of information that the runtime provides. 
We could avoid putting it in the Unknown category.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -121,11 +123,17 @@ internal void TdsLogin(
             // length in bytes
             int length = TdsEnums.SQL2005_LOG_REC_FIXED_LEN;
 
-            string clientInterfaceName = TdsEnums.SQL_PROVIDER_NAME;
+            // Construct client interface name in format: Microsoft SqlClient - {OS}, {Platform} - {architecture}
+            // e.g. "Microsoft SqlClient - Microsoft Windows 10.0.26100, .NET 8.0.11 - X64"
+            string clientInterfaceName = new StringBuilder(TdsEnums.SQL_PROVIDER_NAME)
+                .Append(" - ").Append(GetOSInfo())
+                .Append(", ").Append(RuntimeInformation.FrameworkDescription ?? TdsEnums.UNKNOWN)
+                .Append(" - ").Append(RuntimeInformation.ProcessArchitecture)
```
saurabh500, line 131: I like that you have used `ProcessArchitecture` as opposed to OSArchitecture, since the process running in an x64 emulator on an Arm device, might have the mismatch of the two values. 
Might be a good idea to add the rationale as a comment.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParser.cs
saurabh500: Perhaps, get rid of some information like Framework information if we spill over? 
The problem is that if a user runs into this problem, then they cannot recover from it, with any configuration changes. So a self correctly strategy would make sense. 

There is a lot of information here, we are getting from the runtime, which we dont control. So in that slight chance that we could spill over the Max len, a guardrail should be useful.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -121,11 +123,17 @@ internal void TdsLogin(
             // length in bytes
             int length = TdsEnums.SQL2005_LOG_REC_FIXED_LEN;
 
-            string clientInterfaceName = TdsEnums.SQL_PROVIDER_NAME;
+            // Construct client interface name in format: Microsoft SqlClient - {OS}, {Platform} - {architecture}
```
edwardneal, line 126: Are there any security considerations in sending the precise OS version, instead of just the major OS version? A honeypot SQL Server instance would be able to collect these and discover the OS patch level of its clients.

The .NET SDK and CLI allow the user to [opt out](https://learn.microsoft.com/en-us/dotnet/core/tools/telemetry) of sending telemetry. Could we have something similar here - perhaps using this string as the default application name rather than the client interface name?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParser.cs
cheenamalhotra: Our fallbacks are required to be common across languages and their runtimes, FreeBSD may not be the default fallback in every other language, unless we want to design it that way.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParser.cs
saurabh500: This is a good point. 
instead of opt out flag we should err on the side of sending the least amount of information which can be significant for us. Would sending only the major version make sense?

The concern with an opt out flag is that if this information is considered sensitive then the default behavior would be considered insecure. That makes this a big problem.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParser.cs
saurabh500: I meant that add a condition to check for FreeBSD as well. If none of these are met then the fallback is "unknown "

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParser.cs
edwardneal: Potentially, yes. I'm thinking of it similarly to a browser's user agent. On Chrome, mine is `Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/<major.minor> (KHTML, like Gecko) Chrome/<major>.<minor> Safari/<major>.<minor>`. I can read that and get very basic OS information and architecture details, with more detailed information about the browser information. I'm not sure what the use case is though - I've assumed that it's a request from the Azure SQL team.

I noticed that the JDBC driver added this functionality in 12.10.0 a month or so ago (microsoft/mssql-jdbc#2561). This uses the application name though.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParser.cs
paulmedynski: FreeBSD isn't enumerated in .NET Framework.  We can #if NET around that to get the extra info when it's available.

saurabh500: @paulmedynski  are Linux and MacOS enumerated in Framework?

saurabh500: NVM. They are :)

paulmedynski: Yes - I compiled for .NET Framework 4.6.2 and it has Windows, Linux, MacOS.  .NET (8, 9) additionally have FreeBSD.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParser.cs
cheenamalhotra: We've limited to only fetching OS version and not the OSDescription as it can be very lengthy and detailed.

edwardneal: I agree; it's mainly the OS version which I'm thinking of. If an attacker has compromised a SQL Server instance, or can intercept traffic to or impersonate a SQL Server instance, they'd be able to retrieve the IP addresses and precise OS version (including a measure of detail on the patch level) for all clients. In that kind of scenario, it'd become easier for an attacker or penetration tester to pivot to the clients with version-specific attacks.

The same approach also applies to the .NET version string, although it might be more difficult - an ASP.NET Core web service connecting to a compromised SQL Server would report its .NET version, which could then be cross-referenced to published CVEs for further exploration.

I'm thinking primarily about an on-prem environment for this kind of attack because that's probably more likely to disable SSL altogether on a local network, or to leave permissions on the SQL Server's SSL certificate wide open.

I've brought up browser user agents because it's an example of the client side reporting a truncated version - it reports an OS version of 10.0, rather than one of 10.0.26100.0. If we think there's a security consideration with sending the OS version, this'd be one way to mitigate it. Using this string as the user-customisable application name also provides a way to disable it entirely (if that's a viable option for SqlClient - it seems to be for JDBC.)

paulmedynski:
Superceded by https://github.com/dotnet/SqlClient/pull/3149

PR #3103, opened 2025-01-09
[6.0] Bump SNI to 6.0.1

mdaigle:


ErikEJ:
Getting closer 🥳

JRahnama:
Nuspec file needs to be updated as well.

mdaigle:
> Nuspec file needs to be updated as well.

Thanks, I just realized the same 😄

PR #3126, opened 2025-01-17
Fix down-level SSL/TLS version warnings

edwardneal:
Contributes to #3120. I'll backport this to 6.0 if this is merged.

To verify this, I've expanded our test TDS server, adding support for using a named SSL protocol. It adds to the data for `CertificateTestWithTdsServer.BeginWindowsConnectionTest` and `BeginLinuxConnectionTest`, including an end-to-end connection test to a TDS server which only supports TLS 1.0 & 1.1.

PR #3127, opened 2025-01-21
Port #3126 to release/6.0

benrr101:
Ports #3126 to the release/6.0 branch. A couple merge conflicts were resolved. Please see the original PR for more information.

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserHelperClasses.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserHelperClasses.cs
@@ -768,18 +768,18 @@ namespace Microsoft.Data.SqlClient
                 name = "TLS 1.3";
             }*/
 #pragma warning disable CA5398 // Avoid hardcoded SslProtocols values
-            if ((protocol & SslProtocols.Tls12) == SslProtocols.Tls12)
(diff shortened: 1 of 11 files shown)


benrr101:
@edwardneal no need to make your own PR for this :)

mdaigle (approved):

benrr101:
@edwardneal Ok, there's a good chance I bungled something up with the unit tests, since there were a couple conflicts. Would you be willing to take a peek at the unit test failures? I'm looking into the issue on my end, too.

edwardneal:
These failures look odd. I've pulled your branch locally, and these tests succeed. The diff looks reasonable to me.
The operative error is within SslStream: "The client and server cannot communicate, because they do not possess a common algorithm". It looks as if TLS 1.0 and 1.1 are disabled on the build agents - which makes sense, given their age. We see the same thing on the main CI too.

I don't think we can test non-TLS 1.2 connections in CI, so I'd revert the change to ConnectionTestParametersData.cs. If that passes, I'll replicate the change in main.

benrr101:
@edwardneal, ok, let's try with the changes to the test matrix rolled back. I'm having a hard time confirming locally that it might work, I get the similar (same?) test failures when running main from before #3126 .

I have no idea how this happened but the PR to main didn't have passing CI run. So even main is broken....

edwardneal:
Thanks - once this is passing, I'll submit a PR correcting the `main` build break.

VladimirReshetnikov (approved):

PR #3130, opened 2025-01-22
Disable failing TLS 1.0/1.1 tests

edwardneal:
Rolls up change from original connection test matrix #3127. Corrects #3126, removing its associated test.
CI is currently failing on main branch - it no longer supports TLS 1.0/1.1, so the tests which require TLS 1.2 are failing. This PR removes them.

Does not require porting to release/6.0.

cc @benrr101

### DIFF
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
@@ -10,7 +10,6 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests.DataCommon
 {
     public class ConnectionTestParametersData
     {
-        private const int CASES = 30;
         private string _empty = string.Empty;
         // It was advised to store the client certificate in its own folder.
         private static readonly string s_fullPathToCer = Path.Combine(Directory.GetCurrentDirectory(), "clientcert", "localhostcert.cer");
@@ -22,7 +21,7 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests.DataCommon
 
         public static IEnumerable<object[]> GetConnectionTestParameters()
         {
-            for (int i = 0; i < CASES; i++)
+            for (int i = 0; i < Data.ConnectionTestParametersList.Count; i++)
             {
                 yield return new object[] { Data.ConnectionTestParametersList[i] };
             }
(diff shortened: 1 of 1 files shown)


benrr101 (approved):

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
```diff
@@ -4,13 +4,13 @@
...
 using System.Collections.Generic;
 using System.IO;
-using System.Security.Authentication;
 using Microsoft.SqlServer.TDS.PreLogin;
 
 namespace Microsoft.Data.SqlClient.ManualTesting.Tests.DataCommon
 {
     public class ConnectionTestParametersData
     {
+        private const int CASES = 30;
```
paulmedynski, line 13: There are 31 permutations of the ConnectionTestParameters below.

But this const isn't necessary at all - see comment on line 25.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
```diff
@@ -22,7 +22,7 @@ public class ConnectionTestParametersData
 
         public static IEnumerable<object[]> GetConnectionTestParameters()
         {
-            for (int i = 0; i < Data.ConnectionTestParametersList.Count; i++)
+            for (int i = 0; i < CASES; i++)
```
paulmedynski, line 25: Avoid the fragility of a mismatch between CASES and the actual number of permutations by using a foreach here:
```
foreach (var item in Data.ConnectionTestParameterList)
{
  yield return new object[] { item };
}
```

paulmedynski (changes requested):
See comments in-file.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/ConnectionTestParametersData.cs
edwardneal: Yes, I agree - my original PR corrected this, but we reverted the file to fix the test issue. I've reintroduced the fix.

edwardneal:
Thanks @paulmedynski, I've addressed them. Would you mind triggering the pipeline again please?

paulmedynski (approved):

benrr101:
@edwardneal I'm not 100% sure about these failures, but can you merge/rebase from main and we can try again? @David-Engel made some changes that should fix running CI pipelines from forks

edwardneal:
I've just done that. The pipeline hasn't run automatically though, so needs another trigger.

PR #3142, opened 2025-02-04
Add the ability to use a frozen dictionary for field name lookups

Wraith2:
Looking up field names is an extremely common operation when fetching data. It's faster to use ordinals directly but using ordinals can be fragile if the query changes and is harder to understand because you have to manually to the name to ordinal mapping. One of the most useful additions to this library were the `SqlDataReader.Get*(string name)` overloads which allow people to use the names directly. 

This PR attempts to speed the use of name up very slightly by changing the `Dictionary<string,int>` which is currently used to optimistically use a `FrozenDictionary<string,int>` which can 2/3 the time. This speed is difficult to measure in-situ inside SqlClient because it's not something that is going to turn up in profiles. It can be microbenched outside SqlClient context for example in this blog post: https://dev.to/moh_moh701/immutable-dictionary-and-frozen-dictionary-in-net-2978 which shows that frozen dictionary is faster. 

This PR adds an extra field to the `FieldNameLookup` class but does not increase the size of the class. Checking the layout of the class using https://github.com/SergeyTeplyakov/ObjectLayoutInspector we can see that the current layout has padding of 4 bytes, just long enough to contain the extra field needed to be the counter for an enumerator.
```
Type layout for 'FieldNameLookup'
Size: 32 bytes. Paddings: 4 bytes (%12 of empty space)
|================================================|
| Object Header (8 bytes)                        |
|------------------------------------------------|
| Method Table Ptr (8 bytes)                     |
|================================================|
|   0-7: String[] _fieldNames (8 bytes)          |
|------------------------------------------------|
|  8-15: Dictionary`2 _fieldNameLookup (8 bytes) |
|------------------------------------------------|
| 16-23: CompareInfo _compareInfo (8 bytes)      |
|------------------------------------------------|
| 24-27: Int32 _defaultLocaleID (4 bytes)        |
|------------------------------------------------|
| 28-31: padding (4 bytes)                       |
|================================================|
```
Using this extra field the `FieldNameLookup` class can implement IEnumerable and IEnumerator and become it's own enumerator for `KeyValuePair<string,int>` walking over the fields in the name array backwards (as the current builder logic does). constructing the frozen dictionary using no more allocations than are currently needed.

Is this speed increase worth special casing? Perhaps. I think that the scale of usage of this lookup means that users will see small performance increases that will add up to a useful amount. The counter argument I can see is that the increased complexity here is not worth the cost of maintenance. It's worth some discussion.

Another thing that I've noticed in this code is that we are using the default comparer for the current `Dictionary<string,int>` which will be the default comparer which uses Ordinal comparisons meaning that a simple case mismatch will fall back to a linear search. I think sql server column names are not case sensitive so we could change this explicitly use OrdinalIgnoreCase and allow simple case mismatches to succeed quickly which still preserving fallback behaviour with slower performance.

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/FieldNameLookup.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/FieldNameLookup.cs
@@ -2,6 +2,12 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+
+#if NET8_0_OR_GREATER
+using System;
+using System.Collections;
+using System.Collections.Frozen;
+#endif
 using System.Collections.Generic;
 using System.Data;
 using System.Globalization;
@@ -10,11 +16,17 @@ using Microsoft.Data.Common;
 namespace Microsoft.Data.ProviderBase
 {
     internal sealed class FieldNameLookup
+#if NET8_0_OR_GREATER
+        : IEnumerable<KeyValuePair<string,int>>, IEnumerator<KeyValuePair<string,int>>
+#endif
     {
         private readonly string[] _fieldNames;
         private readonly int _defaultLocaleID;
+#if NET8_0_OR_GREATER
+        private int _enumeratorIndex;
+#endif
 
-        private Dictionary<string, int> _fieldNameLookup;
+        private IDictionary<string, int> _fieldNameLookup;
         private CompareInfo _compareInfo;
 
         public FieldNameLookup(string[] fieldNames, int defaultLocaleID)
@@ -87,6 +99,17 @@ namespace Microsoft.Data.ProviderBase
                 _compareInfo = GetCompareInfo();
             }
 
+#if NET8_0_OR_GREATER
+            // if we have failed a lookup in the frozen dictionary then we're going
+            // to have to modify the dictionary as we do comparison sensitive lookups
+            // and since we can't modify a frozen dictionary we need to revert to a
+            // standard mutable dictionary
+            if (_fieldNameLookup is FrozenDictionary<string,int> frozen)
+            {
+                _fieldNameLookup = new Dictionary<string,int>(frozen);
+            }
+#endif
+
             for (int index = 0; index < _fieldNames.Length; index++)
             {
                 if (_compareInfo.Compare(fieldName, _fieldNames[index], compareOptions) == 0)
@@ -100,9 +123,12 @@ namespace Microsoft.Data.ProviderBase
 
         private void GenerateLookup()
         {
+#if NET8_0_OR_GREATER
+            _fieldNameLookup = this.ToFrozenDictionary();
+#else
+
             int length = _fieldNames.Length;
             Dictionary<string, int> lookup = new Dictionary<string, int>(length);
-
             // walk the field names from the end to the beginning so that if a name exists
             // multiple times the first (from beginning to end) index of it is stored
             // in the hash table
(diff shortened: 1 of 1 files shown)


roji:
FrozenDictionary is for cases where "a dictionary is created infrequently but is used frequently at run time", typically when it's created e.g. once at startup and then reused. I'm not too knowledgeable about SqlClient internals, but doesn't your PR cause a FrozenDictionary to get created for each command execution? If so, my intuition is that this would actually (considerably) regress performance.

In theory, I could see something like this making sense when Prepare() is called: that's a very explicit gesture by the user to indicate that the same command is going to be executed many times, and so it may maker sense to perform the up-front work (slow creation of the FrozenDictionary) to speed up subsequent lookups.

However, I'll point out that any application that's **this** sensitive to performance really should be using ordinal-based lookups anyway, since they should always be faster than any sort of dictionary lookup. So I'm not sure I'd explore this general direction for optimization.

> This speed is difficult to measure in-situ inside SqlClient because it's not something that is going to turn up in profiles.

You mean to say that you couldn't see a difference when benchmarking before and after this PR? Wouldn't that be a strong indication that it isn't an optimization, and is adding complexity for no value?

Wraith2:
> FrozenDictionary is for cases where "a dictionary is created infrequently but is used frequently at run time", typically when it's created e.g. once at startup and then reused. I'm not too knowledgeable about SqlClient internals, but doesn't your PR cause a FrozenDictionary to get created for each command execution? If so, my intuition is that this would actually (considerably) regress performance.

The FieldNameLookup is created once for each SqlDataReader resultset, which is typically once. It contains the field names. The dictionary lookup is only created if you try to lookup a name, so if you use pure ordinal access no dictionary is created. 
This PR does not change the frequency or conditions of creating the dictionary apart from in the case where a name is looked up and fails to be found and in that specific case a standard mutable dictionary is created from the frozen dictionary, I judge this to be an uncommon case because it requires either querying for a name known not to appear or having a name which contains a character sequence which is not copy pasted and uses a slightly different localized format for the matching value.

> However, I'll point out that any application that's **this** sensitive to performance really should be using ordinal-based lookups anyway, since they should always be faster than any sort of dictionary lookup. So I'm not sure I'd explore this general direction for optimization.


> > This speed is difficult to measure in-situ inside SqlClient because it's not something that is going to turn up in profiles.
> 
> You mean to say that you couldn't see a difference when benchmarking before and after this PR? Wouldn't that be a strong indication that it isn't an optimization, and is adding complexity for no value?

No. I didn't attempt to measure it. The time spent doing field lookups is so tiny in comparison to any network communication that it's going to look negligible. I could try to create a horrible microbenchmark that goes to extreme lengths to make it visible but it wouldn't be reflective of the way the library is used. Any such benchmark should be rejected.

This possible optimization only makes sense at a whole-server level or greater and over time. It is not a change which individuals are likely to notice. I had the idea and wanted to try it, It's entirely possible that it isn't worth taking.

roji:
> The FieldNameLookup is created once for each SqlDataReader resultset, which is typically once.

It's once per reader, but then a reader gets created for each and every command execution, and then typically used a pretty small number of times (e.g. something like once per field being read). That's definitely not the kind of scenario that FrozenDictionary was created for, AFAICT. Quoting Toub's .NET 8.0 [perf notes](https://devblogs.microsoft.com/dotnet/performance-improvements-in-net-8/#frozen-collections):

> [...] create once, use a lot. And as such, a lot of attention was paid to overheads involved in reading from the collection, but initially very little time was paid to optimizing construction time. In fact, improving construction time was initially a non-goal, with a willingness to spend as much time as was needed to eke out more throughput for reading. This makes sense if you’re focusing on long-lived services, where you’re happy to spend extra seconds once an hour or day or week to optimize something that will then be used many thousands of times per second.

The key point is that **creating** a FrozenDictionary is quite expensive - concentrating only on the read throughput misses this very important point. And once again, the proposed FrozenDictionary instances will typically get used very little, so that creation overhead wouldn't be justified.

> No. I didn't attempt to measure it. The time spent doing field lookups is so tiny in comparison to any network communication that it's going to look negligible. I could try to create a horrible microbenchmark that goes to extreme lengths to make it visible but it wouldn't be reflective of the way the library is used. Any such benchmark should be rejected.

If that's true, that sounds like something that wouldn't make sense in the driver (non-trivial increase in complexity for no demonstrable benefit). I mean, don't get me wrong - it's always good to explore and experiment, I'm just not sure I see the sense in merging something like this, is all.

Wraith2:
When EF generates code to use the data reader does it go all the way to using ordinals to access fields?

roji:
> When EF generates code to use the data reader does it go all the way to using ordinals to access fields?

Yeah. For an automated generic layer over ADO.NET - such as EF - just using ordinals seems to make the most sense... After all, we generate the SQL at the same time we generate the shaper (which is the piece of code the reads the results from the reader), and so there's really no advantage in getting the fields by their names...

Wraith2:
Closing. After consideration of Roji's feedback and further reflection and work around frozen dictionaries this is clearly not a good set of tradeoffs.

PR #3149, opened 2025-02-10
Enhance TDS Login ClientInterfaceName field with client info

paulmedynski:
Spec is here: https://microsoft-my.sharepoint.com/:w:/p/davidengel/EVybTpCmdQZGixv9r-qzXYQBl-Jp4HH1GGAE6oWxh0DqhQ?e=d0b0uP

- Added ClientInterface class that knows how to build a suitable TDS Login ClientInterfaceName value.
- Added tests for ClientInterface.
- Removed unnecessary SQL_PROVIDER_NAME constant.

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ClientInterface.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ClientInterface.cs
@@ -1,0 +1,453 @@
+using System;
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+using System.Text;
+
+#nullable enable
+
+namespace Microsoft.Data.SqlClient
+{
+    // ========================================================================
+    // This class uses runtime environment information to produce a value
+    // suitable for use in the TDS LOGIN7 Client Interface Name field.
+    //
+    // TODO(ADO.Net-33562): Add a link to the spec document.
+    //
+    internal static class ClientInterface
+    {
+        #region Properties
+        
+        // ====================================================================
+        // Properties
+
+        // The Client Interface Name, never null, never empty, and never larger
+        // than TdsEnum.MAXLEN_CLIENTINTERFACE (currently 128) characters.
+        //
+        // Format:
+        //
+        //   MS-MDS|{OS Type}|{Arch}|{OS Info}|{Runtime Info}
+        //
+        // The {OS Type} will be one of the following strings:
+        //
+        //   Windows
+        //   Linux
+        //   macOS
+        //   FreeBSD
+        //   Unknown
+        //
+        // The {Arch} will be the process architecture, either the bare metal
+        // hardware architecture or the virtualized architecture.  See
+        // System.Runtime.InteropServices.RuntimeInformation.ProcessArchitecture
+        // for possible values.  This value will never be longer than 12
+        // characters.
+        //
+        // The {OS Info} will be sourced from the the
+        // System.Runtime.InteropServices.RuntimeInformation.OSDescription
+        // value, or "Unknown" if that value is empty or all whitespace.
+        //
+        // The {Runtime Info} will be sourced from the
+        // System.Runtime.InteropServices.RuntimeInformation.FrameworkDescription
+        // value, or "Unknown" if that value is empty or all whitespace.
+        //
+        // This adheres to the TDS v37.0 spec, which specifies that the Client
+        // Interface Name has a maximum length as noted above.  If the fully
+        // formed Name length is beyond that limit, it will be truncated to the
+        // maximum with no regard for preserving certain fields or pipe ('|')
+        // delimiters.
+        //
+        // The maximum length is expected to be sufficient to accommodate the
+        // driver name, {OS Type}, and {Arch} fields, but those fields will
+        // be truncated as described above if necessary.
+        //
+        // The {OS Info} and {Runtime Info} fields will share any remaining
+        // space as evenly as possible, being truncated equally if both are
+        // longer than half of the remaining space.  If one of these fields is
+        // shorter than half of the remaining space, the other field will
+        // consume as much remaining space as possible.
+        //
+        // Any characters that are not one of the following are replaced with 
+        // underscore ('_'):
+        //
+        //   - ASCII letters ([A-za-z])
+        //   - ASCII digits ([0-9])
+        //   - Space (' ')
+        //   - Period ('.')
+        //   - Plus ('+')
+        //   - Underscore ('_')
+        //   - Hyphen ('-')
+        //
+        // All known exceptions are caught and handled by injecting the fallback
+        // value of "Unknown".  However, no effort is made to catch all
+        // exceptions, for example process-fatal memory allocation errors.
+        // 
+        public static string Name => _name;
+
+        #endregion Properties
+        
+        #region Helpers
+
+        // ====================================================================
+        // Helpers
+
+        // Static construction builds the Client Interface Name.
+        //
+        // All known exceptions are consumed.
+        //
+        static ClientInterface()
+        {
+            // The max length must not be negative.
+            //
+            // C# doesn't have compile-time type traits, so we can't confirm
+            // that MAXLEN_CLIENTINTERFACE is unsigned.  Instead, we capture
+            // it into a ushort, and let the compiler decide if that is a
+            // permitted operation.
+            //
+            ushort maxLen = TdsEnums.MAXLEN_CLIENTINTERFACE;
+
+            // Determine the OS type.
+            //
+            // This is done outside of Build() to allow tests to inject
+            // specific values.
+            //
+            string osType = Unknown;
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                osType = "Windows";
+            }
+            else if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+            {
+                osType = "Linux";
+            }
+            else if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                osType = "macOS";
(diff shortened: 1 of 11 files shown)


paulmedynski:

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ClientInterface.cs
```diff
@@ -0,0 +1,453 @@
...
+
+            // Determine the OS type.
+            //
+            // This is done outside of Build() to allow tests to inject
+            // specific values.
+            //
+            string osType = Unknown;
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                osType = "Windows";
```
ErikEJ, line 115: namof() ?

paulmedynski: These strings are part of our API promise, so I'd prefer to have them explicitly written out.

Also, nameof(OSPlatform.OSX) wouldn't produce the correct string.

mdaigle (approved):
LGTM

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ClientInterface.cs
```diff
@@ -0,0 +1,453 @@
...
+                    if (remaining >= 2)
+                    {
+                        --remaining;
+                    }
+
+                    // Will both Info fields together be too long?
+                    if (
+                        // If the addition of both lengths would overflow, then
+                        // they are definitely too long.
+                        int.MaxValue - osInfo.Length < runtimeInfo.Length
```
saurabh500, line 249: Can you please explain this? Why are you using `Max - OsInfo.Length`  <  `runtimeInfo.Length` ? For all practical purposes, osInfo will be 10 chars since its been truncated already. This `int.MaxValue - osInfo.Length < runtimeInfo.Length` condition will always be false isn't it?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ClientInterface.cs
```diff
@@ -0,0 +1,453 @@
...
+                // the pieces have been appended.
+                //
+                StringBuilder name = new StringBuilder(maxLen);
+
+                // Start with the (clean) driver name.
+                name.Append(driverName);
+                name.Append('|');
+
+                // Add the OS Type, max length 10.
+                name.Append(Trunc(Clean(osType), 10));
```
saurabh500, line 179: Magic number 10 should be extracted into a const or readonly int. OSTYPE_MAX_LEN or something on these lines.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ClientInterface.cs
```diff
@@ -0,0 +1,453 @@
...
+                name.Append('|');
+
+                // Add the Architecture, max length 12.
+                name.Append(Trunc(Clean(arch.ToString()), 12));
```
saurabh500, line 183: 12 can be extracted into a named const.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ClientInterface.cs
```diff
@@ -0,0 +1,453 @@
...
+                // fallback value.
+                return Unknown;
+            }
+        }
+
+        // Truncate the given value to the given max length, and return the
+        // result.
+        public static string Trunc(string value, ushort maxLength)
+        {
+            if (value.Length <= maxLength)
```
saurabh500, line 424: can this be shortened to `public static string Trunc(string value, ushort maxLength) => value.Length <= maxLength ? value : value.Substring(0, maxLength);`?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ClientInterface.cs
```diff
@@ -0,0 +1,453 @@
...
+            {
+                // We know this won't throw ArgumentOutOfRangeException because
+                // we've already confirmed that Length is greater than maxLen.
+                result = result.Substring(0, maxLen);
+            }
+
+            return result;
+        }
+
+        // Clean the given value of any disallowed characters, replacing them
```
saurabh500, line 333: For these comments on methods, it would be good to stick to /// kind of comments, so that we can leverage VS support of hovering over the method name and looking at the formatted comments.

saurabh500: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/xmldoc/

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ClientInterface.cs
```diff
@@ -0,0 +1,453 @@
...
+            {
+                // StringBuilder failed in an unexpected way, so use our
+                // fallback value.
+                return Unknown;
+            }
+        }
+
+        // Truncate the given value to the given max length, and return the
+        // result.
+        public static string Trunc(string value, ushort maxLength)
```
saurabh500, line 422: Will you consider renaming this to `Truncate`. 
Also xmldocs style comment would be nice

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ClientInterface.cs
```diff
@@ -0,0 +1,453 @@
+using System;
```
saurabh500, line 1: Copyright header is likely needed here. 
Not sure if we moved away from it. I think all files have these

saurabh500 (approved):
Left some comments. Functionality looks good.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ClientInterface.cs
paulmedynski: I find that syntax a bit dense, and I would still want room for the comment about the exception.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ClientInterface.cs
paulmedynski: osType has been truncated.  osInfo has not - this whole block is trying to figure out if/how to truncate osInfo and runtimeInfo.

At this point osInfo.Length could be any positive 32-bit int.

David-Engel (approved):

saurabh500 (approved):

paulmedynski:
The spec has moved here:

https://microsoft.sharepoint-df.com/:w:/t/sqldevx/EZYUHqdnoXhKoQv0kxcM1QUB2V_uTCWN9GHfrDQY2tojmA?e=tf4xco

The OS Info and RuntimeInfo fields have been simplified to have a max length of 44 Unicode characters each, with no efforts to share the remaining space.  The latest commit matches that.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ClientInterface.cs
```diff
@@ -0,0 +1,428 @@
...
+                // Start with the (clean) driver name.
+                name.Append(driverName);
+                name.Append('|');
+
+                // Add the OS Type, truncating to its max length.
+                name.Append(Truncate(Clean(osType), MaxLenOsType));
+                name.Append('|');
+
+                // Add the Architecture, truncating to its max length.
+                name.Append(Truncate(Clean(arch.ToString()), MaxLenArch));
```
cheenamalhotra, line 230: Handle arch == null?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -121,11 +121,10 @@ internal void TdsLogin(
             // length in bytes
             int length = TdsEnums.SQL2005_LOG_REC_FIXED_LEN;
 
-            string clientInterfaceName = TdsEnums.SQL_PROVIDER_NAME;
-            Debug.Assert(TdsEnums.MAXLEN_CLIENTINTERFACE >= clientInterfaceName.Length, "cchCltIntName can specify at most 128 unicode characters. See Tds spec");
+            // Obtain the client interface name.
+            string clientInterfaceName = ClientInterface.Name;
```
paulmedynski, line 125: Hmm... I was expecting an existing test to fail here, but it seems we don't have any!  I will add one.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ClientInterface.cs
paulmedynski: Architecture is an enum (value type) and cannot be null.  I get a compiler error when trying to pass a nullable value, and I can't check for null without an error or a warning.  This applies to all target frameworks.  It might be possible to call this function via reflection or some other trickery and pass a null, but I don't think we need to protect ourselves from that sort of misuse.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParser.cs
paulmedynski: New test added.

PR #3150, opened 2025-02-10
AKV Official Pipeline

benrr101:
**Description**: The current dotnet-sqlclient-official pipeline currently builds MDS and then builds the AKV provider (AKV). For the release process, this is not ideal, since building a production AKV depends on the MDS package being published to nuget. Furthermore, AKV has a different release cadence than MDS, which means version can be out of sync between AKV and MDS. Thus, the current release process is to: set variables for MDS version, run the dotnet-sqlclient-official pipeline w/o the AKV stage, publish the resulting MDS package, reset variables for AKV version, run the dotnet-sqlclient-official pipeline w/o the MDS stage, then publish the resulting AKV package. This is clumsy to say the least.

This PR seeks to resolve the issue by introducing a separate pipeline for building AKV for release. It is based on existing dotnet-sqlclient-official pipeline, but effectively rewrites all of it. The major reason for rewriting it is to decouple tasks from variables, instead relying on parameters being passed into the step templates. 

Another big change with this that the versioning is defined in the yaml instead of in the variables group. This allows the versions to be kept in source control, allowing us to build different versions on eg, release branches, without touching globally defined variables. However, this isn't the perfect solution - versioning for the pipeline is separate from versioning for IDE/CLI builds of the project. Ideally we will have a central location for version to build, but this requires more heavyweight changes to the build.proj.

This new pipeline and the associated job, variables, tasks, etc, are added to the codebase without causing breaking changes to the existing pipelines. This does clutter things up a bit, but ideally it's temporary.

Next step after this PR will be to rewrite the official MDS, following the same pattern as in this new AKV pipeline.

**Testing**:
https://dev.azure.com/SqlClientDrivers/ADO.Net/_build/results?buildId=108404&view=results - Confirmed, build passed, nuget package generated with expected contents!
https://dev.azure.com/SqlClientDrivers/ADO.Net/_build/results?buildId=108995&view=logs&j=842ba44b-d8c2-5068-fa3f-2f994e531989 - Fixes issue where nuget.org was being accessed for vulnerability data despite being blocked. Fixed by making audit conditional to local builds only.
https://dev.azure.com/SqlClientDrivers/ADO.Net/_build/results?buildId=109771&view=results Fixed all open comments as of 2/25/2025


eng/pipelines/steps/compound-publish-symbols-step.yml
```diff
@@ -0,0 +1,159 @@
...
+          SymbolExpirationInDays: 1825 # 5 years
+          SymbolServerType: 'TeamServices'
+          SymbolsArtifactName: '${{ parameters.artifactName }}'
+          SymbolsFolder: '$(ARTIFACT_PATH)/${{ parameters.referenceType }}/bin'
+          SymbolsMaximumWaitTime: 60
+          SymbolsProduct: '$(PACKAGE_NAME)'
+          SymbolsVersion: '${{ parameters.version }}'
+
+    - task: AzureCLI@2
+      displayName: 'Publish Symbols'
```
mdaigle, line 80: Assuming all this is copied from the MDS symbol publishing step?

eng/pipelines/akv-official-pipeline.yml
```diff
@@ -0,0 +1,144 @@
...
+
+        # @TODO: Reassess if these should be enabled for preview builds as well
+        globalSdl:
+            # See https://aka.ms/obpipelines/sdl
+
+            apiscan:
+                # @TODO: Needs to be run for AKV?
+                enabled: ${{ not(parameters.buildPreview) }}
+                softwareFolder: '${{ variables.apiScanDllPath }}'
+                softwareName: 'Microsoft.Data.SqlClient'      # @TODO: Should be separate for AKV?
```
David-Engel, line 65: This name is registered with APIScan so they know what service/product the run is related to. We should leave it as-is.

eng/pipelines/akv-official-pipeline.yml
```diff
@@ -0,0 +1,144 @@
...
+          ref: 'refs/heads/main'
+
+extends:
+    template: 'v2/OneBranch.${{ parameters.oneBranchType }}.CrossPlat.yml@templates'
+
+    parameters:
+        featureFlags:
+            WindowsHostVersion: 1ESWindows2022
+
+        # @TODO: Reassess if these should be enabled for preview builds as well
```
David-Engel, line 57: I would associate these with oneBranchType instead. We should be running all of these for Official builds. We want to catch issues as soon as possible, so Official builds should be scheduled to run from main frequently (daily or min weekly, even if there are no code changes as SDL requirements can and do change).

eng/pipelines/akv-official-pipeline.yml
```diff
@@ -0,0 +1,144 @@
...
+        globalSdl:
+            # See https://aka.ms/obpipelines/sdl
+
+            apiscan:
+                # @TODO: Needs to be run for AKV?
+                enabled: ${{ not(parameters.buildPreview) }}
```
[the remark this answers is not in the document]
benrr101: At this point, the `buildPreview` parameter is only used to turn off SDL tasks, just like you suggested 😅  I'll tweak the naming for it so it makes more sense.

eng/pipelines/steps/compound-publish-symbols-step.yml
benrr101: Yes, mostly, the common/templates/steps/publish-symbols-step.yml is almost identical.

Notable differences:
* Removed condition that the previous task succeeded (since this happens by default), and publishSymbols be true (uh let's just not call the task if that's the case)
* Cleaned up injecting parameters into powershell variables to a format like
  ```
  $var = "${{ parameter.something }}"
  echo "var= $var"
  ```
* Cleaned up the rest of the script a bit, specifically made it so that each step has a log for starting and a log for completed successfully

eng/pipelines/akv-official-pipeline.yml
benrr101: Thanks for verifying!

eng/pipelines/akv-official-pipeline.yml
benrr101: @mdaigle Should these checks also be run on unofficial builds? I think I prefer Malcolm's suggestion that we should just leave them on by default and have a parameter for disabling them if necessary.

As for running from main on an automated schedule, and as a bit of a tangent, right now we won't get any new code unless we sync github main to ado main. I think for having automated builds to be most effective, we should also consider automating syncing github main with ado main.

eng/pipelines/akv-official-pipeline.yml
benrr101: Can you also comment on the SBOM package name? The guidance for that is pretty sketchy these days, and there doesn't appear to be any separate service we had to onboard for it, just a bom that gets generated and dumped in the artifacts folder.

eng/pipelines/akv-official-pipeline.yml
benrr101: ... That's a good question. I think we can turn off TSA to disable reporting results. But I'm not if that's the right behavior. I think we will have the following scenarios, what do you think are the right behaviors here?

1) Manual build against release/x.y branch for release process
   Run SDL tasks as per params, break build on failures, _don't_ report?
2) Scheduled build against release/x.y branch
   Run SDL tasks, break build on failures, report?

3) Manual build against main branch
   Run SDL tasks as per params, break build on failures, report?
4) Scheduled build against main branch
   Run SDL tasks as per params, break build on failures, report?

I guess I don't quite get why we wouldn't want to log failures against release branches. With the infrequent syncing of main to ADO, and the more necessary syncing of release branches, there's ways that SDL failures could get through to a release without being logged.

Also, I'm not sure the downside of always logging the SDL failures. It might be a bit noisier, but we'd want to track it?

mdaigle: I think we should:
1. always break the build unless SDL checks are explicitly disabled
2. report for all builds except PRs

David-Engel: Agreed. In practice, I think you can say if oneBranchType == Unofficial run SDL, don't report to TSA. If oneBranchType == Official, always run SDL checks, always report to TSA. Official is limited to protected branches and if an SDL issue made it into a protected branch, we want to know about it.

A PR validation run will never be Official.

benrr101: Ok, I will update it to disable TSA on unofficial builds (which afaik automatically forces all failures to break the build)

eng/pipelines/akv-official-pipeline.yml
```diff
@@ -0,0 +1,146 @@
...
+
+            sbom:
+                enabled: ${{ parameters.runSdlTasks }}
+                packageName: 'Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider'
+                packageVersion: ${{ variables.nugetPackageVersion }}
+
+            tsa:
+                # OneBranch publishes all sdl results to TSA. If TSA is disabled all SDL tools will
+                # be forced into 'break' build mode.
+                enabled: ${{ parameters.runSdlTasks }}
```
mdaigle, line 120: note to self: expecting to see this one check build type

benrr101: Haha that's more of a note to me - I forgot that I pushed to ADO to test run it and forgot to push it here

eng/pipelines/akv-official-pipeline.yml
```diff
@@ -0,0 +1,146 @@
...
+# See the LICENSE file in the project root for more information.                #
+#################################################################################
+
+name: $(Year:YY)$(DayOfYear)$(Rev:.r)
+
+# @TODO: Add triggers and schedules
+
+parameters:
+    - name: oneBranchType
+      displayName: "OneBranch template"
```
paulmedynski, line 13: Do we need quotes on these displayName values?  If yes, then they should be consistent (double vs single).

paulmedynski (changes requested):
See comment regarding quoting of string values.

eng/pipelines/akv-official-pipeline.yml
```diff
@@ -0,0 +1,146 @@
...
+
+    parameters:
+        featureFlags:
+            WindowsHostVersion:
+                Version: 2022
+
+        globalSdl:
+            # See https://aka.ms/obpipelines/sdl
+
+            apiscan:
```
mdaigle, line 61: Curious to hear about what ended up being the issue with APIScan. Did you need to tweak anything to get it working?

eng/pipelines/jobs/build-akv-official-job.yml
```diff
@@ -0,0 +1,170 @@
...
+          type: windows
+
+      variables:
+          ob_outputDirectory: '$(ARTIFACT_PATH)'
+
+      steps:
+          - template: ../steps/script-output-environment-variables-step.yml@self
+
+          - script: |
+                echo apiScanDllPath=                 ${{ parameters.apiScanDllPath }}
```
mdaigle, line 85: I think we can simplify this using convertToJson: https://learn.microsoft.com/en-us/azure/devops/pipelines/process/expressions?view=azure-devops#converttojson

That way we also automatically pick up any parameter changes we introduce in the future.

mdaigle (approved):
Overall looks good

eng/pipelines/akv-official-pipeline.yml
benrr101: Welcome to YAML where everything is made up and the quotes don't matter!

... well most of the time it doesn't matter. Sometimes it does! So, I try to follow some rules in my own head that are totally unenforceable and I probably don't even follow them myself. Basically, wrap everything in quotes for consistency, but for generally the "first" field of an object (eg, `name` above) I don't quote.

eng/pipelines/akv-official-pipeline.yml
benrr101: For this pipeline, we don't have any apiscan issues... The stuff I was talking to you about was for the MDS pipeline I've been working on.

eng/pipelines/jobs/build-akv-official-job.yml
benrr101: aww, well yes, you're right. And that's probably a better solution. But it's so much uglier :(

benrr101: Ok, with the help of claude, I got it to 1) emit all the parameters AND 2) do it prettily!

eng/pipelines/akv-official-pipeline.yml
benrr101: Ok, I've cleaned up the quotes and quote everything that isn't the first property or obviously boolean values.

paulmedynski (approved):

PR #3155, opened 2025-02-13
[4.0] Remove signing pipeline.

mdaigle:
4.0 is out of support as of November 19, 2024. We have no plans to further service this branch. Deleting the pipeline ensures that it cannot be triggered accidentally. We can pull the pipeline back from history in the case of an emergency.


benrr101 (changes requested):
image

Please provide *some* description of what is happening and why

benrr101 (approved):

PR #3161, opened 2025-02-19
Add `async` snapshot continue capability for multipacket fields

Wraith2:
Split out from https://github.com/dotnet/SqlClient/pull/2608 per discussion detailed in https://github.com/dotnet/SqlClient/pull/2608#issuecomment-2239803005

Adds infrastructure to the async reader snapshot to allow the last packet of the snapshot to be captured as a continue point for the snapshot. On replay if a continue point is available it is used which skips re-doing the previous work improving speed.

The name and behaviour of the appcontext switch being used will need agreement and work. This continue capability requires the partial packet capability so legacy mode enabled on partial packets will force this feature to be disables. I expect there to be some discussion around the names and exact details so the current PR is in dev form, easy to see and change.

Benchmarks for reading a 20Mib string from a local sql server. As payload sizes increase and latency increases speed differences should be more noticeable. I invite and request people to do some benchmarking to make sure we're aware of the performance characteristics of this change.

```
BenchmarkDotNet v0.14.0, Windows 11 (10.0.26100.2314)
.NET SDK 9.0.200
  [Host]     : .NET 8.0.13 (8.0.1325.6609), X64 RyuJIT
  DefaultJob : .NET 8.0.13 (8.0.1325.6609), X64 RyuJIT


| Method | UseContinue | Mean      | Error     | StdDev    | Gen0      | Gen1      | Gen2      | Allocated |
|------- |------------ |----------:|----------:|----------:|----------:|----------:|----------:|----------:|
| Async  | False       | 757.51 ms | 15.053 ms | 36.642 ms | 2000.0000 | 1000.0000 |         - | 101.49 MB |
| Sync   | False       |  39.40 ms |  0.543 ms |  0.508 ms | 2000.0000 |  888.8889 |  777.7778 |  80.14 MB |
| Async  | True        |  49.45 ms |  0.901 ms |  1.376 ms | 4333.3333 | 3555.5556 | 1111.1111 | 101.51 MB |
| Sync   | True        |  40.09 ms |  0.476 ms |  0.445 ms | 2000.0000 |  888.8889 |  777.7778 |  80.14 MB |
```

/cc @MichelZ and @ErikEJ for interest


mdaigle:
> Commenter does not have sufficient privileges for PR 3161 in repo dotnet/SqlClient

Unfortunately, new security requirements restrict our ability to run CI for non-maintainers. Feel free to ping the team alias (@dotnet/sqlclientdevteam) or any of us directly and we'll do our best to trigger a pipeline run ASAP.

See also: #3152

Wraith2:
Based on the feedback provided I have added a new compatibility switch called `Switch.Microsoft.Data.SqlClient.UseCompatibilityAsyncBehaviour` which will disable the new behaviour. Is CompatProcessSni enabled the new switch will automatically return true. The new async behaviour is now enabled by default in this PR. Can someone re-run the CI please.

ErikEJ:
@Wraith2 Is this the "final" PR ?? (I lost track a few years ago)

Wraith2:
Yup, part 3 of 3 of https://github.com/dotnet/SqlClient/pull/2608 so this build should be faster stable async strings.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/LocalAppContextSwitches.cs
```diff
@@ -104,6 +112,40 @@ public static bool UseCompatibilityProcessSni
             }
         }
 
+        /// <summary>
+        /// In TdsParser the async multi-packet column value fetch behaviour is capable of
+        /// using a continue snapshot state in addition to the original replay from start
+        /// logic
```
paulmedynski, line 118: Grammar typos, here and below:

logic  -> logic.

return tru  -> return true

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
```diff
@@ -2659,6 +2695,18 @@ private sealed partial class PacketData
...
+                public int TotalSize;
+
+                internal int GetPacketDataOffset()
+                {
+                    int previous = 0;
+                    if (PrevPacket != null)
+                    {
+                        previous = PrevPacket.TotalSize;
+                    }
+                    return TotalSize - (TotalSize - previous);
```
paulmedynski, line 2707: Isn't this just `return previous;` ?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/LocalAppContextSwitches.cs
```diff
@@ -104,6 +112,40 @@ public static bool UseCompatibilityProcessSni
...
+        /// This switch disables use of the continue snapshot state. This switch will always
+        /// return tru if <see cref="UseCompatibilityProcessSni"/> is enables because the 
+        /// continue state is not stable without the multiplexer.
+        /// </summary>
+        public static bool UseCompatibilityAsyncBehaviour
+        {
+            get
+            {
+                // async continue functionality is not stable without the packet multiplexer
```
paulmedynski, line 127: This is a bit confusing.

If UseCompatibilityProcessSni is false, then UseCompatibilityAsyncBehaviour must be false.

If UseCompatibilityProcessSni is true, then UseCompatibilityAsyncBehaviour can be true or false.

Maybe this statement should be:
```
// Async continue functionality is not stable without the packet multiplexer,
// so if the multiplexer is disabled, then this setting MUST also be disabled.
if (! UseCompatibilityProcessSni)
{
    return false;
}
```

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/LocalAppContextSwitches.cs
```diff
@@ -104,6 +112,40 @@ public static bool UseCompatibilityProcessSni
             }
         }
 
+        /// <summary>
+        /// In TdsParser the async multi-packet column value fetch behaviour is capable of
+        /// using a continue snapshot state in addition to the original replay from start
+        /// logic
+        /// This switch disables use of the continue snapshot state. This switch will always
+        /// return tru if <see cref="UseCompatibilityProcessSni"/> is enables because the
```
paulmedynski, line 120: I'm confused.  Which of the following is true:
- If UseCompatibilityProcessSni is false, then UseCompatibilityAsyncBehaviour must be false.
- If UseCompatibilityProcessSni is true, then UseCompatibilityAsyncBehaviour must be true.
I think it's the first one - AsyncBehaviour isn't stable without ProcessSni.

ProcessSni is stable without AsyncBehaviour.  Is that correct?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
```diff
@@ -3006,6 +3056,7 @@ internal void CheckStack(string trace)
                 }
             }
 #endif
+            public bool ContinueEnabled => !LocalAppContextSwitches.UseCompatibilityAsyncBehaviour;
```
paulmedynski, line 3059: Is the '!' here correct?  We only enable Continue if AsyncBehaviour is disabled?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
```diff
@@ -3127,6 +3193,43 @@ internal void CaptureAsStart(TdsParserStateObject stateObj)
                 AppendPacketData(stateObj._inBuff, stateObj._inBytesRead);
             }
 
+            internal void CaptureAsContinue(TdsParserStateObject stateObj)
+            {
+                if (ContinueEnabled)
+                {
+                    Debug.Assert(_stateObj == stateObj);
```
paulmedynski, line 3200: Why are we taking stateObj as an argument if it must always be _stateObj?  Should this method take no arguments?

paulmedynski (changes requested):
Some comments for discussion, and a couple of typos.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/LocalAppContextSwitches.cs
Wraith2: UseCompatibilityAsyncBehaviour returns true if we want to use the old async behaviour. So true turns OFF continue.
UseCompatibilityAsyncBehaviour is only stable using the new ProcessSni, which is the new default behaviour. The default can be turned off by settings UseCompatibilityProcessSni to true. So if UseCompatibilityProcessSni is true then UseCompatibilityAsyncBehaviour must also return true.

So it's option 2 from your list. 
The appcontext switches containing the word Compatibility are turning things off with their true value.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
Wraith2: Yup. If we've enabled compatibility for async behaviour then we have chosen the old replay behaviour and we cannot use the continue functionality.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
Wraith2: Historic artifact of refactoring I think. There are a number of places in the codebase that do this sort of thing. I think it's there so we can check that in async contexts we're operating on the state object that we expect to be operating on. However given that we don't even run our own tests in debug mode the assertion is effectively useless and I only copied it so we have clear parity between MoveToStart and MoveToContinue.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/LocalAppContextSwitches.cs
Wraith2: fixed.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/LocalAppContextSwitches.cs
Wraith2: changed to:
```
                if (UseCompatibilityProcessSni)
                {
                    // If ProcessSni compatibility mode has been enabled then the packet
                    // multiplexer has been disabled. The new async behaviour using continue
                    // point capture is only stable if the multiplexer is enabled so we must
                    // return true to enable compatibility async behaviour using only restarts.
                    return true;
                }
```
Is that clearer?

mdaigle:
I restarted the failing jobs

Wraith2:
The jobs are failing because the manual tests are timing out. The tests are timing out because stream operations are hitting a path that I wasn't aware of (and didn't break the last two times I reimplemented this) so more work is needed.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
paulmedynski: Heh ok - added this to my (growing) hit list for maintainability improvements.  We plan to enable Debug for unit test pipelines in the near future as well.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/LocalAppContextSwitches.cs
paulmedynski: Ok phew, got it.  I was missing the part where "Compatibility" means "use the old way".

paulmedynski (approved):

Wraith2:
I've made many changes to get streams working. Most things seem to work now but it's hard to tell with how test experience is locally. Can someone on the @dotnet/sqlclientdevteam run the CI please.

Wraith2:
Wow, all green. That isn't what I was expecting but it's good. This is the continue enabled build so if anyone (@ErikEJ ?) would like to build and try the EF tests or some perf testing using that might be useful.

ErikEJ:
@Wraith2 Let me run the EF Test suite with ci package

ErikEJ:
@Wraith2 can you point me to the package?

MichelZ:
@ErikEJ This? https://sqlclientdrivers.visualstudio.com/904996cc-6198-4d39-8540-eca72bdf0b7b/_apis/build/builds/110352/artifacts?artifactName=Artifacts&api-version=7.1&%24format=zip (I believe you can just download the very first one called "Artifacts" which seems to be the .nupkg)

Wraith2:
I've made some refinements and resolved some edge cases around new asserts I have added. I also found that I had regressed the compatibility code path performance significantly so I've fixed that. 
These changes somehow make async string reads reliably slightly faster and more memory efficient than sync ones and compatibility reads a little slower than before.

| Method | UseContinue | Mean      | Error     | StdDev    | Gen0      | Gen1      | Gen2      | Allocated |
|------- |------------ |----------:|----------:|----------:|----------:|----------:|----------:|----------:|
| **Async**  | **False**       | **857.91 ms** | **17.039 ms** | **31.157 ms** | **2000.0000** | **1000.0000** |         **-** | **101.49 MB** |
| Sync   | False       |  42.35 ms |  0.806 ms |  0.990 ms | 2000.0000 |  888.8889 |  777.7778 |  80.14 MB |
| **Async**  | **True**        |  **37.45 ms** |  **0.743 ms** |  **1.431 ms** | **4285.7143** | **3142.8571** | **1000.0000** |  **62.07 MB** |
| Sync   | True        |  42.74 ms |  0.844 ms |  1.183 ms | 2000.0000 |  888.8889 |  777.7778 |  80.14 MB |

Can someone retrigger the CI please.

ErikEJ:
@Wraith2 I will postpone EF  core testing for now.

ErikEJ:
@Wraith2 Should I run against a remote server or just a local instance?

Wraith2:
Either, both, whatever you like or have available. I suppose a remote server is most likely to be representative of real usage. My tests are against a local sql server.

ErikEJ:
@mdaigle @benrr101 could someone kick the CI?

ErikEJ:
Wat??

paulmedynski:
There appear to be conflicts with dotnet:main.  I guess CI wants those fixed before it will run?

Wraith2:
I'll try to look later. rebasing this set of changes is not fun.

Wraith2:
Rebased, kick CI please.

Wraith2:
I recovered the perf for the compatibility path and kept the extra speed on the new path.

Wraith2:
Ci is slowly working through the stages. the package was built and is here https://sqlclientdrivers.visualstudio.com/904996cc-6198-4d39-8540-eca72bdf0b7b/_apis/build/builds/110519/artifacts?artifactName=Artifacts&api-version=7.1&%24format=zip

ErikEJ:
@Wraith2 So I ran the EF Core 9 tests, and then updated to the package above.

The tests run around the same amount of time as previously, but two (one) tests are failing!

I have attached the results.

These are the queries (against Northwind, you can find a .sql script with the database here in the EF Cofre repo: 

`test/EFCore.SqlServer.FunctionalTests/Northwind.sql`

```
p0=NULL (Nullable = false)
SELECT * FROM "Employees" WHERE "ReportsTo" = @p0 OR ("ReportsTo" IS NULL AND @p0 IS NULL)

```

```csharp
    [ConditionalTheory]
    [MemberData(nameof(IsAsyncData))]
    public virtual Task FromSqlRaw_queryable_with_null_parameter(bool async)
    {
        uint? reportsTo = null;

        return AssertQuery(
            async,
            ss => ((DbSet<Employee>)ss.Set<Employee>()).FromSqlRaw(
                NormalizeDelimitersInRawString(
                    // ReSharper disable once ExpressionIsAlwaysNull
                    "SELECT * FROM [Employees] WHERE [ReportsTo] = {0} OR ([ReportsTo] IS NULL AND {0} IS NULL)"), reportsTo),
            ss => ss.Set<Employee>().Where(x => x.ReportsTo == reportsTo));
    }

```

[TestResults.zip](https://github.com/user-attachments/files/19088260/TestResults.zip)

Wraith2:
Thanks, it's a real new bug. I'll investigate.

Wraith2:
Fixed. It looks like the existing tests don't cover the case of a binary non-plp column so the image column in the query needed tweaking to be able to deal with the continue state. I've added covering tests based on the test provided.

@dotnet/sqlclientdevteam can you kick the CI please.

Wraith2:
all green, artifacts here https://sqlclientdrivers.visualstudio.com/904996cc-6198-4d39-8540-eca72bdf0b7b/_apis/build/builds/110607/artifacts?artifactName=Artifacts&api-version=7.1&%24format=zip

If anyone can think of a way to get wider testing of the PR artifacts it might be interesting to try it. It's clear that the CI doesn't give coverage of all code paths and there's no reason to assume that EFCore tests cover regions that SqlClient ones don't. The more we find and fix before it reaches the unwilling public the better it will be.

ErikEJ:
@Wraith2 All EF Core 9 tests passing with latest package, and run duration around the same as with 5.1.6

ErikEJ:
@Wraith2 A way to get wider testing is to include this in a 7.0 preview

Wraith2:
All EF tests passing is good news. Unless the tests have a lot of long strings I wouldn't expect to see any large time differences. In terms of testing reading 1Mib is much the same as reading 100Mib.

It's been marked for inclusion in the next preview but I don't know how much update previews get. If previews are widely used then that will be useful.

mdaigle:
I was able to replicate your performance numbers with the new internal perf tests we're building out. Very exciting!
I'm actively working through this PR and hope to submit any comments today. Thank you!

Wraith2:
Good news. Thanks. Let me know if anything needs explaining I can either try to write it up or arrange another call to go through it.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -5556,6 +5556,12 @@ private TdsOperationStatus TryProcessColumnHeaderNoNBC(SqlMetaDataPriv col, TdsP
         {
             if (col.metaType.IsLong && !col.metaType.IsPlp)
             {
+                if (stateObj.IsSnapshotContinuing())
+                {
+                    length = (ulong)stateObj.GetSnapshotStorageLength<byte>();
+                    isNull = length == 0;
+                    return TdsOperationStatus.Done;
```
mdaigle, line 5563: In this case, we can skip because we'll only have a continuing snapshot if we already read the header?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -6502,6 +6489,48 @@ internal TdsOperationStatus TryReadSqlValue(SqlBuffer value, SqlMetaDataPriv md,
...
 
+        private TdsOperationStatus TryReadByteArrayWithContinue(TdsParserStateObject stateObj, int length, out byte[] bytes)
+        {
+            bytes = null;
+            int offset = 0;
+            byte[] temp = null;
+            (bool isAvailable, bool isStarting, bool isContinuing) = stateObj.GetSnapshotStatuses();
+            if (isAvailable)
+            {
+                if (isContinuing || isStarting)
```
mdaigle, line 6500: Lamenting the fact that we have to check the snapshot state in so many different places 😢 
Not to be addressed in this PR.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -6502,6 +6489,48 @@ internal TdsOperationStatus TryReadSqlValue(SqlBuffer value, SqlMetaDataPriv md,
...
+                {
+                    temp = stateObj.TryTakeSnapshotStorage() as byte[];
+                    Debug.Assert(bytes == null || bytes.Length == length, "stored buffer length must be null or must have been created with the correct length");
```
mdaigle, line 6503: Why do we expect the snapshot storage to have this length? It doesn't seem to me that this would be true in general.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -12934,12 +13063,21 @@ internal TdsOperationStatus TryReadPlpUnicodeChars(
...
+            totalCharsRead = (startOffsetByteCount >> 1);
+            charsLeft -= totalCharsRead;
+            offst = totalCharsRead;
+
+
             while (charsLeft > 0)
             {
+                if (partialReadInProgress)
+                {
+                    goto resumePartialRead;
```
mdaigle, line 13075: I'd prefer to avoid using goto even if it means another layer of nesting.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/TdsParserStateObject.TestHarness.cs
```diff
@@ -159,6 +159,16 @@ public static bool UseCompatibilityProcessSni
                     return (bool)switchesType.GetProperty(nameof(UseCompatibilityProcessSni), BindingFlags.Public | BindingFlags.Static).GetValue(null);
                 }
             }
+
+            public static bool UseCompatibilityAsyncBehaviour
```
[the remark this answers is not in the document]
Wraith2: Yes. Im re-using one of the files from the main project to make sure we can't get out of sync between test and real implementations but that means that various classes and namespaces have to be put in exactly the right places.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCachedBuffer.cs
```diff
@@ -37,18 +37,34 @@ private SqlCachedBuffer(List<byte[]> cachedBytes)
         /// </summary>
         internal static TdsOperationStatus TryCreate(SqlMetaDataPriv metadata, TdsParser parser, TdsParserStateObject stateObj, out SqlCachedBuffer buffer)
```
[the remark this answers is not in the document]
Wraith2: That matches what I found. It's the first time I've encountered this particular class and it was a bit surprising.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCachedBuffer.cs
```diff
@@ -37,18 +37,34 @@ private SqlCachedBuffer(List<byte[]> cachedBytes)
...
-
-            List<byte[]> cachedBytes = new();
             buffer = null;
 
+            (bool isAvailable, bool isStarting, bool isContinuing) = stateObj.GetSnapshotStatuses();
+
+            List<byte[]> cachedBytes = null;
+            if (isAvailable)
+            {
+                cachedBytes = stateObj.TryTakeSnapshotStorage() as List<byte[]>;
```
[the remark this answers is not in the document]
Wraith2: The snapshot storage can contain `byte[]`, `char[]` or `List<byte[]>` depending on the whether we're reading binary, strings or xml. I hadn't realised that there was the `List<byte[]>` option until recently. The spanshot field used to be called `_plpBuffer` but I renamed it to be more general as I found that it needed to contain multiple types.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
Wraith2: Yes, sort of.
In this case we _must_ skip because the length is only present in the first packet before we start reading data. If we aren't in the first packet in the snapshot then a read here is incorrect.
Reading the length from the storage array avoids needing to keep another field in the snapshot to store the length of non-plp fields.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
Wraith2: It's a coordination with the other part of the code which is keeping making sure that the array has been written in the need-more-data state. 

The only way to arrive at this line of code is if we are continuing and that means that we have reached at least the second packet and know that we have hit a packet transition where needs-more-data was returned, in this situation the array that we created to copy the data into while we were in the isStarting state will have been written back to the snapshot storage.

It is also logically incorrect to end an async read with a non-null _storage buffer in the snapshot. That means that you can never find a buffer that some previous operation left hanging around. An async operation must either succeed and have taken the array cleanly or the cleanup process for snapshots will have nulled the array.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
Wraith2: Why? It's a supported and useful language construct which is doing exactly what we need in this case. 
I know there are people who have ideological objections to goto but in this case it's the cleanest way to get the logic we need. Unfortunately the language definite assignment rules prevent it being used to go directly the place we really want.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCachedBuffer.cs
```diff
@@ -59,13 +75,25 @@ internal static TdsOperationStatus TryCreate(SqlMetaDataPriv metadata, TdsParser
                 }
                 do
                 {
+                    bool returnAfterAdd = false;
                     int cb = (plplength > (ulong)MaxChunkSize) ? MaxChunkSize : (int)plplength;
-                    byteArr = new byte[cb];
-                    result = stateObj.TryReadPlpBytes(ref byteArr, 0, cb, out cb);
+                    byte[] byteArr = new byte[cb];
+                    // pass false for the writeDataSizeToSnapshot parameter because we want to only take data
```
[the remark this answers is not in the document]
Wraith2: Yes. In this situation in the cached buffer each read is self contained so we don't want any of the continue logic inside the callee to kick in.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCachedBuffer.cs
```diff
@@ -59,13 +75,25 @@ internal static TdsOperationStatus TryCreate(SqlMetaDataPriv metadata, TdsParser
...
+                    // from the current packet and not try to do a continue-capable multi packet read
+                    result = stateObj.TryReadPlpBytes(ref byteArr, 0, cb, out cb, writeDataSizeToSnapshot: false, compatibilityMode: false);
                     if (result != TdsOperationStatus.Done)
                     {
-                        return result;
+                        if (result == TdsOperationStatus.NeedMoreData && isAvailable && cb == byteArr.Length)
+                        {
+                            // succeeded in getting the data but failed to find the next plp length
```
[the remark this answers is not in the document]
Wraith2: It happens at the end of each packet. We've read the bytes that we know were in the packet, we need more data but when we asked for another packet we got need-more-data back.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
Wraith2: I can appreciate the sentiment.
Consider that there was the possibility that it would need to have changes made in the relationship between the SqlDataReader and the parser which would be much more complicated. I've done my best to minimize the amount of code that needs to be changed to accommodate the new functionality. There may be future opportunities for improvement.

I found that the complexity in this PR is mostly not in the code it's in understanding the sequence of events during async callbacks. In particular the nuance between is `isAvailable`, `isStarting` and `isContinuing` flags isn't big in terms of code but matters immensely for correctness and the ability to turn off the continue capability.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
mdaigle: Mainly because, while correct code, it's unexpected to see it. It's not used within our codebase except for switch statement fall-through, so there's no existing pattern. And the general consensus is that it should be avoided except when truly required. Our code should exemplify best practices for modern c#.

Can we invert the check and move the standard behavior within?

```
if (!partialReadInProgress)
{
    charsRead = ...
}

...partial read behavior...
```

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCachedBuffer.cs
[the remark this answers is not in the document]
mdaigle: I think I had two questions mixed up in here. 

The first was why there are multiple possible types contained by the snapshot storage. Your answer makes sense on that point.

The second was why we would ever need to restore from storage in this case. Looking through the code again, there's no way to add on to a SqlCachedBuffer. The TryCreate method either returns the completed buffer (and done status) or it will be called again until it returns the done status. So there's no higher level concept of adding on the the SqlCachedBuffer itself until it's complete, which is why we may need to restore some partial state from the snapshot.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
mdaigle: Agreed, once I understood the snapshot states things mostly made sense.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
mdaigle: Gotcha, and when we create an array for the snapshot storage, we create it of the required length to fit the data as defined by the packet header, which is `length`.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
edwardneal, line 1: I've not got any comments on the wider design, but diffing the netfx and netcore SqlDataReader files has highlighted an inconsistency in the `TdsOperationStatus` handling.

https://github.com/dotnet/SqlClient/blob/cb4f56802f75fe470efb3d59564924cb12dc6857/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs#L3764-L3770

https://github.com/dotnet/SqlClient/blob/cb4f56802f75fe470efb3d59564924cb12dc6857/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlDataReader.cs#L3776-L3782

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
Wraith2: Ok, Done.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
Wraith2: Unfortunate but hard to avoid. I originally didn't want to make this set of changes twice and wanted to wait for the codebases to be merged. That was several years ago and merging still hasn't been completed. I simply gave up waiting.

Synced in the latest commit.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
```diff
@@ -1997,6 +2072,18 @@ internal TdsOperationStatus TryReadNetworkPacket()
                         stackTrace = Environment.StackTrace;
                     }
 #endif
+                    bool capturedAsContinue = false;
+                    if (_snapshotStatus == SnapshotStatus.ReplayRunning || _snapshotStatus == SnapshotStatus.ReplayStarting)
```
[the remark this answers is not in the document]
Wraith2: It's a style difference and we don't use the old style c# throughout the project. It can be upgraded with a style PR at a later date.

mdaigle (approved):

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlDataReader.cs
edwardneal: Thanks, and I can see why. I think this is the only difference - I've checked the other unmerged files, and searched codebase-wide for `result != TdsOperationStatus.Done` and didn't see any others.

Wraith2:
All Green. Fully approved. Just needs merging...

Wraith2:
Yay! 🚀 

Thanks to everyone involved for helping to get this series of PR's debugged, reviewed and finally merged after so long. 
Based on the emails conversations and branch dates I can find I started work on this set of work over the  Christmas holiday of 2020 and it took me a couple of weeks to get the main body of the work implemented and then I found the stability problem that causes the need for the packet multiplexer and it took another 5 months to track down the cause. So this work has been in progress for almost 5 years now.

brendangooden:
Really appreciate all the work that's gone into this @Wraith2 !

MichelZ:
All the applications that get a nice perf boost out there! Thanks for all the efforts and seeing it through the end

mdaigle:
From the whole @dotnet/sqlclientdevteam, thank you @Wraith2! Your dedication to tracking this issue down and getting it resolved is amazing! We're so lucky to have you involved.

mdaigle:
And thank you as well to all the others who helped with these changes along the way. It was really cool to see the community come together in support of this work!

David-Engel:
THANK YOU @Wraith2 🥳 🚀 👏 !

Love seeing this merged! If you ever find yourself in Redmond, lunch is on me! It's been a long time and I/we very much appreciate the effort you've put into addressing this issue.

Also appreciate the efforts of other community members for the reviews and testing help along the way.

PR #3163, opened 2025-02-19
Merge | LocalDBAPI

benrr101:
_This is a replacement of #3047 that will give us the complete CI_

**Description**: This is one more PR that targets merging another file between the netfx project and the netcore project. This time, it's the LocalDBAPI. This one was a bit complicated because the files were split differently between netfx and netcore. The way I approached it was to merge the operative code into the LocalDBAPI.Windows file and merge the inoperative code to the LocalDBAPI.Unix file. The Windows file will be pulled in on both netfx and netcore projects, with it being conditionally included on netcore if the build is for Windows. The Unix file will be pulled in on netcore only, being conditionally included if the build is for Unix. Otherwise, this is a fairly standard merge with netfx having extra stuff that is `#if`'d in.

please note: cleanup to remove monitor sections, etc, is coming in a later PR.

**Testing**: There miiiiight be some issues with unix build, I might not have fully vetted it on unix.


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/LocalDBAPI.Unix.cs
```diff
@@ -8,6 +8,9 @@ namespace Microsoft.Data
 {
     internal static partial class LocalDBAPI
     {
+        internal static string GetLocalDbInstanceNameFromServerName(string serverName) =>
```
mdaigle, line 11: I think this is a change in behavior, but I agree with it. Right now, we try to get the local db instance name for unix, and may return a non-null value, but later steps can fail with a PlatformNotSupportedException.

mdaigle (approved):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/LocalDBAPI.Unix.cs
benrr101: On second look, yes I think you're right that it's a behavior change. I think the one thing I'm not sure about is if this will raise a platformnotsupported exception or if it will just fail to connect. I'm going to fire this up on linux to verify.

benrr101: Can confirm, on Linux, the behavior is the same - we get the "LocalDB is not supported on this platform" exception when trying to connect.

paulmedynski (approved):

Issue #3170, opened 2025-02-20
Would be very convinent if in SqlBulkCopy, non matching column mapping exception specifies the mismatched column


JNjenga:
### Is your feature request related to a problem?

When doing a bulk copy, if a column mapping that has an invalid destination column name is specified, an exception is thrown. Specifically `InvalidOperationException` with the message `The given ColumnMapping does not match up with any column in the source or destination.`  For tables that have _a lot_ of columns it is really hard to zero in on the mismatching column.

You can repro this behavior by referencing the test case `MissingTargetColumnTest` under `SqlBulkCopyTest`

### Describe the solution you'd like

It would be nice if the exception message gave a hint on which column had an issue. The exception message could include the name of first mismatched column name.

paulmedynski:
Addressed by #3170 and #3183.

PR #3183, opened 2025-02-26
SqlBulkCopy: Explicitly state column name that mismatches(#3170)

JNjenga:
Throwing `SQL.BulkLoadNonMatchingColumnMapping` exception when column mappings mismatch does not provide enough information about where the error occurred. Throwing `SQL.BulkLoadNonMatchingColumnName` exception is more appropriate.

Fixes (#3170)

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
@@ -547,6 +547,7 @@ namespace Microsoft.Data.SqlClient
 
             HashSet<string> destColumnNames = new HashSet<string>();
 
+            Dictionary<string, bool> columnMappingStatusLookup = new Dictionary<string, bool>();
             // Loop over the metadata for each column
             _SqlMetaDataSet metaDataSet = internalResults[MetaDataResultId].MetaData;
             _sortedColumnMappings = new List<_ColumnMapping>(metaDataSet.Length);
@@ -569,9 +570,16 @@ namespace Microsoft.Data.SqlClient
                 int assocId;
                 for (assocId = 0; assocId < _localColumnMappings.Count; assocId++)
                 {
+                    if (!columnMappingStatusLookup.ContainsKey(_localColumnMappings[assocId].DestinationColumn))
+		    {
+                        columnMappingStatusLookup.Add(_localColumnMappings[assocId].DestinationColumn, false);
+		    }
+
                     if ((_localColumnMappings[assocId]._destinationColumnOrdinal == metadata.ordinal) ||
                         (UnquotedName(_localColumnMappings[assocId]._destinationColumnName) == metadata.column))
                     {
+                        columnMappingStatusLookup[_localColumnMappings[assocId].DestinationColumn] = true;
+
                         if (rejectColumn)
                         {
                             nrejected++; // Count matched columns only
@@ -703,6 +711,7 @@ namespace Microsoft.Data.SqlClient
                         break;
                     }
                 }
+
                 if (assocId == _localColumnMappings.Count)
                 {
                     // Remove metadata for unmatched columns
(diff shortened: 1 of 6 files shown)


JNjenga:
@dotnet-policy-service agree

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
```diff
@@ -713,7 +715,16 @@ private string AnalyzeTargetAndCreateUpdateBulkCommand(BulkCopySimpleResultSet i
             // All columnmappings should have matched up
             if (nmatched + nrejected != _localColumnMappings.Count)
             {
-                throw (SQL.BulkLoadNonMatchingColumnMapping());
+                int assocId;
```
paulmedynski, line 718: Simplify to foreach:
```
foreach (var columnMapping in _localColumnMappings)
{
   ...
}
```

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBulkCopyColumnMapping.cs
```diff
@@ -13,6 +13,7 @@ public sealed class SqlBulkCopyColumnMapping
         internal int _destinationColumnOrdinal;
         internal string _sourceColumnName;
         internal int _sourceColumnOrdinal;
+        internal bool _matchedOrRejected;
```
paulmedynski, line 16: Another way to achieve the same result (errors with all unmapped column names), would be to simply build the list of unmapped columns during the processing loops, and then if that list isn't empty, produce the error.  Then you wouldn't need a new boolean here, which adds to the memory footprint for all columns.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
```diff
@@ -713,7 +715,16 @@ private string AnalyzeTargetAndCreateUpdateBulkCommand(BulkCopySimpleResultSet i
...
+                for (assocId = 0; assocId < _localColumnMappings.Count; assocId++)
+                {
+                    SqlBulkCopyColumnMapping columnMapping = _localColumnMappings[assocId];
+
+                    if(!columnMapping._matchedOrRejected)
+                    {
+                        throw (SQL.BulkLoadNonMatchingColumnName(columnMapping.SourceColumn));
```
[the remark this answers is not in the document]
paulmedynski: Agreed.  The current implementation only reports on the first unmatched column.  Try something like this:
```
List<string> unmappedNames = new();
foreach (var columnMapping in _localColumMappings)
{
    if (! columnMapping._matchedOrRejected)
    {
        unmappedNames.Add(columnMapping.SourceColumn);
    }
}
if (unmappedNames.Count > 0)
{
    throw SQL.BulkLoadNonMatchingColumnName(string.Join(',', unmappedNames));
}
```
As suggested below, you could build the unmappedNames list up during the processing loop, and avoid the _matchedOrRejected bool entirely.

paulmedynski, line 725: Does changing from `SQL.BulkLoadNonMatchingColumnMapping()` to `SQL.BulkLoadNonMatchingColumnName()` affect our API promises?  They both end up throwing InvalidOperationException, so probably not.  Callers shouldn't be depending on the message of the exception, but you never know!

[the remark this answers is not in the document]
paulmedynski: It may even be worth adding a SQL.BulkLoadNonMatchingColumnNames() helper that just takes the IEnumerable<string>.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/SqlBulkCopyTest/MissingTargetColumn.cs
```diff
@@ -35,7 +35,9 @@ public static void Test(string srcConstr, string dstConstr, string dstTable)
                             ColumnMappings.Add("LastName", "col2"); // this column does not exist
                             ColumnMappings.Add("FirstName", "col3");
 
-                            string errorMsg = SystemDataResourceManager.Instance.SQL_BulkLoadNonMatchingColumnMapping;
+                            string errorMsg = SystemDataResourceManager.Instance.SQL_BulkLoadNonMatchingColumnName;
+                            errorMsg = string.Format(errorMsg, "LastName");
+
```
paulmedynski, line 40: Let's add a test when multiple columns are unmapped.

paulmedynski (changes requested):
Great idea overall, with a few improvements noted.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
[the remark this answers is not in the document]
JNjenga: Thank you for the feedback. I'll look into the requested changes.

paulmedynski (approved):

paulmedynski:
@JNjenga - Thanks for the changes!

PR #3184, opened 2025-02-26
Allow sni override for testing

mdaigle:
When we make changes to SNI, it's important to test the changes in combination with MDS to avoid regressions. This PR adds the ability to trigger a pipeline run with an overridden SNI version pulled from the internal validation feed in ADO. This functionality is only available to pipeline runs within the ADO.Net project in ADO. Pipeline runs within the public project should not have access to the internal feed and should fail.

NOTE: default values are currently included to prove that runs fail within the public project and will be removed before merging.


eng/pipelines/common/templates/steps/override-sni-version.yml
```diff
@@ -0,0 +1,99 @@
...
+- name: SNIValidationFeed
+  type: string
+
+steps:
+- task: PowerShell@2
+  displayName: Update SNI Version in Nuget.config
+  inputs:
+    targetType: inline
+    script: |
+      #Purpose:  Updates Nuget.config to add SNIValidation Internal Feed and the required credentials
```
paulmedynski, line 20: Missing space between '#' and first character, here and below.

eng/pipelines/common/templates/steps/override-sni-version.yml
```diff
@@ -0,0 +1,99 @@
...
+
+      #Get the SNI Version to test from the user entered version.
+      $SharedSNIVersion = "${{parameters.SNIVersion}}"
+
+      # define file to update
+      $PropsPath = Join-Path -Path '.' -ChildPath 'tools\props\Versions.props'
+      type $PropsPath
+
+      # new version number to update to
+      ##Write-Host "SNI Version to test = $(SNIValidationVersion)"
```
paulmedynski, line 69: Remove commented-out code ?

eng/pipelines/common/templates/steps/override-sni-version.yml
```diff
@@ -0,0 +1,99 @@
...
+      Write-Host "SNI Version to test = $SharedSNIVersion"
+
+
```
paulmedynski, line 72: Extra blank line.

paulmedynski (changes requested):
Looks good, just a few nitpicks.

cheenamalhotra (approved):

eng/pipelines/dotnet-sqlclient-ci-core.yml
```diff
@@ -41,6 +41,18 @@ parameters:
...
 
+- name: SNIVersion
+  displayName: |
+    Use a specific SNI version from the internal validation feed
+  type: string
+  default: ''
+
+- name: SNIValidationFeed
+  displayName: |
+    The specific SNI test version is pulled from this feed
```
benrr101, line 52: This isn't a great display name for the parameter... I'd go for something like "SNI NuGet Feed".

eng/pipelines/dotnet-sqlclient-ci-core.yml
```diff
@@ -41,6 +41,18 @@ parameters:
   type: object
   default: [false, true]
 
+- name: SNIVersion
+  displayName: |
+    Use a specific SNI version from the internal validation feed
```
benrr101, line 46: As a display name, I'm thinking of how this would be displayed in the new run dialog. This is really long, and could be reduced to "SNI NuGet Version Override"

eng/pipelines/dotnet-sqlclient-ci-core.yml
```diff
@@ -41,6 +41,18 @@ parameters:
...
+  type: string
+  default: ''
+
+- name: SNIValidationFeed
+  displayName: |
+    The specific SNI test version is pulled from this feed
+  type: string
+  default: https://sqlclientdrivers.pkgs.visualstudio.com/ADO.Net/_packaging/SNIValidation/nuget/v3/index.json
```
benrr101, line 54: After spending so much time on the AKV pipeline, I've got a lot of preferences about style for pipeline yaml. On the one hand, I could lay it all out in this PR and make things annoying. On the other hand, I anticipate rewriting this pipeline as well, and we can hold off on stylistic changes until then. Which would you prefer?

benrr101 (commented):
I think you and I should have a bit of a conversation about when and how the handful of components need to be tested together (MDS, SNI, AKV, MSS, ETC, WTF, BBQ). I want to get a better idea what depends on what and how the pipelines should be constructed to handle those situations. It's nothing that needs to block this specific PR, but since I think we will want to rebuild this pipeline as part of other engineering goals, I think we'll want to discuss it at some point.

eng/pipelines/dotnet-sqlclient-ci-core.yml
mdaigle: If you're ok with waiting, I'd prefer that. I don't anticipate this to be a long term solution. I had other approaches in mind to facilitate SNI testing, but they were blocked on this pipeline migrating to onebranch. My idea was to keep this small and easily removable.

I'd also be happy to standardize on an IDE and formatter for yaml. Raw yaml editing is not the way to go.

PR #3186, opened 2025-02-28
Fix xUnit Theory serialization warnings

paulmedynski:
xUnit makes 2 passes over unit tests:

1. Discover all of the tests, and for Theory tests, enumerate all of the test data.
2. Run all of the tests, enumerating the test data again.

If the enumeration won't produce the same data in both steps, xUnit emits a warning.  We have many test cases whose Theory data doesn't enumerate identically in both passes, for a variety of reasons.  These tests will still run correctly, but they won't appear in test runner UIs like Visual Studio.  Instead, xUnit collapses them into a single test for these UIs.

The warnings are clutter in the console (where I like to run tests), so xUnit provides a mechanism to disable enumeration completely in the discovery phase.  I have updated the offending tests to use the MemberData DisableDiscoveryEnumeration flag.  This doesn't affect how the tests run, and it doesn't change how UIs display them (collapsed).  All it does is suppress the console warnings.

Further reading here:  https://xunit.net/faq/theory-data-stability-in-vs

I also added 4 new targets to `build.proj` to make running the tests easier.  The existing test running targets function as before.

Commit notes:
- Suppressed xUnit discovery data enumeration warnings in Function and Manual test projects.
- Fixed xUnit test data enumeration warnings specific to .NET Framework (enums in the GAC).
- Added targets to run Windows and Unix tests.
- Cleaned up test run command lines for improved maintenance and runtime logging.

### DIFF
--- a/build.proj
+++ b/build.proj
@@ -178,21 +178,109 @@
   </Target>
 
   <!-- Tests -->
+
+  <!-- Run all unit tests applicable to the host OS. -->
   <Target Name="RunTests" DependsOnTargets="RunFunctionalTests;RunManualTests"/>
-  <Target Name="RunFunctionalTests">
-    <!-- Windows -->
-    <Exec ConsoleToMsBuild="true" Command="$(DotnetPath)dotnet test &quot;@(FunctionalTestsProj)&quot; -p:Configuration=$(Configuration) -p:Target$(TFGroup)Version=$(TF) -p:ReferenceType=$(ReferenceType) --no-build -v n --collect &quot;Code coverage&quot; -p:TestSet=$(TestSet) --results-directory $(ResultsDirectory) -p:TestTargetOS=Windows$(TargetGroup) --filter &quot;category!=non$(TargetGroup)tests&amp;category!=failing&amp;category!=nonwindowstests&quot; &quot;--logger:trx;LogFilePrefix=Functional-Windows$(TargetGroup)-$(TestSet)&quot;" Condition="'$(IsEnabledWindows)' == 'true'"/>
-    <!-- Unix -->
-    <Exec ConsoleToMsBuild="true" Command="$(DotnetPath)dotnet test &quot;@(FunctionalTestsProj)&quot; -p:Configuration=$(Configuration) -p:TargetNetCoreVersion=$(TF) -p:ReferenceType=$(ReferenceType) --no-build -v n -p:TestSet=$(TestSet) --results-directory $(ResultsDirectory) -p:TestTargetOS=Unixnetcoreapp --collect &quot;Code coverage&quot; --filter &quot;category!=nonnetcoreapptests&amp;category!=failing&amp;category!=nonlinuxtests&amp;category!=nonuaptests&quot; &quot;--logger:trx;LogFilePrefix=Functional-Unixnetcoreapp-$(TestSet)&quot;" Condition="'$(IsEnabledWindows)' != 'true'"/>
+
+  <!-- Run all Functional tests applicable to the host OS. -->
+  <Target Name="RunFunctionalTests" DependsOnTargets="RunFunctionalTestsWindows;RunFunctionalTestsUnix"/>
+
+  <!-- Run all Functional tests applicable to Windows. -->
+  <Target Name="RunFunctionalTestsWindows" Condition="'$(IsEnabledWindows)' == 'true'">
+    <PropertyGroup>
+      <TestCommand>
+        $(DotnetPath)dotnet test "@(FunctionalTestsProj)"
+        --no-build
+        -v n
+        -p:Configuration=$(Configuration)
+        -p:Target$(TFGroup)Version=$(TF)
+        -p:ReferenceType=$(ReferenceType)
+        -p:TestSet=$(TestSet)
+        -p:TestTargetOS=Windows$(TargetGroup)
+        --collect "Code coverage"
+        --results-directory $(ResultsDirectory)
+        --filter "category!=non$(TargetGroup)tests&amp;category!=failing&amp;category!=nonwindowstests"
+        --logger:"trx;LogFilePrefix=Functional-Windows$(TargetGroup)-$(TestSet)"
+      </TestCommand>
+      <TestCommand>$(TestCommand.Replace($([System.Environment]::NewLine), " "))</TestCommand>
+    </PropertyGroup>
+    <Message Text=">>> Running Functional test for Windows via command: $(TestCommand)"/>
+    <Exec ConsoleToMsBuild="true" Command="$(TestCommand)"/>
   </Target>
 
-  <Target Name="RunManualTests">
-    <!-- Windows -->
-    <Exec ConsoleToMsBuild="true" Command="$(DotnetPath)dotnet test &quot;@(ManualTestsProj)&quot; -p:Configuration=$(Configuration) -p:Target$(TFGroup)Version=$(TF) -p:ReferenceType=$(ReferenceType) --no-build -l &quot;console;verbosity=normal&quot; --collect &quot;Code coverage&quot; -p:TestSet=$(TestSet) --results-directory $(ResultsDirectory) -p:TestTargetOS=Windows$(TargetGroup) --filter &quot;category!=non$(TargetGroup)tests&amp;category!=failing&amp;category!=nonwindowstests&quot; &quot;--logger:trx;LogFilePrefix=Manual-Windows$(TargetGroup)-$(TestSet)&quot;" Condition="'$(IsEnabledWindows)' == 'true'"/>
-    <!-- Unix -->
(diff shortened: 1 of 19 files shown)


build.proj
```diff
@@ -178,21 +178,109 @@
...
-    <Exec ConsoleToMsBuild="true" Command="$(DotnetPath)dotnet test &quot;@(FunctionalTestsProj)&quot; -p:Configuration=$(Configuration) -p:TargetNetCoreVersion=$(TF) -p:ReferenceType=$(ReferenceType) --no-build -v n -p:TestSet=$(TestSet) --results-directory $(ResultsDirectory) -p:TestTargetOS=Unixnetcoreapp --collect &quot;Code coverage&quot; --filter &quot;category!=nonnetcoreapptests&amp;category!=failing&amp;category!=nonlinuxtests&amp;category!=nonuaptests&quot; &quot;--logger:trx;LogFilePrefix=Functional-Unixnetcoreapp-$(TestSet)&quot;" Condition="'$(IsEnabledWindows)' != 'true'"/>
+
+  <!-- Run all Functional tests applicable to the host OS. -->
+  <Target Name="RunFunctionalTests" DependsOnTargets="RunFunctionalTestsWindows;RunFunctionalTestsUnix"/>
+
+  <!-- Run all Functional tests applicable to Windows. -->
+  <Target Name="RunFunctionalTestsWindows" Condition="'$(IsEnabledWindows)' == 'true'">
+    <PropertyGroup>
+      <TestCommand>
+        $(DotnetPath)dotnet test "@(FunctionalTestsProj)"
```
paulmedynski, line 192: Formatted for readability, and then squished back into a single line since <Exec> treats each line in the Command as a separate command!

build.proj
```diff
@@ -178,21 +178,109 @@
...
+        -p:Target$(TFGroup)Version=$(TF)
+        -p:ReferenceType=$(ReferenceType)
+        -p:TestSet=$(TestSet)
+        -p:TestTargetOS=Windows$(TargetGroup)
+        --collect "Code coverage"
+        --results-directory $(ResultsDirectory)
+        --filter "category!=non$(TargetGroup)tests&amp;category!=failing&amp;category!=nonwindowstests"
+        --logger:"trx;LogFilePrefix=Functional-Windows$(TargetGroup)-$(TestSet)"
+      </TestCommand>
+      <TestCommand>$(TestCommand.Replace($([System.Environment]::NewLine), " "))</TestCommand>
```
paulmedynski, line 205: Modern dotnet tooling supports String.ReplaceLineEndings(), but we're using the old .NET Framework msbuild.exe tooling to build, so we can't have the nice toys :(

src/Microsoft.Data.SqlClient/tests/FunctionalTests/AlwaysEncryptedTests/SqlColumnEncryptionCertificateStoreProviderShould.cs
```diff
@@ -230,7 +230,16 @@ public void EncryptAndDecryptDataSuccessfully()
         }
 
         [Theory]
-        [CEKEncryptionReversalParameters]
+        [MemberData(
+            nameof(CEKEncryptionReversalData)
+#if NETFRAMEWORK
+            // .NET Framework puts system enums in something called the Global
```
paulmedynski, line 236: I've left a comment once in each file where we're using DisableDiscoveryEnumeration, to make it clear why we're using that flag.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/AlwaysEncryptedTests/SqlColumnEncryptionCertificateStoreProviderShould.cs
```diff
@@ -505,33 +519,25 @@ public override IEnumerable<Object[]> GetData(MethodInfo testMethod)
             }
         }
 
-        public class CEKEncryptionReversalParameters : DataAttribute
+        public static IEnumerable<object[]> CEKEncryptionReversalData()
```
paulmedynski, line 522: Since xUnit only supports DisableDiscoveryEnumeration on MemberData attributes, I have converted all of the data sources to be compatible with MemberData.

You will see similar changes throughout the tests, where data source classes become methods.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/AlwaysEncryptedTests/SqlColumnEncryptionCertificateStoreProviderShould.cs
```diff
@@ -230,7 +230,16 @@ public void EncryptAndDecryptDataSuccessfully()
         }
 
         [Theory]
-        [CEKEncryptionReversalParameters]
+        [MemberData(
+            nameof(CEKEncryptionReversalData)
+#if NETFRAMEWORK
```
paulmedynski, line 235: .NET doesn't have a GAC, so these tests will enumerate correctly for those targets.  We only need to disable for .NET Framework targets.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/AmbientTransactionFailureTest.cs
```diff
@@ -4,6 +4,7 @@
 
 using System;
 using System.Runtime.InteropServices;
+using System.Threading;
```
paulmedynski, line 7: I don't think this is necessary.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlMetaDataTest.cs
```diff
@@ -348,15 +401,35 @@ public void ConstructorWithInvalidMaxLengthAndLocaleCompareOptionsBinarySortAndI
...
-        [InlineData(SqlDbType.VarBinary)]
-        [InlineData(SqlDbType.Image)]
+        [MemberData(
+            nameof(ConstructorCharData)
+#if NETFRAMEWORK
+            , DisableDiscoveryEnumeration = true
+#endif
+            )]
+        [MemberData(
+            nameof(ConstructorTextData)
```
paulmedynski, line 422: This adds Text to the mix, which seems to have been missing from the InlineData.

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/ConversionTests.cs
```diff
@@ -565,7 +584,20 @@ private void ExecuteAndCheckForError(SqlCommand sqlCmd, bool expectError)
                 try
                 {
                     sqlCmd.ExecuteNonQuery();
-                    Assert.Fail("We should have gotten an error but passed instead.");
+                    StringBuilder builder = new(
```
paulmedynski, line 587: With broken (now disabled) enumeration, xUnit doesn't include the test data in any failure output, so we need to do it ourselves.

This test fails for me on main as well, so I wanted to see what data set was causing the problem.  Turns out to be the Date data set.

src/Microsoft.Data.SqlClient/tests/tools/Microsoft.Data.SqlClient.TestUtilities/config.default.json
```diff
@@ -1,6 +1,13 @@
 {
+    // Typical Windows SQL Server on the local host via TCP.
```
paulmedynski, line 2: These comments are breaking our CI "Update config.json" task because PowerShell's ConvertFrom-Json fails  But as of PowerShell v6, ConvertFrom-Json handles comments:

https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/convertfrom-json?view=powershell-7.5

Are we really using ancient PowerShell 5.1 from 2016?  I will have to dig into this further.

eng/pipelines/common/templates/steps/update-config-file-step.yml
```diff
@@ -126,7 +126,7 @@ parameters:
 
 steps:
 # All properties should be added here, and this template should be used for any manipulation of the config.json file.
-- powershell: |
+- pwsh: |
     $PSVersionTable
```
paulmedynski, line 130: Remove this once we solve the JSON-with-comments problem.

cheenamalhotra (approved):

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlClientFactoryTest.cs
```diff
@@ -31,7 +31,12 @@ public void InstanceTest()
         };
 
         [Theory]
-        [MemberData(nameof(FactoryMethodTestData))]
+        [MemberData(
+            nameof(FactoryMethodTestData),
+            // xUnit can't consistently serialize the data for this test, so we
+            // disable enumeration of the test data to avoid warnings on the
+            // console.
+            DisableDiscoveryEnumeration = true)]
```
mdaigle, line 39: Would a custom serializer help in this case? Not to say we should do it in this PR, just curious if you experimented with it.

mdaigle: https://xunit.net/docs/getting-started/v3/custom-serialization

paulmedynski: Yes looked into that, but didn't want to figure out serialization for all of the affected tests as part of this PR.  The changes here don't affect how the tests run, and they don't change how they're visualized in IDEs.  They just suppress the warnings on the console.  I've opened an ADO task to track fixing the serialization itself.

mdaigle (approved):

PR #3195, opened 2025-03-04
Add Azure DevOps code coverage config file

paulmedynski:
User Story 34103: [S360] dotnet-sqlclient repository is not meeting Differential Dode Coverage standard, classified as Bucket 2: CodeCoverage data present, missing config file

https://learn.microsoft.com/en-us/azure/devops/pipelines/test/codecoverage-for-pullrequests?view=azure-devops#configuring-coverage-settings

- Added Azure DevOps code coverage config file to the repo root.

### DIFF
--- a/azurepipelines-coverage.yml
+++ b/azurepipelines-coverage.yml
@@ -1,0 +1,16 @@
+# Azure DevOps code coverage settings:
+#
+# https://learn.microsoft.com/en-us/azure/devops/pipelines/test/codecoverage-for-pullrequests?view=azure-devops#configuring-coverage-settings
+#
+coverage:
+  # Code coverage status will be posted to pull requests based on targets
+  # defined below.
+  status:
+    # Off by default. When on, details about coverage for each file changed will
+    # be posted as a pull request comment.
+    comments: on
+    # Diff coverage is code coverage only for the lines changed in a pull
+    # request.
+    diff:
+      # Set this to a desired %. Default is 70%.
+      target: 50%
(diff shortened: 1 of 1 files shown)


azurepipelines-coverage.yml
```diff
@@ -0,0 +1,16 @@
...
+  # defined below.
+  status:
+    # Off by default. When on, details about coverage for each file changed will
+    # be posted as a pull request comment.
+    comments: on
+    # Diff coverage is code coverage only for the lines changed in a pull
+    # request.
+    diff:
+      # Set this to a desired %. Default is 70%.
+      target: 50%
```
paulmedynski, line 16: We're starting with a low threshold, and can adjust up as we get familiar with our current coverage and set some policies going forward.

benrr101 (approved):
Assuming the build doesn't blow up by adding this file - it sounds good to me

paulmedynski:
Hmmm... looks like the build steps that would acquire code coverage were skipped.  I guess we don't bother with those if a subset of the files in the repo don't change?

azurepipelines-coverage.yml
cheenamalhotra: We can let it be 70% - it makes us want to stay compliant in future PRs.

cheenamalhotra:
Path filters were wrongly set to exclude this condition.. I've adjusted those.. builds are running now.

azurepipelines-coverage.yml
paulmedynski: Done.

cheenamalhotra (approved):

PR #3196, opened 2025-03-04
CodeQL | RSA Padding Scheme Upgrades

benrr101:
**Description**: As per CodeQL reports, we need to upgrade our RSA padding schemes. In this PR, there are three main changes:

* Change encrypt/decrypt padding for CNG provider from OaepSHA1 to OaepSHA256
* Change encrypt/decrypt padding for Certificate Store provider from OaepSHA1 to OaepSHA256
* Change sign/verify padding for CNG provider from Pkcs1 to PSS (probabilistic signature scheme)

(there's also some extraneous parens removed for good measure - they were looking at me funny)

**Testing**: These should only be impacting enclave, so ... I need the CI to do the testing.


cheenamalhotra (approved):

samsharma2700 (approved):

benrr101:
Closing as SQL Server column encryption keys (which is what we use these encryption/decryption methods for) only supports RSA_OAEP[SHA1] encryption.

PR #3197, opened 2025-03-04
CodeQL | SqlException Serialization Tests

benrr101:
**Description**: As per CodeQL reports, we have a few bits of code in our SqlException serialization tests that are unsafe. The following changes were made to the tests:

* Removed the test that used BinaryFormatter since it cannot be made safe (as per https://learn.microsoft.com/en-us/dotnet/standard/serialization/binaryformatter-security-guide) - The test only applied to netfx. We are not using BinaryFormatter anywhere else in the code, so there's no real need to make sure serialization/deserialization works with BinaryFormatter. If anyone complains, say "it's not supported anymore".
* Use default behavior for serializing SqlException to JSON - previous behavior was to embed type information into the serialization which is unsafe. Removing that behavior is safer.

**Testing**: Tests are running locally without issue on Windows, will let CI determine if Unix is working, too.

### DIFF
--- a/src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlExceptionTest.cs
+++ b/src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlExceptionTest.cs
@@ -15,52 +15,24 @@ namespace Microsoft.Data.SqlClient.Tests
         private const string badServer = "92B96911A0BD43E8ADA4451031F7E7CF";
 
         [Fact]
-        [ActiveIssue("12161", TestPlatforms.AnyUnix)]
         public void SerializationTest()
         {
             SqlException e = CreateException();
             string json = JsonConvert.SerializeObject(e);
 
-            var settings = new JsonSerializerSettings()
-            {
-                TypeNameHandling = TypeNameHandling.All,
-            };
-
-            // TODO: Deserialization fails on Unix with "Member 'ClassName' was not found."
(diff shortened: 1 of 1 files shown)


src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlExceptionTest.cs
```diff
@@ -15,52 +15,24 @@ public class SqlExceptionTest
...
-        {
-            var formatter = new BinaryFormatter();
-            SqlException e = CreateException();
-            using (var stream = new MemoryStream())
-            {
-                try
-                {
-                    formatter.Serialize(stream, e);
-                    stream.Position = 0;
-                    var e2 = (SqlException)formatter.Deserialize(stream);
```
paulmedynski, line 49: This test wasn't Assert-ing anything anyway, so not very useful.

paulmedynski (approved):

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlExceptionTest.cs
benrr101: I mean it was asserting that no exceptions happened, but that _seems_ like it might just be testing whether BinaryFormatter works. Which is not what unit tests are supposed to do :)

mdaigle (approved):

PR #3198, opened 2025-03-05
CodeQL | Fix JWT Validation

benrr101:
**Description**: As per CodeQL reports, we are bypassing some JWT validation and need to stop doing that. This PR turns back on ValidateAudience for JWT token validation in AzureAttestationBasedEnclaveProvider.

I'm not sure what this might break, @David-Engel was this the change you suspected was a behavior change we need to more closely investigate? The fix is easy, but we have many dependencies deep within enclave code.


paulmedynski (approved):

David-Engel:
As of March 2024, MAA wasn't including the audience claim in their tokens, so MDS is set to not validate it. I'll forward the old email to the team. If our tests aren't exercising MAA attestation endpoints, they won't catch the error.

benrr101:
Withdrawing PR as MAA won't issue audience claim that we can validate. Requesting an exception for CodeQL issue

Issue #25, opened 2018-06-15
Investigate SqlClient connection pool for any performance improvement opportunities


saurabh500:
The repo https://github.com/aspnet/DataAccessPerformance can be used to get a measurement of the throughput of an ADO.Net client. 

When the benchmark is run with Connection Pool enabled, and connections are opened and closed every time a query is performed, I see a TPS number of 40k for async and 50k for sync data retrieval. 

I changed the benchmark to reuse the same connection per thread and this causes the TPS to jump to 65 k and 90k for async and sync scenarios, respectively. 

By changing how the connection is opened, I eliminated the connection pool from the benchmark, as the same connection is reused for the data operations. My modifications are at https://github.com/saurabh500/DataAccessPerformance/blob/modVariation/src/BenchmarkDb/AdoDriver.cs#L106 

This means that there could be a potential to improve the performance of the connection pool in SqlClient ADO.Net Driver to manage connection faster.

It could be suffering from one of the following problems based on my understanding of the connection pool. I have not profiled this piece yet. 
1.	There are problems with fetching the connection in the connection pool causing the slow down.
2.	There are problems while returning the connection to the connection pool causing the slow down.
3.	When a connection is returned to the pool, sp_connection_reset is issued to the server and some other network operations to clean the connection, which happen. I don’t know how much of an overhead this would be, though I anticipate that there are some.

I am opening this issue to see if any performance improvement can be made to SqlClient connection pool to have a better throughput in providing connections. 

If it is 1 or 2 plaguing the driver then there could be client side improvements that can be made, if it is 3, then that would be much harder as there could be protocol changes needed. 

Connection pool should be profiled to find any other optimization opportunities, if any.

GSPP:
> For the record, the optimization with the connection reset was about reducing a roundtrip.

How does this interact with possibly open transactions? It's important that those are killed when the connection is disposed.

Also, temp tables might hold up significant resources.

PR #3199, opened 2025-03-05
Refactor for modular connection pool

mdaigle:
Introduces an abstract class to capture the API of the current connection pool implementation. Gathers pooling related classes into a single namespace (Microsoft.Data.SqlClient.ConnectionPool) and folder. These changes set the stage for swappable connection pool implementations. Targeting main because these are good standalone changes for code cleanliness. Future changes will target a feature branch.

This work is tracking against https://github.com/dotnet/SqlClient/issues/25

PoolBlockingPeriod is not moved because it is public. Moving it's location/namespace would be a breaking API change. It may still make sense to switch its namespace if this can go in with a major version bump.


src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnectionFactory.cs
```diff
@@ -12,6 +12,7 @@
 using Microsoft.Data.Common;
 using Microsoft.Data.ProviderBase;
 using Microsoft.Data.SqlClient.Server;
+using Microsoft.Data.SqlClient.ConnectionPool;
```
benrr101, line 15: nit - should go one line up :)

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnectionHelper.cs
```diff
@@ -12,6 +12,7 @@ namespace Microsoft.Data.SqlClient
     using System.Threading;
     using Microsoft.Data.Common;
     using Microsoft.Data.ProviderBase;
+    using Microsoft.Data.SqlClient.ConnectionPool;
     using System.Transactions;
```
benrr101, line 16: Nit: System.* should go before Microsoft.*

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/DbConnectionPool.cs
```diff
@@ -0,0 +1,120 @@
...
+using System.Transactions;
+using Microsoft.Data.Common;
+using Microsoft.Data.ProviderBase;
+
+namespace Microsoft.Data.SqlClient.ConnectionPool
+{
+    internal abstract class DbConnectionPool
+    {
+        private static int _objectTypeCount; // EventSource counter
+        internal readonly int _objectID = System.Threading.Interlocked.Increment(ref _objectTypeCount);
```
benrr101, line 18: I would not recommend making this internal. Aside from "constants", I generally think member variables should not be visible outside their class.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/DbConnectionPool.cs
```diff
@@ -0,0 +1,120 @@
...
+
+        internal int ObjectID
```
benrr101, line 20: This can be defined as an auto-property, and that resolves the issue of having an externally visible member variable.

benrr101, line 20: Following naming conventions, this should be "ObjectId"

benrr101: Alternatively, I think in other scenarios where I've seen ObjectId used, we're just using it within the same class as it's defined in. In that case, we can dispense with the property and just make the member variable private.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/DbConnectionPool.cs
```diff
@@ -0,0 +1,120 @@
...
+        {
+            get
+            {
+                return _objectID;
+            }
+        }
+
+        private DbConnectionPoolState _state;
+        internal DbConnectionPoolState State
```
benrr101, line 29: Auto property, please :)

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/DbConnectionPool.cs
benrr101: ALTERNATIVELY, if the intention is for this to be accessible within child classes, use protected property.

To summarize:
* Need to access it outside the class/children -> internal auto-property
* Need to access it from self and children only -> protected auto-property
* Need to access it from self only -> private member variable

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/DbConnectionPool.cs
```diff
@@ -0,0 +1,120 @@
...
+        {
+            get => _state;
+            set => _state = value;
+        }
+
+        #region Abstract Properties
+        internal abstract int Count
+        {
+            get;
```
benrr101, line 38: nit: These should be on the same line as the definition (ie `internal abstract int Count { get; }`)

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/DbConnectionPoolState.cs
```diff
@@ -0,0 +1,13 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.Data.SqlClient.ConnectionPool
+{
+    internal enum DbConnectionPoolState
+    {
+        Initializing,
```
benrr101, line 9: I'm not sure if this would raise issues with our code quality thingies, but I recall getting warnings in the past (on a different project) that enums should have a "=0" value defined (in most cases like an "unknown"/"not specified" state). 

ah here it is https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1008

benrr101: It's probably outside the scope of this PR, but we should consider if we might expect more states that a pool might be in. Right now this enum is just a cut/paste from the waithandle implementation - just want to make sure we cover our bases as we make things extensible.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/WaitHandleDbConnectionPool.cs
```diff
@@ -474,17 +461,17 @@ private int CreationTimeout
             get { return PoolGroupOptions.CreationTimeout; }
         }
 
-        internal int Count
+        internal override int Count
```
benrr101, line 464: Can these be auto-properties?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/WaitHandleDbConnectionPool.cs
```diff
@@ -503,7 +490,7 @@ private bool NeedToReplenish
         {
             get
             {
-                if (State.Running != _state) // Don't allow connection create when not running.
+                if (State != Running) // Don't allow connection create when not running.
```
benrr101, line 493: nit: `is not`?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/WaitHandleDbConnectionPool.cs
```diff
@@ -825,20 +804,6 @@ private DbConnectionInternal CreateObject(DbConnection owningObject, DbConnectio
 #endif
                 }
 
-                // If the old connection belonged to another pool, we need to remove it from that
```
benrr101, line 828: No longer necessary operation, I guess?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/WaitHandleDbConnectionPool.cs
```diff
@@ -926,7 +891,7 @@ private void DeactivateObject(DbConnectionInternal obj)
             {
                 // NOTE: constructor should ensure that current state cannot be State.Initializing, so it can only
                 //   be State.Running or State.ShuttingDown
-                Debug.Assert(_state == State.Running || _state == State.ShuttingDown);
+                Debug.Assert(State == Running || State == ShuttingDown);
```
benrr101, line 894: nit: `is Running or ShuttingDown`?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/WaitHandleDbConnectionPool.cs
```diff
@@ -936,7 +901,7 @@ private void DeactivateObject(DbConnectionInternal obj)
                     // transaction object will ensure that it is owned (not lost),
                     // and it will be certain to put it back into the pool.                    
 
-                    if (_state == State.ShuttingDown)
+                    if (State == ShuttingDown)
```
benrr101, line 904: nit: `is ShuttingDown` :)

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/WaitHandleDbConnectionPool.cs
```diff
@@ -1235,7 +1200,7 @@ internal bool TryGetConnection(DbConnection owningObject, TaskCompletionSource<D
                 allowCreate = true;
             }
 
-            if (_state != State.Running)
+            if (State != Running)
```
benrr101, line 1203: nit: `is not Running`

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/WaitHandleDbConnectionPool.cs
```diff
@@ -1613,7 +1578,7 @@ private void PoolCreateRequest(object state)
             long scopeID = SqlClientEventSource.Log.TryPoolerScopeEnterEvent("<prov.DbConnectionPool.PoolCreateRequest|RES|INFO|CPOOL> {0}", ObjectID);
             try
             {
-                if (State.Running == _state)
+                if (State == Running)
```
benrr101, line 1581: nit: ... you know :P

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/Common/SystemDataInternals/ConnectionPoolHelper.cs
```diff
@@ -14,23 +14,24 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests.SystemDataInternals
     internal static class ConnectionPoolHelper
     {
         private static Assembly s_MicrosoftDotData = Assembly.Load(new AssemblyName(typeof(SqlConnection).GetTypeInfo().Assembly.FullName));
-        private static Type s_dbConnectionPool = s_MicrosoftDotData.GetType("Microsoft.Data.ProviderBase.DbConnectionPool");
-        private static Type s_dbConnectionPoolGroup = s_MicrosoftDotData.GetType("Microsoft.Data.ProviderBase.DbConnectionPoolGroup");
-        private static Type s_dbConnectionPoolIdentity = s_MicrosoftDotData.GetType("Microsoft.Data.ProviderBase.DbConnectionPoolIdentity");
+        private static Type s_dbConnectionPool = s_MicrosoftDotData.GetType("Microsoft.Data.SqlClient.ConnectionPool.DbConnectionPool");
```
benrr101, line 17: Doesn't manual tests have a reference to the SqlClient project? I'm really confused why we need to get these types via reflection instead of direct reference.

benrr101 (changes requested):
Overall I like these changes. Mostly stylistic changes I'm requesting, a couple bigger questions around what the longer term plan is.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/Common/SystemDataInternals/ConnectionPoolHelper.cs
mdaigle: It does, but the internals aren't visible. I spent a good couple hours trying to make that happen, but declared defeat in the context of this PR. I'm sure we can make it happen, but we'll need to dedicate some more time to it.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/DbConnectionPool.cs
mdaigle: I see one usage outside the class used for tracing which pools are pruned. So I can't make it private/protected without altering that trace. I'll leave it as internal.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/WaitHandleDbConnectionPool.cs
mdaigle: Yeah, forgot to add a comment on this. The old connection is already doomed, so it will be cleared from the other pool when that pool gets pruned.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/DbConnectionPoolState.cs
mdaigle: The most appropriate setting right now would probably be "Initializing". But I'll keep this in mind as we move forward.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/Common/SystemDataInternals/ConnectionPoolHelper.cs
benrr101: oh damn. I swear we fixed that a long time ago in the clientx project... Ok no biggie.

benrr101 (approved):
Looks much better now, thanks!

cheenamalhotra (approved):

Issue #3067, opened 2024-12-09
SqlCommand async Task may never complete because of unhandled exception


SergeyBokhantsev:
### Describe the bug

Async Task of SqlCommand sometimes could never go to Completed state (if exception occurs during command execution). 
Thus, if you await the SqlCommand then your code may stuck forever at waiting the task completion.

Observed case:
There is a trivial code to interact with Azure SQL DB via EntityFramework. We noticed that sometimes the module that is working with DB freezes forever. Analyzing memory dumps we saw that it is always stuck awaiting EntityFramework operations like Select, Find, SaveChangesAsync etc.
In every case there is an unobserved exception logged also with the following details about the root exception:
_Invalid operation. The connection is closed. 
System.AggregateException: System.InvalidOperationException: 
at Microsoft.Data.SqlClient.SqlConnection.GetOpenTdsConnection 
at Microsoft.Data.SqlClient.SqlCommand+<>c__DisplayClass186_0.<CreateLocalCompletionTask>b__0 
at System.Threading.ExecutionContext.RunFromThreadPoolDispatchLoop 
at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw 
at System.Threading.ExecutionContext.RunFromThreadPoolDispatchLoop 
at System.Threading.Tasks.Task.ExecuteWithThreadLocal_

### To reproduce
Unfortunately I don't have code to reproduce the issue because in our case the issue was caused by a sporadic network connection problem.
But taking into account the above exception call stack, the issue in source code can be seen:

- Because of some exception (network issue in our case) that happens here https://github.com/dotnet/SqlClient/blob/bb4c3b7a3a3125b23bc8ddce105ff3ba67a25414/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs#L2407

- the execution gets here into the catch block https://github.com/dotnet/SqlClient/blob/bb4c3b7a3a3125b23bc8ddce105ff3ba67a25414/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs#L2448C29-L2448C70

[...]

PR #3202, opened 2025-03-06
Add try-catch and ensure task exception is set in SqlCommand

Wraith2:
fixes https://github.com/dotnet/SqlClient/issues/3067

Adds a try-catch block around getting a connection which can fail. This change ensures that the task completion source that tasks are waiting on is signalled.

Also found a trivial continue that was capturing the tcs so I changed it to static and passed the tcs as state avoiding the lambda capture.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
@@ -2445,9 +2445,16 @@ namespace Microsoft.Data.SqlClient
                                 _cachedAsyncState.ResetAsyncState();
                             }
 
-                            _activeConnection.GetOpenTdsConnection().DecrementAsyncCount();
+                            try
+                            {
+                                _activeConnection.GetOpenTdsConnection().DecrementAsyncCount();
 
-                            globalCompletion.TrySetException(e);
+                                globalCompletion.TrySetException(e);
+                            }
+                            catch (Exception e2)
+                            {
+                                globalCompletion.TrySetException(e2);
+                            }
                         }
                         else
                         {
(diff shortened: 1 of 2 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
```diff
@@ -2445,9 +2445,16 @@ long firstAttemptStart
...
+                            try
+                            {
+                                _activeConnection.GetOpenTdsConnection().DecrementAsyncCount();
 
-                            globalCompletion.TrySetException(e);
+                                globalCompletion.TrySetException(e);
+                            }
+                            catch (Exception e2)
+                            {
+                                globalCompletion.TrySetException(e2);
```
paulmedynski, line 2456: Is there a way to avoid losing the original exception (e) here?  Assign it to e2.InnerException if not set maybe?  C# Exceptions don't seem to have a 'related exceptions' list.

paulmedynski (changes requested):
See question about exception loss.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
Wraith2: If you look further down from the change in the else block you'll see that the retry block is surrounded in a catch that drops the inner connection and uses the outer, that's why I chose to go with this method. 
Arguably whatever the inner exception is the fact that the connection has been closed is probably more informative to the consumer.
We could create an aggregate exception I suppose but I'm not a huge fan of those in general.

paulmedynski: Fair enough - no worse than the existing behaviour below :)

paulmedynski (approved):

cheenamalhotra (approved):

PR #3204, opened 2025-03-06
Tests | Remove hardcoded credentials from ManualTests

benrr101:
_This is a remake of #3090 so that we can get all of the CI to run, original description follows_

---

This PR mops up almost all of the remaining hardcoded credentials, and does three things:
* Removes references to `CertificateUtility.CreateCertificate`, which always used a hardcoded certificate.
* Forces the tests to dynamically generate keys in Azure Key Vault.
* Removes references to `CertificateUtilityWin` in `CspProviderExt`, then the class itself.

#### Removal of hardcoded credentials ####

The removal of CertificateUtility.CreateCertificate is pretty uncontroversial. However, the [`AKVTest.TestRoundTripWithAKVAndCertStoreProvider`](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/AKVTests.cs#L132-L146) test tries to round-trip encrypted column data, encrypting it with the hardcoded certificate and decrypting it with the Azure Key Vault key. I've thus treated this key as another static credential and removed it.

It's worth noting that the [`CoreCryptoTests`](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/CoreCryptoTests.cs) class loads hardcoded credentials from TCECryptoNativeBaselineRsa.txt. I've not touched this because I don't know what to do with it. It isn't testing any SqlClient-specific functionality, just .NET's ability to decrypt text which is encrypted by native code. I'm not sure whether this is necessary any more though: we already test this implicitly with the end-to-end AE tests. There's no modification required here - just a choice on whether to keep or delete the test entirely.

#### CspProviderExt tests ####

One partially-related change was to remove CertificateUtilityWin. This was only ever used by the [`CspProviderExt`](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/CspProviderExt.cs) tests, which used RSACryptoServiceProvider to generate a key in a specific CSP and encrypt/decrypt data with it. This consisted of three tests:
1. TestKeysFromCertificatesCreatedWithMultipleCryptoProviders, which looked for all CSPs with a name containing "RSA and AES", then generated a key in that CSP and tested the Always Encrypted provider with it.
2. TestRoundTripWithCSPAndCertStoreProvider, which generated a certificate and tried to round-trip between the certificate-based AE provider and the CSP-based AE provider. XUnit was already being told to skip this test.
3. TestEncryptDecryptWithCSP, which did exactly the same thing as test 1, but with a hard-coded CSP of "Microsoft Enhanced RSA and AES Cryptographic Provider".

Tests 1 and 3 are actually identical. "Microsoft Enhanced RSA and AES Cryptographic Provider" is the only CSP which fits the criteria for Test 1. I've thus eliminated test 1, and modified test 3 to run against all matching CSPs.

Of the remaining two tests, TestEncryptDecryptWithCSP would run a PowerShell script to generate a certificate, then extract the private key and use this as the CSP. It didn't need to do this, it could simply instantiate an RSACryptoServiceProvider directly and use that. I've thus eliminated the certificate generation. The round-trip test did need the certificate though, so I've just switched it to use the same certificate generation fixture as the rest of the tests. It still fails if I force it to run, but it fails with the same error as before. Once these changes were done, CertificateUtilityWin was no longer referenced and could be deleted wholesale.

The CspProviderExt/CertificateUtilityWin changes comprise around half of the changes. If it's easier to review, I can split them into a separate PR.


benrr101:
@edwardneal I could use a wee bit of help figuring out the errors - it looks to me that a table doesn't exist, so we can't delete the entries in it before running the test. Since we're not getting any errors creating the tables, I suspect we're dropping the table somewhere between tests. But I can't seem to repro it locally (although that could be an issue with my setup not being the same as test machine). 

I've also kicked the failing tests again to see if maybe it's transient. But since it's consistent across all the enclave tests I suspect it isn't transient.

edwardneal:
Sorry for the delay here @benrr101, I've just been able to get to look at this.

The fixture here is `PlatformSpecificTestContext`. This instantiates a `SQLSetupStrategyCertStoreProvider`, which calls the base `SQLSetupStrategy` - `CreateTables`, followed by `SetupDatabase`.

The effect of `CreateTables` which we care about is that it populates the `TrustedMasterKeyPathsTestTable` and returns it; the caller adds this to the `databaseObjects` field.

`SetupDatabase` iterates over each of these objects, calling their `Create` method. Then, it calls `InsertSampleData`, which wipes the table and adds a new record to it.

The net effect is that we create a new table, then immediately insert sample data into it. This fails, so the test fixture can't be instantiated and every test which uses the fixture fails. I'm not able to reproduce it either.

I'd like to compare the logs of a few CI runs - could you retrigger this please?

benrr101:
@edwardneal Ok, that's mostly what I suspected that it was an issue with the database setup steps causing the fixture to fail to be instantiated. Stuff like that working locally but not on the test machine suggests to me some kind of race condition, but that's just wild speculation. As per your request, I've kicked off another CI run. Let me know if you need more - I appreciate you looking into this :)

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/TestFixtures/SQLSetupStrategy.cs
```diff
@@ -89,6 +90,14 @@ internal virtual void SetupDatabase()
                 }
 
                 // Insert data for TrustedMasterKeyPaths tests.
+                InsertSampleData(TrustedMasterKeyPathsTestTable.Name);
+            }
```
edwardneal, line 94:
```suggestion
            }
            // Insert data for TrustedMasterKeyPaths tests.
            InsertSampleData(TrustedMasterKeyPathsTestTable.Name);
```

If `AEConnStringsSetup` has multiple values, I think this could be the reason why tests are failing. We create the table in the first data source, then call `InsertSampleData` - but this iterates over every data source, so when it reaches the second it's trying to address a table which doesn't yet exist.

I think there might be another test failure to come, but this should at least let us see it...

benrr101:
@edwardneal Any further thoughts on this one?

src/Microsoft.Data.SqlClient/tests/tools/Microsoft.Data.SqlClient.TestUtilities/Fixtures/CertificateFixtureBase.cs
```diff
@@ -54,7 +56,7 @@ protected static X509Certificate2 CreateCertificate(string subjectName, IEnumera
 #if NET9_0_OR_GREATER
```
edwardneal, line 56:
```suggestion
#if NET
```
The .NET 8 target on Ubuntu is trying to use the Windows-specific path. This should resolve that.

src/Microsoft.Data.SqlClient/tests/tools/Microsoft.Data.SqlClient.TestUtilities/Fixtures/CertificateFixtureBase.cs
```diff
@@ -85,7 +87,7 @@ protected static X509Certificate2 CreateCertificate(string subjectName, IEnumera
             using (X509Certificate2 ephemeral = request.CreateSelfSigned(notBefore, notAfter))
             {
                 return X509CertificateLoader.LoadPkcs12(ephemeral.Export(X509ContentType.Pkcs12, password), password,
-                    X509KeyStorageFlags.PersistKeySet | X509KeyStorageFlags.Exportable);
+                    X509KeyStorageFlags.PersistKeySet | X509KeyStorageFlags.Exportable, new Pkcs12LoaderLimits(Pkcs12LoaderLimits.Defaults) { PreserveStorageProvider = true, PreserveKeyName = true });
```
edwardneal, line 90: It looks like MDS.TestUtilities is built against net9.0;net462 locally, and net8.0;net9.0;net462 in CI - so there are CI-based compile errors which don't show up when I compile locally.

I'm not sure what the best way to fix that problem is, but in the mean time, the underlying compilation issue is that `X509CertificateLoader` doesn't exist in net8.0. I've fixed this locally by replacing these two lines with:

```csharp
#if NET9_0_OR_GREATER
return X509CertificateLoader.LoadPkcs12(ephemeral.Export(X509ContentType.Pkcs12, password), password,
    X509KeyStorageFlags.PersistKeySet | X509KeyStorageFlags.Exportable, new Pkcs12LoaderLimits(Pkcs12LoaderLimits.Defaults) { PreserveStorageProvider = true, PreserveKeyName = true });
#else
return new X509Certificate2(ephemeral.Export(X509ContentType.Pkcs12, password), password,
    X509KeyStorageFlags.PersistKeySet | X509KeyStorageFlags.Exportable);
#endif
```

paulmedynski:
In general I think we should:
- Build using the latest SDK on whatever OSes we choose to support _development_ on.  For example:
  - Build for all supported Target Frameworks on Windows using both VS2022 SDK _and_ .NET 9.0 SDK.
  - Build for all supported Target Frameworks on .NET 9.0 SDK on Linux and macOS (see #3311).
  - I don't think we should support _building_ using older SDKs.
- Run tests on all _target runtimes_ we support, on whichever OSes those runtimes exist.  For example:
  - Test net462, net47, net471, net472, net48, net481 _runtimes_ on Windows.
    - There are official Windows Server containers that support this (except net462 - see https://github.com/microsoft/Windows-Containers/issues/562).
  - Test net8.0, net9.0 _runtimes_ on Windows, Linux, macOS.
    - We can install whatever .NET runtimes we want as part of CI.
 
Today, we are building some things on some OSes, other things on othe OSes, and testing some things here, some there, etc.  We definitely need to clean it all up.

edwardneal:
Can you re-run CI please? We're compiling, but the AE legs are now timing out and I don't have any logs to use for troubleshooting.

edwardneal:
CI has timed out again, but I was able to see the cause in the realtime logs a few hours ago. It looks like we're not receiving any response from 10.0.0.4 and 10.0.0.5 prior to running the tests. Is the AE Enclave test environment healthy?

saurabh500:
@cheenamalhotra does this have to do with the VM decommissioning?

@edwardneal there are test infrastructure issues we are trying to iron out.

David-Engel:
10.0.0.5 is the one that was deallocated due to Azure capacity pressure, yes. It let me start it back up so I've re-run failed jobs for this PR.

edwardneal:
Thanks - I can see that most of these tests are passing now (with the exception of Azure Key Vault - I'll try to reproduce that locally.)

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/TestFixtures/SQLSetupStrategyAzureKeyVault.cs
```diff
@@ -42,10 +53,20 @@ public static void RegisterGlobalProviders(SqlColumnEncryptionAzureKeyVaultProvi
             IsAKVProviderRegistered = true;
         }
 
+        private void SetupAzureKeyVault()
+        {
+            JsonWebKey rsaImport = new JsonWebKey(ColumnMasterKeyCertificate.GetRSAPrivateKey(), true);
```
edwardneal, line 58:
```suggestion
        private static RSA CopyKey(RSA rsa)
        {
#if NET8_0
            // In .NET Framework, the key is exportable in plaintext. In .NET 9.0+, we use X509CertificateLoader2 to maintain this functionality.
            // We need to manually handle this in .NET 8.0 with an non-plaintext export.
            RSA replacementKey = RSA.Create(rsa.KeySize);
            Span<byte> passwordBytes = stackalloc byte[32];
            PbeParameters pbeParameters = new PbeParameters(PbeEncryptionAlgorithm.Aes256Cbc, HashAlgorithmName.SHA256, 10000);

            Random.Shared.NextBytes(passwordBytes);

            replacementKey.ImportEncryptedPkcs8PrivateKey(
                passwordBytes,
                rsa.ExportEncryptedPkcs8PrivateKey(passwordBytes, pbeParameters),
                out _);
            return replacementKey;
#else
            return rsa;
#endif
        }
        
        private void SetupAzureKeyVault()
        {
            RSA rsaCopy = CopyKey(ColumnMasterKeyCertificate.GetRSAPrivateKey());
            JsonWebKey rsaImport = new JsonWebKey(rsaCopy, true);
```

It looks like creating a new X509Certificate2 resets the AllowPlaintextExport flag on the underlying private key, which prevents us from instantiating a JsonWebKey. Performing an encrypted export and import into another RSA instance allows that.

This should hopefully resolve the remaining test failures.

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/TestFixtures/SQLSetupStrategyAzureKeyVault.cs
```diff
@@ -4,6 +4,9 @@
 
 using System;
 using System.Collections.Generic;
+using System.Security.Cryptography.X509Certificates;
```
edwardneal, line 7:
```suggestion
using System.Security.Cryptography;
using System.Security.Cryptography.X509Certificates;
```

paulmedynski (changes requested):
This is a large review, and I've only skimmed it so far.  I will need a subject-matter expert to walk me through most of it.

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/TestFixtures/SQLSetupStrategyAzureKeyVault.cs
```diff
@@ -42,10 +53,42 @@ public static void RegisterGlobalProviders(SqlColumnEncryptionAzureKeyVaultProvi
             IsAKVProviderRegistered = true;
         }
 
+        private static RSA CopyKey(RSA rsa)
+        {
+#if NET8_0
```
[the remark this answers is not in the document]
edwardneal: It's *only* for .NET 8.0. This problem doesn't occur for .NET Framework, and in .NET 9.0 the X509Certificate2 constructor is marked as obsolete.

paulmedynski (commented):
Clearing my 'Request changes' review, but not approving.  In the interest of getting this merged, other team members with a better understanding of the changes should review for approval.

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/TestTrustedMasterKeyPaths.cs
```diff
@@ -147,18 +147,14 @@ public void TestTrustedColumnEncryptionMasterKeyPathsWithMultipleServers(string
...
             {
-                string tempThumbprint = new string('F', CertificateUtility.CreateCertificate().Thumbprint.Length);
-                string invalidKeyPath = string.Format(@"{0}/my/{1}", StoreLocation.CurrentUser.ToString(), tempThumbprint);
-                server1TrustedKeyPaths.Add(invalidKeyPath);
-            }
-
-            // Add the key path used by the test
-            server1TrustedKeyPaths.Add(columnMasterKeyPath);
+                // Add some random key paths
+                string.Format(@"{0}/my/{1}", StoreLocation.CurrentUser.ToString(), dummyThumbprint),
```
mdaigle, line 153: ✏️  duplicate

benrr101:
@paulmedynski - I can't review this on myself, so that means we've only got you, @mdaigle, @samsharma2700, and @cheenamalhotra who can review it.

cheenamalhotra (approved):

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/ConversionTests.cs
```diff
@@ -29,7 +30,7 @@ public sealed class ConversionTests : IDisposable
         private const decimal SmallMoneyMinValue = -214748.3648M;
         private const int MaxLength = 10000;
         private int NumberOfRows = DataTestUtility.EnclaveEnabled ? 10 : 100;
-        private static X509Certificate2 certificate;
+        private X509Certificate2 certificate;
```
paulmedynski: private readonly

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/CspProviderExt.cs
```diff
@@ -25,185 +24,48 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests.AlwaysEncrypted
...
+            byte[] encryptedColumnEncryptionKeyUsingCSP = cspProvider.EncryptColumnEncryptionKey(cspPath, @"RSA_OAEP", columnEncryptionKey);
+            byte[] columnEncryptionKeyReturnedCSP2Cert = certProvider.DecryptColumnEncryptionKey(certificatePath, @"RSA_OAEP", encryptedColumnEncryptionKeyUsingCSP);
+            Assert.True(columnEncryptionKey.SequenceEqual(columnEncryptionKeyReturnedCSP2Cert));
         }
 
         [ConditionalTheory(typeof(DataTestUtility), nameof(DataTestUtility.AreConnStringSetupForAE))]
-        [ClassData(typeof(AEConnectionStringProvider))]
-        public void TestEncryptDecryptWithCSP(string connectionString)
+        [ClassData(typeof(AEConnectionStringProviderWithCspParameters))]
+        public void TestEncryptDecryptWithCSP(string connectionString, CspParameters cspParameters)
```
paulmedynski: Should this test just take the ProviderType and ProviderName as direct arguments?  We're not using anything else from CspParameters.

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/TestFixtures/AEConnectionStringProviderWithCspParameters.cs
```diff
@@ -0,0 +1,46 @@
...
+using System.Security.Cryptography;
+
+namespace Microsoft.Data.SqlClient.ManualTesting.Tests.AlwaysEncrypted
+{
+    public class AEConnectionStringProviderWithCspParameters : IEnumerable<object[]>
+    {
+        public IEnumerator<object[]> GetEnumerator()
+        {
+            const string ProvidersRegistryKeyPath = @"SOFTWARE\Microsoft\Cryptography\Defaults\Provider";
+            using Microsoft.Win32.RegistryKey defaultCryptoProvidersRegistryKey = Microsoft.Win32.Registry.LocalMachine.OpenSubKey(ProvidersRegistryKeyPath);
```
paulmedynski: Does this function on non-Windows?  Should its filename have a .windows suffix?

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/TestFixtures/SQLSetupStrategy.cs
```diff
@@ -59,15 +59,16 @@ public class SQLSetupStrategy : IDisposable
...
 
         public SQLSetupStrategy()
+            : base(true)
         {
-            if (certificate == null)
-            {
-                certificate = CertificateUtility.CreateCertificate();
-            }
-            keyPath = string.Concat(StoreLocation.CurrentUser.ToString(), "/", StoreName.My.ToString(), "/", certificate.Thumbprint);
+            ColumnMasterKeyPath = string.Format("{0}/{1}/{2}", StoreLocation.CurrentUser, StoreName.My, ColumnMasterKeyCertificate.Thumbprint);
```
paulmedynski: Prefer interpolation?

src/Microsoft.Data.SqlClient/tests/tools/Microsoft.Data.SqlClient.TestUtilities/Fixtures/AzureKeyVaultKeyFixtureBase.cs
```diff
@@ -0,0 +1,63 @@
...
+// See the LICENSE file in the project root for more information.using System;
+
+using System;
+using System.Collections.Generic;
+using Azure.Core;
+using Azure.Security.KeyVault.Keys;
+
+namespace Microsoft.Data.SqlClient.TestUtilities.Fixtures
+{
+    public abstract class AzureKeyVaultKeyFixtureBase : IDisposable
```
paulmedynski, line 12: Does this class need to be abstract?

src/Microsoft.Data.SqlClient/tests/tools/Microsoft.Data.SqlClient.TestUtilities/Fixtures/CertificateFixtureBase.cs
```diff
@@ -14,6 +14,8 @@ namespace Microsoft.Data.SqlClient.TestUtilities.Fixtures
 {
     public abstract class CertificateFixtureBase : IDisposable
     {
+        private const string CspProviderName = "Microsoft Enhanced RSA and AES Cryptographic Provider";
```
paulmedynski, line 23: What's the significance of this string?  Does it have to be this value, or could we use anything, something else specific... ?

src/Microsoft.Data.SqlClient/tests/tools/Microsoft.Data.SqlClient.TestUtilities/Fixtures/ColumnMasterKeyCertificateFixture.cs
```diff
@@ -0,0 +1,29 @@
...
+    public class ColumnMasterKeyCertificateFixture : CertificateFixtureBase
+    {
+        public ColumnMasterKeyCertificateFixture()
+            : this(true)
+        {
+        }
+
+        public X509Certificate2 ColumnMasterKeyCertificate { get; }
+
+        protected ColumnMasterKeyCertificateFixture(bool createCertificate)
```
paulmedynski, line 19: Is this ever called with false?

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/TestFixtures/SQLSetupStrategy.cs
```diff
@@ -9,15 +9,15 @@
...
 namespace Microsoft.Data.SqlClient.ManualTesting.Tests.AlwaysEncrypted
 {
-    public class SQLSetupStrategy : IDisposable
+    public class SQLSetupStrategy : ColumnMasterKeyCertificateFixture
     {
         internal const string ColumnEncryptionAlgorithmName = @"AEAD_AES_256_CBC_HMAC_SHA256";
 
-        protected static X509Certificate2 certificate;
-        public string keyPath { get; internal set; }
+        public string ColumnMasterKeyPath { get; protected set; }
```
paulmedynski: Is this ever assigned to from a protected context?

paulmedynski (changes requested):
It all makes sense to me, with a few comments.

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/CspProviderExt.cs
edwardneal: It could - there's no significance to CspParameters here, it's just a transport type.

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/TestFixtures/AEConnectionStringProviderWithCspParameters.cs
edwardneal: Agreed

src/Microsoft.Data.SqlClient/tests/tools/Microsoft.Data.SqlClient.TestUtilities/Fixtures/AzureKeyVaultKeyFixtureBase.cs
edwardneal: It does, yes. The goal here is that someone who needs to use an AKV key in the future would derive a class based upon their use case and call `CreateKey` in the constructor.

src/Microsoft.Data.SqlClient/tests/tools/Microsoft.Data.SqlClient.TestUtilities/Fixtures/CertificateFixtureBase.cs
edwardneal: It needs to be this value specifically, because certificates created by PowerShell using another provider aren't accessible from `RSACryptoServiceProvider`, which means that we can't round-trip between `SqlColumnEncryptionCertificateStoreProvider` and `SqlColumnEncryptionCspProvider`.

src/Microsoft.Data.SqlClient/tests/tools/Microsoft.Data.SqlClient.TestUtilities/Fixtures/ColumnMasterKeyCertificateFixture.cs
edwardneal: Yes. `SQLSetupStrategyAzureKeyVault` inherits from `SQLSetupStrategy`, which inherits from `ColumnMasterKeyCertificateFixture`. I'm unhappy with the inheritance hierarchy here, but shuffling the responsibilities around between the various SQLSetupStrategy classes seemed to need a deeper refactor than I wanted to do in this PR.

The same situation also assigns to `ColumnMasterKeyPath`.

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/ConversionTests.cs
benrr101: I'll do you one better and just move it to be a local variable
(the subsequent member variable can be made local, too)

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/CspProviderExt.cs
benrr101: Agreed, if we're just creating a CspParameters object to generate another CspParameters object, it seems unnecessary. I've also take the liberty to just roll the ClassData into a MemberData in the test class - it's only being used in this class for this one test, so it seems better to me to keep related test and test data together.

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/TestFixtures/AEConnectionStringProviderWithCspParameters.cs
benrr101: Deleted the file :)

src/Microsoft.Data.SqlClient/tests/tools/Microsoft.Data.SqlClient.TestUtilities/Fixtures/CertificateFixtureBase.cs
benrr101: Ok, I'll roll that into a comment for the const.

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/TestFixtures/SQLSetupStrategy.cs
benrr101: It is not. In fact, it is never set outside of the constructor, so it can be converted to a get-only propertly

benrr101:
@paulmedynski I addressed most of the comments you added, some a bit more heavily-handed (heavy-handidly?) than others. I also merged main into it b/c I couldn't build without a fix we added a looong time ago.

paulmedynski (approved):

mdaigle (approved):

PR #3085, opened 2024-12-18
Merge | Cleanup Pre 2005 Support in TdsParser

MichelZ:
Now that #2839 is merged, the Pre 2005 Code can be removed from netfx to align netfx & netcore code for future merge of TdsParser.

It also removes asserts that were used to gate 2005-only code

The code does not aim to remove Pre 2005 Code from other classes

This is working towards #2953

PR #3206, opened 2025-03-07
Merge | Cleanup Pre 2005 Support in TdsParser

benrr101:
_Rebuilding #3085 under the dotnet repo in order to complete full CI, original description is as follows_

---

Now that #2839 is merged, the Pre 2005 Code can be removed from netfx to align netfx & netcore code for future merge of TdsParser.

It also removes asserts that were used to gate 2005-only code

The code does not aim to remove Pre 2005 Code from other classes

This is working towards #2953


src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -273,13 +270,7 @@ internal EncryptionOptions EncryptionOptions
...
         }
 
-        internal bool Is2005OrNewer
-        {
-            get
-            {
-                return _is2005;
-            }
-        }
+        internal bool Is2005OrNewer => true;
```
mdaigle, line 273: This is confusing to me. Why keep the _is2005 above (and set to false) if we always return true here. I think I'd prefer to remove the field and adjust checks to look for _is2008. We can keep Is2005OrNewer to avoid refactors in other files for now.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -2860,7 +2850,6 @@ private TdsOperationStatus TryProcessEnvChange(int tokenLength, TdsParserStateOb
                     case TdsEnums.ENV_CHARSET:
                         // we copied this behavior directly from luxor - see charset envchange
                         // section from sqlctokn.c
-                        Debug.Assert(!_is2000, "Received ENV_CHARSET on non 7.0 server!");
```
mdaigle, line 2863: This is asserting that only 2000 or earlier hits this line. Should we remove the whole case? Or was this just a bad assert?

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -3095,7 +3080,6 @@ private TdsOperationStatus TryProcessEnvChange(int tokenLength, TdsParserStateOb
                         break;
 
                     case TdsEnums.ENV_USERINSTANCE:
-                        Debug.Assert(!_is2005, "Received ENV_USERINSTANCE on non 9.0 server!");
```
mdaigle, line 3098: Same here. It feels like these were just bad asserts.

mdaigle (changes requested):

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
edwardneal: MS-TDS says that ENVCHANGE types 3, 5 and 6 are only sent back to clients running TDS 7.0 or earlier ([src](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-tds/2b3eb7e5-d43d-4d1b-bf4d-76b9e3afc791).) I think that means we can remove ENV_CHARSET, ENV_LOCALEID and ENV_COMPFLAGS. In theory, I'd expect collation changes to be picked up by ENVCHANGE type 7.

ENV_USERINSTANCE was probably a bad assert. MS-TDS says it contains information for user instances, but that functionality was introduced in TDS 7.2.

paulmedynski (approved):

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
benrr101: Sounds good, if we connect it's implied we're connected to 2005 or newer. No reason to set state that indicates such.
There is another check deeper in the parser that disables MARS on <2005, so I will remove that as well (since we can't connect to <2005, MARS will always be enabled)

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
benrr101: Sounds good to me, I'll prune those as well

mdaigle (approved):

cheenamalhotra (approved):

PR #3208, opened 2025-03-10
Code Health | Make SqlDependencyProcessDispatcherStorage not `unsafe`

benrr101:
**Description**: Fairly small change, but worthy of its own PR. This PR changes the SqlDependencyProcessDispatcherStorage class to be safe (and not `unsafe`). The data for the property is now stored as an `IntPtr` which is the safe version of a `void*` that it was previously stored as. After changing that, we no longer need to use a `fixed` block to get an unsafe pointer to data.

The other improvement is that this PR removes the spin-lock and replaces it with a `lock` block.

**Testing**: Project still builds, and tests are still working.

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Interop/Windows/Sni/SqlDependencyProcessDispatcherStorage.netfx.cs
+++ b/src/Microsoft.Data.SqlClient/src/Interop/Windows/Sni/SqlDependencyProcessDispatcherStorage.netfx.cs
@@ -8,27 +8,24 @@ using System;
 using System.Diagnostics;
 using System.Runtime.InteropServices;
 using System.Runtime.Versioning;
-using System.Threading;
 
 namespace Interop.Windows.Sni
 {
-    internal unsafe class SqlDependencyProcessDispatcherStorage
+    internal class SqlDependencyProcessDispatcherStorage
     {
-        private static void* s_data;
+        private static readonly object s_lockObj = new();
+        private static IntPtr s_data;
         private static int s_size;
-        private static volatile int s_lock; // Int used for a spin-lock.
 
         [ResourceExposure(ResourceScope.Process)] // SxS: there is no way to set scope = Instance, using Process which is wider
         [ResourceConsumption(ResourceScope.Process, ResourceScope.Process)]
         public static byte[] NativeGetData()
         {
-            IntPtr ptr = (IntPtr)s_data;
-            
             byte[] result = null;
-            if (ptr != IntPtr.Zero)
+            if (s_data != IntPtr.Zero)
             {
                 result = new byte[s_size];
-                Marshal.Copy(ptr, result, 0, s_size);
+                Marshal.Copy(s_data, result, 0, s_size);
             }
 
             return result;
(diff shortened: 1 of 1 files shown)


Wraith2:
The new version is still using pointers and not safehandles. The difference between void* and IntPtr seems academic. The code is as safe as it was before it simply lacks the `unsafe` keyword but that doesn't make it better (or worse) simply different. In a future version of the runtime the new version may still require an `unsafe` context.

If it isn't broken then I don't see this adding value.

EgorBo:
While I agree that this PR in fact doesn't make code any safer, the hand-made SpinLock implementation is indeed better to be replaced by a lock IMO. The lock itself is optimized for cases without no contention (thin-lock) and it also does a little of spin-waiting under the hood first. The hand-made spin wait with Sleep(50) does look like it could cause troubles.

benrr101:
@Wraith2 I think the lock change stands on its own. Being able to remove `unsafe` keywords seems like an improvement to me, but maybe it's six of one, half dozen of the other.

The reason for not calling the lock change out in the first place was the code changes have been sitting around for a long time waiting for other PRs to go through. When I finally got to submitting it, I didn't really remember all the changes I made and only skimmed the diff.

EgorBo:
>  Being able to remove unsafe keywords seems like an improvement to me,

Like it was mentioned before, the new code will most likely be annotated with `unsafe` anyway, see https://github.com/dotnet/designs/pull/330

src/Microsoft.Data.SqlClient/src/Interop/Windows/Sni/SqlDependencyProcessDispatcherStorage.netfx.cs
```diff
@@ -38,29 +35,18 @@ public static byte[] NativeGetData()
...
-                {
-                    s_data = Marshal.AllocHGlobal(data.Length).ToPointer();
-
-                    Trace.Assert(s_data != null);
-
-                    Buffer.MemoryCopy(pDispatcher, s_data, data.Length, data.Length);
-
-                    Trace.Assert(0 == s_size); // Size should still be zero at this point.
+                    s_data = Marshal.AllocHGlobal(data.Length);
+                    Trace.Assert(s_data != IntPtr.Zero);
```
paulmedynski, line 43: I don't think this Assert() is helpful.  AllocHGlobal() either allocates memory successfully and returns an IntPtr to it, or it throws.  Granted, the AllocHGlobal() docs don't explicitly say the returned IntPtr.Zero will never be true, but so what if it is?  Will Copy() work if s_data.Zero is true?  Those docs don't say either way, so I'm not sure we should be making any assumptions here.  I think this Assert() gives a false sense of program state below (i.e. s_data.Zero == false).  If it's important that we never call Copy() with a Zero IntPtr, then we need an if-statement.

This is just FYI - no need to make any changes - just wanted to point it out.

paulmedynski (approved):

mdaigle (approved):
Even if it's still technically unsafe, less pointer type conversions and better locking still feel worthwhile.

PR #3209, opened 2025-03-11
Feature Extension Changes for Vector Datatype Support 

apoorvdeshmukh:
This PR adds feature extension support for vector datatype that allows MDS to fetch vector data in native binary format in which it is stored at the server.


src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/GenericTDSServer.cs
```diff
@@ -590,6 +598,32 @@ protected virtual TDSMessageCollection OnAuthenticationCompleted(ITDSServerSessi
...
             }
 
+            // Check if Vector is supported
+            if (session.IsVectorSupportEnabled)
+            {
+                // Create ack data (1 byte: Version number)
+                byte[] data = new byte[1];
+                data[0] = (byte)1;
+
+                // Create Json support as a generic feature extension option
```
Wraith2, line 608: json -> vector ?

Wraith2:
What is the format of vector data? what does it look like and what type would you expect GetFieldValue to return for it?

apoorvdeshmukh:
> What is the format of vector data? what does it look like and what type would you expect GetFieldValue to return for it?

@Wraith2 We are still in the process of updating the TDS document with the relevant details for vector data. These changes are meant for the feature branch, and we plan to update the TDS document and finalize the SqlClient APIs before we start preparing changes for the main. Will share the details soon.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -208,6 +208,9 @@ internal bool IsDNSCachingBeforeRedirectSupported
         // Json Support Flag
         internal bool IsJsonSupportEnabled = false;
 
+        //Vector Support Flag
```
paulmedynski, line 211: `// Vector Support Flag`

(Space between // and Vector)

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -2914,6 +2918,24 @@ internal void OnFeatureExtAck(int featureId, byte[] data)
                         break;
                     }
 
+                case TdsEnums.FEATUREEXT_VECTORSUPPORT:
+                    {
+                        SqlClientEventSource.Log.TryAdvancedTraceEvent("<sc.SqlInternalConnectionTds.OnFeatureExtAck|ADV> {0}, Received feature extension acknowledgement for VECTORSUPPORT", ObjectID);
+                        if (data.Length != 1)
+                        {
+                            SqlClientEventSource.Log.TryTraceEvent("<sc.SqlInternalConnectionTds.OnFeatureExtAck|ERR> {0}, Unknown token for JSONSUPPORT", ObjectID);
```
paulmedynski, line 2926: `JSONSUPPORT` -> `VECTORSUPPORT`

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -2914,6 +2918,24 @@ internal void OnFeatureExtAck(int featureId, byte[] data)
...
+                            throw SQL.ParsingError(ParsingErrorState.CorruptedTdsStream);
+                        }
+                        byte vectorSupportVersion = data[0];
+                        if (vectorSupportVersion == 0 || vectorSupportVersion > TdsEnums.MAX_SUPPORTED_JSON_VERSION)
```
paulmedynski, line 2930: `MAX_SUPPORTED_JSON_VERSION` -> `MAX_SUPPORTED_VECTOR_VERSION`

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -2914,6 +2918,24 @@ internal void OnFeatureExtAck(int featureId, byte[] data)
...
+                        {
+                            SqlClientEventSource.Log.TryTraceEvent("<sc.SqlInternalConnectionTds.OnFeatureExtAck|ERR> {0}, Invalid version number for JSONSUPPORT", ObjectID);
```
paulmedynski, line 2932: `JSONSUPPORT` -> `VECTORSUPPORT`

Should we also emit the byte value we received in hexadecimal, and the max version we support?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -8486,6 +8486,24 @@ internal int WriteFedAuthFeatureRequest(FederatedAuthenticationFeatureExtensionD
             return len;
         }
 
+        internal int WriteVectorSupportFeatureRequest(bool write /* if false just calculates the length */)
```
paulmedynski, line 8489: Can you add proper method comments?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -8486,6 +8486,24 @@ internal int WriteFedAuthFeatureRequest(FederatedAuthenticationFeatureExtensionD
...
+        {
+            int len = 6; // 1byte = featureID, 4bytes = featureData length, 1 bytes = Version
```
paulmedynski, line 8491: Let's prefer comments on their own line above the code they document.

paulmedynski, line 8491: const int

Also, any reason this isn't uint, since it can never be negative?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsEnums.cs
```diff
@@ -996,6 +998,9 @@ internal enum FedAuthInfoId : byte
         // JSON Support constants
         internal const byte MAX_SUPPORTED_JSON_VERSION = 0x01;
 
+        // JSON Support constants
```
paulmedynski, line 1001: JSON -> VECTOR

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/GenericTDSServer.cs
```diff
@@ -590,6 +598,32 @@ protected virtual TDSMessageCollection OnAuthenticationCompleted(ITDSServerSessi
                 }
             }
 
+            // Check if Vector is supported
+            if (session.IsVectorSupportEnabled)
```
cheenamalhotra, line 602: Can we write tests to enable Vector feature and assert on `IsVectorSupportEnabled`?

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/GenericTDSServer.cs
```diff
@@ -239,6 +239,14 @@ public virtual TDSMessageCollection OnLogin7Request(ITDSServerSession session, T
                                 session.IsJsonSupportEnabled = true;
                                 break;
                             }
+
+                        case TDSFeatureID.VectorSupport:
+                            {
+                                // Enable Vector Support
+                                session.IsVectorSupportEnabled = true;
```
paulmedynski, line 246: Are we missing a TDSServer option that forces vector support off?  How will we test that the driver _doesn't_ enable vector support after receiving the login ack without that?

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/GenericTDSServer.cs
paulmedynski: And a test that the client asks for Vector support, but the server says no.

paulmedynski: And tests where client max vector version > server max and vice versa.  Make sure they negotiate the highest common version.

paulmedynski (changes requested):
We need some tests that the feature negotiation works as expected.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/GenericTDSServer.cs
apoorvdeshmukh: @cheenamalhotra : Added testcase to validate the feature extension data being sent from client and received in the ack. 
@paulmedynski : Great suggestion! I'll incorporate this specific scenario for v2 implementation.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/GenericTDSServer.cs
apoorvdeshmukh: This will be taken care of in the v2 implementation of feature extension for vectors.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/GenericTDSServer.cs
apoorvdeshmukh: @paulmedynski I have added testcase for negotiating the highest common version.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlConnectionBasicTests.cs
```diff
@@ -524,5 +527,114 @@ public void ConnectionTestDeniedVersion(int major, int minor, int build)
 
             Assert.Throws<InvalidOperationException>(() => conn.Open());
         }
+
+        // Test that we send correct feature extension data for vector support
```
paulmedynski, line 531: Shouldn't we test these 2 scenarios:
- Client asks for Vector, server says no.
- Client asks for Vector, server says yes.

Also, I think you can combine this test and the next one.  They have very similar setup and checks.

```
[Theory]
[InlineData(true, 0x02, 0x01)]
[InlineData(false, 0x02, 0x01)]
...
public void ConnectionOpenVectorNegotiation(
    bool serverSupportsVector,
    byte serverVectorVersion,
    byte negotiatedVectorVersion)
{
    ...
}
```

This could then be extended to handle the future where MDS supports version 2 and we negotiate down to version 1.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/GenericTDSServer.cs
paulmedynski: Why wait until then?  Our initial vector support needs to handle the case where the server _doesn't_ support vectors.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/GenericTDSServer.cs
```diff
@@ -27,6 +27,50 @@ namespace Microsoft.SqlServer.TDS.Servers
     /// </summary>
     public class GenericTDSServer : ITDSServer
     {
+        /// <summary>
+        /// Delegate to be called when a LOGIN7 request has been received
+        /// and vector feature extension is present in the request.
+        /// This is called before any authentication work is done,
+        /// and before any response is sent.
+        /// </summary>
+        public delegate void OnLogin7VectorFeatureExtDelegate(
```
paulmedynski, line 36: I understand why you chose to make these new delegates very specific to Vector processing, since pulling the feature extension token from the Login7 packet is a bit of a pain.  However, it might be better to define delegates that return the Login7 packets and let the tests pick out what they care about.  That would be more generally useful.  If picking bits and pieces from a packet is too hard, maybe we should solve that problem instead.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/GenericTDSServer.cs
```diff
@@ -27,6 +27,50 @@ namespace Microsoft.SqlServer.TDS.Servers
...
+
+        /// <summary>
+        /// Version for vector FeatureExtension
+        /// </summary>
+        public const byte MaxSupportedVectorFeatureExtVersion = 0x01;
+
+        /// <summary>
+        /// Server version for vector FeatureExtension
+        /// </summary>
+        private byte _serverSupportedVectorFeatureExtVersion = MaxSupportedVectorFeatureExtVersion;
```
paulmedynski, line 56: Can we just use the property instead?
```
public byte ServerSupportedVectorFeatureExtVersion { private get; set; } = MaxSupportedVectorFeatureExtVersion;
```
I'm not sure why the getter is currently private though.

paulmedynski (changes requested):
I think we're still missing some tests for combinations of server support and version negotiation.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlConnectionBasicTests.cs
apoorvdeshmukh: Refactored the test case to include the scenario where server doesn't support vector feature extension

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/GenericTDSServer.cs
apoorvdeshmukh: Makes sense. Added delegates for Login7 request and TDSMessage response.

cheenamalhotra (approved):

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlConnectionBasicTests.cs
```diff
@@ -528,112 +529,91 @@ public void ConnectionTestDeniedVersion(int major, int minor, int build)
...
         // Test to verify that the server and client negotiate
         // the common feature extension version.
+        // MDS currently supports vector feature ext version 0x1.
         [Theory]
-        [InlineData(0x2, 0x1)] 
-        [InlineData(0x0, 0x0)] 
-        [InlineData(0x1, 0x1)] 
-        public void TestVectorFeatExtVersionNegotiation(byte serverVersion, byte expectedVersion)
+        [InlineData(true, 0x2, 0x1)]
+        [InlineData(false, 0x0, 0x0)]
```
paulmedynski, line 538: I don't think this case should fail to connection.open().  If the server sends back a version we don't recognize, I think we should treat that as not having vector support.  Imagine if SQL Server releases a version that misbehaves and sends vector VE v0.  We don't want to prevent MDS from connecting to it entirely.  We just want to disable vector support.

So I guess I'm saying, we should never reject a connection due to bad/missing/not-understood vector FE negotiation.  We should just disable vector support on the connection.

This is a philosophical change to your existing implementation, so we should probably discuss it with the team.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlConnectionBasicTests.cs
```diff
@@ -528,112 +529,91 @@ public void ConnectionTestDeniedVersion(int major, int minor, int build)
...
-                    actualFeatureExtVersion = optionToken.Data[0];
-                };
-
-            byte actualFeatureExtAckId = 0;
-            byte actualFeatureExtAckVersion = 0;
-
-            // Assign a delegate to save the feature extension ack token.
-            server.OnAuthenticationVectorFeatAckValidated =
-            (TDSFeatureExtAckGenericOption ackToken) =>
+            using (var server = TestTdsServer.StartTestServer())
```
paulmedynski, line 544: The code would be easier to read (less indented) if you avoid the using blocks:
```
using var server = TestTdsServer.StartTestServer();
...
using var connection = new SqlConnection(server.ConnectionString);
...
```
No nested blocks is preferred.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlConnectionBasicTests.cs
```diff
@@ -528,112 +529,91 @@ public void ConnectionTestDeniedVersion(int major, int minor, int build)
...
+                byte actualFeatureExtAckId = 0;
+                byte actualFeatureExtAckVersion = 0;
 
-            // Connect to the test TDS server.
-            using SqlConnection connection = new(server.ConnectionString);
-
-            if (expectedVersion != 0)
-            {
-                connection.Open();
+                server.OnLogin7Validated = loginToken =>
```
paulmedynski, line 555: I think we also need to remember if these callbacks were actually called _and_ the expected tokens found.  As-is, we can't tell the difference between not called, called but no token found, and called with token but values that match the initialized values above.
```
// Line 554
bool loginValuesFound = false;
bool responseValuesFound = false;

...

// Line 597
Assert.True(loginValuesFound);
Assert.True(responseValuesFound);
```

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/GenericTDSServer.cs
```diff
@@ -28,48 +28,41 @@ namespace Microsoft.SqlServer.TDS.Servers
...
-            TDSFeatureExtAckGenericOption vectorFeatExtAck);
-        public OnAuthenticationCompletedDelegate OnAuthenticationVectorFeatAckValidated { private get; set; }
+            TDSMessage response);
+        public OnAuthenticationCompletedDelegate OnAuthenticationResponseCompleted { private get; set; }
 
         /// <summary>
-        /// Version for vector FeatureExtension
+        /// Default feature extension version supported on the server for vector support.
         /// </summary>
         public const byte MaxSupportedVectorFeatureExtVersion = 0x01;
```
paulmedynski, line 50: Let's call this `DefaultVectorFeatureExtensionVersion`, since we can set whatever actual value we want via the property below.  This isn't enforced as a maximum.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/GenericTDSServer.cs
```diff
@@ -286,10 +279,12 @@ public virtual TDSMessageCollection OnLogin7Request(ITDSServerSession session, T
 
                         case TDSFeatureID.VectorSupport:
                             {
-                                // Enable Vector Support
-                                session.IsVectorSupportEnabled = true;
-                                _clientSupportedVectorFeatureExtVersion = ((TDSLogin7GenericOptionToken)option).Data[0];
-                                OnLogin7VectorFeatureValidated?.Invoke((TDSLogin7GenericOptionToken)option);
+                                if (ServerSupportedVectorFeatureExtVersion != 0xFF)
```
paulmedynski, line 282: What is this if-statement for?  Is 0xFF a magic value that _disables_ server vector support?

I think it would be clearer to have a specific boolean property to enable/disable vector support entirely.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/GenericTDSServer.cs
```diff
@@ -636,7 +633,7 @@ protected virtual TDSMessageCollection OnAuthenticationCompleted(ITDSServerSessi
             {
                 // Create ack data (1 byte: Version number)
                 byte[] data = new byte[1];
-                data[0] = _serverSupportedVectorFeatureExtVersion > _clientSupportedVectorFeatureExtVersion ? _clientSupportedVectorFeatureExtVersion : _serverSupportedVectorFeatureExtVersion;
+                data[0] = ServerSupportedVectorFeatureExtVersion > _clientSupportedVectorFeatureExtVersion ? _clientSupportedVectorFeatureExtVersion : ServerSupportedVectorFeatureExtVersion;
```
paulmedynski, line 636: Is this how SQL Server behaves?  Does it send back min(client FE version, server FE version), or does it just send back server FE version and let the client decide what to do?

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlConnectionBasicTests.cs
paulmedynski: Do we need to re-visit other FE negotiation code (i.e. JSON)?

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/GenericTDSServer.cs
apoorvdeshmukh: Yes. Server sends back the min(clientFE version, serverFE version).  The idea is clients will tell its version to the server and let the server decide what to do. This allows server to send data in a compatible format that client can understand.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlConnectionBasicTests.cs
apoorvdeshmukh: We follow same strategy for other FEs like JSON. Server is always expected to return feature data that is understood by the client. Getting a version that is not supported by client breaks that promise and we fail that connection.

David-Engel: > We follow same strategy for other FEs like JSON. Server is always expected to return feature data that is understood by the client. Getting a version that is not supported by client breaks that promise and we fail that connection.

This is the way. No valid server should violate the spec here, and if it does, we want it detected quickly. The server is responding that it's going to send vector data in an unknown way, despite what the client has requested, and whether we have disabled support on the client side or not. That will result in unexpected downstream TDS tokens/behavior.

paulmedynski: Ok that's the piece I was missing.  If we can't understand the FE data or version, then we can't reliably parse TDS, so we're hooped.  Makes sense.  No need to discuss further.

paulmedynski (approved):

PR #3213, opened 2025-03-12
Merge | Diagnostic Files

benrr101:
**Description**: More work as part of the merge project. This one is pretty straightforward - the SqlDiagnostic* classes only belonged to the netcore project, so this PR:

* Moves the files over to the common project in a Diagnostics folder
* Split the files into one class per file
* Moved classes that weren't in the SqlClient.Diagnostics namespace into that namespace
* Moved the extension methods for DiagnosticsListener class into the DiagnosticListener class
  * I can revert this if there's significant backlash to the change, but considering the code is all internal, to me, there's no real reason to keep these as extension methods.
* Also created stubs for SqlCommand and SqlConnection in common project to make it a bit easier to work on the Diagnostic files.

**Testing**: No functional changes to the code, CI should be able to validate.


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnection.stub.cs
cheenamalhotra, line 1: Is this needed in this PR for merging Diagnostics or can be done later?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlClientDiagnostic.cs
cheenamalhotra, line 1: I don't see a need to split this class into subclasses since all of them have a single purpose and will not likely see any major works in foreseeable future.
This class could stay in single file targeted for netcore project.

benrr101: No, I disagree, it should be properly split into separate files. It messes with developer expectations to have multiple classes within a single file. C# standards suggest each class should get its own file. There are a lot of bad examples in this codebase where multiple classes have been jammed into the same file, and it makes it much more difficult to know where each class is supposed to be.
If it truly is only used within the confines of another class, maybe we consider making it a nested class.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnection.stub.cs
benrr101: This is only added to aid the migration process - while I'm working within the common project, it resolves missing methods/classes so I can work without red squiggles everywhere. This will be removed when the sqlconnection class is merged.

paulmedynski (approved):

PR #3217, opened 2025-03-13
Complete removal of SQL 2000 support

edwardneal:
This finishes the work from #3206 and prior PRs. It removes the `Is2005OrNewer` and `Is2000` properties from `SqlCommand`, `SqlConnection`, `SqlInternalConnectionSmi`, `SqlInternalConnectionTds`, `SqlInternalConnection` and `TdsParser`, then chases those references through the codebase.

I'd appreciate a CI run against this, although I don't think there'll be any surprises here - all of these code paths should be dead.


src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
```diff
@@ -549,7 +531,7 @@ private string AnalyzeTargetAndCreateUpdateBulkCommand(BulkCopySimpleResultSet i
 
             StringBuilder updateBulkCommandText = new StringBuilder();
 
-            if (_connection.Is2000 && 0 == internalResults[CollationResultId].Count)
+            if (0 == internalResults[CollationResultId].Count)
```
cheenamalhotra, line 534: Removing this condition may trigger this error where it wouldn't have before.. can you remove this check instead?

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlInternalConnectionSmi.cs
```diff
@@ -162,22 +162,6 @@ override internal bool IsLockedForBulkCopy
...
-
-        override internal bool Is2005OrNewer
-        {
-            get
-            {
-                return true;    // Must be direct connecting to 2005 or newer.
-            }
-        }
-
         override internal bool Is2008OrNewer
```
cheenamalhotra, line 165: Not removing this?

edwardneal: Preferably not in this PR. The last major release only barred connections to SQL 2000; it's still technically possible (just very far out of support) to connect to SQL 2005 without the library throwing an exception.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnection.cs
```diff
@@ -2538,12 +2514,15 @@ private static void ChangePassword(string connectionString, SqlConnectionString
             // Normally we would simply create a regular connectoin and open it but there is no other way to pass the
             // new password down to the constructor. Also it would have an unwanted impact on the connection pool
             //
-            using (SqlInternalConnectionTds con = new SqlInternalConnectionTds(null, connectionOptions, credential, null, newPassword, newSecurePassword, false, null, null, null, null))
+            SqlInternalConnectionTds con = null;
```
paulmedynski, line 2517: Ahh yes, the weird C# requirement that you must _name_ a resource when using, but then it complains when you don't actually _use_ the named resource (because you only want the side effects of acquisition).

So now we have to manually write the (fragile, error prone) using/disposal code to avoid unused variable warnings :)

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnection.cs
```diff
@@ -2538,12 +2514,15 @@ private static void ChangePassword(string connectionString, SqlConnectionString
...
             {
-                if (!con.Is2005OrNewer)
-                {
-                    throw SQL.ChangePasswordRequires2005();
-                }
+                con = new SqlInternalConnectionTds(null, connectionOptions, credential, null, newPassword, newSecurePassword, false, null, null, null, null);
+            }
+            finally
+            {
+                if (con != null)
```
paulmedynski, line 2524: Can you add braces here, or use null-conditional `con?.Dispose()` ?

edwardneal: Yes - fixed. I've updated the netcore code to keep it in sync.

paulmedynski (approved):

cheenamalhotra (approved):

Issue #3188, opened 2025-02-28
Deprecate Authentication=ActiveDirectoryPassword


David-Engel:
Given the direction of Azure authentication and enforcement of mandatory MFA, we should deprecate the ActiveDirectoryPassword Authentication option on SqlConnections. MSAL.NET and Azure.Identity are also going to deprecate their corresponding APIs.

For more details of the Azure plan, see [Planning for mandatory multifactor authentication for Azure and other admin portals](https://learn.microsoft.com/en-us/entra/identity/authentication/concept-mandatory-multifactor-authentication?tabs=dotnet)

I imagine this includes annotating:

```
public enum SqlAuthenticationMethod
{
    [Obsolete("ActiveDirectoryPassword is deprecated.")]
    ActiveDirectoryPassword
```

Updating the API docs (SqlAuthenticationMethod.xml), public docs, and including it in the next release announcement.

Let me know if I'm missing anything else.

ErikEJ:
Maybe some connection string docs?

mdaigle:
I added a task to our backlog to investigate.

David-Engel:
Reopened to make sure we don't forget to redo this for 7.0.

PR #3219, opened 2025-03-14
Deprecate 'Active Directory Password' authentication mode.

cheenamalhotra:
Addresses #3188

### DIFF
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectivityTests/AADConnectionTest.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectivityTests/AADConnectionTest.cs
@@ -49,12 +49,12 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests
                 return new SqlAuthenticationToken(result.AccessToken, result.ExpiresOn);
             }
 
-            #pragma warning disable 0618
             public override bool IsSupported(SqlAuthenticationMethod authenticationMethod)
             {
+                #pragma warning disable 0618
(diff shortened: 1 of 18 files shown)


src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectivityTests/AADConnectionTest.cs
```diff
@@ -49,10 +49,12 @@ public override async Task<SqlAuthenticationToken> AcquireTokenAsync(SqlAuthenti
                 return new SqlAuthenticationToken(result.AccessToken, result.ExpiresOn);
             }
 
+            #pragma warning disable 0618
```
paulmedynski, line 52: Should we tighten these #pragmas up to include only the ActiveDirectoryPassword symbol?  There are a few other places where we're including unrelated symbols in the #pragma as well:
```
    return authenticationMethod.Equals(
#pragma warning disable 0618
        SqlAuthenticationMethod.ActiveDirectoryPassword
#pragma warning restore 0618
    );
```

paulmedynski (changes requested):
Overall looks good.  Asking to think about narrowing #pragma scope.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectivityTests/AADConnectionTest.cs
cheenamalhotra: Sure

David-Engel (approved):

paulmedynski (approved):

PR #2985, opened 2024-11-07
Merge | Align TdsExecuteRPC between netcore and netfx

MichelZ:
Splitting up TdsExecuteRPC to multiple functions to align netcore and netfx

Part of #2953

PR #3222, opened 2025-03-17
Merge | TdsParser cosmetic changes

edwardneal:
This cleans up the diff of TdsParser. Contributes to #2953.

I recommend [disabling whitespace](https://github.com/dotnet/SqlClient/pull/3222/files?w=1) when viewing this, it'll bring the PR size down to +239/-148.

Most of TdsParser is pretty similar. We've refactored a few methods, some of the members are in different orders and there are other cosmetic differences though, so this needs to be aligned before we can see where the true differences lie. This PR is the first of a few which do that. I'm fairly sure that once we've done that, only a few dozen lines of code will truly be different.

This dovetails with #2985, which handles some of the refactoring.

If someone from the SqlClient team can run CI, this should be enough to validate it. I don't expect that anything will have changed in the compiled DLL.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
@@ -1007,7 +1007,8 @@ namespace Microsoft.Data.SqlClient
                             ON,
                             NOT_SUP,
                             REQ,
-                            LOGIN
+                            LOGIN,
+                            OPTIONS_MASK = 0x3f
                         } */
 
                         // Any response other than NOT_SUP means the server supports encryption.
@@ -6382,7 +6383,7 @@ namespace Microsoft.Data.SqlClient
                     }
                     else
                     {
-                        value.SqlBinary = SqlBinary.WrapBytes(b);
+                        value.SqlBinary = SqlBinary.WrapBytes(b); // doesn't copy the byte array
                     }
                     break;
 
@@ -6741,7 +6742,7 @@ namespace Microsoft.Data.SqlClient
                         {
                             return result;
                         }
-                        value.SqlBinary = SqlBinary.WrapBytes(b);
+                        value.SqlBinary = SqlBinary.WrapBytes(b); // doesn't copy the byte array
 
                         break;
                     }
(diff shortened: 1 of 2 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -4619,63 +4642,84 @@ internal int GetCodePage(SqlCollation collation, TdsParserStateObject stateObj)
 
         internal void DrainData(TdsParserStateObject stateObj)
         {
+#if NETFRAMEWORK
+            RuntimeHelpers.PrepareConstrainedRegions();
```
paulmedynski, line 4646: I thought @benrr101 recently did some work to remove constrained regions.  Should we be adding them back here, removing them from the netfx code, or keeping these parts of the netcore and netfx files different in purpose?

paulmedynski (changes requested):
Need comment from @benrr101 regarding constrainted regions.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
benrr101: Close! - the changes I did more recently was to remove "reliability sections" which were a thing we invented for debug builds to validate certain objects were being properly disposed (or something like that - they're gone now, so I forget). Constrained regions remain (netfx only, and really only here to prevent SQL Server from crashing when referencing MDS from CLR objects)

It's a little odd to add `#if NETFRAMEWORK` sections to the netcore project (since they'd never be included), but I get it from the perspective of making the netcore and netfx projects look more similar. I'm not entirely sure about this merge strategy (eg, this hasn't particularly made it easier for sqltransaction merge), but I see where it's coming from.

edwardneal: Thanks. I'm trying to make sure that everything's in place to merge TdsParser by simply deleting the netfx file and moving the netcore version, but it definitely looks odd in the interim.

I considered moving identical methods into a partial class in the shared directory, but that's easily ~8k LOC - we'd need 16-20 PRs to make that reviewable.

paulmedynski (approved):

mdaigle (approved):

PR #3224, opened 2025-03-17
Release notes for v6.0.2

cheenamalhotra:
Release notes for milestone: [v6.0.2](https://github.com/dotnet/SqlClient/milestone/68)

### DIFF
--- a/release-notes/6.0/6.0.2.md
+++ b/release-notes/6.0/6.0.2.md
@@ -1,6 +1,6 @@
 # Release Notes
 
-## [Stable release 6.0.2] - 2025-04-24
+## [Stable release 6.0.2] - 2025-04-25
 
 This update brings the below changes over the previous stable release:
 
--- a/release-notes/6.0/6.0.2.md
+++ b/release-notes/6.0/6.0.2.md
@@ -1,6 +1,6 @@
 # Release Notes
 
-## [Stable release 6.0.2] - 2025-03-19
+## [Stable release 6.0.2] - 2025-04-09
 
 This update brings the below changes over the previous stable release:
 
--- a/release-notes/6.0/6.0.2.md
+++ b/release-notes/6.0/6.0.2.md
@@ -1,6 +1,6 @@
 # Release Notes
 
-## [Stable release 6.0.2] - 2025-04-09
+## [Stable release 6.0.2] - 2025-04-17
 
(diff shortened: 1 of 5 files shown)


release-notes/6.0/6.0.2.md
```diff
@@ -0,0 +1,55 @@
...
+- Microsoft.Bcl.Cryptography 8.0.0
+- Microsoft.Data.SqlClient.SNI 6.0.2
+- Microsoft.Extensions.Caching.Memory 8.0.1
+- Microsoft.IdentityModel.JsonWebTokens 7.5.0
+- Microsoft.IdentityModel.Protocols.OpenIdConnect 7.5.0
+- System.Buffers 4.5.1
+- System.Security.Cryptography.Pkcs 8.0.1
+- System.Text.Encodings.Web 6.0.0
+- System.Text.Json 6.0.10
+- System.Data.Common 4.3.0
```
mdaigle, line 31: Curious about this one. I don't see it mentioned in any of the release notes since 2.0. It's in our nuspec file, but it doesn't show on nuget.org.

cheenamalhotra: It doesn't show up because it wasn't included in nuspec before, while it's a required dependency, it should have.

cheenamalhotra: That might be something to do with how NuGet manages dependencies and type forwards, but it's part of the NuGet package dependency: 
image

mdaigle: When I follow any System.Data.Common references in the netfx project through to their source, they say that they're coming from the System.Data dll within .NET framework. And I only see the package referenced in the ref project and the common project.

cheenamalhotra: The type '[IDbColumnSchemaGenerator](https://learn.microsoft.com/en-us/dotnet/api/system.data.common.idbcolumnschemagenerator?view=net-9.0)' is only available in [System.Data.Common](https://www.nuget.org/packages/System.Data.Common/) NuGet package for .NET Framework. This is one of the types that was ported back to .NET Framework through a DLL that type-forwards to System.Data.dll all other types in the namespace..

mdaigle (approved):

paulmedynski (approved):

PR #3229, opened 2025-03-19
Update dependencies to latest

cheenamalhotra:


paulmedynski (approved):

mdaigle (approved):

mdaigle (approved):

benrr101 (changes requested):
Looks like they made some changes to the behavior in the xunit utilities... Here's the error message on testset 3

> Tests marked with Timeout are only supported for async tests

Just let me know if you don't have time to get to it.

benrr101 (approved):

PR #3230, opened 2025-03-19
Tidy up performance tests

paulmedynski:
- Tidied up the existing Performance benchmarks.
- Added environment variables to specify config files.
- Removed Newtonsoft.Json as a dependency.
- Simplified csproj a bit.

### DIFF
--- a/BUILDGUIDE.md
+++ b/BUILDGUIDE.md
@@ -342,11 +342,30 @@ dotnet test <test_properties...> --collect:"XPlat Code Coverage"
 ### Running Performance test project directly
 
 Project location from Root: `src\Microsoft.Data.SqlClient\tests\PerformanceTests\Microsoft.Data.SqlClient.PerformanceTests.csproj`
-Configure `runnerconfig.json` file with connection string and preferred settings to run Benchmark Jobs.
+
+Create an empty database for the benchmarks to use:
+```bash
+> create database [sqlclient-perf-db]
+> go
+```
+
+Configure `runnerconfig.json` file with connection string and preferred settings
+to run Benchmark Jobs.
 
 ```bash
 cd src\Microsoft.Data.SqlClient\tests\PerformanceTests
 dotnet run -c Release -f net8.0
+```
+
+Optionally, to avoid polluting your git workspace, copy `runnerconfig.json` to a
+new file, make your edits there, and then specify the new file with the
+RUNNER_CONFIG environment variable.
+
+```bash
+cd src/Microsoft.Data.SqlClient/tests/PerformanceTests
+cp runnerconfig.json ~/.configs/runnerconfig.json
+# Make edits to ~/.configs/runnerconfig.json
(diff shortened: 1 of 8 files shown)


src/Microsoft.Data.SqlClient/tests/PerformanceTests/Program.cs
```diff
@@ -2,22 +2,20 @@
...
 
 namespace Microsoft.Data.SqlClient.PerformanceTests
 {
     public class Program
     {
-        private static Config s_config;
+        private readonly Config _config;
 
-        public static void Main()
+        public Program()
```
paulmedynski, line 14: Minimize the static stuff by embracing Program's inherent object-ness.

tools/props/Versions.props
```diff
@@ -53,7 +53,7 @@
   </PropertyGroup>
   <!-- Test Project Dependencies -->
   <PropertyGroup>
-    <BenchmarkDotNetVersion>0.13.2</BenchmarkDotNetVersion>
+    <BenchmarkDotNetVersion>0.14.0</BenchmarkDotNetVersion>
```
paulmedynski, line 56: Needed to update BenchmarkDotNet so it would work with .NET 9 SDK/runtime.  No other projects use this.

src/Microsoft.Data.SqlClient/tests/PerformanceTests/Config/Loader.cs
```diff
@@ -0,0 +1,78 @@
...
+              : Environment.GetEnvironmentVariable(envOverride)
+                ?? path;
+
+            T config = null;
+            Exception error = null;
+            try
+            {
+                using var stream = File.OpenRead(configFile);
+                config =
+                    JsonSerializer.Deserialize<T>(
```
paulmedynski, line 55: Deserialize<T>() returns a nullable T, but doesn't document if/when null may be returned.  This seems to be a pervasive oversight in much of the .NET documentation.  As a result, we must check for null _and_ exceptions.

mdaigle (approved):

BUILDGUIDE.md
```diff
@@ -342,11 +342,30 @@ dotnet test <test_properties...> --collect:"XPlat Code Coverage"
...
 
+Optionally, to avoid polluting your git workspace, copy `runnerconfig.json` to a
+new file, make your edits there, and then specify the new file with the
+RUNNER_CONFIG environment variable.
+
+```bash
+cd src/Microsoft.Data.SqlClient/tests/PerformanceTests
+cp runnerconfig.json ~/.configs/runnerconfig.json
+# Make edits to ~/.configs/runnerconfig.json
+RUNNER_CONFIG=~/.configs/runnerconfig.json dotnet run -c Release -f net9.0
```
ErikEJ, line 368: Is this a single line? On Windows?

paulmedynski: The existing example claimed to be bash, so I assumed Linux.  I will add examples for both shells.

samsharma2700 (approved):

Issue #2210, opened 2023-11-07
Emit OpenTelemetry tracing


roji:
OpenTelemetry has become the de-facto standard for distributed, cross-platform tracing; it is being rapidly adopted both inside the .NET ecosystem and outside. OTel defines semantic conventions for how a database client emits information; [these specifications](https://opentelemetry.io/docs/specs/semconv/database/database-spans/) are currently in experimental state, but stabilization is likely start happening soon.

There is a separate instrumentation library, [OpenTelemetry.Instrumentation.SqlClient](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Instrumentation.SqlClient/README.md#steps-to-enable-opentelemetryinstrumentationsqlclient), which AFAIK uses the DiagnosticSource instrumentation already emitted by SqlClient to then emit OpenTelemetry tracing.

However, ideally SqlClient would emit this tracing data directly, without requiring a 3rd-party library.

Wraith2:
Why is it better to emit third party specific information rather than have their adapter deal with the diagnostic information we already emit?

roji:
@Wraith2 this isn't 3rd-party specific information, but rather the emerging standard for how database clients emit tracing information; fundational components in the .NET ecosystem (such as ASP.NET itself, but also Npgsql) are being instrumented to emit OpenTelemetry tracing directly, which is the best user experience and guarantees both accurate tracing and best performance. This really belongs in the driver rather than some 3rd-party adapter which may or may not have access to the needed information in an efficient way.

Wraith2:
If there is additional information needed in the diagnostic data that we emit then we should identify and add it so that it can be available to the open telemetry adapter. I've already got an [issue open](https://github.com/open-telemetry/opentelemetry-dotnet/issues/5011) at open telemetry about changing the diagnostic information to be strongly typed so it's more aot friendly so adding any new information would be best done along with that work.

I don't see why it's beneficial to have open telemetry as s direct dependency/implementation inside sqlclient though. It adds complexity and another set of diagnostic logging (we already have 2) and isn't needed by many people. If we want to make things pay-for-play then an external adapter that people can opt into sounds like the best approach to me.

roji:
> I don't see why it's beneficial to have open telemetry as s direct dependency/implementation inside sqlclient though.

[...]

PR #3231, opened 2025-03-19
Merge | SqlTransaction

benrr101:
**Description**: In the latest installment of "Ben Tries To Combine Two Projects Into One", we are merging the SqlTransaction class. This is actually a pretty easy merge - there are ~4 methods that need to be merged into a single file in the common project.

The big headache with these are that they use the constrained execution region pattern that only applies to netfx - this means each of these methods have an entirely different set of exception handling for special exceptions that can only happen in netfx constrained execution regions. However, the netcore implementation has a diagnostic listener that expects an exception to be set on it if one occurs. This presents a challenge of how to retain these differences while also making it readable. There are two patterns we could adopt:

1) All exception handles have a conditional for netcore to set the diagnostic listener eg:
  ```c#
  //...
  catch (OutOfMemoryException e) {
    #if NET
    diagnosticListener.SetException(e)
    #endif

    _connection.Abort();
   throw;
  } catch (ThreadAbortException e ) {
    // .. repeat from above
  } catch (Exception e) {
    #if NET
    diagnosticListener.SetException(e);
    #endif
    throw;
  }
  ```
2) Conditional exception handles for netfx/netcore
  ```C#
  #if NETFRAMEWORK
  catch (OutOfMemoryException e) {
    _connection.Abort();
    throw;
  } catch (ThreadAbortException e) {
    // .. repeat from above
  }
  #else
  catch (Exception e) {
    diagnosticListener.SetException(e);
    throw;
  }
  #endif
  ```

In my humble opinion, option 2 is somewhat better, and it is the pattern I have adopted in this PR.

The other merge as part of this PR is moving SqlTransaction.Common back into the SqlTransaction file. Thus, no more partial for SqlTransaction. The Common file looks to be an early attempt to merge what could be merged, so since we have reconciled the differences, we can bring them all together again.

Testing: Once again, no functional differences, just code moving around.

@edwardneal Please don't be upset if I sniped your next merge target 😄


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlTransaction.cs
```diff
@@ -0,0 +1,468 @@
...
+
+        /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlTransaction.xml' path='docs/members[@name="SqlTransaction"]/Commit/*' />
+        public override void Commit()
+        {
+            #if NET
+            using DiagnosticTransactionScope diagnosticScope = s_diagnosticListener.CreateTransactionCommitScope(
+                _isolationLevel,
+                _connection,
+                InternalTransaction);
+            #endif
```
edwardneal, line 96: No action here as part of this PR, but this is only going to get uglier as we progress through SqlConnection and SqlCommand...

What do you think about lifting this design (along with TryEventScope.Create, TryCorrelationTraceEvent, etc.) into a common tracing interface before merging classes which perform a lot of logging, and having the common tracing interface call the diagnostic listener or ETW event source (and eventually #2210)?

samsharma2700 (approved):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlTransaction.cs
cheenamalhotra: Would like to support anything needed to move faster and remove impediments, however I want to understand how adding another tracing layer benefits us to merge better and what are your thoughts and proposal. As I see usage of TryEventScope.Event and similar ETW events are applicable to both netfx/netcore projects so when merging them, we should consider picking newer trace patterns (from netcore project).

cheenamalhotra (approved):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlTransaction.cs
benrr101: Hmmm. That's a tough question. My gut reaction is that it's beyond the scope of the project merge, especially since it will still require rewriting the existing logging. I definitely agree that these are getting ugly, and I'd like to do what I can to reduce the number of #if's we have to use.

I'm also considering something that we can call in both netfx and netcore but make it clear it only applies to netfx or netcore (kinda like how Debug.* only applies when DEBUG is defined). It might be a better stop-gap for the merge project since opentelemetry it more/less happening in parallel.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlTransaction.cs
```diff
@@ -0,0 +1,468 @@
...
+using System.ComponentModel;
+using System.Data;
+using System.Data.Common;
+using System.Diagnostics;
+using System.Threading;
+using Microsoft.Data.Common;
+
+#if NET
+using Microsoft.Data.SqlClient.Diagnostics;
+#else
```
paulmedynski, line 15: Is the intention here to use CompilerServices if NET _isn't_ defined, or if NETFRAMEWORK _is_ defined?

I don't think those are the same thing.  The absence of NET doesn't imply NETFRAMEWORK.

So I think maybe this would be better:
```
#if NET
using Microsoft.Data.SqlClient.Diagnostics;
#endif

#if NETFRAMEWORK
using System.Runtime.CompilerServices;
#endif
```

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlTransaction.cs
```diff
@@ -0,0 +1,468 @@
...
+using System.Runtime.CompilerServices;
+#endif
+
+namespace Microsoft.Data.SqlClient
+{
+    /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlTransaction.xml' path='docs/members[@name="SqlTransaction"]/SqlTransaction/*' />
+    public sealed class SqlTransaction : DbTransaction
+    {
+        #if NET
+        private static readonly SqlDiagnosticListener s_diagnosticListener = new(SqlDiagnosticListener.DiagnosticListenerName);
```
paulmedynski, line 25: Would it be better to make SqlDiagnosticListener exist for .NET Framework as well, but just have empty function bodies (put #if NET around each body)?  Then you could just call it regardless, and let the compiler elide the empty function calls for .NET Framework.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlTransaction.cs
```diff
@@ -0,0 +1,468 @@
...
+            {
+                SqlStatistics statistics = null;
+
+                SqlClientEventSource.Log.TryCorrelationTraceEvent(
+                    "SqlTransaction.Commit | API | Correlation | Object Id {0}, Activity Id {1}, Client Connection Id {2}",
+                    ObjectId,
+                    ActivityCorrelator.Current,
+                    Connection?.ClientConnectionId);
+
+                #if NETFRAMEWORK
```
paulmedynski, line 110: Heh - GitHub's syntax highlighting doesn't make these obvious, so now they're a bit hidden due to indentation.  Can't win!

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlTransaction.cs
```diff
@@ -0,0 +1,468 @@
...
+                #if NET
+                catch (Exception ex)
+                {
+                    diagnosticScope.SetException(ex);
+                    throw;
+                }
+                #endif
+                finally
+                {
+                    SqlStatistics.StopTimer(statistics);
```
paulmedynski, line 168: You could consolidate the SetException() calls here if you save the exception in the catch blocks.

That would re-order operations in the timeout case above, but is that important?

paulmedynski, line 168: Should we stop the timer as early as possible?  We're currently measuring all of the time spent handling exceptions.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlTransaction.cs
```diff
@@ -0,0 +1,468 @@
...
+
+                    if (!IsZombied && !Is2005PartialZombie)
+                    {
+                        InternalTransaction.Dispose();
+                    }
+                }
+                catch (OutOfMemoryException e)
+                {
+                    _connection.Abort(e);
+                    throw;
```
paulmedynski, line 197: Do we expect Abort() to succeed if there is no memory?  Can we even call Abort() if the stack just overflowed?

What is the caller supposed to do if Dispose() throws?  Seems strange to allow disposal to throw anything.  Dispose() is your last chance to cleanup (assuming the runtime is behaving) - there's no "try again".

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlTransaction.cs
```diff
@@ -0,0 +1,468 @@
...
+                    #if NETFRAMEWORK
+                    bestEffortCleanupTarget = SqlInternalConnection.GetBestEffortCleanupTarget(_connection);
+                    #endif
+
+                    if (!IsZombied && !Is2005PartialZombie)
+                    {
+                        InternalTransaction.Dispose();
+                    }
+                }
+                catch (OutOfMemoryException e)
```
paulmedynski, line 194: No \#if NETFRAMEWORK around these exceptions like in the other methods?  Here and below in Save().

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlTransaction.cs
```diff
@@ -0,0 +1,468 @@
...
+                        _isFromApi = false;
+                    }
+                }
+            }
+        }
+
+        /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlTransaction.xml' path='docs/members[@name="SqlTransaction"]/RollbackTransactionName/*' />
+        #if NET
+        public override void Rollback(string transactionName)
+        #else
```
paulmedynski, line 303: Another case where `#if ! NET` != `#if NETFRAMEWORK` .  Here and below on Save().

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlTransaction.cs
```diff
@@ -0,0 +1,468 @@
...
+                    bestEffortCleanupTarget = SqlInternalConnection.GetBestEffortCleanupTarget(_connection);
+                    #endif
+
+                    statistics = SqlStatistics.StartTimer(Statistics);
+
+                    _isFromApi = true;
+
+                    InternalTransaction.Commit();
+                }
+                #if NETFRAMEWORK
```
paulmedynski, line 126: What's the harm in removing the \#if here?  C# doesn't have checked exceptions, so maybe worst case we end up with some bytecode in a .NET runtime that never runs.

paulmedynski (changes requested):
Some comments/questions about preprocessor directives that are likely relevant to this PR.  Other comments about existing code/patterns that we could maybe improve upon since we're here already.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlTransaction.cs
benrr101: The way I've understood it for so long is:
On netcore (net8.0, net9.0, etc), NET is defined, NETFRAMEWORK is not
On netfx (net462, etc) NETFRAMEWORK is defined, NET is not.

See also https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives

So, for our purposes (net462, net8.0, and net9.0), I'd have to disagree with you - absence of NET does imply NETFRAMEWORK

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlTransaction.cs
benrr101: This has been suggested before, but I'm fairly strongly against it. If I were an engineer coming into the code fresh, I would be bamboozled to find out that the netfx implementation of the SqlDiagnosticListener does nothing. Having code that subverts expectations is bad practice, imo.

That being said, there is precedent for code in dotnet that does nothing if a flag isn't set, eg, `Debug.Assert`. I'm ok with this because it's understood that if `DEBUG` isn't set, `Debug.Assert` does nothing. I would be ok with implementing a similar situation for the SqlDiagnosticListener if we made it clear that it only applied to netfx or netcore from the name of the class/method.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlTransaction.cs
benrr101: heh, yeah ... which is why I want to avoid using them as much as possible!

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlTransaction.cs
benrr101: Well, because as stated in the description, if I want to make sure that all exceptions do the same thing in netcore (since netcore only has the Exception catch-all), each exception handler has to have an
```
#if NET
diagnosticScope.SetException(e)
#endif
```
which to me is worse than having one big #if block that adds exception handlers for only netfx.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlTransaction.cs
benrr101: To be honest, I'm not sure how important the order of operations is in this case. It's one of those things where I don't fully understand how it needs to come together, but I also don't really want to mess with it.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlTransaction.cs
benrr101: Well, that was what the merge gave me... But, I also know @edwardneal was trying to unite the implementation of these prior to merging. I think in his changes, the goal was to leave the exception blocks for both netfx and netcore (which I kinda reverted). I suppose the better choice here would be to follow suit and have these wrapped in #if like the rest of them.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlTransaction.cs
benrr101: I'm not 1000% certain about this. CER seems to be some interesting magic baked in at a very low level of the framework. As per [this stack overflow](https://stackoverflow.com/questions/1101147/code-demonstrating-the-importance-of-a-constrained-execution-region), with constrained execution regions, the exception handlers are specially treated to allow execution of "safe things" before blowing exploding.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlTransaction.cs
benrr101: As per above comment, I disagree :)

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlTransaction.cs
benrr101: Oh - for this specific one, if I wrap the exception handlers in #if NETFRAMEWORK, then there won't be any exception handlers, which means I'd need to wrap the try in a #if NETFRAMEWORK, too - which .... is just more ugly that I can sign onto.

But I can apply the change to the Save

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlTransaction.cs
edwardneal: I'm thinking mainly about the current `SqlDiagnosticsListener` / `SqlClientEventSource` split.

At the moment, we have `TryEventScope.Create` in [SqlClientEventSource.cs](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlClientEventSource.cs#L1170-L1191), which returns a struct which records to the ETW event source when the instance enters and leaves scope. This is fairly similar to objects like [`DiagnosticScope.CreateCommandScope`](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Diagnostics/DiagnosticScope.netcore.cs), to the point where SqlCommand.ExecuteScalar calls both methods on [netcore](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs#L1127-L1128) and only TryEventScope.Create on [netfx](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlCommand.cs#L1313). The core difference is that when the return value of DiagnosticScope.CreateCommandScope is disposed of, it writes to the SqlDiagnosticListener.

The SqlDiagnosticListener type is only available on netcore, so dealing with that difference at the point of logging seems like it'll become difficult pretty quickly. It might be cleaner to expose DiagnosticScope.CreateCommandScope and its sibling types to both projects, then make the writing to SqlDiagnosticsListener / SqlClientEventSource an implementation detail, locked behind `#if` within those siblings.

Hopefully we could do this piecemeal - to use the example of SqlTransaction.Commit, we'd use `DiagnosticTransactionScope.CreateTransactionCommitScope`, and this would replace the inner `TryEventScope` and the manual call to `TryCorrelationTraceEvent`; DiagnosticTransactionScope would be responsible for logging to SqlClientEventSource, then (on netcore) for logging to SqlDiagnosticListener.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlTransaction.cs
paulmedynski: I got the willies when I saw this in our netfx csproj:
```
    <DefineConstants>$(DefineConstants);NETFRAMEWORK;</DefineConstants>
```
I'm not truly concerned about \#else vs \#if NETFRAMEWORK, but I thought it was worth some discussion.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlTransaction.cs
paulmedynski: It's a question of sprinkling \#if throughout the code vs having it in a single, well-documented place.  I agree that Surprising Behaviour ® should be avoided, but sometimes it's superior to copy-pasta.  Let's keep this in mind for some future code cleanup.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlTransaction.cs
paulmedynski: Yeah, something to consider for future cleanup.  It just caught my eye.

paulmedynski: Another "caught my eye" thing that I thought someone might know whether or not it's intentional (unlikely?).

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlTransaction.cs
paulmedynski: Wild stuff, let's leave it alone for now.  In general though, I think we need to be careful about throwing drom disposal methods.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlTransaction.cs
paulmedynski: Interesting - Are these exceptions documented to be thrown in .NET as well as .NET Framework for this particular case but not the other cases?  Was the original code missing the `catch (Exception)` path for .NET?

It's hard to tell if this deviation from the pattern is intentional here or not.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlTransaction.cs
paulmedynski: Ahh true - we can't really tell if those specific exceptions _aren't_ being thrown in .NET, so you would need the \#if NET for each one, unless you consolidate that into a finally or something.

Let's leave these as-is and maybe revisit during a targeted cleanup later if desired.

paulmedynski (approved):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlTransaction.cs
benrr101: 😨 My hope/suspicion is that this was some misguided attempt to "fix" something that it didn't actually fix. This seems like something that should be aggressively removed since all it does is confuse us.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlTransaction.cs
benrr101: Yes I agree. I think I will put some effort into coming up with a better solution to this. SqlTransaction is a pretty small class for this issue, but other classes will be much worse.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlTransaction.cs
benrr101: - ThreadAbortException
  > .NET Core and .NET 5+ only: Even though this type exists in .NET Core and .NET 5+, since [Abort](https://learn.microsoft.com/en-us/dotnet/api/system.threading.thread.abort?view=net-9.0) is not supported, the common language runtime won't ever throw [ThreadAbortException](https://learn.microsoft.com/en-us/dotnet/api/system.threading.threadabortexception?view=net-9.0).

- OutOfMemory
  Looks like it can be thrown, but my guess is it's all best effort at that point. The runtime is blowing up, not much we can do even if we want to (unless it's within a CER)

- StackOverflow
  I suspect its the same as OutOfMemory

> Was the original code missing the catch(Exception)

Correct - netfx didn't have a general catch, netcore didn't have any try/catch

PR #3232, opened 2025-03-19
Cleanup | SqlDiagnosticListener Classes

benrr101:
**Description**: Based on the reception of my last cleanup PR, i don't know how this one will be received. This PR just tidies up the SqlDiagnosticListener classes after they were merged. 

* Removing Name - although a name is required to be passed to the DiagnosticListener class on construction, we only ever use the same one, and the class is internal. Thus, it doesn't make a ton of sense to me to expose the name within the project. So, I removed it from the constructor and just pass in the hardcoded name we use. If we end up having multiple names later, we can reinstate this.
* The rest of the changes are just breaking up long lines and some light reordering of member variables.

**Testing**: Once more, no real functional behavior change. Just a bit of tweaks to syntax and a change that doesn't impact anything external (the name thing).

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
@@ -159,7 +159,7 @@ namespace Microsoft.Data.SqlClient
         private static bool _forceRetryableEnclaveQueryExecutionExceptionDuringGenerateEnclavePackage = false;
 #endif
 
-        private static readonly SqlDiagnosticListener s_diagnosticListener = new SqlDiagnosticListener(SqlDiagnosticListener.DiagnosticListenerName);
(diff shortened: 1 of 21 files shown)


mdaigle (approved):
Are you using any kind of linter for these changes? I see an editorconfig file checked into the repo, but I don't have much experience with c# linting. Is there a good way to leverage that across IDEs so that we can all follow the same style?

paulmedynski (approved):

benrr101:
@mdaigle Sort of... The tooling I use raises a lot more suggestions/hints than plain VS, but it's not raising line length suggestions. Partly because the entire codebase would light up brighter than a christmas tree, partly because I don't know if it's even a supported feature, partly because it may already be turned off in .editorconfig, and partly because I don't really like line length being an enforced thing - I try to follow no lines >120 characters, but sometimes it's only over by a few characters and splitting it would make it less readable. In that situation I don't really want to be forced to split it, I don't want to have it be permanently lit up, and I don't want to have to put (possibly tooling-specific) warning suppressions in the code. It's all very subjective :)

PR #3236, opened 2025-03-20
Merge | AssemblyCache

benrr101:
**Description**: Moving one small class from netfx project to common project. I made a couple changes to it along the way:

* Removed comments that make literally no sense anymore
* Made the class static (it only has two static methods in it anymore)
* Couple low-impact stylistic changes

**Testing**: Just moving things around. No funny business here.

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AssemblyCache.netfx.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AssemblyCache.netfx.cs
@@ -4,41 +4,19 @@
 
 #if NETFRAMEWORK
 
-//------------------------------------------------------------------------------
-//ThisAssembly class keeps a map of the following:
-//AssemblyName to AssemblyID
-//TypeName to TypeID
-//AssemblyID to AssemblyRef and State
-//TypeID to TypeRef and AssemblyId
-//
-//Adding an assembly to this class will NOT enable users to create types from that assembly. Users should explicitly add type details and link types to assemblies.
-// This class also registers for assembly resolve events so that dependent assemblies can be resolved if they are registered.
-//This class does NOT know anything about assembly dependencies. It simply loads assemblies as handed over to it.
-//Users can take advantage of connection pooling by tying this instance to a pooling-aware component.
-//------------------------------------------------------------------------------
-
 using System;
 using System.Diagnostics;
 using Microsoft.Data.SqlClient.Server;
 
 namespace Microsoft.Data.SqlClient
 {
-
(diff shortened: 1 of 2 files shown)


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AssemblyCache.netfx.cs
```diff
@@ -0,0 +1,44 @@
...
+
+namespace Microsoft.Data.SqlClient
+{
+    internal static class AssemblyCache
+    {
+        internal static int GetLength(object inst)
+        {
+            // Caller should have allocated enough, based on MaxByteSize
+            return SerializationHelperSql9.SizeInBytes(inst);
+        }
```
edwardneal, line 19: This method is only called by SqlParameter.GetActualSize, [here](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlParameter.cs#L1617-L1622). We can remove the `#if` in SqlParameter, and delete the AssemblyCache.GetLength method completely.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AssemblyCache.netfx.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        internal static SqlUdtInfo GetInfoFromType(Type t)
```
edwardneal, line 21: This method is present in netcore's SqlConnection class, [here](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnection.cs#L2578-L2597). Not sure how you want to tackle that - either to remove the method from netcore's SqlConnection and include AssemblyCache in both projects, or add it to netfx's SqlConnection and remove AssemblyCache from both projects.

mdaigle (approved):
Looks good as is for a pure move. But I also agree with @edwardneal 's method cleanup suggestion.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AssemblyCache.netfx.cs
benrr101: Fantastic catch! Yes that's a great change!!

mdaigle (approved):

samsharma2700 (approved):

edwardneal (approved):

paulmedynski (approved):

Issue #3223, opened 2025-03-17
Fix Functional test failures in CI


cheenamalhotra:
Functional tests are failing in CI, but not failing the build, link to a sample error:
https://sqlclientdrivers.visualstudio.com/public/_build/results?buildId=111108&view=logs&j=45bbf68e-8822-54cd-de70-7b288193beb0&t=e66ff1ae-4024-561f-a615-46b9f468658b&l=872

Needs investigation and fixing.

PR #3237, opened 2025-03-24
[Tests] Diagnose Windows cert creation failures

paulmedynski:
Diagnostic changes for #3223.  We're now emitting the PowerShell script content used to generate a self-signed cert and add it to the Windows system cert store.  This may help figure out why the cert operations fail intermittently.


edwardneal:
I think I introduced this in #3034, which used PowerShell to generate certificates more frequently. Apologies if so!

The only difference in the PS invocation which I can see between my version and the preceding one is that we no longer load the user profile. Does setting `ProcessStartInfo.LoadUserProfile` to true fix this?

paulmedynski:
> I think I introduced this in #3034, which used PowerShell to generate certificates more frequently. Apologies if so!
> 
> The only difference in the PS invocation which I can see between my version and the preceding one is that we no longer load the user profile. Does setting `ProcessStartInfo.LoadUserProfile` to true fix this?

`LoadUserProfile` is already being set to true, so likely not the issue.  I've run some manual tests in a Windows Server container, and the PS script works fine.  Since the failures are intermittent, I suspect a flaky test environment.

We're having some ADO CI pipeline issues that are preventing me from running CI reliably right now, so I will circle back once the overall CI stuff settles down.

paulmedynski:
The latest commit adds a retry within the `CertificateFixtureBase.CreateCertificate` that seems to work around the `ACCESS DENIED` issue.  We now see this in the test logs:
```
2025-03-31T17:22:11.0473456Z   PowerShell command failed with exit code 1 on attempt 1 of 3; retrying in 5 seconds...
...
2025-03-31T17:22:19.7322237Z     Passed Microsoft.Data.SqlClient.Tests.AlwaysEncryptedTests.ExceptionsAlgorithmErrors.TestInvalidCipherText [44 ms]
```
There are no warnings or errors for the CI runs.

eng/pipelines/common/templates/jobs/ci-run-tests-job.yml
```diff
@@ -236,7 +236,7 @@ jobs:
     - powershell: |
         dotnet --info
         
-        Invoke-WebRequest https://dot.net/v1/dotnet-install.ps1 -OutFile dotnet-install.ps1
+        Invoke-WebRequest https://builds.dotnet.microsoft.com/dotnet/scripts/v1/dotnet-install.ps1 -OutFile dotnet-install.ps1
```
paulmedynski, line 239: This avoids `Invoke-WebRequest` failures due to some dead IPs in the DNS results for `dot.net`, and is the proper canonical DNS name for the dotnet scripts.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNICommon.cs
```diff
@@ -174,7 +174,7 @@ internal static bool ValidateSslServerCertificate(Guid connectionId, string targ
                 {
                     try
                     {
-#if NET9_0
+#if NET9_0_OR_GREATER
```
paulmedynski, line 177: Now we're future-proofed!

The docs for the preprocessor symbols were missing `NET9_0_OR_GREATER`, so I notified the owners and they will be updated accordingly.  The symbol itself has been present since .NET 9.0 release though.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft.Data.SqlClient.csproj
```diff
@@ -16,7 +16,6 @@
      We should remove ResolveComReferenceSilent as soon as we can remove the dependency on mscoree. -->
     <ResolveComReferenceSilent>True</ResolveComReferenceSilent>
     <EnableDefaultCompileItems>false</EnableDefaultCompileItems>
-    <DefineConstants>$(DefineConstants);NETFRAMEWORK;</DefineConstants>
```
paulmedynski, line 19: We don't need to explicitly define `NETFRAMEWORK`.  It has been a built-in symbol for ages.

src/Microsoft.Data.SqlClient/tests/tools/Microsoft.Data.SqlClient.TestUtilities/Fixtures/CertificateFixtureBase.cs
```diff
@@ -138,24 +136,44 @@ exit 1
...
+            // We will retry it a few times with a short delay to avoid spurious
+            // failures in CI pipeline runs.
+            //
+            // See ADO issue for more details:
+            //
+            //  Issue 34304: #3223 Fix Functional test failures in CI
+            //
+            //  https://sqlclientdrivers.visualstudio.com/ADO.Net/_workitems/edit/34304
+            //
+            const int delay = 5;
```
paulmedynski, line 169: These seem like reasonable values.  I observed 2 retries across all of the jobs, so it's not adding significant delay.

paulmedynski (commented):
Ready for review, with my comments added for context.

src/Microsoft.Data.SqlClient/tests/tools/Microsoft.Data.SqlClient.TestUtilities/Fixtures/CertificateFixtureBase.cs
```diff
@@ -166,15 +184,23 @@ exit 1
...
-                {
-                    throw new Exception($"PowerShell command raised exception: {commandOutput}");
-                }
+                
+                Console.WriteLine(
+                    $"PowerShell command failed with exit code {code} on " +
+                    $"attempt {attempt} of {retries}; " +
+                    $"retrying in {delay} seconds...");
+                
+                System.Threading.Thread.Sleep(delay * 1000);
```
benrr101, line 198: Two nits:
* If you add a using up top, you don't need the full System.Threading...
* Might be clearer to use TimeSpan.FromSeconds(delay)

eng/pipelines/common/templates/jobs/ci-run-tests-job.yml
benrr101: This feels a bit fishy since we have the UseDotNet task we could run before this...

benrr101 (approved):
Overall pretty good - if it works, I won't complain too hard.

mdaigle (approved):

samsharma2700 (approved):

eng/pipelines/common/templates/jobs/ci-run-tests-job.yml
paulmedynski: Agreed, I'm getting conflicting information:

https://devblogs.microsoft.com/dotnet/critical-dotnet-install-links-are-changing/

Says that the canonical CDN is `builds.dotnet.microsoft.com`.

https://learn.microsoft.com/en-us/dotnet/core/install/windows#install-with-powershell

Says the URL is `https://dot.net/v1/dotnet-install.ps1`

Those DNS names resolve to different IPs:
```
;; ANSWER SECTION:
dot.net.                143     IN      A       20.112.250.133
dot.net.                143     IN      A       20.236.44.162
dot.net.                143     IN      A       20.70.246.20
dot.net.                143     IN      A       20.76.201.171
dot.net.                143     IN      A       20.231.239.246
```
```
;; ANSWER SECTION:
builds.dotnet.microsoft.com. 2922 IN    CNAME   dotnetcli.trafficmanager.net.
dotnetcli.trafficmanager.net. 60 IN     CNAME   builds.dotnet.microsoft.com.edgesuite.net.
builds.dotnet.microsoft.com.edgesuite.net. 9469 IN CNAME a441.dscd.akamai.net.
a441.dscd.akamai.net.   20      IN      A       24.222.184.168
a441.dscd.akamai.net.   20      IN      A       24.222.184.138
```

I can revert it back, but dead IPs in the `dot.net` records were definitely causing job failures earlier.

ErikEJ: Use `builds.dotnet.microsoft.com`

paulmedynski: Heh, the `dot.net` hosts just HTTP redirect to the other URL:
```
$ curl -I https://dot.net/v1/dotnet-install.ps1
HTTP/2 301
date: Fri, 04 Apr 2025 19:15:47 GMT
server: Kestrel
location: https://builds.dotnet.microsoft.com/dotnet/scripts/v1/dotnet-install.ps1
strict-transport-security: max-age=31536000
```
I guess that one of them is/was dead.  I think I'll keep it as-is.

mdaigle (approved):

PR #3239, opened 2025-03-24
Test | Expose internal types to ManualTests

mdaigle:
Internal types are not accessible outside of their home project. However, they still need to be tested. This PR exposes internal types to the `ManualTests` and `FunctionalTests` projects so that they can be used directly instead of via reflection.

This PR does not update any tests that currently use reflection. Those can be addressed separately as tests are updated or as code health PRs.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
@@ -33,6 +33,9 @@
     <TargetFrameworkMonikerAssemblyAttributesPath>$([System.IO.Path]::Combine('$(IntermediateOutputPath)','$(TargetFramework)','$(TargetFrameworkMoniker).AssemblyAttributes$(DefaultLanguageSourceExtension)'))</TargetFrameworkMonikerAssemblyAttributesPath>
   </PropertyGroup>
   <ItemGroup>
+    <InternalsVisibleTo Include="ManualTests" />
+  </ItemGroup>
+  <ItemGroup>
     <EmbeddedFiles Include="$(GeneratedAssemblyInfoFile)" />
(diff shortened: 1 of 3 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
```diff
@@ -32,6 +32,10 @@
   <PropertyGroup>
     <TargetFrameworkMonikerAssemblyAttributesPath>$([System.IO.Path]::Combine('$(IntermediateOutputPath)','$(TargetFramework)','$(TargetFrameworkMoniker).AssemblyAttributes$(DefaultLanguageSourceExtension)'))</TargetFrameworkMonikerAssemblyAttributesPath>
   </PropertyGroup>
+  <ItemGroup>
+    <InternalsVisibleTo Include="ManualTests" />
```
mdaigle, line 36: This syntax is only available in SDK style projects that use .NET 5+. Our only option for netfx was direct assembly tags.

paulmedynski, line 36: I love it, but do we need to restrict via strong names and public keys?

https://learn.microsoft.com/en-us/dotnet/fundamentals/runtime-libraries/system-runtime-compilerservices-internalsvisibletoattribute

As-is, I think this would let anyone pair their assembly named "ManualTests" with our MDS package, and easily access internals.  Maybe that isn't a concern though, since they can bypass the whole visibility system via reflection anyway.  Thoughts?

paulmedynski (approved):

benrr101 (changes requested):
Love the idea - we've needed it for a long time - but need to see it pass before I can approve :)

benrr101 (approved):
Let's gooooooooooooooooooooooooooooooooooooo 🚢 🚢 🚢 🚀

PR #3241, opened 2025-03-24
Merge | PacketHandle.*.cs

benrr101:
**Description**: Very simple move of PacketHandle.Windows.cs and PacketHandle.Unix from the netcore project to the common project. File names updated to reflect that they only apply to netcore.

**Testing**: Literally no functional changes, just moving files.


mdaigle (approved):

src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
```diff
@@ -886,9 +888,11 @@
     <Compile Include="$(CommonSourceRoot)Microsoft\Data\SqlClient\LocalDb\LocalDbApi.Unix.cs">
       <Link>Microsoft\Data\SqlClient\LocalDb\LocalDbApi.Unix.cs</Link>
     </Compile>
+    <Compile Include="$(CommonSourceRoot)Microsoft\Data\SqlClient\PacketHandle.netcore.Windows.cs">
```
paulmedynski, line 891: Windows -> Unix

paulmedynski (changes requested):
Oops - are you compiling the Unix file?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
benrr101: Oops indeed!

paulmedynski (approved):

Wraith2:
The reason there are two files is that at the time the common practice in this library was to avoid using compiler directives and use msbuild conditions. Since then I think everyone involved has got tired of the complexity involved and the amount of difficulty debugging msbuild so it might be worth considering just combining the two files and using compiler directives. 

/cc @David-Engel

benrr101:
@Wraith2 yes, the eventual plan is to use compiler directives for the different OS builds. Not there yet, but it's in the plan.

PR #3250, opened 2025-04-01
Fix crash when Data Source starts with a comma

### LINKED ISSUE #3110: ArgumentOutOfRange when datasource starts with comma
### Describe the bug
When a connection string's data source starts with a comma, a ArgumentOutOfRangeException is thrown. Here is the stack trace:

```
Unhandled exception. System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'startIndex')
   at System.String.LastIndexOf(Char value, Int32 startIndex, Int32 count)
   at Microsoft.Data.Common.ADP.IsEndpoint(String dataSource, String prefix)
   at Microsoft.Data.Common.ADP.IsAzureSynapseOnDemandEndpoint(String dataSource)
   at Microsoft.Data.SqlClient.SqlConnection.CacheConnectionStringProperties()
   at Microsoft.Data.SqlClient.SqlConnection.set_ConnectionString(String value)
   at Microsoft.Data.SqlClient.SqlConnection..ctor(String connectionString)
   at Program.<Main>$(String[] args) in C:\Projects\[redacted]\Program.cs:line 21
```

### To reproduce
Very easy to reproducte:

```c#
var connection = new SqlConnection("Data Source=,");
```

### Expected behavior
I think ideally either the comma should be accepted or a more explanatory "invalid data source" should be returned.

### Further technical details
Microsoft.Data.SqlClient version: 6.0.0-preview3.24332.3
.NET target: .NET 8
SQL Server version: n/a (exception happens before connection is even attempted)
Operating system: Windows 11 (Version 10.0.26100 Build 26100)

Midhunnnk:
Fixes #3110
this PR fixes a crash in `ADP.IsEndpoint(...)` caused by an unsafe call to `LastIndexOf('\\', length - 1, length - 1)` when the Data Source in the connection string starts with a comma (e.g. `"Data Source=,"`).

When `length == 0`, the method would previously throw an `ArgumentOutOfRangeException`.

---Fix
Added a defensive check to ensure `length > 0` before calling `LastIndexOf`, which prevents invalid index access.

--- Tests

- Added a new unit test: `ConnectionString_WithOnlyComma_ShouldNotThrow`
- The test confirms that the exception is no longer thrown
- Verified the test fails without the fix and passes with it

This is my first open-source contribution — excited to be part of the project! Let me know if you'd like any changes.

### DIFF
--- a/src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlConnectionTest.cs
+++ b/src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlConnectionTest.cs
@@ -1082,6 +1082,18 @@ namespace Microsoft.Data.SqlClient.Tests
             Assert.Equal(1, (int)field.GetValue(cn));
         }
 
+
+
+        [Fact]
+        public void ConnectionString_WithOnlyComma_ShouldNotThrow()
+        {
+            SqlConnection cn = new SqlConnection("Data Source=,;Initial Catalog=master;Integrated Security=True");
+            Exception ex = Record.Exception(() => cn.Open());
+
+            Assert.NotNull(ex);
+            Assert.False(ex is ArgumentOutOfRangeException);
+        }
+
         [Theory]
         [InlineData("myserver.database.windows.net")]
         [InlineData("myserver.database.cloudapi.de")]
--- a/src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlConnectionTest.cs
+++ b/src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlConnectionTest.cs
@@ -1085,13 +1085,15 @@ namespace Microsoft.Data.SqlClient.Tests
 
 
         [Fact]
-        public void ConnectionString_WithOnlyComma_ShouldNotThrow()
+        public void ConnectionString_WithOnlyComma()
         {
-            SqlConnection cn = new SqlConnection("Data Source=,;Initial Catalog=master;Integrated Security=True");
-            Exception ex = Record.Exception(() => cn.Open());
+            // Test Case for https://github.com/dotnet/SqlClient/issues/3110
+            // Validates that a single-comma Data Source (e.g., "Data Source=,") no longer causes ArgumentOutOfRangeException
+            // Instead, it should throw a SqlException indicating a connection failure
 
-            Assert.NotNull(ex);
-            Assert.False(ex is ArgumentOutOfRangeException);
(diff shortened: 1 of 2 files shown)


Midhunnnk:
@dotnet-policy-service agree

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlConnectionTest.cs
```diff
@@ -1082,6 +1082,18 @@ public void ConnectionRetryForNonAzureEndpoints()
...
+
+
+        [Fact]
+        public void ConnectionString_WithOnlyComma_ShouldNotThrow()
+        {
+            SqlConnection cn = new SqlConnection("Data Source=,;Initial Catalog=master;Integrated Security=True");
+            Exception ex = Record.Exception(() => cn.Open());
+
+            Assert.NotNull(ex);
+            Assert.False(ex is ArgumentOutOfRangeException);
```
ErikEJ, line 1094: What type IS ex?? Maybe assert that?

paulmedynski: It's probably an `SqlException`, so try this instead of lines 1091-1094:
```
Assert.Throws<SqlException>(() => { cn.Open(); });
```

It looks like we pass the `,` down the stack to the TCP socket as-is, so the error is network related.  You can check the exception `Message` for some key content if you want, but I don't think it's necessary here.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlConnectionTest.cs
```diff
@@ -1082,6 +1082,18 @@ public void ConnectionRetryForNonAzureEndpoints()
             Assert.Equal(1, (int)field.GetValue(cn));
         }
 
+
+
+        [Fact]
+        public void ConnectionString_WithOnlyComma_ShouldNotThrow()
```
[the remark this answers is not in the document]
paulmedynski: It also wouldn't hurt to leave a comment on this test explaining why it exists, and link to the GitHub issue/PR.

If I come across this test in a few years, it will look a bit funny to have a whole test just for comma

paulmedynski (changes requested):
Great fisrt open-source contribution!  Little bugs like these can be very frustrating for users, so we appreciate you taking the time to fix one :)

paulmedynski (approved):
Changes look great - thanks!

cheenamalhotra (approved):

PR #3251, opened 2025-04-01
Merge | Merge metrics/performance counter methodology

edwardneal:
Relates to #1261, but also lays the groundwork for #2211.

SqlClientcurrently has two different ways to record metrics - .NET Core has counters on the EventSource, and .NET Framework has performance counters. These are *usually* recorded at the same time, but they're separated by various `#if NET` conditional compilation blocks.

This PR consolidates both metric surfaces into a single `SqlClientMetrics` class, with various `HardConnectRequest` methods which increment a performance counter or event counter. This means that we've no longer got this type of conditional compilation scattered throughout the connection logic.

@benrr101 - this is the outcome of the [conversation](https://github.com/dotnet/SqlClient/pull/2910#discussion_r1796431098) we had while merging DbConnectionInternal. I've split it out so it's easier to review.

I'd appreciate a CI run please.


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/DbConnectionPoolGroup.cs
```diff
@@ -132,9 +132,7 @@ internal int Clear()
                     if (pool != null)
                     {
                         DbConnectionFactory connectionFactory = pool.ConnectionFactory;
-#if NETFRAMEWORK
```
mdaigle, line 135: These are covered by DbConnectionFactory.QueuePoolForRelease

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Diagnostics/SqlClientMetrics.cs
```diff
@@ -0,0 +1,346 @@
...
+        {
+#if NET
+            EnableEventCounters();
+#endif
+        }
+
+        /// <summary>
+        /// The number of actual connections that are being made to servers
+        /// </summary>
+        internal void HardConnectRequest()
```
mdaigle, line 46: This is definitely a step in the right direction. I still don't love all of the branching in here and I find reaching across partial classes to access fields confusing. But I'm struggling to think of any other way to structure things that'll still be useful when adding OTEL support in the future.

mdaigle (commented):
Overall looks good, but I don't know enough about the test coverage of this functionality to approve just based on a successful CI run. Have you done any manual validation that the events and metrics are still emitted as expected?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Diagnostics/SqlClientMetrics.cs
edwardneal: Another arrangement which I'd thought a little about was:

```csharp
// SqlClientMetrics.netfx.cs:
using PlatformSpecificMetric = PerformanceCounter;

// SqlClientMetrics.netcore.cs
using PlatformSpecificMetric = long;

// Both files
void IncrementPlatformSpecificMetric(ref PlatformSpecificMetric metric) { }
void DecrementPlatformSpecificMetric(ref PlatformSpecificMetric metric) { }

// SqlClientMetrics.cs
internal void EnterNonPooledConnection()
{
    IncrementPlatformSpecificMetric(ref _nonPooledConnectionsCounter);
}
```

This needs a higher LangVersion (12.0), but I can see #3240 is working on that.

There'll still be two conditional compilations (netcore has two more event counters than netfx, for the number of active hard & soft connections) but it should be an improvement.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/DbConnectionPoolGroup.cs
edwardneal: There was test coverage for the event counters in EventCounterTest, but nothing for the performance counters. The code coverage looked reasonable, but it was probably being exercised through the rest of the tests.

I've widened EventCounterTest to cover both metrics, so we've got more explicit test coverage now.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Diagnostics/SqlClientMetrics.cs
```diff
@@ -0,0 +1,346 @@
...
+
+        public SqlClientMetrics(SqlClientEventSource eventSource)
+        {
+            _eventSource = eventSource;
+            EnableDefaultMetrics();
+        }
+
+        private void EnableDefaultMetrics()
+        {
+#if NETFRAMEWORK
```
benrr101, line 31: Still not a fan of having methods that become empty based on a compiler flag... For this one, can we just roll it into the constructor?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Diagnostics/SqlClientMetrics.cs
```diff
@@ -0,0 +1,346 @@
...
+        /// <summary>
+        /// The number of actual connections that are being made to servers
+        /// </summary>
+        internal void HardConnectRequest()
+        {
+#if NET
+            Interlocked.Increment(ref _activeHardConnectionsCounter);
+            Interlocked.Increment(ref _hardConnectsCounter);
+#endif
+#if NETFRAMEWORK
```
benrr101, line 52: There's been some discussion on this, but the way I'm looking at it, if NET isn't set NETFRAMEWORK will be set. I think we can just use an `#else` for these.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Diagnostics/SqlClientMetrics.cs
benrr101: Yeah, the partials kinda sketches me out a bit, though I acknowledge it would do the job pretty well. Your new suggestion is pretty clever, but I think we could do it all in one file with a plan like that. I think if we can get it all in one file without it being a mess of #ifs, then that'd be the ideal 👍

benrr101 (approved):
Overall, love it! 🚀 🚀 🚀 
I'd love it more if we can get it all into a single file, but I'll leave that to your judgement.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Diagnostics/SqlClientMetrics.cs
edwardneal: I think so - I'll test a few ideas and see if there's a clean way to deal with this.

edwardneal: What's your opinion on this version? The variable names are a little fragile and I'd probably look askance at the `IncrementPlatformSpecificMetric`/`DecrementPlatformSpecificMetric` methods if approaching for the first time, but we perform far less conditional compilation.

benrr101: I think it's great!

benrr101:
@edwardneal ok, so looking at these failures, I thiiiink I understand why these failures are happening. Sorta. The MetricsTest file is only pulled into the test project if it's on test set 3 (or no test set), that's why only some of the test jobs are failing (specifically the ones that end with 3). That test file has references to the event source class, which is internal. Since we don't have internalsvisibleto the test projects (yet ... @mdaigle did you give up on that?), I don't think we can access that class from the test project.

The part I don't get tho, if I build it in the IDE, it doesn't complain at all. But that build specifies no test set, which makes me wonder if something including a file or property that makes it all work. Nothing stood out at me, but I didn't go through it with a fine toothed comb.

benrr101:
@edwardneal Update - apparently something is really messed up with how we have our projects set up.... As per @mdaigle, the ref projects are being used for the internalsvisibleto when building from command line, but the implementation projects are being used when building from IDE. Thus, we see the failures of your unit tests when we build the project during CI. I don't think we have a good plan for fixing this anytime soon (it's sort of predicated on a lot of other things being fixed). But in the meantime, Malcolm is going to back out the internalsvisibleto change so people don't get the wrong idea that they can use internal classes in test projects safely.

To address this ... I don't know what's best. I'd like to keep the tests in the codebase so we can use them someday, but maybe just comment them out? (it grosses me out to even type that)

edwardneal:
Good catch, thanks - I'd not yet seen the test results. It's worth keeping in mind that this test originally used reflection; I'll swap it back to doing so, and we can avoid the problem completely.

edwardneal:
I've adjusted MetricsTest to use reflection again - could I have another CI run please?

benrr101 (approved):

mdaigle (approved):

benrr101:
@edwardneal please don't mind that I just fixed the merge conflict for you :)

PR #3256, opened 2025-04-03
Merge | DbConnectionString.Common

benrr101:
**Description**: This is a bit more heavyweight change. DbConnectionString.Common.cs contains several classes and a large collection of utility methods. I have made the following changes:

* Created a new namespace Microsoft.Data.Common.ConnectionString to house these connection string utilities
* Separated DbConnectionStringDefaults
* Separated DbConnectionStringKeywords
* Separated DbConnectionStringSynonyms
* Renamed DbConnectionStringBuilderUtil to DbConnectionStringUtilities
* Extracted AttestationProtocolUtilities, IpAddressPreferenceUtilities, and PoolBlockingUtilities from DbConnectionStringUtilities (these were especially called out with regions in the previous).
* Moved all aforementioned files into the common project
* Wired up the name changes

This also identified a lot of pointless code that just forwards method calls or duplicates enums. I will be cleaning these up in a future PR.

**Testing**: Stuff moved but wasn't functionally changed, no changes to public APIs. Projects still build.


src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/ConnectionString/DbConnectionStringKeywords.cs
```diff
@@ -0,0 +1,72 @@
...
+
+        // OleDb
+        internal const string FileName = "File Name";
+        internal const string OleDbServices = "OLE DB Services";
+        internal const string Provider = "Provider";
+
+        // OracleClient
+        internal const string Unicode = "Unicode";
+        internal const string OmitOracleConnectionName = "Omit Oracle Connection Name";
+        #endif
```
edwardneal, line 70: All of these except for `Driver` could be removed altogether, they're never referenced.

`Driver` is also technically only referenced in a dead code path in DbConnectionOptions, but might be a little out of scope.

edwardneal (commented):
There's one `#if NETFRAMEWORK` conditional compilation block which we can pare down - and in future, remove completely.

Besides that, I agree with your TODO comments about removing the redundant code.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/ConnectionString/DbConnectionStringKeywords.cs
benrr101: Hmm. That's a valid point... I think the origin of this file was to basically be a mirror of the keywords from System.Data.Common. Otherwise, why would we have oracle nonsense in here?
In the cleanup PR I have queued, I'm getting rid of all but the one you mentioned that is still around.

samsharma2700 (approved):

PR #3262, opened 2025-04-08
Merge | SqlClientPermission

benrr101:
**Description**: This change is a bit heavier than it should be because of reasons... 

* Move SqlClientPermission from netfx project to common project
* Separate SqlClientPermissionAttribute from the SqlClientPermission file into its own file in the common project
* Move NameValuePermission into SqlClientPermission class, as a subclass
  * This class is only used for SqlClientPermission, and I really didn't like it hanging out in the middle by itself.
* Applied some reorg/cleanup of SqlClientPermission
  * I would have gone further with this, but because this class overrides most of the base implementation (System.Data.Common.DBDataPermission), it's a bit more difficult to untangle.
* Removed XML encode and decode methods
  * This is available via built-in methods, so it is not ideal to write our own.

**Testing**: For the most past, there are no functional changes. Code is moved around. CI should validate.


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlClientPermission.netfx.cs
```diff
@@ -0,0 +1,702 @@
...
+using System.Security;
+using System.Security.Permissions;
+using Microsoft.Data.Common;
+using DBDataPermission = System.Data.Common.DBDataPermission;
+
+namespace Microsoft.Data.SqlClient
+{
+    /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlClientPermission.xml' path='docs/members[@name="SqlClientPermission"]/SqlClientPermission/*' />
+    [Serializable]
+    public sealed class SqlClientPermission : DBDataPermission
```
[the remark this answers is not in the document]
mdaigle: It feels a little weird to add obsolete methods to the API surface. 

If someone is multitargeting with SDS, won't they already have the necessary conditional compilation to skip usage of this type?

I'm also worried that people may have apps where the deprecation warning is globally disabled. In that situation, if they add .NET as a target, they could get unexpected behavior when the checks aren't respected at runtime.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlClientPermission.netfx.cs
```diff
@@ -388,34 +389,6 @@ private void CopyFrom(SqlClientPermission permission)
             }
         }
         
-        private string DecodeXmlValue(string value)
```
mdaigle: I agree that it makes sense to use built ins, but they may perform other substitutions that could make this a breaking change. How is the connection string ultimately used in this situation? I would want to know if it's ever exposed in a way that customer code could depend on or if it's just an internal detail of the security element.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlClientPermission.netfx.cs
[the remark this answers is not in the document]
benrr101: I personally would be opposed to re-introducing an API just to mark it obsolete. I can see the value in smoothing over the transition from SDS, but to me it doesn't seem right to add useless code in order to do it.

Regardless, if we did decide to introduce them, them we can just as easily add it to the netcore project and drop the suffix/wrapper.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlClientPermission.netfx.cs
```diff
@@ -0,0 +1,702 @@
...
+        private bool IsEmpty()
+        {
+            ArrayList keyvalues = _keyvalues;
+            bool flag = !IsUnrestricted() && !AllowBlankPassword && (keyvalues == null || (0 == keyvalues.Count));
+            return flag;
+        }
+        
+        #endregion
+
+        private sealed class NameValuePermission : IComparable<NameValuePermission>
```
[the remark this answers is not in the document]
benrr101: While true, it is marked as serializable, I don't see anywhere in the code where we're serializing it, and since it's private I figured it wouldn't make much difference to leave out the serializable attribute (especially since I can't pull up mdac bugs) 🤔 I can add it back, but I don't know what difference it will or won't make...

Likewise, I'm not sure what difference the generic vs non-generic IComparable will make. But I can add them back...

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlClientPermissionAttribute.netfx.cs
```diff
@@ -0,0 +1,32 @@
...
+        /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlClientPermissionAttribute.xml' path='docs/members[@name="SqlClientPermissionAttribute"]/ctor/*' />
+        public SqlClientPermissionAttribute(SecurityAction action) : base(action)
+        {
+        }
+
+        /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlClientPermissionAttribute.xml' path='docs/members[@name="SqlClientPermissionAttribute"]/CreatePermission/*' />
+        public override IPermission CreatePermission()
+        {
+            return new SqlClientPermission(this);
+        }
```
cheenamalhotra, line 28:
```suggestion
        public override IPermission CreatePermission() => new SqlClientPermission(this);
```

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlClientPermission.netfx.cs
```diff
@@ -0,0 +1,701 @@
...
+            CopyFrom(permission);
+        }
+        
+        #endregion
+        
+        #region Properties
+        
+        private bool _IsUnrestricted
+        {
+            get
```
cheenamalhotra, line 87: get => base.IsUnrestricted();

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlClientPermission.netfx.cs
```diff
@@ -0,0 +1,701 @@
...
+        {
+            DBConnectionString constr = new DBConnectionString(connectionString, restrictions, behavior, SqlConnectionString.GetParseSynonyms(), false);
+            AddPermissionEntry(constr);
+        }
+        
+        /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlClientPermission.xml' path='docs/members[@name="SqlClientPermission"]/Copy/*' />
+        public override IPermission Copy()
+        {
+            return new SqlClientPermission(this);
+        }
```
cheenamalhotra, line 117:
```suggestion
        public override IPermission Copy() => new SqlClientPermission(this);
```

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlClientPermission.netfx.cs
```diff
@@ -0,0 +1,701 @@
...
+                }
+
+                // mid-chain failure
+                return hasMatch;
+            }
+
+            internal NameValuePermission CopyNameValue()
+            {
+                return new NameValuePermission(this);
+            }
```
cheenamalhotra, line 591:
```suggestion
            internal NameValuePermission CopyNameValue() => new NameValuePermission(this);
```

cheenamalhotra (approved):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlClientPermission.netfx.cs
paulmedynski: Personally, I like function bodies enclosed in braces.  And in this case, it matches the format of the sibling setter.  The => notation increases code density, which isn't necessarily a good thing.  Personal preference I guess.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlClientPermission.netfx.cs
```diff
@@ -0,0 +1,701 @@
...
+            
+            string version = securityElement.Attribute(XmlStr._Version);
+            if (version != null && !version.Equals(XmlStr._VersionNumber))
+            {
+                throw ADP.InvalidXMLBadVersion();
+            }
+
+            string unrestrictedValue = securityElement.Attribute(XmlStr._Unrestricted);
+            _IsUnrestricted = unrestrictedValue != null && bool.Parse(unrestrictedValue);
```
paulmedynski, line 146: Just FYI, but this seemingly benign assignment can (surprisingly) throw reflection exceptions, which we don't document.  In fact, the docs don't claim this function will throw anything!

paulmedynski (approved):

PR #3263, opened 2025-04-08
Merge | TaskToApm

benrr101:
**Description**: This PR removes the TaskToApm class. This class was used to take task-based asynchronous operations (TPL) and convert them to the Asynchronous Programming Model (APM) pattern. It was only used in one situation in the SqlSequentialStream class.

MSDN has its own simple guide on how to convert from tasks to callbacks, which *seems* like it would eliminate the need for this bulky class. Please see https://learn.microsoft.com/en-us/dotnet/standard/parallel-programming/tpl-and-traditional-async-programming#implement-the-apm-pattern-by-using-tasks 

**Testing**: Since this is a fairly significant change, it will definitely need to be tested.


ErikEJ:
Looking forward to the day when all this effort can be spent on user facing fixes and features.. but nice to see you are getting closer.

paulmedynski (approved):

mdaigle:
Looks like TaskToAPM was pulled from .NET Framework: https://referencesource.microsoft.com/#mscorlib/system/threading/Tasks/TaskToApm.cs,419004d6a72826c7,references

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlSequentialStream.cs
```diff
@@ -328,11 +329,24 @@ public override int EndRead(IAsyncResult asyncResult)
...
+            byte[] array,
+            int offset,
+            int count,
+            AsyncCallback asyncCallback,
+            object asyncState)
+        {
+            Task<int> readTask = ReadAsync(array, offset, count, CancellationToken.None);
+            if (asyncCallback is not null)
+            {
+                readTask.ContinueWith(result => asyncCallback(result));
```
mdaigle, line 342: The difference I can see here is that you're returning a reference to the read task, not the callback. So the caller has no way to await the callback completion. See the TAP to APM section here for an example: https://learn.microsoft.com/en-us/dotnet/standard/asynchronous-programming-patterns/interop-with-other-asynchronous-patterns-and-types

In that example, they create a new task completion source that the callback signals to and return the Task associated with that completion source.

mdaigle (changes requested):

edwardneal:
> Looks like TaskToAPM was pulled from .NET Framework: https://referencesource.microsoft.com/#mscorlib/system/threading/Tasks/TaskToApm.cs,419004d6a72826c7,references

It was also made public in .NET Core, as [TaskToAsyncResult](https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.tasktoasyncresult?view=net-9.0). There are some differences in the state management, ([code](https://github.com/dotnet/runtime/blob/main/src/libraries/Common/src/System/Threading/Tasks/TaskToAsyncResult.cs)) but I've not checked whether those code paths are ever hit.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlSequentialStream.cs
benrr101: Hmm. The example from the link I linked in the description didn't wait for the callback to finish... Not sure why they wouldn't do that.

Either way, I've rewritten it to follow the pattern in the link you provided.

benrr101:
@edwardneal Ok, I've rewritten the netcore implementation to use TaskToAsyncResult.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlSequentialStream.cs
benrr101: @mdaigle is that good enough for you? 😅

mdaigle: After reading more about the Begin* and End* APIs for Stream, I think my comment here was off base. The expectation from callers is that the returned result does not include the execution of the callback. The callback runs as a fire and forget task. I would say let's stick with the built in implementation for .NET Framework that we were previously using.

paulmedynski (approved):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlSequentialStream.cs
benrr101: @mdaigle Ok, I've reverted that change of the netfx implementation. There isn't really a "built-in" implementation like there is for netcore, so I hope this is what you meant.

mdaigle (approved):

PR #3264, opened 2025-04-08
Merge | BufferWriterExtensions

benrr101:
**Description**: ~~This one is super easy, move the BufferWriterExtension class to the common project. Specifically, to a new Utilities namespace, with the .netfx suffix. While we're at it, also move the ArrayBufferWriter class to a Utilities namespace and rename it to use the .netfx suffix.~~

**Update**: There is now scope creep in this PR... BufferWriterExtension class has now been moved to a new Utilities namespace in the common project. The original change moved ArrayBufferWriter to the utilities namespace, but this was not ideal. Since ArrayBufferWriter is meant to be like a polyfill for netfx (ArrayBufferWriter was not added to dotnet until netstandard2.0), it should (unfortunatly) be in the System.Buffers namespace (otherwise we will need a #if for each time it is used, to change namespace). As such, this file was moved to the System/Buffers folder in the common project. The netfx project was updated as such.

I considered merging the BufferWriterExtension into ArrayBufferWriter but 1) ArrayBufferWriter is an exact copy of the dotnet runtime implementation (since it is not part of net framework), 2) there are no tests for it, 3) we can't add tests because the type is `internal`, and 4) ArrayBufferWriter is technically generic, while BufferWriterExtensions is not generic. It just makes it kind of a low roi change, so I'll just live it as-is.

**Scope Creep**: During CI, a reference to ArrayBufferWriter was failing in the SqlObjectPools class. Since this is definitely a utility, I moved it and the SqlObjectPool classes to the utilities namespace. I then renamed them to drop the "Sql" from the class names. I also think that the thread-safe implementation of SqlObjectPools is a bit overcomplicated, so I replaced it with a much simpler `Lazy` object (which is thread-safe already).

**Testing**: Moving the files should not have issues, rewriting SqlObjectPool to use Lazy shouldn't have impact (since it's still thread-safe).


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Utilities/ObjectPools.cs
```diff
@@ -0,0 +1,24 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Buffers;
+
+namespace Microsoft.Data.SqlClient.Utilities
+{
+    /// <summary>
+    /// This is a collection of general object pools that can be reused as needed.
```
paulmedynski, line 11: This doesn't seem general.  It's a pool of ArrayBufferWriter<byte>s.  Should the class name be ArrayBufferWriterPools ?

paulmedynski (changes requested):
Just one comment about class name.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Utilities/ObjectPools.cs
benrr101: I think the theory is that the class will have more object pools added to it as we go. I'm not sure what other pools might be added, but I think that's the idea.

paulmedynski: Oh I see.  This class is basically a namespace to hold static instances of a variety of ObjectPool specializations.

More of a "collection of specific ObjectPool instances that can be reused..."

paulmedynski (approved):

mdaigle (approved):

samsharma2700 (approved):

PR #3269, opened 2025-04-10
Remove comments from config json (default)

cheenamalhotra:
Small change, fixes pipelines that update and save this json file using PS script.


cheenamalhotra:
Got a workaround, change not needed anymore.

PR #3270, opened 2025-04-10
Handle null socket when receiving packets of data

cheenamalhotra:
This will wrap null socket error in `SqlException` to prevent exception occurring in the finally block.
* Also, using Timeout.Infinite (-1) now that is equivalent to 0 as timeout value.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
@@ -841,7 +841,7 @@ namespace Microsoft.Data.SqlClient.SNI
                     else if (timeoutInMilliseconds == -1)
                     {
                         // SqlClient internally represents infinite timeout by -1, and for TcpClient this is translated to a timeout of 0
-                        _socket.ReceiveTimeout = 0;
+                        _socket.ReceiveTimeout = Timeout.Infinite;
                     }
                     else
                     {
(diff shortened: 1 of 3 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
```diff
@@ -829,7 +841,7 @@ public override uint Receive(out SNIPacket packet, int timeoutInMilliseconds)
                     else if (timeoutInMilliseconds == -1)
                     {
                         // SqlClient internally represents infinite timeout by -1, and for TcpClient this is translated to a timeout of 0
```
paulmedynski, line 843: Stale comment:  "timeout of 0"

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
```diff
@@ -884,7 +896,9 @@ public override uint Receive(out SNIPacket packet, int timeoutInMilliseconds)
                 }
                 finally
                 {
-                    _socket.ReceiveTimeout = 0;
+                    // Reset the socket timeout to -1 (Timeout.Infinite) after the receive operation is done
```
paulmedynski, line 899: `// Reset the socket timeout to Timeout.Infinite after ... `

paulmedynski (changes requested):
Some documentation fixes.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNITcpHandle.cs
cheenamalhotra, line 843:
```suggestion
                    {
```

paulmedynski (approved):

mdaigle (approved):

PR #3271, opened 2025-04-11
Version upgrade for next releases

cheenamalhotra:
It would make sense to contain Major/Minor/Patch/Revision within branch since with OneBranch we can no longer work with different pipelines maintaining different versioning.
* Upgraded version to 6.1.0

### DIFF
--- a/eng/pipelines/libraries/mds-variables.yml
+++ b/eng/pipelines/libraries/mds-variables.yml
@@ -7,6 +7,20 @@
 variables:
   - group: Release Variables
 
+  # Update this after every release. This is used to generate the MDS NuGet package version.
+  - name: Major
+    value: '6'
+  - name: Minor
+    value: '1'
+  - name: Patch
+    value: '0'
+
+    # Update this for preview releases.
+  - name: Preview
+    value: '-preview'
+  - name: Revision
+    value: '1'
+
   - name: NugetPackageVersion
     value: $(Major).$(Minor).$(Patch)
   - name: PreviewNugetPackageVersion
--- a/eng/pipelines/dotnet-sqlclient-ci-core.yml
+++ b/eng/pipelines/dotnet-sqlclient-ci-core.yml
@@ -71,7 +71,7 @@ parameters:
 
 variables:
   - template: libraries/ci-build-variables.yml@self
-
+  - template: libraries/mds-variables.yml@self
   - name: artifactName
     value: Artifacts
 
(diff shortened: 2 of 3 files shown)


eng/pipelines/libraries/mds-variables.yml
```diff
@@ -7,6 +7,20 @@
 variables:
   - group: Release Variables
 
+  # Update this after every release. This is used to generate the MDS NuGet package version.
+  - name: Major
```
paulmedynski, line 11: Are these values used in place of any environment/pipeline variables, like these:

https://sqlclientdrivers.visualstudio.com/ADO.Net/_library?itemType=VariableGroups&view=VariableGroupView&variableGroupId=17&path=Release%20Variables

paulmedynski (changes requested):
Question about how these variables interact with pipeline Library or environment variables.

eng/pipelines/libraries/mds-variables.yml
cheenamalhotra: Yes, they should take precedence over other pipeline variable sources.

mdaigle (approved):
I'm in favor of these changes! The more isolated we can keep each release branch, the better.

samsharma2700 (approved):

eng/pipelines/libraries/mds-variables.yml
```diff
@@ -6,26 +6,3 @@
 
 variables:
   - group: Release Variables
```
paulmedynski, line 8: Can we remove this entirely?

This injects all of the variables defined in the ADO Libray named "Release Variables", here:

https://sqlclientdrivers.visualstudio.com/ADO.Net/_library?itemType=VariableGroups&view=VariableGroupView&variableGroupId=17&path=Release%20Variables

I think the only variables from that Library that we use in our pipelines are:
- SymbolServer
- SymbolTokenUri

Can we include those variables here?  They don't appear to be secrets.

There are other variables in that Library that I couldn't find us using.  Are they used by templates we include/use?

paulmedynski (changes requested):
Can we remove "Release Variables" entirely?

eng/pipelines/libraries/mds-variables.yml
cheenamalhotra: I want to review the symbols flow altogether in a separate PR, we may not need these variables is what my thought is.

paulmedynski (approved):

PR #3275, opened 2025-04-13
Merge | SqlConnection cosmetic changes

edwardneal:
Contributes to #1261.

This is easier to review commit-by-commit - it performs formatting changes, reorders members, synchronises comments and other cosmetic changes. The end result is that SqlConnection is reduced to a -315/+305 diff.

The remaining diff is 1/3 context connection handling, 1/3 logging changes, 1/6 unported performance improvements and 1/6 being the "real" set of behavioural changes.)

Could someone run CI against this please?

### DIFF
--- a/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnection.cs
+++ b/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnection.cs
@@ -392,11 +392,7 @@ namespace Microsoft.Data.SqlClient
         /// </summary>
         internal string EnclaveAttestationUrl
         {
-            get
-            {
-                SqlConnectionString opt = (SqlConnectionString)ConnectionOptions;
-                return opt.EnclaveAttestationUrl;
-            }
+            get => ((SqlConnectionString)ConnectionOptions).EnclaveAttestationUrl;
         }
 
         /// <summary>
@@ -404,11 +400,7 @@ namespace Microsoft.Data.SqlClient
         /// </summary>
         internal SqlConnectionAttestationProtocol AttestationProtocol
         {
-            get
-            {
-                SqlConnectionString opt = (SqlConnectionString)ConnectionOptions;
-                return opt.AttestationProtocol;
-            }
+            get => ((SqlConnectionString)ConnectionOptions).AttestationProtocol;
         }
 
         /// <summary>
--- a/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnection.cs
+++ b/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnection.cs
@@ -710,7 +710,6 @@ namespace Microsoft.Data.SqlClient
         {
             get
             {
-                string result = _accessToken;
                 // When a connection is connecting or is ever opened, make AccessToken available only if "Persist Security Info" is set to true
                 // otherwise, return null
                 SqlConnectionString connectionOptions = (SqlConnectionString)UserConnectionOptions;
@@ -2553,12 +2552,6 @@ namespace Microsoft.Data.SqlClient
                 Debug.Assert(o != null, "object could NOT be created");
                 return o;
             }
-        }
-
-        internal byte[] GetBytes(object o)
-        {
-            Format format = Format.Native;
-            return GetBytes(o, out format, out int maxSize);
         }
 
         internal byte[] GetBytes(object o, out Format format, out int maxSize)
--- a/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnection.cs
+++ b/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnection.cs
@@ -16,14 +16,12 @@ using System.IO;
 using System.Reflection;
 using System.Runtime.CompilerServices;
 using System.Runtime.ConstrainedExecution;
-using System.Runtime.InteropServices;
-using System.Runtime.Versioning;
 using System.Security;
 using System.Security.Permissions;
 using System.Security.Principal;
-using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.Data.Common;
 using Microsoft.Data.ProviderBase;
 using Microsoft.Data.SqlClient.ConnectionPool;
 using Microsoft.SqlServer.Server;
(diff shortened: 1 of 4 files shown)


src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnection.cs
```diff
@@ -755,12 +735,7 @@ public string AccessToken
                 // When a connection is connecting or is ever opened, make AccessToken available only if "Persist Security Info" is set to true
```
paulmedynski, line 714: result is unused now.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnection.cs
```diff
@@ -2621,8 +2579,7 @@ internal object GetUdtValue(object value, SqlMetaDataPriv metaData, bool returnD
         internal byte[] GetBytes(object o)
         {
             Format format = Format.Native;
-            int maxSize = 0;
-            return GetBytes(o, out format, out maxSize);
+            return GetBytes(o, out format, out int maxSize);
```
paulmedynski, line 2561: Can we discard the out parameters?  They're not used here, and aren't used as input to GetBytes().

`return GetBytes(o, out _, out _);`

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnection.cs
```diff
@@ -586,7 +586,7 @@ private bool UsesClearUserIdOrPassword(SqlConnectionString opt)
             bool result = false;
             if (opt != null)
             {
-                result = (!ADP.IsEmpty(opt.UserID) || !ADP.IsEmpty(opt.Password));
```
paulmedynski, line 651: Wow, ADP.IsEmpty() needs to go.  I've opened an internal task to remove it:

paulmedynski (changes requested):
Cleanup is always welcome!  A couple of minor tweaks.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnection.cs
edwardneal: Agreed. The good news here is that this is practically always called from netfx-specific code paths, and I'm purging it from those as part of the merge.

samsharma2700 (approved):

cheenamalhotra (approved):

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnection.cs
```diff
@@ -80,16 +101,28 @@ private static Dictionary<string, SqlColumnEncryptionKeyStoreProvider> s_systemC
...
         /// Key to the dictionary is a SQL Server Name
         /// IList contains a list of trusted key paths.
         /// </summary>
-        static private readonly ConcurrentDictionary<string, IList<string>> _ColumnEncryptionTrustedMasterKeyPaths
+        private static readonly ConcurrentDictionary<string, IList<string>> _ColumnEncryptionTrustedMasterKeyPaths
             = new(concurrencyLevel: 4 * Environment.ProcessorCount /* default value in ConcurrentDictionary*/,
                                                             capacity: 1,
                                                             comparer: StringComparer.OrdinalIgnoreCase);
 
+        internal bool ForceNewConnection
```
benrr101: I'm assuming the goal here is just to get it into the same place as it is in netcore? Otherwise I'd be a stickler and say properties belong after constructors

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnection.cs
```diff
@@ -827,8 +730,83 @@ public override int ConnectionTimeout
             }
         }
 
+        /// <include file='../../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnection.xml' path='docs/members[@name="SqlConnection"]/CommandTimeout/*' />
+        [DesignerSerializationVisibility(DesignerSerializationVisibility.Hidden)]
+        [ResDescription(StringsHelper.ResourceNames.SqlConnection_ConnectionTimeout)]
+        public int CommandTimeout
```
benrr101, line 695: I don't get this ordering scheme at all.... If we're just trying to get one side in line with the other side, then I won't complain. But if we're not, we've got privates before publics, properties before constructors, properties after constructors...

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnection.cs
```diff
@@ -178,43 +173,35 @@ public SqlConnection(string connectionString, SqlCredential credential) : this()
                 {
                     throw ADP.InvalidMixedArgumentOfSecureCredentialAndIntegratedSecurity();
                 }
-
-                if (UsesContextConnection(connectionOptions))
+                else if (UsesContextConnection(connectionOptions))
```
benrr101, line 176: nit: this one is one I've butted heads with people in the past. Technically, the else is redundant here since if the condition is triggered, we exit the method. In my IDE, I get notices about them, so *I* usually try to avoid them. But I admit, it is tidier to have the else-if chain. Though it would be even better if this could be written as a switch block :/

benrr101 (approved):
I think you goal here is to get one side in line with the other. In that case I can be a lot more forgiving to stuff like order of properties. No need to address the comments, just raising them as things to think on.
Otherwise, glad to get closer to merging the big beasts.

benrr101:
@edwardneal ok that was more dangerous, but yeahhhh I just rebased your branch off of main and fixed the merge conflicts (please don't be mad)

edwardneal:
Thanks rebasing this @benrr101. I noticed that CI was failing on the netfx project, so I've adjusted it slightly.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnection.cs
edwardneal: Exactly - the goal's to make a Visual Studio diff useful. I've tested reordering properties, but it'd rapidly increase the size of the diff: even moving a dozen properties resulted in a 300-line change.

I agree it'd be a good idea to deal with code style fixups though. We could do this after the codebases are merged, or file-by-file through the merged codebase. The WPF repo is going through this at the moment, and the approach they've taken is dotnet/wpf#10270.

PR #3279, opened 2025-04-16
[5.1] CVE-2024-43485 - Update System.Text.Json to 6.0.11

paulmedynski:
- Added an explicit dependency on System.Text.Json 6.0.11 to avoid transitive CVE for .NET Framework.
- Reorganized some PackageReference entries to be alphabetically ordered.
- Fixed incorrect Microsoft.Extensions.Caching.Memory version in AKV .nuspec to suppress https://github.com/dotnet/SqlClient/security/dependabot/13
- Fixed incorrect MDS versions in AKV .nuspec.
- Fixed transitive downgrade of System.Text.Encodings.Web.
- Added TSAUpload bug workaround to TSA options config.

### DIFF
--- a/.config/tsaoptions.json
+++ b/.config/tsaoptions.json
@@ -7,7 +7,7 @@
     "repositoryName": "SqlClient",
     "codebaseName": "SqlClient",
     "allTools": true,
-    "template": "MSDATA_RevolutionR",
+    "template": "MSDATA_RevolutionR_Overloaded0",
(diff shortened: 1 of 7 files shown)


cheenamalhotra (approved):

cheenamalhotra:
Could also use 6.0.11?

mdaigle (approved):

PR #3291, opened 2025-04-22
Merge | TdsParserStateObjectFactory, TdsParserStateObject

edwardneal:
Fairly small merge here, adjacent to TdsParser. This completely merges TdsParserStateObjectFactory, which should be fairly simple.

More importantly, it also starts to reorganise the netfx class hierarchy to align with netcore. This is the key point I'd like some feedback on. I think we'll need to finish this to make more progress with some of the remaining files.

netfx currently has a single TdsParserStateObject class. In netcore this class is abstract, and all of the native functionality is located in a derived TdsParserStateObjectNative class. To start migrating netfx to this structure, I've:

* Created a stub TdsParserStateObjectNative class in the netfx project, ensuring it inherits from TdsParserStateObject
* Marked the original TdsParserStateObject class as abstract, redirected TdsParserStateObjectFactory to instantiate the Native derived class
* Started to move the abstract method definitions in TdsParserStateObject's netcore definition to netfx
* Where the netcore project's TdsParserStateObjectNative implementation of these abstract method definitions matches netfx, merge the two implementations

I've started small, with five methods which just pass through to already-shared code and don't maintain state.

I'd appreciate a CI run please.


benrr101 (approved):
Overall, I think this is a good step! I'm curious about your larger scope ideas and how we can combine our efforts to get merging the projects wrapped up. The team is interested in chatting more with you about it, if you're open to it, feel free to shoot me an email at the email on my profile (since you're pretty secretive, haha)

edwardneal:
Thanks, I've just emailed.

The biggest hurdle for the merge is TdsParser, and the most complex part of that is instantiating and calling into `TdsParserStateObject`. I'm trying to work out a path towards merging that by dealing with the instantiation, then proving the concept of swapping out the simplest virtual method calls. Accessing variables becomes a little more complex as we merge parts of `TdsParserStateObjectNative`, I'm not sure whether to merge every method which touches the same set of variables or to just change their scope to `protected` in netfx's `TdsParserStateObject` definition - there'll be some testing needed.

samsharma2700 (approved):

PR #3292, opened 2025-04-22
Merge | Align TdsExecuteRPC between netcore and netfx

benrr101:
_Recreation of #2985 to resolve merge conflicts and get full CI runs_

Splitting up TdsExecuteRPC to multiple functions to align netcore and netfx

Part of #2953

### DIFF
--- a/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
+++ b/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
@@ -9710,9 +9710,9 @@ namespace Microsoft.Data.SqlClient
                                     if (releaseConnectionLock)
                                     {
                                         task.ContinueWith(
-                                            static (Task _, object state) => ((TdsParser)state)._connHandler._parserLock.Release(),
-                                            state: this,
-                                            scheduler: TaskScheduler.Default
+                                            static (Task _, object state) => ((SqlInternalConnectionTds)state)._parserLock.Release(),
+                                            state: _connHandler,
+                                            TaskScheduler.Default
                                         );
                                         releaseConnectionLock = false;
                                     }
--- a/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
+++ b/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
@@ -9751,10 +9751,9 @@ namespace Microsoft.Data.SqlClient
                             task = completion.Task;
                         }
 
-                        bool taskReleaseConnectionLock = releaseConnectionLock;
-                        execFlushTask.ContinueWith(tsk => ExecuteFlushTaskCallback(tsk, stateObj, completion, taskReleaseConnectionLock), TaskScheduler.Default);
+                        TDSExecuteRPCParameterSetupFlushCompletion(stateObj, completion, execFlushTask, releaseConnectionLock);
 
-                        // ExecuteFlushTaskCallback will take care of the locks for us
+                        // TDSExecuteRPCParameterSetupFlushCompletion calling ExecuteFlushTaskCallback will take care of the locks for us
                         releaseConnectionLock = false;
 
                         return task;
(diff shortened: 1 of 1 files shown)


src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -10231,6 +9797,493 @@ internal Task TdsExecuteRPC(SqlCommand cmd, IList<_SqlRPC> rpcArray, int timeout
...
+                    notificationRequest,
+                    stateObj,
+                    isCommandProc,
+                    sync,
+                    completion,
+                    startRpc,
+                    startParam
+                ),
+                onFailure: exc => TdsExecuteRPC_OnFailure(exc, stateObj)
+            );
```
edwardneal, line 10278: This corresponds to lines 10138-10146, and it swaps `AsyncHelper.ContinueTaskWithState` for `AsyncHelper.ContinueTask`. The passing of state only seems to have an effect on the `onFailure` lambda, but we also no longer pass `connectionToDoom`.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -10171,7 +9730,15 @@ internal Task TdsExecuteRPC(SqlCommand cmd, IList<_SqlRPC> rpcArray, int timeout
...
-                            stateObj.WriteByte(TdsEnums.SQL2005_RPCBATCHFLAG);
+                            if (_is2005)
+                            {
+                                stateObj.WriteByte(TdsEnums.SQL2005_RPCBATCHFLAG);
+
+                            }
+                            else
+                            {
+                                stateObj.WriteByte(TdsEnums.SQL2000_RPCBATCHFLAG);
+                            }
```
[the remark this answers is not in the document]
benrr101: 🤦‍♂️ That's a good catch

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
benrr101: Are you saying I should be replacing this code with what's in 10138-10146 in main? Or just noting what the difference is for others. 
Doing the rebase on this one was a bit tricky (since I don't trust myself touching TdsParser), but I generally just took whatever the newest code was. In this case, @MichelZ's version was newer (the previous was written by @Wraith2 in 2021), so I took that.

Wraith2: If I remember correctly the connection to doom is present in netfx but not netcore. The difference between ContinueTask and ContinueTaskWithState should only be that the second avoids a closure on the common path, closures on exception paths are irrelevant since it's going to allocate exception, errors, stack trace etc.

edwardneal: I agree that the closure is probably a drop in the ocean (allocation-wise) - but I think we should replace the code with L10138-10146. It's not worth removing the closure's state management, and the exception handling which comes with connectionToDoom would be a regression on netfx. The connectionToDoom parameter is present in both targets now, it looks like that was ported over pretty recently.

Looking more broadly at TdsParser, it's surprisingly similar to SqlConnection - there's a fairly large diff, but it's mostly comprised of non-functional changes. TdsExecuteRPC makes up about 20% of that diff.

Wraith2: The core, TDSParser, SqlConnectionInternal are fairly heavily conserved between the two codebases. Most people are rightfully afraid to touch them unless required.

benrr101: Ok, I replaced it with L10138-10146. Hopefully that's what the agreement seems to be 😬

samsharma2700 (approved):

apoorvdeshmukh (approved):

cheenamalhotra (approved):

samsharma2700 (approved):

PR #3293, opened 2025-04-23
Use NonOfficial template for signing test

cheenamalhotra:
Addresses test signing pipeline trigger issues

### DIFF
--- a/eng/pipelines/dotnet-sqlclient-signing-pipeline.yml
+++ b/eng/pipelines/dotnet-sqlclient-signing-pipeline.yml
@@ -52,7 +52,7 @@ parameters: # parameters are shown up in ADO UI in a build queue time
 
 - name: oneBranchType
   displayName: 'Select OneBranch template'
-  default: Official
+  default: NonOfficial
   values:
   - NonOfficial
   - Official


paulmedynski (approved):

mdaigle (changes requested):
This file is also used as the definition for dotnet-sqlclient-Official. I assume that pipeline will fail if we use the NonOfficial template.

benrr101 (changes requested):
I'd like to get a better understanding of the goal of this change. The default behavior of the nonofficial template is that it'll break the build on APIScan errors, while the official template will raise APIScan errors as TSA bugs (which are currently not working as expected). The template can be changed when starting the pipeline, so I presume this change is only supposed to apply to nightly-type builds. Afaik, this will cause our nightly builds to fail all the time since we still have unresolved APIScan errors.

I think we need to determine what we expect builds to do in each case (and modify the pipeline accordingly):
* Manual builds for release process - imho, should break the build
* Manual builds for testing purposes - imho, should break the build
* Automatic builds for nightly stuff - imho, should raise TSA bugs.

We can make those changes, but it will be a slightly more complex change than just changing the default onebranch template. And if we want to adopt my suggestions, this change basically does the opposite of all of those 😅

paulmedynski:
> This file is also used as the definition for dotnet-sqlclient-Official. I assume that pipeline will fail if we use the NonOfficial template.

My changes in #3298 allow the dotnet-sqlclient-Official pipeline in ADO to run as NonOfficial and succeed.  See: https://sqlclientdrivers.visualstudio.com/ADO.Net/_build/results?buildId=114046&view=results

cheenamalhotra:
> My changes in #3298 allow the dotnet-sqlclient-Official pipeline in ADO to run as NonOfficial and succeed. See: https://sqlclientdrivers.visualstudio.com/ADO.Net/_build/results?buildId=114046&view=results

Once we port this to main, we can then enable non-official build by default.
PR #3327 port this over.

PR #3299, opened 2025-04-24
Cleanup of AKV Pipeline After ESRP Changes

benrr101:
**Description**: This PR cleans up the AKV official pipeline after the recent ESRP changes. There's a handful of changes going on here:

* Change parameter naming conventions such that above the level of the signing task, signing parameters have "Signing" prefix - this helps it be obvious the parameters/variables are used for signing (eg, AkvName isn't clear that it is used for signing - AKV could mean any key vault thing, any especially in the AKV provider pipeline it needs clarification)
* Also, enforce naming conventions of camelCase for parameters and PascalCase for variable group variables
* Enforce alphabetical ordering of parameters
* Move variable group to signing-variables-v2 - this helps us have versionable variable groups if we ever need to update the list of variables the pipeline uses.
* Also ensuring that the variable group is pulled into the pipeline - I don't think this was actually tested previously, since the variable group was not being included, and would fail

**Testing**: Build was successful https://sqlclientdrivers.visualstudio.com/ADO.Net/_build/results?buildId=114057&view=results

### DIFF
--- a/eng/pipelines/steps/compound-esrp-code-signing-step.yml
+++ b/eng/pipelines/steps/compound-esrp-code-signing-step.yml
@@ -5,20 +5,15 @@
 #################################################################################
 
 parameters:
-    - name: ESRPConnectedServiceName
-      type: string
-
     - name: appRegistrationClientId
       type: string
 
     - name: appRegistrationTenantId
       type: string
 
-    - name: EsrpClientId
-      type: string
-
-    - # Name of the Azure Key Vault to retrieve ESRP Code Signing certificate from.
-      name: AuthAkvName
+    - # Name of the Azure Key Vault to retrieve certificates from.
+      # note: This has nothing to do with the AKV provider package.
+      name: authAkvName
       type: string
 
     - name: authSignCertName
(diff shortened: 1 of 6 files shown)


paulmedynski (approved):

eng/pipelines/steps/compound-esrp-code-signing-step.yml
```diff
@@ -5,20 +5,15 @@
 #################################################################################
 
 parameters:
-    - name: ESRPConnectedServiceName
-      type: string
-
     - name: appRegistrationClientId
```
cheenamalhotra, line 8: Can you rename these too for consistency?

eng/pipelines/variables/esrp-signing-variables.yml
```diff
@@ -0,0 +1,17 @@
...
+# Licensed to the .NET Foundation under one or more agreements.                 #
+# The .NET Foundation licenses this file to you under the MIT license.          #
+# See the LICENSE file in the project root for more information.                #
+#################################################################################
+
+# These variables are used for running ESRP signing tasks. All names start with "Signing" to make
+# it clear that these variables are used for signing (as opposed to other msc tasks).
+
+variables:
+    - group: 'esrp-variables-v2'
```
cheenamalhotra, line 11: Can you just call it esrp-variables for now? If this ever needs updates in config in future, we can start with v2 later.

cheenamalhotra: Btw, if you're changing the group, do update all release branches/SNI branches next thing too - because we don't want to keep duplicate groups for such configs that serve all repos and branches with same configuration.

cheenamalhotra:
Can you also run a build on sqlclient official and link it here? Thanks!

eng/pipelines/steps/compound-esrp-code-signing-step.yml
benrr101: There's no real reason to, since at this point all the parameters have to do with signing (it is the _signing_ task after all :) )

eng/pipelines/variables/esrp-signing-variables.yml
benrr101: I opted to call it -v2 because to me it is a second version of the signing variables. Since we've stored these variables in various, inconsistently named places, to me it makes more sense to make some indication that this group isn't one of the old ones.

As for changing all the other branches, yes, that is the goal. But, like I said at the pipeline meeting, I want the new pipelines (AKV provider, only so far) to reference the new tasks, variable groups, etc, as they are created. I'd like it to be entirely parallel to the existing ones since touching the existing ones is a lot riskier. But, I can send out a PR with those changes so we can expedite getting rid of the other group.

cheenamalhotra: Well, at the end of the day only this group will remain, so v2 doesn't really mean anything later. I'm saying this from a final view of what will remain after we cleanup (which we must asap) - this and many other groups that are contradictory with each other. There should only 1 group called 'esrp-variables-v2' (if you keep v2) that would contain the esrp stuff.

cheenamalhotra (approved):

PR #3310, opened 2025-04-29
Merge | Fix/commands null

deusanyjunior:
Currently the method is returning **Object reference not set to an instance of an object** when we try to run a batch without commands. This PR returns **SqlBatchCommand list has not been initialized** instead.

I'd appreciate a CI & Test run.

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBatch.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBatch.cs
@@ -225,7 +225,7 @@ namespace Microsoft.Data.SqlClient
             }
             if (_commands is null) 
             {
-                throw ADP.ArgumentNull(nameof(_commands));
+                throw ADP.InvalidOperation("SqlBatchCommand list has not been initialized.");
             }
             _batchCommand.Connection = Connection;
             _batchCommand.Transaction = Transaction;
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBatch.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBatch.cs
@@ -4,6 +4,7 @@
 
 #if NET
 
+using System;
 using System.Collections.Generic;
 using System.Data;
 using System.Data.Common;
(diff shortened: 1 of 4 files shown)


deusanyjunior:
@dotnet-policy-service agree company="VTEX"

benrr101:
It looks like a good idea - the change is pretty simple. However, we want to make sure that these get caught by our localization process (and that we have consistent error messaging). Could you put the error message in our /resources/string.resx file and reference that when throwing the exception, please?

paulmedynski (approved):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBatch.cs
```diff
@@ -223,6 +223,10 @@ private void SetupBatchCommandExecute()
             {
                 throw ADP.ConnectionRequired(nameof(SetupBatchCommandExecute));
             }
+            if (_commands is null) 
+            {
+                throw ADP.InvalidOperation("SqlBatchCommand list has not been initialized.");
```
paulmedynski, line 228: Per @benrr101's comment, please add an entry to our string resources and reference that here for localization.

deusanyjunior: @paulmedynski and @benrr101 , please, let me know if this is the proper way to add a resource string.

paulmedynski (approved):

src/Microsoft.Data.SqlClient/src/Resources/Strings.Designer.cs
```diff
@@ -1301,7 +1301,18 @@ internal static string ADP_NoQuoteChange {
...
-        
+
+        /// <summary>
+        ///   Looks up a localized string similar to SqlBatchCommands list has not been initialized..
+        /// </summary>
+        internal static string ADP_NoSqlBatchCommandList
+        {
+            get
+            {
+                return ResourceManager.GetString("ADP_NoSqlBatchCommandList", resourceCulture);
```
benrr101, line 1312: Looks real good *except* ... Either this line should be `ADP_NoSqlBatchCommandsList` (with an `s`) OR the resx file should be `ADP_NoSqlBatchCommandList` (without an `s`). I personally would vote for the second choice

As it stands right now, the mismatch is causing the unit test to fail due to ArgumentNullException coming from looking up the resx.

benrr101 (changes requested):
Found the test failure! One character change, should be easy fix. :)

src/Microsoft.Data.SqlClient/src/Resources/Strings.Designer.cs
deusanyjunior: Good catch! Great suggestion. Just fixed.

paulmedynski (approved):

benrr101 (approved):
If the build passes, I'm on board with this change 🚢

deusanyjunior:
@benrr101 Do you have any clue about tests? It seems the results are not good. =\

paulmedynski:
Looks like some network blips.  I'm re-running the failed tests.

PR #2980, opened 2024-11-06
Upgrade / align packages in test projects

MichelZ:
- Upgrade packages only used in test projects to latest versions
- Upgrade XUnit / MSTest SDK
- Remove unused variables
  - MicrosoftNETCoreRuntimeCoreCLRVersion
  - MicrosoftDotNetPlatformAbstractionsVersion
  - SystemTextEncodingCodePagesVersion
- Remove reference to xunit.runner.utility, this should not be linked as per https://github.com/xunit/visualstudio.xunit/issues/409#issuecomment-2480747672

According to https://github.com/dotnet/SqlClient/pull/2379#issuecomment-1979544650, the Remote Executor package and XUnit was held back by .NET6, now that .NET6 is removed, this should work fine

PR #3312, opened 2025-04-29
Tests | Update Package Versions

benrr101:
_Recreation of #2980 to rebase it onto main, the original pull request description is below_

---

- Upgrade packages only used in test projects to latest versions
- Upgrade XUnit / MSTest SDK
- Remove unused variables
  - MicrosoftNETCoreRuntimeCoreCLRVersion
  - MicrosoftDotNetPlatformAbstractionsVersion
  - SystemTextEncodingCodePagesVersion
- Remove reference to xunit.runner.utility, this should not be linked as per https://github.com/xunit/visualstudio.xunit/issues/409#issuecomment-2480747672

According to https://github.com/dotnet/SqlClient/pull/2379#issuecomment-1979544650, the Remote Executor package and XUnit was held back by .NET6, now that .NET6 is removed, this should work fine

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -99,7 +99,6 @@
 
   <!-- Provides Version properties -->
   <Import Project="$(ToolsDir)props\Versions.props" />
-  <Import Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net9.0'))" Project="$(ToolsDir)props\VersionsNet9OrLater.props" />
   <!-- Provides Tool properties -->
   <Import Project="$(ToolsDir)props\Tools.props" />
   <!-- Provides Build properties -->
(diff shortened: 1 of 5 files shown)


src/Directory.Build.props
```diff
@@ -99,7 +99,6 @@
 
   <!-- Provides Version properties -->
   <Import Project="$(ToolsDir)props\Versions.props" />
-  <Import Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net9.0'))" Project="$(ToolsDir)props\VersionsNet9OrLater.props" />
```
paulmedynski, line 102: Are we including these versions somewhere else?

paulmedynski (approved):

src/Directory.Build.props
benrr101: No, the original changes removed the net8+ version.props file, claiming it wasn't necessary anymore. So in the rebase, I just removed the equivalent for net9+ versions. But, looking at the list of versions in there, the net9+ uses newer versions, so I think we probably still need the net9+ versions. I'll add it back, let's see what happens :)

paulmedynski (approved):

cheenamalhotra (approved):

PR #3323, opened 2025-05-01
Merge | SqlInternalConnectionTds preparation

edwardneal:
This contains the preparatory work for the merge of SqlInternalConnectionTds. We're working with a smaller code file here, so I've merged every part of the two files which avoids a behavioural change.

This should be easier to review comment-by-comment - it's bulky, but not complex. I'd also appreciate a CI run.

At the end of this, the diff is +43/-80. Remaining differences here are:
* Optimisation: netcore will clear the array of session state when the connection resets, netfx will replace it with a new array.
* netfx will consider four extra error codes from the server to be transient: 40020, 40143, 40166, 40540.
* netfx contains a little context connection support.
* netfx contains support for Transparent Network IP Resolution, while netcore does not.
* netfx supports the Network Library connection string parameter.
* netcore will throw an exception if it receives an ENV_LOGSHIPNODE ENVCHANGE when the connection string's Application Intent is read-only, while netfx will not.
* When getting the federated authentication token, netfx will stop retrying if the next pause between each attempt to acquire a token would cause the connection timeout to be expired.
* netcore supports passing the User ID parameter to MSAL as an initial username; netfx does not.
* Some exception messages are more precise in netfx than they are in netcore.

What should we do with Transparent Network IP Resolution? This functionality doesn't exist in netcore at all - does it need to be ported? If it only exists in netfx for backwards compatibility, should it be made obsolete?

I'd also be interested to know what the right style of exception messages is. If (for example) the connection receives an unexpected session recovery feature extension acknowledgement, which exception message should I merge in favour of?
> Internal connection fatal error. Error state: {ParsingErrorState}, Feature Id: {ID of feature}.

> Internal connection fatal error.

The former is netfx, the latter is netcore. I prefer the detail of netfx, but I imagine netcore changed this for a reason.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnectionFactory.cs
+++ b/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnectionFactory.cs
@@ -145,7 +145,7 @@ namespace Microsoft.Data.SqlClient
                     opt = new SqlConnectionString(opt, instanceName, false /* user instance=false */, null /* do not modify the Enlist value */);
                     poolGroupProviderInfo = null; // null so we do not pass to constructor below...
                 }
-                result = new SqlInternalConnectionTds(identity, opt, key.Credential, poolGroupProviderInfo, "", null, redirectedUserInstance, userOpt, recoverySessionData, applyTransientFaultHandling: applyTransientFaultHandling, key.AccessToken, pool, key.AccessTokenCallback);
+                result = new SqlInternalConnectionTds(identity, opt, key.Credential, poolGroupProviderInfo, "", null, redirectedUserInstance, userOpt, recoverySessionData, applyTransientFaultHandling, key.AccessToken, pool, key.AccessTokenCallback);
             }
             return result;
         }
--- a/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnectionFactory.cs
+++ b/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnectionFactory.cs
@@ -145,7 +145,7 @@ namespace Microsoft.Data.SqlClient
                     opt = new SqlConnectionString(opt, instanceName, false /* user instance=false */, null /* do not modify the Enlist value */);
                     poolGroupProviderInfo = null; // null so we do not pass to constructor below...
                 }
-                result = new SqlInternalConnectionTds(identity, opt, key.Credential, poolGroupProviderInfo, "", null, redirectedUserInstance, userOpt, recoverySessionData, pool, key.AccessToken, applyTransientFaultHandling: applyTransientFaultHandling, key.AccessTokenCallback);
+                result = new SqlInternalConnectionTds(identity, opt, key.Credential, poolGroupProviderInfo, "", null, redirectedUserInstance, userOpt, recoverySessionData, applyTransientFaultHandling: applyTransientFaultHandling, key.AccessToken, pool, key.AccessTokenCallback);
             }
             return result;
         }
--- a/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
+++ b/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
@@ -1582,16 +1582,16 @@ namespace Microsoft.Data.SqlClient
 
             ResolveExtendedServerName(serverInfo, !redirectedUserInstance, connectionOptions);
 
-            Boolean disableTnir = ShouldDisableTnir(connectionOptions);
+            bool disableTnir = ShouldDisableTnir(connectionOptions);
 
             long timeoutUnitInterval = 0;
 
-            Boolean isParallel = connectionOptions.MultiSubnetFailover || (connectionOptions.TransparentNetworkIPResolution && !disableTnir);
-
+            bool isParallel = connectionOptions.MultiSubnetFailover || (connectionOptions.TransparentNetworkIPResolution && !disableTnir);
 
             if (isParallel)
             {
                 float failoverTimeoutStep = connectionOptions.MultiSubnetFailover ? ADP.FailoverTimeoutStep : ADP.FailoverTimeoutStepForTnir;
+
                 // Determine unit interval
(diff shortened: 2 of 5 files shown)


src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -251,7 +251,54 @@ internal bool IsDNSCachingBeforeRedirectSupported
         // If the context is expiring within the next 10 mins, then create a new context, irrespective of if another thread is trying to do the same.
         private static readonly TimeSpan _dbAuthenticationContextUnLockedRefreshTimeSpan = new TimeSpan(hours: 0, minutes: 10, seconds: 00);
 
-        private static HashSet<int> s_transientErrors = new HashSet<int>();
+        // The errors in the transient error set are contained in
+        // https://azure.microsoft.com/en-us/documentation/articles/sql-database-develop-error-messages/#transient-faults-connection-loss-and-other-temporary-errors
+        private static readonly HashSet<int> s_transientErrors = new HashSet<int>
```
benrr101, line 256: Dumb question - why are we using a hashset here instead of just a statically defined array?

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnectionFactory.cs
```diff
@@ -145,7 +145,7 @@ override protected DbConnectionInternal CreateConnection(DbConnectionOptions opt
                     opt = new SqlConnectionString(opt, instanceName, false /* user instance=false */, null /* do not modify the Enlist value */);
                     poolGroupProviderInfo = null; // null so we do not pass to constructor below...
                 }
-                result = new SqlInternalConnectionTds(identity, opt, key.Credential, poolGroupProviderInfo, "", null, redirectedUserInstance, userOpt, recoverySessionData, pool, key.AccessToken, applyTransientFaultHandling: applyTransientFaultHandling, key.AccessTokenCallback);
+                result = new SqlInternalConnectionTds(identity, opt, key.Credential, poolGroupProviderInfo, "", null, redirectedUserInstance, userOpt, recoverySessionData, applyTransientFaultHandling: applyTransientFaultHandling, key.AccessToken, pool, key.AccessTokenCallback);
```
benrr101, line 148: To me, it doesn't seem to be necessary to have the argument label here when the variable is the same as the argument name

benrr101 (approved):
Looks pretty good to me, my only comments are probably out of scope since the overall goal is to align the two codebases, not necessarily clean them up. Feel free to address or ignore and comment on my comments.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -107,12 +107,13 @@ internal sealed class SqlInternalConnectionTds : SqlInternalConnection, IDisposa
...
 
-        // Connection re-route limit
-        internal const int _maxNumberOfRedirectRoute = 10;
-
         // CONNECTION AND STATE VARIABLES
         private readonly SqlConnectionPoolGroupProviderInfo _poolGroupProviderInfo; // will only be null when called for ChangePassword, or creating SSE User Instance
         private TdsParser _parser;
+
+        // Connection re-route limit
+        internal const int _maxNumberOfRedirectRoute = 10;
```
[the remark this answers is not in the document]
edwardneal: It was purely to align the codebases - I didn't look at code style. If we're changing it anyway, we might as well correct it where it makes sense.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -297,21 +297,21 @@ internal SessionData CurrentSessionData
...
         internal class SyncAsyncLock
         {
-            private SemaphoreSlim semaphore = new SemaphoreSlim(1);
+            private SemaphoreSlim _semaphore = new SemaphoreSlim(1);
 
             internal void Wait(bool canReleaseFromAnyThread)
             {
-                Monitor.Enter(semaphore); // semaphore is used as lock object, no relation to SemaphoreSlim.Wait/Release methods
-                if (canReleaseFromAnyThread || semaphore.CurrentCount == 0)
+                Monitor.Enter(_semaphore); // semaphore is used as lock object, no relation to SemaphoreSlim.Wait/Release methods
```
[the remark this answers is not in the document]
edwardneal: Some fairly specific locking is needed within TdsParser to prevent a deadlock when closing the connection, and so it needs to manually Enter/Exit the lock. We're likely using the semaphore rather than a dedicated lock object to reduce memory usage.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
edwardneal: Likely to indicate that the error codes are meant to be unique. It might also be to align with the SqlConfigurableRetryFactory class' definition of transient error numbers.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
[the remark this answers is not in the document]
benrr101: Definitely outside the scope of this PR, so again, I'll overlook it :)

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
benrr101: I guess I'd look at a static array of error codes and just assume that they are unique. I could see using the hashset if we allowed the user to add more errors to the list.
Again, outside the scope of this PR, so I won't make a fuss about it.

benrr101 (approved):
Thanks for looking at the comments - looking forward to getting this merged 🚀 🚢

paulmedynski (approved):

PR #3326, opened 2025-05-02
Merge | Netfx SMI Classes

benrr101:
**Description**: This PR is a bit bigger than usual, but it's mostly cut/paste of classes from the netfx project to the common project. There are a few additional changes going on in this PR though that are worth calling out:

* Some light cleanup of `SmiContextFactory` to just make sure it's following some style guidelines
* Making `SmiEventStream.Dispose` obsolete (since it is notated to be obsolete and only throws)
* Removing majority of the code in SmiRequestExecutor since it is notated as obsolete and should have never been used.
  * This class is weird - it abstract and used in `SqlCommand`, but there are no classes that inherit from it. All the virtual code paths overridden in it throw, so any codepath that uses it will always throw. I suspect it is ultimately not used. I will investigate if this entire codepath can be deleted.
* `TriggerAction` enum definitions derive directly from `EMDEventType` enum definitions, which seems completely pointless to me. So, I rolled the `EMDEventType` values into the `TriggerAction` enum and deleted `EMDEventType`.
* The two `SmiXetterAccessMap` classes were combined together with `#if NETFRAMEWORK`.
* Renamed `SmiEventSink_DeferedProcessing` to `SmiEventSink_DeferredProcessing` (typo in Deferred)

**Testing**: Project builds locally, will wait for CI to call it good.


benrr101:
> I'm pretty sure that everything here will be deleted when the context connection PR set (#2996 and its follow-ups) is finished. Every path here is already unused because we block context connections in the connection string. Do you still want to merge the files?
> 
> I completely agree with removing the event sink code from SqlCommand. I think the original SQLCLR methodology had a set of custom event sinks and a custom SmiLink derivative in the SqlAccess assembly, so it's used to support that.

nooooooooooooooooooooooooooooooooooooooooooooooooooooooo 😭 
I *just* "discovered" that almost all of the context connection code can be obliterated. And i was really excited to blow it all up ☹️ However, I'm not super sure about all of it being exploded. There's some code paths that are moved in this PR that I can't confidently say are not being used (eg, abstract classes that are never inherited from, but they are still being referenced in a way I can't safely say is never called). I mean for my own satisfaction/ego, I'd like to do the merge and then pare it away in future PRs. But I also want to be nice to our wonderful contributors and let them get the glory when they deserve it.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SmiXetterAccessMap.cs
```diff
@@ -0,0 +1,125 @@
...
+        private const bool _ = false;
+
+        #if NETFRAMEWORK
+        private static readonly bool[,] s_isGetterAccessValid =
+        {
+            // Getters as columns (abbreviated from XetterTypeCode names)
+            // SqlDbTypes as rows
+            //            bool, byte, bytes, chars, strng, int16, int32, int64, singl, doubl, sqldec, date,  guid, varmd, Xetr, time, dtost
+            /*BigInt*/   { _,    _,    _,     _,     _,     _,     _,     X,     _,     _,     _,      _,     _,    _,     _,    _,    _, },
+            /*Binary*/   { _,    _,    X,     _,     _,     _,     _,     _,     _,     _,     _,      _,     _,    _,     _,    _,    _, },
```
mdaigle, line 25: This is some crazy stuff 😟

mdaigle (approved):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SmiXetterAccessMap.cs
```diff
@@ -0,0 +1,125 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System.Diagnostics;
+
+namespace Microsoft.Data.SqlClient.Server
+{
+    /// <summary>
+    /// Formal encoding of SMI's metadata-to-ITypedSetter/-from-ITypedGetter validity rules
+    /// </summary>
+    internal partial class SmiXetterAccessMap
```
paulmedynski, line 12: What's a Xetter?

paulmedynski (approved):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SmiXetterAccessMap.cs
benrr101: Afaik it's [G|S]etter - so this has both getters and setters

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SmiXetterAccessMap.cs
benrr101: What's crazier - a lot of this is just going to disappear in a later PR :)

PR #3329, opened 2025-05-04
Fix SqlCached buffer async read with continue edge case.

### LINKED ISSUE #3319: System.Xml.XmlException: Unexpected end tag
### Describe the bug
Unable to iterate through results with an xml column with a sql reader.

```
System.Xml.XmlException: Unexpected end tag. Line 1, position 13.
   at System.Xml.XmlTextReaderImpl.Throw(Exception e)
   at System.Xml.XmlTextReaderImpl.Throw(String res, String arg)
   at System.Xml.XmlTextReaderImpl.ParseDocumentContent()
   at System.Xml.XmlWriter.WriteNode(XmlReader reader, Boolean defattr)
   at System.Data.SqlTypes.SqlXml.get_Value()
   at Microsoft.Data.SqlClient.SqlCachedBuffer.ToString()
   at Microsoft.Data.SqlClient.SqlBuffer.get_String()
   at Microsoft.Data.SqlClient.SqlDataReader.GetString(Int32 i)
```

### To reproduce
https://github.com/kstrohminfor/SqlReaderIssue

### Expected behavior
The sql reader should be able to iterate through results with an xml column without error.

### Further technical details
Microsoft.Data.SqlClient version: 6.1.0-preview1.25120.4
.NET target: .NET 8.0
SQL Server version: 2022
Operating system: Windows 11

Wraith2:
fixes #3319 

SqlCachedBuffer is used only when reading xml data. When adding the async-continue feature in https://github.com/dotnet/SqlClient/pull/3161 the read function was changed to enable the use of stored snapshots and this edge case was missed. 

When reading in async continue mode the first two packets are read in replay mode and at the end of the second packet if NeedsMoreData is returned an it is safe to mark as a continue point the snapshot marked the packet end of the second packet as the continue point. The code was incorrectly discarding the data that had already been read into the List<byte[]> from the first two packets which lead to the contents of the read being the remaining portion of the data from the packets after the continue point.

This change alters the logic used to decide whether to store the data which has been read into the snapshot and the logic used to decide whether the data fetched from the snapshot should be used.

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCachedBuffer.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCachedBuffer.cs
@@ -45,9 +45,9 @@ namespace Microsoft.Data.SqlClient
             if (isAvailable)
             {
                 cachedBytes = stateObj.TryTakeSnapshotStorage() as List<byte[]>;
-                if (cachedBytes != null && !isStarting && !isContinuing) 
+                if (isStarting)
                 {
-                    stateObj.SetSnapshotStorage(null);
+                    cachedBytes = null;
                 }
             }
  
@@ -63,7 +63,6 @@ namespace Microsoft.Data.SqlClient
             {
                 return result;
             }
-
 
             // For now we  only handle Plp data from the parser directly.
             Debug.Assert(metadata.metaType.IsPlp, "SqlCachedBuffer call on a non-plp data");
@@ -105,10 +104,7 @@ namespace Microsoft.Data.SqlClient
 
                     if (returnAfterAdd)
                     {
-                        if (isStarting || isContinuing)
-                        {
-                            stateObj.SetSnapshotStorage(cachedBytes);
-                        }
+                        stateObj.SetSnapshotStorage(cachedBytes);
                         return result;
                     }
 
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCachedBuffer.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCachedBuffer.cs
@@ -39,7 +39,7 @@ namespace Microsoft.Data.SqlClient
         {
             buffer = null;
 
-            (bool isAvailable, bool isStarting, bool isContinuing) = stateObj.GetSnapshotStatuses();
+            (bool isAvailable, bool isStarting, _) = stateObj.GetSnapshotStatuses();
 
             List<byte[]> cachedBytes = null;
             if (isAvailable)
@@ -55,7 +55,6 @@ namespace Microsoft.Data.SqlClient
             {
                 cachedBytes = new List<byte[]>();
             }
-
 
             // the very first length is already read.
             TdsOperationStatus result = parser.TryPlpBytesLeft(stateObj, out ulong plplength);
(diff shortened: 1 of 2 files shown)


ErikEJ:
Possible to add a test? Seems like a test was missing.

Wraith2:
Yes. Working on that today. It's a little complicated to reproduce cleanly.

Wraith2:
Test added. I tried for several hours to make an effective reduction but the combination of packet spanning and lengths just isn't conducive to simplification. Ultimately the failure causing logic is copied from the reproduction supplied relatively directly and altered for the test harness that we're using.

I'm unable to test locally because the build fails, I've raised https://github.com/dotnet/SqlClient/issues/3330 to cover that problem.
@dotnet/sqlclientdevteam could someone trigger the CI please.

ErikEJ:
@Wraith2 Forgot to push the test? 

To get pwsh on your system:

`winget install --id Microsoft.PowerShell --source winget`

Wraith2:
> @Wraith2 Forgot to push the test?
> 
> To get pwsh on your system:
> 
> `winget install --id Microsoft.PowerShell --source winget`

The test is there, it's called CanReadXmlData. 
I have powershell installed. As far as I know it's installed by default on 11. It's there and usable it's just not called `pwsh`.

ErikEJ:
Pwsh is Powershell Core, based on .NET Core. There are two "Powershells"

Wraith2:
I forgot xml reads to string canonicalize. test is updated. please trigger again.

Wraith2:
I've reduced the iterations on the new test by 10x by finding a packet size which causes the failure much earlier. Previously it needed 800 iterations to find the failure, this one will find it in <100 iterations. That means the overall iterations to ensure failure is much lower. I hope this makes the test now succeed meaning that the only problem was that it was too slow in combination with all the other tests in stage 2 of the tests. Please re-run CI.

paulmedynski (approved):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCachedBuffer.cs
```diff
@@ -39,15 +39,15 @@ internal static TdsOperationStatus TryCreate(SqlMetaDataPriv metadata, TdsParser
...
 
-            (bool isAvailable, bool isStarting, bool isContinuing) = stateObj.GetSnapshotStatuses();
+            (bool isAvailable, bool isStarting, _) = stateObj.GetSnapshotStatuses();
 
             List<byte[]> cachedBytes = null;
             if (isAvailable)
             {
                 cachedBytes = stateObj.TryTakeSnapshotStorage() as List<byte[]>;
-                if (cachedBytes != null && !isStarting && !isContinuing) 
+                if (isStarting)
```
mdaigle, line 48: Relates to this question posed on the original change:
https://github.com/dotnet/SqlClient/pull/3161/files#r1994120794

Resetting the snapshot storage here wiped out the data in the snapshot that future SqlCachedBuffers would need.
Now, we only set `cachedBytes` to null when _starting_. Any new data we read is added to the `cachedBytes` and stored in the snapshot as needed, not to be overwritten until we've completed this read.

mdaigle (approved):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCachedBuffer.cs
Wraith2: Yes. Overall while it works I'm not happy with how complex the code is but I can't find a way to do everything it needs to more simply.

PR #3347, opened 2025-05-14
Reset negotiateAuth if SNI doesn't work

twsouthwick:


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/NegotiateSspiContextProvider.cs
```diff
@@ -10,13 +11,16 @@ namespace Microsoft.Data.SqlClient
 {
     internal sealed class NegotiateSspiContextProvider : SspiContextProvider
     {
-        private NegotiateAuthentication? _negotiateAuth = null;
+        private NegotiateAuthentication? _negotiateAuth;
```
vonzshik, line 14: A bit unrelated, but isn't `NegotiateAuthentication` `IDisposable`? And I do not see it ever being disposed.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -441,12 +437,14 @@ internal void Connect(ServerInfo serverInfo,
 
             _connHandler.pendingSQLDNSObject = null;
 
+            string[] serverSpn = null;
```
paulmedynski, line 440: The CreatePhysicalSNIHandle() methods don't seem to have any API docs, but the `serverSpn` argument being passed here appears to be an array of SPNs, and is called `spns` on the method definitions.  Can we make this variable name plural to match?  `serverSpns`

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -13556,15 +13552,14 @@ private TdsOperationStatus TryProcessUDTMetaData(SqlMetaDataPriv metaData, TdsPa
...
-                                       + "         _sniSpnBuffer = {17}\n\t"
-                                       + "         _errors = {18}\n\t"
-                                       + "         _warnings = {19}\n\t"
-                                       + "         _attentionErrors = {20}\n\t"
-                                       + "         _attentionWarnings = {21}\n\t"
-                                       + "         _statistics = {22}\n\t"
-                                       + "         _statisticsIsInTransaction = {23}\n\t"
-                                       + "         _fPreserveTransaction = {24}"
-                                       + "         _fParallel = {25}"
+                                       + "         _errors = {17}\n\t"
```
paulmedynski, line 13555: Oh my - this format string is a programming error magnet!  Just commentary - no action requested.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/NegotiateSspiContextProvider.cs
```diff
@@ -10,13 +11,16 @@ namespace Microsoft.Data.SqlClient
...
 
         protected override bool GenerateSspiClientContext(ReadOnlySpan<byte> incomingBlob, IBufferWriter<byte> outgoingBlobWriter, SspiAuthenticationParameters authParams)
         {
             NegotiateAuthenticationStatusCode statusCode = NegotiateAuthenticationStatusCode.UnknownCredentials;
 
             _negotiateAuth ??= new(new NegotiateAuthenticationClientOptions { Package = "Negotiate", TargetName = authParams.Resource });
+
+            Debug.Assert(_negotiateAuth.TargetName == authParams.Resource, "SSPI resource does not match TargetName");
```
paulmedynski, line 22: Is this mismatch important to notice in the field?  What's the impact of a mismatch?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/SspiContextProvider.cs
```diff
@@ -10,14 +11,34 @@ internal abstract class SspiContextProvider
...
+#else
+            string[] serverSpns
+#endif
+            )
         {
             _parser = parser;
             _physicalStateObj = physicalStateObj;
             _serverInfo = serverInfo;
 
+#if NETFRAMEWORK
```
paulmedynski, line 37: This would become a simple foreach loop.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/SspiContextProvider.cs
```diff
@@ -27,46 +48,48 @@ private protected virtual void Initialize()
 
         protected abstract bool GenerateSspiClientContext(ReadOnlySpan<byte> incomingBlob, IBufferWriter<byte> outgoingBlobWriter, SspiAuthenticationParameters authParams);
 
-        internal void SSPIData(ReadOnlySpan<byte> receivedBuff, IBufferWriter<byte> outgoingBlobWriter, string serverSpn)
+        internal void WriteSSPIContext(ReadOnlySpan<byte> receivedBuff, IBufferWriter<byte> outgoingBlobWriter)
         {
             using var _ = TrySNIEventScope.Create(nameof(SspiContextProvider));
 
-            if (!RunGenerateSspiClientContext(receivedBuff, outgoingBlobWriter, serverSpn))
+            if (_authParamValue is SspiAuthenticationParameters authParam && RunGenerateSspiClientContext(receivedBuff, outgoingBlobWriter, authParam))
```
paulmedynski, line 55: This disappears and you just have the foreach from line 61.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/SspiContextProvider.cs
```diff
@@ -27,46 +48,48 @@ private protected virtual void Initialize()
...
-
-        internal void SSPIData(ReadOnlySpan<byte> receivedBuff, IBufferWriter<byte> outgoingBlobWriter, ReadOnlySpan<string> serverSpns)
-        {
-            using var _ = TrySNIEventScope.Create(nameof(SspiContextProvider));
-
-            foreach (var serverSpn in serverSpns)
+            else if (_authParamValue is SspiAuthenticationParameters[] authParams)
             {
-                if (RunGenerateSspiClientContext(receivedBuff, outgoingBlobWriter, serverSpn))
+                foreach (var p in authParams)
```
paulmedynski, line 61: `p` -> `authParam` or `param`

paulmedynski (commented):
Some comments about handling multipe and single SPNs.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
twsouthwick: Yes and yes - it would be a great candidate for an interpolated string (especially on .NET 8+).

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/NegotiateSspiContextProvider.cs
twsouthwick: I've updated the debug statement to explain it a bit more

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/SspiContextProvider.cs
twsouthwick: I'm not sure what you mean. If that if statement is true, it's returned. There's a catch all error handling at the end.

To make it clearer, I've separated things out into separate methods so it reads a bit better.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/SspiContextProvider.cs
```diff
@@ -35,9 +34,9 @@ string[] serverSpns
             _serverInfo = serverInfo;
 
 #if NETFRAMEWORK
-            _authParamValue = CreateAuthParams(serverSpn);
+            _authParam = CreateAuthParams(serverSpn);
```
paulmedynski, line 37: I guess this is a decent tradeoff since the netfx implementation still exists.  Once it is merged into a single implementation, I suspect we won't need a singular argument and the accompanying code here.

paulmedynski (approved):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/NegotiateSspiContextProvider.cs
paulmedynski: Yes separate PR is good since I think the changes will be spread out.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/SspiContextProvider.cs
paulmedynski: I was thinking that you would have only a List of auth params, and it could contain multiple entries until you get a successful SSPI interaction, and then you just replace it with a list of 1.

Your latest changes are fine as-is though.

paulmedynski (approved):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/SspiContextProvider.cs
```diff
@@ -27,46 +47,64 @@ private protected virtual void Initialize()
...
-            foreach (var serverSpn in serverSpns)
+        /// <summary>
+        /// If we have multiple, we need to loop through them, and then identify the correct one for future use.
+        /// </summary>
+        private bool TryRunMultiple(ReadOnlySpan<byte> receivedBuff, IBufferWriter<byte> outgoingBlobWriter)
+        {
+            if (_authParams is { })
             {
-                if (RunGenerateSspiClientContext(receivedBuff, outgoingBlobWriter, serverSpn))
+                foreach (var authParam in _authParams)
```
vonzshik, line 76: Correct me if I wrong, but I feel like the whole SSPI code is extremely problematic.

From my understanding, the way SSPI handshake works is that the client first generates some data to send to the server (in this case, it's calling `_negotiateAuth.GetOutgoingBlob` with an empty `ReadOnlySpan`).
Then, the resulting data is sent to the server, which responds with some other data, which is fed to `_negotiateAuth.GetOutgoingBlob`, and the cycle repeats until `_negotiateAuth.GetOutgoingBlob` returns either an error or it's successful.

The main point is, the first call (with an empty `ReadOnlySpan`) will always be successful and as result no other SPN will ever be tried. This is especially important if SqlClient supports login retrying, as that will force each retry to always start with the first SPN.

twsouthwick: I am personally not that familiar with SSPI, but I think you're onto something. The initial call is just to get the first blob; it does appear to always return something with "ContinueNeeded".

I dug into what the `serverSPNs` value even is. It appears to just be an array that is either a single or doubl entry; the double entry has a port included at times. It appears that for this, there should be no looping - instead I need to grab the final entry.

This feels really hacky, but the change to refactor the serverSPNs will probably be a larger change and I'll do a separate PR for that to keep it targeted.

vonzshik (approved):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/SspiContextProvider.cs
twsouthwick: Ok - it wasn't as big of a change as I thought. Now the SNI code paths will just return the value it needs - no need for an array.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/SspiContextProvider.cs
twsouthwick: no need to have this anymore :)

paulmedynski (approved):

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -13700,7 +13696,6 @@ internal string TraceString()
                             _connHandler == null ? "(null)" : _connHandler.ObjectID.ToString((IFormatProvider)null),
                             _fMARS ? bool.TrueString : bool.FalseString,
                             _sessionPool == null ? "(null)" : _sessionPool.TraceString(),
-                            _serverSpn == null ? "(null)" : _serverSpn.Length.ToString((IFormatProvider)null),
```
mdaigle, line 13703: It would be beneficial to keep the resolved SPN in the trace logs.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/NegotiateSspiContextProvider.cs
```diff
@@ -29,6 +31,9 @@ protected override bool GenerateSspiClientContext(ReadOnlySpan<byte> incomingBlo
                 return true;
             }
 
+            _negotiateAuth.Dispose();
```
mdaigle, line 34: Can you please add a comment in code describing why we want to dispose of _negotiateAuth here?

mdaigle (changes requested):

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
twsouthwick: moved the logging into the SSPI provider

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SNI/SNIProxy.cs
```diff
@@ -169,12 +170,12 @@ private static string[] GetSqlServerSPNs(string hostNameOrAddress, string portOr
                 string serverSpnWithDefaultPort = serverSpn + $":{DefaultSqlServerPort}";
                 // Set both SPNs with and without Port as Port is optional for default instance
                 SqlClientEventSource.Log.TryAdvancedTraceEvent("SNIProxy.GetSqlServerSPN | Info | ServerSPNs {0} and {1}", serverSpn, serverSpnWithDefaultPort);
-                return new[] { serverSpn, serverSpnWithDefaultPort };
```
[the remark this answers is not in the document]
twsouthwick: @mdaigle I've pushed an update based on our discussion

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/InstanceNameTest/InstanceNameTest.cs
```diff
@@ -212,10 +212,8 @@ private static string GetSPNInfo(string dataSource, string inInstanceName)
             string serverSPN = "";
             MethodInfo getSqlServerSPNs = sniProxyObj.GetType().GetMethod("GetSqlServerSPNs", BindingFlags.Static | BindingFlags.Public | BindingFlags.NonPublic, null, CallingConventions.Any, getSqlServerSPNsTypesArray, null);
 
-            string[] result = (string[])getSqlServerSPNs.Invoke(sniProxyObj, new object[] { dataSrcInfo, serverSPN });
-
-            string spnInfo = result[0];
-
+            string spnInfo = (string)getSqlServerSPNs.Invoke(sniProxyObj, new object[] { dataSrcInfo, serverSPN });
```
mdaigle, line 215: I think this is spitting out your new object, and so can't be cast directly to a string

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SSPI/NegotiateSspiContextProvider.cs
```diff
@@ -31,11 +31,29 @@ protected override bool GenerateSspiClientContext(ReadOnlySpan<byte> incomingBlo
...
+        {
+            if (_negotiateAuth is { })
+            {
+                if (string.Equals(_negotiateAuth.TargetName, authParams.Resource, StringComparison.Ordinal))
+                {
+                    return _negotiateAuth;
+                }
+
+                // Dispose of it since we're not going to use it now
+                _negotiateAuth?.Dispose();
```
paulmedynski, line 52: Shouldn't need the `?` here.

mdaigle (changes requested):

paulmedynski (dismissed):

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/InstanceNameTest/InstanceNameTest.cs
twsouthwick: yeah - my changes keep affecting that taste. Yay for reflection

paulmedynski (dismissed):

paulmedynski (approved):

paulmedynski:
@twsouthwick - Did you have to resolve any conflicts from your previous commits when merging main?  It's hard to tell from the large merge commit.

twsouthwick:
@paulmedynski I did have to resolve a couple of things - mostly that I had added a type into the SNI directory, and it has been renamed to ManagedSni. I did the merge resolution on GH and hadn't built locally... I've pushed a fix that should succeed

paulmedynski (approved):

mdaigle (dismissed):

paulmedynski (approved):

twsouthwick:
@mdaigle fixed a merge error that only failed on a few builds so it's ready now :) thanks for all the feedback!

mdaigle (approved):

PR #3353, opened 2025-05-16
Merge | TdsParserStateObject PacketHandle usage

edwardneal:
Relates to #1261.

This continues the `TdsParserStateObject` merge work. At present, there's a deviation between netcore and netfx: netcore uses a `PacketHandle` type to shuttle the native or managed packet handles around, while netfx just treats PacketHandle as an alias for `IntPtr`. This PR is focused strictly on merging that handling, so we can move on to the identical methods which use the type.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParserStateObject.netfx.cs
+++ b/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParserStateObject.netfx.cs
@@ -18,8 +18,6 @@ using Microsoft.Data.ProviderBase;
 
 namespace Microsoft.Data.SqlClient
 {
-    using PacketHandle = IntPtr;
-
     internal partial class TdsParserStateObject
     {
         protected SNIHandle _sessionHandle = null;              // the SNI handle we're to work on
--- a/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParserStateObject.netfx.cs
+++ b/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParserStateObject.netfx.cs
@@ -136,27 +136,35 @@ namespace Microsoft.Data.SqlClient
                 ipPreference, cachedDNSInfo, hostNameInCertificate);
         }
 
-        internal bool IsPacketEmpty(PacketHandle readPacket) => readPacket == default;
+        internal bool IsPacketEmpty(PacketHandle readPacket)
+        {
+            Debug.Assert(readPacket.Type == PacketHandle.NativePointerType || readPacket.Type == 0, "unexpected packet type when requiring NativePointer");
+            return IntPtr.Zero == readPacket.NativePointer;
+        }
 
         internal PacketHandle ReadSyncOverAsync(int timeoutRemaining, out uint error)
         {
             SNIHandle handle = Handle ?? throw ADP.ClosedConnectionError();
-            PacketHandle readPacket = default;
-            error = SniNativeWrapper.SniReadSyncOverAsync(handle, ref readPacket, timeoutRemaining);
-            return readPacket;
+            IntPtr readPacketPtr = IntPtr.Zero;
+            error = SniNativeWrapper.SniReadSyncOverAsync(handle, ref readPacketPtr, timeoutRemaining);
+            return PacketHandle.FromNativePointer(readPacketPtr);
         }
 
         internal PacketHandle ReadAsync(SessionHandle handle, out uint error)
         {
-            PacketHandle readPacket = default;
-            error = SniNativeWrapper.SniReadAsync(handle.NativeHandle, ref readPacket);
-            return readPacket;
+            IntPtr readPacketPtr = IntPtr.Zero;
+            error = SniNativeWrapper.SniReadAsync(handle.NativeHandle, ref readPacketPtr);
+            return PacketHandle.FromNativePointer(readPacketPtr);
         }
 
         internal uint CheckConnection() => SniNativeWrapper.SniCheckConnection(Handle);
 
-        internal void ReleasePacket(PacketHandle syncReadPacket) => SniNativeWrapper.SniPacketRelease(syncReadPacket);
-        
+        internal void ReleasePacket(PacketHandle syncReadPacket)
+        {
+            Debug.Assert(syncReadPacket.Type == PacketHandle.NativePointerType, "unexpected packet type when requiring NativePointer");
+            SniNativeWrapper.SniPacketRelease(syncReadPacket.NativePointer);
+        }
+
         [ReliabilityContract(Consistency.WillNotCorruptState, Cer.Success)]
         internal int DecrementPendingCallbacks(bool release)
         {
@@ -375,7 +383,13 @@ namespace Microsoft.Data.SqlClient
 
         private uint GetSniPacket(PacketHandle packet, ref uint dataSize)
         {
-            return SniNativeWrapper.SniPacketGetData(packet, _inBuff, ref dataSize);
+            return SniPacketGetData(packet, _inBuff, ref dataSize);
+        }
+
+        private uint SniPacketGetData(PacketHandle packet, byte[] _inBuff, ref uint dataSize)
+        {
+            Debug.Assert(packet.Type == PacketHandle.NativePointerType, "unexpected packet type when requiring NativePointer");
+            return SniNativeWrapper.SniPacketGetData(packet.NativePointer, _inBuff, ref dataSize);
         }
 
         public void ReadAsyncCallback(IntPtr key, PacketHandle packet, uint error)
@@ -409,7 +423,7 @@ namespace Microsoft.Data.SqlClient
             bool processFinallyBlock = true;
             try
             {
-                Debug.Assert(IntPtr.Zero == packet || IntPtr.Zero != packet && source != null, "AsyncResult null on callback");
+                Debug.Assert(CheckPacket(packet, source), "AsyncResult null on callback");
 
                 if (_parser.MARSOn)
                 {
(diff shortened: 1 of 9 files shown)


mdaigle (dismissed):
All of your changes look reasonable.

Would you be able to add a high level summary to the PacketHandle class docs that covers when the different packet types are used?  In particular, it's not obvious to me when we would use NativePointerType vs NativePacketType.

Bigger picture (and not blocking for this PR, just looking to discuss), do you have any insight into what this design pattern gives us? It feels like we're fighting the type system. I can't think of a reason why we'd want this pattern vs just using the underlying types directly or a ref struct per underlying type, but I may be missing something.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/PacketHandle.Windows.cs
```diff
@@ -22,21 +20,33 @@ internal readonly ref struct PacketHandle
     {
         public const int NativePointerType = 1;
         public const int NativePacketType = 2;
+#if NET
```
benrr101, line 23: So, I saw a similar scenario in the SNIHandle class, and I'm not sure this is a good approach. My thoughts here is that this is a good example where we should have an abstract class and a ManagedPacketHandle and a NativePacketHandle. These are very lightweight classes that just wrap the actual packet handle, which makes me wonder if they're even necessary in the first place. So, my big question is: 
* option a, should we just merge it like this and maybe address it later
* option b, should we introduce some abstraction
* option c, should we consider removing the handle class altogether

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParserStateObject.netfx.cs
```diff
@@ -138,27 +136,35 @@ internal void CreatePhysicalSNIHandle(
                 ipPreference, cachedDNSInfo, hostNameInCertificate);
         }
 
-        internal bool IsPacketEmpty(PacketHandle readPacket) => readPacket == default;
+        internal bool IsPacketEmpty(PacketHandle readPacket)
```
benrr101, line 139: It seems to me this is something that should be moved into the packet handle itself?

edwardneal:
Thanks.

The PacketHandle and SessionHandle designs were added in dotnet/corefx#34044, and they were added to avoid boxing. This boxing was introduced as part of the original implementation of the managed SNI in dotnet/corefx#16589. The original native SNI design had the two types we care about (IntPtr and SNIPacket) and this has evolved from that.

To speculate on the original design pattern: it feels like we're driven partially by marshalling here. When a PacketHandle's `Type == NativePacketType`, we know that it's transporting an `SNIPacket` instance. This type is a SafeHandle derivative, and it's marshalled *into* the unmanaged SNI when we set a packet's data and write it out to the remote instance. Conversely, a PacketHandle's `Type == NativePointerType` when we know that it's transporting nothing more than an IntPtr. This is marshalled *out* of the unmanaged SNI when we receive a packet and read its data in managed code.

Strictly speaking, I'm not sure we'd need to maintain that separation; I'd personally be surprised if the unmanaged SNI DLL had a type separation between the inbound and the outbound packets, so the managed library might not need it either. If there's no separation, we might be able to rely on the normal SafeHandle marshalling mechanisms to let us use SNIPacket instances from end to end. This'd simplify matters, but it'd also increase GC traffic (because every packet received from SQL Server is now a full object with its own finalizer, rather than just an IntPtr.)

If we wanted to avoid that specific regression, another way of representing this might be a generics-based design similar to below:

```csharp
internal abstract class TdsParserStateObject<TOutputPacket, TInputPacket, TSession>
{
    public abstract TSession Session { get; }
    public abstract SetPacketData(TOutputPacket packet, byte[] buffer, int bytesUsed);
    public abstract void ReleasePacket(TInputPacket syncReadPacket);

    // This was in TdsParser, but would need to be be moved into TdsParserStateObject - it's the
    // only place outside this class (besides NativeSspiContextProvider) where PacketHandle and SessionHandle are used
    public void PostReadAsyncForMars();
    // ...
}

internal sealed class TdsParserStateObjectNative : TdsParserStateObject<SNIPacket, IntPtr, SNIHandle>
{ }

internal sealed class TdsParserStateObjectManaged : TdsParserStateObject<SNI.SNIPacket, SNI.SNIPacket, SNI.SNIHandle>
{ }
```

It might be worth considering and benchmarking after these are merged.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParserStateObject.netfx.cs
```diff
@@ -377,7 +383,13 @@ private void ReadSniError(TdsParserStateObject stateObj, uint error)
 
         private uint GetSniPacket(PacketHandle packet, ref uint dataSize)
         {
-            return SniNativeWrapper.SniPacketGetData(packet, _inBuff, ref dataSize);
+            return SniPacketGetData(packet, _inBuff, ref dataSize);
+        }
+
+        private uint SniPacketGetData(PacketHandle packet, byte[] _inBuff, ref uint dataSize)
```
benrr101, line 389: To me this seems unnecessary - just put the contents of this method into GetSniPacket. But if you made this change because netcore does it this way, I'd say we should make netcore do it the simpler way.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParserStateObject.netfx.cs
```diff
@@ -790,7 +810,15 @@ private Task SNIWritePacket(SNIPacket packet, out uint sniError, bool canAccumul
 
 #pragma warning restore 420
 
-        internal bool IsValidPacket(PacketHandle packetPointer) => packetPointer != default;
+        internal bool IsValidPacket(PacketHandle packetPointer)
+        {
+            Debug.Assert(packetPointer.Type == PacketHandle.NativePointerType || packetPointer.Type == PacketHandle.NativePacketType, "unexpected packet type when requiring NativePointer");
+            return (
```
benrr101, line 816: nit: technically the parenthesis at the beginning here are redundant.
if it were me, I'd consider rewriting this like:
```
return (xyz && abc)
    || (xyz && abc);
```

benrr101 (changes requested):
Would like to think through the PacketHandle class before we approve this. But otherwise, I think these will be good changes to move us through the merge classes.
Also, this is going to run into conflicts with #3345...

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/PacketHandle.Windows.cs
edwardneal: I think we can set option b aside - this is a ref struct, so it can't inherit anything. There might be a path to replace this via generics, but there's already a design gap between netcore and netfx. I think we should merge the type as-is, finish the merge of TdsParserStateObject and TdsParserStateObjectNative, then revisit it.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParserStateObject.netfx.cs
edwardneal: This needs to remain as-is, I'm afraid. In netcore, SniPacketGetData is an abstract method. This change syncs up the method definitions so they can be merged in #3355 - commit is [here](https://github.com/dotnet/SqlClient/pull/3355/commits/e06dba95519f717c59e207a174f9f5cb85b8ceb2).

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParserStateObject.netfx.cs
edwardneal: In principle, yes - but this is another abstract method which we merge in #3355, with implementations which vary between TdsParserStateObjectNative and TdsParserStateObjectManaged. The assertions verify that we're not taking a PacketHandle populated for the native SNI and passed it to the managed SNI (and vice versa.)

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParserStateObject.netfx.cs
edwardneal: Agreed. I'll make this change in #3355, when the method is ready to merge.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/PacketHandle.Windows.cs
benrr101: Ok, option A it is
I keep forgetting about ref structs ... kinda annoying technicalities imho

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParserStateObject.netfx.cs
benrr101: Hmm, ok... I guess the change I'm suggesting would make more sense if the packet handles were typed native and managed. But as you said above, we can't really do that today.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParserStateObject.netfx.cs
benrr101: I'll try to hold you to it :)

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParserStateObject.netfx.cs
benrr101: Ok, i think that's ok for now

benrr101 (approved):

mdaigle (approved):

PR #3357, opened 2025-05-19
APIScan | Fix Builder Mistake

benrr101:
**Description**: Made a mistake when I wrote #3354 - forgot to store the modified builder in the builder variable.

**Testing**: Kinda disappointed it didn't get caught in any CI pathways. Ideally I should write up a test for this specific scenario, but I'm concerned this will be a huge undertaking versus fixing it before it gets released.

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ActiveDirectoryAuthenticationProvider.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ActiveDirectoryAuthenticationProvider.cs
@@ -556,7 +556,7 @@ namespace Microsoft.Data.SqlClient
             #if NETFRAMEWORK
             if (_iWin32WindowFunc is not null)
             {
-                builder.WithParentActivityOrWindow(_iWin32WindowFunc);
+                builder = builder.WithParentActivityOrWindow(_iWin32WindowFunc);
             }
             #endif
 


mdaigle (commented):
I don't think we have any test coverage of UI based app usage and it's hard to do in an automated way. But as far as I know, this is actually the only bit of code that's specific to UI based apps. What do you think about doing a manual verification?

benrr101:
@mdaigle If I had any idea how to do that ... sure

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ActiveDirectoryAuthenticationProvider.cs
```diff
@@ -556,7 +556,7 @@ private IPublicClientApplication CreateClientAppInstance(PublicClientAppKey publ
             #if NETFRAMEWORK
             if (_iWin32WindowFunc is not null)
             {
-                builder.WithParentActivityOrWindow(_iWin32WindowFunc);
+                builder = builder.WithParentActivityOrWindow(_iWin32WindowFunc);
```
paulmedynski, line 559: Is this actually necessary?  Chaining typically returns the same object the method was called on, not a new object.  Pretty sure WithParentActivityOrWindow() is setting internal state on builder and then returning builder.  I'd be very surprised if it was returning a different instance of a Builder.

The API docs don't say either way, but the source code returns `this`:

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/428838e8ac9588e34c115e30e410e66855700b52/src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs#L287

I think your original change was fine (i.e. it worked) and maybe that's why there were no test failures.  This might be slightly more robust, but I doubt MSAL will change its chaining behaviour, and break a bunch of calling code.

paulmedynski (changes requested):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ActiveDirectoryAuthenticationProvider.cs
cheenamalhotra: Because config is local to builder creation, I think it is necessary to store the updated config in the builder instance.

You can validate the builder assignment need by updating another config that overrides something else that can be validated easily in a standalone console app, where you can directly work with MSAL APIs for testing the theory.

benrr101:
Backing PR out since as per discussion w/ @paulmedynski the builder just updates it's internal state and returns itself. No need to store the output of any With* call since the builder is modified internally.
Although both storing and not storing are valid, why have an extra PR for something you don't need?

PR #3309, opened 2025-04-28
Merge | Port SqlDataSourceConverter, widen test coverage

edwardneal:
#3243 reminded me that we didn't completely port the TypeConverters for `SqlConnectionStringBuilder`.

This PR makes sure that the WinForms PropertyGrid behaviour is consistent between .NET Core and .NET Framework for the `DataSource` and the `FailoverPartner` properties.

Both this and `NetworkLibraryConverter` also had [very poor](https://app.codecov.io/gh/dotnet/SqlClient/blob/main/src%2FMicrosoft.Data.SqlClient%2Fsrc%2FMicrosoft%2FData%2FSqlClient%2FSqlConnectionStringBuilder.cs#L712) code coverage; I've added tests to broaden this.

I'd appreciate a CI & Codecov run.

PR #3369, opened 2025-05-20
NativeAOT | Remove DataSet.ReadXml, correct type name

edwardneal:
## Description

This PR's diff is much smaller than it looks! There's an embedded XML resource, and adjusting this meant that the indentation levels changed. When that's accounted for, the PR becomes much smaller - around 300 lines, rather than 7800.

This PR relates to #1942 and #1261. At present, there are two identical copies of the `Microsoft.Data.SqlClient.SqlMetaData.xml` embedded resource in the netfx and the netcore projects. This PR merges these with a single file. More importantly though, this embedded resource is used only within `DbMetaDataFactory`; it's deserialized using `DataSet.ReadXml`, where it is eventually used in `SqlConnection.GetSchema`.

`DataSet.ReadXml` isn't trim-safe - it uses XML serialization in the background. I've replaced this with a simple parsing routine which runs against the embedded resource and adds the rows manually.

The net result is smaller, faster and more efficient. Sizoscope confirms that this PR cuts the file size of a simple application by 3.8MB, from 20.4MB to 16.6MB. Benchmark results for this in isolation are:

| Method               | Mean       | Error    | StdDev   | Ratio | RatioSD | Gen0     | Gen1    | Allocated | Alloc Ratio |
|--------------------- |-----------:|---------:|---------:|------:|--------:|---------:|--------:|----------:|------------:|
| Before | 1,819.7 μs | 25.07 μs | 23.45 μs |  1.00 |    0.02 | 187.5000 | 39.0625 | 829.81 KB |        1.00 |
| After |   738.9 μs |  3.25 μs |  2.71 μs |  0.41 |    0.01 |  62.5000 | 19.5313 | 307.74 KB |        0.37 |

Separately, I noticed that my earlier PR #3309 generated two warnings when publishing because it wasn't using a fully-qualified type name. I've adjusted this.

## Issues

Relates to #1942.
Relates to #1261.
Follows up #3309.

## Testing

All tests continue to pass.

There's a lot of validation of the structure of the XML in the embedded resource, and this is done with debug assertions. I've treated this as hardcoded at the point of build, so haven't used exceptions for this. I can change this if needed.


src/Microsoft.Data.SqlClient/netfx/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -1038,7 +1038,7 @@ public SqlConnectionStringBuilder(string connectionString) { }
         /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnectionStringBuilder.xml' path='docs/members[@name="SqlConnectionStringBuilder"]/DataSource/*'/>
         [System.ComponentModel.DisplayNameAttribute("Data Source")]
         [System.ComponentModel.RefreshPropertiesAttribute(System.ComponentModel.RefreshProperties.All)]
-        [System.ComponentModel.TypeConverter("Microsoft.Data.SqlClient.SqlConnectionStringBuilder+SqlDataSourceConverter")]
+        [System.ComponentModel.TypeConverter("Microsoft.Data.SqlClient.SqlConnectionStringBuilder+SqlDataSourceConverter, Microsoft.Data.SqlClient")]
```
mdaigle, line 1041: what do these additions do?

src/Microsoft.Data.SqlClient/src/Resources/Microsoft.Data.SqlClient.SqlMetaData.xml
mdaigle, line 1: What is the actual reason for having this file? Is it an optimization to avoid a round trip to the server to retrieve the common schema elements? It feels odd to have it fully duplicated here if the server owns the definitive schema.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbMetaDataFactory.cs
```diff
@@ -502,18 +478,229 @@ private bool IncludeThisColumn(DataColumn sourceColumn, string[] hiddenColumnNam
...
+                IgnoreWhitespace = true
             };
             using XmlReader reader = XmlReader.Create(XmlStream, settings);
-            _metaDataCollectionsDataSet.ReadXml(reader);
+
+            // Build up the schema DataSet manually, then load data from XmlStream. The schema of the DataSet is defined at:
+            // * https://learn.microsoft.com/en-us/sql/connect/ado-net/common-schema-collections
+            // * https://learn.microsoft.com/en-us/sql/connect/ado-net/sql-server-schema-collections
+            // Building the schema manually is necessary because DataSet.ReadXml uses XML serialization. This is slow, and it
+            // increases the binary size of AOT assemblies by approximately 4MB.
```
mdaigle, line 497: Can you help me understand how the choice of loading here impacts the AOT binary size? Doesn't this all happen at runtime?

mdaigle (commented):
A few questions to help with my understanding

src/Microsoft.Data.SqlClient/netfx/ref/Microsoft.Data.SqlClient.cs
edwardneal: They fully qualify the type name of the SqlDataSourceConverter. I noticed when testing that although a standard application find the property's TypeConverter, the NativeAOT linker raises a warning unless I include the assembly name.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbMetaDataFactory.cs
edwardneal: There are two core issues, both of which come down to the fact that DataSet.ReadXml uses reflection-based XML serialization.

1. If one of the types in SqlMetaData.xml isn't used elsewhere, it won't be included in the published app and the deserialization routine won't be able to find it. That's not a problem for core types like `int`, but might be an issue for `System.Data.Common.GroupByBehavior`. This issue would appear at runtime.
2. ReadXml is the only method which requires the XML serialization infrastructure. The AOT linker notes the requirement and includes the infrastructure in the published binary (which is where the ~3.8MB size improvement from its removal comes from.)

src/Microsoft.Data.SqlClient/src/Resources/Microsoft.Data.SqlClient.SqlMetaData.xml
edwardneal: Not exactly. It bootstraps some of the collections returned from SqlConnection.GetSchema, providing the SQL queries needed to retrieve the server-side schema collections (list of these [here](https://learn.microsoft.com/en-us/sql/connect/ado-net/sql-server-schema-collections).) Separately, it contains the hardcoded values returned when someone calls SqlConnection.GetSchema and queries for the ADO provider schema - the list of server-side schema collections, the list of reserved words, a description of how the various SQL Server types map to .NET types and their capabilities (i.e. nullability, which value in the SqlDbType enum it maps to.)

The master list of those schema collections is [here](https://learn.microsoft.com/en-us/sql/connect/ado-net/common-schema-collections).

paulmedynski: I don't really understand what the data in this file is used for, but could we generate some C# code at compile time instead of parsing XML at runtime?  Is the XML expected to change at runtime such that we need to re-parse it?

paulmedynski (changes requested):
Asking for some clarification on why we're reading this XML at runtime.

src/Microsoft.Data.SqlClient/src/Resources/Microsoft.Data.SqlClient.SqlMetaData.xml
edwardneal: It's not expected to change at runtime - it's completely constant.

I did originally have a version which inserted the records via code, but found this was harder to initially error-check; it meant reviewing both the method and the data we insert. I'm happy to proceed with that version if you'd prefer?

paulmedynski: The changes you made are a step in the right direction - no need to hold this work up that does improve the AOT experience.  I've opened #3372 to explore this further.

paulmedynski (approved):

mdaigle (approved):

PR #3376, opened 2025-05-26
Merge | Port #528 and #2091 to netfx

edwardneal:
## Description

This ports an earlier set of memory usage reductions from netcore to netfx.

## Issues

Relates to #1261.
Ports #528 and #2091.

## Testing

Unit tests pass locally. I don't see anything specifically targeting these changes, so I assume that they're covered by the overall test suite - they're fairly frequently exercised.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
@@ -35,7 +35,7 @@ namespace Microsoft.Data.SqlClient
     {
         private static int _objectTypeCount; // EventSource Counter
         private const int MaxRPCNameLength = 1046;
-        internal readonly int ObjectID = Interlocked.Increment(ref _objectTypeCount); private string _commandText;
+        internal readonly int ObjectID = Interlocked.Increment(ref _objectTypeCount);
 
         internal sealed class ExecuteReaderAsyncCallContext : AAsyncCallContext<SqlCommand, SqlDataReader, CancellationTokenRegistration>
         {
@@ -113,6 +113,7 @@ namespace Microsoft.Data.SqlClient
             }
         }
 
+        private string _commandText;
         private CommandType _commandType;
         private int? _commandTimeout;
         private UpdateRowSource _updatedRowSource = UpdateRowSource.Both;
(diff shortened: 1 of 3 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
```diff
@@ -2645,7 +2646,7 @@ private Task<int> InternalExecuteNonQueryAsync(CancellationToken cancellationTok
             {
                 s_diagnosticListener.WriteCommandError(operationId, this, _transaction, e);
                 source.SetException(e);
-                context.Dispose();
+                context?.Dispose();
```
paulmedynski, line 2649: I don't see how context can be null here.  It is constructed on line 2616, and then never assigned to again.

Is the compiler maybe confused due to the same identifier being used on line 2634 in the static lambda?

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlCommand.cs
```diff
@@ -2910,39 +3007,67 @@ private Task<int> InternalExecuteNonQueryAsync(CancellationToken cancellationTok
...
+
+                        command.CleanupAfterExecuteNonQueryAsync(task, source, operationId);
+                    },
+                    scheduler: TaskScheduler.Default
+                );
             }
             catch (Exception e)
             {
                 source.SetException(e);
+                context?.Dispose();
```
paulmedynski, line 3045: Same here - context won't be null.

paulmedynski (changes requested):
A couple of spurious ? checks.

I didn't dig into the full logic that was added/modified - I don't have enough experience with SqlCommand for that yet.

paulmedynski (approved):

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlCommand.cs
```diff
@@ -2992,6 +3117,7 @@ private Task<SqlDataReader> InternalExecuteReaderAsync(CommandBehavior behavior,
         {
             SqlClientEventSource.Log.TryCorrelationTraceEvent("<sc.SqlCommand.ExecuteReaderAsync|API|Correlation> ObjectID {0}, behavior={1}, ActivityID {2}", ObjectID, (int)behavior, ActivityCorrelator.Current);
             SqlConnection.ExecutePermission.Demand();
+            Guid operationId = default(Guid);
```
benrr101, line 3120: Nit: Technically `(Guid)` isn't necessary here.

benrr101 (approved):
Since this is porting changes into netfx, I'm onboard with these changes

Issue #3331, opened 2025-05-06
6.1 Preview 1: SqlDataReader.GetChars ignoring BufferIndex?


Tornhoof:
### Describe the bug
I tried the new 6.1 PReview1 bits, a few of my tests failed with wrong data output, I tracked it down to sqlDataReader.GetChars apparently ignoring the buffer index. In the GetPooledChars loop below the buffer is always written at bufferIndex 0 ignoring the offset parameter. If you run the same code with 6.0.2 it works.

### To reproduce
```csharp
using System.Buffers;
using System.Data;
using Microsoft.Data.SqlClient;

namespace SqlRepro
{
    internal class Program
    {
        static async Task Main(string[] args)
        {
            const CommandBehavior commandBehavior = CommandBehavior.SequentialAccess | CommandBehavior.SingleResult;
            await using var sqlConnection = new SqlConnection("Server=localhost;Database=DatabaseTests;TrustServerCertificate=True;Trusted_Connection=True;");
            await sqlConnection.OpenAsync();
            const int length = 32000;
            const string sqlCharWithArg = "SELECT CONVERT(BIGINT, 1) AS [Id], CONVERT(NVARCHAR(MAX), @input) AS [Value];";
            var input = string.Create(length, length, (span, state) =>
            {
                for (int i = 0; i < state; i++)
                {
                    span[i] = (char) Random.Shared.Next(0x30, 0x5A);
                }
            });
            await using var sqlCommand = new SqlCommand();
            sqlCommand.Connection = sqlConnection;
            sqlCommand.CommandTimeout = 0;
            sqlCommand.CommandText = sqlCharWithArg;
            sqlCommand.Parameters.Add(new SqlParameter("@input", SqlDbType.NVarChar, -1) {Value = input});
            await using var sqlReader = await sqlCommand.ExecuteReaderAsync(commandBehavior);
            var succ = await sqlReader.ReadAsync();
            long id = sqlReader.GetInt64(0);
            if (id != 1)
            {
                Console.WriteLine("Id not 1");
            }

            var sliced = GetPooledChars(sqlReader, 1);
            if (!sliced.SequenceEqual(input.ToCharArray()))
            {
                Console.WriteLine("sliced != input");
            }
        }

        private static char[] GetPooledChars(SqlDataReader sqlDataReader, int ordinal)
        {
            var buffer = ArrayPool<char>.Shared.Rent(8192);
            int offset = 0;
            while (true)
            {
[...]

Issue #3274, opened 2025-04-12
SqlClient 6.0.1 DataReader hangs on trying to read 1 row with 150Mb value in nvarchar(max) column - consume huge amount of memory, and do not react to cancelationToken


AlexanderKot:
### Describe the bug

In app I am developing happens error as result it stored several huge strings in nvarchar(max) column - 150Mb json.
Then trying to read such record with DataReader application starts consume huge amount of memory and do not react to cancellation attempts using cancellationToken.
 

It looks like this
![Image](https://github.com/user-attachments/assets/9e837f5d-9e4d-42a3-934d-edf10424933e)
On picture resource consumption by App with only 1 operation running - reading this row.
Trying to read 1 row with 150Mb column it consume up to 10G of mem.

Let assume app receive 150Mb from DB, and then convert it to 300Mb UTF16 string
What for it consume other 9.5Gb?

Data was saved to DB with same client, quick and with adequate resource consumption.

Trying to debug next state can be observed:
![Image](https://github.com/user-attachments/assets/b6ff0516-db49-4a89-9cb3-852e6b72cb5f)

In our app code execution stopped on  await _reader.ReadAsync(cancellationToken)  
and cancellationToken is activated by closing web page.

At this time control is in TP Worker thread processing read request

![Image](https://github.com/user-attachments/assets/eca73b25-ab9f-44d0-a830-da5007eda452)
There some cancellationToken token checked only at the beginning of operation and seems that is another token, as it is not activated.

Most of the time thread can be stopped in next methods called from SqlDataReader:

![Image](https://github.com/user-attachments/assets/54ea860e-18c4-417e-891b-177aa395c03a)

![Image](https://github.com/user-attachments/assets/62f0ee6f-294f-4498-8f4a-08272ea310a0)

### Expected behavior
1. React to cancellation token
2. Do read operation with adequate memory consumption.

Wraith2:
You're hitting https://github.com/dotnet/SqlClient/issues/593 
Speed and memory consumption should be improved for large strings by https://github.com/dotnet/SqlClient/pull/3161 which will be in the next preview release.

Wraith2:
performance and memory usage change slighlty. see benchmarks in https://github.com/dotnet/SqlClient/pull/3161#issuecomment-2696001820 for some information. 
The memory usage is never going to be a lot better while reading into a string. The best I can ever do is 2x string length + some overhead. It is currently more like 3x + overhead.

[...]

PR #3377, opened 2025-05-26
Improve async string perf and fix reading chars with initial offset.

Wraith2:
## Description

For performance related changes see background information in https://github.com/dotnet/SqlClient/issues/3274#issuecomment-2869878530 and the preceeding conversation. Reading a string in async mode was being slower than expected so I investigated. I will annotate each change with an explanation.

## Issues

fixes https://github.com/dotnet/SqlClient/issues/3331
fixes https://github.com/dotnet/SqlClient/issues/3274 

## Testing

Tests have been added which cover both issues that are being fixed.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
@@ -13088,7 +13088,7 @@ namespace Microsoft.Data.SqlClient
                 ||
                 (buff.Length >= offst + len)
                 ||
-                (buff.Length == (startOffsetByteCount >> 1) + 1),
+                (buff.Length >= (startOffsetByteCount >> 1) + 1),
                 "Invalid length sent to ReadPlpUnicodeChars()!"
             );
             charsLeft = len;
@@ -13152,7 +13152,10 @@ namespace Microsoft.Data.SqlClient
                         }
                         else
                         {
-                            newbuf = new char[offst + charsRead];
+                            // grow by an arbitrary number of packets to avoid needing to reallocate
+                            //  the newbuf on each loop iteration of long packet sequences which causes
+                            //  a performance problem as we spend large amounts of time copying and in gc
+                            newbuf = new char[offst + charsRead + (stateObj.GetPacketSize() * 8)];
                             rentedBuff = false;
                         }
 
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
@@ -13096,9 +13096,9 @@ namespace Microsoft.Data.SqlClient
             // If total length is known up front, the length isn't specified as unknown 
             // and the caller doesn't pass int.max/2 indicating that it doesn't know the length
             // allocate the whole buffer in one shot instead of realloc'ing and copying over each time
-            if (buff == null && stateObj._longlen != TdsEnums.SQL_PLP_UNKNOWNLEN && len < (int.MaxValue >> 1))
+            if (buff == null && stateObj._longlen != TdsEnums.SQL_PLP_UNKNOWNLEN && stateObj._longlen < (int.MaxValue >> 1))
             {
-                if (supportRentedBuff && len < 1073741824) // 1 Gib
+                if (supportRentedBuff && stateObj._longlen < 1073741824) // 1 Gib
                 {
                     buff = ArrayPool<char>.Shared.Rent((int)Math.Min((int)stateObj._longlen, len));
                     rentedBuff = true;
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
@@ -13133,8 +13133,7 @@ namespace Microsoft.Data.SqlClient
 
             totalCharsRead = (startOffsetByteCount >> 1);
             charsLeft -= totalCharsRead;
-            offst = totalCharsRead;
-
+            offst += totalCharsRead;
 
             while (charsLeft > 0)
             {
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
@@ -1360,6 +1360,11 @@ namespace Microsoft.Data.SqlClient
             return false;
         }
 
+        internal int GetPacketSize()
+        {
+            return _inBuff.Length;
+        }
+
         ///////////////////////////////////////
         // Buffer read methods - data values //
         ///////////////////////////////////////
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
@@ -4027,6 +4027,28 @@ namespace Microsoft.Data.SqlClient
                     Debug.Assert(_stateObj._permitReplayStackTraceToDiffer || prev.Stack == trace, "The stack trace on subsequent replays should be the same");
                 }
             }
+
+            public int CurrentPacketIndex
+            {
+                get
+                {
+                    int value = -1;
+                    if (_current != null)
+                    {
+                        PacketData current = _firstPacket;
+                        while (current != null)
+                        {
+                            value += 1;
+                            if (current == _current)
+                            {
+                                break;
+                            }
+                            current = current.NextPacket;
+                        } 
+                    }
+                    return value;
+                }
+            }
 #endif
             public bool ContinueEnabled => !LocalAppContextSwitches.UseCompatibilityAsyncBehaviour;
 
--- a/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
+++ b/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
@@ -13267,7 +13267,7 @@ namespace Microsoft.Data.SqlClient
         {
             int charsRead = 0;
             int charsLeft = 0;
-            
+
             if (stateObj._longlen == 0)
             {
                 Debug.Assert(stateObj._longlenleft == 0);
@@ -13277,11 +13277,11 @@ namespace Microsoft.Data.SqlClient
 
             Debug.Assert((ulong)stateObj._longlen != TdsEnums.SQL_PLP_NULL, "Out of sync plp read request");
             Debug.Assert(
-                (buff == null && offst == 0) 
-                || 
+                (buff == null && offst == 0)
+                ||
                 (buff.Length >= offst + len)
                 ||
-                (buff.Length == (startOffsetByteCount >> 1) + 1), 
+                (buff.Length >= (startOffsetByteCount >> 1) + 1),
                 "Invalid length sent to ReadPlpUnicodeChars()!"
             );
             charsLeft = len;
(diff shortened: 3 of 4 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -13133,8 +13133,7 @@ bool writeDataSizeToSnapshot
 
             totalCharsRead = (startOffsetByteCount >> 1);
             charsLeft -= totalCharsRead;
-            offst = totalCharsRead;
```
Wraith2, line 13136: This is all that's required to fix #3331, done in both netcore and netfx.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -13088,7 +13088,7 @@ bool writeDataSizeToSnapshot
                 ||
                 (buff.Length >= offst + len)
                 ||
-                (buff.Length == (startOffsetByteCount >> 1) + 1),
+                (buff.Length >= (startOffsetByteCount >> 1) + 1),
```
Wraith2, line 13091: This change simply allows the target buffer we're copying into to be a bit larger than we actually need. This is required so that we can make the later change which grows the buffer more than we need.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -13152,7 +13152,10 @@ bool writeDataSizeToSnapshot
                         }
                         else
                         {
-                            newbuf = new char[offst + charsRead];
+                            // grow by an arbitrary number of packets to avoid needing to reallocate
```
Wraith2, line 13154: As this comment explains we grow the buffer in a slightly unusual way. Typically buffer sizes are doubled or use some reasonably fraction (like 1.5). This can be unhelpful when we know that we're likely to run in low memory conditions like containers and that the data could be very large. It would be very easy to have a 512 Mib buffer and be unable to resize it to 1Gib because of limits. 

The reason that we need to resize the buffer is that we've encountered some more data and we aren't certain of the total data size. In the case of a multipacket string we would keep encountering more data with each packet and keep creating new buffers, copying data and then doing it all again resulting in huge amounts of memory throughput and time spent in memcpy, see profile traces in https://github.com/dotnet/SqlClient/issues/3274#issuecomment-2869878530

This strategy allows the buffer to be slightlty oversized each time more space is needed. The multiple of packetsize means that we can accept at least that number of packets before needing another resize (because packet data cannot be entirely made up of char data) which reduces the reallocation frequenty by the multiplier. 8 is an entirely arbitrary choice guided by the opposing needs to be 1) conservative about growth and 2) reduce frequency of reallocation.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -13096,9 +13096,9 @@ bool writeDataSizeToSnapshot
             // If total length is known up front, the length isn't specified as unknown 
             // and the caller doesn't pass int.max/2 indicating that it doesn't know the length
             // allocate the whole buffer in one shot instead of realloc'ing and copying over each time
-            if (buff == null && stateObj._longlen != TdsEnums.SQL_PLP_UNKNOWNLEN && len < (int.MaxValue >> 1))
+            if (buff == null && stateObj._longlen != TdsEnums.SQL_PLP_UNKNOWNLEN && stateObj._longlen < (int.MaxValue >> 1))
```
Wraith2, line 13099: I've found this line of code suspicious for quite a while but haven't been confident about changing it. Understanding what it's doing requires both understanding what the internals of the code are doing and  how plp data is represented in the tds protocol, please see TDS spec 2.2.5.2.3:

> A value sent as PLP can be either NULL, a length followed by chunks (as defined by PLP_CHUNK), or an unknown length token followed by chunks

The important part is that the total length can be sent up front. In practice when we encounter a plp data stream we read the length and store the value in `stateObj._longlen`. The `len` parameter is not the same thing, it's a parameter to the function to tell it how much data we're expecting it to read and if we're reading chunks we usually don't know. 

Now look at the check it's checking the buffer, then the  plp total length to see if it's unknown, then the length we're asking for. Why does it matter if we're asking for "as much as possible" which is what int.MaxValue /2 is interpretted as? It doesn't matter as far as I can see in this context. The context is deciding how large of a buffer to allocate. 

Now imagine we change it like this. If we have no buff yet, and we know the `stateObj._longlen` (which is what the protocol gave us as the known total length) is not the sprecific value for unknown, and the value of `stateObj._longlen` is less than the largest possible char buffer we're allowed to allocate by the runtime, then we could allocate a buffer of that size.

With that change all tests continue to pass and we're more likely to allocate correctly size buffers for plp data at the start of the read process. This avoids later copying resizing etc.

Wraith2:
/cc @tomhoof, @AlexanderKot this should fix your issues.

@dotnet/sqlclientdevteam can I get a CI kick? I'd really really like to get this in to preview 2 because it fixes all the known issues with async continue.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
```diff
@@ -4027,6 +4032,28 @@ internal void CheckStack(string trace)
                     Debug.Assert(_stateObj._permitReplayStackTraceToDiffer || prev.Stack == trace, "The stack trace on subsequent replays should be the same");
                 }
             }
+
+            public int CurrentPacketIndex
```
cheenamalhotra, line 4036: What's the significance of this method? Was this something you were using for debugging?

paulmedynski (dismissed):

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
Wraith2: len will be unknown length, which is int.max/2 as the original comment mentions. So all valid plp longlen's will be less than it in that case. If we are passed a len value which is less than the chunk size it'll read the chunk length then read the len bytes that it wants from that chunk. The min(len, _longlen) still makes sense which is why I didn't change it.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -13088,17 +13088,17 @@ bool writeDataSizeToSnapshot
...
                 (buff.Length >= offst + len)
                 ||
-                (buff.Length == (startOffsetByteCount >> 1) + 1),
+                (buff.Length >= (startOffsetByteCount >> 1) + 1),
                 "Invalid length sent to ReadPlpUnicodeChars()!"
             );
             charsLeft = len;
 
             // If total length is known up front, the length isn't specified as unknown 
             // and the caller doesn't pass int.max/2 indicating that it doesn't know the length
```
benrr101, line 13097: I think with this change, the comment will need to be changed as well - for this check we no longer care what the caller requested.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
benrr101: I'm a bit suspicious of this change ... `SQL_PLP_UNKNOWNLEN` is a magic number that indicates the plp data is of unknown length, but `(int.MaxValue >> 1)` is magic number that indicates the *caller* doesn't know the length. Also, _longlen is a ulong but len is just an int. So I'm confused why the magic maxint/2 value would be useful to the _longlen. Ie, there's a wide range between `UNKNOWNLEN` and `int.MaxValue >> 1` that isn't being accounted for, afaict.

Either way, rewriting the comment above the check and giving int.MaxValue >> 1 a constant would probably help with understanding what's going on here.

benrr101 (changes requested):
I can't say I know much about the internals here, so I appreciate the callouts on what changes are being made. However, I'm not understanding what the _longlen changes are really doing. Please take a look at my comments and I'll approve if you can explain it without code changes :)

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
Wraith2: My interpretation of the int.MaxValue >> 1 value is that it's simply a derivation of the SQL_PLP_UNKNOWNLEN constant which is a byte count converted to a char count since chars are twice as long. As far as I know the protocol only deals with byte count lengths. If you're reading bytes you'd expect the unknown value and if you're working in chars you /2, there are a number of other places in the codebase that this operation is done.

I've spent a lot of time debugging through this code over a few years because of the async-continue work and other improvements before that. When doing so I've observed that the piece of code that I'm changing here is almost never executed because the len is the sentinel int.max/2 value. Inspecting the state on several of those occasions it wasn't clear to me why the check was written the way that it is. 
As I mentioned in https://github.com/dotnet/SqlClient/pull/3377#discussion_r2107841161 I found this suspicious, but _only_ suspicious. It's not clearly incorrect and the code functions but it does so less well than it possibly can. I appreciate the need for confidence in the change but without being the original author I can't identify the original intention of the check and whether it guards against some obscure condition that we can't identify. Unfortunately I can't claim to have complete knowledge of why it is the way it is so I also can't say that the change is entirely safe. It's just too complicated for that level of assertion.

Without this change everything continues to work. If we don't allocate the entire array up front then we carry on into the code path where we repeatedly resize the array which I've changed in this PR. If you'd prefer to drop this particular change from the PR and keep all the other changes I'd be fine with it, it's still a bug fix and net gain PR.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
benrr101: Haha, I'd like something that explains what the check is doing. Right now, it says that the if statment is checking for int.max/2 from the caller (ie, `len`), but the if statement no longer checks `len`. Even if it's not the perfect explanation what's going on, it'd still be good to get some understanding of what the check tries to do or why it's being made. Basically, I just don't want stale/inaccurate information in the comments - it'll make our lives a lot easier if the code and comments agree with each other.

Alternatively, if you think the comment is still accurate as-is, let me know and maybe we can come up with a better wording to explain the behavior.

Wraith2: "Invalid length sent to ReadPlpUnicodeChars()!" to "Too short buffer passed to ReadPlpUnicodeChars()!" ?

or for the other one:
```csharp
// If total data length is known up front from the plp header by being not SQL_PLP_UNKNOWNLEN
//  and the number of chars required is less than int.max/2 allocate the entire buffer now to avoid
//  later needing to repeatedly allocate new target buffers and copy data as we discover new data
```

benrr101: The other one - the line I added the comment to. I think what you suggested is a good revision 👍

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
ericsampson: @Wraith2, I just wanted to express that you're a hero for keeping at this : )

Wraith2: Thanks.

Any thought on:
> If you'd prefer to drop this particular change from the PR and keep all the other changes I'd be fine with it, it's still a bug fix and net gain PR.

team?

mdaigle (approved):

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
mdaigle: Your explanation makes sense to me, but I do wonder why callers would still pass in int.MaxValue >> 1 if the length is known. At least this guards against that mistake as much as possible.

mdaigle:
@benrr101 this one needs an updated review from you

benrr101 (approved):
Thanks for making the changes!

PR #3380, opened 2025-05-27
Add unit test project

mdaigle:
## Description
This PR adds a new unit test project to the solution and grants that project access to the driver's internal types. The intention of this project is to hold strict unit tests that do not rely on outside resources (azure/local dbs, AKV, etc.). Tests in the unit test project are run as part of the pipeline, before functional tests.

Access to internal types is achieved using the `InternalsVisibleTo` assembly directive. Internal types must be included in the reference assembly for the project to compile via MSBuild. Automatic reference assembly generation is disabled to avoid conflicts between the auto generated and manually generated assemblies.

Many tests that currently live in the functional tests project could move to this project. I've left that as a future improvement to keep this change targeted.

### Why is this important?
Today, most of our testing of internal types uses reflection, but this is problematic for a number of reasons:
 - reflection is slow, error-prone, not type-safe, not refactor-safe, not discoverable (via IDE references), and increases friction to write tests
 - usage of reflection as standard practice encourages access of truly private members of a class
We need a better unit test setup where we can quickly and easily write comprehensive unit tests for internal classes.

### Why not expose internals to integration tests (a.k.a manual tests)?
We could do this, but integration tests are the best way we have to test the public API of our driver. As much as possible, public API testing shouldn't invoke private/internal functionality to generate testing conditions. Integration tests should be stable even as internals are updated. I'm working towards the following testing methodology:
- Unit tests
  - lowest level
  - test units of code in isolation
  - fast
  - no external resources
  - access to internal types
- Integration tests
  - high level
  - test public API surfaces
  - slower
  - can use external resources
  - no access to internal types

## Testing

I added one existing internal type to the reference assemblies and added a test to access that type to prove that internals are visible to the unit test project.

## Guidelines

Please review the contribution guidelines before submitting a pull request:

- [Contributing](/CONTRIBUTING.md)
- [Code of Conduct](/CODE_OF_CONDUCT.md)
- [Best Practices](/policy/coding-best-practices.md)
- [Coding Style](/policy/coding-style.md)
- [Review Process](/policy/review-process.md)


build.proj
```diff
@@ -56,14 +56,24 @@
     <NetCoreDriver Include="**/netcore/**/Microsoft.Data.SqlClient*.csproj" />
     <AKVProvider Include="**/add-ons/**/AzureKeyVaultProvider/*.csproj" />
 
+    <UnitTests Include="**/tools/TDS/TDS/TDS.csproj" />
+    <UnitTests Include="**/tools/TDS/TDS.EndPoint/TDS.EndPoint.csproj" />
+    <UnitTests Include="**/tools/TDS/TDS.Servers/TDS.Servers.csproj" />
```
paulmedynski: Not sure if this counts as "external resources" or not.  It will require a functioning network stack, etc.  Most of our TDS generation and parsing should be testable without any of that.

I think we should leave all of these helpers/utilities out until we actually need them in this project (if ever).

eng/pipelines/common/templates/steps/run-all-tests-step.yml
```diff
@@ -53,6 +53,20 @@ steps:
     condition: succeededOrFailed()
 
 - ${{if eq(parameters.operatingSystem, 'Windows')}}:
+  - task: MSBuild@1
```
paulmedynski: Not sure why we have 2 ways to run tests in the pipelines.  This one uses MSBuild and our build.proj targets, and the ones above use dotnet CLI and the csproj files directly.  We should probably just pick one method.

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -1941,6 +1945,41 @@ public sealed class SqlConfigurableRetryFactory
         public static SqlRetryLogicBaseProvider CreateNoneRetryProvider() { throw null; }
     }
 }
+namespace Microsoft.Data.SqlClient.ConnectionPool
```
paulmedynski: So why does MSBuild need this in the ref API?  Is it so our ReferenceType=Package targets can build/run?  Is it also a problem if we build/run via dotnet CLI?

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft.Data.SqlClient.UnitTests.csproj
```diff
@@ -0,0 +1,82 @@
...
+    <OutputPath>$(BinFolder)$(Configuration).$(Platform).$(AssemblyName)</OutputPath>
+    <IsTestProject>true</IsTestProject>
+  </PropertyGroup>
+  <!-- XUnit and XUnit extensions -->
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="$(MicrosoftNETFrameworkReferenceAssembliesVersion)" Condition="$(TargetGroup) == 'netfx'">
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+      <PrivateAssets>all</PrivateAssets>
+    </PackageReference>
+    <PackageReference Condition="$(TargetGroup) == 'netfx'" Include="System.Runtime.InteropServices.RuntimeInformation" Version="$(SystemRuntimeInteropServicesRuntimeInformationVersion)" />
```
paulmedynski: Do we need all of these references?  Can you remove any unused ones?

paulmedynski (changes requested):
Looks great overall, just a few comments/questions.

build.proj
mdaigle: Thanks, copy pasted these from functional tests and forgot to remove them.

eng/pipelines/common/templates/steps/run-all-tests-step.yml
mdaigle: I agree, I think I'd prefer to standardize on dotnetcli. It's getting more active attention than msbuild.

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
mdaigle: Exactly. When we do a ReferenceType=package test run, we ingest MDS from a nuget package. That nuget package contains the reference assembly compiled from these files in the ref directory. Because the ref directly is manually maintained, we need to explicitly add anything we want to show up in the final ref assembly that ends up in the nuget package. When you compile using a dependency from a nuget package, you preferentially use any reference assemblies included in the package, then swap in the full dll at runtime. If we don't add these files here, the tests won't compile when using ReferenceType=package, even if we've marked internals as visible to the UnitTests assembly.

paulmedynski: Alright that makes sense for now.  In the future, I think we should stop unit testing via package, and only integration test via package (without internals).  That will allow our unit tests to access all the internals like they need without polluting our ref APIs.  It will also ensure our integration testing is truly black-box.

paulmedynski (dismissed):

paulmedynski (approved):

build.proj
```diff
@@ -179,8 +203,57 @@
...
+
   <!-- Run all unit tests applicable to the host OS. -->
-  <Target Name="RunTests" DependsOnTargets="RunFunctionalTests;RunManualTests"/>
+  <Target Name="RunUnitTests" DependsOnTargets="RunUnitTestsWindows;RunUnitTestsUnix"/>
+  
+  <!-- Run all unit tests applicable to Windows. -->
+  <Target Name="RunUnitTestsWindows" Condition="'$(IsEnabledWindows)' == 'true'">
+    <PropertyGroup>
+      <TestCommand>
+        $(DotnetPath)dotnet test "@(UnitTestsProj)"
```
benrr101, line 216: I'd really like it if we can start as simply as possible with these tests. The way I look at it, the current tests are encumbered by a lot of cruft that's made it much more difficult to maintain. Ideally, for this new project, we should try to start as bare-bones as possible and only add cruft when it's absolutely necessary to make it work with our existing infra.

So, things I'd like to see removed:
* ReferenceType (unit tests shouldn't depend on package, only project)
* TestTargetOS (This just seems very confusing - target framework, target os, whatever, and then filters based on os/platform?)
* TestSet (there should be no need for multiple sets in unit tests, run them all, all the time)
* Filters (there should be no need for filters in unit tests, run them all, all the time)

Really, I'd like it if we can just do `dotnet test ${unittestproj)` and be done with it. That should be the baseline, and we should only add to it if absolutely necessary.

eng/pipelines/common/templates/steps/build-and-run-tests-netcore-step.yml
```diff
@@ -64,11 +64,18 @@ steps:
     msbuildArchitecture: x64
     msbuildArguments: '-t:BuildTestsNetCore -p:ReferenceType=${{parameters.referenceType }} -p:TargetNetCoreVersion=${{parameters.TargetNetCoreVersion }} -p:TestMicrosoftDataSqlClientVersion=${{parameters.NugetPackageVersion }} -p:Configuration=${{parameters.configuration }}'
 
+- task: DotNetCoreCLI@2
+  displayName: 'Run Unit Tests for ${{parameters.TargetNetCoreVersion }}'
+  inputs:
+    command: test
+    projects: 'src\Microsoft.Data.SqlClient\tests\UnitTests\Microsoft.Data.SqlClient.UnitTests.csproj'
+    arguments: '-p:Platform=${{parameters.platform }} -p:TestTargetOS="${{parameters.TestTargetOS }}" -p:TargetNetCoreVersion=${{parameters.TargetNetCoreVersion }}  -p:ReferenceType=${{parameters.referenceType }} -p:Configuration=${{parameters.configuration }} -p:TestMicrosoftDataSqlClientVersion=${{parameters.NugetPackageVersion }} --no-build -v n --filter "category!=nonnetcoreapptests&category!=failing&category!=nonwindowstests"'
```
benrr101: Please see similar comments from the build.proj...

Specifically, filters seem completely unnecessary
* Platform tests should be included/excluded by preprocessor directives (same as code in common project)
* Failing tests should be marked with `Fact(Skip=...)` which auto skips them w/o need for category
* NonWindowsTests should be handled via TestTargetOS - why do we need both???

src/Microsoft.Data.SqlClient/netfx/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -152,23 +154,23 @@ public class SqlAuthenticationParameters
         /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlAuthenticationParameters.xml' path='docs/members[@name="SqlAuthenticationParameters"]/ctor/*'/>
         protected SqlAuthenticationParameters(Microsoft.Data.SqlClient.SqlAuthenticationMethod authenticationMethod, string serverName, string databaseName, string resource, string authority, string userId, string password, System.Guid connectionId, int connectionTimeout) { }
         /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlAuthenticationParameters.xml' path='docs/members[@name="SqlAuthenticationParameters"]/AuthenticationMethod/*'/>
-        public Microsoft.Data.SqlClient.SqlAuthenticationMethod AuthenticationMethod { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
+        public Microsoft.Data.SqlClient.SqlAuthenticationMethod AuthenticationMethod { [System.Runtime.CompilerServices.CompilerGeneratedAttribute] get { throw null; } }
```
benrr101: Kinda think all these changes shouldn't be part of this PR...

src/Microsoft.Data.SqlClient/netfx/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -1956,6 +1958,41 @@ public sealed class SqlConfigurableRetryFactory
         public static SqlRetryLogicBaseProvider CreateNoneRetryProvider() { throw null; }
     }
 }
+namespace Microsoft.Data.SqlClient.ConnectionPool
```
benrr101: Also not sure what this has to do with unit tests ...

src/Microsoft.Data.SqlClient/netfx/src/AssemblyInfo.cs
```diff
@@ -0,0 +1,3 @@
+﻿using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("UnitTests")]
```
benrr101: Out of curiosity, can this be done via the new-fangled csproj properties?

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft.Data.SqlClient.UnitTests.csproj
```diff
@@ -0,0 +1,49 @@
...
+    <AssemblyName>UnitTests</AssemblyName>
+    <TargetGroup Condition="$(TargetFramework.StartsWith('net4'))">netfx</TargetGroup>
+    <TargetGroup Condition="$(TargetGroup) == ''">netcoreapp</TargetGroup>
+    <RuntimeIdentifier Condition="'$(TargetGroup)'=='netfx'">win</RuntimeIdentifier>
+    <RuntimeIdentifier Condition="'$(TargetGroup)'=='netfx' AND $(ReferenceType.Contains('Package')) AND !$(Platform.Contains('AnyCPU'))">win-$(Platform)</RuntimeIdentifier>
+    <IntermediateOutputPath>$(ObjFolder)$(Configuration).$(Platform).$(AssemblyName)</IntermediateOutputPath>
+    <OutputPath>$(BinFolder)$(Configuration).$(Platform).$(AssemblyName)</OutputPath>
+    <IsTestProject>true</IsTestProject>
+  </PropertyGroup>
+  <!-- XUnit and XUnit extensions -->
```
benrr101: In common project, I'm moving to break these into three itemgroups:
* References for all platforms
* References for netfx
* References for netcore

This way we don't need to repeat conditions for each package reference, we can have the condition once for the itemgroup

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
edwardneal: I think it'd also be a good idea to control whether we enable `InternalsVisibleTo` on MDS based on the same build configuration. Besides making sure that we don't expose parts of the test harness to downstream clients, it also means that the IL trimmer can work more effectively once we enable that functionality.

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft.Data.SqlClient.UnitTests.csproj
```diff
@@ -0,0 +1,49 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <AssemblyName>UnitTests</AssemblyName>
+    <TargetGroup Condition="$(TargetFramework.StartsWith('net4'))">netfx</TargetGroup>
```
benrr101, line 4: This confirms my suspicion that the parameters we're feeding dotnet for this project are mostly cruft. I'd really like it if we can all the conditions going into this and the below properties

src/Microsoft.Data.SqlClient/netfx/src/AssemblyInfo.cs
mdaigle: Only in .NET
Framework doesn't support the csproj property.

build.proj
paulmedynski: I agree with all of this, but only if we can easily mark runtime/OS specific tests such that by default they only execute where they are supported.  Maybe the Arcade stuff can help here, or maybe we can just roll our own.

eng/pipelines/common/templates/steps/build-and-run-tests-netcore-step.yml
paulmedynski: Again, I think tests that aren't supported on the OS running the tests should simply not execute.  No need for filters.

src/Microsoft.Data.SqlClient/netfx/ref/Microsoft.Data.SqlClient.cs
paulmedynski: He's adding the UnitTests project _and_ the first tests of ConnectionPool in the same PR.  Yes, they could be split apart.

mdaigle: It shows that the InternalsVisibleTo attribute is functioning properly, which is an important part of this project. It doesn't matter that it's connection pool related, I just happened to pick an internal type that I'll need later on for connection pool testing.

build.proj
```diff
@@ -187,8 +202,51 @@
...
+  <!-- Run all unit tests applicable to Unix. -->
+  <Target Name="RunUnitTestsUnix" Condition="'$(IsEnabledWindows)' != 'true'">
+    <PropertyGroup>
+      <TestCommand>
+        $(DotnetPath)dotnet test "@(UnitTestsProj)"
+        --no-build
+        -v n
+        -p:Configuration=$(Configuration)
+        -p:TargetNetCoreVersion=$(TF)
+        -p:TestTargetOS=Unixnetcoreapp
```
mdaigle, line 241: I need to keep this split between windows and linux. The TestTargetOS feeds back into the Directory.build.props to determine the target frameworks. Refactoring that would impact other test projects and is a large change than I'd like to tackle in this PR.

benrr101: Understood, I kinda figured that would break a lot of things. Don't worry, we'll wield the muckrake soon enough.

src/Microsoft.Data.SqlClient/netfx/ref/Microsoft.Data.SqlClient.cs
benrr101: Yep, discussed offline, no problems with the logic for adding it as a demonstration.

Though - @mdaigle  discovered if we only run the tests with project reference type we can avoid doing it altogether! (if I understood it correctly)

eng/pipelines/common/templates/steps/build-and-run-tests-netcore-step.yml
benrr101: Yep, @paulmedynski I agree. This will all be worked out as part of the last steps of the common project. The teaser I'll give is I basically want to be able to #if WINDOWS. There's plenty of examples for it, but the way we have things right now, it's quite complicated to work out.

build.proj
benrr101: Agreed. We'll get there. Malcolm was able to strip some of it out, so I'm happy with that for now.

benrr101 (approved):
Looks much better now, thanks!

build.proj
```diff
@@ -182,13 +177,9 @@
     <MSBuild Projects="@(ManualTests)" Properties="TestTargetOS=$(TestOS)netcoreapp;$(TestProjectProperties);Platform=$(Platform);" Condition="$(ReferenceType.Contains('Package'))" />
   </Target>
 
-  <Target Name="BuildUnitTestsNetFx" DependsOnTargets="RestoreTestsNetFx" Condition="'$(IsEnabledWindows)' == 'true'">
+  <Target Name="BuildUnitTestsNetFx" DependsOnTargets="RestoreTestsNetFx" Condition="'$(IsEnabledWindows)' == 'true' AND $(ReferenceType.Contains('Project'))">
```
paulmedynski, line 181: Here's a case where we should be able to _build_ the UnitTests for .NET Framework on any OS.  I will figure this out with my Linux build work.

build.proj
paulmedynski: Yeah, I have made a bunch of progress on this on my Linux build branch.  I'll figure it all out eventually :)

paulmedynski (approved):

build.proj
mdaigle: I think there's no technical limitation, there's just a lot of intricate artifact path determination we do based on OS.

benrr101 (approved):

paulmedynski (approved):

PR #3383, opened 2025-05-28
Fix sequential text reader bug and add covering test

### LINKED ISSUE #3361: SqlSequentialTextReader ReadAsync returns 0 before the end of a file
### Description
SqlSequentialTextReader ReadAsync returns 0 before the end of a file is reached when the number of characters requested varies.

It reproduces when the number of characters requested is different on each read. If reading with the same constant count of characters to read (e.g. 4096) it will always return slightly oversized buffer that will trigger a read on each iteration from SqlDataReader and everything will be fine.

In PrepareByteBuffer ArrayPool<byte>.Shared returns instance with 8192 size for requested byteBufferSize 4097 (4096 chars).  _decoder. Convert returns 4096 bytesUsed so it's not an encoding issue.
The second iteration requests 1 chare less. It creates a buffer with 4096 length that is entirely in _leftOverBytes size limit. It skips the read from SqlDataReader here https://github.com/dotnet/SqlClient/blob/6d2473d801cadada160815f47f4d672395fb0aca/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlSequentialTextReader.cs#L175
and charsRead 0 at this time (unbuffered characters count read) returned as read bytes, which means the end of a file for the consumer.

### To reproduce

```c#
	var reader = await command.ExecuteReaderAsync(CommandBehavior.SequentialAccess, cancellationToken);
	var textReader = dataReader.DbDataReader.GetTextReader(0);
        
        var buffer = new char[4096];
        var result = -1;
        var start = 0;
        while (result != 0)
        {
            result = await textReader.ReadAsync(buffer, start, buffer.Length - start);
            start++;
        }
```

### Expected behavior
A number of characters ready to process on current iteration returned even if it was buffered earlier.

### Further technical details
Microsoft.Data.SqlClient version: 6.0.2
.NET target: .NET 9.0
SQL Server version: SQL Server 2022
Operating system: Windows 10

Wraith2:
Fixes #3361 

When using single byte input the test provided triggered a condition where the next read would skip data in available internal buffer because it was exactly, and not less than, the required amount.

/cc @m0a0k0s

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlSequentialTextReader.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlSequentialTextReader.cs
@@ -172,7 +172,7 @@ namespace Microsoft.Data.SqlClient
                         byte[] byteBuffer = PrepareByteBuffer(charsNeeded, out int byteBufferUsed);
 
                         // Permit a 0 byte read in order to advance the reader to the correct column
-                        if ((byteBufferUsed <= byteBuffer.Length) || (byteBuffer.Length == 0))
+                        if (byteBufferUsed <= byteBuffer.Length || byteBuffer.Length == 0)
                         {
                             SqlDataReader reader = _reader;
                             if (reader != null)
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlSequentialTextReader.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlSequentialTextReader.cs
@@ -172,7 +172,7 @@ namespace Microsoft.Data.SqlClient
                         byte[] byteBuffer = PrepareByteBuffer(charsNeeded, out int byteBufferUsed);
 
                         // Permit a 0 byte read in order to advance the reader to the correct column
-                        if ((byteBufferUsed < byteBuffer.Length) || (byteBuffer.Length == 0))
+                        if ((byteBufferUsed <= byteBuffer.Length) || (byteBuffer.Length == 0))
                         {
                             SqlDataReader reader = _reader;
                             if (reader != null)
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataReaderTest/DataReaderTest.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataReaderTest/DataReaderTest.cs
@@ -606,6 +606,56 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests
         }
 
         [ConditionalFact(typeof(DataTestUtility), nameof(DataTestUtility.AreConnStringsSetup))]
+        public static async Task CanReadSequentialDecreasingChunks()
+        {
+            const string baseString = "ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789";
+            StringBuilder inputBuilder = new StringBuilder();
+            while (inputBuilder.Length < (64 * 1024))
+            {
+                inputBuilder.Append(baseString);
+                inputBuilder.Append(' ');
+            }
+
+            string input = inputBuilder.ToString();
+
+            StringBuilder resultBuilder = new StringBuilder();
+            CancellationTokenSource cts = new CancellationTokenSource();
+            using (var connection = new SqlConnection(DataTestUtility.TCPConnectionString))
+            {
+                await connection.OpenAsync(cts.Token);
+
+                using (var command = connection.CreateCommand())
+                {
+                    command.CommandText = "SELECT CONVERT(varchar(max),@str) as a";
+                    command.Parameters.AddWithValue("@str", input);
+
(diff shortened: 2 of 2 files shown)


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlSequentialTextReader.cs
```diff
@@ -172,7 +172,7 @@ public override Task<int> ReadAsync(char[] buffer, int index, int count)
                         byte[] byteBuffer = PrepareByteBuffer(charsNeeded, out int byteBufferUsed);
 
                         // Permit a 0 byte read in order to advance the reader to the correct column
-                        if ((byteBufferUsed < byteBuffer.Length) || (byteBuffer.Length == 0))
+                        if ((byteBufferUsed <= byteBuffer.Length) || (byteBuffer.Length == 0))
```
benrr101, line 175: nit: remove redundant parenthesis while you're here

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataReaderTest/DataReaderTest.cs
```diff
@@ -605,6 +605,56 @@ integrated into a comprehensive development
...
+            {
+                inputBuilder.Append(baseString);
+                inputBuilder.Append(' ');
+            }
+
+            string input = inputBuilder.ToString();
+
+            StringBuilder resultBuilder = new StringBuilder();
+            CancellationTokenSource cts = new CancellationTokenSource();
+            using (var connection = new SqlConnection(DataTestUtility.TCPConnectionString))
```
benrr101, line 623: More curiosity than request: for unit tests, I'm usually more liberal with usage of using statements vs blocks, so I'm wondering if your preference for the blocks is to trigger a disposal before running assertions or if behavior has no impact here and it's just a style preference.

Wraith2: Purely style. It would work with blocks the same way as with statements. I learnt c# at the first version and there are some newer idioms that I simply prefer not to use. In general I like to have very clear scoped blocks to help me visually understand what's in and out of scope.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlSequentialTextReader.cs
Wraith2: Done.

paulmedynski (changes requested):
Minor simplicity suggestion.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataReaderTest/DataReaderTest.cs
```diff
@@ -605,6 +605,56 @@ integrated into a comprehensive development
             }
         }
 
+        [ConditionalFact(typeof(DataTestUtility), nameof(DataTestUtility.AreConnStringsSetup))]
+        public static async Task CanReadSequentialDecreasingChunks()
+        {
+            const string baseString = "ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789";
```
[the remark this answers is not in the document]
Wraith2: Using repeats with specific patterns makes debugging a lot easier because you can easily identify if chunks are correctly related to each other. It's not important to the final test but it is important when the test is failing to let you see where the problem is occurring. I'd prefer to keep it for that reason.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataReaderTest/DataReaderTest.cs
benrr101: Understood - I agree, there's some newer features I don't plan on using (eg, auto-constructor assignment of member variables). Just wanted to make sure the blocks weren't integral to the behavior of the test.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataReaderTest/DataReaderTest.cs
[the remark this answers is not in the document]
benrr101: That's a decent reason I can support 👍

benrr101 (dismissed):

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataReaderTest/DataReaderTest.cs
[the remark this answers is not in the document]
paulmedynski: Yes agreed, can we get your commentary as a comment in the code?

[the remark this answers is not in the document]
Wraith2: Done.

mdaigle (approved):

PR #3388, opened 2025-06-03
Merge | DbConnectionOptions (The Rest Of It)

benrr101:
## Description
In the previous PR, we moved the debug-only code for DbConnectionOptions into its own partial class and renamed the remaining partials to follow the merge pattern (ie, *.netcore.cs and *.netfx.cs). This PR completes the process by moving the platform-specific partials into the common DbConnectionOptions.cs file. The code is wrapped in pre-processor directives to ensure they are only compiled on the appropriate platforms.

## Issues
Continues work towards #1261

## Testing
Code has moved but the functionality has not changed. Project still builds, and CI should be enough for validation.


src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/ConnectionString/DbConnectionOptions.cs
```diff
@@ -566,5 +574,287 @@ internal NameValuePair ReplacePasswordPwd(out string constr, bool fakePassword)
...
+            }
+        }
+
+        // SxS notes:
+        // * this method queries "DataDirectory" value from the current AppDomain.
+        //   This string is used for to replace "!DataDirectory!" values in the connection string, it is not considered as an "exposed resource".
+        // * This method uses GetFullPath to validate that root path is valid, the result is not exposed out.
+        [ResourceExposure(ResourceScope.None)]
+        [ResourceConsumption(ResourceScope.Machine, ResourceScope.Machine)]
+        internal static string ExpandDataDirectory(string keyword, string value, ref string datadir)
```
edwardneal: This is only ever called with a `datadir` value of `null`. With that accounted for, the only differences between this and the netcore version are some slightly more fragile and OS-specific path handling and the `ResourceExposure`/`ResourceConsumption` attributes.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/ConnectionString/DbConnectionOptions.cs
```diff
@@ -566,5 +574,287 @@ internal NameValuePair ReplacePasswordPwd(out string constr, bool fakePassword)
...
+                    }
+
+                    return (_parsetable.TryGetValue(DbConnectionStringKeywords.UserID, out value) && !string.IsNullOrEmpty(value)) || 
+                           (_parsetable.TryGetValue(DbConnectionStringSynonyms.UID, out value) && !string.IsNullOrEmpty(value));
+                }
+                return false;
+            }
+        }
+
+        internal string ExpandKeyword(string keyword, string replacementValue)
```
edwardneal, line 689: This is effectively a generalised version of netcore's `ExpandAttachDbFileName`. It's only ever called with `keyword` of AttachDBFilename or (in the event that CAS is used) Data Source and Failover Partner.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/ConnectionString/DbConnectionOptions.cs
```diff
@@ -566,5 +574,287 @@ internal NameValuePair ReplacePasswordPwd(out string constr, bool fakePassword)
...
+                    // <val'ue> -> <"val'ue">
+                    // <=value> -> <"=value">
+                    // <;value> -> <";value">
+                    // < value> -> <" value">
+                    // <va lue> -> <"va lue">
+                    // <va'"lue> -> <"va'""lue">
+                    builder.Append('\"');
+                    builder.Append(keyValue.Replace("\"", "\"\""));
+                    builder.Append('\"');
+                }
```
edwardneal: The method is only ever called with `useOdbcRules` of false. When the directly redundant code is removed, I think this is all that's left. There's also significant overlap between this functionality and `GetKeyValuePair` - have these key values already been sanitised?

This place is one of the only places which uses Regex, so there are AOT implications if it's truly already unused.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/ConnectionString/DbConnectionOptions.cs
```diff
@@ -225,6 +232,7 @@ private enum ParserState
             NullTermination,
         };
 
+        // @TODO: This should probably be extracted into its own parser class
```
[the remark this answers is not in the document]
benrr101: Go home copilot, you're drunk

mdaigle (approved):
From a pure merge perspective, it looks good to me. Interested in some of the additional improvements that @edwardneal has suggested.

paulmedynski (dismissed):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/ConnectionString/DbConnectionOptions.cs
benrr101: I can see the redundancy here ... but trying to detangle it is a bit outside the scope of this PR.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/ConnectionString/DbConnectionOptions.cs
benrr101: Sounds good - I love merges like that

mdaigle (approved):

PR #3392, opened 2025-06-04
User Story 37652: Perform Azure Authentication Split

paulmedynski:
- Setup new Extensions and Azure extension projects.
- Moved shared types over into Extensions.
- Moved Azure auth types into Azure.
- Updated netcore and netfx projects to reference Extensions.
- Updated netcore and netfx projects to build on Linux.


ErikEJ:
Maybe refer to related and most voted issue?

How will this be delivered (which NuGet packages) ?

edwardneal:
I realise this is still in draft, but PR #2884 had a couple of interesting conversations about configuring the Azure authentication provider to use WAM without dragging in the System.Windows.Forms dependencies - in particular, around the `SetIWin32WindowFunc` and the `SetParentActivityOrWindow` methods. It also introduced the use of `SynchronisationContext`, to help prevent deadlocks in the UI thread. The current PR's introducing a breaking change, so perhaps this reorganisation is the right opportunity to resuscitate that PR and cut the hard dependency on System.Windows.Forms.

paulmedynski:
This is an exploratory draft PR meant for internal consumption for the time being.  It's the easiest way to get some pretty diffs that we can use to discuss the path forward.  I'll watch here for external commentary though, and we will take everything under consideration.

paulmedynski (commented):
Marking where I have reviewed to.

mdaigle:
> I realise this is still in draft, but PR #2884 had a couple of interesting conversations about configuring the Azure authentication provider to use WAM without dragging in the System.Windows.Forms dependencies - in particular, around the `SetIWin32WindowFunc` and the `SetParentActivityOrWindow` methods. It also introduced the use of `SynchronisationContext`, to help prevent deadlocks in the UI thread. The current PR's introducing a breaking change, so perhaps this reorganisation is the right opportunity to resuscitate that PR and cut the hard dependency on System.Windows.Forms.

@edwardneal that's my intention :) I've been waiting for this work to kick off to pick up the WAM work again.

src/Microsoft.Data.SqlClient/extensions/Extensions/src/ISqlAuthenticationParameters.cs
```diff
@@ -0,0 +1,36 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.Data.SqlClient;
+
+/// <include file='../doc/ISqlAuthenticationParameters.xml' path='docs/members[@name="ISqlAuthenticationParameters"]/ISqlAuthenticationParameters/*'/>
+public interface ISqlAuthenticationParameters
```
saurabh500: @paulmedynski You may want to reconsider the usage of an interface for a public API. 
Interfaces cannot be evolved with new members. 
If a new method/property is added to interface, then older implementations become incompatible. With base class and virtual methods, the evolution of public API is possible. 

I thought I will leave this thought here, even though I know this PR is WIP.

Frulfump:
PR Closed without comment is that bad news for the split?

paulmedynski:
This was a draft PR for internal consumption.  I am splitting the work up in to smaller chunks.  You will see more PRs as this work becomes ready.

christothes:
@paulmedynski 
This might be a good opportunity to decouple Azure.Identity from your auth story. We initially explored such options with the [AccessTokenCallback](https://learn.microsoft.com/en-us/dotnet/api/microsoft.data.sqlclient.sqlconnection.accesstokencallback?view=sqlclient-dotnet-core-6.0). I realize one of the challenges is how to allow something like the `DefaultAzureCredential` functionality via connection string alone, but perhaps there are other approaches we could consider.

PR #3399, opened 2025-06-06
Use UTF8 instance that doesn't emit BOM

### LINKED ISSUE #3397: SqlBulkCopy streaming ends up prefixing BOM to UTF8 source column with VARCHAR(MAX) type
### Describe the bug

Scenario: SqlBulkCopy

Source SQL Server >= 2019 with UTF-8 support.
Target  SQL Server >= 2019 with UTF-8 support.

Source Schema: UTF-8 Encoded Varchar(max) column with one of the UTF-8 Collation. 
Target Schema: Same as source

When the source table is copied to target table with Streaming turned on SqlBulkCopy, then a BOM is prefixed to the data, which causes the data in the target to change. This causes unexpected side effects while working with data since the original data has lost its fidelity. 

Scope of the problem: UTF8 encoded source table with VARCHAR(MAX) column, copied to target SQL server using Streaming.

### To reproduce
Include a complete code listing (or project/solution) that we can run to reproduce the issue.

Partial code listings, or multiple fragments of code, will slow down our response or cause us to push the issue back to you to provide code to reproduce the issue.

```sql
drop table if exists utf8_data, utf8_data_bak;
create table utf8_data (
 str_col varchar(max) collate Latin1_General_100_CS_AS_KS_WS_SC_UTF8
);
create table utf8_data_bak (
 str_col varchar(max) collate Latin1_General_100_CS_AS_KS_WS_SC_UTF8
);
insert into utf8_data select top(10) name from sys.tables;
```
```c#
    {

        String sourceConnectionString = @"Server=172.20.107.10;Database=ScrapDb;user id=<>;encrypt=false";
        String targetConnectionString = @"Server=172.20.107.10;Database=ScrapDb;user id=<sa>;encrypt=false";

        SqlConnection sqlsourceConnection = new SqlConnection(sourceConnectionString);
        SqlConnection sqlTargetConnection = new SqlConnection(targetConnectionString);

        sqlsourceConnection.Open(); 
        sqlTargetConnection.Open();

        SqlCommand sqc = sqlsourceConnection.CreateCommand();
        sqc.CommandText = $"SELECT str_col FROM utf8_data;";
        const int bulktimeout = 3600;

        SqlCommand sqd = sqlsourceConnection.CreateCommand();
        sqd.CommandText = $"TRUNCATE TABLE utf8_data_bak;";
        sqd.ExecuteNonQuery();


        SqlBulkCopy sbc = new SqlBulkCopy(sqlTargetConnection);
        sbc.DestinationTableName = $"utf8_data_bak";
        sbc.BulkCopyTimeout = bulktimeout;
        sbc.BatchSize = 1000;
        sbc.EnableStreaming = true;
        long crows = 0;
        int ccols = 0;

        try
        {
            using (SqlDataReader sdr = sqc.ExecuteReader())
            {
                ccols = sdr.FieldCount;
                sbc.WriteToServer(sdr);
                //crows += sbc.RowsCopied64;
            }
            sbc.Close();

            Console.WriteLine("Columns count: {0}, Rows Inserted: {0}", ccols, crows);

            Console.WriteLine("Source Table: {0}", "utf8_data");
            sqc.CommandText = $"SELECT str_col, datalength(str_col) FROM utf8_data;";
            using (SqlDataReader sdr = sqc.ExecuteReader())
            {
                Console.WriteLine("str_col type: {0}", sdr.GetDataTypeName(0));
                Console.WriteLine();
                while (sdr.Read())
                {
                    Console.WriteLine("{0},{1}", sdr.GetSqlString(0), sdr.GetSqlInt64(1));
                }
            }

            Console.WriteLine();
            Console.WriteLine("Destination Table: {0}", "utf8_data_bak");
            sqd.CommandText = $"SELECT str_col, datalength(str_col) FROM utf8_data_bak;";
            using (SqlDataReader sdr = sqd.ExecuteReader())
            {
                Console.WriteLine("str_col type: {0}", sdr.GetDataTypeName(0));
                Console.WriteLine();
                while (sdr.Read())
                {
                    Console.WriteLine("{0},{1}", sdr.GetSqlString(0), sdr.GetSqlInt64(1));
                }
            }
        }
        catch (Exception ex)
        {
            Console.WriteLine(ex.Message);
        }
        finally
        {
            sqlsourceConnection.Close();
            sqlTargetConnection.Close();
        }
    }
}
```

The above code copies data from the source to the target table, however when we look at the bytes of the data, a UTF-8 BOM prefix is added. 

To validate, run the following query on the target table
```sql
select cast(str_col as varbinary) from utf8_data_bak;
```

The output is 
```
0xEFBBBF757466385F64617461
0xEFBBBF757466385F646174615F62616B
```

The first 3 bytes `EFBBBF` is unexpected.

This shouldn't be prefixed to the data. 

I tested this with MDS 1.0 to the latest and the source of this problem is the inclusion of UTF8 support

### Expected behavior
no BOM should be pre-fixed in the target table's data.

### Further technical details
Microsoft.Data.SqlClient version: Tested on 1.0 - 6.x. Exists on all. 
.NET target: .NET Framework 4.6.2, .NET 8.0
SQL Server version:  SQL Server 2022, SQL Server 2018
Operating system: Client on Win 11

**Additional context**

This is likely happening because UTF8 encoding by default adds a BOM. 
The data is sent to SQL server using `StringWriter` with the encoding of the column received from the server. The server provides UTF8 encoding, which is translated to C# UTF8. The same encoding is used to transmit the data to SQL server. 

TdsParser.cs
```
using (ConstrainedTextWriter writer = new ConstrainedTextWriter(new StreamWriter(new TdsOutputStream(this, stateObj, null), encoding), size))
```

A likely fix could be to check if the encoding is UTF8 and to reinitialize the encoding and indicate that no bom is emitted

```
if (encoding == UTF8Encoding.UTF8)
{
    encoding = new UTF8Encoding(encoderShouldEmitUTF8Identifier: false); // no BOM
}
```

apoorvdeshmukh:
## Description
Fixes #3397.  Used UTF8Encoding(false) instance that doesn't emit BOM instead of
using System.Encoding.UTF8 which adds BOM while transmitting the data.
Added test case based on the available repro which fails without the fix.

## Issues
#3397 

## Testing
Added sync and async SqlBulkCopy testcase that does bulkcopy from UTF8 source table to another table with UTF8 collation
with MARS = true/false, EnableStraming = true/false.
Without the fix, when the destination table is read after bulkcopy, contains BOM in the data.


src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/SqlBulkCopyTest/TestBulkCopyWithUTF8.cs
```diff
@@ -0,0 +1,82 @@
...
+            Assert.Equal(1, finalCount - initialCount);
+
+            using SqlCommand verifyCommand = new SqlCommand($"SELECT str_col FROM {destinationTable}", destinationConnection);
+            using SqlDataReader verifyReader = verifyCommand.ExecuteReader(CommandBehavior.SequentialAccess);
+
+            byte[] expectedBytes = Encoding.UTF8.GetBytes("test");
+
+            Assert.True(verifyReader.Read(), "No data found in destination table after bulk copy.");
+
+            byte[] actualBytes = Encoding.UTF8.GetBytes(verifyReader.GetString(0));
```
saurabh500: Instead of getting the col as  string, can you please get the str_col as Byte array by casting it to a binary in the query? 
The intention is to avoid going through the string conversions/processing of .Net and look at the binary that the server sees. 
This would be a better way of testing, since this is how the bug was found in the first place. 

Essentially the query in SqlCommand would change to `select cast(str_col as varbinary) from {destinationTable};` 

Get the bytes array using GetObject on the reader and then do the binary comparison. 

Let me know if you think otherwise.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -178,6 +178,8 @@ internal sealed partial class TdsParser
         internal bool isTcpProtocol { get; set; }
         internal string FQDNforDNSCache { get; set; }
 
+        private static readonly Encoding s_utf8EncodingWithoutBom = new UTF8Encoding(encoderShouldEmitUTF8Identifier: false);
```
saurabh500: Isn't there a common TdsParser.cs where this static can be placed?

saurabh500 (commented):
Approach looks good. 
Left some comments on the test approach

edwardneal:
It might be a good idea to make sure that we have test coverage for scenarios where v6.1 tries to read UTF8 data which starts with a BOM, to handle cases where v6.1+ has to read data which was copied by v6.0...

saurabh500:
@edwardneal   The scenario you're describing is not something we intend to support. Adding a BOM (Byte Order Mark) can make the data unusable in many cases.

We're operating under the assumption that no one has relied on this buggy behavior for bulk copy operations. As such, ensuring interoperability between the old and corrected behavior is not within the scope of this fix.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/SqlBulkCopyTest/TestBulkCopyWithUTF8.cs
apoorvdeshmukh: That is what my original thought was too. Just out of curiosity, I decided to check if the retrieved string contained BOM when data was retrieved with `GetString()` and then converted it into bytes and I could see BOM bytes being present in the byte[] array. 
However, I agree, converting it to `varbinary` leaves no room for any confusion.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
apoorvdeshmukh: Yes. Moved it to common TdsParser.cs

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/SqlBulkCopyTest/TestBulkCopyWithUTF8.cs
```diff
@@ -0,0 +1,82 @@
...
+                Assert.Fail($"Bulk copy failed: {ex.Message}");
+            }
+
+            long finalCount = Convert.ToInt64(countCommand.ExecuteScalar());
+            Assert.Equal(1, finalCount - initialCount);
+
+            using SqlCommand verifyCommand = new SqlCommand($"SELECT cast(str_col as varbinary) FROM {destinationTable}", destinationConnection);
+            using SqlDataReader verifyReader = verifyCommand.ExecuteReader(CommandBehavior.SequentialAccess);
+
+            byte[] expectedBytes = Encoding.UTF8.GetBytes("test");
```
paulmedynski: Let's verify that this is [0x74, 0x65, 0x73, 0x74] to ensure we don't have a BOM later on line 79.  In fact, I would suggest hardcoding the byte array to avoid any UTF8 encoder perturbing the test.

apoorvdeshmukh: Makes sense. Hardcoded the expected bytes.

paulmedynski (dismissed):

saurabh500 (approved):

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/SqlBulkCopyTest/TestBulkCopyWithUTF8.cs
```diff
@@ -0,0 +1,82 @@
...
+
+            using SqlBulkCopy bulkCopy = new SqlBulkCopy(destinationConnection)
+            {
+                EnableStreaming = true,
+                DestinationTableName = destinationTable
+            };
+
+            try
+            {
+                bulkCopy.WriteToServer(reader);
```
cheenamalhotra, line 110: Can you add a test for WriteToServerAsync as well?

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/SqlBulkCopyTest/TestBulkCopyWithUTF8.cs
```diff
@@ -0,0 +1,82 @@
...
+        [ConditionalFact(typeof(DataTestUtility), nameof(DataTestUtility.AreConnStringsSetup), nameof(DataTestUtility.IsNotAzureServer), nameof(DataTestUtility.IsNotAzureSynapse))]
+        public void BulkCopy_Utf8Data_ShouldMatchSource()
+        {
+            string sourceTable = DataTestUtility.GetUniqueName("SrcUtf8DataTable");
+            string destinationTable = DataTestUtility.GetUniqueName("DstUtf8DataTable");
+            string insertQuery = $"INSERT INTO {sourceTable} VALUES('test')";
+
+            using SqlConnection sourceConnection = new SqlConnection(_connectionString);
+            sourceConnection.Open();
+            SetupTables(sourceConnection, sourceTable, destinationTable, insertQuery);
```
cheenamalhotra: 1. Delete table as well.
2. Instead of setting and destroying tables in test, use Xunit helper methods that run before and after the tests.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/SqlBulkCopyTest/TestBulkCopyWithUTF8.cs
```diff
@@ -0,0 +1,82 @@
...
+
+namespace Microsoft.Data.SqlClient.ManualTesting.Tests
+{
+    public class TestBulkCopyWithUtf8
+    {
+        private readonly string _connectionString;
+
+        public TestBulkCopyWithUtf8()
+        {
+            _connectionString = new SqlConnectionStringBuilder(DataTestUtility.TCPConnectionString){MultipleActiveResultSets = true}.ConnectionString;
```
cheenamalhotra: Run the test for MARS=true/false both cases.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/SqlBulkCopyTest/TestBulkCopyWithUTF8.cs
apoorvdeshmukh: Added the async version of the testcase

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/SqlBulkCopyTest/TestBulkCopyWithUTF8.cs
apoorvdeshmukh: Added the per-test setup and teardown functions.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/SqlBulkCopyTest/TestBulkCopyWithUTF8.cs
apoorvdeshmukh: Parameterized the test for MARS and SqlBulkCopy streaming behavior.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/SqlBulkCopyTest/TestBulkCopyWithUTF8.cs
```diff
@@ -4,54 +4,64 @@
...
-            string sourceTable = DataTestUtility.GetUniqueName("SrcUtf8DataTable");
-            string destinationTable = DataTestUtility.GetUniqueName("DstUtf8DataTable");
-            string insertQuery = $"INSERT INTO {sourceTable} VALUES('test')";
-
-            using SqlConnection sourceConnection = new SqlConnection(_connectionString);
+            using SqlConnection sourceConnection = new SqlConnection(GetConnectionString(isMarsEnabled));
             sourceConnection.Open();
-            SetupTables(sourceConnection, sourceTable, destinationTable, insertQuery);
-
             using SqlCommand countCommand = new SqlCommand($"SELECT COUNT(*) FROM {destinationTable}", sourceConnection);
```
paulmedynski: Should this be using destinationConnection (for consistency)?  Some comments throughout these tests explaining the intention would help.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/SqlBulkCopyTest/TestBulkCopyWithUTF8.cs
```diff
@@ -63,20 +73,74 @@ public void BulkCopy_Utf8Data_ShouldMatchSource()
             {
                 Assert.Fail($"Bulk copy failed: {ex.Message}");
             }
-
+            
+            reader.Close();
             long finalCount = Convert.ToInt64(countCommand.ExecuteScalar());
             Assert.Equal(1, finalCount - initialCount);
```
paulmedynski: Would it be better to `Assert.Equal(0, initialCount)` above?  And then `Assert.Equal(1, finalCount)` here?  The destination  table is supposed to be new and empty as an invariant of the test setup.  You wouldn't need the initialCount and finalCount variables then either:

```
// Verify that our destination table is empty before bulk copying.
Assert.Equal(0, Convert.ToInt64(countCommand.ExecuteScalar()));
```

```
// Verify that the 1 row from the source table has been copied into our destination table.
Assert.Equal(1, Convert.ToInt64(countCommand.ExecuteScalar()));
```

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/SqlBulkCopyTest/TestBulkCopyWithUTF8.cs
```diff
@@ -63,20 +73,74 @@ public void BulkCopy_Utf8Data_ShouldMatchSource()
...
-
             using SqlCommand verifyCommand = new SqlCommand($"SELECT cast(str_col as varbinary) FROM {destinationTable}", destinationConnection);
             using SqlDataReader verifyReader = verifyCommand.ExecuteReader(CommandBehavior.SequentialAccess);
-
             byte[] expectedBytes = new byte[] { 0x74, 0x65, 0x73, 0x74 };
```
paulmedynski: A comment here to explain this would be helpful.

`// This is 'test' in UTF-8 bytes.`

And this could be a class constant, up beside insertQuery for context.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/SqlBulkCopyTest/TestBulkCopyWithUTF8.cs
```diff
@@ -4,54 +4,64 @@
...
+using System.Threading.Tasks;
 using Xunit;
 
 namespace Microsoft.Data.SqlClient.ManualTesting.Tests
 {
-    public class TestBulkCopyWithUtf8
+    public sealed class TestBulkCopyWithUtf8 : IDisposable
     {
-        private readonly string _connectionString;
+        private static string sourceTable = DataTestUtility.GetUniqueName("SrcUtf8DataTable");
```
paulmedynski: We use `s_` prefix for private static fields.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/SqlBulkCopyTest/TestBulkCopyWithUTF8.cs
```diff
@@ -63,20 +73,74 @@ public void BulkCopy_Utf8Data_ShouldMatchSource()
...
+            using SqlCommand verifyCommand = new SqlCommand($"SELECT cast(str_col as varbinary) FROM {destinationTable}", destinationConnection);
+            using SqlDataReader verifyReader = await verifyCommand.ExecuteReaderAsync(CommandBehavior.SequentialAccess);
+            byte[] expectedBytes = new byte[] { 0x74, 0x65, 0x73, 0x74 };
+            Assert.True(await verifyReader.ReadAsync(), "No data found in destination table after bulk copy.");
             byte[] actualBytes = verifyReader.GetSqlBinary(0).Value;
             Assert.Equal(expectedBytes.Length, actualBytes.Length);
             Assert.Equal(expectedBytes, actualBytes);
         }
+
+        public void Dispose()
```
paulmedynski: I typically put Disopse() beside the constructors, so it's easy to see resource setup/cleanup together.

paulmedynski (changes requested):
Some changes requested.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/SqlBulkCopyTest/TestBulkCopyWithUTF8.cs
```diff
@@ -50,43 +82,60 @@ private void SetupTables(SqlConnection connection, string sourceTable, string de
         [InlineData(false, false)]
         public void BulkCopy_Utf8Data_ShouldMatchSource(bool isMarsEnabled, bool enableStreaming)
         {
-            using SqlConnection sourceConnection = new SqlConnection(GetConnectionString(isMarsEnabled));
+            // Setup conneections for source and destination tables
```
paulmedynski: connections

paulmedynski (approved):

Issue #1947, opened 2023-03-07
NativeAOT support: get to zero warnings


roji:
Right now, when publishing a NativeAOT app using SqlClient, many warnings are emitted. To properly support NativeAOT, SqlClient will need to get to a place where no warnings are emitted, so the user can be 100% confident that the application will function correctly.

Note that this is related, but not the same as #1942, which is about reducing the binary size. However, the same technique can be used for both: if an entry point (e.g. slim data source builder) is introduced, which by default doesn't include functionality that is large and/or unsafe for NativeAOT, then programs not explicitly bringing in that functionality won't get warnings etc.

kant2002:
I add project settings to be enable trimming analysis here https://github.com/dotnet/SqlClient/pull/1943 , That allow bookkeeping  warnings which should be fixed.

steveberdy:
@roji @jkotas I'd like to take a stab at this if you could assign it to me.

roji:
@steveberdy this is SqlClient - neither @jkotas nor myself are owners/maintainers in this repo. I highly suggested first discussing the plan around this with the SqlClient owners and proceeding step-by-step.

steveberdy:
Thanks, sorry about that. @cheenamalhotra are you a code owner? It looks like you have approved and merged PRs. If so, it'd be great if we could have a discussion on implementing AOT compatibility

cheenamalhotra:
Hi @steveberdy 

Yes, I am one of the engineers on the team. I noticed your https://github.com/dotnet/SqlClient/issues/2742#issuecomment-2522196270 as well, did you want to propose the changes to Active Directory Auth provider as you mentioned here? Let us know your thoughts, and we're here to review and support.

steveberdy:
@cheenamalhotra I'd be happy too! I'm working on the changes, and I'm setting up AOT-compatibility, trim warnings, etc. I'll make a PR. If there are any trim or AOT issues, I'd be happy to check them out as well.

charlesroddie:
I put in two PRs to tackle some easy cases and whittle this down.

- https://github.com/dotnet/SqlClient/pull/4683
- https://github.com/dotnet/SqlClient/pull/4684

A harder case is the use of DiagnosticListener. A switch to turn that off should work. I'll look at that next.

charlesroddie:
I added 3 more PRs and they are all ready for review.

The only remaining trim warnings are:
[...]

PR #3401, opened 2025-06-06
Enable IL trimming and NativeAOT on AzureKeyVaultProvider

edwardneal:
## Description

We can safely enable IL trimming and NativeAOT support on the AzureKeyVaultProvider project for the .NET targets. This PR simply sets the `IsTrimmable` and `IsAotCompatible` properties.

This doesn't enable these features on SqlClient itself, and the addon isn't particularly useful on its own - but when the library itself is compatible, the addon isn't going to hold it back.

## Issues

Relates to #1947.

## Testing

The IL trimming and NativeAOT support is mostly flagged by compile-time warnings. A sample application which directly references the DLL runs successfully - I can successfully publish and run the code below.

```csharp
const string AkvPath = "...";
var keyVaultProvider = new SqlColumnEncryptionAzureKeyVaultProvider(new DefaultAzureCredential(true));
var signedCMKM = keyVaultProvider.SignColumnMasterKeyMetadata(AkvPath, false);
var verifiedMetadata = keyVaultProvider.VerifyColumnMasterKeyMetadata(AkvPath, false, signedCMKM);

if (!verifiedMetadata)
    throw new Exception("Signed CMK metadata does not match.");

var cek = new byte[32];
var encryptedCek = keyVaultProvider.EncryptColumnEncryptionKey(AkvPath, "RSA-OAEP", cek);
var decryptedCek = keyVaultProvider.DecryptColumnEncryptionKey(AkvPath, "RSA-OAEP", encryptedCek);

if (!decryptedCek.SequenceEqual(cek))
    throw new Exception("Column encryption key does not roundtrip.");
```

I've not added a unit test - as per dotnet/runtime#97013 this isn't a good way to perform testing.

### DIFF
--- a/src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider.csproj
+++ b/src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider.csproj
@@ -15,6 +15,8 @@
     <BuildProjectReferences>false</BuildProjectReferences>
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
     <PackageRequireLicenseAcceptance>true</PackageRequireLicenseAcceptance>
+    <IsTrimmable Condition="'$(TargetGroup)'=='netcoreapp'">true</IsTrimmable>
+    <IsAotCompatible Condition="'$(TargetGroup)'=='netcoreapp'">true</IsAotCompatible>
   </PropertyGroup>
    <!--Generating Strong Name-->
   <PropertyGroup Condition="'$(CDP_BUILD_TYPE)'=='Official'">


src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider.csproj
```diff
@@ -15,6 +15,8 @@
     <BuildProjectReferences>false</BuildProjectReferences>
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
     <PackageRequireLicenseAcceptance>true</PackageRequireLicenseAcceptance>
+    <IsTrimmable Condition="'$(TargetGroup)'=='netcoreapp'">true</IsTrimmable>
```
edwardneal, line 18: Technically this is implied by setting `IsAotCompatible`. I think having both properties is slightly clearer until we have end-to-end support for trimming, but that's purely a matter of taste.

paulmedynski (approved):

benrr101 (approved):
image
LGTM! (that stands for "Let's get this merged", right?)

PR #3402, opened 2025-06-07
Tests | Move fixtures to common project

edwardneal:
## Description

This PR performs some simple cleanup of the tests. #3371 created a new Common project which contains common code for test projects. We add this project to the Visual Studio solution, and move the encryption-based fixtures to it. Indentation changes make up the vast majority of this PR...

I've also rolled one changes which caused headaches in #3204 by allowing Visual Studio to run tests against the net8.0 target. This caused some pain when working with certificates - the recommended API changed between net8.0 and net9.0.

There was also one piece of feedback on #3204 about preferring string interpolation. This was fixed in one place, but there was another instance of it. I've made the spot-fix.

## Issues

Relates to #3371 and #3204.

## Testing

All automated tests continue to pass, on all three targets.

### DIFF
--- a/src/Microsoft.Data.SqlClient/tests/Common/Common.csproj
+++ b/src/Microsoft.Data.SqlClient/tests/Common/Common.csproj
@@ -53,8 +53,8 @@
 
   <!-- Fixture-specific References -->
   <ItemGroup>
-    <PackageReference Include="Azure.Identity" Version="$(AzureIdentityVersion)" />
-    <PackageReference Include="Azure.Security.KeyVault.Keys" Version="$(AzureSecurityKeyVaultKeysVersion)" />
+    <PackageReference Include="Azure.Identity" />
+    <PackageReference Include="Azure.Security.KeyVault.Keys" />
   </ItemGroup>
 
 </Project>


paulmedynski:
Hey @edwardneal - Thanks for starting to clean up some of these test helpers.

My goal with the Common test project was for it to contain things that are shared between other test projects.  I see that some of the fixtures you moved are only used in a single test project, so maybe they're better off moving into those projects?

- AzureKeyVaultKeyFixtureBase - used only in ManualTests.
- CertificateFixtureBase - used in FunctionalTests and ManualTests, so it belongs in Common as you have moved it.
- ColumnEncryptionCertificateFixture - used only in FunctionalTests.
- ColumnMasterKeyCertificateFixture - used only in ManualTests.
- CspCertificateFixture - used only in ManualTests.

I see that ManualTests already has a TestFixtures sub-directory.  Maybe FunctionalTests needs the same (preferrably both named just Fixtures)?

I will discuss with the team and report back before asking for any changes here.

edwardneal:
Thanks @paulmedynski. The tests are in an odd middle ground at the moment - I've moved the fixtures from one "common-ish" project to the purpose-built Common project, but introducing a convention of a project-specific `Fixtures` namespace would work just as well for our purposes.

It highlights some slightly odd test placement: I'd have expected some functional tests for `CspCertificateFixture` and `AzureKeyVaultKeyFixtureBase` to make sure they can round-trip values. It might be a good idea to keep these two fixtures in Common, ready for some future test restructuring.

paulmedynski:
I discussed with the team, and we're ok with these moves.  We can make further changes once the project merging is complete.

src/Microsoft.Data.SqlClient/tests/Common/Common.csproj
```diff
@@ -51,4 +51,10 @@
     </ContentWithTargetPath>
   </ItemGroup>
 
+  <!-- Fixture-specific References -->
+  <ItemGroup>
+    <PackageReference Include="Azure.Identity" Version="$(AzureIdentityVersion)" />
```
cheenamalhotra: Use central package management here.

src/Microsoft.Data.SqlClient/tests/Directory.Build.props
```diff
@@ -22,7 +22,7 @@
   <!--These properties can be modified locally to target .NET version of choice to build and test entire test suite-->
   <PropertyGroup>
     <TargetNetFxVersion Condition="'$(TargetNetFxVersion)' == ''">net462</TargetNetFxVersion>
-    <TargetNetCoreVersion Condition="'$(TargetNetCoreVersion)' == ''">net9.0</TargetNetCoreVersion>
+    <TargetNetCoreVersion Condition="'$(TargetNetCoreVersion)' == ''">net8.0;net9.0</TargetNetCoreVersion>
```
cheenamalhotra, line 25: Use only net8.0 instead of using both targets, because its supposed to feed 'TargetFramework' property.

edwardneal: It does, and CI will set TargetNetCoreVersion/TargetNetFxVersion appropriately to run tests against net462, net8.0 and net9.0. Visual Studio doesn't set TargetNetCoreVersion though, so the test projects are only currently built against (and only test) net462 and net9.0.

This caused a few problems when generating certificates in the test project - the recommended usage of X509Certificate2 changed between net8.0 and net9.0, so we had a few build issues which appeared in CI but not in VS.

src/Microsoft.Data.SqlClient/tests/Common/Common.csproj
cheenamalhotra: @edwardneal please address these soon, so we can get this merged. Thanks!

edwardneal: Thanks @cheenamalhotra. I've just fixed this part of the review - I'm not sure what the best approach would be around the `TargetNetCoreVersion` point, so have left it alone.

cheenamalhotra (approved):

PR #3405, opened 2025-06-09
Merge | SqlBulkCopy

benrr101:
## Description
This PR merges the SqlBulkCopy from both netfx and netcore projects into a single instance in the common project. This was surprisingly easy, as the vast majority of the two versions were identical. I'll add callouts in the diff for the few differences that needed to be addressed, though in most cases, the netcore versions were taken since they were slightly more performant than the netfx version.

There is also a lot of stubs added to common projects. This was just to make it easier to verify that the file is error-free when working in the common project. If necessary, I can drop these from the PR.

## Issues
More work for #1261

## Testing
Code is mostly just moved, so CI should validate the behavior.


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlUtil.cs
```diff
@@ -416,26 +416,30 @@ internal static void SetTimeoutException(TaskCompletionSource<object> completion
                 );
             }
         }
-
-#if NET
-        internal static void SetTimeoutExceptionWithState(TaskCompletionSource<object> completion, int timeout, object state, Func<object, Exception> onFailure, CancellationToken cancellationToken)
+        
+        internal static void SetTimeoutExceptionWithState(
```
benrr101, line 420: By adding this project to the netfx project, we can use the slightly more performant "with state" version of this method.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlUtil.cs
```diff
@@ -416,26 +416,30 @@ internal static void SetTimeoutException(TaskCompletionSource<object> completion
...
+            TaskCompletionSource<object> completion,
+            int timeout,
+            object state,
+            Func<object, Exception> onFailure,
+            CancellationToken cancellationToken)
         {
-            if (timeout > 0)
+            if (timeout <= 0)
```
benrr101, line 427: Flipped the condition to reduce nesting

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
```diff
@@ -492,8 +490,8 @@ private string CreateInitialQuery()
         private Task<BulkCopySimpleResultSet> CreateAndExecuteInitialQueryAsync(out BulkCopySimpleResultSet result)
         {
             string TDSCommand = CreateInitialQuery();
-            SqlClientEventSource.Log.TryTraceEvent("<sc.SqlBulkCopy.CreateAndExecuteInitialQueryAsync|INFO> Initial Query: '{0}'", TDSCommand);
-            SqlClientEventSource.Log.TryCorrelationTraceEvent("<sc.SqlBulkCopy.CreateAndExecuteInitialQueryAsync|Info|Correlation> ObjectID {0}, ActivityID {1}", ObjectID, ActivityCorrelator.Current);
+            SqlClientEventSource.Log.TryTraceEvent("SqlBulkCopy.CreateAndExecuteInitialQueryAsync | Info | Initial Query: '{0}'", TDSCommand);
```
benrr101, line 493: Took the netcore version of the traces as per discussion a while ago.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
```diff
@@ -554,7 +552,11 @@ private string AnalyzeTargetAndCreateUpdateBulkCommand(BulkCopySimpleResultSet i
 
             // Keep track of any result columns that we don't have a local
             // mapping for.
+            #if NETFRAMEWORK
             HashSet<string> unmatchedColumns = new();
+            #else
+            HashSet<string> unmatchedColumns = new(_localColumnMappings.Count);
```
benrr101, line 558: Believe it or not, netfx doesn't support a constructor with specifying the size.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
```diff
@@ -1202,6 +1206,19 @@ private bool ReadFromRowSource()
 
         private SourceColumnMetadata GetColumnMetadata(int ordinal)
         {
+            bool IsMetadataDataStream(_SqlMetaData metadata)
```
benrr101, line 1209: I opted to rewrite this using a local function to make switching in the netfx-only check cleaner

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
```diff
@@ -1375,7 +1386,11 @@ private void RunParserReliably(BulkCopySimpleResultSet bulkCopyHandler = null)
             internalConnection.ThreadHasParserLockForClose = true;
             try
             {
-                _parser.RunReliably(RunBehavior.UntilDone, null, null, bulkCopyHandler, _stateObj);
+                #if NETFRAMEWORK
```
benrr101, line 1389: I'm not 100% sure the implications of using one versus the other.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
```diff
@@ -1844,25 +1875,18 @@ public Task WriteToServerAsync(DataRow[] rows, CancellationToken cancellationTok
...
-                    {
-                        source.SetCanceled();
-                    }
-                    else
-                    {
-                        source.SetResult(null);
-                    }
-                    resultTask = source.Task;
-                    return resultTask; // nothing to do. user passed us an empty array. Return a completed Task.
+                    return cancellationToken.IsCancellationRequested
```
benrr101, line 1887: Took the netcore implementation that is way cleaner

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
```diff
@@ -2043,20 +2078,24 @@ private Task WriteRowSourceToServerAsync(int columnCount, CancellationToken ctok
 
             bool finishedSynchronously = true;
             _isBulkCopyingInProgress = true;
+            
+            CreateOrValidateConnection(nameof(WriteToServer));
```
benrr101, line 2080: Switching to `nameof` allows us to remove the SQL.WriteToServer constant.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
```diff
@@ -3002,14 +3049,27 @@ private void WriteToServerInternalRestAsync(CancellationToken cts, TaskCompletio
                         TaskCompletionSource<object> cancellableReconnectTS = new TaskCompletionSource<object>();
                         if (cts.CanBeCanceled)
                         {
-                            regReconnectCancel.Value = cts.Register(() => cancellableReconnectTS.TrySetCanceled());
+                            regReconnectCancel.Value = cts.Register(
```
benrr101, line 3050: These methods, taken from netcore implementation, are being used since they use static lambdas and are slightly more performant.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
edwardneal: It's purely CER-related, if the method existed in netcore in its current form then there'd be no side-effects to calling it on both targets.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
```diff
@@ -1987,30 +2018,34 @@ public Task WriteToServerAsync(DataTable table, DataRowState rowState, Cancellat
...
             finally
             {
                 SqlStatistics.StopTimer(statistics);
             }
-            return resultTask;
         }
 
         private Task WriteRowSourceToServerAsync(int columnCount, CancellationToken ctoken)
         {
+            #if NETFRAMEWORK
```
edwardneal: This behaviour is a little odd - it's never been in netcore (even back to S.D.S's first checkin in 2015.) The logic behind it is sound though. I'd suggest enabling it in netcore, but have the method return `Task.FromCanceled(ctoken)` instead.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
```diff
@@ -198,6 +194,10 @@ public SourceColumnMetadata(ValueMethod method, bool isSqlType, bool isDataFeed)
         private DataRowState _rowStateToSkip;
         private IEnumerator _rowEnumerator;
 
+        #if NETFRAMEWORK
+        private bool _rowSourceIsSqlDataReaderSmi;
+        #endif
```
edwardneal, line 199: SqlDataReaderSmi can only be instantiated from context connections, so this will always be false on netfx. That's got knock-on implications throughout the file.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
```diff
@@ -1269,7 +1285,7 @@ private SourceColumnMetadata GetColumnMetadata(int ordinal)
                         method = ValueMethod.DataFeedStream;
                     }
                     // For text and XML there is memory gain from streaming on destination side even if reader is non-sequential
-                    else if (((metadata.type == SqlDbType.VarChar) || (metadata.type == SqlDbType.NVarChar) || (metadata.type == SqlDbTypeExtensions.Json)) && (mtSource.IsCharType) && (mtSource.SqlDbType != SqlDbType.Xml))
+                    else if (metadata.type is SqlDbType.VarChar or SqlDbType.NVarChar or SqlDbTypeExtensions.Json && mtSource.IsCharType && mtSource.SqlDbType != SqlDbType.Xml)
```
mdaigle: did parentheses removal change this check? `(x || y || z) && a && b` became `x || y || z && a && b`

mdaigle (changes requested):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
paulmedynski: I think this syntax is described here:

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/patterns#logical-patterns

The `or`s aren't synonyms for ||.  They appear to bind to the `is` operator, so the entire first part `x is A or B or C` is one operator, which is then logical-anded `&&` with the next operand as expected.

It's a bit esoteric, so extra parentheses wouldn't hurt, even if the IDE says they are unnecessary.

paulmedynski (approved):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
benrr101: @paulmedynski is correct - I don't think it's particularly esoteric, but I can add parens to make it a bit clearer

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
benrr101: What's even better is that's the only usage of ADP.CreatedTaskWithCancellation so I can just delete that, too

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
benrr101: You are right, and it's being tackled as part of part 2 of SMI dead code deletion 🚮

paulmedynski (approved):

PR #3414, opened 2025-06-12
Cleanup | Remove More Dead SMI Code (Part 2)

benrr101:
## Description
While working on another chunk of code, I had another breakthrough on provably deleting SMI code. There's a tiny class `InOutOfProcHelper` that contains a single property `InProc` that always returns `false`. Pulling on this thread revealed that the entirety of SmiContextFactory and SmiContext could be deleted.

The process I used for safely deleting these was to basically take the hard coded values and propagate them down to lower levels. At each step of propagating, I delete branches of the code became unreachable, and keep pushing it further down. Each commit builds on its own, so it's possible to step through the changes. Trying to explain each step would probably be impossible...

Removing SmiContext from ValueUtilsSmi became a bit tricky since there are circular references that pass an SmiContext between each other. However, knowing that SmiContext should always be null, I was able to work it out by passing null into the calls to the methods, then making the argument have a null default, then eliminating nulls from the calls to the methods, then removing the "internal" usages of the argument since logically they could only be null.

There are a couple annotated spots where the code must've never been called because it will literally always throw.

## Testing
These branches should not have been reached previously, so CI should be sufficient validate nothing broke.


edwardneal:
There are a few types which are only ever used from these code paths too.
* SqlStreamChars
* SqlClientWrapperSmiStream
* SqlClientWrapperSmiStreamChars

Besides this, SqlSequentialStreamSmi and SqlSequentialTextReaderSmi will also disappear shortly, once SqlDataReaderSmi is removed.

benrr101:
@edwardneal Very good catch!! Let's get that deleted code counter over 1000!

mdaigle (approved):

PR #3417, opened 2025-06-12
Fix | PoolGroupProviderInfo Typing

benrr101:
## Description
This PR does two main changes:
1) Reverts a small change in netfx SqlConnectionFactory class from #3393.
    I thought it was a safe change to make, but I neglected to notice that the poolgroupproviderinfo was being passed in from the caller of the method. I noticed this issue while working on merging SqlConnectionFactory, since I made this change on the netfx version but not on the netcore version.
2) Makes SqlConnectionFactory CreateConnection classes have strongly typed poolgroupproviderinfo arguments.
    Not sure why these were always being passed in as object, but they are now being passed in as DbConnectionPoolGroupProviderInfo, which is what they should have been from the beginning.

## Issues
No issue currently correspond to these changes.

## Testing
Well, it still builds. Kinda disappointed the initial issue didn't cause CI failures.

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbConnectionFactory.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbConnectionFactory.cs
@@ -91,10 +91,8 @@ namespace Microsoft.Data.ProviderBase
             }
         }
 
-        internal virtual DbConnectionPoolProviderInfo CreateConnectionPoolProviderInfo(DbConnectionOptions connectionOptions)
-        {
-            return null;
-        }
+        internal abstract DbConnectionPoolProviderInfo CreateConnectionPoolProviderInfo(
+            DbConnectionOptions connectionOptions);
 
         protected virtual DbMetaDataFactory CreateMetaDataFactory(DbConnectionInternal internalConnection, out bool cacheMetaDataFactory)
         {
@@ -139,10 +137,8 @@ namespace Microsoft.Data.ProviderBase
             return newConnection;
         }
 
-        internal virtual DbConnectionPoolGroupProviderInfo CreateConnectionPoolGroupProviderInfo(DbConnectionOptions connectionOptions)
-        {
-            return null;
-        }
+        internal abstract DbConnectionPoolGroupProviderInfo CreateConnectionPoolGroupProviderInfo(
+            DbConnectionOptions connectionOptions);
 
         private Timer CreatePruningTimer() =>
             ADP.UnsafeCreateTimer(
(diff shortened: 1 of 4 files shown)


cheenamalhotra (approved):

saurabh500:
>> Not sure why these were always being passed in as object, but they are now being passed in as DbConnectionPoolGroupProviderInfo, which is what they should have been from the beginning.


In System.Data.dll which had ODBC/OleDB/(once OracleClient) and SqlClient, the base classes for conn pool were being used to manage pools for all the providers.
DbConnectionPoolGroupProviderInfo was provider specific class and which was being passed to provider specific pool implementation (ODBC pool would have its own quirks and so would OleDb). 

The code you are modifying was passthrough, which needed to pass around an object. 
Now that there is only SqlClient and its pool, the change you have made, makes sense.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbConnectionFactory.cs
```diff
@@ -646,12 +642,23 @@ internal void QueuePoolGroupForRelease(DbConnectionPoolGroup poolGroup)
...
         }
 
-        virtual protected DbConnectionInternal CreateConnection(DbConnectionOptions options, DbConnectionPoolKey poolKey, object poolGroupProviderInfo, IDbConnectionPool pool, DbConnection owningConnection, DbConnectionOptions userOptions)
+        protected virtual DbConnectionInternal CreateConnection(
+            DbConnectionOptions options,
+            DbConnectionPoolKey poolKey,
+            DbConnectionPoolGroupProviderInfo poolGroupProviderInfo,
+            IDbConnectionPool pool,
+            DbConnection owningConnection,
+            DbConnectionOptions userOptions)
```
paulmedynski, line 651: Do we know why this is throwing away userOptions?

paulmedynski (commented):
One question.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbConnectionFactory.cs
benrr101: If I had a piece of paper, I would've traced it out, but you'll have to accept my half-butted explanation for now. As near as I can tell, there's 4 CreateConnection methods - only one has an implementation, the others just call into others or are abstract. It's more confusing since the DbConnectionFactory one _with_ userOptions discards the options, while the SqlConnectionFactory _without_ userOptions just calls into overload with null userOptions. But since the DbConnectionFactory ones are virtual, I just get more confused.

Ultimately - I just deleted all but the one with a complete implementation and the abstract in the base class. The other ones weren't being used anyhow aside from forwarding to overloads. Let me know what you think.

paulmedynski: The answer seems to be "We don't really know", but you didn't change it, so status quo FTW!

paulmedynski (approved):

PR #3420, opened 2025-06-13
Merge | SqlConnectionFactory

benrr101:
## Description
Does what it says on the tin - merges the SqlConnectionFactory class from the netfx and netcore projects into a single implementation in the common project.

One noteworthy change that's a biiit outside the scope: There were two overloads of CreateConnection in SqlConnectionFactory and DbConnectionFactory, one that takes user connection options one that does not. The one that does not, would just call into the one that does passing null as the user connection options. This was confused by the fact that both methods are virtual. To make this easier, I just dropped the overload that doesn't take user connection options, and made the implementation in the base class abstract.

## Issues
Continues work on #1261

## Testing
Code moved, it still compiles, shouldn't be any issues.


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionFactory.cs
```diff
@@ -146,8 +140,11 @@ protected override DbConnectionInternal CreateConnection(
                 //       retrieved from the previous SSE connection. For this UserInstance connection 'Enlist=True'.
                 // options immutable - stored in global hash - don't modify
                 opt = new SqlConnectionString(opt, instanceName, userInstance: false, setEnlistValue: null);
+                //netcore---
```
paulmedynski: What do these comments mean?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionFactory.cs
```diff
@@ -78,7 +72,7 @@ protected override DbConnectionInternal CreateConnection(
 
             // Pass DbConnectionPoolIdentity to SqlInternalConnectionTds if using integrated security.
```
paulmedynski: This comment is stale now.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionFactory.cs
```diff
@@ -193,16 +190,16 @@ override protected DbConnectionPoolGroupOptions CreateConnectionPoolGroupOptions
...
                     connectionTimeout *= 1000;
                 }
                 else if (connectionTimeout >= int.MaxValue / 1000)
                 {
                     connectionTimeout = int.MaxValue;
                 }
-
-                if (opt.Authentication == SqlAuthenticationMethod.ActiveDirectoryInteractive || opt.Authentication == SqlAuthenticationMethod.ActiveDirectoryDeviceCodeFlow)
+                
+                if (opt.Authentication is SqlAuthenticationMethod.ActiveDirectoryInteractive or SqlAuthenticationMethod.ActiveDirectoryDeviceCodeFlow)
```
paulmedynski: Hmmm, not sure I like this syntax.  Is it more/less/the same efficiency?  You're using good ol' `==` on lines 75 and 217.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.stub.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Security;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Data.SqlClient.ConnectionPool;
+
+namespace Microsoft.Data.SqlClient
+{
+    internal class SqlInternalConnectionTds
+    {
+        internal SqlInternalConnetionTds(
```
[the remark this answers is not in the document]
paulmedynski: Seems legit.

paulmedynski (changes requested):
Style/consistency questions.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.stub.cs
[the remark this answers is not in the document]
benrr101: I was thinking how is it possible that this isn't breaking things - then I realized it's the stub file... 🤦‍♂️

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionFactory.cs
benrr101: Whoops... I missed that one.

If you check out commit where I added a "merge file" that's my own little syntax for notating when a block that's specific to a platform exists. Forgot to resolve this one.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionFactory.cs
benrr101: Yes, it's more/less the same efficiency. The only difference is that the values that are compared against must be compile-time constants (eg, enum values), and *if* the type has an override for `==`/`!=` operator, `is`/`is not` will not use the override. 

I personally prefer it when checking against >1 values (since you can omit the first half of the comparison). But I have limits - I won't go so far as to use the (eg) `x is {prop: 4}` syntax, and I try to make the spacing make as much sense. As such, I think I can tweak that line to be more visible what's going on. Let me know what you think.

mdaigle (approved):

paulmedynski (approved):

Issue #3421, opened 2025-06-15
TdsParser: Fatal connection error


Kobuntu:
### Describe the bug
When table field named is "List"

System.InvalidOperationException: Fatal connection error.
   at Microsoft.Data.SqlClient.TdsParser.TryRun(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj, Boolean& dataReady)
   at Microsoft.Data.SqlClient.SqlDataReader.TryCloseInternal(Boolean closeReader)
   at Microsoft.Data.SqlClient.SqlDataReader.Close()
   at Microsoft.Data.SqlClient.SqlDataReader.Dispose(Boolean disposing)
   at System.Data.Common.DbDataReader.DisposeAsync()

### To reproduce
When field named is "List". If field exclude from query, no exception handing.

```c#
SELECT
	[gs].[UserGlobalSet_ID],     <varchar(22)>
	[gs].[User_ID],              <varchar(22)>
	[gs].[List]                  <ntext>                <-- This field !
FROM
	[dbo].[Expl_User_Global_Sets] [gs]
```

### Further technical details
Microsoft.Data.SqlClient version: 6.1.0-preview1.25120.4
.NET: 10.0.0-preview.5
SQL Server version: SQL Server (ver. 15.00.4410)
Operating system: Windows 10

This exception no reprodused on Microsoft.Data.SqlClient version: 6.0.2

ErikEJ:
@Wraith2 sounds familiar?

ErikEJ:
@Kobuntu please share a fully runnable repro (console app and CREATE stayement)

Kobuntu:
I finally managed to reproduce the error
I assume that the problem is in the contents of the field "List

[SqlClientExample.zip](https://github.com/user-attachments/files/20746404/SqlClientExample.zip)

"

ErikEJ:
It could well be fixed in 6.1 preview 2

Kobuntu:
Thank you!

ErikEJ:
But you need to provide a simple repro, using just ADO.NET, please

Kobuntu:
[SqlClientExample.zip](https://github.com/user-attachments/files/20746620/SqlClientExample.zip)

ErikEJ:
Thanks. This also repros on the latest main package @Wraith2 FYI

Simplified repro:


```c#
[...]

PR #3422, opened 2025-06-15
Change TryReadString to use TryReadBytesWithContinue

Wraith2:
Fixes https://github.com/dotnet/SqlClient/issues/3421

Digging into the replication provided I found that if the data type of the string column were changed from `ntext` to `nvarchar(max)` everything worked as expected. The reason for this is that the two types are handled differently internally despite being value equivalent to the user. nvarchar() is a plp type and ntext is not. non-plp strings pass through the TryReadString function and it was tripping up when hitting the continue point because it had not been altered to deal with it. 

TryReadString was using TryReadByteArray and in previous work I have written a version of this which encapsulates continue capable logic, TryReadByteArrayWithContinue. I moved this method down from TdsParser to TdsParserStateObject so it sits at the same level as the original method and then called it from TryReadString. This resolves the problem with minimal changes and no new code. 

- The ability to use the existing api with minor rearrangements is pleasing and I think promising that we've got all the tools we need to handle continue.
 - The code coverage for strings is alarming. I thought that we would have a lot better coverage than we seem to. Strings are a very common path and yet we've found at least 3 combinations that I would have expected to have coverage.
- This really need to go into Preview 2 to make sure anyone testing the preview has as many fixes as possible and that we pick up any new issues cleanly with preview2. @dotnet/sqlclientdevteam @David-Engel 

/cc @ErikEJ @Kobuntu


src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataReaderTest/DataReaderTest.cs
```diff
@@ -790,6 +790,76 @@ static char[] Resize(char[] buffer)
             }
         }
 
+        [ConditionalFact(typeof(DataTestUtility), nameof(DataTestUtility.AreConnStringsSetup))]
```
ErikEJ, line 793: Would it add value to add a `text` column to the test as well?

Wraith2: Probably not, that is handled by TryReadStringWithEncoding which is already continue-aware.

paulmedynski (approved):

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataReaderTest/DataReaderTest.cs
```diff
@@ -790,6 +790,76 @@ static char[] Resize(char[] buffer)
...
+            using (var cn = new Microsoft.Data.SqlClient.SqlConnection(DataTestUtility.TCPConnectionString))
+            {
+                await cn.OpenAsync();
+
+                using (var cmd = cn.CreateCommand())
+                {
+                    cmd.CommandText = $"""
+                   if exists (select *
+                                  from sysobjects
+                                  where name = 'Expl_User_Global_Sets_'
```
David-Engel: The table name needs to be randomized or it will stomp on other parallel test runs. (Mainly an issue when the tests target Azure SQL DB.) It also needs to be dropped at the end.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataReaderTest/DataReaderTest.cs
```diff
@@ -790,6 +790,95 @@ static char[] Resize(char[] buffer)
...
+                await cn.OpenAsync();
+
+                string tableName = DataTestUtility.GenerateObjectName();
+
+                try
+                {
+                    using (var cmd = cn.CreateCommand())
+                    {
+                        cmd.CommandText = $"""
+                   if exists (select *
```
ErikEJ: @Wraith2 I think you can just remove this "drop" section - and maybe put the table name in [ ] ?

ErikEJ (commented):
LGTM

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataReaderTest/DataReaderTest.cs
```diff
@@ -790,6 +790,95 @@ static char[] Resize(char[] buffer)
...
+                                  from sysobjects
+                                  where name = 'Expl_User_Global_Sets_'
+                                    and xtype = 'U')
+                                    
+                   begin
+                        drop table dbo.Expl_User_Global_Sets_
```
cheenamalhotra: Shouldn't exist anymore.

cheenamalhotra (dismissed):
Small change to remove old cleanup query, rest LGTM!

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataReaderTest/DataReaderTest.cs
```diff
@@ -790,6 +790,95 @@ static char[] Resize(char[] buffer)
...
...k6NwoEaToyMgoEaTo1MgoFaToxNDYKBWk6MTk0CgRpOjM2CgRpOjQwCgVpOjE2MgoFaToyNDUKBWk6MTk5CgRpOjY5CgRpOjY3CgVpOjE1NAoFaToxNzYKBWk6MjIwCgVpOjIzMwoFaToxNTYKBWk6MTkzCgRpOjg2CgVpOjE5OAoFaToyNzcKBWk6MjA5CgVpOjIxMAoFaToyMzIKBWk6MjE1CgRpOjYyCgRpOjMyCgVpOjE3MAoFaToxNTcKBWk6MjUyCgRpOjU0CgRpOjQxCgVpOjE1NQoFaToyNzkKBWk6MTY5CgVpOjE0NwoFaToxNDgKBGk6MzcKBGk6MjkKBWk6MTc4CgVpOjE3NwoFaToyMTMKBGk6MzMKBWk6MjgzCgVpOjE3NQoEaTo1OAoEaToyMAoEaTo2MwoEaToxNQoDaTozCgVpOjIwMAoDaToyCgVpOjI2MQoFaToyODYKBWk6MTM1CgVpOjEzNgoFaToyNzYKBGk6NDcKA2k6OAoEaTo5NwoFaToyODgKBWk6MTM5CgVpOjI4NwoFaToxOTAKBWk6MTkyCgVpOjE5MQoEaToxOAoEaTo3NgoFaToxMzAKBWk6MjA0CgVpOjI2MAoFaToyMzEKBWk6MjE0CgRpOjYxCgVpOjE0MQoFaToyODAKBWk6MjI4CgVpOjIyOQoFaToyMzQKBWk6MjM1CgVpOjI1MAoFaToyNTEKBWk6MTg2CgVpOjE4OAoFaToxODcKBWk6MjMwCgRpOjQyCgVpOjE5NgoFaToyNDAKBWk6MjM5CgVpOjE0MAoEaToyMwoFaToyNDEKBWk6MTc0CgRpOjcyCgVpOjIwNQoFaToyNTYKBWk6MjU1CgRpOjE5CgRpOjQ4CgNpOjYKBWk6MTAyCgVpOjIxMQoFaToyMTIKBGk6NTcKBWk6MjYyCgVpOjI0NAoFaToyMzYKBWk6MjM3CgVpOjIzOAoDaTo1CgVpOjEwNQoFaToyNDgKBWk6MjQ5CgVpOjE4MwoFaToxODQKBGk6MzEKBWk6MTgwCgN0OjMKBHQ6MTEKBHQ6MTAKBHQ6MjEKA3Q6NAoDdDo2CgN0OjkKA3Q6NwoDdDo4CgR0OjEzCgR0OjE4CgR0OjEyCgR0OjE5CgR0OjE3CgN0OjUKBHQ6MTUKA3Q6MQoEdDoyMAoDdDoyCgR0OjE2CgR0OjE0";
```
benrr101, line 796: Why not generate it... we have lots of code that generates random values, could be a good opportunity to add a utility to common

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataReaderTest/DataReaderTest.cs
benrr101: I think your indents got messed up here

benrr101 (approved):
Low stakes comments on the test more than anything

paulmedynski (approved):

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataReaderTest/DataReaderTest.cs
Wraith2: fixed.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataReaderTest/DataReaderTest.cs
Wraith2: Fixed.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataReaderTest/DataReaderTest.cs
Wraith2: Because i know this specific value has the required characteristics when split over multiple packets to cause the original problem 100% effectively. I could spend time trying to reduce it but it gains me nothing.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataReaderTest/DataReaderTest.cs
Wraith2: it's using """'s so the indents are handled by the compiler. Not a feature I've used before but it's part of the original reproduction and it works so I left it alone.

paulmedynski (dismissed):

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataReaderTest/DataReaderTest.cs
```diff
@@ -790,6 +790,78 @@ static char[] Resize(char[] buffer)
...
+                      values ('80004Q4WZ1350KO8NT59RM', '_', 1, '{stringValue}', 1, 2, 1, 1, newid(), 1);
+
+                   """;
+
+                        await cmd.ExecuteNonQueryAsync();
+
+                        cmd.CommandText = $""""
+                  SELECT
+                  	[gs].[List],
+                  	[gs].[UseProtoSerializer],
```
cheenamalhotra: The test is currently failing with:

Microsoft.Data.SqlClient.SqlException : Incorrect syntax near the keyword 'FROM'.

```suggestion
                  	[gs].[UseProtoSerializer] 
```

cheenamalhotra (approved):

benrr101 (approved):

PR #3423, opened 2025-06-16
Tests | Expand UDT serialization code coverage

edwardneal:
## Description

I've been looking at a set of performance improvements and code health improvements to the serialization of CLR UDTs - [SqlNormalizer.cs](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SqlNormalizer.cs) and [SqlSer.cs](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SqlSer.cs). At present though, there isn't sufficient test coverage to prevent regressions. This PR widens that coverage, adding a set of new unit tests for the serialization and deserialization of CLR UDTs. By the end of this, I think we cover every reachable branch.

This is also the first set of additional tests in the new UnitTests project, and the first tests which test current library functionality. I've got no objection if anyone wants to start the new project with a consistent code style.

## Issues

None.

## Testing

All tests pass, all values match existing behaviour.

### DIFF
--- a/src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft/Data/SqlTypes/SqlVectorTest.cs
+++ b/src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft/Data/SqlTypes/SqlVectorTest.cs
@@ -11,8 +11,6 @@ using Xunit;
 
 namespace Microsoft.Data.SqlTypes.UnitTests;
 
-#pragma warning disable CS1591 // Test classes do not require XML documentation comments
-
 public class SqlVectorTest
 {
     #region Tests
@@ -246,5 +244,3 @@ public class SqlVectorTest
     
     #endregion
 }
-
-#pragma warning restore CS1591


src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft.Data.SqlClient.UnitTests.csproj
```diff
@@ -25,6 +25,7 @@
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
     <PackageReference Include="Microsoft.DotNet.XUnitExtensions" />
+    <ProjectReference Include="$(SqlServerSource)Microsoft.SqlServer.Server.csproj" />
```
paulmedynski: Can this be a PackageReference instead?  The UnitTests project shouldn't be testing anything from the SqlServer package.

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft/Data/SqlClient/UdtSerialization/NativeSerializationTest.cs
```diff
@@ -0,0 +1,268 @@
...
+        yield return [new ShortWrapperStruct { Field1 = 0x1234 },
+            new byte[] { 0x92, 0x34 }];
+        yield return [new UIntWrapperStruct { Field1 = 0xFFFFFFFF },
+            new byte[] { 0xFF, 0xFF, 0xFF, 0xFF }];
+        yield return [new IntWrapperStruct { Field1 = -0x12345678 },
+            new byte[] { 0x6D, 0xCB, 0xA9, 0x88 }];
+        yield return [new ULongWrapperStruct { Field1 = ulong.MaxValue },
+            new byte[] { 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF }];
+        yield return [new LongWrapperStruct { Field1 = long.MinValue },
+            new byte[] { 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00 }];
```
paulmedynski: Shouldn't the first byte be 0x80?

https://learn.microsoft.com/en-us/dotnet/api/system.int64.minvalue?view=net-9.0

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft/Data/SqlClient/UdtSerialization/SerializedTypes.cs
```diff
@@ -0,0 +1,207 @@
...
+using Microsoft.SqlServer.Server;
+using System;
+using System.Data.SqlTypes;
+using System.IO;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.Data.SqlClient.UnitTests.UdtSerialization.SerializedTypes;
+
+// Simple cases: a struct containing one of the designated primitive types
+[SqlUserDefinedType(Format.Native)] [StructLayout(LayoutKind.Sequential)] public struct BoolWrapperStruct { public bool Field1; }
```
paulmedynski: I think I'd prefer the attribute on one line, and the struct definition on the next.

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft/Data/SqlClient/UdtSerialization/InvalidSerializationTest.cs
```diff
@@ -0,0 +1,41 @@
...
+using Xunit;
+
+namespace Microsoft.Data.SqlClient.UnitTests.UdtSerialization;
+
+public class InvalidSerializationTest
+{
+    [Fact]
+    public void RequiresSqlUserDefinedTypeAttribute()
+    {
+        using MemoryStream stream = new MemoryStream();
```
paulmedynski: For new code, thoughts on avoiding duplicating type names for new() ?  I prefer keep the type name on the left:
```
using MemoryStream stream = new();
```

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft/Data/SqlClient/UdtSerialization/NativeSerializationTest.cs
```diff
@@ -0,0 +1,268 @@
...
+    /// Attempts to serializes an instance of a class.
+    /// </summary>
+    /// <seealso cref="CannotSerializeNestedClass"/>
+    [Fact]
+    public void CanSerializeTopLevelClass()
+    {
+        NestedBoolWrapperClass validWrapper = new NestedBoolWrapperClass()
+        {
+            Field1 = true,
+            Field2 = new BoolWrapperStruct() { Field1 = true }
```
paulmedynski, line 423: I'm on the fence about using `new(){ Field1 = true }` here, since that would eliminate the type name entirely.  But for tests, maybe that's fine?

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft/Data/SqlClient/UdtSerialization/NativeSerializationTest.cs
```diff
@@ -0,0 +1,268 @@
...
+        };
+        using MemoryStream stream = new MemoryStream();
```
paulmedynski: Is it worth making this an instance field, rather than having each test create one?  Same for the other test classes.  The "Constructor and Dispose" approach seems sufficient here.

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft/Data/SqlClient/UdtSerialization/InvalidSerializationTest.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        InvalidUdtException exception = Assert.Throws<InvalidUdtException>(
```
paulmedynski: I'd be fine with `var exception` here, since it's obvious what Assert.Throws<>() returns.

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft/Data/SqlClient/UdtSerialization/NativeSerializationTest.cs
```diff
@@ -0,0 +1,268 @@
...
+
+        SerializationHelperSql9.Serialize(stream, validWrapper);
+    }
+
+    /// <summary>
+    /// Attempts to serializes a field referring to an instance of a class.
+    /// Verifies that this succeeds, and that Native format serialization only operates with primitive types
```
paulmedynski: succeeds -> fails  ?

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft/Data/SqlClient/UdtSerialization/NativeSerializationTest.cs
```diff
@@ -0,0 +1,268 @@
...
+
+    /// <summary>
+    /// Serializes an object, verifies the value and the size of the object, then roundtrips it and verifies the result is identical.
+    /// </summary>
+    /// <param name="inputValue">Object to serialize.</param>
+    /// <param name="expectedValue">Expected serialization output.</param>
+    private static void RoundtripType(object inputValue, byte[] expectedValue)
+    {
+        using MemoryStream stream = new MemoryStream();
+        object readPrimitive;
```
paulmedynski: Declare and assign together on line 258.

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft/Data/SqlClient/UdtSerialization/UserDefinedSerializationTest.cs
```diff
@@ -0,0 +1,97 @@
...
+
+        Assert.Throws<InvalidCastException>(
+            () => SerializationHelperSql9.Serialize(stream, new UserDefinedDoesNotImplementIBinarySerialize()));
+    }
+
+    private static void RoundtripType<T>(T userObject)
+        where T : IFormattingProgress
+    {
+        using MemoryStream stream = new MemoryStream();
+        byte[] serializedValue;
```
paulmedynski: Declare and assign together, this line and the next.

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft/Data/SqlClient/UdtSerialization/NativeSerializationTest.cs
paulmedynski: Also, I think we should enable nullable for the entire UnitTest project.  Can you do that?  Maybe implicit usings too.

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft/Data/SqlClient/UdtSerialization/InvalidSerializationTest.cs
edwardneal: That sounds good to me. I'm in favour of any style which makes the variable/field type clear, and keeping the type name on the left sounds sensible enough to me. This covers the `stream` and the `exception` variables, but not the `BoolWrapperStruct` field.

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft/Data/SqlClient/UdtSerialization/NativeSerializationTest.cs
edwardneal: Afraid not. With signed bytes/ints/longs, MS-SSCLRT will flip the top bit of the most significant byte. This is also the reason why a `ShortWrapperStruct` with a value of `0x1234` persists to a value of `0x9234`.

paulmedynski: Wild!

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft/Data/SqlClient/UdtSerialization/InvalidSerializationTest.cs
edwardneal: I've changed this. I'm also testing the approach of enforcing the code style via editorconfig, so we can catch this earlier in the new project.

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft/Data/SqlClient/UdtSerialization/NativeSerializationTest.cs
edwardneal: Fixed. I've also enabled XML documentation, so we get a CS1591 warning forcing us to add a description to each test. Enforcing that might be overkill for simpler unit tests - I'm happy to remove it if so.

paulmedynski (dismissed):

paulmedynski (dismissed):

edwardneal:
I've merged main, with the unreachable code paths removed. Could someone re-run CI so we can verify the up-to-date coverage please?

paulmedynski (approved):

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft.Data.SqlClient.UnitTests.csproj
mdaigle: It's definitely not directly under test, but if we move all of our testing under here we'll lose coupling between changes to the SqlServer.Server project and the test cases that cover it indirectly.

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft/Data/SqlClient/ConnectionPool/ChannelDbConnectionPoolTest.cs
```diff
@@ -11,6 +11,7 @@
 
 namespace Microsoft.Data.SqlClient.UnitTests
 {
+#pragma warning disable CS1591 // Test classes do not require XML documentation comments
```
benrr101: Agreed, but are we going to need this on each file? If so, can we just add it at the project level?

And ... why isn't this breaking our builds - are you using more advanced analyzers than we currently use?

src/Microsoft.Data.SqlClient/tests/UnitTests/InternalsVisibleToTest.cs
```diff
@@ -4,8 +4,14 @@
 
 namespace Microsoft.Data.SqlClient.UnitTests
 {
+    /// <summary>
```
benrr101: We can probably avoid adding these if we disable CS1519 at the project level.

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft.Data.SqlClient.UnitTests.csproj
```diff
@@ -7,6 +7,8 @@
     <IntermediateOutputPath>$(ObjFolder)$(Configuration).$(Platform).$(AssemblyName)</IntermediateOutputPath>
     <OutputPath>$(BinFolder)$(Configuration).$(Platform).$(AssemblyName)</OutputPath>
     <IsTestProject>true</IsTestProject>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
```
benrr101: What benefit does this grant us?

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft/Data/SqlClient/UdtSerialization/NativeSerializationTest.cs
```diff
@@ -0,0 +1,290 @@
...
+    {
+        _stream.Dispose();
+    }
+
+    /// <summary>
+    /// Provides a collection of test data representing non-null primitive type values and their corresponding
+    /// serialized byte arrays.
+    /// </summary>
+    /// <see cref="SerializePrimitiveType"/>
+    public static IEnumerable<object[]> SerializedNonNullPrimitiveTypeValues()
```
benrr101: Can we use the new TheoryData types here? https://andrewlock.net/creating-strongly-typed-xunit-theory-test-data-with-theorydata/

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft/Data/SqlClient/UdtSerialization/NativeSerializationTest.cs
```diff
@@ -0,0 +1,290 @@
...
+
+    /// <summary>
+    /// Attempts to serialize various structs containing non-null primitive types.
+    /// Verifies that the method does not throw, that serialized byte output is correct, and that the value round-trips.
+    /// </summary>
+    /// <param name="primitive">Primitive to serialize and to compare against.</param>
+    /// <param name="expectedValue">Expected byte output.</param>
+    [Theory]
+    [MemberData(nameof(SerializedNonNullPrimitiveTypeValues))]
+    public void SerializePrimitiveType(object primitive, byte[] expectedValue)
```
benrr101: Nit: `=>` on same line as declaration, pls

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft/Data/SqlClient/UdtSerialization/NativeSerializationTest.cs
```diff
@@ -0,0 +1,290 @@
...
+        SerializationHelperSql9.Serialize(_stream, validWrapper);
+    }
+
+    /// <summary>
+    /// Attempts to serialize a field referring to an instance of a class.
+    /// Verifies that this fails, and that Native format serialization only operates with primitive types and value types containing these.
+    /// </summary>
+    /// <seealso cref="CanSerializeTopLevelClass"/>
+    [Fact]
+    public void CannotSerializeNestedClass()
```
benrr101: I'd really love it if we can have our unit tests named in a format like: `Method_Scenario_ExpectedResult` or at least with `Method_Scenario`. This would look like `Serialize_NestedClass_Throws`

Open to arguing over this one though.

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft/Data/SqlClient/UdtSerialization/NativeSerializationTest.cs
```diff
@@ -0,0 +1,290 @@
...
+    {
+        InvalidNestedBoolWrapperClass invalidWrapper = new()
+        {
+            Field1 = true,
+            Field2 = new BoolWrapperClass() { Field1 = true }
+        };
+
+        var ex = Assert.Throws<Exception>(() => SerializationHelperSql9.Serialize(_stream, invalidWrapper));
```
benrr101, line 443: I'm hesitant to even mention it ... in a previous job I had it drilled into my head that all unit tests should follow a arrange/act/asset pattern (with comments to indicate as such). I can see you've got them structured as such (even if it's without the comments). But, the one thing they insisted on was that for assert-that-it-throws tests, the "act" stage is assigning a lambda and the "assert" stage is asserting the lambda throws - two separate lines. Idk if it's stockholm syndrome or what, but now I think that's the "better" way to do unit tests. I won't block it if you don't do it, but if you agree and decide to do it, that'd be cool 👍

src/Microsoft.Data.SqlClient/tests/UnitTests/.editorconfig
```diff
@@ -0,0 +1,11 @@
+# editorconfig.org
```
benrr101, line 1: Ehhh, I'm not opinionated enough about these behaviors to enforce them in the editor config. I'd kinda prefer it if we don't add this editorconfig.

benrr101 (changes requested):
Glad to have more tests! A handful of requests, but nothing too crazy

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft.Data.SqlClient.UnitTests.csproj
edwardneal: It forces an XML comment on every publicly-visible member, generating CS1591 if that comment's not present. Since every test's publicly visible, it forces us to document each test case. It's very similar to the changes to editorconfig, in that it forces PRs to be written without the nits which sometimes come up during PR review.

It's up for discussion, but I think it might be a good idea to start enforcing a code style in the new UnitTests project (then eventually rolling the editorconfig out repo-wide.)

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft/Data/SqlClient/UdtSerialization/NativeSerializationTest.cs
edwardneal: Yes - and it looks like there might be an [analyzer](https://xunit.net/xunit.analyzers/rules/xUnit1042) to enforce this everywhere (depending on your opinions on the project file's comment.)

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft.Data.SqlClient.UnitTests.csproj
benrr101: I'm ok with enforcing code style in unit tests, but I really don't think we should force xmldocs on test classes/methods. I'm all for documenting real classes, but my standards on tests are a lot lower.

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft/Data/SqlClient/UdtSerialization/NativeSerializationTest.cs
benrr101: "Enforce" might be too strong - like I said above, my standards for tests are a lot lower :) For me, they show up as hints as I go, so that's how I found out they existed in the first place. But I think there might be some cases where it's not feasible to use them.

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft.Data.SqlClient.UnitTests.csproj
benrr101: @paulmedynski do you have opinions?

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft/Data/SqlClient/UdtSerialization/NativeSerializationTest.cs
```diff
@@ -0,0 +1,290 @@
...
+            new byte[] { 0x01, 0x80, 0x00, 0x00, 0x00, 0x00, 0x00, 0x2A, 0xF8 }];
+    }
+
+    /// <summary>
+    /// Provides a collection of test data representing serialized values of nested non-null primitive types.
+    /// </summary>
+    /// <see cref="SerializeNestedPrimitiveType"/>
+    public static IEnumerable<object[]> SerializedNestedNonNullPrimitiveTypeValues()
+    {
+        yield return [new NestedBoolWrapperStruct { Field1 = true, Field2 = new BoolWrapperStruct { Field1 = false } },
```
benrr101: Nit: If we have multiple lines, can we always put the item on the next line?
eg
```
yield return [
    new ...,
    new ...,
];


And I see the byte arrays are split across two lines when they could fit on one line. Was there a reason for that?

File: src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft/Data/SqlClient/UdtSerialization/SerializedTypes.cs [hunk 20]
```diff
@@ -0,0 +1,335 @@
...
+using System.Data.SqlTypes;
+using System.IO;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.Data.SqlClient.UnitTests.UdtSerialization.SerializedTypes;
+
+// Simple cases: a struct containing one of the designated primitive types
+[SqlUserDefinedType(Format.Native)]
+[StructLayout(LayoutKind.Sequential)]
+internal struct BoolWrapperStruct { public bool Field1; }
```
benrr101, line 16: Just thinking, would this be a good case for a record? I seem to recall the syntax for those are a bit simpler.

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft.Data.SqlClient.UnitTests.csproj
paulmedynski: I generally don't care about Intellisense for non-public-API things.  Unit tests aren't public APIs, so I'm not sure we need the structure.  The intention to promote good documentation habits is noble though, so I applaud that effort.  I could certainly improve my test commentary sometimes.

I find the XML documentation style annoying to create and maintain, so minimizing its use is fine by me.  I prefer plain ol' English comments with as much detail as necessary to understand how/why/when to use something.  We have a lot of existing XML docs that don't meet those standards.  Will enforcing XML docs actually help?

Having said all of that, I'm open to enabling this flag and "seeing how it goes".  A few PRs with it enabled will be enough for me to decide if I like it or not :wink:

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft/Data/SqlClient/UdtSerialization/SerializedTypes.cs
edwardneal: I don't think it'd do much in this case, we'd end up with any number of types which look similar to the below.

```cs
[SqlUserDefinedType(Format.Native)]
[StructLayout(LayoutKind.Sequential)]
internal record struct BoolWrapperStruct (bool Field1);
```

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft/Data/SqlClient/UdtSerialization/NativeSerializationTest.cs
edwardneal: I'll reformat the square brackets. The layout of the byte arrays is purely a matter of readability: we have each serialized representation of a field on a new line. Would an inline comment similar to the below be simpler?

```cs
yield return [
    new NestedSByteWrapperStruct { Field1 = -0x01, Field2 = new SByteWrapperStruct { Field1 =0x01 } },
    new byte[] { /* Field1: */ 0x7F, /* Field2.Field1: */ 0x81 }
];

yield return [
    new NestedSqlInt64WrapperStruct { Field1 = long.MinValue, Field2 = new SqlInt64WrapperStruct { Field1 = long.MaxValue } },
    new byte[] { /* Field1: */ 0x01, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
        /* Field2.Field1: */ 0x01, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF }
];
```

edwardneal: Taking a second look, the line break formatting actually had to change as part of the TheoryData swap. Could you re-check this PR's diff as of the latest commit please?

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft.Data.SqlClient.UnitTests.csproj
edwardneal: I've just merged and pushed to the branch after noticing this evening's CI run failed due to the missing XML comments on UserAgentInfoTest.

I'm in two minds on the change. It's irritating to deal with, but the suppressions are highlighting situations where the test  case descriptions aren't ideal: ChannelDbConnectionPoolTest doesn't have any, and UserAgentInfoTest uses a mix of fully-fledged XML comments and standard comments above the method. It might cast too wide a net to be useful, depending on whether we actually need to care about those points.

If it seems to generate unnecessary work in writing tests, I've got no objection to removing it.

paulmedynski (approved):

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft/Data/SqlClient/UdtSerialization/NativeSerializationTest.cs
benrr101: Ok, I get it now - it makes sense to split the byte array into multiple lines for the multiple fields. However, I think the part we're implicitly disagreed on is the positioning of the first new line

eg imo less good
```
new byte[] { 0x01, 0x02, 0x03,
    0x01, 0x02, 0x03 };
```

eg imo better (optionally with explanation at the end of the line)
```
new byte[] {     // (actually this might have to go on the next line too, I forget)
    0x01, 0x02, 0x03,
    0x01, 0x02, 0x03,
}
```

I know it's probably silly, but imo it improves readability to have all sub-"objects" (values, arguments, etc) either all on the same line or all on separate lines.

But otherwise, the changes in the commit look ok

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft.Data.SqlClient.UnitTests.csproj
benrr101: I will continue to push for "let's just turn it off" 😅
Especially considering in the latest commit we turn it off *temporarily* and say it's not required. If it's not required, why do we have a requirement for it? 😂 

(Idk, my philosophy for unit tests are that they should be so byte-sized it's pretty hard to misunderstand what they're testing, but the name should give a good explanation, and if anything needs more explanation it's commented inside the test. The UserAgenInfo tests could stand more comments but imo are doing too much in one test and not covering all scenarios, so I'll have to do a better job reviewing those)

src/Microsoft.Data.SqlClient/tests/UnitTests/.editorconfig
benrr101: Gonna run these past the team and see what they prefer.

benrr101 (changes requested):
Still going to request we remove the requirement for comments on unit tests. And will get back with results of poll for variable declaration style.

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft/Data/SqlClient/UdtSerialization/NativeSerializationTest.cs
edwardneal: That makes sense, I've just corrected the formatting.

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft.Data.SqlClient.UnitTests.csproj
edwardneal: Fair enough - I've just switched it off.

src/Microsoft.Data.SqlClient/tests/UnitTests/.editorconfig
edwardneal: Speaking from the peanut gallery: I've seen feedback on PRs about these style comments. Once the merge is complete, we should try to reach a point where we have a consistent solution-wide code style, enforced at the point of compilation without any warnings/messages/errors in existing code. This is a first step towards that goal, enforcing points which have already been commented on in PRs on a new project.

benrr101: It seems the consensus it for what your changes enforce. 👍 

Hey you in the peanut gallery, quiet over there! *throws shoe*
I'm always hesitant to "enforce" style, especially style that's enforced at compilation or check-in. I've probably said it before but I generally think of coding style as more of an artform than a hard science. Sometimes I'll use one approach sometimes another depending on which ... _flows_ better 🤷‍♂️And my style comments are generally me trying to keep our codebase in a nicely flowing state. But yeah, what works in my head doesn't necessarily work in everyone else's heads. So, maybe we do need some kind of enforceable rules - at least in some cases. And I'd support cleaning up the codebase as per those rules.

PR #3431, opened 2025-06-18
Cleanup | SMI Cleanup (part 3)

benrr101:
## Description
This PR is a continuation of work done in parts 1 and 2 to clean up dead code in SMI. This time I took a slightly more heavy handed approach - if the class isn't being constructed anywhere, I deleted it. I then went through and removed any methods, properties, and variables that weren't being used in the Server folder, using IDE analysis of the files.

One of the biggest not-strictly-delete changes is to the methods in ITypedGetters/ITypedSetters. The methods in ITypedGetters/ITypedSetters did not take an EventSink in their arguments, but the interfaces themselves were not being used. The methods in ITypedGettersV3 and ITypedSettersV3 were being used around the codebase and take an EventSink in their arguments, but were never using the sink. As such, I deleted the ITypedGetters/ITypedSetters, and modified the ITypedGettersV3/ITypedSettersV3 to not take an EventSink in their arguments.

## Issues
Believe it or not, this actually contributes to #1261 - ~~two~~ **four** of the classes wholesale deleted from the project were slated for merging! 
* SqlDataReaderSmi.cs
* SqlInternalConnectionSmi
* SqlSequentialStreamSmi
* SqlSequentialTextReaderSmi

## Testing
I attempted to run tests locally and identified a few that were failing due to reflection accessing methods that were removed. I suspect there will be more of these, but I was too impatient with the tests to wait for more to tests to complete.


src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
```diff
@@ -1687,15 +1686,11 @@ public void WriteToServer(DbDataReader reader)
...
                 statistics = SqlStatistics.StartTimer(Statistics);
+                
                 ResetWriteToServerGlobalVariables();
                 _rowSource = reader;
                 _dbDataReaderRowSource = reader;
                 _sqlDataReaderRowSource = reader as SqlDataReader;
-
-                if (_sqlDataReaderRowSource != null)
-                {
-                    _rowSourceIsSqlDataReaderSmi = _sqlDataReaderRowSource is SqlDataReaderSmi;
```
benrr101: SqlDataReaderSmi has been removed, so this check would always be false

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnection.cs
```diff
@@ -1906,33 +1906,24 @@ private bool TryOpenInner(TaskCompletionSource<DbConnectionInternal> retry)
                 // GetBestEffortCleanup must happen AFTER OpenConnection to get the correct target.
                 bestEffortCleanupTarget = SqlInternalConnection.GetBestEffortCleanupTarget(this);
 
-                var tdsInnerConnection = (InnerConnection as SqlInternalConnectionTds);
-                if (tdsInnerConnection == null)
+                var tdsInnerConnection = (SqlInternalConnectionTds)InnerConnection;
```
benrr101, line 1909: This cast can be safely made without type checking as SqlInternalConnectionTds is the only type of connection that exists. In the future, I hope to merge SqlInternalConnection and SqlInternalConnectionTds to avoid making these (guaranteed to be safe) casts.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SqlMetaData.cs
```diff
@@ -97,80 +97,44 @@ public sealed partial class SqlMetaData
...
-            new SqlMetaData("date", SqlDbType.Date,
-                    3, 10,0, 0, SqlCompareOptions.None, false),                // SqlDbType.Date
-            new SqlMetaData("time", SqlDbType.Time,
-                    5, 0, 7, 0, SqlCompareOptions.None, false),                // SqlDbType.Time
-            new SqlMetaData("datetime2", SqlDbType.DateTime2,
-                    8, 0, 7, 0, SqlCompareOptions.None, false),                // SqlDbType.DateTime2
-            new SqlMetaData("datetimeoffset", SqlDbType.DateTimeOffset,
-                   10, 0, 7, 0, SqlCompareOptions.None, false),                // SqlDbType.DateTimeOffset
-            };
+        private static readonly SqlMetaData[] s_defaults =
```
benrr101, line 100: This block of changes was just because the `isPartialLength` argument is not being used (since it was deleted)

benrr101 (commented):
Calling out some notable non-deletion changes

mdaigle (dismissed):

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnection.cs
```diff
@@ -1906,33 +1906,24 @@ private bool TryOpenInner(TaskCompletionSource<DbConnectionInternal> retry)
...
-                    {
-                        _statistics._openTimestamp = ADP.TimerCurrent();
-                        tdsInnerConnection.Parser.Statistics = _statistics;
-                    }
-                    else
-                    {
-                        tdsInnerConnection.Parser.Statistics = null;
-                        _statistics = null; // in case of previous Open/Close/reset_CollectStats sequence
-                    }
+                    tdsInnerConnection.Parser.Statistics = null;
```
paulmedynski, line 1925: This should check that Parser isn't null.  I think `Parser?.Statistics` is sufficient.  And then remove the unnecessary Debug.Assert() above.

paulmedynski (changes requested):
One improvement suggested.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnection.cs
benrr101: @paulmedynski I'm not sure about what you're suggesting... We can't assign to `tdsInnerConnection.Parser?.Statistics`, and because the Assert is there above, we can still safely assume that Parser exists.

paulmedynski: Ah yes, null conditional assignment is in preview.
```
else
{
    if (tdsInnerConnection.Parser is not null)
    {
        tdsInnerConnection.Parser.Statistics = null;
    }
    _statistics = null;
}
```
The `Debug.Assert()` doesn't exist in Release builds, so we must avoid a null exception programmatically.  The old code has this problem as well, but I think it's worth fixing.  I guess we don't have an existing test for this either.

paulmedynski: Upon further digging, this is an existing wart that isn't worth addressing.  Parser may be null before a login attempt, and during/after disposal, so it _could_ happen here, but probably not due to the rest of the machinery.

mdaigle (approved):

PR #3433, opened 2025-06-19
Introduce SqlVectorFloat32 SqlType class for vector datatype support

### LINKED ISSUE #3317: Native Vector datatype support
Tracking issue for introducing native vector datatype support in MDS.

apoorvdeshmukh:
## Description
Adds Vector datatype support for Float32 with Sqltype SqlVectorFloat32

### Design Considerations

- Some inputs have been considered from #3382 
- Between a generic class definition SqlVector<T> and concrete type SqlVectorFloat32, after comparison it was decided to use concrete types.
- In future we may consider using SqlVector<T> as base class to isolate common functionality between different Vector types but the public API surface will continue to use concrete definitions so that developers know the exact supported type they are working with.
- SqlVectorFloat32 will be encouraged as the only way to work with Float32 vector data when using SqlDbType enum value for Vector.
- Output parameters for reading SqlVectorFloat32 return value will be initialized through SqlVectorFloat32(n) constructor for SqlParameter.Value to  allow users to correctly specify the size and type for the vector data they expect to read.

## Issues
Closes #3317 

## Testing
Tests are split into three categories
Unit Tests for SqlVectorFloat32 under SqlVectorFloat32Tests.cs
Backward Compatibility Test Suite to validate behavior when exchanging vector data as varchar(max)
Native Vector Support Test Suite highlighting the supported functionality


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlTypes/SqlVectorFloat32.cs
```diff
@@ -0,0 +1,175 @@
...
+        internal SqlVectorFloat32(byte[] rawBytes)
+        {
+            if (!ValidateRawBytes(rawBytes))
+            {
+                throw ADP.InvalidVectorHeader();
+            }
+            _rawBytes = rawBytes;
+            _elementCount = rawBytes[2] | (rawBytes[3] << 8);
+            _elementType = rawBytes[4];
+            _elementSize = sizeof(float);
```
samsharma2700: `_elementSize` can be either made a const with value of 4 (since `sizeof(float)` causes a compile-time error outside an `unsafe` context) or it can be `static readonly` if you prefer to set it at runtime.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlTypes/SqlVectorFloat32.cs
```diff
@@ -0,0 +1,175 @@
...
+            _elementCount = length;
+            _elementType = (byte)MetaType.SqlVectorElementType.Float32;
+            _elementSize = sizeof(float);
+            _rawBytes = Array.Empty<byte>();
+        }
+
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlTypes/SqlVectorFloat32.xml' path='docs/members[@name="SqlVectorFloat32"]/ctor2/*' />
+        public SqlVectorFloat32(ReadOnlyMemory<float> values)
+        {
+            if (values.IsEmpty)
```
samsharma2700: Since the TDS protocol does not prohibit empty vectors (length 0), we should allow values to be empty here and defer handling to the server. Rejecting it at the client level might be unnecessarily strict unless there's a known server-side constraint.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlTypes/SqlVectorFloat32.cs
```diff
@@ -0,0 +1,175 @@
...
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlTypes/SqlVectorFloat32.xml' path='docs/members[@name="SqlVectorFloat32"]/Values/*' />
+        public float[] Values
+        {
+            get
+            {
+                if (_rawBytes.Length == 0)
+                {
+                    return Array.Empty<float>();
+                }
+                int elementCount = _rawBytes[2] | (_rawBytes[3] << 8);
```
samsharma2700: Maybe use `_elementCount` here instead of recalculating `elementCount`, since it's already assigned earlier.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlTypes/SqlVectorFloat32.cs
```diff
@@ -0,0 +1,175 @@
...
+#if NETFRAMEWORK
+                // Allocate array and copy bytes into it
+                float[] result = new float[elementCount];
+                Buffer.BlockCopy(_rawBytes, 8, result, 0, elementCount * sizeof(float));
```
samsharma2700: Reminder from our discussion: Instead of copying the block into a new array, consider using a `ReadOnlyMemory<float>` or `ReadOnlySpan<float>` with the known offset, since you already have the element count and data starting position here.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlTypes/SqlVectorFloat32.cs
```diff
@@ -0,0 +1,175 @@
...
+        #region ISqlVectorProperties
+        byte ISqlVector.ElementType => _elementType;
+        byte ISqlVector.ElementSize => _elementSize;
+        byte[] ISqlVector.VectorPayload => _rawBytes;
+        #endregion
+
+        #region Helpers
+        private void InitializeVectorBytes(ReadOnlyMemory<float> values)
+        {
+            //Header Bytes
```
samsharma2700: nit: Consider adding a brief comment here as a quick reference for readers scanning the code.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlTypes/SqlVectorFloat32.cs
```diff
@@ -0,0 +1,175 @@
...
+using Microsoft.Data.Common;
+using Microsoft.Data.SqlClient;
+
+#nullable enable
+namespace Microsoft.Data.SqlTypes
+{
+    /// <include file='../../../../doc/snippets/Microsoft.Data.SqlTypes/SqlVectorFloat32.xml' path='docs/members[@name="SqlVectorFloat32"]/SqlVectorFloat32/*' />
+    public sealed class SqlVectorFloat32 : INullable, ISqlVector
+    {
+        #region Constants
```
samsharma2700, line 14: nit: Might be worth adding a quick explanation here to aid readability, especially for someone unfamiliar with this part of the logic and fields.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlTypes/SqlVectorFloat32.cs
```diff
@@ -0,0 +1,175 @@
...
+
+        private bool ValidateRawBytes(byte[] rawBytes)
+        {
+            if (rawBytes.Length == 0 || rawBytes.Length < TdsEnums.VECTOR_HEADER_SIZE)
+                return false;
+            
+            if (rawBytes[0] != VecHeaderMagicNo || rawBytes[1] != VecVersionNo || rawBytes[4] != VecTypeIndicator)
+                 return false;
+            
+            int elementCount = rawBytes[2] | (rawBytes[3] << 8);
```
samsharma2700: Since this little-endian 16-bit read is used repeatedly, consider encapsulating it in a utility method to clarify intent and centralize the byte access logic.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlTypes/SqlVectorFloat32.cs
apoorvdeshmukh: Server is known to throw the following error in case empty vector is sent from the client.
Unhandled exception. Microsoft.Data.SqlClient.SqlException (0x80131904): Line 1: Length or precision specification 0 is invalid.
However, this is a generic error from the server and may not convey why the empty vector value is invalid.
As confirmed with the server team,  server expects at least 12 bytes ( 8 byte header + 4byte for single float element) for a 1 element vector column.
I will remove this check and let server error out.

doc/snippets/Microsoft.Data.SqlClient/SqlDataReader.xml
```diff
@@ -959,6 +959,15 @@ The <xref:Microsoft.Data.SqlClient.SqlDataReader.GetSchemaTable%2A> method retur
       <summary>Gets the value of the specified column as a <see cref="T:Microsoft.Data.SqlTypes.SqlJson"/>.</summary>
       <returns>A <see cref="T:Microsoft.Data.SqlTypes.SqlJson"/> object representing the column at the given ordinal.</returns>
     </GetSqlJson>
+    <GetSqlVectorFloat32>
+      <param name="i"></param>
+      <summary>
+        Gets the value of the specified column as a <see cref="T:Microsoft.Data.SqlTypes.SqlVectorFloat32"/>.
+      </summary>
+      <returns>
+        A <see cref="T:Microsoft.Data.SqlTypes.SqlVectorFloat32"/> object representing the column at the given ordinal.
```
paulmedynski, line 973: Does this method throw anything?

doc/snippets/Microsoft.Data.SqlTypes/SqlVectorFloat32.xml
```diff
@@ -0,0 +1,44 @@
+<?xml version="1.0"?>
+<docs>
+  <members name="SqlVectorFloat32">
+    <SqlVectorFloat32>
+      <summary>Represents the vector of float datatype in SQL Server.</summary>
```
paulmedynski: Represents the 32-bit float vector datatype in SQL Server.

doc/snippets/Microsoft.Data.SqlTypes/SqlVectorFloat32.xml
```diff
@@ -0,0 +1,44 @@
...
+    </SqlVectorFloat32>
+    <ctor1>
+      <param name="length"></param>
+      <summary>
+        Takes a <see cref="int"/> as input and represents null value for vector column of float element type.
```
paulmedynski: Constructs a null vector of the given length.  SQL Server requires vector arguments to specify their length even when null.

paulmedynski: I believe this constructor throws.

doc/snippets/Microsoft.Data.SqlTypes/SqlVectorFloat32.xml
```diff
@@ -0,0 +1,44 @@
...
+      </summary>
+    </ctor1>
+    <ctor2>
+      <param name="values"></param>
+      <summary>
+        Takes an array of float values as input and initializes a new instance of the SqlVectorFloat32 class.
```
paulmedynski: Constructs a vector with the given values.

Does this throw?

doc/snippets/Microsoft.Data.SqlTypes/SqlVectorFloat32.xml
```diff
@@ -0,0 +1,44 @@
...
+      </summary>
+    </ctor2>
+    <IsNull>
+      <inheritdoc/>
+    </IsNull>
+    <Null>
+        <inheritdoc/>
```
paulmedynski: Where does this inherit docs from?  Also indent is off.

doc/snippets/Microsoft.Data.SqlTypes/SqlVectorFloat32.xml
```diff
@@ -0,0 +1,44 @@
...
+    </Null>
+    <Length>
+      <summary>
+        Indicates the count of the elements in the float vector.
```
paulmedynski: Returns the number of elements in the vector.

doc/snippets/Microsoft.Data.SqlTypes/SqlVectorFloat32.xml
```diff
@@ -0,0 +1,44 @@
...
+      </summary>
+    </Length>
+    <Size>
+      <summary>
+        Indicates the size in bytes for a SqlVectorFloat32 value for a vector column of <see cref="Length"/> elements of <see cref="T:System.Single"/> type.
```
paulmedynski: Returns the number of bytes required to represent this vector when communicating with SQL Server.

doc/snippets/Microsoft.Data.SqlTypes/SqlVectorFloat32.xml
```diff
@@ -0,0 +1,44 @@
...
+      </summary>
+    </Size>
+    <Values>
+      <summary>Returns the internal float array representing the float vector.</summary>
```
paulmedynski: Returns the vector values as a memory region.  No copies are made.

And for the ToArray() method:

Returns the vector values as an array of floats.  One copy is made the first time you call this method per instance.

doc/snippets/Microsoft.Data.SqlTypes/SqlVectorFloat32.xml
```diff
@@ -0,0 +1,44 @@
...
+      <returns>An array of <see cref="T:System.Single"/> values.</returns>
+    </Values>
+    <ToString>
+      <summary>Returns a JSON string representing the internal float array representing the float vector.</summary>
```
paulmedynski: Returns a JSON string representation of the vector.  A new string is generated each time you call this method.

doc/snippets/Microsoft.Data.SqlClient/SqlDataReader.xml
```diff
@@ -959,6 +959,15 @@ The <xref:Microsoft.Data.SqlClient.SqlDataReader.GetSchemaTable%2A> method retur
       <summary>Gets the value of the specified column as a <see cref="T:Microsoft.Data.SqlTypes.SqlJson"/>.</summary>
       <returns>A <see cref="T:Microsoft.Data.SqlTypes.SqlJson"/> object representing the column at the given ordinal.</returns>
     </GetSqlJson>
+    <GetSqlVectorFloat32>
+      <param name="i"></param>
```
paulmedynski, line 968: Can we choose a better argument name, and document it?

src/Microsoft.Data.SqlClient.sln
```diff
@@ -172,6 +172,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Microsoft.Data.SqlTypes", "
 	ProjectSection(SolutionItems) = preProject
 		..\doc\snippets\Microsoft.Data.SqlTypes\SqlFileStream.xml = ..\doc\snippets\Microsoft.Data.SqlTypes\SqlFileStream.xml
 		..\doc\snippets\Microsoft.Data.SqlTypes\SqlJson.xml = ..\doc\snippets\Microsoft.Data.SqlTypes\SqlJson.xml
+		..\doc\snippets\Microsoft.Data.SqlTypes\SqlFloatVector.xml = ..\doc\snippets\Microsoft.Data.SqlTypes\SqlFloatVector.xml
```
paulmedynski: Does this file exist?

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -117,6 +119,27 @@ public SqlJson(System.Text.Json.JsonDocument jsonDoc) { }
         /// <include file='../../../../doc/snippets/Microsoft.Data.SqlTypes/SqlJson.xml' path='docs/members[@name="SqlJson"]/Value/*' />
         public string Value { get { throw null; } }
     }
+
+    /// <include file='../../../../doc/snippets/Microsoft.Data.SqlTypes/SqlVectorFloat32.xml' path='docs/members[@name="SqlVectorFloat32"]/SqlVectorFloat32/*' />
+    public class SqlVectorFloat32 : System.Data.SqlTypes.INullable
+    {
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlTypes/SqlVectorFloat32.xml' path='docs/members[@name="SqlVectorFloat32"]/ctor1/*' />
+        public SqlVectorFloat32(int length) { }
```
paulmedynski, line 129: Should the constructors throw null as well?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
```diff
@@ -673,12 +677,16 @@ private string AnalyzeTargetAndCreateUpdateBulkCommand(BulkCopySimpleResultSet i
...
+                                    case TdsEnums.SQLVECTOR:
+                                        size = MetaType.GetVectorElementCount(metadata.length, metadata.scale);
+                                        break;
                                     default:
                                         break;
                                 }
                                 updateBulkCommandText.AppendFormat((IFormatProvider)null, "({0})", size);
                             }
-                            else if (metadata.metaType.IsPlp && metadata.metaType.SqlDbType != SqlDbType.Xml && metadata.metaType.SqlDbType != SqlDbTypeExtensions.Json)
+                            else if (metadata.metaType.IsPlp && metadata.metaType.SqlDbType != SqlDbType.Xml
```
paulmedynski: This might be a good candidate for Ben's fancy new is-or syntax:
```
else if (
    element.metaType.IsPlp &&
    (! metadata.metaType.SqlDbType is
        SqlDbType.Xml or
        SqlDbTypeExtensions.Json or
        SqlDbTypeExtensions.Vector))
{
    ...
}
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
```diff
@@ -6435,6 +6435,13 @@ internal string BuildParamList(TdsParser parser, SqlParameterCollection paramete
                     paramList.Append(scale);
                     paramList.Append(')');
                 }
+                else if (mt.SqlDbType == SqlDbTypeExtensions.Vector)
+                {
+                    var sqlVectorProps = (ISqlVector)sqlParam.Value;
```
paulmedynski, line 6443: It would be helpful to document that this cast will throw if someone specified Vector but then provided something else, and that throwing here is expected and OK.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -6442,6 +6470,18 @@ internal TdsOperationStatus TryReadSqlValue(SqlBuffer value,
                     }
                     break;
 
+                case TdsEnums.SQLVECTOR:
+                    result = TryReadByteArrayWithContinue(stateObj, length, out b);
+                    if (result != TdsOperationStatus.Done)
+                    {
+                        return result;
+                    }
+                    value.SqlBinary = SqlBinary.WrapBytes(b);
```
paulmedynski, line 6484: Comments explaining what is happening here would be helpful.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -9608,7 +9696,10 @@ private Task TDSExecuteRPCAddParameter(TdsParserStateObject stateObj, SqlParamet
             if (param.Direction == ParameterDirection.Output)
             {
                 isSqlVal = param.ParameterIsSqlType;  // We have to forward the TYPE info, we need to know what type we are returning.  Once we null the parameter we will no longer be able to distinguish what type were seeing.
-                param.Value = null;
+                if (mt.SqlDbType != SqlDbTypeExtensions.Vector)
```
paulmedynski, line 9669: Why is Vector special here?  Comment would help.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -9941,6 +10032,12 @@ private Task TDSExecuteRPCAddParameter(TdsParserStateObject stateObj, SqlParamet
                             maxsize = 1;
                     }
 
+                    if (mt.SqlDbType == SqlDbTypeExtensions.Vector)
+                    {
+                        var sqlVectorProps = ((ISqlVector)param.Value);
+                        maxsize = TdsEnums.VECTOR_HEADER_SIZE + sqlVectorProps.Length * sqlVectorProps.ElementSize;
```
paulmedynski: I feel like this calculation belongs in ISqlVector implementations.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -9941,6 +10032,12 @@ private Task TDSExecuteRPCAddParameter(TdsParserStateObject stateObj, SqlParamet
                             maxsize = 1;
                     }
 
+                    if (mt.SqlDbType == SqlDbTypeExtensions.Vector)
+                    {
+                        var sqlVectorProps = ((ISqlVector)param.Value);
```
paulmedynski, line 10009: Again, comment that this may throw, and that's OK.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ISqlVector.cs
```diff
@@ -0,0 +1,28 @@
...
+    /// </summary>
+    internal interface ISqlVector
+    {
+        /// <summary>
+        /// Gets the number of elements in the vector.
+        /// </summary>
+        int Length { get; }
+
+        /// <summary>
+        /// Gets the type of the elements in vector.
```
paulmedynski: ... as a TDS Vector Header `DimensionType` value.

Can you link to the TDS spec here, and maybe note which sections define these values?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -10042,7 +10142,8 @@ private Task TDSExecuteRPCAddParameter(TdsParserStateObject stateObj, SqlParamet
                 {
                     // for codePageEncoded types, WriteValue simply expects the number of characters
                     // For plp types, we also need the encoded byte size
-                    writeParamTask = WriteValue(value, mt, isParameterEncrypted ? (byte)0 : param.GetActualScale(), actualSize, codePageByteSize, isParameterEncrypted ? 0 : param.Offset, stateObj, isParameterEncrypted ? 0 : param.Size, isDataFeed);
+                    byte writeScale = mt.SqlDbType == SqlDbTypeExtensions.Vector ? ((ISqlVector)param.Value).ElementType : param.GetActualScale();
```
paulmedynski, line 10119: Comment describing why Vector is special here.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBuffer.cs
```diff
@@ -76,6 +77,12 @@ internal struct DateTimeOffsetInfo
             internal short _offset;
         }
 
+        internal struct VectorInfo
+        {
+            internal int _vectorElementCount;
```
paulmedynski: Do we need `vector` in these field names?  The struct is already called `VectorInfo`.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBuffer.cs
```diff
@@ -133,6 +142,15 @@ private SqlBuffer(SqlBuffer value)
 
         internal StorageType VariantInternalStorageType => _type;
 
+        internal Storage GetVectorInfo()
+        {
+            if (StorageType.Vector == _type)
```
paulmedynski: `_type == StorageType.Vector`

Here and several other places below.  Feel free to swap other adjacent checks as well.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBuffer.cs
```diff
@@ -486,7 +507,17 @@ internal string String
             get
             {
                 ThrowIfNull();
-
+                if (StorageType.Vector == _type)
+                {
+                    byte elementType = _value._vectorInfo._vectorElementType;
```
paulmedynski: Maybe cast this to SqlVectorElementType instead of casting the enum value to byte below.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBuffer.cs
```diff
@@ -941,6 +980,8 @@ internal SqlString SqlString
 
         internal SqlJson SqlJson => (StorageType.Json == _type) ? (IsNull ? SqlTypes.SqlJson.Null : new SqlJson((string)_object)) : (SqlJson)SqlValue;
 
+        internal SqlVectorFloat32 SqlVectorFloat32 => (StorageType.Vector == _type) ? (IsNull ? new SqlVectorFloat32(_value._vectorInfo._vectorElementCount) : new SqlVectorFloat32(SqlBinary.Value)) : (SqlVectorFloat32)SqlValue;
```
paulmedynski: Improve formatting for readability:
```
internal SqlVectorFloat32 SqlVectorFloat32 =>
    StorageType.Vector == _type
    ? (IsNull
        ? new SqlVectorFloat32(_value._vectorInfo._vectorElementCount)
        : new SqlVectorFloat32(SqlBinary.Value))
    : (SqlVectorFloat32)SqlValue;
```

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBuffer.cs
```diff
@@ -975,6 +1016,15 @@ internal object SqlValue
                         return SqlString;
                     case StorageType.Json:
                         return SqlJson;
+                    case StorageType.Vector:
+                        byte elementType = _value._vectorInfo._vectorElementType;
```
paulmedynski: Same here, cast to SqlVectorElementType rather than byte below.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBuffer.cs
```diff
@@ -1238,7 +1292,7 @@ internal void Clear()
             _object = null;
         }
 
-        #if NETFRAMEWORK
+#if NETFRAMEWORK
```
paulmedynski: Undo outdent.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlEnums.cs
```diff
@@ -63,6 +63,15 @@ internal sealed class MetaType
         internal readonly bool Is90Supported;
         internal readonly bool Is100Supported;
 
+        // SqlVector Element Types
+        internal enum SqlVectorElementType : byte
+        {
+            Float32 = 0
+            // Future values:
```
paulmedynski: Remove these - we will define them once they actually exist.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlEnums.cs
```diff
@@ -1027,6 +1045,21 @@ internal static int GetTimeSizeFromScale(byte scale)
             return 5;
         }
 
+        internal static int GetVectorElementSize(byte type)
+        {
+            switch (type)
+            {
+                case 0: return sizeof(float);
```
paulmedynski, line 1049: I think SqlVectorFloat32 should be the owner of this value.  IOW, the only place we should see `sizeof(float)` is somewhere inside SqlVectorFloat32.

This actually makes SqlVector<> an attractive place for this to live.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlEnums.cs
```diff
@@ -1027,6 +1045,21 @@ internal static int GetTimeSizeFromScale(byte scale)
...
+                default:
+                    throw SQL.VectorTypeNotSupported(type.ToString());
+            }
+        }
+
+        internal static int GetVectorElementCount(int size, byte elementType)
+        {
+            return (size - TdsEnums.VECTOR_HEADER_SIZE) / GetVectorElementSize(elementType);
```
paulmedynski, line 1057: I think this calculation should live in SqlVectorFloat32 somewhere too.  (Or SqlVector<>, since it doesn't vary by underlying type.)

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlParameter.cs
```diff
@@ -1896,6 +1926,11 @@ private MetaType GetMetaTypeOnly()
         {
             if (_metaType != null)
             {
+                if (_metaType.SqlDbType == SqlDbTypeExtensions.Vector && (_direction == ParameterDirection.Input) && (_value == null || _value == DBNull.Value))
```
paulmedynski: Unnecessary `()` around second logical AND segment.  Formatting would help too:
```
if (_metaType.SqlDbType == SqlDbTypeExtensions.Vector &&
    _direction == ParameterDirection.Input &&
    (_value == null || _value == DBNull.Value))
```

Here and below.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlParameter.cs
```diff
@@ -2149,6 +2190,10 @@ private int ValueSize(object value)
                 }
                 return sqlString.Value.Length;
             }
+            if (value is ISqlVector sqlVector)
+            {
+                return TdsEnums.VECTOR_HEADER_SIZE + (sqlVector.Length * sqlVector.ElementSize);
```
paulmedynski: Can this calculation live in ISqlVector or its implementors?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlParameter.cs
```diff
@@ -2320,6 +2365,21 @@ internal static object CoerceValue(object value, MetaType destinationType, out b
                         value = ((TimeOnly)value).ToTimeSpan();
                     }
 #endif
+                    else if (currentType == typeof(SqlVectorFloat32))
+                    {
+                        value = (value as ISqlVector).VectorPayload;
```
paulmedynski: Did you mean to cast here?  `as` will just return null if value isn't an ISqlVector and then you'll get a different exception.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlParameter.cs
```diff
@@ -2320,6 +2365,21 @@ internal static object CoerceValue(object value, MetaType destinationType, out b
...
+                    }
+                    else if (currentType == typeof(string) && destinationType.SqlDbType == SqlDbTypeExtensions.Vector)
+                    {
+                        try
+                        {
+                            value = (new SqlVectorFloat32(JsonSerializer.Deserialize<float[]>(value as string)) as ISqlVector).VectorPayload;
```
paulmedynski, line 2380: Why are we deserializing JSON to get a vector value?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlParameter.cs
```diff
@@ -2320,6 +2365,21 @@ internal static object CoerceValue(object value, MetaType destinationType, out b
...
+                        }
+                        catch (Exception ex)
```
paulmedynski: Catch only those exceptions documented to be thrown.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlDbTypeExtensions.cs
```diff
@@ -14,6 +14,12 @@ public static class SqlDbTypeExtensions
         public const SqlDbType Json = SqlDbType.Json;
 #else
         public const SqlDbType Json = (SqlDbType)35;
+#endif
+        /// <include file='../../../../doc/snippets/Microsoft.Data/SqlDbTypeExtensions.xml' path='docs/members[@name="SqlDbTypeExtensions"]/SqlVector[@name="default"]' />
+#if NET10_0
```
paulmedynski: `#if NET10_0_OR_GREATER` ?

paulmedynski (changes requested):
I have reviewed all of the implementation except SqlVectorFloat32.  I'll wait to review that once you make the changes we discussed on Friday.

I will look at the tests after that.

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
cheenamalhotra: Not necessary. The goal for ref assembly is to provide minimal API implementation, and as long as file size can be minimum, we are okay to keep the ctors empty.

doc/snippets/Microsoft.Data.SqlClient/SqlDataReader.xml
```diff
@@ -959,6 +959,15 @@ The <xref:Microsoft.Data.SqlClient.SqlDataReader.GetSchemaTable%2A> method retur
...
+      <summary>
+        Gets the value of the specified column as a <see cref="T:Microsoft.Data.SqlTypes.SqlVectorFloat32"/>.
+      </summary>
+      <returns>
+        A <see cref="T:Microsoft.Data.SqlTypes.SqlVectorFloat32"/> object representing the column at the given ordinal.
+      </returns>
+    </GetSqlVectorFloat32>
```
cheenamalhotra, line 987: Also include remarks as in other Getters
+include remarks in Json as well.

doc/snippets/Microsoft.Data.SqlClient/SqlDataReader.xml
cheenamalhotra: We have this for all other APIs too.. 🙈

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -5,6 +5,8 @@
 // NOTE: The current Microsoft.VSDesigner editor attributes are implemented for System.Data.SqlClient, and are not publicly available.
 // New attributes that are designed to work with Microsoft.Data.SqlClient and are publicly documented should be included in future.
 
+using System;
```
cheenamalhotra: We avoid importing globally in ref files.

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -117,6 +119,27 @@ public SqlJson(System.Text.Json.JsonDocument jsonDoc) { }
...
         public string Value { get { throw null; } }
     }
+
+    /// <include file='../../../../doc/snippets/Microsoft.Data.SqlTypes/SqlVectorFloat32.xml' path='docs/members[@name="SqlVectorFloat32"]/SqlVectorFloat32/*' />
+    public class SqlVectorFloat32 : System.Data.SqlTypes.INullable
+    {
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlTypes/SqlVectorFloat32.xml' path='docs/members[@name="SqlVectorFloat32"]/ctor1/*' />
+        public SqlVectorFloat32(int length) { }
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlTypes/SqlVectorFloat32.xml' path='docs/members[@name="SqlVectorFloat32"]/ctor2/*' />
+        public SqlVectorFloat32(ReadOnlyMemory<float> values) { }
```
cheenamalhotra: Avoid global imports
```suggestion
        public SqlVectorFloat32(System.ReadOnlyMemory<float> values) { }
```

src/Microsoft.Data.SqlClient/netfx/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -5,6 +5,8 @@
 // NOTE: The current Microsoft.VSDesigner editor attributes are implemented for System.Data.SqlClient, and are not publicly available.
 // New attributes that are designed to work with Microsoft.Data.SqlClient and are publicly documented should be included in future.
 
+using System;
```
cheenamalhotra: Similarly here, avoid global imports.

doc/snippets/Microsoft.Data.SqlClient/SqlDataReader.xml
apoorvdeshmukh: Yes. Similar as this [one](https://learn.microsoft.com/en-us/dotnet/api/microsoft.data.sqlclient.sqldatareader.getsqlxml?view=sqlclient-dotnet-core-6.0#exceptions)
Will add these.

doc/snippets/Microsoft.Data.SqlTypes/SqlVectorFloat32.xml
apoorvdeshmukh: This will not throw as we have decided to allow users to pass empty array.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlParameter.cs
apoorvdeshmukh: This is for one of the backward compatibility scenarios, in which a bulk copy operation may be performed from a varchar(max) column to vector column.  We get vector data as a JSON string in that case, and we then create vector value out of it.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlTypes/SqlVectorFloat32.cs
apoorvdeshmukh: These are explained further when `_rawBytes` is initialized further in `InitializeVectorBytes()`

doc/snippets/Microsoft.Data.SqlTypes/SqlVectorFloat32.xml
apoorvdeshmukh: Added note for this one that  'This is equivalent to the C# <c>null</c> value.'
We have added this for making SqlVectorFloat32 compatible with DataTable.

doc/snippets/Microsoft.Data.SqlTypes/SqlVectorFloat32.xml
apoorvdeshmukh: I am excluding the copy part here. We will refactor TdsParser later which will ensure no copies are made.

doc/snippets/Microsoft.Data.SqlClient/SqlDataReader.xml
apoorvdeshmukh: I will update this argument for all other APIs as separate PR later. For now, will keep it consistent with other APIs.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
apoorvdeshmukh: The validate function for SqlParameter usually gets called much earlier before we reach this point and will throw if any unexpected value is assigned for SqlDbType Vector. Added comment documenting this.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlEnums.cs
apoorvdeshmukh: I agree. If we decide to go ahead with SqlVector<> approach I can keep it there but until then I can move it to VectorUtils.cs(new file) or keep it here for now as these are applicable for all vectors of unmanaged types.

src/Microsoft.Data.SqlClient.sln
apoorvdeshmukh: Missed renaming it here. Corrected it.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
edwardneal, line 1: This needs to be removed - an earlier PR merged it.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlDataReader.cs
```diff
@@ -3325,6 +3358,13 @@ private T GetFieldValueFromSqlBufferInternal<T>(SqlBuffer data, _SqlMetaData met
                 }
                 else
                 {
+                    if (typeof(T) == typeof(string) && metaData.metaType.SqlDbType == SqlDbTypeExtensions.Vector)
```
edwardneal, line 3361: Nit: the JSON support was done slightly differently. That pattern seems pretty reasonable.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlTypes/SqlVectorFloat32.cs
```diff
@@ -0,0 +1,175 @@
...
+            // | Number of Dimensions   | 2               | NN                    | Number of vector elements                                   |
+            // | Dimension Type         | 1               | 0x00                     | Element type indicator (e.g. 0x00 for float32)           |
+            // | Reserved               | 3               | 0x00 0x00 0x00        | Reserved for future use                                     |
+            // | Stream of Values       | NN * sizeof(T)  | [float bytes...]      | Raw bytes for vector elements                               |
+            // +------------------------+-----------------+----------------------+--------------------------------------------------------------+
+
+            _rawBytes[0] = VecHeaderMagicNo;
+            _rawBytes[1] = VecVersionNo;
+            _rawBytes[2] = (byte)(_elementCount & 0xFF);
+            _rawBytes[3] = (byte)((_elementCount >> 8) & 0xFF);
```
edwardneal: We can use BinaryPrimitives.WriteUInt16LittleEndian for this sort of work, rather than manually shifting bits around (here and elsewhere.)

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlTypes/SqlVectorFloat32.cs
```diff
@@ -0,0 +1,175 @@
...
+            }
+            else
+            {
+                Buffer.BlockCopy(values.ToArray(), 0, _rawBytes, TdsEnums.VECTOR_HEADER_SIZE, values.Length * _elementSize);
+            }
+#else
+            // Fast span-based copy
+            var byteSpan = MemoryMarshal.AsBytes(values.Span);
+            byteSpan.CopyTo(_rawBytes.AsSpan(TdsEnums.VECTOR_HEADER_SIZE));
+#endif
```
edwardneal: In other parts of the codebase, we're sensitive to endianness and specify it explicitly when reading and writing structures. `MemoryMarshal.AsBytes` simply uses the current CPU architecture.

If this is a problem, I'd suggest iterating over `values.Span` and calling `BinaryPrimitives.WriteSingleLittleEndian` instead. This also removes the conditional compilation and the slightly unusual call to `MemoryMarshal.TryGetArray`.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/VectorTest/NativeVectorFloat32Tests.cs
```diff
@@ -0,0 +1,593 @@
...
+            yield return new object[] { 4, SqlVectorFloat32.Null, Array.Empty<float>(), sizeInbytes, vectorColumnLength };
+        }
+    }
+
+    public sealed class NativeVectorFloat32Tests : IDisposable
+    {
+        private readonly ITestOutputHelper _output;
+        private static readonly string s_connectionString = ManualTesting.Tests.DataTestUtility.TCPConnectionString;
+        private static readonly string s_tableName = DataTestUtility.GetUniqueName("VectorTestTable");
+        private static readonly string s_bulkCopySrcTableName = DataTestUtility.GetUniqueName("VectorBulkCopyTestTable");
```
edwardneal, line 56: Do we need to generate this (and the other unique names) on a per-test basis?

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/VectorTest/NativeVectorFloat32Tests.cs
```diff
@@ -0,0 +1,593 @@
...
+using Microsoft.Data.SqlTypes;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.Data.SqlClient.ManualTesting.Tests.SQL.VectorTest
+{
+    public static class VectorFloat32TestData
+    {
+        public const int VectorHeaderSize = 8;
+        public static float[] testData = new float[] { 1.1f, 2.2f, 3.3f };
```
edwardneal, line 16: How does this behave with unusual values?
* `float.MinValue`: a value which is within the range of a .NET float, but outside the range of a SQL Server real
* `float.NegativeZero`: is the sign preserved?
* `float.NaN`: does this throw?
* `1.01`: this can't be precisely rendered as an IEE 754 float - how does it roundtrip?

I'm fairly sure all of these values will fail. If they do, perhaps we should filter some of these values in `SqlVectorFloat32`'s constructor.

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -117,6 +119,29 @@ public SqlJson(System.Text.Json.JsonDocument jsonDoc) { }
         /// <include file='../../../../doc/snippets/Microsoft.Data.SqlTypes/SqlJson.xml' path='docs/members[@name="SqlJson"]/Value/*' />
         public string Value { get { throw null; } }
     }
+
+    /// <include file='../../../../doc/snippets/Microsoft.Data.SqlTypes/SqlVectorFloat32.xml' path='docs/members[@name="SqlVectorFloat32"]/SqlVectorFloat32/*' />
+    public class SqlVectorFloat32 : System.Data.SqlTypes.INullable
```
edwardneal:
```suggestion
    public sealed class SqlVectorFloat32 : System.Data.SqlTypes.INullable
```
Here and in netfx

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -1368,6 +1393,8 @@ public override void Close() { }
         public virtual object GetSqlValue(int i) { throw null; }
         /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlDataReader.xml' path='docs/members[@name="SqlDataReader"]/GetSqlValues/*'/>
         public virtual int GetSqlValues(object[] values) { throw null; }
+        /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlDataReader.xml' path='docs/members[@name="SqlDataReader"]/GetSqlVectorFloat32/*'/>
+        public virtual object GetSqlVectorFloat32(int i) { throw null; }
```
edwardneal:
```suggestion
        public virtual SqlVectorFloat32 GetSqlVectorFloat32(int i) { throw null; }
```

Here and in netfx.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlDataReader.cs
```diff
@@ -3187,6 +3211,15 @@ private T GetFieldValueFromSqlBufferInternal<T>(SqlBuffer data, _SqlMetaData met
                 return (T)(object)data.TimeOnly;
             }
 #endif
+            else if (typeof(T) == typeof(SqlVectorFloat32))
```
edwardneal, line 3214: This might also need documentation updates for SqlDataReader.GetFieldValues.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlEnums.cs
```diff
@@ -978,6 +991,8 @@ internal static string GetStringFromXml(XmlReader xmlreader)
 
         internal static readonly MetaType s_MetaJson = new(255, 255, -1, false, true, true, TdsEnums.SQLJSON, TdsEnums.SQLJSON, MetaTypeName.JSON, typeof(string), typeof(string), SqlDbTypeExtensions.Json, DbType.String, 0);
 
+        internal static readonly MetaType s_MetaVector = new(255, 255, -1, false, false, false, TdsEnums.SQLVECTOR, TdsEnums.SQLVECTOR, MetaTypeName.VECTOR, typeof(byte[]), typeof(SqlBinary), SqlDbTypeExtensions.Vector, DbType.Binary, 2);
```
edwardneal, line 994: Microsoft.Data.SqlClient.SqlMetaData.xml contains a list of SQL Server data types, and also needs to be updated.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBuffer.cs
edwardneal, line 1: It's worth noting that there's also `SqlDataRecord` support to consider for this.

apoorvdeshmukh: Thanks @edwardneal  for valuable feedback. We have not considered the TVP and SqlDataRecord support for the initial vector datatype support. I will, however, track this under #3440.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlEnums.cs
apoorvdeshmukh: To be addressed as part of #3440

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/VectorTest/NativeVectorFloat32Tests.cs
apoorvdeshmukh: Sql server does not complain for these values except for float.NaN which it considers as invalid float value for vector element.

edwardneal: Thanks for checking here. Do they roundtrip (both as JSON and as elements in a strongly-typed `SqlVectorFloat32`)?

The example of float.NaN suggests that we should validate the contents of the ReadOnlyMemory<float> that users can pass to the new type.

apoorvdeshmukh: They do roundtrip as both JSON and as SqlVectorFloat32.
As JSON for `float.NegativeZero` it does not preserve the sign and we have noticed some other strange issues when exchanging the data as JSON but that is expected and also why the strongly typed SqlVectorFloat32 is encouraged. 
We are maintaining the support for exchanging vector data for backward compatibility purpose since server supported vectors initially as JSON string.
For errors that arise from server due to user passing incompatible data, driver generally doesn't enforce validations and lets the server error out.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/VectorTest/NativeVectorFloat32Tests.cs
apoorvdeshmukh: Following the pattern from other tests.  Just to ensure that tests don't interfere with each other.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/VectorTest/NativeVectorFloat32Tests.cs
edwardneal: Interesting. I definitely agree on the need for the strongly-typed vector object. Could you add those unusual values to the test case please?

If you think it's worth it, we could describe the behaviour when roundtripping float.NegativeZero to/from JSON in a test too.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlDataReader.cs
apoorvdeshmukh: For JSON we had a CLR type `string` for which we would return String from SqlBuffer.
For vector datatype we are making SqlVectorFloat32 as the preferred type to interact with vector data.

cheenamalhotra (approved):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBuffer.cs
```diff
@@ -1190,6 +1249,8 @@ internal Type GetTypeFromStorageType(bool isSqlType)
                         return typeof(DateTimeOffset);
                     case StorageType.Json:
                         return typeof(string);
+                    case StorageType.Vector:
+                        return typeof(byte[]);
```
rhuijben, line 1253: Note from the future:
```c#
return typeof(SqlVectorFloat32);
```

PR #3435, opened 2025-06-23
Cleanup | SqlConnectionFactory <- DbConnectionFactory

benrr101:
## Description
This is the first PR in a bit of a fun experiment where we see if we can seriously clean up the codebase. According to @saurabh500 the codebase we've inherited dates back to the days when it was part of System.Data that not only contained the SqlClient but OdbcClient and even OracleClient. This is why there's Db* classes buried in the codebase that are passed around instead of the Sql* classes. In the best cases, these base implementations are fully abstract or just provide fallback implementations that throw. In the worst cases, they create confusing loops of implementations. And regardless, we are constantly passing them around as Db* and doing a lot of unnecessary checks and casts to the Sql* classes. These seemingly redundant base classes have a lot of unnecessary code and make the codebase as a whole more confusing. If we can do away with them, it would be idea. So, as an experiment, I decided SqlConnectionFactory might be a good place to see if we can simply merge the Db* class into the Sql* class.

As with most of my other merge project PRs, this one can be digested commit by commit (although be forewarned these ones are a bit stranger than usual due to protection levels). Important callouts have been added via my own review of the commit. Keep in mind although this could easily turn into a gigantic PR that removes all Db* classes, I've tried my best to be surgical and only eliminate the DbConnectionFactory class.

## Issues
N/A

## Testing
I'm still having trouble running test projects locally, so I have attempted to run them, but I can't guarantee I caught all the issues. CI will hopefully catch the remainder.


src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/AdapterUtil.cs
```diff
@@ -130,7 +130,10 @@ internal static Exception ExceptionWithStackTrace(Exception e)
             }
         }
 
-        internal static Timer UnsafeCreateTimer(TimerCallback callback, object state, int dueTime, int period)
+        internal static Timer UnsafeCreateTimer(TimerCallback callback, object state, int dueTime, int period) =>
+            UnsafeCreateTimer(callback, state, TimeSpan.FromSeconds(dueTime), TimeSpan.FromSeconds(period));
```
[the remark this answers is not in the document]
benrr101: Yep, you were right, thanks

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -3049,10 +3049,12 @@ internal bool ThreadHasParserLockForClose
...
-        internal override bool TryReplaceConnection(DbConnection outerConnection, DbConnectionFactory connectionFactory, TaskCompletionSource<DbConnectionInternal> retry, DbConnectionOptions userOptions)
-        {
-            return base.TryOpenConnectionInternal(outerConnection, connectionFactory, retry, userOptions);
-        }
+        internal override bool TryReplaceConnection(
+            DbConnection outerConnection,
+            SqlConnectionFactory connectionFactory,
+            TaskCompletionSource<DbConnectionInternal> retry,
+            DbConnectionOptions userOptions) =>
+            TryOpenConnectionInternal(outerConnection, connectionFactory, retry, userOptions);
```
paulmedynski: Hmm... this syntax has disguised the function body as a function agrument.  Would this be more readable as a classic {} block?  Same in the netfx version.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbConnectionClosed.cs
```diff
@@ -36,13 +37,22 @@ internal override void CloseConnection(DbConnection owningObject, DbConnectionFa
...
         public override void EnlistTransaction(System.Transactions.Transaction transaction) => throw ADP.ClosedConnectionError();
 
-        protected internal override DataTable GetSchema(DbConnectionFactory factory, DbConnectionPoolGroup poolGroup, DbConnection outerConnection, string collectionName, string[] restrictions)
+        protected internal override DataTable GetSchema(
+            SqlConnectionFactory factory,
+            DbConnectionPoolGroup poolGroup,
+            DbConnection outerConnection,
+            string collectionName,
+            string[] restrictions)
             => throw ADP.ClosedConnectionError();
```
paulmedynski: Do we put the `=>` on the next line, or the previous line?  Or would all of these methods with multi-line argument lists look better as {} blocks?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/IDbConnectionPool.cs
paulmedynski, line 1: Maybe this class should just be called `IConnectionPool`?  Question for @mdaigle

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
benrr101: Yeah, it doesn't stand out very well. I'm fine using {} in this case

src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbConnectionClosed.cs
benrr101: I don't know if there's a standard... I put them on the previous line, then put the expression on the same line for properties and next line for methods.

I can change to {} though to match the previous comment.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/IDbConnectionPool.cs
mdaigle: I'm fine with that :)

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnectionHelper.cs
```diff
@@ -157,7 +157,7 @@ override protected DbCommand CreateDbCommand()
             using (TryEventScope.Create("<prov.DbConnectionHelper.CreateDbCommand|API> {0}", ObjectID))
             {
                 DbCommand command = null;
-                DbProviderFactory providerFactory = ConnectionFactory.ProviderFactory;
+                DbProviderFactory providerFactory = SqlConnectionFactory.ProviderFactory;
```
mdaigle: ProviderFactory always calls through to the singleton

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionFactory.cs
```diff
@@ -18,26 +22,562 @@
 
 namespace Microsoft.Data.SqlClient
 {
-    sealed internal class SqlConnectionFactory : DbConnectionFactory
```
mdaigle, line 21: I did find it useful to add stub implementations for DbConnectionFactory when writing unit tests for the connection pool. It's the best spot to stub opening a new connection. If this class becomes sealed, that pattern will break (can't even do inheritance and override). I'd consider adding an interface (It's just a bummer that this class has such an unnecessarily big API at the moment.)

paulmedynski (approved):

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnectionHelper.cs
benrr101: Makes sense to remove - my guess is it was some kinda thing where everything comes from the connection factory, implying that there's more than one kind of connection factory.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionFactory.cs
benrr101: I agree. Singletons/static classes are always a kind of code smell - totally valid to have, but they should used as interfaces so they can be replaced in test scenarios (what I think of as a facade pattern). 

I kinda don't want to try and implement an entire facade pattern as part of this change. I'll add a note to facade it. But what if I unseal the class for now?

paulmedynski (dismissed):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionFactory.cs
mdaigle: that works for me

paulmedynski (approved):

PR #3438, opened 2025-06-24
Cleanup | Remove SmiEventSink_Default

edwardneal:
## Description

This follows up #3431 - sorry if I'm treading on your toes here @benrr101. It's a fairly bulky set of changes, but most of this bulk comes from removing the methods in ValueUtilsSmi which accepted and passed `SmiEventSink_Default` parameters. The majority of the work is handled in the first three methods.

My process here is clearest when considered in the context of [SmiEventSink_Default.cs](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SmiEventSink_Default.cs):
1. `ProcessMessagesAndThrow` only does work if called when `HasMessages` is true.
2. `HasMessages` is only true if `_errors` or `_warnings` is non-null.
3. Both variables are private in scope. They are only ever assigned to in `ProcessMessages`, and in both cases they're set to null.
4. Based on this, `HasMessages` is always false, and `ProcessMessagesAndThrow` never does anything. It can be removed.
5. `ProcessMessagesAndThrow` was the only member (besides the constructor) accessed on the class.

I thus removed the entire class and every reference to it. This has knock-on effects on SqlDataReader's target-specific files, (which can be merged and removed) but those can be cleaned up later.

## Testing

Unit tests and manual tests pass, although I'd appreciate a CI run.


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SqlDataRecord.cs
```diff
@@ -377,13 +376,11 @@ public SqlDataRecord(params SqlMetaData[] metaData)
                 _columnSmiMetaData[i] = MetaDataUtilsSmi.SqlMetaDataToSmiExtendedMetaData(_columnMetaData[i]);
             }
 
-            _eventSink = new SmiEventSink_Default();
             #if NETFRAMEWORK
```
benrr101: Since these are a lot closer now, can we simplify it to:

```
_recordBuffer = new MemoryRecordBuffer(_columnSmiMetaData);

#if NETFRAMEWORK
_usesStringStorageForXml = true;
#endif
```

benrr101 (dismissed):
Looks great! One small tweak you can make if you feel like it. No worries about stepping on my toes - I kinda thought I had hit the end of the line for removing SMI code, so I'm extra glad there's more to chuck 😄

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SqlDataRecord.cs
edwardneal: Yes - thanks. Corrected and committed!

benrr101 (approved):

paulmedynski (approved):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SqlDataRecord.netcore.cs
```diff
@@ -20,18 +20,18 @@ private Type GetFieldTypeFrameworkSpecific(int ordinal) =>
             MetaType.GetMetaTypeFromSqlDbType(GetSqlMetaData(ordinal).SqlDbType, false).ClassType;
 
         private object GetValueFrameworkSpecific(int ordinal) =>
-            ValueUtilsSmi.GetValue200(_eventSink, _recordBuffer, ordinal, GetSmiMetaData(ordinal));
+            ValueUtilsSmi.GetValue200(_recordBuffer, ordinal, GetSmiMetaData(ordinal));
```
samsharma2700, line 23: Great work! Should we keep a note of target-specific `SqlDataReader` files that should removed in future?

edwardneal: Thanks. It'll be a short-lived note (it's the next PR I'm planning, and it'll wrap up the task started by #2996) but my pending work is:
1. Remove the various `Get*ValueFrameworkSpecific` methods, replacing them with the ValueUtilsSmi calls at the original call site.
2. Remove the optional parameter from `ValueUtilsSmi.SetTimeSpan` and `.SetDateTimeOffset` - these methods are always called with a value of true.
3. Remove the various `Set*Time*FrameworkSpecific` methods, treating them the same way as point 1.
4. Move the `SetValue*FrameworkSpecific` method to SqlDataRecord.cs and delete the target-specific files
5. Remove the [`SmiRecordBuffer`](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SmiRecordBuffer.cs) class
6. Clean up the now-unused resource strings, AdapterUtil and SqlUtil methods referring to context connections.

PR #3446, opened 2025-06-26
Fix ADO pipline missing variable

paulmedynski:
SQL pipeline variable doesn't exist, so explicitly setting TSA and ApiScan to enabled.


eng/pipelines/dotnet-sqlclient-signing-pipeline.yml
```diff
@@ -90,9 +90,9 @@ extends:
         # The OneBranch template will set 'break' to false for the other SDL
         # tools when TSA is enabled.  This allows TSA to gather the results
         # and publish them for downstream analysis.
-        enabled: ${{parameters.enableAllSdlTools }}
```
paulmedynski, line 93: This pipeline variable doesn't exist, so this has been broken either since it was added, or since the variable was removed.

samsharma2700 (approved):

PR #1647, opened 2022-06-17
Remove union overlay design and use reflection in SqlTypeWorkarounds

David-Engel:
This replaces the union overlay design against netfx with reflection queries to see if appropriate internal methods exist and utilizes them. If they don't exist, the code falls back to slower public APIs.

The union overlay design is retained against netcore because most of the internal methods used were not referenced by netcore itself and removed by the netcore build process so they don't actually exist. We will push for public APIs in .NET and migrate to them when they are available. Calling the current public APIs in netcore would result in significant performance penalties in this hot path.

PR #3448, opened 2025-06-26
Merge | Netfx SqlTypeWorkarounds

benrr101:
## Description
This PR is less of a merge and more of a total rewrite 😅 . There are a handful of methods that are used in the code to convert between internal representations of a handful of Sql* types (SqlBinary, SqlDecimal, SqlGuid, and SqlMoney). In netcore, we can use built-in methods to get at these internal representations, but netfx will not have access to them. There are equivalent conversion methods available, but they almost always make copies of the data. Thus, we have these methods that directly access the data or internal conversion method for the best performance when reading these types.

The original implementations of these were clunky and confusing, and although it'd be easy to copy-paste them into the common project file, I wanted to make sure we knew what was going on (and whether they were being used). Thus, with a bit of AI magic, I rewrote these conversion methods. Now, I'm sure no one will trust that I did it right - I mean, I don't even trust myself! So, this is my first PR where I'm introducing *true* unit tests! These were checked against the behavior of the original implementations, so it can be safely assumed that the new implementations do things correctly. I have also added comments to the methods to give a better understanding of how they work, for future people.

Also, say goodbye to dynamic IL method creation!

## Issues
#1261 

## Testing
Added **unit tests** for the rewritten methods!


David-Engel:
@benrr101 - This is a sensitive hot path. I'd perf check your changes. For reference, check out https://github.com/dotnet/SqlClient/pull/1647#issuecomment-1171792576. You might be able to reuse the BDN project on that comment.

edwardneal:
It's good to get the trailing file merged, but I'm not sure switching to using pointers is a real simplification. The dynamic method is well-documented and although it's unusual, it doesn't raise questions about the managed/unmanaged memory layout of SqlDecimal, whether the value needs to be pinned or any performance issues which might arise from copying structs if they're passed by value.

I was planning to use a DynamicMethod when simplifying the UDT serialization so might be slightly biased, but I think keeping the initial approach of emitting a single IL method might be simpler here. The method we emit isn't much more than a chain of field accesses.

benrr101:
@David-Engel As per your request, here's comparison of perf for old vs new, fast vs fallback. As you can see, almost all of them are faster or are negligibly slower (ie <1%). Though the question is whether this is compliant or not (I'll run a test to see what it says).

[Numbers removed b/c they're out of date now]

---

@edwardneal I'd have to hard disagree on using dynamic methods. Unless there's a huge performance difference between not using them and using them (like 20%+) I think readability and maintainability outweigh it. I think for most cases (at least for our case), engineers are going to be more familiar with pointers than IL opcodes. And in this case, removing the dynamic method actually improves perf by 11% (though some of that might be due to removing checks and some layers of indirection).

Though this did lead me down a trail to learn more about marshaling and blittable types. As near as I can figure, for structs, as long as they aren't defined as using `LayoutKind.Automatic` there's no difference between unmanaged and managed layouts (and if they are, marshaling will throw). `SqlDecimal`, doesn't specify a `LayoutKind`, so I assume it uses `0`, which corresponds to `Layout.Sequential`. Though, I'm curious to learn more about it, if you've got better info for me

benrr101:
Update, some more information has come to light about considerations for these code paths, and I need to reimplement a couple of them. As such, I'm going to mark this as draft until it's ready. Updated perf numbers will be provided when ready.

And as a heads up to @edwardneal the DynamicMethod will go away, but so will accessing member variables.

benrr101:
Alrighty, updates are ready for review. 

@David-Engel as per your request, here's the perf numbers:
| Method                           | Mean         | Error      | StdDev     | Pct Change  |
|--------------------------------- |-------------:|-----------:|-----------:|------------:|
| SqlMoney_FromLongFast_Old        |     8.169 ns |  0.0798 ns |  0.0708 ns |             |
| SqlMoney_FromLongFast_New        |     7.633 ns |  0.0955 ns |  0.0893 ns | 6.6% faster |
| SqlMoney_FromLongSlow_Old        |     8.096 ns |  0.1033 ns |  0.0966 ns |             |
| SqlMoney_FromLongSlow_New        |     7.634 ns |  0.0458 ns |  0.0382 ns | 5.7% faster |
| SqlMoney_ToLongFast_Old          |     7.266 ns |  0.0738 ns |  0.0691 ns |             |
| SqlMoney_ToLongFast_New          |     7.259 ns |  0.0650 ns |  0.0608 ns | <1% faster  |
| SqlMoney_ToLongSlow_Old          |    55.909 ns |  1.0257 ns |  0.9595 ns |             |
| SqlMoney_ToLongSlow_New          |     7.532 ns |  0.0832 ns |  0.0778 ns | 86% faster  |
| SqlDecimal_WriteTdsValueFast_Old |    47.304 ns |  0.6128 ns |  0.5732 ns |             |
| SqlDecimal_WriteTdsValueFast_New |    49.798 ns |  0.3473 ns |  0.3079 ns | 5.3% slower |
| SqlDecimal_WriteTdsValueSlow_Old |    50.678 ns |  0.7565 ns |  0.6706 ns |             |
| SqlDecimal_WriteTdsValueSlow_New |    50.547 ns |  0.6062 ns |  0.5374 ns | 0.2% faster |
| SqlBinary_FromByteArrayFast_Old  | 4,370.409 ns | 81.8799 ns | 76.5905 ns |             |
| SqlBinary_FromByteArrayFast_New  | 4,276.846 ns | 34.5898 ns | 28.8840 ns | 2.1% faster |
| SqlBinary_FromByteArraySlow_Old  | 4,283.448 ns | 45.3417 ns | 40.1942 ns |             |
| SqlBinary_FromByteArraySlow_New  | 4,273.694 ns | 57.0698 ns | 47.6559 ns | <1% faster  |
| SqlGuid_FromByteArrayFast_Old    |    79.946 ns |  1.0679 ns |  0.9989 ns |             |
| SqlGuid_FromByteArrayFast_New    |    80.508 ns |  1.2926 ns |  1.2091 ns | <1% slower  |
| SqlGuid_FromByteArraySlow_Old    |    80.722 ns |  1.5184 ns |  1.5593 ns |             |
| SqlGuid_FromByteArraySlow_New    |    79.871 ns |  1.0893 ns |  1.0189 ns | 1.1% faster |

edwardneal:
Thanks for checking @benrr101; I've just done the same, and agree - the managed and unmanaged layouts are going to be the same (although that's pretty much academic now.)

The removal of DynamicMethod is probably the reason why SqlDecimal_WriteTdsValueFast has regressed (and it'd be good to see the benchmark's memory allocation figures) but if you're happy with the performance/complexity tradeoff then everything else looks good to me.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlTypes/SqlTypeWorkarounds.cs
```diff
@@ -90,5 +91,192 @@ internal static DateTime SqlDateTimeToDateTime(int daypart, int timepart)
...
+            
+        #endregion
+
+        private static unsafe Func<TValue, TInstance> CreateFactory<TInstance, TValue, TIgnored>(
+            Func<TValue, TInstance> fallbackFactory)
+            where TInstance : struct
+        {
+            // The logic of this method is that there are special internal methods that can create
+            // Sql* types without the need for copying. These methods are internal to System.Data,
+            // so we cannot access them, even they are so much faster. To get around this, we
```
paulmedynski, line 225: even _though_ they

ErikEJ:
5.3 % slower on decimal? Is that OK - or am I misreading the numbers?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlTypes/SqlTypeWorkarounds.cs
```diff
@@ -18,39 +23,35 @@ namespace Microsoft.Data.SqlTypes
...
             Debug.Assert(closeInput || !async, "Currently we do not have pre-created settings for !closeInput+async");
 
-            XmlReaderSettings settingsToUse = closeInput ?
-                (async ? s_defaultXmlReaderSettingsAsyncCloseInput : s_defaultXmlReaderSettingsCloseInput) :
-                s_defaultXmlReaderSettings;
+            XmlReaderSettings settingsToUse = closeInput
+                ? async 
+                    ? s_defaultXmlReaderSettingsAsyncCloseInput
+                    : s_defaultXmlReaderSettingsCloseInput
+                : s_defaultXmlReaderSettings;
```
samsharma2700, line 50: This is so much better to read 😄

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlTypes/SqlTypeWorkarounds.cs
```diff
@@ -90,5 +91,192 @@ internal static DateTime SqlDateTimeToDateTime(int daypart, int timepart)
...
+        {
+            try
+            {
+                // Note: Although it would be faster to use the m_value member variable in
+                //    SqlMoney, but because it is not documented, we cannot use it. The method
+                //    we are calling below *is* documented, despite it being internal.
+                // Documentation for internal method:
+                // https://learn.microsoft.com/en-us/dotnet/framework/additional-apis/system.data.sqltypes.sqlmoney.tosqlinternalrepresentation
+                
+                MethodInfo method = typeof(SqlMoney).GetMethod(
```
samsharma2700, line 184: Invoking a statically cached method would be faster than redoing Reflection lookup. Something like `static readonly MethodInfo s_toSqlInternal = typeof(SqlMoney).GetMethod(...);`

samsharma2700 (changes requested):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlTypes/SqlTypeWorkarounds.cs
benrr101: I hope I'm addressing your comment correctly...
Up on line 157, we call CreateSqlMoneyToLongFactory at static construction, so we only do the reflection lookup once at static construction. We also warm up the method before caching it, so it's always going as fast as possible. All calls to SqlMoneyToLong go through the cached method. So, I don't think there's any issue here?

samsharma2700: I see, thanks for explaining! looks good to me then.

benrr101:
> 5.3 % slower on decimal? Is that OK - or am I misreading the numbers?

@ErikEJ Yes, you read that right ... but, we're bound by legal reasons to implement the method this way over the previous way.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlTypes/SqlTypeWorkarounds.cs
benrr101: Since this is the only change requested and fixing would kick off another round of CI and re-review, this change is too small to take.

PR #3451, opened 2025-06-27
Add Feature Extension for User Agent

samsharma2700:
## Description

Adding support for User Agent Feature Extension(part 1) according to design doc here(https://microsoft.sharepoint-df.com/:w:/t/sqldevx/ERIWTt0zlCxLroNHyaPlKYwBI_LNSff6iy_wXZ8xX6nctQ?e=iP8q75):

- Introducing necessary constants.
- Added function to write appropriate packet data.
- Added Unit tests for the changes

Part 2 will include the `userAgentJsonPayload` related implementations followed by functional tests.

## Testing
Builds are passing.
Unit tests are passing.


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -8529,6 +8529,32 @@ internal int WriteFedAuthFeatureRequest(FederatedAuthenticationFeatureExtensionD
             return len;
         }
 
+        internal int WriteUserAgentFeatureRequest(byte[] userAgentJsonPayload,
```
apoorvdeshmukh, line 8599: Nit: Add descriptive comment for the function about what it does, what it returns and format of the feature request.
You may refer to WriteVectorSupportFeatureRequest

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -207,6 +207,9 @@ internal bool IsDNSCachingBeforeRedirectSupported
         // Json Support Flag
         internal bool IsJsonSupportEnabled = false;
 
+        // User Agent Flag
+        internal bool IsUserAgentEnabled = true;
```
apoorvdeshmukh, line 211: Does this not depend on the ack received from server?  We generally send the feature request to the server and based on the ack received from server, we will set this to true.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -208,6 +208,9 @@ internal bool IsDNSCachingBeforeRedirectSupported
         // Json Support Flag
         internal bool IsJsonSupportEnabled = false;
 
+        // User Agent Flag
+        internal bool IsUserAgentEnabled = true;
```
apoorvdeshmukh, line 215: Do check here as well if default should be false and needs to be set to true based on the server response.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsEnums.cs
```diff
@@ -253,7 +254,8 @@ public enum FeatureExtension : uint
             DataClassification = 1 << (TdsEnums.FEATUREEXT_DATACLASSIFICATION - 1),
             UTF8Support = 1 << (TdsEnums.FEATUREEXT_UTF8SUPPORT - 1),
             SQLDNSCaching = 1 << (TdsEnums.FEATUREEXT_SQLDNSCACHING - 1),
-            JsonSupport = 1 << (TdsEnums.FEATUREEXT_JSONSUPPORT - 1)
+            JsonSupport = 1 << (TdsEnums.FEATUREEXT_JSONSUPPORT - 1),
+            UserAgent = 1 << (TdsEnums.FEATUREEXT_USERAGENT - 1)
```
apoorvdeshmukh, line 261: Need to resolve some branch conflicts I guess as I don't see vector feature extension here.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -1426,6 +1429,10 @@ private void Login(ServerInfo server, TimeoutTimer timeout, string newPassword,
             requestedFeatures |= TdsEnums.FeatureExtension.SQLDNSCaching;
             requestedFeatures |= TdsEnums.FeatureExtension.JsonSupport;
 
+        #if DEBUG
```
ErikEJ, line 1436: Not in Release  build?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
samsharma2700: Server will not send an ack for this FE. We probably won't need this flag but wanted to add it till I finalize the JsonPayload work and comply with current FE functions.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
samsharma2700: Not yet but I have an issue to track and remove these once part 2 and 3 are merged.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
samsharma2700: Good point, will add a summary.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
samsharma2700: It will be set to true by default, there is no case from spec to disable. Server will just ignore it based on version. I'll still keep a note to remove it entirely if no use case arises from the rest of FE work.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -8579,6 +8579,32 @@ internal int WriteVectorSupportFeatureRequest(bool write)
...
+                                                  bool write /* if false just calculates the length */)
+        {
+            // 1byte (Feature Version) + size of UTF-8 encoded JSON payload 
+            int dataLen = 1 + userAgentJsonPayload.Length;
+            // 1byte (Feature ID) + 4bytes (Feature Data Length) + 1byte (Version) + N(JSON payload size)
+            int totalLen = 1+ 4 + dataLen;
```
paulmedynski: `1+ 4` -> `1 + 4`

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsEnums.cs
```diff
@@ -241,6 +241,7 @@ public enum EnvChangeType : byte
         public const byte FEATUREEXT_SQLDNSCACHING = 0x0B;
         public const byte FEATUREEXT_JSONSUPPORT = 0x0D;
         public const byte FEATUREEXT_VECTORSUPPORT = 0x0E;
+        public const byte FEATUREEXT_USERAGENT = 0x0F;
```
paulmedynski, line 245: This byte value will compete with the feature extension that @mdaigle is negotiating for the routing+database work.  Just FYI - we will need to coordinate which feature gets which byte value.

paulmedynski (changes requested):
One minor fix.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsEnums.cs
samsharma2700: Ack, will keep a note and leave a to-do here.

apoorvdeshmukh: We need to document this value in TDS spec as well if we haven't done that yet.
@mdaigle will have to do the same. Maybe we could make both of these updates at once in the TDS spec once the values are finalized.

samsharma2700: Yes that sounds good, will track this as a task.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -8579,6 +8579,49 @@ internal int WriteVectorSupportFeatureRequest(bool write)
...
+        /// </param>
+        /// <returns>The length of the feature request in bytes.</returns>
+        /// <remarks>
+        /// The feature request consists of:
+        /// - 1 byte for the feature ID.
+        /// - 4 bytes for the feature data length.
+        /// - 1 byte for the version number.
+        /// - N bytes for the JSON payload
+        /// </remarks>
+        internal int WriteUserAgentFeatureRequest(byte[] userAgentJsonPayload,
```
mdaigle, line 8599: How can we test that this function works as expected? Is there any opportunity to add a unit test?

apoorvdeshmukh: @samsharma2700 Let's add testcase  to test this feature request. You can refer to TestConnWithVectorFeatExtVersionNegotiation to inspect login packet.

samsharma2700: Ack, added unit tests

src/Microsoft.Data.SqlClient/tests/UnitTests/TdsParserInternalsTest.cs
```diff
@@ -0,0 +1,115 @@
...
+using Xunit;
+
+#nullable enable
+
+namespace Microsoft.Data.SqlClient.UnitTests
+{
+    public class TdsParserInternalsTest
+    {
+        // Selects and returns the first non-public instance constructor for TdsParser 
+        private static TdsParser CreateParserInstance()
```
paulmedynski: The UnitTests project has access to MDS internals, so you shouldn't need any reflection.

src/Microsoft.Data.SqlClient/tests/UnitTests/TdsParserInternalsTest.cs
```diff
@@ -0,0 +1,115 @@
...
+                .Select(p => p.ParameterType.IsValueType
+                    ? Activator.CreateInstance(p.ParameterType)
+                    : null)
+                .ToArray();
+
+            return (TdsParser)ctor.Invoke(ctorArgs);
+        }
+
+        // Helper function to extract private _physicalStateObj fields raw buffer and no. of bytes written so far 
+        private static (byte[] buffer, int count) ExtractOutputBuffer(TdsParser parser)
```
paulmedynski, line 20: This is unavoidable until we can split the policy and mechanism intertwined within TdsParser and his friends.

Please create a User Story to do this split work, and add a comment here referencing it:
```
// TODO(ADO-1234): Avoid reflection by exposing a way for tests to intercept outbound TDS packets.
```

src/Microsoft.Data.SqlClient/tests/UnitTests/TdsParserInternalsTest.cs
paulmedynski: This function can just become a private member:
```
private readonly TdsParser _parser = new(false, false);
```

src/Microsoft.Data.SqlClient/tests/UnitTests/TdsParserInternalsTest.cs
```diff
@@ -0,0 +1,115 @@
...
+            int count = (int)usedField.GetValue(stateObj)!;
+
+            return (buffer, count);
+        }
+
+        [Fact]
+        public void WriteUserAgentFeatureRequest_WriteFalse_LengthOnlyReturn()
+        {
+            byte[] payload = Encoding.UTF8.GetBytes("{\"kel\":\"sier\"}");
+            var parser = CreateParserInstance();
```
paulmedynski: And here is where you will just use _parser.

src/Microsoft.Data.SqlClient/tests/UnitTests/TdsParserInternalsTest.cs
```diff
@@ -0,0 +1,115 @@
...
+
+            int lengthOnly = parser.WriteUserAgentFeatureRequest(payload, write: false);
+
+            // assert: total = 1 (feat-ID) + 4 (len field) + [1 (version) + payload.Length]
+            int expectedDataLen = 1 + payload.Length;
+            int expectedTotalLen = 1 + 4 + expectedDataLen;
+            Assert.Equal(expectedTotalLen, lengthOnly);
```
paulmedynski, line 59: Missing an assert that nothing was written to whatever buffer it would otherwise have been.

src/Microsoft.Data.SqlClient/tests/UnitTests/TdsParserInternalsTest.cs
```diff
@@ -0,0 +1,115 @@
...
+        public void WriteUserAgentFeatureRequest_WriteTrue_AppendsOnlyExtensionBytes()
+        {
+            byte[] payload = Encoding.UTF8.GetBytes("{\"kel\":\"sier\"}");
+            var parser = CreateParserInstance();
+
+            var (bufferBefore, countBefore) = ExtractOutputBuffer(parser);
+
+            int returnedLength = parser.WriteUserAgentFeatureRequest(payload, write: true);
+
+            var (bufferAfter, countAfter) = ExtractOutputBuffer(parser);
```
paulmedynski: Do we expect that bufferBefore and bufferAfter are the same object?
```
Assert.Same(bufferBefore, bufferAfter);
```

src/Microsoft.Data.SqlClient/tests/UnitTests/TdsParserInternalsTest.cs
```diff
@@ -0,0 +1,115 @@
...
+
+            int dataLenFromStream = BitConverter.ToInt32(bufferAfter, start + 1);
+            int expectedDataLen = 1 + payload.Length;
+            Assert.Equal(expectedDataLen, dataLenFromStream);
+
+            Assert.Equal(
+                TdsEnums.SUPPORTED_USER_AGENT_VERSION,
+                bufferAfter[start + 5]);
+
+            byte[] writtenPayload = bufferAfter
```
paulmedynski: You can avoid any copies:
```
ReadOnlySpan writtenPayload = new(bufferAfter, 6, appended - 6);
```
The `Assert.Equal()` below may be smart enough to conerce the `payload` byte[] into a span, or you may have to do it by hand.

src/Microsoft.Data.SqlClient/tests/UnitTests/TdsParserInternalsTest.cs
```diff
@@ -0,0 +1,115 @@
...
+                .Skip(start + 6)
+                .Take(payload.Length)
+                .ToArray();
+            Assert.Equal(payload, writtenPayload);
+
+            Assert.Equal(returnedLength, appended);
```
paulmedynski: You already checked this on line 89.

paulmedynski (changes requested):
Looks good, with some minor tweaks.

src/Microsoft.Data.SqlClient/tests/UnitTests/TdsParserInternalsTest.cs
samsharma2700: Added the task(ADO-37888)

src/Microsoft.Data.SqlClient/tests/UnitTests/TdsParserInternalsTest.cs
samsharma2700: Added `Assert.Equal(countBefore, countAfter);`

src/Microsoft.Data.SqlClient/tests/UnitTests/TdsParserInternalsTest.cs
samsharma2700: Good catch! removed with updated tests.

src/Microsoft.Data.SqlClient/tests/UnitTests/TdsParserInternalsTest.cs
samsharma2700: Ack, now we slice into the existing buffer.

src/Microsoft.Data.SqlClient/tests/UnitTests/TdsParserInternalsTest.cs
samsharma2700: Yes, we do. Added it as an Assert as well

src/Microsoft.Data.SqlClient/tests/UnitTests/TdsParserInternalsTest.cs
samsharma2700: Nice, using this in test now.

paulmedynski (approved):

src/Microsoft.Data.SqlClient/tests/UnitTests/TdsParserInternalsTest.cs
```diff
@@ -0,0 +1,107 @@
...
+            Assert.Equal(expectedDataLen, dataLenFromStream);
+
+            Assert.Equal(
+                TdsEnums.SUPPORTED_USER_AGENT_VERSION,
+                bufferAfter[start + 5]);
+
+            // slice into the existing buffer
+            ReadOnlySpan<byte> writtenSpan = new ReadOnlySpan<byte>(
+                bufferAfter,
+                start + 6,
```
apoorvdeshmukh, line 97: Nit: Maybe some comment explaining these constants would be helpful.

samsharma2700: Fair, will add more details along with other tests with next PR(part 2 of this change)

PR #3452, opened 2025-06-27
Official pipeline tweaks - round 3

paulmedynski:
## Description

Another round of tweaks to get the Official pipeline working:
- Added netstandard2.0 to the list of expected runtimes when validating an Official MDS package.
- Added comments to Versions.props warning of fragile validation logic.
- Improved comments in the validation pipeline job.

### DIFF
--- a/eng/pipelines/libraries/mds-validation-variables.yml
+++ b/eng/pipelines/libraries/mds-validation-variables.yml
@@ -17,7 +17,7 @@ variables:
   - name: expectedFolderNames
     value: lib,ref,runtimes
   - name: expectedDotnetVersions
-    value: net462,net8.0,net9.0
+    value: netstandard2.0,net462,net8.0,net9.0
   - name: Database
     value: Northwind
   - name: platform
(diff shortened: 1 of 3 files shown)


eng/pipelines/libraries/mds-validation-variables.yml
```diff
@@ -17,7 +17,7 @@ variables:
   - name: expectedFolderNames
     value: lib,ref,runtimes
   - name: expectedDotnetVersions
-    value: net462,net8.0,net9.0
+    value: netstandard2.0,net462,net8.0,net9.0
```
paulmedynski, line 20: This is the actual fix, since we've added .NET Standard 2.0 back.

tools/props/Versions.props
```diff
@@ -4,12 +4,23 @@
   <PropertyGroup>
     <MdsVersionDefault>6.1.0</MdsVersionDefault>
     <BuildNumber Condition="'$(BuildNumber)' == ''">0</BuildNumber>
+
+    <!--
+      GOTCHA: These two properties must appear in the first <PropertyGroup> in
```
paulmedynski, line 9: The validation scripts are very fragile, so we need to leave warnings like this in our files.

priyankatiwari08 (approved):

PR #3454, opened 2025-06-27
Cleanup | Merge SqlDataRecord and remove context connection remnants

### LINKED ISSUE #2838: .NET Framework SQLCLR support

edwardneal:
## Description

This picks up where #3438 left off. With `SmiEventSink_Default` removed, almost all of the code in SqlDataRecord.netfx.cs and SqlDataRecord.netcore.cs is identical and can be removed. We do so.

The `SmiRecordBuffer` class is also removed, since this isn't doing anything meaningful. I'm not pleased with the overall type hierarchy here, but I want to think through the consequences of changing them in the wider context of ValueUtilsSmi.

With that work done, we just remove the now-unused resource strings and exception methods from SqlUtil, AdapterUtil and the string resources. There are other unused strings and exception methods here, for anyone interested in picking up a bulky commit.

Closes #2838.

## Issues

Follows #3438.
Resolves #2838.
Contributes to #1261.

## Testing

Local unit tests for SqlDataRecord pass.


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SqlDataRecord.cs
```diff
@@ -54,10 +54,24 @@ public virtual string GetDataTypeName(int ordinal)
 #if NET
         [return: DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicProperties | DynamicallyAccessedMemberTypes.PublicFields)]
 #endif
-        public virtual Type GetFieldType(int ordinal) => GetFieldTypeFrameworkSpecific(ordinal);
+        public virtual Type GetFieldType(int ordinal)
+        {
+            SqlMetaData md = GetSqlMetaData(ordinal);
+
+#if NETFRAMEWORK
+            if (md.SqlDbType == SqlDbType.Udt)
```
benrr101, line 62: Since this check only applies to netframework and will return, let's avoid the braces for netcore

```c#
#if NETFRAMEWORK
if (md.SqlDbType == SqlDbType.Udt)
{
    return md.Type;
}
#endif

return MetaType.GetMetaTypeFromSqlDbType(md.SqlDbType, false).ClassType;

Comment (benrr101) [reply src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SqlDataRecord.cs:62]: Also tiny request, can we get an argument label for the `false` here? :)

**Review by benrr101 — APPROVED** [review]
We do so 🫡 

(one small request, but overall it's great)

File: src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SqlDataRecord.cs (hunk 1 above)
Comment (edwardneal) [reply src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SqlDataRecord.cs:62]: Yes - done!

**Review by paulmedynski — APPROVED** [review]

File: src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SqlDataRecord.cs [hunk 2]
```diff
@@ -246,10 +258,73 @@ public virtual int GetSqlValues(object[] values)
...
-        public virtual int SetValues(params object[] values) => SetValuesFrameworkSpecific(values);
+        public virtual int SetValues(params object[] values)
+        {
+            if (values == null)
+            {
+                throw ADP.ArgumentNull(nameof(values));
+            }
+
+            // Allow values array longer than FieldCount, just ignore the extra cells.
+            int copyLength = (values.Length > FieldCount) ? FieldCount : values.Length;
```
samsharma2700, line 269: We can do some early exit optimization here `if (copyLength == 0) return 0;`

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SqlDataRecord.cs
```diff
@@ -246,10 +258,73 @@ public virtual int GetSqlValues(object[] values)
...
+                    values[i],
+                    metaData.Type);
+                if (typeCodes[i] == ExtendedClrTypeCode.Invalid)
+                {
+                    throw ADP.InvalidCast();
+                }
+            }
+
+            // Now move the data (it'll only throw if someone plays with the values array between
+            //      the validation loop and here, or if an invalid UDT was sent).
```
samsharma2700: We can make a shallow copy of the portion of array to be used 
```
object[] stableValues = new object[copyLength];
Array.Copy(values, stableValues, copyLength);
```
 then use `stableValues[i]` in both loops, guaranteeing stability of inputs between validation and writing phase.

samsharma2700 (changes requested):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SqlDataRecord.cs
edwardneal: Nothing actually tampers with the `values` array so I don't think we need to make a defensive copy. I'll adjust the comment accordingly.

paulmedynski (approved):

samsharma2700 (approved):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SqlDataRecord.cs
samsharma2700: Thank you!

PR #3458, opened 2025-06-29
Tests | Move various unit tests to UnitTests project

edwardneal:
## Description

This is another PR which is a little smaller than it looks. It locates some of the smaller tests which made heavy use of reflection to access internal SqlClient types, moves them from FunctionalTests to UnitTests, removes the reflection and adds descriptions to the test cases. The steps are visible if reviewed commit-by-commit.

Besides some basic test movements, it also lays the groundwork for expanding the test coverage of SqlBuffer. At the moment, we only have explicit coverage of Guid and SqlGuid. A future PR will cover the other types.

There's a little overlap between this and #3423; I've tried to maintain the same standards as there, and I'll revisit them if #3423 is merged first.

More broadly here, there are around 1200 tests in FunctionalTests. I've assumed that the new UnitTests project is eventually going to contain most of the tests in FunctionalTests, is that correct?

## Issues

None.

## Testing

All unit tests pass.


paulmedynski (approved):

edwardneal:
Thanks @mdaigle - that all makes sense. I'll put new tests there and port over older ones as I touch them.

I wasn't sure about fixing that part of the SqlCommandSetTests, but I've changed the visibility of those members from private to internal and removed the reflection.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
```diff
@@ -114,7 +114,7 @@ protected override void AfterCleared(SqlCommand owner)
         }
 
         private string _commandText;
-        private CommandType _commandType;
+        internal CommandType _commandType;
```
paulmedynski: Hmmm - Looking at SqlCommand's CommantType setter (line 738), it never contains a value other than Text or StoredProcedure.  Why is SqlCommandText performing the same check (line 108)?  I think we can remove the check from SqlCommandSet - at most replace it with a Debug.Assert().  Then we don't need the associated test for SqlCommandSet, and we don't need to expose _commandType here.

Unfortunately, the docs for SqlCommand.CommandType don't declare this API promise, or that setting CommandType can throw.  But that's out of scope for these changes.

edwardneal: That's just odd! It looks like the validation was part of the codebase even back to System.Data.SqlClient, #1377 added coverage for it as part of a wider drive and I ported it into the new project by mistake. Thanks for catching that.

I've demoted the validation to a simple assertion, and ditched the test.

paulmedynski (approved):
Looks great - and one less test to worry about :)

mdaigle (approved):

Issue #1773, opened 2022-09-22
SQLException.Number property differs on Linux OS and Windows.


g-lashuk:
Hello.

Could you please clarify the behavior of the following code example:

```cs
public static void Main(string[] args)
{
    try
    {
        var connectionString =
            "Server=non-existent-server;Database=nonexistentDb;User ID=sa;Password=invalidPassword;max pool size=1000";
        var sqlConnection = new SqlConnection(connectionString);
        sqlConnection.Open();
        Console.WriteLine("Opened!");
        sqlConnection.Close();
    }
    catch (SqlException sqlException)
    {
        Console.WriteLine(sqlException.ToString());
        Console.WriteLine($"SQLException.Number: {sqlException.Number}");
    }
}
```

Running it on Windows 10 Pro 19044.2006 results in:

```
Microsoft.Data.SqlClient.SqlException (0x80131904): A network-related or instance-specific error occurred while establishing a connection to SQL Server. The server was not found or was not accessible. Verify that the instance name is correct and that SQL Server is configured to allow remote connections. (provider: Named Pipes Provider, error: 40 - Could not open a connection to SQL Server)
 ---> System.ComponentModel.Win32Exception (53): The network path was not found.
   at Microsoft.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObject stateObj, Boolean callerHasConnectionLock, Boolean asyncClose)
   at Microsoft.Data.SqlClient.TdsParser.Connect(ServerInfo serverInfo, SqlInternalConnectionTds connHandler, Boolean ignoreSniOpenTimeout, Int64 timerExpire, SqlConnectionString connectionOptions, Boolean withFailover)
   at Microsoft.Data.SqlClient.SqlInternalConnectionTds.AttemptOneLogin(ServerInfo serverInfo, String newPassword, SecureString newSecurePassword, Boolean ignoreSniOpenTimeout, TimeoutTimer timeout, Boolean withFailover)
   at Microsoft.Data.SqlClient.SqlInternalConnectionTds.LoginNoFailover(ServerInfo serverInfo, String newPassword, SecureString newSecurePassword, Boolean redirectedUserInstance, SqlConnectionString connectionOptions, SqlCredential credential, TimeoutTimer timeout)
[...]

PR #3461, opened 2025-07-03
Align SqlException Numbers across platforms

David-Engel:
## Description

Attempt to better align SqlException Numbers across platforms. Wondering if it might be this simple.

## Issues

#1773

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ManagedSni/SniError.netcore.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ManagedSni/SniError.netcore.cs
@@ -5,6 +5,8 @@
 #if NET
 
 using System;
+using System.ComponentModel;
+using System.Net.Sockets;
 
 namespace Microsoft.Data.SqlClient.ManagedSni
 {
@@ -14,17 +16,18 @@ namespace Microsoft.Data.SqlClient.ManagedSni
     internal class SniError
     {
         // Error numbers from native SNI implementation
-        internal const uint CertificateValidationErrorCode = 2148074277;
+        // This is signed int representation of the error code 0x80090325 
+        internal const int CertificateValidationErrorCode = -2146893019;
 
         public readonly SniProviders provider;
         public readonly string errorMessage;
-        public readonly uint nativeError;
+        public readonly int nativeError;
         public readonly uint sniError;
         public readonly string function;
         public readonly uint lineNumber;
         public readonly Exception exception;
 
-        public SniError(SniProviders provider, uint nativeError, uint sniErrorCode, string errorMessage)
+        public SniError(SniProviders provider, int nativeError, uint sniErrorCode, string errorMessage)
         {
             lineNumber = 0;
             function = string.Empty;
@@ -35,12 +38,25 @@ namespace Microsoft.Data.SqlClient.ManagedSni
             exception = null;
         }
 
-        public SniError(SniProviders provider, uint sniErrorCode, Exception sniException, uint nativeErrorCode = 0)
+        public SniError(SniProviders provider, uint sniErrorCode, Exception sniException, int nativeErrorCode = 0)
         {
             lineNumber = 0;
             function = string.Empty;
             this.provider = provider;
             nativeError = nativeErrorCode;
+            if (nativeErrorCode == 0)
+            {
+                if (sniException is SocketException socketException)
+                {
+                    // SocketErrorCode values are cross-plat consistent in .NET (matching native Windows error codes)
+                    // underlying type of SocketErrorCode is int
+                    nativeError = (int)socketException.SocketErrorCode;
+                }
+                else if (sniException is Win32Exception win32Exception)
+                {
+                    nativeError = win32Exception.NativeErrorCode; // Replicates native SNI behavior
+                }
+            }
             sniError = sniErrorCode;
             errorMessage = string.Empty;
             exception = sniException;
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ManagedSni/SniTcpHandle.netcore.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ManagedSni/SniTcpHandle.netcore.cs
@@ -375,6 +375,8 @@ namespace Microsoft.Data.SqlClient.ManagedSni
 
                 IEnumerable<IPAddress> ipAddresses = GetHostAddressesSortedByPreference(serverName, ipPreference);
 
+                SocketException lastSocketException = null;
+
                 foreach (IPAddress ipAddress in ipAddresses)
                 {
                     bool isSocketSelected = false;
@@ -426,7 +428,7 @@ namespace Microsoft.Data.SqlClient.ManagedSni
                             {
                                 if (timeout.IsExpired)
                                 {
-                                    return null;
+                                    throw new Win32Exception(258, "The operation has timed out.");
                                 }
 
                                 int socketSelectTimeout =
@@ -442,10 +444,24 @@ namespace Microsoft.Data.SqlClient.ManagedSni
 
                                 Socket.Select(checkReadLst, checkWriteLst, checkErrorLst, socketSelectTimeout);
                                 // nothing selected means timeout
+                                SqlClientEventSource.Log.TrySNITraceEvent(nameof(SniTcpHandle), EventType.INFO,
+                                    "Socket.Select results: checkReadLst.Count: {0}, checkWriteLst.Count: {1}, checkErrorLst.Count: {2}",
+                                    checkReadLst.Count, checkWriteLst.Count, checkErrorLst.Count);
                             } while (checkReadLst.Count == 0 && checkWriteLst.Count == 0 && checkErrorLst.Count == 0);
 
                             // workaround: false positive socket.Connected on linux: https://github.com/dotnet/runtime/issues/55538
                             isConnected = socket.Connected && checkErrorLst.Count == 0;
+                            if (!isConnected)
+                            {
+                                // Retrieve the socket error code
+                                int socketErrorCode = (int)socket.GetSocketOption(SocketOptionLevel.Socket, SocketOptionName.Error);
+                                SocketError socketError = (SocketError)socketErrorCode;
+
+                                SqlClientEventSource.Log.TrySNITraceEvent(nameof(SniTcpHandle), EventType.ERR,
+                                "Socket connection failed. SocketError: {0} ({1})", socketError, socketErrorCode);
+
+                                lastSocketException = new SocketException(socketErrorCode);
+                            }
                         }
 
                         if (isConnected)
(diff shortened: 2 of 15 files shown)


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ManagedSni/SniError.netcore.cs
```diff
@@ -40,7 +42,23 @@ public SniError(SniProviders provider, uint sniErrorCode, Exception sniException
             lineNumber = 0;
             function = string.Empty;
             this.provider = provider;
-            nativeError = nativeErrorCode;
+            if (nativeErrorCode == 0)
+            {
+                nativeError = nativeErrorCode;
```
campersau: I guess this could still be done always before the if so the else isn't needed.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ManagedSni/SniTcpHandle.netcore.cs
```diff
@@ -479,6 +482,14 @@ private static Socket Connect(string serverName, int port, TimeoutTimer timeout,
                     }
                 }
 
+                if (lastSocketException != null)
+                {
+                    SqlClientEventSource.Log.TryAdvancedTraceEvent(
+                        "{0}.{1}{2}Last Socket Exception: {3}",
+                        nameof(SniTcpHandle), nameof(Connect), EventType.ERR, lastSocketException);
+                    throw lastSocketException;
```
paulmedynski, line 508: I'd love to see some tests for these changes.

paulmedynski (changes requested):
Do we have existing tests we could update?  This is pretty critical functionality.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ManagedSni/SniError.netcore.cs
David-Engel: Done

paulmedynski (dismissed):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ManagedSni/SniError.netcore.cs
```diff
@@ -41,6 +43,18 @@ public SniError(SniProviders provider, uint sniErrorCode, Exception sniException
             function = string.Empty;
             this.provider = provider;
             nativeError = nativeErrorCode;
+            if (nativeErrorCode == 0)
+            {
+                if (sniException is SocketException socketException)
+                {
+                    // SocketErrorCode values are cross-plat consistent in .NET (matching native Windows error codes)
+                    nativeError = (uint)socketException.SocketErrorCode;
```
cheenamalhotra: This is a little concerning, as it will translate [SocketError -1](https://learn.microsoft.com/en-us/dotnet/api/system.net.sockets.socketerror?view=net-8.0) to a positive long (4294967295) and will likely throw an exception here, as there is no unchecked block.

paulmedynski: The same negative int -> uint cast is happening below with Win32Exception too.  But in my test program, there is no unchecked exception thrown.  The int -1 value happily becomes uint 4294967295.  There is no loss of data since both types are 32-bit.  I can't tell from this code whether or not the caller expects a large unsigned value for nativeError - none of these public fields are documented.  It looks like nativeError ends up in SNIErrorDetails.nativeError, which is also a uint, so I guess that's fine.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ManagedSni/SniTcpHandle.netcore.cs
```diff
@@ -426,7 +428,7 @@ private static Socket Connect(string serverName, int port, TimeoutTimer timeout,
                             {
                                 if (timeout.IsExpired)
                                 {
-                                    return null;
+                                    throw new Win32Exception(258, "The operation has timed out.");
```
cheenamalhotra: Localize error message?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ManagedSni/SniError.netcore.cs
cheenamalhotra: But can we consider moving away from uint? Since ErrorCodes can be negative too, I wonder if we should adjust accordingly now that we have the opportunity.

We also seem to cast it back to _int_ later when creating `SqlError`:

https://github.com/dotnet/SqlClient/blob/9e5eb97540ea840d788d2425d14cd5f7d42f058b/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs#L1636-L1637

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ManagedSni/SniTcpHandle.netcore.cs
```diff
@@ -444,10 +444,24 @@ private static Socket Connect(string serverName, int port, TimeoutTimer timeout,
...
+                                    "Socket.Select results: checkReadLst.Count: {0}, checkWriteLst.Count: {1}, checkErrorLst.Count: {2}",
+                                    checkReadLst.Count, checkWriteLst.Count, checkErrorLst.Count);
                             } while (checkReadLst.Count == 0 && checkWriteLst.Count == 0 && checkErrorLst.Count == 0);
 
                             // workaround: false positive socket.Connected on linux: https://github.com/dotnet/runtime/issues/55538
                             isConnected = socket.Connected && checkErrorLst.Count == 0;
+                            if (!isConnected)
+                            {
+                                // Retrieve the socket error code
+                                int socketErrorCode = (int)socket.GetSocketOption(SocketOptionLevel.Socket, SocketOptionName.Error);
```
paulmedynski, line 459: Weird that GetSocketOption() returns object? , but then the docs don't describe when it could be null.

paulmedynski (approved):

paulmedynski (dismissed):

paulmedynski:
Let's resolve the conflicts and promote from draft to normal.

paulmedynski:
I rebased to main, fixed the conflicts, and squashed commits.  Please re-review.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ManagedSni/SniTcpHandle.netcore.cs
paulmedynski: Use SNI_WAIT_TIMEOUT from TdsEnums.cs ?

ErikEJ:
Will this be a breaking change?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ManagedSni/SniTcpHandle.netcore.cs
```diff
@@ -654,9 +696,19 @@ private static Socket ParallelConnect(IPAddress[] serverAddresses, int port, Tim
 
                 if (connectedSocket == null)
                 {
+                    if (timeout.IsExpired)
+                    {
+                        throw new Win32Exception(258, "The operation has timed out.");
```
benrr101: Probably want to also change this one to use the string resource

benrr101 (approved):
Overall, seems pretty good!

paulmedynski (dismissed):

paulmedynski:
> Will this be a breaking change?

We don't think so, but I will check the updated code paths to see if anything sticks out.  These changes improve surfacing of underlying connection error codes on non-Windows, which boils down to exposing existing specific error codes at the appropriate times.

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/ApiShould.cs
```diff
@@ -47,10 +47,20 @@ public sealed class ApiShould : IClassFixture<PlatformSpecificTestContext>, IDis
 
         private HashSet<string> _cancellationExceptionMessages = new HashSet<string>()
         {
-            "A severe error occurred on the current command.  The results, if any, should be discarded.\r\nOperation cancelled by user.",
-            "A severe error occurred on the current command.  The results, if any, should be discarded.",
+            "A severe error occurred on the current command.  " +
```
paulmedynski, line 50: Unrelated fixes to unit tests that assumed Windows line endings, and were failing on other OSes.

ErikEJ:
So it will not affect code relying on certain values to be thrown?

David-Engel:
> So it will not affect code relying on certain values to be thrown?

The affected scenarios are all currently surfacing an error number of "0", which is very generic and arguably not actionable. This change should provide actionable error numbers. Linked issues generally refer to this as a "bug" since the underlying errors are actually being hidden. While you could argue that this behavior change is breaking, we've decided to take it for 6.1 since 6.1 will likely be a LTS release and we would be less inclined to make a behavior change in a hotfix. The behavior change will definitely be documented in the release notes.

samsharma2700 (approved):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ManagedSni/SniTcpHandle.netcore.cs
```diff
@@ -426,7 +428,9 @@ private static Socket Connect(string serverName, int port, TimeoutTimer timeout,
                             {
                                 if (timeout.IsExpired)
                                 {
-                                    return null;
+                                    throw new Win32Exception(
```
paulmedynski, line 431: This change will now cause TdsParser.Connect() to throw the Win32Exception created here.  The old code would cause TdsParser.Connect() to throw a SqlException with an error code based on whatever is in SniLoadHandle.SingletonInstance.LastError, which is likely 0.

The Win32Exception appears to bubble all the way up to SqlConnection.Open().  The stack is very convoluted so I may be mistaken here, but if true, then is an API change since Open() doesn't document that it throws Win32Exception.

David-Engel: Any exceptions thrown here (SniTcpHandle.Connect) will be caught in the SniTcpHandle() ctor (where all this logic flows through) and surfaced as an SNI error. All errors still surface as SqlExceptions.

paulmedynski: Ok thanks for clarifying that!

PR #3462, opened 2025-07-06
Cleanup | Remove unused UDT serialization code

edwardneal:
## Description

This sits alongside #3423. Where the earlier PR widens test coverage of the UDT serialization logic, this one simplifies conditions which are constant at runtime. The two primary variables here are `_isTopLevelUdt` and `_skipNormalize`.

`_isTopLevelUdt` is a readonly variable, which is set in the sole `BinaryOrderedUdtNormalizer` constructor to a hardcoded value of `true`. There are then a set of if conditions which test if the variable is false, and which will thus never be called. This PR removes these. The ripple effects of these result in the `_padBuffer` and `_nullInstance` variables never being set (and the `IsNullable` property never being referenced.) These members are also removed.

`_skipNormalize` is another variable, this time defined on the `Normalizer` base class. It is only ever set to `false`, and thus every condition which tests if it is false will succeed. This PR removes the test and runs the code unconditionally.

This accounts for the bulk of the changes. There is one parameter on the constructor to `BinaryOrderedUdtNormalizer` which is never used, a comment which is incorrect, a call to `GetFormat` which discards the result, and several fields on `SqlUdtInfo` which are written to, but never read.

Review of the wider area will highlight some code paths (such as the `Size` property) which remain unused. There are also quirks such as the use of a per-thread type cache, the implementation of the non-generic `IComparable` interface, significant numbers of heap allocations, and likely others. The next set/sets of changes will resolve these, so I've deliberately left these alone to keep this PR focused on the removal of dead code paths.

## Issues

None.

## Testing

All tests pass. I've also tested against the new tests in #3423, and these also pass.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnection.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnection.cs
@@ -2545,11 +2545,10 @@ namespace Microsoft.Data.SqlClient
             }
         }
 
-        internal byte[] GetBytes(object o, out Format format, out int maxSize)
+        internal byte[] GetBytes(object o, out int maxSize)
         {
             SqlUdtInfo attr = GetInfoFromType(o.GetType());
             maxSize = attr.MaxByteSize;
-            format = attr.SerializationFormat;
 
             if (maxSize < -1 || maxSize >= ushort.MaxValue)
             {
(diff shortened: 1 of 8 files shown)


edwardneal:
Thanks. The binary logic is as described in [MS-SSCLRT 2.3.1.2](https://learn.microsoft.com/en-us/openspecs/sql_server_protocols/ms-ssclrt/5a3d583d-64c1-4dc5-b02f-ac8011b7b933), but it's not particularly efficient, on several counts.

PR #3434, opened 2025-06-23
Merge | TdsParser.GetSniErrorDetails

edwardneal:
## Description

This is a slightly smaller merge, but it's the first of its type.

TdsParser.Windows.cs and TdsParser.Unix.cs have a number of methods which vary in behaviour based upon `TdsParserStateObjectFactory.UseManagedSNI`. This PR handles those methods by moving the method to the `TdsParserStateObject` base class, implementing them in the Native and Managed child classes, then updating the references to use the state object.

Once this process is complete, we'll be able to remove the OS-specific TdsParser files.

## Issues

Contributes to #1261.

## Testing

Unit tests pass locally, but I'd like a CI run please.

PR #3469, opened 2025-07-09
Merge | Remove OS-specific TdsParser files

edwardneal:
## Description

This picks up where #3434 left off. The earlier PR handled the simple case of merging OS-specific functionality (which was actually just different between the native and managed SNIs.) This one expands the principle, moving `PostReadAsyncForMars` and `WaitForSSLHandShakeToComplete` from `TdsParser` to `TdsParserStateObject`'s child classes. This removes TdsParser.Unix.cs and TdsParser.Windows.cs.

This has also involved modifying the call site for the method within TdsParser to reference TdsParserStateObject. This is particularly noticeable in EnableSsl.

Finally, while it's not a new gap, this process has highlighted that the existing functionality of warning when down-level SSL is in use only functions on Windows. I've maintained this gap and made it a little more explicit.

## Issues

Contributes to #1261 and #2953.

## Testing

Testing with SSL- and MARS-enabled connection strings works, I'd appreciate a CI run.


paulmedynski (approved):

benrr101:
@edwardneal I wanted to resolve the conflicts for you, but there's one with TdsParserStateObject.PostReadAsyncForMars that I'm not sure the right way to resolve. Seems to me we shouldn't be checking for managed SNI in the native state object, but I'm not an expert here.

I'll give the overall change a review, but let me know when the conflict is resolved and we can get a CI run going.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObjectNative.cs
```diff
@@ -385,6 +385,43 @@ internal override uint DisableSsl()
...
+
+            PacketHandle temp = default;
+            uint error = TdsEnums.SNI_SUCCESS;
+
+#if NETFRAMEWORK
+            RuntimeHelpers.PrepareConstrainedRegions();
+#endif
+            try
+            { }
+            finally
```
benrr101, line 402: I don't get this at all ... isn't the point of a finally block to always run regardless of what happens in the try block, or what exceptions are caught? If there's nothing to run in the try block, doesn't it just immediately skip to the finally block? And if it does, what's the point of the try/finally in the first place?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObjectNative.cs
```diff
@@ -399,43 +436,43 @@ internal override uint EnableSsl(ref uint info, bool tlsFirst, string serverCert
         internal override uint SetConnectionBufferSize(ref uint unsignedPacketSize)
             => SniNativeWrapper.SniSetInfo(Handle, QueryType.SNI_QUERY_CONN_BUFSIZE, ref unsignedPacketSize);
 
-        internal override uint WaitForSSLHandShakeToComplete(out int protocolVersion)
+        internal override uint WaitForSSLHandShakeToComplete(out uint protocolVersion)
```
benrr101: If we're doing this, would it just make more sense to just use SslProtocols as the parameter type?

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -996,19 +974,26 @@ private void EnableSsl(uint info, SqlConnectionEncryptOption encrypt, bool integ
...
+
             // in the case where an async connection is made, encryption is used and Windows Authentication is used,
             // wait for SSL handshake to complete, so that the SSL context is fully negotiated before we try to use its
             // Channel Bindings as part of the Windows Authentication context build (SSL handshake must complete
             // before calling SNISecGenClientContext).
-            error = SniNativeWrapper.SniWaitForSslHandshakeToComplete(_physicalStateObj.Handle, _physicalStateObj.GetTimeoutRemaining(), out uint protocolVersion);
-
-            if (error != TdsEnums.SNI_SUCCESS)
+#if NET
+            if (OperatingSystem.IsWindows())
```
benrr101, line 970: My hope is eventually we will have compile time constants for this so we can replace the whole check with `#if WINDOWS` around the whole if block.

edwardneal: Being able to use `#if WINDOWS` or `#if UNIX` would be good, it'd simplify the OS-specific files in general. In this specific case, I'm planning to enable the SSL protocol warning in a separate PR.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObjectNative.cs
edwardneal: This is part of the CER infrastructure. The finally block will always run, even in the event of a ThreadAbortException. Without the call to `PrepareConstrainedRegions` and a `try` block, this wouldn't be guaranteed.

I think the goal in this specific case is to make sure that if a thread is aborted while setting up the physical MARS connection, the same physical connection is left in a known-good state.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObjectNative.cs
edwardneal: It broadens the number of files impacted, but I agree it'd be cleaner.

edwardneal:
Thanks @benrr101 - I've handled the remaining code review feedback.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
benrr101: Ok no worries for now. This was a non-actionable note 👍

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObjectNative.cs
benrr101: *sigh* I can't stand this CER code... But I'll accept it for now.

benrr101 (approved):

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObjectNative.cs
benrr101: Actually ... I still don't get it. Even if a ThreadAbortExtpion happens ... where? There's no code that's running as a constrained execution region... 
I guess I don't fully understand how these CERs work.

edwardneal: This code (and its [pattern](https://grep.app/search?f.repo.pattern=dotnet%2FSqlClient&regexp=true&q=try%5Cs*%7B%5Cs*%7D)) is a noninterruptible CER block. This means that if a ThreadAbortException happens while the `finally` block is running, the finally block is guaranteed to run completely - in this case, to preserve the physical MARS connection.

Issue #3463, opened 2025-07-08
TdsParserStateObject.TryReadPlpBytes: System.ArgumentException: Offset and length were out of bounds


frankbuckley:
Since updating to 6.1.0-preview2.25178.5, I am seeing occassional exceptions with the following stack trace:

```
System.ArgumentException: Offset and length were out of bounds for the array or count is greater than the number of elements from index to the end of the source collection.
   at System.Buffer.BlockCopy(Array src, Int32 srcOffset, Array dst, Int32 dstOffset, Int32 count)
   at Microsoft.Data.SqlClient.TdsParserStateObject.TryReadPlpBytes(Byte[]& buff, Int32 offset, Int32 len, Int32& totalBytesRead, Boolean canContinue, Boolean writeDataSizeToSnapshot, Boolean compatibilityMode)
   at Microsoft.Data.SqlClient.TdsParser.TryReadSqlStringValue(SqlBuffer value, Byte type, Int32 length, Encoding encoding, Boolean isPlp, TdsParserStateObject stateObj)
   at Microsoft.Data.SqlClient.TdsParser.TryReadSqlValue(SqlBuffer value, SqlMetaDataPriv md, Int32 length, TdsParserStateObject stateObj, SqlCommandColumnEncryptionSetting columnEncryptionOverride, String columnName, SqlCommand command)
   at Microsoft.Data.SqlClient.SqlDataReader.TryReadColumn(Int32 i, Boolean setTimeout, Boolean allowPartiallyReadColumn, Boolean forStreaming)
   at Microsoft.Data.SqlClient.SqlDataReader.ReadAsyncExecute(Task task, Object state)
   at Microsoft.Data.SqlClient.SqlDataReader.ContinueAsyncCall[T](Task task, SqlDataReaderBaseAsyncCallContext`1 context)
   at System.Threading.Tasks.ContinuationResultTaskFromResultTask`2.InnerInvoke()
   at System.Threading.ExecutionContext.RunFromThreadPoolDispatchLoop(Thread threadPoolThread, ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location ---
   at System.Threading.ExecutionContext.RunFromThreadPoolDispatchLoop(Thread threadPoolThread, ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Threading.Tasks.Task.ExecuteWithThreadLocal(Task& currentTaskSlot, Thread threadPoolThread)
[...]

Issue #3385, opened 2025-05-30
TdsParserStateObject.ResetSnapshot() Debug.Assert failing


paulmedynski:
Failure is here:
```
    /home/paul/dev/SqlClient/dev/paul/app-context-switch-test-helper/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs(3441): error TESTERROR:
      Microsoft.Data.SqlClient.ManualTesting.Tests.DataReaderTest.CanReadXmlData (354ms): Error Message: Microsoft.VisualSt
      udio.TestPlatform.TestHost.DebugAssertException : Method Debug.Fail failed with 'snapshot._storage == null
      ', and was translated to Microsoft.VisualStudio.TestPlatform.TestHost.DebugAssertException to avoid terminating the p
      rocess hosting the test.
```

Here's the code:
```
            if (_snapshot != null)
            {
                StateSnapshot snapshot = _snapshot;
                _snapshot = null;
                Debug.Assert(snapshot._storage == null);
                snapshot.Clear();
                Interlocked.CompareExchange(ref _cachedSnapshot, snapshot, null);
            }

```

I'm not sure what the intention of the Assert() is.

paulmedynski:
All 3 of the commented-out `Debug.Assert()` statements are causing this test to fail:

`Microsoft.Data.SqlClient.ManualTesting.Tests.DataReaderTest.CanReadXmlData`

You can run all of the DataReader tests on the command line like this after uncommenting them one at a time:

```
.../ManualTests$ dotnet build
.../ManualTests$ dotnet test --no-build -f net9.0 --filter DataReaderTest
```

PR #3470, opened 2025-07-10
Fix TryReadPlpBytes throws ArgumentException

Wraith2:
fixes https://github.com/dotnet/SqlClient/issues/3463
fixes https://github.com/dotnet/SqlClient/issues/3385

In the reproduction shared in https://github.com/dotnet/SqlClient/issues/3463 multiple reads of text fields are ddone in sequence with different lengths until a particular scenario occurs. The situation that causes the error is that a multi-packet text read completes on the third packet of an async operation and the lengths of the text sections in the packets align such that the third packet  completes one text read followed by the another text read from the same packet but for a different column. 
This hard to reach set of circumstances causes the read logic to see a text read starting in a ContinueReplay state when another read has finished on the same packet and the error is caused by the first completed read leaving behind the data packet byte count used when it completed.

The fix is much easier than trying to replicate and understand the error. We simply need to clean up the data lengths and buffer when we successfully read a value. I've added this logic to the place where the bug was found and precautionarily to the other place in the codebase that does similar logic (one for bytes, one for chars).

Creating a test and runs quickly enough to be reliable in the CI is prohibitively difficult. I haven't been able to go anything that needs fewer than 1600 iterations of large text reads. In this case I feel that the fix is obvious enough that existing coverage may be enough.

This also removes the asserts pointed out as problematic in https://github.com/dotnet/SqlClient/issues/3385 i've reviewed them and they're no longer applicable, they were too stringent for the more relaxed buffer handing used when moving from replay to continue states.

@dotnet/sqlclientdevteam can I get a CI run please

/cc @frankbuckley

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
@@ -2280,11 +2280,7 @@ namespace Microsoft.Data.SqlClient
                         else if (canContinue && result == TdsOperationStatus.NeedMoreData)
                         {
                             SetSnapshotStorage(buff);
-                            if (writeDataSizeToSnapshot)
-                            {
-                                // data bytes read from the current packet must be 0 here so do not save the snapshot data size
-                                //SetSnapshotDataSize(bytesRead);
-                            }
+                            // data bytes read from the current packet must be 0 here so do not save the snapshot data size
                         }
                         return result;
                     }
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
@@ -1512,7 +1512,7 @@ namespace Microsoft.Data.SqlClient
 
                 if (writeDataSizeToSnapshot)
                 {
-                    SetSnapshotDataSize(bytesToRead);
+                    AddSnapshotDataSize(bytesToRead);
                 }
 
                 AssertValidState();
@@ -2058,7 +2058,7 @@ namespace Microsoft.Data.SqlClient
             // bool firstchunk = false;
             bool isNull = false;
 
-            Debug.Assert(_longlenleft == 0, "Out of synch length read request");
+            Debug.Assert(_longlenleft == 0, "Out of sync length read request");
             if (_longlen == 0)
             {
                 // First chunk is being read. Find out what type of chunk it is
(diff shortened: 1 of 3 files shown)


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
```diff
@@ -2281,7 +2282,8 @@ internal TdsOperationStatus TryReadPlpBytes(ref byte[] buff, int offset, int len
                             SetSnapshotStorage(buff);
                             if (writeDataSizeToSnapshot)
                             {
-                                SetSnapshotDataSize(bytesRead);
+                                // data bytes read from the current packet must be 0 here so do not save the snapshot data size
```
paulmedynski: Remove the entire if block here, and keep the comment.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
```diff
@@ -3605,12 +3607,18 @@ internal void SetSnapshotStorage(object buffer)
         /// countOfBytesCopiedFromCurrentPacket to be calculated
         /// </summary>
         /// <param name="countOfBytesCopiedFromCurrentPacket"></param>
-        internal void SetSnapshotDataSize(int countOfBytesCopiedFromCurrentPacket)
+        internal void AddSnapshotDataSize(int countOfBytesCopiedFromCurrentPacket)
         {
             Debug.Assert(_snapshot != null && _snapshot.ContinueEnabled, "_snapshot must exist to store packet data size");
             _snapshot.SetPacketDataSize(countOfBytesCopiedFromCurrentPacket);
```
paulmedynski, line 3609: Should this get the `_snapshot?.` treatment like below?

paulmedynski (changes requested):
A couple of minot comments.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
Wraith2: No, this is just a rename and we know that this function is never called when a snapshot isn't present. If you like we can add the ? just in case it ever is but it's not needed. Arguably we don't need it in the Clear method either but clearing feels like something we might want to scatter around as a precaution whereas setting the value is something you have to be careful about. 
Let me know what you feel, I'm happy to change if you think it needs it.

paulmedynski: I was just confirming.  Some `#nullable enable` love would help here, but that's a problem for another day :)

Wraith2: I got nullable into one file in the managed networking implementation. Doing it in TdsParser though... that's going to be a big job.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
Wraith2: Done.

paulmedynski (approved):

PR #3471, opened 2025-07-10
WIP | Create Extensions package

paulmedynski:
## Description

*Work in Progress* - This PR is a draft while I get the CI machinery working.

As part of the Azure split work, a new Extensions package is being created that contains types shared between MDS and its future extensions (Azure will be the first).  This PR creates that package (with no meaningful content) to setup the MDS dependency, and get testing and CI setup accordingly.

I'm also experimenting with simplified .slnx files and some project directory structure changes.

## Issues

The first step towards #1108.

## Testing

- Unit tests for the sample class to prove that CI can run them successfully.
- There are no MDS tests to prove out the new dependency yet.  Those will naturally happen when the first real bit of Azure functionality is moved out into the new Azure extension package.


NuGet.config
```diff
@@ -2,10 +2,16 @@
...
+    <!-- We use a curated, vetted ADO feed for our dependencies. -->
+    <add key="curated" value="https://sqlclientdrivers.pkgs.visualstudio.com/public/_packaging/sqlclient/nuget/v3/index.json" />
+
+    <!-- During development, our sibling packages may only be available locally. -->
+    <add key="local" value="packages/" />
   </packageSources>
   <auditSources>
     <clear />
+
+    <!-- There is no curated, vetted ADO feed for autiding, so we use the official source. -->
```
paulmedynski: auditing

eng/pipelines/dotnet-sqlclient-ci-core.yml
```diff
@@ -79,11 +83,29 @@ variables:
     value: 'Azure Pipelines'
 
 stages:
+
+  # Build the Extensions package, and publish it to the pipeline artifacts
+  # under the defaule name 'ExtensionsPackageArtifacts'.
```
paulmedynski: default

eng/pipelines/dotnet-sqlclient-ci-core.yml
```diff
@@ -327,9 +349,8 @@ stages:
             # ways to detect if they were present) won't run consistently across forks and non-forks.
             ${{ if eq(variables['system.pullRequest.isFork'], 'False') }}:
               AADPasswordConnectionString: $(AAD_PASSWORD_CONN_STR)
-            AADServicePrincipalId: $(AADServicePrincipalId)
-            ${{ if eq(variables['system.pullRequest.isFork'], 'False') }}:
```
paulmedynski, line 331: My YAML linter was complaining about duplicate keys because these `${{if ...}` lines are the same.  Similar changes below for the enclave test configurations.

eng/pipelines/stages/ci-build-extensions-package-stage.yml
```diff
@@ -0,0 +1,148 @@
...
+  - stage: extensions_package_stage
+    displayName: CI Stage for Extensions Package
+
+    jobs:
+
+    - job: extensions_package_job
+      displayName: CI Job for Extensions Package
+      pool:
+        name: ADO-CI-1ES-Pool
+        vmImage: ubuntu-latest
```
paulmedynski, line 145: I'm only building on Linux for now.  We can expand this to also build on Windows if we think that is necessary.

eng/pipelines/stages/ci-build-extensions-package-stage.yml
```diff
@@ -0,0 +1,148 @@
...
+
+      - task: DotNetCoreCLI@2
+        displayName: Build Solution
+        inputs:
+          command: custom
+          custom: build
+          projects: $(project)
+          arguments: $(commonArguments) --no-restore
+
+      - task: DotNetCoreCLI@2
```
paulmedynski, line 322: Tests are only run on Linux for now.  I will expand this to run a job on Windows that pulls the artifacts and runs tests there as well.

tools/props/Versions.props
```diff
@@ -1,29 +1,61 @@
...
+  <PropertyGroup>
+    <MdsVersionDefault>6.1.0</MdsVersionDefault>
+
+    <!--
+      TODO: The remaining properties in this group should be renamed to be
+      specific to MDS, and then used in the MDS project.  As-is, these names are
+      used by the build tooling and may be unintentionally included in other
+      (non-MDS) projects.
+    -->
+    <NugetPackageVersion Condition="'$(NugetPackageVersion)' == ''">$(MdsVersionDefault).$(BuildNumber)-dev</NugetPackageVersion>
```
paulmedynski, line 42: Minor tweak here - I added `$(BuildNumber)` - not sure if that's helpful, but it matches how the Extensions package does things.

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.csproj
paulmedynski, line 1: Missing associated MDS .nuspec file changes.

eng/pipelines/stages/ci-build-extensions-package-stage.yml
```diff
@@ -0,0 +1,351 @@
...
+      - task: UseDotNet@2
+        displayName: Install .NET 9.0 SDK
+        inputs:
+          packageType: sdk
+          version: 9.x
+
+      - task: UseDotNet@2
+        displayName: Install .NET 8.0 Runtime
+        inputs:
+          packageType: sdk
```
Frulfump: Display name says "Runtime" but `packageType` is `sdk` (which will install both the sdk and the runtime so it still works but might shorten the install time a bit if it's `runtime`)

paulmedynski: Nice catch!

src/Microsoft.Data.SqlClient.Extensions/Extensions/Extensions.slnx
paulmedynski, line 1: I decided to give .slnx format a try, and so far the tooling supports it.

paulmedynski (commented):
Adding commentary for reviewers.

edwardneal:
Thanks for this - I've got no comments or feedback on the design, other than the cosmetic note that Microsoft.Extensions.Logging and Microsoft.Extensions.DependencyInjection rely upon packages with the ".Abstractions" suffix, rather than ".Extensions".

Interestingly, there's also a sizable improvement in AOT application size as a result of removing the Azure components from the main package. In my sample application, doing so cut the size of a 13.2MB application by about a third.

paulmedynski:
> Thanks for this - I've got no comments or feedback on the design, other than the cosmetic note that Microsoft.Extensions.Logging and Microsoft.Extensions.DependencyInjection rely upon packages with the ".Abstractions" suffix, rather than ".Extensions".

I like the `Abstractions` name better than just `Extensions`.  I'll discuss with my colleagues after I get back from vacation before I do a huge renaming :)  I'm thinking:

|Package|Description|
|-|-|
|Microsoft.Data.SqlClient|The main MDS package.|
|Microsoft.Data.SqlClient.Abstractions|Shared types, classes, etc; used by the other packages.|
|Microsoft.Data.SqlClient.Extensions.Azure|Functionality that relies on Azure dependencies.|

Is there any benefit in `Microsoft.DatalSqlClient.Extensions.Abstractions` versus the above?

> Interestingly, there's also a sizable improvement in AOT application size as a result of removing the Azure components from the main package. In my sample application, doing so cut the size of a 13.2MB application by about a third.

Excellent - more tangible benefits!

Frulfump:
Maybe this issue would cause changes you would want to incorporate here?
https://github.com/dotnet/SqlClient/issues/3509

paulmedynski:
Closing in favour of #3567 .

PR #3473, opened 2025-07-14
Merge | SqlCommand (Part 1)

benrr101:
## Description
This PR is a collection of changes to both SqlCommand implementations to bring them closer together. Although the bulk of SqlCommand is the same, it is a huge class and very hard to reconcile all the differences cleanly. I have no idea what's the best way to handle this merge, so I figure I'll do it in chunks that will be digestible.

In this installment, I've tried to tackle the small differences that are just really annoying to go through. These changes includes:
* Aligning method arguments that are `[CallerMemberName]` and calls to such methods
* Sync formatting for method attributes
* Sync method signatures:
  * TriggerInternalEndAndRetryIfNecessary
  * PrepareForTransparentEncryption
  * TryFetchInputParameterEncryptionInfo
  * ReadDescribeEncryptionParameterResults
  * RunExecuteNonQueryTds
  * RunExecuteReaderTdsWithTransparentParameterEncryption
  * RunExecuteReaderTds
  * OnDone
* Align names
  * KeysToBeSentToEnclave
  * RequiresEnclaveComputations
  * ShouldCacheEncryptionMetadata
* Align visibility
  * CancelIgnoreFailureCallback
  * CancelIgnoreFailure
* Align ordering
  * ExecuteReader methods
  * CancelIgnoreFailureCallback
  * CancelIgnoreFailure
  * Clone, ICloneable.Clone
* Remove unsed debug methods CompletePendingReadWithSuccess CompletePendingReadWithFailure
* Sync method body expression formatting
  * BeginExecuteReader*
  * ExecuteNonQueryAsync*
  * ExecuteReaderAsync*
  * ExecuteXmReaderAsync*
* Sync _prepareHandle type
* Rename CachedAsyncState to AsyncState, change name of property to CachedAsyncState, replace usages of _cachedAsyncState to the property **this could be dangerous**
* Sync async parameters/variables to `isAsync`
* Sync inRetry parameters to `isRetry`
* Sync event/trace/correlation messages to netcore style
* Sync/fix typos, low-risk formatting changes

## Testing
Code changes are pretty low stakes syncing between the two implementations. CI should be suitable to vaidate.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
@@ -1115,8 +1115,16 @@ namespace Microsoft.Data.SqlClient
             base.Dispose(disposing);
         }
 
-        private SqlDataReader RunExecuteReaderWithRetry(CommandBehavior cmdBehavior, RunBehavior runBehavior, bool returnStream, [CallerMemberName] string method = "")
-            => RetryLogicProvider.Execute(this, () => RunExecuteReader(cmdBehavior, runBehavior, returnStream, method));
+        private SqlDataReader RunExecuteReaderWithRetry(
+            CommandBehavior cmdBehavior,
+            RunBehavior runBehavior,
+            bool returnStream,
+            [CallerMemberName] string method = "")
+        {
+            return RetryLogicProvider.Execute(
+                this,
+                () => RunExecuteReader(cmdBehavior, runBehavior, returnStream, method));
+        }
 
         /// <include file='../../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlCommand.xml' path='docs/members[@name="SqlCommand"]/ExecuteScalar/*'/>
         public override object ExecuteScalar()
@@ -1137,9 +1145,9 @@ namespace Microsoft.Data.SqlClient
                 {
                     statistics = SqlStatistics.StartTimer(Statistics);
                     WriteBeginExecuteEvent();
-                    SqlDataReader ds = IsProviderRetriable ?
-                        RunExecuteReaderWithRetry(0, RunBehavior.ReturnImmediately, returnStream: true) :
-                        RunExecuteReader(0, RunBehavior.ReturnImmediately, returnStream: true, method: nameof(ExecuteScalar));
+                    SqlDataReader ds = IsProviderRetriable
+                        ? RunExecuteReaderWithRetry(0, RunBehavior.ReturnImmediately, returnStream: true)
+                        : RunExecuteReader(0, RunBehavior.ReturnImmediately, returnStream: true);
                     success = true;
                     return CompleteExecuteScalar(ds, _batchRPCMode);
                 }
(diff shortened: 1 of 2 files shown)


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
```diff
@@ -3303,7 +3378,7 @@ internal void DeriveParameters()
             }
 
             // validate that we have a valid connection
-            ValidateCommand(false /*not async*/, nameof(DeriveParameters));
```
mdaigle, line 3306: Looks like nameof(DeriveParameters) got dropped in netfx and netcore

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlCommand.cs
```diff
@@ -2356,8 +2356,8 @@ private SqlDataReader ExecuteReaderWithRetry(CommandBehavior behavior, string me
                 {
                     statistics = SqlStatistics.StartTimer(Statistics);
                     return IsProviderRetriable ?
-                            ExecuteReaderWithRetry(CommandBehavior.Default, nameof(ExecuteReader)) :
```
mdaigle, line 2293: methodName dropped

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
```diff
@@ -2374,14 +2362,8 @@ private IAsyncResult BeginExecuteReaderInternal(CommandBehavior behavior, AsyncC
...
                         localCompletion,
-                        endFunc: static (SqlCommand command, IAsyncResult asyncResult, bool isInternal, string endMethod) =>
-                        {
-                            return command.InternalEndExecuteReader(asyncResult, isInternal, endMethod);
-                        },
-                        retryFunc: static (SqlCommand command, CommandBehavior behavior, AsyncCallback callback, object stateObject, int timeout, bool inRetry, bool asyncWrite) =>
-                        {
-                            return command.BeginExecuteReaderInternal(behavior, callback, stateObject, timeout, inRetry, asyncWrite);
-                        },
+                        InternalEndExecuteReader,
```
mdaigle: I think this might result in a new closure to capture SqlCommand?

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlCommand.cs
```diff
@@ -356,16 +357,14 @@ internal void SetAsyncReaderState(SqlDataReader ds, RunBehavior runBehavior, str
...
+        // @TODO: This is never null, so we can remove the null checks from usages of it.
+        private AsyncState CachedAsyncState
         {
             get
             {
-                if (_cachedAsyncState == null)
-                {
-                    _cachedAsyncState = new CachedAsyncState();
-                }
+                _cachedAsyncState ??= new AsyncState();
```
mdaigle, line 367: At this point, maybe we just initialize it at construction time. It seems like it's always initialized one way or another. That way there's no possibility of accidentally referencing the backing field.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
benrr101: This is intentional, since the argument is a [CallerMemberName] argument, so it will automatically fill it with the name of the method that called it. The only time we are providing it now is when the call is deeper than the name we want to pass

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
benrr101: Hmmmmmmmm... I'm not entirely sure about that. Let me do a bit of research and find out.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlCommand.cs
benrr101: I think I agree with you. The theory I guess is that we want to initialize the state when it's needed, rather than at command construction time. While I'm working on subsequent merges, I see this pattern being used in several places, and I'm not sure if we want to continue using that pattern or just initialize on construction.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlCommand.cs
benrr101: This is intentional, since the argument is a [CallerMemberName] argument,

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
benrr101: Claude says:
> Your colleague is absolutely right to raise this concern. 
😂 

I will address this by reintroducing the static lambdas. However, I'll add a couple todos for how we can clean up these calls a bit. Right now my goal is just to merge - and I'm using every ounce of restraint I have to not just go full cowboy 🤠

mdaigle (approved):

PR #3488, opened 2025-07-17
Refactor Test TDS Servers. Expand functional testing of connection pooling and transient failures.

mdaigle:
## Description
The TDS.Servers project provides mocked server implementations for a variety of use cases. The construction and endpoint initialization flows for these mock servers was complicated and resulted in duplicated code across TDS.Servers, FunctionalTests, and ManualTests. This PR consolidates endpoint initialization and makes test server constructor parameters more explicit (no more defaults buried a few layers deep).

One downside of this approach is that it's now slightly harder to get the connection string for a test server. Before, some servers had a ConnectionString property you could access. The issue with these connection strings is they had a lot of non-standard default behavior coded into them at the time they were constructed. This makes it hard to understand which properties you're actually using when connecting to a test server. With my changes, you now need to manually construct a connection string using SqlConnectionStringBuilder and explicitly set connection parameters. Any parameters you don't set use the stock connection string defaults.

## Testing

I added some testing around pool invalidation during failover and routing behavior during transient errors.

### DIFF
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectionTestWithSSLCert/CertificateTestWithTdsServer.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectionTestWithSSLCert/CertificateTestWithTdsServer.cs
@@ -15,6 +15,7 @@ using Microsoft.Data.SqlClient.ManualTesting.Tests.DataCommon;
 using Microsoft.SqlServer.TDS.Servers;
 using Microsoft.Win32;
 using Xunit;
+#nullable enable
 
 namespace Microsoft.Data.SqlClient.ManualTesting.Tests
 {
@@ -132,11 +133,7 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests
 
             using TdsServer server = new TdsServer(new TdsServerArguments
             {
-                #if NET9_0_OR_GREATER
-                EncryptionCertificate = X509CertificateLoader.LoadPkcs12FromFile(s_fullPathToPfx, "nopassword", X509KeyStorageFlags.UserKeySet),
-                #else
-                EncryptionCertificate = new X509Certificate2(s_fullPathToPfx, "nopassword", X509KeyStorageFlags.UserKeySet),
-                #endif
+                EncryptionCertificate = GetEncryptionCertificate(s_fullPathToPfx, "nopassword", X509KeyStorageFlags.UserKeySet),
                 EncryptionProtocols = connectionTestParameters.EncryptionProtocols,
                 Encryption = connectionTestParameters.TdsEncryptionType,
             });
@@ -237,6 +234,22 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests
             }
         }
 
+        /// <summary>
+        /// Loads the specified certificate.
+        /// </summary>
+        /// <param name="fileName">The full path of the certificate.</param>
+        /// <param name="password">The certificate's password.</param>
+        /// <param name="keyStorageFlags">Key storage flags to apply when loading the certificate</param>
+        /// <returns>An <see cref="X509Certificate2"/> instance.</returns>
+        private X509Certificate2 GetEncryptionCertificate(string fileName, string? password, X509KeyStorageFlags keyStorageFlags)
+        {
+#if NET9_0_OR_GREATER
+            return X509CertificateLoader.LoadPkcs12FromFile(fileName, password, keyStorageFlags);
+#else
+            return new X509Certificate2(fileName, password, keyStorageFlags);
+#endif
+        }
+
         private void RemoveCertificate()
         {
             string thumbprint = File.ReadAllText(s_fullPathTothumbprint);
@@ -255,7 +268,7 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests
 
         private static void RemoveForceEncryptionFromRegistryPath(string registryPath)
         {
-            RegistryKey key = Registry.LocalMachine.OpenSubKey(registryPath, true);
+            RegistryKey? key = Registry.LocalMachine.OpenSubKey(registryPath, true);
             key?.SetValue("ForceEncryption", 0, RegistryValueKind.DWord);
             key?.SetValue("Certificate", "", RegistryValueKind.String);
             ServiceController sc = new($"{s_instanceNamePrefix}{s_instanceName}");
--- a/src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionRoutingTests.cs
+++ b/src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionRoutingTests.cs
@@ -48,15 +48,9 @@ namespace Microsoft.Data.SqlClient.UnitTests.SimulatedServerTests
                 Encrypt = false,
             };
             using SqlConnection connection = new(builder.ConnectionString);
-            try
-            {
-                // Act
-                connection.Open();
-            }
-            catch (Exception e)
-            {
-                Assert.Fail(e.Message);
-            }
+
+            // Act
+            connection.Open();
 
             // Assert
             Assert.Equal(ConnectionState.Open, connection.State);
--- a/src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/AuthenticatingTdsServer.cs
+++ b/src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/AuthenticatingTdsServer.cs
@@ -17,8 +17,8 @@ namespace Microsoft.SqlServer.TDS.Servers
         /// <summary>
         /// Initialization constructor
         /// </summary>
-        public AuthenticatingTdsServer() :
-            this(new AuthenticatingTdsServerArguments())
+        public AuthenticatingTdsServer() 
+            : this(new AuthenticatingTdsServerArguments())
         {
         }
 


src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlConnectionBasicTests.cs
```diff
@@ -153,6 +183,63 @@ public void TransientFaultDisabledTest(uint errorCode)
...
+            };
+
+            using SqlConnection connection = new(builder.ConnectionString);
+            connection.Open();
+
+            // Act
+            initialServer.SetErrorBehavior(true, errorCode, "Transient fault occurred.");
+            using SqlConnection failoverConnection = new(builder.ConnectionString);
+            // Should fail over to the failover server
+            failoverConnection.Open();
```
paulmedynski: Don't we need to Assert() that the connection connected where we expect it to?  Something like:

`Assert.Equal(failoverConnection.RemoteEndpoint, failoverServer.Endpoint);`

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionTests.cs
```diff
@@ -310,8 +397,14 @@ public void ConnectionTestValidCredentialCombination()
         public void ConnectionTimeoutTest(int timeout)
         {
             // Start a server with connection timeout from the inline data.
-            using TestTdsServer server = TestTdsServer.StartTestServer(false, false, timeout);
-            using SqlConnection connection = new SqlConnection(server.ConnectionString);
+            //TODO: do we even need a server for this test?
```
paulmedynski, line 466: IMO we shouldn't need a server for most of these tests.  We're missing a way to intercept outbound TDS packets and inject inbound TDS packets.  If we had that, we wouldn't need TdsServer for these kinds of unit tests, and we could clearly see in each test exactly what was being tested.  Right now, it isn't obvious what TdsServer is doing or how it is verifying our expectations.  This is something I think we need to work towards - nothing for this PR to address.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlConnectionReadOnlyRoutingTests.cs
```diff
@@ -136,5 +140,54 @@ public async Task AsyncConnectionRoutingLimit()
...
+            catch (Exception e)
+            {
+                Assert.Fail(e.Message);
+            }
+
+            // Assert
+            Assert.Equal(ConnectionState.Open, connection.State);
+
+            // Failures should prompt the client to return to the original server, resulting in a login count of 2
+            Assert.Equal(2, router.PreLoginCount);
```
paulmedynski: We haven't verified what Open() actually did though - just what the side-effects on our test servers were.  Did Open() make a bunch of unexpected failed connection attempts elsewhere?  Which test server, if any, is it currently connected to?  We don't know.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/AuthenticatingTDSServer.cs
paulmedynski, line 1: Don't we need to rename these files to match their classes?

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/GenericTDSServer.cs
```diff
@@ -69,36 +72,42 @@ public delegate void OnAuthenticationCompletedDelegate(
...
+        protected T Arguments { get; set; }
 
         /// <summary>
         /// Query engine instance
         /// </summary>
         protected QueryEngine Engine { get; set; }
 
         /// <summary>
-        /// Default constructor
+        /// Counts unique pre-login requests to the server.
```
paulmedynski: The term `unique` here implies that duplicates are a thing, and they're not counted.  Does this just count each pre-login packet received, or does it ignore some of them?

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/GenericTDSServer.cs
```diff
@@ -158,7 +179,7 @@ public virtual TDSMessageCollection OnPreLoginRequest(ITDSServerSession session,
             TDSPreLoginToken preLoginToken = new TDSPreLoginToken(Arguments.ServerVersion, serverResponse, false); // TDS server doesn't support MARS
 
             // Cache the received Nonce into the session
-            (session as GenericTDSServerSession).ClientNonce = preLoginRequest.Nonce;
+            (session as GenericTdsServerSession).ClientNonce = preLoginRequest.Nonce;
```
paulmedynski: Should we cast this once and re-use it?

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/TdsServer.cs
```diff
@@ -0,0 +1,31 @@
...
+{
+    public class TdsServer : GenericTdsServer<TdsServerArguments>
+    {
+        /// <summary>
+        /// Default constructor
+        /// </summary>
+        public TdsServer() : this(new TdsServerArguments())
+        {
+        }
+        /// <summary>
```
paulmedynski, line 16: Missing a blank line here.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/TdsServer.cs
```diff
@@ -0,0 +1,31 @@
...
+        /// Constructor with arguments
+        /// </summary>
+        public TdsServer(TdsServerArguments arguments) : base(arguments)
```
paulmedynski, line 19: You won't need the default constructor if you:

`public TdsServer(TdsServerArguments arguments = new()) : base(arguments)`

paulmedynski (changes requested):
Looks good to me, with a few suggestions.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/TdsServer.cs
mdaigle: default arguments need to be compile time constant or a ValType.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.EndPoint/TDSServerEndPointConnection.cs
```diff
@@ -79,7 +80,7 @@ public abstract class ServerEndPointConnection
         /// <summary>
         /// The flag indicates whether server is being stopped
```
paulmedynski: This property isn't really a flag any more.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/TransientFaultTdsServer.cs
```diff
@@ -49,52 +47,45 @@ private static string GetErrorMessage(uint errorNumber)
             return "Unknown server error occurred";
         }
 
-        /// <summary>
-        /// Handler for login request
-        /// </summary>
+            /// <summary>
```
paulmedynski: Extra indent.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/TdsServer.cs
paulmedynski: Frown - that's an unfortunate restriction.  Can TdsServerArguments be a struct?

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/TransientTimeoutTdsServer.cs
```diff
@@ -0,0 +1,82 @@
...
+            Arguments.IsEnabledPermanentTimeout = isEnabledPermanentTimeout;
+            Arguments.SleepDuration = sleepDuration;
+        }
+
+        /// <summary>
+        /// Handler for login request
+        /// </summary>
+        public override TDSMessageCollection OnLogin7Request(ITDSServerSession session, TDSMessage request)
+        {
+            // Check if we're still going to raise transient error
```
paulmedynski, line 40: How/where does this cause a transient error?

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/TransientTimeoutTdsServerArguments.cs
```diff
@@ -0,0 +1,33 @@
...
+        /// Flag to consider when simulating a timeout on each request.
+        /// </summary>
+        public bool IsEnabledPermanentTimeout { get; set; }
+
+        /// <summary>
+        /// Constructor to initialize
+        /// </summary>
+        public TransientTimeoutTdsServerArguments()
+        {
+            SleepDuration = TimeSpan.FromSeconds(0);
```
paulmedynski: Can you just set these on the properties, and avoid a constructor?

```
public TimeSpan SleepDuration { get; set; } = TimeSpan.FromSeconds(0);
```

That makes it easy to see the default value for each property.

paulmedynski (changes requested):
A few tidying-up suggestions.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/TransientTimeoutTdsServer.cs
paulmedynski: I think this server is intended to delay responses, eliciting a driver-side timeout.  So this isn't really a timeout server - it's a delay server.

You could also make the count of delayed requests configurable instead of hardcoding it to 1.  Then you could test a variety of retry counts/mechanisms in the driver.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/TdsServer.cs
mdaigle: Not easily. You can't inherit from a struct, so we would have to flatten all of these server-type specific arguments down into one flat struct. It would work but wouldn't be super usable.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.EndPoint/TDSServerEndPointConnection.cs
```diff
@@ -78,9 +78,9 @@ public abstract class ServerEndPointConnection : IDisposable
         protected TcpClient Connection { get; set; }
 
         /// <summary>
-        /// The flag indicates whether server is being stopped
+        /// Cancellation token source for managing cancellation of the processing thread
         /// </summary>
-        protected CancellationTokenSource CancellationTokenSource { get; set; }
+        private CancellationTokenSource CancellationTokenSource = new CancellationTokenSource();
```
paulmedynski: I think we use `_cancellationTokenSource` for private fields.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/TransientDelayTdsServer.cs
```diff
@@ -59,11 +58,11 @@ public override TDSMessageCollection OnLogin7Request(ITDSServerSession session,
             return base.OnLogin7Request(session, request);
         }
 
+        /// <inheritdoc/>
         public override TDSMessageCollection OnSQLBatchRequest(ITDSServerSession session, TDSMessage message)
         {
-            // Check if we're still going to raise transient error
             if (Arguments.IsEnabledPermanentTimeout ||
-                (Arguments.IsEnabledTransientTimeout && RequestCounter < 1)) // Fail first time, then connect
+                (Arguments.IsEnabledTransientTimeout && RequestCounter < 1))
```
paulmedynski: `RequestCounter < Arguments.RepeatCount` ?

paulmedynski (changes requested):
A couple of suggestions.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/GenericTdsServer.cs
```diff
@@ -142,8 +159,11 @@ public virtual void CloseSession(ITDSServerSession session)
         /// </summary>
         public virtual TDSMessageCollection OnPreLoginRequest(ITDSServerSession session, TDSMessage request)
         {
+            Interlocked.Increment(ref _preLoginCount);
+
             // Inflate pre-login request from the message
             TDSPreLoginToken preLoginRequest = request[0] as TDSPreLoginToken;
+            GenericTdsServerSession genericTdsServerSession = session as GenericTdsServerSession;
```
paulmedynski, line 173: Why are we casting?  Should we take GenericTdsServerSession instead?  I'm not sure what ITDSServerSession is doing for us.

src/Microsoft.Data.SqlClient/tests/UnitTests/ADPHelper.cs
```diff
@@ -0,0 +1,29 @@
...
+        List<string> _originalAzureSqlServerEndpoints;
+
+        internal ADPHelper()
+        {
+            _originalAzureSqlServerEndpoints = [.. ADP.s_azureSqlServerEndpoints];
+        }
+
+        internal void AddAzureSqlServerEndpoint(string endpoint)
+        {
+            ADP.s_azureSqlServerEndpoints.Add(endpoint);
```
paulmedynski, line 20: Does this work when multiple tests use it in parallel?

paulmedynski (changes requested):
Still some outstanding requests from earlier PRs as well.  I will go back through them and see if they have been addressed.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlConnectionBasicTests.cs
mdaigle: All tests were updated to assert the data source of the connection.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlConnectionReadOnlyRoutingTests.cs
mdaigle: All test cases were updated to verify call counts on each relevant server and the ultimate data source of the connection.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/GenericTdsServer.cs
mdaigle: There are several cases of unnecessary inheritance within the test server classes. We can remove them as each only has a single subclass/implementer. Just trying to limit the scope of this PR. There's a lot more we could do in this space :)

src/Microsoft.Data.SqlClient/tests/UnitTests/ADPHelper.cs
mdaigle: No, this was one of the sources of the flakiness of the tests. xunit test collections and categories are to fragile to coordinate this kind of static state manipulation. I would either need 1. a refactor of how the azure endpoints are stored, to make them connection instance specific or 2. actually rewrite these tests as unit tests with the logic under test more directly available.

Right now, I need to draw the line on these tests, but I'll evaluate these options moving forward.

paulmedynski:
Two small comments remain.  Should this convereted from a draft now?

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -1713,8 +1713,11 @@ private void LoginNoFailover(ServerInfo serverInfo,
                     }
 
                     if (_parser == null
-                        || TdsParserState.Closed != _parser.State
                         || IsDoNotRetryConnectError(sqlex)
+                        // If state != closed, indicates that the parser encountered an error while processing the
```
benrr101: Appreciate the comment here - but two requests to improve readability?
* Can we put the comment outside the if statement?
* Can we flip the `TdsParserState.Closed != _parser.State` to `_parser.State != TdsParserState.Closed`
* Can we maintain having `parser == null` and `parser.State != TdsParserState.Closed` conditions next to each other?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/ConnectionString/DbConnectionStringDefaults.cs
```diff
@@ -57,7 +57,7 @@ internal static class DbConnectionStringDefaults
         
         #if NETFRAMEWORK
         internal const bool ConnectionReset = true;
-        internal static readonly bool TransparentNetworkIpResolution = !LocalAppContextSwitches.DisableTnirByDefault;
+        internal static bool TransparentNetworkIpResolution => !LocalAppContextSwitches.DisableTnirByDefault;
```
benrr101, line 60: Semantically these are different (though admittedly I'm not sure what the compiler will do) - making this a property means it will calculate it each time the property is called, while leaving it a readonly field it's calculated once on static construction. I suppose if the idea is to make sure we load it from the context switches each time because it might change, that's a valid change. But as it stands now, it seems more performant to have this be a readonly field.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/AdapterUtil.cs
```diff
@@ -787,7 +788,7 @@ internal static Version GetAssemblyVersion()
         /// <remarks>This array includes endpoint URLs for Azure SQL in global, Germany, US Government,
         /// China, and Fabric environments. These endpoints are used to identify and interact with Azure SQL services 
         /// in their respective regions or environments.</remarks>
-        internal static readonly string[] s_azureSqlServerEndpoints = { AZURE_SQL,
+        internal static readonly List<string> s_azureSqlServerEndpoints = new() { AZURE_SQL,
```
benrr101, line 791: nit: first element of a multi-line definition like this should go on the next line

benrr101: But it's a nit - this file needs a lot of cleanup

benrr101, line 791: I'm assuming the goal of making this a List is to add/remove elements during testing?
I'm ... not sure about that. Seems like the overhead for test-only scenarios. Especially with the "make read-only code below")

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/AdapterUtil.cs
```diff
@@ -823,11 +824,11 @@ internal static bool IsAzureSynapseOnDemandEndpoint(string dataSource)
         
         internal static bool IsAzureSqlServerEndpoint(string dataSource)
         {
-            return IsEndpoint(dataSource, s_azureSqlServerEndpoints);
+            return IsEndpoint(dataSource, s_azureSqlServerEndpoints.AsReadOnly());
         }
 
         // This method assumes dataSource parameter is in TCP connection string format.
-        private static bool IsEndpoint(string dataSource, string[] endpoints)
+        private static bool IsEndpoint(string dataSource, IReadOnlyList<string> endpoints)
```
benrr101: Do we even need to pass in the list?

src/Microsoft.Data.SqlClient/tests/FunctionalTests/LocalizationTest.cs
```diff
@@ -55,9 +56,9 @@ private string GetLocalizedErrorMessage(string culture)
             Thread.CurrentThread.CurrentCulture = new CultureInfo(culture);
             Thread.CurrentThread.CurrentUICulture = new CultureInfo(culture);
 
-            using TestTdsServer server = TestTdsServer.StartTestServer();
-            var connStr = server.ConnectionString;
-            connStr = connStr.Replace("localhost", "dummy");
+            using TdsServer server = new TdsServer(new TdsServerArguments() { });
```
benrr101: the `{ }` seem redundant here

src/Microsoft.Data.SqlClient/tests/FunctionalTests/LocalizationTest.cs
```diff
@@ -55,9 +56,9 @@ private string GetLocalizedErrorMessage(string culture)
...
+            server.Start();
+            var connStr = new SqlConnectionStringBuilder() { DataSource = $"dummy,{server.EndPoint.Port}" }.ConnectionString;
```
benrr101: It'd be nice if there were a way to preserve the old behavior since it's much tidier. Maybe like a "GetConnectionString" method that takes a SqlConnectionStringBuilder. I don't know exactly how it could be done. What about just a DataSource property?

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectionTestWithSSLCert/CertificateTestWithTdsServer.cs
```diff
@@ -129,18 +130,23 @@ private void ConnectionTest(ConnectionTestParameters connectionTestParameters)
...
-#else
-                new X509Certificate2(s_fullPathToPfx, "nopassword", X509KeyStorageFlags.UserKeySet),
-#endif
-                encryptionProtocols: connectionTestParameters.EncryptionProtocols,
-                encryptionType: connectionTestParameters.TdsEncryptionType);
-
-            builder = new(server.ConnectionString)
+            using TdsServer server = new TdsServer(new TdsServerArguments
             {
+                #if NET9_0_OR_GREATER
```
benrr101: Now might be a good time to introduce a helper for this :)

src/Microsoft.Data.SqlClient/tests/ManualTests/TracingTests/DiagnosticTest.cs
```diff
@@ -717,10 +717,19 @@ private static void CollectStatisticsDiagnostics(Action<string> sqlOperation, bo
             {
 
                 Console.WriteLine(string.Format("Test: {0} Enabled Listeners", methodName));
-                using (var server = TestTdsServer.StartServerWithQueryEngine(new DiagnosticsQueryEngine(), enableLog: enableServerLogging, methodName: methodName))
+
+                using (var server = new TdsServer(new DiagnosticsQueryEngine(), new TdsServerArguments()))
```
benrr101, line 721: This only appears to be using a using block b/c there's console logging that happens when the server is disposed. You know about how these tests are being done, but I generally think it's not good to log to console during a test. So we could delete those logs, and make this a using statement.

benrr101: (also the "new experience" doesn't let me make comments on unchanged lines?)

src/Microsoft.Data.SqlClient/tests/ManualTests/TracingTests/DiagnosticTest.cs
```diff
@@ -906,11 +915,17 @@ private static async Task CollectStatisticsDiagnosticsAsync(Func<string, Task> s
             using (DiagnosticListener.AllListeners.Subscribe(diagnosticListenerObserver))
             {
                 Console.WriteLine(string.Format("Test: {0} Enabled Listeners", methodName));
-                using (var server = TestTdsServer.StartServerWithQueryEngine(new DiagnosticsQueryEngine(), methodName: methodName))
+                using (var server = new TdsServer(new DiagnosticsQueryEngine(), new TdsServerArguments()))
```
benrr101, line 918: Another instance where unnecessary console logging prevents us from using using statements.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.EndPoint/TDSServerEndPointConnection.cs
```diff
@@ -171,7 +155,7 @@ private void _ConnectionHandler()
                 PrepareForProcessingData(rawStream);
 
                 // Process the packet sequence
-                while (Connection.Connected && !StopRequested)
+                while (Connection.Connected)
                 {
                     // Check incoming buffer
                     if (rawStream.DataAvailable)
```
benrr101, line 169: Probably out of scope for this PR, but we can definitely make reading from the more async than polling it for available data.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/AuthenticatingTdsServer.cs
```diff
@@ -12,20 +12,20 @@ namespace Microsoft.SqlServer.TDS.Servers
...
     /// </summary>
-    public class AuthenticatingTDSServer : GenericTDSServer
+    public class AuthenticatingTdsServer : GenericTdsServer<AuthenticatingTdsServerArguments>
     {
         /// <summary>
         /// Initialization constructor
         /// </summary>
-        public AuthenticatingTDSServer() :
-            this(new AuthenticatingTDSServerArguments())
+        public AuthenticatingTdsServer() :
```
benrr101: nit: I thought the `:` was supposed to go on the next line 🤔

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/AuthenticatingTdsServer.cs
```diff
@@ -39,10 +39,10 @@ public override TDSMessageCollection OnLogin7Request(ITDSServerSession session,
             TDSLogin7Token loginRequest = request[0] as TDSLogin7Token;
 
             // Check if arguments are of the authenticating TDS server
-            if (Arguments is AuthenticatingTDSServerArguments)
+            if (Arguments is AuthenticatingTdsServerArguments)
```
benrr101: pls use the `Arguments is AuthenticatingTdsServerArguments serverArguments` form instead of separate check/cast

benrr101: Also ... can we rename ServerArguments to the propert camelCase?

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/AuthenticatingTdsServerArguments.cs
```diff
@@ -7,43 +7,31 @@ namespace Microsoft.SqlServer.TDS.Servers
...
     /// Arguments for authenticating TDS Server
     /// </summary>
-    public class AuthenticatingTDSServerArguments : TDSServerArguments
+    public class AuthenticatingTdsServerArguments : TdsServerArguments
     {
         /// <summary>
         /// Type of the application intent filter
         /// </summary>
-        public ApplicationIntentFilterType ApplicationIntentFilter { get; set; }
+        public ApplicationIntentFilterType ApplicationIntentFilter = ApplicationIntentFilterType.All;
```
benrr101: Keep in mind this changes it from a property to a field... Generally public fields are not good practice, so I'd recommend keeping it a property and setting the default value, like:

```
public ApplicationIntentFilterType ApplicationIntentFilter { get; set; } = ApplicationIntentFilterType.All
```

Alternatively (and I don't recommend this) change the name of fields to match convention, eg: `_applicationIntentFilter`

(comment applies to all following changes in this file)

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/AuthenticatingTdsServerArguments.cs
```diff
@@ -7,43 +7,31 @@ namespace Microsoft.SqlServer.TDS.Servers
...
 
         /// <summary>
         /// Filter for server name
         /// </summary>
-        public string ServerNameFilter { get; set; }
+        public string ServerNameFilter = string.Empty;
```
benrr101: I suppose the question is do we never want null here? If not, maybe this is a good case for `#enable nullable`. But I'd personally say null is a good default value here?

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/FederatedAuthenticationNegativeTdsScenarioType.cs
```diff
@@ -4,7 +4,7 @@
 
 namespace Microsoft.SqlServer.TDS.Servers
 {
-    public enum FederatedAuthenticationNegativeTDSScenarioType : int
+    public enum FederatedAuthenticationNegativeTdsScenarioType : int
```
benrr101, line 7: I might be wrong, but I thought enums by default inherit from int

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/FederatedAuthenticationNegativeTdsServer.cs
```diff
@@ -39,18 +39,18 @@ public override TDSMessageCollection OnPreLoginRequest(ITDSServerSession session
             TDSMessageCollection preLoginCollection = base.OnPreLoginRequest(session, request);
 
             // Check if arguments are of the Federated Authentication server
-            if (Arguments is FederatedAuthenticationNegativeTDSServerArguments)
+            if (Arguments is FederatedAuthenticationNegativeTdsServerArguments)
```
benrr101: Same `is Xyz xyz` should be used here

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/FederatedAuthenticationNegativeTdsServerArguments.cs
```diff
@@ -7,18 +7,11 @@ namespace Microsoft.SqlServer.TDS.Servers
...
         /// </summary>
-        public FederatedAuthenticationNegativeTDSScenarioType Scenario { get; set; }
-
-        /// <summary>
-        /// Initialization constructor
-        /// </summary>
-        public FederatedAuthenticationNegativeTDSServerArguments()
-        {
-        }
+        public FederatedAuthenticationNegativeTdsScenarioType Scenario = FederatedAuthenticationNegativeTdsScenarioType.ValidScenario;
```
benrr101: Should also be left as a property

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/FederatedAuthenticationNegativeTdsServer.cs
benrr101: *actually* now that I look at the GenericTdsServer, `Arguments` will always be the type T, which means this check and cast is totally unnecessary.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/GenericTdsServer.cs
```diff
@@ -69,36 +72,42 @@ public delegate void OnAuthenticationCompletedDelegate(
...
+
+        private TDSServerEndPoint _endpoint;
+
+        public IPEndPoint EndPoint => _endpoint.ServerEndPoint;
+
         /// <summary>
         /// Server configuration
         /// </summary>
-        protected TDSServerArguments Arguments { get; set; }
+        protected T Arguments { get; set; }
```
benrr101, line 98: Just adding it again, because this is now strongly typed to `T`, we don't need to check if `Arguments` is the right type in the child classes.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/GenericTdsServer.cs
```diff
@@ -69,36 +72,42 @@ public delegate void OnAuthenticationCompletedDelegate(
...
 
         /// <summary>
-        /// Default constructor
+        /// Counts pre-login requests to the server.
         /// </summary>
-        public GenericTDSServer() :
-            this(new TDSServerArguments())
-        {
-        }
+        public int PreLoginCount => _preLoginCount;
```
benrr101, line 108: This can be made an auto-property with private setter

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/GenericTdsServer.cs
```diff
@@ -69,36 +72,42 @@ public delegate void OnAuthenticationCompletedDelegate(
...
 
         /// <summary>
         /// Initialization constructor
         /// </summary>
-        public GenericTDSServer(TDSServerArguments arguments) :
+        public GenericTdsServer(T arguments) :
```
benrr101: ... technically in C# style constructors should go before properties, but I will overlook this

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/GenericTdsServer.cs
```diff
@@ -110,6 +119,14 @@ public GenericTDSServer(TDSServerArguments arguments, QueryEngine queryEngine)
             Engine.Log = Arguments.Log;
         }
 
+        public void Start([CallerMemberName] string methodName = "")
+        {
+            _endpoint = new TDSServerEndPoint(this) { ServerEndPoint = new IPEndPoint(IPAddress.Any, 0) };
+            _endpoint.EndpointName = methodName;
+            _endpoint.EventLog = Arguments.Log;
+            _endpoint.Start();
+        }
```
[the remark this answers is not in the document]
benrr101: Good point, I'd recommend throwing if endpoint already exists

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/GenericTdsServer.cs
```diff
@@ -244,7 +264,7 @@ public virtual TDSMessageCollection OnLogin7Request(ITDSServerSession session, T
                                 TDSLogin7SessionRecoveryOptionToken sessionStateOption = option as TDSLogin7SessionRecoveryOptionToken;
 
                                 // Inflate session state
-                                (session as GenericTDSServerSession).Inflate(sessionStateOption.Initial, sessionStateOption.Current);
+                                (session as GenericTdsServerSession).Inflate(sessionStateOption.Initial, sessionStateOption.Current);
```
benrr101, line 274: Removing these casts would be nice since `as` will return `null` if it's not the right type. Not saying it isn't the right type often, but I'm pretty sure I'll get warnings in my IDE about possible null reference exception

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/GenericTdsServer.cs
```diff
@@ -654,6 +674,16 @@ protected virtual TDSMessageCollection OnAuthenticationCompleted(ITDSServerSessi
                 }
             }
 
+            if (!String.IsNullOrEmpty(Arguments.FailoverPartner))
```
benrr101: Holdover from Java days - `string.IsNullOrEmpty` please :)

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/GenericTdsServer.cs
```diff
@@ -880,5 +910,7 @@ private bool AreEqual(byte[] left, byte[] right)
 
             return left.SequenceEqual<byte>(right);
         }
+
+        public virtual void Dispose() => _endpoint?.Dispose();
```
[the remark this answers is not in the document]
benrr101: Good suggestion again

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/RoutingTdsServerArguments.cs
```diff
@@ -7,43 +7,31 @@ namespace Microsoft.SqlServer.TDS.Servers
...
     /// </summary>
-    public class RoutingTDSServerArguments : TDSServerArguments
+    public class RoutingTdsServerArguments : TdsServerArguments
     {
         /// <summary>
-        /// Routing destination protocol
+        /// Routing destination protocol.
         /// </summary>
-        public int RoutingProtocol { get; set; }
+        public int RoutingProtocol = 0;
```
benrr101: Please keep these as properties

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/TdsServerArguments.cs
```diff
@@ -28,76 +28,56 @@ public class TDSServerArguments
         /// <summary>
         /// Log to which send TDS conversation
         /// </summary>
-        public TextWriter Log { get; set; }
+        public TextWriter Log = null;
```
benrr101: Please also leave these as properties

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/TransientDelayTdsServer.cs
```diff
@@ -0,0 +1,82 @@
...
+            Arguments.DelayDuration = sleepDuration;
+        }
+
+        /// <summary>
+        /// Handler for login request
+        /// </summary>
+        public override TDSMessageCollection OnLogin7Request(ITDSServerSession session, TDSMessage request)
+        {
+            // Check if we're still going to raise transient error
+            if (Arguments.IsEnabledPermanentDelay ||
```
benrr101: If `IsEnabledPermanentDelay` is true, will this method ever stop? Obviously that's the point, but how do we clean up after the server is done?

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/TransientDelayTdsServerArguments.cs
```diff
@@ -0,0 +1,31 @@
...
+using System;
+
+namespace Microsoft.SqlServer.TDS.Servers
+{
+    public class TransientDelayTdsServerArguments : TdsServerArguments
+    {
+        /// <summary>
+        /// The duration for which the server should sleep before responding to a request.
+        /// </summary>
+        public TimeSpan DelayDuration = TimeSpan.FromSeconds(0);
```
benrr101: Please make these properties

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/TransientDelayTdsServer.cs
```diff
@@ -0,0 +1,82 @@
...
+        {
+            Arguments.IsEnabledTransientDelay = isEnabledTransientTimeout;
+            Arguments.IsEnabledPermanentDelay = isEnabledPermanentTimeout;
+            Arguments.DelayDuration = sleepDuration;
+        }
+
+        /// <summary>
+        /// Handler for login request
+        /// </summary>
+        public override TDSMessageCollection OnLogin7Request(ITDSServerSession session, TDSMessage request)
```
benrr101, line 38: Since this is net-new code (right?) I'd request that methods be in proper order (ie, alphabetical)

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/TransientTdsErrorTdsServer.cs
```diff
@@ -0,0 +1,101 @@
...
+    /// <summary>
+    /// TDS Server that returns TDS error token on login request for the specified number of times
+    /// </summary>
+    public class TransientTdsErrorTdsServer : GenericTdsServer<TransientTdsErrorTdsServerArguments>, IDisposable
+    {
+        private int RequestCounter = 0;
+
+        public void SetErrorBehavior(bool isEnabledTransientFault, uint errorNumber, int repeatCount = 1, string message = null)
+        {
+            Arguments.IsEnabledTransientError = isEnabledTransientFault;
```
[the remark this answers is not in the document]
benrr101: Good point, and the other question is why we renamed the class from Fault to Error

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/TransientTdsErrorTdsServer.cs
```diff
@@ -0,0 +1,101 @@
...
+            Arguments.Number = errorNumber;
+            Arguments.Message = message;
+            Arguments.RepeatCount = repeatCount;
+        }
+
+        public TransientTdsErrorTdsServer(TransientTdsErrorTdsServerArguments arguments) : base(arguments)
```
benrr101, line 28: nit: `: base(arguments)` should probably go on the next line

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/TransientTdsErrorTdsServerArguments.cs
```diff
@@ -4,31 +4,26 @@
...
 namespace Microsoft.SqlServer.TDS.Servers
 {
-    public class TransientFaultTDSServerArguments : TDSServerArguments
+    public class TransientTdsErrorTdsServerArguments : TdsServerArguments
     {
         /// <summary>
         /// Transient error number to be raised by server.
         /// </summary>
-        public uint Number { get; set; }
+        public uint Number = 0;
```
benrr101: Please keep these as properties

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft/Data/SqlClient/SqlConnectionStringTest.cs
```diff
@@ -0,0 +1,68 @@
...
+        [InlineData("my.test.server", true, Tristate.True, true)]
+        [InlineData("my.test.server", false, Tristate.True, false)]
+        [InlineData("my.test.server", null, Tristate.True, false)]
+        [InlineData("my.test.server", true, Tristate.False, true)]
+        [InlineData("my.test.server", false, Tristate.False, false)]
+        [InlineData("my.test.server", null, Tristate.False, true)]
+        [InlineData("my.test.server", true, Tristate.NotInitialized, true)]
+        [InlineData("my.test.server", false, Tristate.NotInitialized, false)]
+        [InlineData("my.test.server", null, Tristate.NotInitialized, true)]
+        public void TestDefaultTNIR(string dataSource, bool? tnirEnabledInConnString, Tristate tnirDisabledAppContext, bool expectedValue)
```
benrr101: Since this is net new code, I'm going to ask that the name be `TestDefaultTnir`

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionFailoverTests.cs
```diff
@@ -0,0 +1,544 @@
...
+            };
+            using SqlConnection connection = new(builder.ConnectionString);
+            try
+            {
+                // Act
+                connection.Open();
+            }
+            catch (Exception e)
+            {
+                Assert.Fail(e.Message);
```
benrr101: Isn't this redundant? Like, the test will fail (with the full exception object) if an exception is thrown, right?

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionFailoverTests.cs
```diff
@@ -0,0 +1,544 @@
...
+            using TdsServer failoverServer = new(
+                new TdsServerArguments
+                {
+                    // Doesn't need to point to a real endpoint, just needs a value specified
+                    FailoverPartner = "localhost:1234",
+                });
+            failoverServer.Start();
+
+            using TransientTdsErrorTdsServer server = new(
+                new TransientTdsErrorTdsServerArguments()
```
benrr101, line 353: After looking at this pattern a good few times, I'm wondering if maybe we should consider making the arguments be properties on the server itself. 🤔

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionReadOnlyRoutingTests.cs
```diff
@@ -0,0 +1,158 @@
...
+                ApplicationIntent = ApplicationIntent.ReadOnly,
+                Encrypt = SqlConnectionEncryptOption.Optional
+             };
+            using SqlConnection connection = new(builder.ConnectionString);
+            await connection.OpenAsync();
+        }
+
+        [Fact]
+        public void RoutedConnection()
+            => RecursivelyRoutedConnection(1);
```
benrr101: Nit: I thought the `=>` were supposed to be on the previous line...

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionReadOnlyRoutingTests.cs
```diff
@@ -0,0 +1,158 @@
...
+
+namespace Microsoft.Data.SqlClient.ScenarioTests
+{
+    [Collection("SimulatedServerTests")]
+    public class ConnectionReadOnlyRoutingTests
+    {
+        [Fact]
+        public void NonRoutedConnection()
+        {
+            using TdsServer server = new();
```
benrr101, line 20: Why'd we give up on the Arrange / Act / Assert pattern for these tests?

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionRoutingTests.cs
```diff
@@ -0,0 +1,207 @@
...
+
+using System;
+using System.Data;
+using Microsoft.Data.Common;
+using Microsoft.SqlServer.TDS.Servers;
+using Xunit;
+
+namespace Microsoft.Data.SqlClient.ScenarioTests
+{
+    [Trait("Category", "flaky")]
```
benrr101, line 13: This is scary ☹️

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionRoutingTests.cs
```diff
@@ -0,0 +1,207 @@
...
+            };
+            using SqlConnection connection = new(builder.ConnectionString);
+            try
+            {
+                // Act
+                connection.Open();
+            }
+            catch (Exception e)
+            {
+                Assert.Fail(e.Message);
```
benrr101: Another case where it seems unnecessary to catch these as the test will always fail on exception

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionRoutingTests.cs
```diff
@@ -0,0 +1,207 @@
...
+            Assert.Equal(2, server.PreLoginCount);
+        }
+
+        [Theory]
+        [InlineData(40613)]
+        [InlineData(42108)]
+        [InlineData(42109)]
+        public void TransientFaultAtRoutedLocation_RetryDisabled_ShouldFail(uint errorCode)
+        {
+            // Arrange
```
benrr101, line 71: Where's the Act and Assert?

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionRoutingTestsAzure.cs
```diff
@@ -0,0 +1,203 @@
...
+                    IsEnabledTransientError = true,
+                    Number = errorCode,
+                });
+
+            server.Start();
+
+            using RoutingTdsServer router = new RoutingTdsServer(
+                new RoutingTdsServerArguments()
+                {
+                    //RoutingTCPHost = server.EndPoint.Address.ToString() == IPAddress.Any.ToString() ? IPAddress.Loopback.ToString() : server.EndPoint.Address.ToString(),
```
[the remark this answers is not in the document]
benrr101: Agree with this one, too

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionRoutingTests.cs
```diff
@@ -0,0 +1,207 @@
...
+using System;
+using System.Data;
+using Microsoft.Data.Common;
+using Microsoft.SqlServer.TDS.Servers;
+using Xunit;
+
+namespace Microsoft.Data.SqlClient.ScenarioTests
+{
+    [Trait("Category", "flaky")]
+    [Collection("SimulatedServerTests")]
```
benrr101, line 14: I'm not sure about this ... Generally speaking the namespace should be used for categorizing tests and the namespace should match the folder structure. Neither of these are met
I guess I'd consider how you want to categorize this, and make the folder and namespace match.

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionTests.cs
```diff
@@ -20,26 +19,36 @@
...
 {
-    public class SqlConnectionBasicTests
+    public class ConnectionTests
     {
         [Fact]
         public void ConnectionTest()
         {
-            using TestTdsServer server = TestTdsServer.StartTestServer();
-            using SqlConnection connection = new SqlConnection(server.ConnectionString);
+            using TdsServer server = new(new TdsServerArguments() { });
```
benrr101, line 29: No Arrange/Act/Assert for these either? :(

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionTests.cs
```diff
@@ -49,117 +58,264 @@ public void IntegratedAuthConnectionTest()
...
-                Assert.Fail(e.Message);
-            }
+            connection.Open();
+            Assert.Equal(ConnectionState.Open, connection.State);
+            Assert.Equal($"localhost,{server.EndPoint.Port}", connection.DataSource);
+            Assert.Equal(2, server.PreLoginCount);
         }
 
-        [ConditionalTheory(typeof(TestUtility), nameof(TestUtility.IsNotArmProcess))]
+        [Trait("Category", "flaky")]
```
benrr101, line 128: ☹️

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionTests.cs
```diff
@@ -341,29 +501,34 @@ public void ConnectionTimeoutTest(int timeout)
...
 
             // Dispose the server to force connection timeout 
             server.Dispose();
 
             // Measure the actual time it took to timeout and compare it with configured timeout
             Stopwatch timer = new();
-            Exception ex = null;
+            Exception? ex = null;
 
             // Open a connection with the server disposed.
```
benrr101, line 527: Outside the scope of this PR, but this can be rewritten much better

```
// Act
using CancellationTokenSource cts = new(timeout * 1000); // Please replace with TimeSpan.FromXyz so we know the units
   
timer.Start();
Assert.ThrowsAsync<SqlException>(() => );
timer.Stop();

// Assert
// Ok well technically we asserted above but ... I can't rewrite it more pretty
Assert.True(timer.Elapsed.TotalSeconds <= timeout + 3)
```

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/TransientTdsErrorTdsServer.cs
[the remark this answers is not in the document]
mdaigle: It's named error because it returns a TDSError token.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.EndPoint/TDSServerEndPointConnection.cs
```diff
@@ -44,12 +45,12 @@ public override void ProcessData(Stream rawStream)
...
     /// Connection to a single client
     /// </summary>
-    public abstract class ServerEndPointConnection
+    public abstract class ServerEndPointConnection : IDisposable
     {
         /// <summary>
         /// Worker thread
         /// </summary>
-        protected Thread ProcessorThread { get; set; }
+        protected Task ProcessorTask { get; set; }
```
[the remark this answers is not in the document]
mdaigle: I couldn't get an await on the ProcessorTask to complete. There's some deadlock condition that gets triggered when we try to await ProcessorTask. More work to be done here.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/ConnectionString/DbConnectionStringDefaults.cs
mdaigle: I want to start to unwind how much global static config we have because it really hampers our ability to test. In this case, it's a static readonly field wrapping another static field, so even if you update the context switch, the default is still locked in. Once the context switch is initialized this call is just a bool comparison under the hood, so shouldn't have any performance impact. From a user perspective there's also no change to the experience because we don't provide a public method to update a LocalAppContextSwitch value.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/AdapterUtil.cs
mdaigle: What I really want is an internal only path to make this injectable per connection. We have tons of paths that branch on "isAzure" that we can't hit today. The change to a list is sort of a half-measure to allow updates without doing that whole refactor. We iterate through this set of endpoints when checking isAzure, so array vs. list shouldn't impact performance.

I'm switching the method that iterates over the list to accept an ICollection<string>.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/AdapterUtil.cs
mdaigle: yes, this method is called in two places, IsAzureSqlServerEndpoint and IsAzureSqlServerEndpoint, which use different lists.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/LocalizationTest.cs
mdaigle: The challenge is that the TDS.Servers project doesn't have a dependency on the MDS project and all of the connection string building functionality lives in MDS. I think this is correct because otherwise they could become too tightly coupled. A change in MDS could automatically make a compensating change here, resulting in a behavior change but not breaking related tests.

Also, one of the key features of these tests is that the connection string parameters that you use really matter. I think that hiding the connection string creation away under layers of abstraction will make it too easy to accidentally introduce new default values that lots of tests pick up. For example, before these changes, the servers had defaults for encryption (not supported) that don't match the driver's default setting (mandatory).

src/Microsoft.Data.SqlClient/tests/ManualTests/TracingTests/DiagnosticTest.cs
mdaigle: I think you're right, but I tried not to rewrite existing tests except as necessary to get them to compile.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.EndPoint/TDSServerEndPointConnection.cs
mdaigle: I tried to introduce some async behavior in here but hit my time box 🤷‍♂️ There's more cleanup to be done in the interaction between the TDSServer and TDSServerEndpointConnection. Some of the lifecycles and task management are very odd.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/AuthenticatingTdsServer.cs
mdaigle: Removed the cast as the arguments are already strongly typed.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/AuthenticatingTdsServerArguments.cs
mdaigle: Probably! I tried not to change the internals of the existing server implementations too much to try to keep the line count down. There are probably many more similar improvements that can be made.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/FederatedAuthenticationNegativeTdsScenarioType.cs
mdaigle: Similar to above. I tried to focus on project structure and consolidation in this PR, so there are many more cleanup items we could address.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/FederatedAuthenticationNegativeTdsServer.cs
mdaigle: Yep, I did make that change in the new server types, but not in the existing ones.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/GenericTdsServer.cs
mdaigle: Unfortunately, it doesn't play nicely with Interlocked.Increment, so I need a backing field.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/GenericTdsServer.cs
mdaigle: I'm thinking of just removing these interfaces because there's only one implementation and they're cluttering the code. Would like to take this change in a separate PR to keep this one targeted.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/TransientDelayTdsServer.cs
mdaigle: Really this should be done with async/await and a cancellation token that we can trigger when we want to shut down the server to break us out of any ongoing handlers. I'd like to take this change in a future PR as it'll require a refactor of all the handlers.

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionFailoverTests.cs
mdaigle: I'm not opposed, but we'd have some redundancy because some of the inherited properties from TdsServerArguments are used in all of the server implementations or in the base class.

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionReadOnlyRoutingTests.cs
mdaigle: This is why I hate ambiguous whitespace/newline rules haha. Can you pick a configurable option that we can set in the .editorconfig? I'd like to set these once and have the compiler tell me straight away.

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionReadOnlyRoutingTests.cs
mdaigle: These are all existing tests, moved over to a new location

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionTests.cs
mdaigle: These were existing tests

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionRoutingTests.cs
mdaigle: Meant to switch these all to SimulatedServerTests. Will update this.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/AdapterUtil.cs
benrr101: That's what I figured you were going for, and I agree that's probably a decent approach for making it injectable at runtime.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/ConnectionString/DbConnectionStringDefaults.cs
benrr101: Ok, that's a valid explanation 👍 
I suppose if we really want to split hairs, by having `!xyz` we're still doing a calculation each time the property is called, but ... it's small.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectionTestWithSSLCert/CertificateTestWithTdsServer.cs
benrr101: (I meant for this entire switch - introduce a helper to get the X509 cert)

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionFailoverTests.cs
benrr101: We could put those properties in the base server class?
Idk, it's ok as-is. Just seemed a bit annoying to have to new up both classes every time

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionRoutingTests.cs
benrr101: So, we can remove [Collection] now? They should be grouped now by namespace appropriately, right?

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionRoutingTests.cs
benrr101: 👍 but not fixed 😄 - just moved without changes?

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/GenericTdsServer.cs
benrr101: Ah. Fair enough.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/GenericTdsServer.cs
benrr101: Ok that's fine

src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/AdapterUtil.cs
```diff
@@ -816,19 +816,19 @@ internal static Version GetAssemblyVersion()
 
         internal static readonly string[] s_azureSynapseOnDemandEndpoints = [.. s_azureSqlServerOnDemandEndpoints, .. s_azureSynapseEndpoints];
 
-        internal static bool IsAzureSynapseOnDemandEndpoint(string dataSource)
+        internal static bool C(string dataSource)
```
benrr101: ## C
uhh, is that intentional? 😂

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionRoutingTests.cs
mdaigle: Thought I got all these, but I must have missed this one.

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionRoutingTests.cs
mdaigle: The Collection trait also serves to serialize test execution. The app context switch and azure endpoint updates in some of these tests don't play nicely in a parallel environment. https://xunit.net/docs/running-tests-in-parallel#custom-test-collections. This is another area for improvement.

benrr101: Ugh, that's not particularly good naming for the attribute, but that's not on you

paulmedynski (approved):

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionReadOnlyRoutingTests.cs
benrr101: As per offline discussion - there isn't a rule for this in .editorconfig, but I'm interested in experimenting with writing a roslyn analyzer for it

PR #3489, opened 2025-07-17
Add Json Payload Functionality for User Agent Feature Extension

samsharma2700:
## Description

Part 2 of UserAgent work. Previous PR: https://github.com/dotnet/SqlClient/pull/3451

- Adding necessary classes for UserAgentInfo json payload.
- Implements the current spec guidelines for this FE(https://microsoft.sharepoint-df.com/:w:/t/sqldevx/ERIWTt0zlCxLroNHyaPlKYwBI_LNSff6iy_wXZ8xX6nctQ?e=iP8q75)
- The payload is defined by a DTO(Data Transfer Object) class and a UserAgentInfo that implements spec constraints(character limits, bytes limits and field dropping logic etc.)
- Part 3 will make the changes to write the payload as a part of login flow

## Testing
Builds are running and added unit tests to verify changes.


roji:
@samsharma2700 @cheenamalhotra great to see this work happening. I've taken a look at the spec doc linked above, and could you clarify how we'd go about injecting EF information into the user agent? We'd need to identify EF Core (as opposed to direct users of SqlClient), it's version, etc.

paulmedynski:
Hi @roji - There is a V2 of the design that adds a public API for middleware like EF to pass version information, but we need to go through a security and privacy review before we can implement it.  This V1 phase only includes values we can pull from the runtime that don't have any such concerns.

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft.Data.SqlClient.UnitTests.csproj
```diff
@@ -48,4 +48,7 @@
       <TargetPath>xunit.runner.json</TargetPath>
     </ContentWithTargetPath>
   </ItemGroup>
+  <ItemGroup>
+    <Folder Include="microsoft.data.sqlclient\tests\unittests\microsoft\data\sqlclient\" />
+  </ItemGroup>
```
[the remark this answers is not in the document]
paulmedynski: The UnitTests project doesn't explicitly specify any of the source files to compile.  We let the SDK figure it out.  This ItemGroup shouldn't be necessary.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/UserAgentInfo.cs
paulmedynski, line 1: These new files should live in the src/ project, and shouldn't be duplicated in netcore and netfx.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/UserAgentInfo.cs
```diff
@@ -0,0 +1,421 @@
...
+using System.Management;
+#endif
+
+namespace Microsoft.Data.SqlClient
+{
+    /// <summary>
+    /// Gathers driver + environment info, enforces size constraints,
+    /// and serializes into a UTF-8 JSON payload.
+    /// </summary>
+    public static class UserAgentInfo
```
paulmedynski: This isn't part of our public API, so should be internal.

The UnitTests project has access to internals, so shouldn't have any effect on testing.

paulmedynski (changes requested):
Need to move the UserAgent files into the src/ project and avoid duplicating them.  I'll complete my review after the move.

roji:
@paulmedynski ok - can you clarify what's planned for the 6.1.0 (v1, v2?)

paulmedynski:
The V1 work will arrive in 7.0.0.  The V2 work to support EF and other middleware may also make it into that timeframe, but depends on which reviews we need and how long they take.  There is also ongoing SQL Server support that we need to align with to make any of this driver side work useful.

@samsharma2700 will be completing the V1 work over several PRs.  We have (and will) put them into the 7.0.0-preview1 milestone.

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft.Data.SqlClient.UnitTests.csproj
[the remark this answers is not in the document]
samsharma2700: Sounds good.Will update.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/UserAgentInfo.cs
samsharma2700: Good point! moved to src/

roji:
@paulmedynski thanks for the clarifications!

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
```diff
@@ -0,0 +1,421 @@
...
+using System.Text;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using System.Text.RegularExpressions;
+using Microsoft.Data.Common;
+
+#if WINDOWS
+using System.Management;
+#endif
+
```
paulmedynski, line 12: `#nullable enable`

Add this to all new files, and we can remove it later when we enable nullable everywhere.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
```diff
@@ -0,0 +1,421 @@
...
+namespace Microsoft.Data.SqlClient.UserAgent
```
paulmedynski: Prefer namespace declaration:

`namespace Microsoft.Data.SqlClient.UserAgent;`

Then the entire definition isn't unnecessarily indented.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
```diff
@@ -0,0 +1,421 @@
...
+        /// <summary>
+        /// Maximum number of characters allowed for the driver runtime.
+        /// </summary>
+        private const int RuntimeMaxChars = 128;
+
+        /// <summary>
+        /// Maximum number of bytes allowed for the user agent json payload.
+        /// payloads larger than this may be rejected by the server.
+        /// </summary>
+        public const int JsonPayloadMaxBytesSpec = 2047;
```
paulmedynski: What is the significance of the `Spec` suffix here?  I think `JsonPayloadMaxBytes` is accurate enough.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
```diff
@@ -0,0 +1,421 @@
...
+
+#if WINDOWS
+using System.Management;
+#endif
+
+namespace Microsoft.Data.SqlClient.UserAgent
+{
+    /// <summary>
+    /// Gathers driver + environment info, enforces size constraints,
+    /// and serializes into a UTF-8 JSON payload.
```
paulmedynski: It's probably worthwhile adding a link to the spec document here.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
```diff
@@ -0,0 +1,421 @@
...
+        private const int ArchMaxChars = 16;
+
+        /// <summary>
+        /// Maximum number of characters allowed for the driver runtime.
+        /// </summary>
+        private const int RuntimeMaxChars = 128;
+
+        /// <summary>
+        /// Maximum number of bytes allowed for the user agent json payload.
+        /// payloads larger than this may be rejected by the server.
```
paulmedynski: payloads -> Payloads

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
```diff
@@ -0,0 +1,421 @@
...
+        /// </summary>
+        public const int JsonPayloadMaxBytesSpec = 2047;
+
+        /// <summary>
+        /// Maximum number of bytes allowed before we drop multiple fields 
+        /// and only send bare minimum useragent info.
+        /// </summary>
+        public const int UserAgentPayloadMaxBytes = 10000;
```
paulmedynski: I'm curious why this exists...

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
```diff
@@ -0,0 +1,421 @@
...
+
+
+        private const string DefaultJsonValue = "Unknown";
+        private const string DefaultDriverName = "MS-MDS";
```
paulmedynski: This _is_ the driver name, not the default.  I think `DriverName` would be more accurate.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
```diff
@@ -0,0 +1,421 @@
...
+
+        // JSON Payload for UserAgent
+        private static readonly string driverName;
```
paulmedynski: Can we avoid these entirely and just assign directly into the DTO?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
```diff
@@ -0,0 +1,421 @@
...
+                if (OperatingSystem.IsMacOS())   return OsType.macOS;
+#endif
+                // second we fallback to OSplatform checks
+                if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+                    return OsType.Windows;
+                if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+                    return OsType.Linux;
+                if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+                    return OsType.macOS;
+
```
paulmedynski: `OSPlatform.FreeBSD` exists for .NET, so you can check that conditionally here.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
```diff
@@ -0,0 +1,421 @@
...
+                // final fallback is inspecting OSdecription
+                var desc = RuntimeInformation.OSDescription?.ToLowerInvariant() ?? "";
+                if (desc.Contains("android"))
```
paulmedynski: How did you determine these are the correct strings to search for?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
```diff
@@ -0,0 +1,421 @@
...
+            return OsType.Unknown;
+        }
+
+        /// <summary>
+        /// Given an OsType enum, returns the edition/distro string.
+        /// passing the enum makes search less expensive
+        /// </summary>
+        private static string DetectOsDetails(OsType os)
+        {
+            try
```
paulmedynski: I think we just want to use OSDescription here regardless of OS.  The runtime will faithfully populate it with something useful for us.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
```diff
@@ -0,0 +1,421 @@
...
+        /// <summary>
+        /// Detects and reports whatever CPU architecture the guest OS exposes
+        /// </summary>
+        private static string DetectArchitecture()
+        {
+            try
+            {
+                // Returns “X86”, “X64”, “Arm”, “Arm64”, etc.
+                // This is the architecture of the guest process it's running in
+                // it does not see through to the physical host.
```
paulmedynski: Run-on sentence.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
```diff
@@ -0,0 +1,421 @@
...
+        /// <summary>
+        /// Reads the Microsoft.Data.SqlClient assembly’s informational version
+        /// or falls back to its AssemblyName.Version.
+        /// </summary>
+        private static string DetectRuntime()
+        {
+            // 1) Try the built-in .NET runtime description
+            try
+            {
+                string fw = RuntimeInformation.FrameworkDescription;
```
paulmedynski: I think it's fine to use FrameworkDescription here, and give up if it throws or is empty.  No need to interrogate further.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
```diff
@@ -0,0 +1,421 @@
...
+                if (!string.IsNullOrWhiteSpace(fw))
+                    return fw.Trim();
+            }
+            catch
+            {
+                // ignore and fall back
+            }
+
+            // 2) On Linux, ask glibc what version it is
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
```
paulmedynski: We can count on .NET on Linux will always having a FrameworkDescription.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
```diff
@@ -0,0 +1,421 @@
...
+                    if (!string.IsNullOrWhiteSpace(glibc))
+                        return "glibc " + glibc.Trim();
+                }
+                catch
+                {
+                    // ignore
+                }
+            }
+
+            // 3) If running under Mono, grab its internal display name
```
paulmedynski: We don't support Mono, so this is unnecessary.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
```diff
@@ -0,0 +1,421 @@
...
+                // If the payload is over 10KB, we only send the bare minimum fields
+                dto.OS.Details = null; // drop OS.Details
+                dto.Runtime = null; // drop Runtime
+                dto.Arch = null; // drop Arch
+                payload = JsonSerializer.SerializeToUtf8Bytes(dto, options);
+            }
+
+            // Last check to ensure we are within the limits(in case remaining fields are still too large)
+            return payload.Length > UserAgentPayloadMaxBytes
+                ? JsonSerializer.SerializeToUtf8Bytes(new { }, options)
```
paulmedynski: We must always send at least the driver name.  Should driver version be mandatory too?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
paulmedynski: Does this need to be public?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
```diff
@@ -0,0 +1,421 @@
...
+        private const string DefaultDriverName = "MS-MDS";
+
+        // JSON Payload for UserAgent
+        private static readonly string driverName;
+        private static readonly string version;
+        private static readonly string osType;
+        private static readonly string osDetails;
+        private static readonly string architecture;
+        private static readonly string runtime;
+        private static readonly byte[] _cachedPayload;
```
paulmedynski: This can be a public property to make life easier in the tests.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
```diff
@@ -0,0 +1,421 @@
...
+            };
+            byte[] payload = JsonSerializer.SerializeToUtf8Bytes(dto, options);
+
+            // Note: server will likely reject payloads larger than 2047 bytes
+            // Try if the payload fits the max allowed bytes
+            if (payload.Length <= JsonPayloadMaxBytesSpec)
+            { 
+                return payload;
+            }
+            if (payload.Length > UserAgentPayloadMaxBytes)
```
paulmedynski: Since we're pretty sure the worst-case size is < 2047 bytes, let's skip this and just null-out the extra details and try once more.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfoDto.cs
```diff
@@ -0,0 +1,44 @@
...
+        // the DTO or relying on JsonProperty attribute resolution at runtime. The small overhead of 
+        // maintaining constants is justified by the performance and allocation savings.
+        public const string DriverJsonKey = "driver";
+        public const string VersionJsonKey = "version";
+        public const string OsJsonKey = "os";
+        public const string ArchJsonKey = "arch";
+        public const string RuntimeJsonKey = "runtime";
+
+        [JsonPropertyName(DriverJsonKey)]
+        public string Driver { get; set; }
```
paulmedynski: All of these properties should be `init` instead of `set`.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfoDto.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System.Text.Json;
+using System.Text.Json.Serialization;
+namespace Microsoft.Data.SqlClient.UserAgent
+{
+    internal class UserAgentInfoDto
```
paulmedynski: Should this be a readonly struct?

src/Microsoft.Data.SqlClient/tests/UnitTests/UserAgentInfoTests.cs
```diff
@@ -0,0 +1,130 @@
+﻿using System;
+using System.Reflection;
```
paulmedynski: We don't need any reflection in the UnitTests project.  It has access to MDS internals.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
```diff
@@ -0,0 +1,421 @@
...
+
+        }
+
+        /// <summary>
+        /// Truncates a string to the specified maximum length or returns a default value if input is null or empty.
+        /// </summary>
+        /// <param name="jsonStringVal">The string value to truncate</param>
+        /// <param name="maxChars">Maximum number of characters allowed</param>
+        /// <returns>Truncated string or default value if input is invalid</returns>
+        private static string TruncateOrDefault(string jsonStringVal, int maxChars)
```
paulmedynski: private -> internal   will make it easier to test.

src/Microsoft.Data.SqlClient/tests/UnitTests/UserAgentInfoTests.cs
```diff
@@ -0,0 +1,130 @@
...
+            Assert.NotNull(payload);
+            Assert.InRange(payload.Length, 1, UserAgentInfo.JsonPayloadMaxBytesSpec);
+        }
+
+        // 2. TruncateOrDefault respects null, empty, fit, and overflow cases
+        [Theory]
+        [InlineData(null, 5, "Unknown")]              // null returns default
+        [InlineData("", 5, "Unknown")]                // empty returns default
+        [InlineData("abc", 5, "abc")]                // within limit unchanged
+        [InlineData("abcdef", 5, "abcde")]          // overflow truncated
```
paulmedynski: Test exactly max chars in the input.

src/Microsoft.Data.SqlClient/tests/UnitTests/UserAgentInfoTests.cs
```diff
@@ -0,0 +1,130 @@
...
+                name: "AdjustJsonPayloadSize",
+                bindingAttr: BindingFlags.NonPublic | BindingFlags.Static);
+            Assert.NotNull(mi);
+
+            byte[] payload = (byte[])mi!.Invoke(null, new object?[] { dto })!;
+
+            // Final payload must satisfy limits
+            Assert.InRange(payload.Length, 1, UserAgentInfo.UserAgentPayloadMaxBytes);
+
+            // Convert to string for field presence checks
```
paulmedynski: Probably better to parse as a JsonDocument and then explictily test the fields.

src/Microsoft.Data.SqlClient/tests/UnitTests/UserAgentInfoTests.cs
paulmedynski, line 1: It would be nice to also see tests that prove your trunctate, adjust, and serialize all works together.  You could put all of that into an internal function that takes in the field values, and then you can check the output is as expected.

src/Microsoft.Data.SqlClient/tests/UnitTests/UserAgentInfoTests.cs
```diff
@@ -0,0 +1,130 @@
...
+            Assert.Contains(UserAgentInfoDto.VersionJsonKey, json);
+            Assert.Contains(UserAgentInfoDto.OsJsonKey, json);
+
+            // Low‑priority fields removed
+            Assert.DoesNotContain(UserAgentInfoDto.ArchJsonKey, json);
+            Assert.DoesNotContain(UserAgentInfoDto.RuntimeJsonKey, json);
+            Assert.DoesNotContain(UserAgentInfoDto.OsInfo.DetailsJsonKey, json);
+        }
+
+        // 4. DTO serializes with expected JSON property names
```
paulmedynski: What about testing that the values are serialized as expected as well?

paulmedynski (changes requested):
You're on the right track here.  We can discuss my comments and suggestions.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
samsharma2700: Yes agreed, the file-scoped namespace declarations do make the files cleaner. I assumed our project convention is to use traditional namespace declaration with a block, so continued with that.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
samsharma2700: So `Spec` marks the specification defined limit for the payload, but we still send the payload if it is higher than that. We drop a few fields if payload is larger than `UserAgentPayloadMaxBytes`. Added a suffix to distinguish them easily. Made it public for easier testing but I'll hard code this value for now.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
samsharma2700: In our discussion we decided to let payloads bigger than the spec limit to go through unless they reach this value(`UserAgentPayloadMaxBytes`) that is closer to packet limits. Over this value we drop certain fields of the payload and send a smaller user agent payload.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
samsharma2700: Will update.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfoDto.cs
samsharma2700: Don't think so, it is mutable by design, we modify properties post creation. There is also hierarchical structure. Serialization behavior is safer and more predictable with classes.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
samsharma2700: It's not available for all versions. Ran into build issues.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
samsharma2700: Mostly trial and error, inspecting the string .NET returns on different platforms. There is no formal specification so using this as the final fallback.

src/Microsoft.Data.SqlClient/tests/UnitTests/UserAgentInfoTests.cs
samsharma2700: Good point, will check both in one go.

src/Microsoft.Data.SqlClient/tests/UnitTests/UserAgentInfoTests.cs
samsharma2700: Yes good point, added more tests.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
```diff
@@ -1,421 +1,318 @@
...
+        Unknown
+    }
+
+    static UserAgentInfo()
+    {
+        _dto = BuildDto();
+        _cachedPayload = AdjustJsonPayloadSize(_dto);
+    }
+
+    static UserAgentInfoDto BuildDto()
```
paulmedynski: This should have internal visibility.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
```diff
@@ -1,421 +1,318 @@
...
+        // TDS Prelogin7 packets support up to 65,535 bytes (including headers), but many server versions impose
+        // stricter limits for prelogin payloads.
+        //
+        // As a safety measure:
+        // - If the serialized payload exceeds **10 KB**, we fallback to transmitting only essential fields:
+        //   'driver', 'version', and 'os.type'
+        // - If the payload exceeds 2,047 bytes but remains within sensible limits, we still send it, but note that
+        //   some servers may silently drop or reject such packets — behavior we may use for future probing or diagnostics.
+        // - If payload exceeds 10KB even after dropping fields , we send an empty payload.
+        var driverName = TruncateOrDefault(DriverName, DriverNameMaxChars);
```
paulmedynski: We can just assign directly to the DTO properties/fields instead of making temporaries here:
```
return new()
{
  Driver = TruncateOrDefault(DriverName, DriverNameMaxChars).
  ...
}
```

You don't even need to repeat `UserAgentInfoDto` since the compiler knows what the return type is.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfoDto.cs
```diff
@@ -1,44 +1,47 @@
...
+    public const string DriverJsonKey = "driver";
+    public const string VersionJsonKey = "version";
+    public const string OsJsonKey = "os";
+    public const string ArchJsonKey = "arch";
+    public const string RuntimeJsonKey = "runtime";
+
+
+    // TODO: Does this need to be nullable?
+    [JsonPropertyName(DriverJsonKey)]
+    public string? Driver { get; set; }
```
paulmedynski: These two should be non-nullable.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfoDto.cs
```diff
@@ -1,44 +1,47 @@
...
-            public string Type { get; set; }
-
-            [JsonPropertyName(DetailsJsonKey)]
-            public string Details { get; set; }
-        }
+        public const string TypeJsonKey = "type";
+        public const string DetailsJsonKey = "details";
+
+        [JsonPropertyName(TypeJsonKey)]
+        public string? Type { get; set; }
```
paulmedynski: This doesn't need to be nullable.

src/Microsoft.Data.SqlClient/tests/UnitTests/UserAgentInfoTests.cs
```diff
@@ -69,39 +60,36 @@ public void AdjustJsonPayloadSize_StripsLowPriorityFields_When_PayloadTooLarge()
...
-            // Final payload must satisfy limits
-            Assert.InRange(payload.Length, 1, UserAgentInfo.UserAgentPayloadMaxBytes);
+            // Structural checks using JsonDocument
+            using JsonDocument doc = JsonDocument.Parse(payload);
+            JsonElement root = doc.RootElement;
 
-            // Convert to string for field presence checks
-            string json = Encoding.UTF8.GetString(payload);
+            // High‑priority fields must survive.
+            Assert.True(root.TryGetProperty(UserAgentInfoDto.DriverJsonKey, out _));
```
paulmedynski: You can at least verify the driver name is "MS-MDS".  You could also probably verify the version if you can access the ADP helper that returns it.

src/Microsoft.Data.SqlClient/tests/UnitTests/UserAgentInfoTests.cs
```diff
@@ -69,39 +60,36 @@ public void AdjustJsonPayloadSize_StripsLowPriorityFields_When_PayloadTooLarge()
...
 
-            // We either receive the minimal payload with only high‑priority fields,
-            // or we receive an empty payload in case of overflow despite dropping fields.
-            if (payload.Length <= 2)
+            // If the "os" object survived, only its "type" sub‑field may remain.
+            if (root.TryGetProperty(UserAgentInfoDto.OsJsonKey, out JsonElement os))
             {
-                Assert.Equal("{}", json.Trim());
-                return;
+                Assert.True(os.TryGetProperty(UserAgentInfoDto.OsInfo.TypeJsonKey, out _));
```
paulmedynski: You should verify the value here, since we have a constrained set of allowable values and you can predict what they will be.

src/Microsoft.Data.SqlClient/tests/UnitTests/UserAgentInfoTests.cs
```diff
@@ -69,39 +60,36 @@ public void AdjustJsonPayloadSize_StripsLowPriorityFields_When_PayloadTooLarge()
...
-            Assert.Contains(UserAgentInfoDto.OsJsonKey, json);
-
-            // Low‑priority fields removed
-            Assert.DoesNotContain(UserAgentInfoDto.ArchJsonKey, json);
-            Assert.DoesNotContain(UserAgentInfoDto.RuntimeJsonKey, json);
-            Assert.DoesNotContain(UserAgentInfoDto.OsInfo.DetailsJsonKey, json);
         }
 
-        // 4. DTO serializes with expected JSON property names
+        // 4. DTO JSON contract - verify names and values
```
paulmedynski: Test serialization when the nullable fields are null.

paulmedynski (changes requested):
Very close now; just a few more comments.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
```diff
@@ -0,0 +1,318 @@
+﻿using System;
+using System.Runtime.InteropServices;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using Microsoft.Data.Common;
+
+#nullable enable
+
+#if WINDOWS
```
benrr101: Is this actually supported in netfx?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
```diff
@@ -0,0 +1,318 @@
...
+{
+    /// <summary>
+    /// Maximum number of characters allowed for the driver name.
+    /// </summary>
+    private const int DriverNameMaxChars = 16;
+
+    /// <summary>
+    /// Maximum number of characters allowed for the driver version.
+    /// </summary>
+    private const int VersionMaxChars = 16;
```
benrr101: Please order alphabetically

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
```diff
@@ -0,0 +1,318 @@
...
+            Driver = driverName,
+            Version = version,
+            OS = new UserAgentInfoDto.OsInfo
+            { 
+                Type = osType,
+                Details = osDetails
+            },
+            Arch = architecture,
+            Runtime = runtime
+
```
benrr101: nit: extra line

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
```diff
@@ -0,0 +1,318 @@
...
+    /// <summary>
+    /// Detects the OS platform and returns the matching OsType enum.
+    /// </summary>
+    private static OsType DetectOsType()
+    {
+        try
+        {
+            // first we try with built-in checks (Android and FreeBSD also report Linux so they are checked first)
+#if NET6_0_OR_GREATER
+            if (OperatingSystem.IsAndroid())  return OsType.Android;
```
benrr101: Ok, sorry to do this but, please *always always always* use braces around if/else blocks

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
```diff
@@ -0,0 +1,318 @@
...
+            if (OperatingSystem.IsFreeBSD())  return OsType.FreeBSD;
+            if (OperatingSystem.IsWindows()) return OsType.Windows;
+            if (OperatingSystem.IsLinux())   return OsType.Linux;
+            if (OperatingSystem.IsMacOS())   return OsType.macOS;
+#endif
+            // second we fallback to OSPlatform checks
+#if NETCOREAPP3_0_OR_GREATER || NET5_0_OR_GREATER
```
benrr101: We don't support < net6, so these conditions seem unnecessary

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
```diff
@@ -0,0 +1,318 @@
...
+            return osDetails;
+        }
+        
+        return DefaultJsonValue;
+    }
+
+    /// <summary>
+    /// Detects and reports whatever CPU architecture the guest OS exposes
+    /// </summary>
+    private static string DetectArchitecture()
```
benrr101: Also please organize these alphabetically

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
```diff
@@ -0,0 +1,318 @@
...
+    {
+        try
+        {
+            // Returns the architecture of the current process (e.g., "X86", "X64", "Arm", "Arm64").
+            // Note: This reflects the architecture of the running process, not the physical host system.
+            return RuntimeInformation.ProcessArchitecture.ToString();
+        }
+        catch
+        {
+            // In case RuntimeInformation isn’t available or something unexpected happens
```
benrr101: Since there's only one line in the try block, can we just move the default return value to inside the catch - empty catch blocks are a code smell to me

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
```diff
@@ -0,0 +1,318 @@
...
+    }
+
+    /// <summary>
+    /// Returns the framework description as a string.
+    /// </summary>
+    private static string DetectRuntime()
+    {
+        // FrameworkDescription is never null, but IsNullOrWhiteSpace covers it anyway
+        var desc = RuntimeInformation.FrameworkDescription;
+        if (string.IsNullOrWhiteSpace(desc))
```
benrr101: BRACES!!!!

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfoDto.cs
```diff
@@ -0,0 +1,47 @@
...
+internal class UserAgentInfoDto
+{
+    // Note: JSON key names are defined as constants to avoid reflection during serialization.
+    // This allows us to calculate their UTF-8 encoded byte sizes efficiently without instantiating
+    // the DTO or relying on JsonProperty attribute resolution at runtime. The small overhead of 
+    // maintaining constants is justified by the performance and allocation savings.
+    public const string DriverJsonKey = "driver";
+    public const string VersionJsonKey = "version";
+    public const string OsJsonKey = "os";
+    public const string ArchJsonKey = "arch";
```
benrr101, line 24: Sort alphabetically please

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfoDto.cs
```diff
@@ -0,0 +1,47 @@
...
+    public const string RuntimeJsonKey = "runtime";
+
+
+    // TODO: Does this need to be nullable?
```
benrr101: According to @paulmedynski - no!

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfoDto.cs
```diff
@@ -0,0 +1,47 @@
+﻿using System.Text.Json;
+using System.Text.Json.Serialization;
+
+#nullable enable
+
+namespace Microsoft.Data.SqlClient.UserAgent;
+internal class UserAgentInfoDto
```
benrr101, line 12: It seems unnecessary to make this a class - it could probably be done as a struct

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfoDto.cs
```diff
@@ -0,0 +1,47 @@
...
+    [JsonPropertyName(OsJsonKey)]
+    public OsInfo? OS { get; set; }
+
+    [JsonPropertyName(ArchJsonKey)]
+    public string? Arch { get; set; }
+
+    [JsonPropertyName(RuntimeJsonKey)]
+    public string? Runtime { get; set; }
+
+    public class OsInfo
```
benrr101, line 42: Likewise, this could probably be a struct?

src/Microsoft.Data.SqlClient/tests/UnitTests/UserAgentInfoTests.cs
```diff
@@ -0,0 +1,154 @@
...
+    ///   2. Truncation verification
+    ///   3. Payload size adjustment and field dropping
+    ///   4. DTO JSON contract (key names and values)
+    ///   5. Combined truncation, adjustment, and serialization
+    /// </summary>
+    public class UserAgentInfoTests
+    {
+        // 1. Cached payload is within the 2,047‑byte spec and never null
+        [Fact]
+        public void CachedPayload_IsNotNull_And_WithinSpecLimit()
```
benrr101, line 31: If we can, it would be good to structure these as 

```
// Arrange
.. any kind of setup work

// Act
.. preferably one line of action to exercise the desired code

// Assert
.. whatever checks need to be done to validate the behavior
```

benrr101 (changes requested):
😬 I'm gonna force braces for if statements if it's the last thing I do

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
samsharma2700: Should not by default, but need to remove this as we changed the implementation. Good catch!

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
samsharma2700: Is it a new guideline we are following? Haven't seen alphabetical ordering in other classes.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
samsharma2700: We do target `net462` which does not support `OSPlatform.FreeBSD` : https://apisof.net/catalog/138b234eefee5c6692118fe8c4d64920
Will change the check to reflect that.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
samsharma2700: Fair point, will do that.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfoDto.cs
samsharma2700: We do drop this field though.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfoDto.cs
samsharma2700: DTO does not satisfy criteria for a struct: https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/choosing-between-class-and-struct
1). Instance size should be under 16 bytes for a struct. `DTO` holds 5 strings and one 'OSInfo', around 48+bytes, triple the guideline.
2). We are mutating this object in `UserAgentInfo` by dropping fields, serializing and dropping again if needed(structs need to be immutable).
3). There won't be any perf benefits as a struct would introduce boxing allocations during serializations.

src/Microsoft.Data.SqlClient/tests/UnitTests/UserAgentInfoTests.cs
samsharma2700: Yes, agreed. I have tried to follow this structure as much as possible but grouped statements logically together. Will double check if there are any changes in accordance with the structure.

src/Microsoft.Data.SqlClient/tests/UnitTests/UserAgentInfoTests.cs
samsharma2700: In this particular test all values are replaced as `xxxx...` so driver name and version will not be the default values.

src/Microsoft.Data.SqlClient/tests/UnitTests/UserAgentInfoTests.cs
samsharma2700: Same as above, all values are still `xxxx...` since we pass the DTO values directly and are only testing `AdjustJsonPayloadSize` functionality.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
paulmedynski: These currently reflect the order of the JSON fields defined in the spec.  I would prefer we retain that order throughout the implementation.  Perhaps a comment explaining that here would help.

src/Microsoft.Data.SqlClient/tests/UnitTests/UserAgentInfoTests.cs
paulmedynski: Let's not introduce a bunch of temp variables just to capture state from `Act` and then `Assert` it a few lines below.  That becomes error prone and can obscure what's happening in the tests.  I would prefer to see:
```
Assert.Equal(foo.SomeAction(), "expected result");
```

than

```
var actualResult = foo.SomeAction();
...
Assert.Equal(actualResult, expectedResult);
```

If there is complicated actions and it makes sense to capture output along the way, then sure.  But I would advise against introducing ways to break a test just to follow a pattern.

src/Microsoft.Data.SqlClient/tests/UnitTests/UserAgentInfoTests.cs
paulmedynski: Is there a way to test that you produce the expected real values?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfoDto.cs
paulmedynski: I thought we drop the entire Os field, but never just the Type field.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
benrr101: We're definitely going to have something like this once the common project is finished up. Didn't want you jumping the gun :)

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfoDto.cs
benrr101: Ok fair enough. I'll have to give those docs a read. I was primarily operating under the assumption that it'd perform better to put the object on the stack rather allocating on the heap, but your point about boxing is a good reason not to do that.

src/Microsoft.Data.SqlClient/tests/UnitTests/UserAgentInfoTests.cs
benrr101: I think it's important to make it very clear what the test is testing. We can discuss this more, but it is much clearer to have

```C#
// Arrange
var foo = new MyObject();

// Act
var result = foo.SomeAction()

// Assert
Assert.Equal("expected result", result);
```

This indicates that the test is specifically testing what SomeAction does (not the creation of the object). And it doesn't bury the testing action inside the assertion. 

Granted, this is a contrived example. But I'd still like to see comments to separate the steps of the test. In some situations I'll sign off on something like

```C#
// Act / Assert
Assert.Equal("expected result", foo.SomeAction());
```

But, it still should be called out that the action under test is occurring within.

Sidenote: yeah ... I dislike xunit's `(expected, actual)` pattern, too.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
```diff
@@ -53,10 +49,10 @@ internal static class UserAgentInfo
     /// Maximum number of bytes allowed for the user agent json payload.
     /// Payloads larger than this may be rejected by the server.
     /// </summary>
-    public const int JsonPayloadMaxBytes = 2047;
+    internal const int JsonPayloadMaxBytes = 2047;
 
     private const string DefaultJsonValue = "Unknown";
-    private const string DriverName = "MS-MDS";
+    internal const string DriverName = "MS-MDS";
```
benrr101, line 60: *sigh* I know I'm looking like a big grump now, but ... I'm really gonna be a stickler for ordering stuff as per https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1201.md where internal should go before private.

src/Microsoft.Data.SqlClient/tests/UnitTests/UserAgentInfoTests.cs
```diff
@@ -89,33 +97,94 @@ public void AdjustJsonPayloadSize_StripsLowPriorityFields_When_PayloadTooLarge()
...
+            var options = new JsonSerializerOptions
+            {
+                PropertyNamingPolicy = null,
+                DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull,
+                WriteIndented = false
             };
 
-            string json = JsonSerializer.Serialize(dto);
-            using JsonDocument doc = JsonDocument.Parse(json);
+            string json = JsonSerializer.Serialize(dto, options);
```
benrr101, line 248: See, what I'm saying about the test structure is that it's not easy to see at-a-glance which bit is actually under test. It might be clear in your head (and I'm guilty of this too), but I have no idea which part is the action.

src/Microsoft.Data.SqlClient/tests/UnitTests/UserAgentInfoTests.cs
```diff
@@ -89,33 +97,94 @@ public void AdjustJsonPayloadSize_StripsLowPriorityFields_When_PayloadTooLarge()
...
-            Assert.Equal("r", root.GetProperty(UserAgentInfoDto.RuntimeJsonKey).GetString());
+            // always expected
+            Assert.Equal(driver,
+                root.GetProperty(UserAgentInfoDto.DriverJsonKey).GetString());
+            Assert.Equal(version,
+                root.GetProperty(UserAgentInfoDto.VersionJsonKey).GetString());
 
-            // Assert – nested os object
+            // optional Arch
+            if (dto.Arch is null)
```
benrr101: In my opinion, this test is doing too many scenarios...

benrr101 (changes requested):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
samsharma2700: Makes sense, I've reordered them.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfoDto.cs
samsharma2700: As we discussed, we'll retain this order for this class to reflect the spec doc but will sort alphabetically otherwise.

src/Microsoft.Data.SqlClient/tests/UnitTests/UserAgentInfoTests.cs
samsharma2700: Valid points, I have tried to strike balance between both of these philosophies and preferring logical grouping in case test becomes too convoluted. I have also added summary and comments grouping steps together. Hope this improves readability.

src/Microsoft.Data.SqlClient/tests/UnitTests/UserAgentInfoTests.cs
samsharma2700: Added a new test that verifies we populate the correct default values for driver name and version.

src/Microsoft.Data.SqlClient/tests/UnitTests/UserAgentInfoTests.cs
samsharma2700: Added a new test to verify the field value handling.

This is a special case because `UserAgentInfo` normally populates default values automatically. Here, we intentionally bypass `UserAgentInfo` to create our own `DTO` directly, without calling `BuildDto()` or `TruncateOrDefault()`.
As a result, no truncation of field values occurs. The main purpose of this test is to validate the field-dropping logic under these conditions.

src/Microsoft.Data.SqlClient/tests/UnitTests/UserAgentInfoTests.cs
samsharma2700: Yes, good point. Added a summary along with `Arrange`, `Act` and `Assert` grouping comments.

src/Microsoft.Data.SqlClient/tests/UnitTests/UserAgentInfoTests.cs
samsharma2700: I understand the test may feel tedious. I’ve added additional comments to improve readability.
This section needs to be tested in one run to properly verify the field-dropping logic.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
```diff
@@ -20,46 +16,49 @@ namespace Microsoft.Data.SqlClient.UserAgent;
...
-    private const int VersionMaxChars = 16;
+    private const int DriverNameMaxChars = 16;
 
     /// <summary>
-    /// Maximum number of characters allowed for the operating system type.
+    /// Maximum number of bytes allowed for the user agent json payload.
+    /// Payloads larger than this may be rejected by the server.
     /// </summary>
-    private const int OsTypeMaxChars = 16;
+    internal const int JsonPayloadMaxBytes = 2047;
```
paulmedynski, line 36: Yuck - now the JSON field constants are mixed in with other constants, and they're no longer in logical order.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
```diff
@@ -20,46 +16,49 @@ namespace Microsoft.Data.SqlClient.UserAgent;
...
-    private const string DriverName = "MS-MDS";
+    internal const string DefaultJsonValue = "Unknown";
+    internal const string DriverName = "MS-MDS";
 
-    private static readonly UserAgentInfoDto _dto;
-    public static readonly byte[] _cachedPayload;
+    private static readonly UserAgentInfoDto s_dto;
+    private static readonly byte[] s_cachedPayload;
+
+    public static byte[] CachedPayload => s_cachedPayload;
```
paulmedynski: This public property deserves some documentation, most importantly that it doesn't calculate anything.

Can this be readonly?

Is `CachedPayload` the best name?  Maybe it's fine when you add documentation.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
```diff
@@ -73,11 +72,55 @@ private enum OsType
...
+        }
+
+        payload = JsonSerializer.SerializeToUtf8Bytes(dto, options);
+        if (payload.Length <= JsonPayloadMaxBytes)
+        {
+            return payload;
+        }
+            
+        dto.OS = null; // drop OS entirely
+        // Last attempt to send minimal payload driver + version only
```
paulmedynski: Note that we know driver + version cannot be larger than the max, pointing to the explanation in the `BuildDto()` commentary.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
```diff
@@ -73,11 +72,55 @@ private enum OsType
...
+        return JsonSerializer.SerializeToUtf8Bytes(dto, options);
     }
 
-    static UserAgentInfoDto BuildDto()
+    internal static UserAgentInfoDto BuildDto()
     {
         // Note: We serialize 6 fields in total:
```
paulmedynski, line 97: I think this entire comment belongs in `AdjustJsonPayloadSize()`.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfoDto.cs
```diff
@@ -10,19 +11,20 @@ internal class UserAgentInfoDto
...
     public const string VersionJsonKey = "version";
     public const string OsJsonKey = "os";
     public const string ArchJsonKey = "arch";
     public const string RuntimeJsonKey = "runtime";
 
-
-    // TODO: Does this need to be nullable?
     [JsonPropertyName(DriverJsonKey)]
-    public string? Driver { get; set; }
+    public string Driver { get; set; } = UserAgentInfo.DriverName;
```
paulmedynski: These initialization values don't belong here.  The DTO doesn't know anything about where its values come from.  That's `UserAgentInfo`'s responsibility.

I think `string.Empty` is the correct choice here.

src/Microsoft.Data.SqlClient/tests/UnitTests/UserAgentInfoTests.cs
paulmedynski: You can still verify that the present fields contain the expected 'xxx...' values, truncated to their respective max lengths.

src/Microsoft.Data.SqlClient/tests/UnitTests/UserAgentInfoTests.cs
```diff
@@ -60,62 +91,140 @@ public void AdjustJsonPayloadSize_StripsLowPriorityFields_When_PayloadTooLarge()
...
             Assert.True(root.TryGetProperty(UserAgentInfoDto.VersionJsonKey, out _));
 
-            // Low‑priority fields must be gone.
+            // Low-priority fields should have been removed(arch and runtime)
             Assert.False(root.TryGetProperty(UserAgentInfoDto.ArchJsonKey, out _));
             Assert.False(root.TryGetProperty(UserAgentInfoDto.RuntimeJsonKey, out _));
 
-            // If the "os" object survived, only its "type" sub‑field may remain.
+            // if OS block remains, only its Type sub-field may survive
             if (root.TryGetProperty(UserAgentInfoDto.OsJsonKey, out JsonElement os))
```
paulmedynski: This if-statement allows the test to pass whether or not the OS block is present.  So we're not really testing one of the adjust scenarios.  Here's what we need:

1.  Keep this test, rename it to indicate that _all_ low priority fields are stripped, and then replace this if-statement with `Assert.False(root.TryGetProperty(OsJsonKey))`.
2.  Add a new test that checks the middle-ground adjusting, where only arch, runtime, and os.description are stripped (i.e. the return on UserAgentInfo line 115).

src/Microsoft.Data.SqlClient/tests/UnitTests/UserAgentInfoTests.cs
```diff
@@ -60,62 +91,140 @@ public void AdjustJsonPayloadSize_StripsLowPriorityFields_When_PayloadTooLarge()
...
+                    Assert.False(os.TryGetProperty(UserAgentInfoDto.OsInfo.DetailsJsonKey, out _));
+                }
+                else
+                {
+                    Assert.Equal(dto.OS.Details, os.GetProperty(UserAgentInfoDto.OsInfo.DetailsJsonKey).GetString());
+                }
+            }
         }
 
         // 5. End-to-end test that combines truncation, adjustment, and serialization
```
paulmedynski: Maybe this is the missing test I was referring to above, just not named/documented accurately.

paulmedynski (changes requested):
Getting closer!

src/Microsoft.Data.SqlClient/tests/UnitTests/UserAgentInfoTests.cs
```diff
@@ -0,0 +1,263 @@
+﻿using System.Text;
```
cheenamalhotra: Include .NET headers

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfoDto.cs
```diff
@@ -0,0 +1,49 @@
+﻿using System.Text.Json;
```
cheenamalhotra: Include headers

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
```diff
@@ -0,0 +1,348 @@
+﻿using System;
```
cheenamalhotra: Include headers

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfoDto.cs
```diff
@@ -0,0 +1,49 @@
...
+using System.Text.Json.Serialization;
+using Microsoft.Data.Common;
+
+#nullable enable
+
+namespace Microsoft.Data.SqlClient.UserAgent;
```
cheenamalhotra, line 11: For consistency in reading, keep curly brackets for namespace definition, as in other classes.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
samsharma2700: Yeah we have to decide between logical groupings or lexicographical ordering. The groupings are maintained in case of `UserAgentInfoDto`

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfo.cs
samsharma2700: Will add some documentation here. We cannot mark properties as `readonly`. In case of `s_cachedPayload` despite it being `readonly` someone can still modify it after initialization. I think `CachedPayload` can be `ReadOnlyMemory<byte>`, which would communicate the read-only intent at the API boundary.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/UserAgent/UserAgentInfoDto.cs
samsharma2700: There was feedback to move to file-scoped namespace declarations : https://github.com/dotnet/SqlClient/pull/3489#discussion_r2222370372

src/Microsoft.Data.SqlClient/tests/UnitTests/UserAgentInfoTests.cs
samsharma2700: Great catch! will do that.

paulmedynski (dismissed):

benrr101 (dismissed):
👍

paulmedynski:
Abandoned in favour of #3582.

PR #3492, opened 2025-07-19
Cleanup | Centralise AppContext switches

edwardneal:
## Description

We try to centralise all access to SqlClient's AppContext switches via the `LocalAppContextSwitches` class. There are currently a few cases where we don't do this though:
* netcore's globalization invariant mode detection lives in SqlConnection
* Switch.Microsoft.Data.SqlClient.TruncateScaledDecimal is handled by TdsParser on both targets
* Switch.Microsoft.Data.SqlClient.UseManagedNetworkingOnWindows is handled by TdsParserStateObjectFactory

The first port of call is to relocate all of these to the existing LocalAppContextSwitches class. This slightly helps in the merge of SqlConnection.

I've also added an `ILLink.Substitutions.xml` resource to the output. This is a trimming improvement, and is documented [here](https://github.com/dotnet/runtime/blob/main/docs/tools/illink/data-formats.md). It allows referencing applications to specify a `RuntimeHostConfigurationOption` to set an AppContext switch, which the IL trimmer will reference and remove the entire managed or native SNI. There's a little more work to come on this once trimming is ready.

For comparison purposes, this is modelled in part on CsWinRT:
* [ILLink.Substitutions.xml](https://github.com/microsoft/CsWinRT/blob/master/src/WinRT.Runtime/Configuration/ILLink.Substitutions.xml)
* Underlying [FeatureSwitches](https://github.com/microsoft/CsWinRT/blob/master/src/WinRT.Runtime/Configuration/FeatureSwitches.cs) type
* Usage of [`RuntimeHostConfigurationOption`](https://github.com/microsoft/CsWinRT/blob/master/nuget/Microsoft.Windows.CsWinRT.targets#L339-L374)

I've used sizoscope to verify that I can use RuntimeHostConfigurationOption, publish an application and see the corresponding SNI's types are removed from the output.

## Issues

Partially relates to #1261 and #1942.

## Testing

Unit tests pass locally.

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/LocalAppContextSwitches.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/LocalAppContextSwitches.cs
@@ -327,7 +327,7 @@ namespace Microsoft.Data.SqlClient
         internal const string UseManagedNetworkingOnWindowsString = "Switch.Microsoft.Data.SqlClient.UseManagedNetworkingOnWindows";
 
         private static Tristate s_globalizationInvariantMode;
-        private static Tristate s_useManagedNetworkingOnWindows;
+        private static Tristate s_useManagedNetworking;
 
         /// <summary>
         /// .NET Core 2.0 and up supports Globalization Invariant mode, which reduces the size of the required libraries for
@@ -389,22 +389,22 @@ namespace Microsoft.Data.SqlClient
         {
             get
             {
-                if (s_useManagedNetworkingOnWindows == Tristate.NotInitialized)
+                if (s_useManagedNetworking == Tristate.NotInitialized)
                 {
                     if (!OperatingSystem.IsWindows())
                     {
-                        s_useManagedNetworkingOnWindows = Tristate.True;
+                        s_useManagedNetworking = Tristate.True;
                     }
                     else if (AppContext.TryGetSwitch(UseManagedNetworkingOnWindowsString, out bool returnedValue) && returnedValue)
                     {
-                        s_useManagedNetworkingOnWindows = Tristate.True;
+                        s_useManagedNetworking = Tristate.True;
                     }
                     else
                     {
-                        s_useManagedNetworkingOnWindows = Tristate.False;
+                        s_useManagedNetworking = Tristate.False;
                     }
                 }
-                return s_useManagedNetworkingOnWindows == Tristate.True;
+                return s_useManagedNetworking == Tristate.True;
             }
         }
 #else
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/LocalAppContextSwitches.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/LocalAppContextSwitches.cs
@@ -15,14 +15,14 @@ namespace Microsoft.Data.SqlClient
             True = 2
         }
 
-        internal const string MakeReadAsyncBlockingString = @"Switch.Microsoft.Data.SqlClient.MakeReadAsyncBlocking";
-        internal const string LegacyRowVersionNullString = @"Switch.Microsoft.Data.SqlClient.LegacyRowVersionNullBehavior";
-        internal const string SuppressInsecureTlsWarningString = @"Switch.Microsoft.Data.SqlClient.SuppressInsecureTLSWarning";
-        internal const string UseMinimumLoginTimeoutString = @"Switch.Microsoft.Data.SqlClient.UseOneSecFloorInTimeoutCalculationDuringLogin";
-        internal const string LegacyVarTimeZeroScaleBehaviourString = @"Switch.Microsoft.Data.SqlClient.LegacyVarTimeZeroScaleBehaviour";
-        internal const string UseCompatibilityProcessSniString = @"Switch.Microsoft.Data.SqlClient.UseCompatibilityProcessSni";
-        internal const string UseCompatibilityAsyncBehaviourString = @"Switch.Microsoft.Data.SqlClient.UseCompatibilityAsyncBehaviour";
-        internal const string UseConnectionPoolV2String = @"Switch.Microsoft.Data.SqlClient.UseConnectionPoolV2";
+        private const string MakeReadAsyncBlockingString = @"Switch.Microsoft.Data.SqlClient.MakeReadAsyncBlocking";
+        private const string LegacyRowVersionNullString = @"Switch.Microsoft.Data.SqlClient.LegacyRowVersionNullBehavior";
+        private const string SuppressInsecureTlsWarningString = @"Switch.Microsoft.Data.SqlClient.SuppressInsecureTLSWarning";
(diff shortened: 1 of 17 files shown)


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/LocalAppContextSwitches.cs
```diff
@@ -296,5 +298,125 @@ public static bool UseConnectionPoolV2
...
+                    }
+                }
+                return s_truncateScaledDecimal == Tristate.True;
+            }
+        }
+
+#if NET
+        private const string GlobalizationInvariantModeString = @"System.Globalization.Invariant";
+        private const string GlobalizationInvariantModeEnvironmentVariable = "DOTNET_SYSTEM_GLOBALIZATION_INVARIANT";
+        internal const string UseManagedNetworkingOnWindowsString = "Switch.Microsoft.Data.SqlClient.UseManagedNetworkingOnWindows";
```
paulmedynski: private ?

paulmedynski (changes requested):
Looks good, but need to also update this test helper:

`src/Microsoft.Data.SqlClient/tests/Common/LocalAppContextSwitchesHelper.cs`

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/LocalAppContextSwitches.cs
```diff
@@ -296,5 +298,125 @@ public static bool UseConnectionPoolV2
...
+                    {
+                        s_truncateScaledDecimal = Tristate.False;
+                    }
+                }
+                return s_truncateScaledDecimal == Tristate.True;
+            }
+        }
+
+#if NET
+        private const string GlobalizationInvariantModeString = @"System.Globalization.Invariant";
```
mdaigle: Please place these with the other strings.

src/Microsoft.Data.SqlClient/src/Resources/ILLink.Substitutions.Windows.xml
```diff
@@ -0,0 +1,9 @@
+<linker>
+  <assembly fullname="Microsoft.Data.SqlClient">
+    <type fullname="Microsoft.Data.SqlClient.LocalAppContextSwitches">
+      <!-- Enable the unused SNI to be trimmed based upon the publish-time value of the Switch.Microsoft.Data.SqlClient.UseManagedNetworkingOnWindows switch -->
+      <method signature="System.Boolean get_UseManagedNetworking()" body="stub" value="false" feature="Switch.Microsoft.Data.SqlClient.UseManagedNetworkingOnWindows" featurevalue="false" />
```
mdaigle, line 5: I'm a little bit hesitant to use this type of substitution because it's hard to know that it's there when you're just reading through the source code. Did you consider using a preprocessor directive wrapping the get_UseManagedNetworking() method? We could set that at build time and have a path that implements it to always return a set value. It would also be obvious in the code itself that there's a compile time option that modifies it.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/LocalAppContextSwitches.cs
```diff
@@ -296,5 +298,125 @@ public static bool UseConnectionPoolV2
...
+
+        private static Tristate s_globalizationInvariantMode;
+        private static Tristate s_useManagedNetworking;
+
+        /// <summary>
+        /// .NET Core 2.0 and up supports Globalization Invariant mode, which reduces the size of the required libraries for
+        /// applications which don't need globalization support. SqlClient requires those libraries for core functionality,
+        /// and will throw exceptions later if they are not present. This switch allows SqlClient to detect this mode early.
+        /// </summary>
+        public static bool GlobalizationInvariantMode
```
mdaigle, line 305: Moving these switches here enables caching of the switch value. The only way to get a new value is to start a new process. That's probably a good thing so that all of our switches work in the same way, but it's possible someone relies on being able to modify these switch values dynamically.  For that reason, I'd consider it a breaking change and want to make sure the release notes reflect that.

edwardneal: The ParametersTest failures highlight this too; I'm currently vaccillating between whether it'd be better to fix the behaviour or to change the test.

The [documentation](https://learn.microsoft.com/en-us/sql/connect/ado-net/appcontext-switches?view=sql-server-ver17) suggests to me that it can only be set once though, so I think someone relying on the newer switches being dynamically changeable was always relying on an implementation detail:

> To enable the previous behavior of truncation, you can set the AppContext switch "Switch.Microsoft.Data.SqlClient.TruncateScaledDecimal" to true at application startup

> To enable the use of a managed SNI implementation, you can set the AppContext switch "Switch.Microsoft.Data.SqlClient.UseManagedNetworkingOnWindows" to true at application startup

I'd read this as something which we can only set once. I'm not sure we've ever tested situations where some connections in the pool use the native SNI and others use the managed SNI.

src/Microsoft.Data.SqlClient/src/Resources/ILLink.Substitutions.Windows.xml
edwardneal: It definitely feels fragile, and I'm not particularly fond of the way we need to have another platform-specific XML file for non-Windows targets.

I did consider a preprocessor directive, but AOT/trimmed publishing is just triggered by a new MSBuild target in the downstream library's code; we'd need to build more combinations of the library and insert a step into the MSBuild target which swaps the correct platform's AOT-compatible binary in. That seems like it'd be more fragile, and it'd leak the details of our trim compatibility internals to downstream rather than letting the linker work.

It definitely makes sense to add a comment to the property referring to the substitutions resource though.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/LocalAppContextSwitches.cs
edwardneal: This is still referenced by some logging in TdsParserStateObjectFactory. It'd be nice to make these switch names private though, I'll reorganise it.

src/Microsoft.Data.SqlClient/src/Resources/ILLink.Substitutions.Windows.xml
mdaigle: I see, makes sense. I'm happy then as long as we have that comment.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/LocalAppContextSwitches.cs
mdaigle: I read it the same way. We should have the behavior match our published docs. It doesn't _really_ make sense to change them dynamically anyway.

edwardneal:
Thanks both. I've added the review comments, and the failing `TestScaledDecimalParameter_*` tests should now run correctly (as should the `AdjustPrecScaleForBulkCopy` test.) There are no other references to the AppContext switches in the tests as far as I can see.

Could you run CI please?

PR #3494, opened 2025-07-21
Tests | Active Issues 1

benrr101:
## Description
Re-enables DataStreamTests.RunAllTestsForSingleServer_TCP previously blocked by #5540

In Release mode, it works just fine without any changes. In Debug mode, though, it has issues with async behavior. The test harness for async test created (but did not start) an async operation, attempted to run the operation in parallel (which should fail as another async operation is in progress), then disposes the test harness and waits on the first async task. This fails because the connection is closed when the test harness is disposed, failing the test. It's unclear to me why the test harness disposes the connection or why it exists at all, but that's not the point. So, to resolve those issues, I followed a pattern used in another part of the test code that asserts the final wait throws. Totally unnecessary imho, but if we have to wait something that'll throw, then we might as well make sure it throws.

Another issue happens deep in the TdsParserStateObject in Debug mode. There is a Debug.Assert in there that validates the previousTimeoutState is Stopped. This check is failing in the test scenarios. However, immediately following the assert, we have an if statement that checks the exact same thing. Theoretically if the assert is correct, we always run the code in the if statement. But, since we have the conditional, we must expect that the asserted condition might not be met. It implies to me that the assert shouldn't exist. So I removed it and the test now passes.

Since this is a lengthy explanation, I figure I should submit this PR by itself.

## Issues
#5540 - which I don't think exists in GH.

## Testing
Running the test locally, it passes


paulmedynski (approved):

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
```diff
@@ -595,10 +595,21 @@ public static string GetUniqueNameForSqlServer(string prefix, bool withBracket =
                 Environment.MachineName,
                 DateTime.Now.ToString("yyyy_MM_dd", CultureInfo.InvariantCulture));
             string name = GetUniqueName(extendedPrefix, withBracket);
-            if (name.Length > 128)
+
+            // Truncate to no more than 128 characters.
+            const int maxLen = 128;
+            if (name.Length > maxLen)
             {
-                throw new ArgumentOutOfRangeException("the name is too long - SQL Server names are limited to 128");
```
paulmedynski, line 600: Running the tests on macOS generates names longer than 128, so throwing here isn't helpful.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataStreamTest/DataStreamTest.cs
```diff
@@ -757,7 +764,10 @@ private static void ExecuteXmlReaderTest(string connectionString)
                         "select employeeId from employees where employeeid < 3 for xml auto;";
                     using (xr = cmd.ExecuteXmlReader())
                     {
-                        string[] expectedResults =
+                        // The XML elements may be returned in any order, so we
```
paulmedynski, line 767: macOS was returning rows in a different order than the other OSes.

paulmedynski (approved):

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
```diff
@@ -595,14 +595,10 @@ public static string GetUniqueNameForSqlServer(string prefix, bool withBracket =
...
+            // Truncate to no more than 128 characters.
             const int maxLen = 128;
-            if (name.Length > maxLen)
-            {
-                throw new ArgumentOutOfRangeException(
-                    $"GetUniqueNameForSqlServer(): Generated name \"{name}\" " +
-                    $"exceeds SQL Server limit of {maxLen} characters");
-            }
-            return name;
+            return name.Length > maxLen ? name.Substring(0, maxLen) : name;
```
mdaigle: I've seen this error pop up on a few builds randomly. I wonder if 1ES changed the machine names or something and made them longer.

PR #3495, opened 2025-07-21
CodeQL | RSA Encryption Padding (Part 2?)

benrr101:
## Description
This is a test to see if tests will pass with Pss padding enabled. I don't think it will, but I need to verify.

## Issues
S360 issue for RSA encryption padding scheme.

## Testing
Local tests that should use that codepath tests, so I need a CI run to validate it further


paulmedynski (approved):

benrr101:
Closing as this will break backwards compat and cross-driver compatibility :(

PR #3498, opened 2025-07-22
Remove BuildTests target

paulmedynski:
## Description

Removed the BuildTests target from `build.proj` since it doesn't function.  MSBuild has problems when building in "solution mode" and performing multiple different TargetFramework restores, so I just removed the target.  The `BUILDGUIDE.md` was updated accordingly.

Additional tweaks were made to allow `dotnet build` to function on Linux using the default target.

Some stale scripts in the repo root were removed.

## Issues

Fixes discussion #3147.


apoorvdeshmukh (approved):
LGTM. Tried dotnet build on WSL and it works fine.

override.ps1
cheenamalhotra, line 1: @mdaigle could you confirm if this file isn't being used/unwanted for the repository?

mdaigle: It's not used. The script is duplicated in a task group called by one of the SNI pipelines.

mdaigle: This version also has a hardcoded sni version

apoorvdeshmukh (approved):

PR #3505, opened 2025-07-23
add test and simplify code to fix problems

### LINKED ISSUE #3504: SqlClient gets stuck asynchronously reading string values in certain conditions
### Describe the bug
On a highly specific arrangement of data, SqlClient will be stuck in a forever loop in `TryReadPlpUnicodeChars`, attempting to asynchronously read data. 

### To reproduce
This issue seems to be latency sensitive as well so using a remote database like SQL Azure is advised.

Setup a database and test project as described in the following gist:
https://gist.github.com/Suchiman/1b70ab4d8b549b5c647f495533ee39c3

If you run the test project, it will print:
```
Starting query
Retrieved records in 00:00:00.5410259
Starting query
```
and now the program is forever stuck on the second iteration.

### Expected behavior
Will continously print
```
Starting query
Retrieved records in ...
```
without ever getting stuck

### Further technical details
Microsoft.Data.SqlClient version: 6.1.0-preview2.25178.5
.NET target: .NET 9
SQL Server version: SQL Azure / SQL Server 2022
Operating system: App Service / Windows 11

Wraith2:
Fixes #3504

The sample data provided in the original issue by @Suchiman showed me that there were multiple problems happening when reading long strings. The data in the new test would not successfully complete the first iteration without asserting and would throw an exception on the second iteration. The lengths of the values and their combinations over multiple packets are important, even the column name lengths contribute to the offsets so they have not been changed. The data itself is randomised from a production sample.

The changes add in asserts to ensure that if we are expecting to continue that we have a stored buffer to continue with. There was a case where no buffer was present at a continue point causing the tail end of a multi-packet string to be returned to the user successfully causing an incorrect data read.

The conditions in which buffers and lengths are stored in the snapshot are relaxed. This brings the code more into line with the more relaxed handling of replay to continue that was added in previous commits. Overall I'm happier with this lessening of complexity.


Wraith2:
@dotnet/sqlclientdevteam can I get a CI run please

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataReaderTest/DataReaderTest.cs
```diff
@@ -887,5 +887,196 @@ public static void CheckLegacyNullRowVersionIsEmptyArray()
...
...247YlBiTAXXn9OfWA5nfVHzAnREFf4lkxQmX7XHbBiD8hsjm3wHDsLLUsR9II9kDGUJSDp5RAW6gfXuk0K8Tf20I0lDbWF7w9fGJkPKtRdgoOJsPN1El0i3CWDkW5FxGwj4kRsLh4o02JKM371NAg3SwIRYFd8Sl15WN5yk6ZB9q3uheNGinAehq2CORuYlWnPjLBR1gI8ons9cwQtwWbZUJ1BkoG0k517gtsk0cbfB0OpyliVOQGWH0XX2Mwj5jymwuNlqjNudb9HYE1XmnyCTpDgKTY0NkDeP7nRidmUlv8UyoDD7Z2L7ZlmJ0Oz5yQx8ccUV9dJO0L8dT08VOUIs5TQ0TWSYkd5hYSaAGvgYXykZLSYNQs4aaMO1wa7LbK6OTjIJnj50o3JRCudq86s08SQ4fwSJcRO9oq2UWtdeZX62z10OdkrVrbRIdTyZlEQSVbTEb77qipRmfORcvlIaz7WGgWBUz7lpUdmOQzfgmEpiNUJiW5IZJTdYmRNYdEM92tLLdYPhiMsjrCqJ73Vd80RoOnrnUwrlNbhBRydHwCjJsYcHbuFywvay5VyqYAJ7cvn1VUGARXxO5yHDO5XJeb11BYQ3NiLcLpNyrU0IdEcffqTFtBq4S0e02aoVUEJAHWFgsGyMJ9MkdnsY11X18YRLqe0TCN0sRMeXRjQaLL3mY6Qf12jqj6L645DVUXGoRkadj6EwJRouxkQiWP7crgHknwyywSSBLKgJsl6kDEsYCSjQP9Q2LSDdWnIbW1qIc4ksKn5TXXB3MgQ4i251utgQf665Plxg33VJpRvP1ueEGLudi2VOC8YTMDULevIXYLwzfIUA53EHPnjlzfIS8v6BzBL2DxoIBWns0J5Ci9rArjPWB3h6UkeAeqXLJsPrfKqesYL5ZJAJPopIXdzRuqseqie1akOWQDcOxxDQsvEWVBfjuZpvaectHK7HlxzgeXP14Aa2tqyUfVpM5meDtcGcPSd1SfNf0RGvD16rgiq225T7yx4c0uBJTRQekF2CXvGUDTDeLxR08GKscXlgBCZOgeOjd4kX3UGXSY7T0KPpKNgLLjwLkPQBqXrJcThYDimnm1zn73HfoPO7qsPqXmB8S380lMQqrXOLYizumMutVQX1D2FWLNLkYlKpUQb8jkT0QX5AfcNPFFKOefyZltq7JCvf4gh7oPTMKe8VVPoGA
```
cheenamalhotra: Why not randomize and generate instead?

cheenamalhotra: Remove extra lines

cheenamalhotra: Use `DataTestUtility.DropTable(connection, tableName);` instead.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataReaderTest/DataReaderTest.cs
```diff
@@ -887,5 +887,196 @@ public static void CheckLegacyNullRowVersionIsEmptyArray()
...
+            {
+                Id = id;
+                DocumentIdentificationId = documentIdentificationId;
+                Name = name;
+                Value = value;
+            }
+        }
+
+        [ConditionalFact(typeof(DataTestUtility), nameof(DataTestUtility.AreConnStringsSetup))]
+        public static async Task CanReadAwkwardDataLengths()
```
cheenamalhotra, line 909: You seem to have modified 3 paths (TryReadStringWithEncoding, TryReadByteArrayWithContinue, and TryReadPlpUnicodeCharsWithContinue) where this issue can occur - can you add test coverage for all of the changes?

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataReaderTest/DataReaderTest.cs
Wraith2: Because while debugging it remembering the substrings and finding them in the code was important. As I noted in the original post the lengths of all the fields are important and it took a couple of hours to get a stable reproduction from the original reporting user that worked for me. The more changes I make the more likely it is to disrupt and make the test useless.

Wraith2: Tidied indenting. There aren't any extra lines.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataReaderTest/DataReaderTest.cs
Wraith2: All the paths already have coverage. The code changes retain the existing functionality. Proving they were susceptible to the problem found with chars specifically would require that I write some sort of fuzzer to generate large amounts of varied input data and make sure it can be read correctly. It's an idea that I've considered but is not sensible to do for this PR because of the need to quickly fix it, it can cause silent bad data to be returned to users, highly dangerous.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataReaderTest/DataReaderTest.cs
```diff
@@ -887,5 +887,191 @@ public static void CheckLegacyNullRowVersionIsEmptyArray()
...
```
mdaigle, line 909: Please add some comments to this test describing its purpose and advising that the test data should not be changed.

mdaigle (commented):
one small request

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataReaderTest/DataReaderTest.cs
Wraith2: Done:
```csharp
// The test data is taken from a production sample which exhibited multiple problems around async string reads.
//  The lengths of the field names and data values are important to reproducing the problems. The values have been
//  changed to randomized values. Changes to the column names or value lengths can cause the problems to move or
//  fail to reproduce entirely because of packet and data alignment. The data is not randomly generated to help with
//  debugging failures in the test.
```

Suchiman: Missing word:
> lengths can CAUSE the problems

Wraith2: Fixed.

paulmedynski (approved):

cheenamalhotra (approved):

PR #3516, opened 2025-07-25
Update AKV to depend on any MDS 6.x package

paulmedynski:
Updated the AKV .nuspec file to declare it depends on any MDS 6.x package.  This will make it compatible with the latest MDS 6.1.0 release, and any future MDS 6.x releases.


benrr101 (approved):
I guess since our next scheduled release is 7.0, it's safe to make this change. Otherwise, I'd say we are supposed to cap it at 6.2

apoorvdeshmukh (dismissed):

benrr101 (approved):

Issue #2971, opened 2024-11-04
Change SqlParameter to allow empty IEnumerable<SqlDataRecord>?


wjrogers:
At the beginning of the `GetActualFieldsAndProperties` branch for `IEnumerable<SqlDataRecord>`, this comment suggests it will set the value to `null` when the enumerable is empty:
https://github.com/dotnet/SqlClient/blob/9d5ca32e666b3da1cc27b4ab145367308998b147/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlParameter.cs#L1291-L1292

However, at the end, it throws an exception:
https://github.com/dotnet/SqlClient/blob/9d5ca32e666b3da1cc27b4ab145367308998b147/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlParameter.cs#L1413-L1416
> System.ArgumentException: There are no records in the SqlDataRecord enumeration. To send a table-valued parameter with no rows, use a null reference for the value instead.

This means I have to check for an empty `IEnumerable<SqlDataRecord>` before setting `SqlParameter.Value`, which means I have to materialize and/or enumerate the source. For efficiency, I would like to avoid that. Considering `SqlParameter` is already peeking at the enumeration to initialize the metadata, could it be changed to handle an empty enumerable the same as the value `null`?

sharafabacery:
hey can I contribute to solve this issue I see good first issue tag that mean would good to start as contributer but it will be first time it will be enough to solve it?

sharafabacery:
I need to try to get comfortable in open source could you assign it to me and give small example how this issue happen
?

PR #3518, opened 2025-07-26
Fix | Pass Empty List for SqlDataRecord (#2971)

sharafabacery:
## Description
Important topics to cover include:
Enable Passing Empty List of type SqlDataRecord and make it functionality like when value is null

- Description of the functionality.
I remove exception from SqlParameter.cs and in SqlCommand I figure where it if parameter is null what happen and adding small condition

## Issues
Fixes issue #2971 

## Testing
I create database and delete it when test finish to make test work and 
first test when list is empty and second one when it null and it works correctly
Screenshot 2025-07-25 025730
and this photo captured from my sql server profiler to prove that it sends default

### DIFF
--- a/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlCommand.cs
+++ b/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlCommand.cs
@@ -24,7 +24,6 @@ using Microsoft.Data.Sql;
 using Microsoft.Data.SqlClient.Server;
 using System.Transactions;
 using System.Collections.Concurrent;
-using System.Collections;
 
 // NOTE: The current Microsoft.VSDesigner editor attributes are implemented for System.Data.SqlClient, and are not publicly available.
 // New attributes that are designed to work with Microsoft.Data.SqlClient and are publicly documented should be included in future.
@@ -6266,9 +6265,7 @@ namespace Microsoft.Data.SqlClient
                         // Don't assume a default value exists for parameters in the case when
                         // the user is simply requesting schema.
                         // TVPs use DEFAULT and do not allow NULL, even for schema only.
-                        bool isEmptyValue = parameter.Value is null || (parameter.Value is ICollection collection && collection.Count == 0);
-                        bool allowsDefault = !inSchema || parameter.SqlDbType is SqlDbType.Structured;
-                        if (isEmptyValue && allowsDefault)
+                        if (parameter.Value == null && (!inSchema || SqlDbType.Structured == parameter.SqlDbType))
                         {
                             options |= TdsEnums.RPC_PARAM_DEFAULT;
                         }


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
```diff
@@ -1065,13 +1066,7 @@ public override void Cancel()
                         }
 
                         if (!_pendingCancel)
-                        {
```
benrr101, line 1068: I don't understand this change at all, especially since they seem to be completely unrelated to the focus of the PR. We don't want to be removing braces for if statements, and in this particular instance it completely upends the logic of the method.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
```diff
@@ -1763,7 +1758,6 @@ private Task InternalExecuteNonQuery(
                     asyncWrite,
                     isRetry,
                     methodName);
-
```
benrr101, line 1766: This doesn't really help readability.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
```diff
@@ -2587,7 +2581,6 @@ long firstAttemptStart
                                         {
                                             completion.TrySetResult(retryTask.Result);
                                         }
-                                    },
```
benrr101, line 2590: This change also makes no sense, and I don't see how it compiles at all. Why was this change made?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
```diff
@@ -3105,7 +3098,6 @@ private async Task<object> ExecuteScalarUntilEndAsync(SqlDataReader reader, Canc
         }
 
         /// <include file='../../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlCommand.xml' path='docs/members[@name="SqlCommand"]/ExecuteXmlReaderAsync[@name="default"]/*'/>
-        public Task<XmlReader> ExecuteXmlReaderAsync() =>
```
benrr101, line 3108: This change seems completely random, unrelated to the focus of the PR, and likely doesn't compile.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
```diff
@@ -4871,7 +4863,6 @@ internal SqlDataReader RunExecuteReader(
                 task: out unused,
                 usedCache: out _,
                 method: method);
-
```
benrr101, line 4874: Again, removing this line doesn't really aid readability and is unrelated to the PR focus.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlCommandTest.cs
```diff
@@ -518,5 +520,110 @@ public void ParameterCollectionTest()
                 cmd.Parameters.Remove(cmd.Parameters[0]);
             }
         }
+
+        #region fixing "Change SqlParameter to allow empty IEnumerable<SqlDataRecord>? #2971"
```
benrr101, line 524: I'm not sure what to do about these tests. I think someone with a bit more experience in the test area could provide more concrete feedback. @mdaigle ?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
```diff
@@ -6192,7 +6183,8 @@ private void SetUpRPCParameters(_SqlRPC rpc, bool inSchema, SqlParameterCollecti
                         // Don't assume a default value exists for parameters in the case when
                         // the user is simply requesting schema.
                         // TVPs use DEFAULT and do not allow NULL, even for schema only.
-                        if (parameter.Value == null && (!inSchema || SqlDbType.Structured == parameter.SqlDbType))
+                        if ((parameter.Value == null || (parameter.Value is ICollection collection && collection.Count == 0)) &&
```
benrr101, line 6186: Multi-line if conditions are really hard to read ... can we take these two main cases and extract them to variables right above the if statement?
like 
```
bool isEmptyValue = parameter.Value is null || (parameter.Value is ICollection collection && collection.Count == 0);
bool allowsDefault = !inSchema || parameter.SqlDbType is SqlDbType.Structured;
```

sharafabacery: Ok I will do it

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlCommandTest.cs
sharafabacery: All changes above could be because I enable formatting document after saving

sharafabacery: @benrr101

sharafabacery: If there any test need to be done i will do it?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
sharafabacery: All changes above could be because I enable formatting document after saving
@benrr101

edwardneal:
Without wanting to step on @benrr101's toes, does this address the issue and support `IEnumerable<SqlDataRecord>`?

I've written a simple test - gist for it is [here](https://gist.github.com/edwardneal/2e185ff82fb13c4aff9eb9fee41337e7). This test returns an enumerator which isn't backed by an ICollection (and can't be - it's an iterator method.) The test as-is fails with the error:

> `The incoming tabular data stream (TDS) remote procedure call (RPC) protocol stream is incorrect. Table-valued parameter 1 ("@tvp"), row 0, column 0: Data type 0xF3 (user-defined table type) is null and not set to default.  A null table-valued parameter is required to be sent as a default parameter.`

One possible approach which might allow iterator methods to work is [here](https://github.com/dotnet/SqlClient/compare/main...edwardneal:scratchpad/pr-3518). It's enough to prove the principle and hopefully it helps a little, but it's definitely not complete. For a quick overview:
* We add an explicit `Empty` flag to the structure which holds the peekahead information, indicating that the enumerator is empty.
* I took this slightly further and have a shared `EmptyPeekahead` instance so that we're not creating a large number of objects.
* `SqlParameter.GetActualFieldsAndProperties` is the only method which populates this peekahead structure. Rather than throwing `SQL.IEnumerableOfSqlDataRecordHasNoRows`, we set the peekahead to `EmptyPeekahead`.
* `TdsParser.WriteSmiParameter` is the only method which calls `GetActualFieldsAndProperties` and does anything with the peekahead information. It also has a `sendDefault` parameter, so we make sure that if the `Empty` flag on the peakahead is set, sendDefault is also forced to true. My test started to pass.
* Since we no longer used the `SQL.IEnumerableOfSqlDataRecordHasNoRows` method, I deleted it.
* I added the test in the gist to prove that this works. It passes on .NET Core.

I've not ported this over to the .NET Framework version. I've also only tested this in one specific situation - it's worth testing to prove this behaves properly with multiple table-valued parameters, empty lists, empty arrays and any other situations which come to mind. I'm sure there'll be some feedback on the way the test is written too - it's just to demonstrate the existing code which eliminates the need to create and drop databases manually.

sharafabacery:
@edwardneal  so impressive what you do,IF I need to still on this issue i need to do? what i need to learn? to complete this issue?

edwardneal:
Sorry for the delayed response. I think we're dealing with the problem in nearly the same place - the difference here is that your approach sets the `options` variable correctly in the first instance (based on the type of the parameter) while my approach changes the way that TdsParser responds to that variable a little later in the process (based on whether the IEnumerable has a value.)

I think the `EmptyDynamicIEnumerableShouldPassDefault` test will prove whether or not #2971 has been fixed (although benrr101 or mdaigle might have some extra feedback.) I'd suggest reverting the changes to SqlCommandTest.cs and using `EmptyDynamicIEnumerableShouldPassDefault` as your test case to prove whether or not the issue is fixed.

Ben's already pointed out a few of the unrelated changes to SqlCommand.cs, those need to be reverted. Once that's done (and the PR only contains the needed changes) then you can find a working method to fix the problem. After you've found that method, just make sure the two SqlCommand code files (one in netcore and one in netfx) have been changed in the same way.

sharafabacery:
@edwardneal thanks for help,curently I working on it by solution based on my previous one if your testcase fail i will try yours beacuse we try soolve the same problem and each of us look on it some respective, but whenI fetch orgin and run testcases and it build the package it responed with errors
Metadata file 'source\repos\SqlClient\artifacts\Project\bin\Windows_NT\Debug.AnyCPU\Microsoft.Data.SqlClient\netcore\net8.0\Microsoft.Data.SqlClient.dll' could not be found 
and there is some syntax error I showed

benrr101:
This PR doesn't compile in its current state, and it has merge conflicts with the current main branch. Please let us know what you want to do with it in the next week otherwise we'll probably close this PR.

sharafabacery:
@benrr101 close this issue the project in my machine doesn't compile also for that I will stop in this issue,I am sorry if I waste your time

paulmedynski:
Closing on author's request.

PR #3521, opened 2025-07-26
Update SqlVector<T>.Null refs

### LINKED ISSUE #3520: Method not found: Microsoft.Data.SqlTypes.SqlVector`1.get_Null()'
Using version 6.1.0:

```c#
// <PackageReference Include="Microsoft.Data.SqlClient" Version="6.1.0" />

using Microsoft.Data.SqlTypes;

Console.WriteLine(SqlVector<float>.Null.IsNull);
```

Results in:

```
Unhandled exception. System.MissingMethodException: Method not found: 'Microsoft.Data.SqlTypes.SqlVector`1<!0> Microsoft.Data.SqlTypes.SqlVector`1.get_Null()'.
   at Program.<Main>$(String[] args)
```

Is it because?...

Here, the return value is not defined as nullable:

https://github.com/dotnet/SqlClient/blob/b89ac491038bc9d0e90fdf7af45134465de6635a/src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs#L133-L134

While here, the return value is defined as nullable:

https://github.com/dotnet/SqlClient/blob/b89ac491038bc9d0e90fdf7af45134465de6635a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlTypes/SqlVector.cs#L107-L108

frankbuckley:
Updates refs for `SqlVector<T>.Null` to reflect it returns `Nullable<SqlVector<T>>`.

Should fix #3520


benrr101 (approved):
I think the major note is that this change is updating the ref definition to match the implementation definition.

PR #3524, opened 2025-07-28
Adjust TestSqlCommandCancel condition to reflect key store config requirement

David-Engel:
## Description

Adjusts the ConditionalTheory on TestSqlCommandCancel to only run if AE is configured with an appropriate key store for Windows and non-Windows. The Kerberos test pipeline doesn't use/configure AKV and due to recent changes to this test, it now fails in that pipeline on the Linux job (no AKV configured). This change skips the test on Linux if AKV isn't configured.


cheenamalhotra (approved):

David-Engel:
> We need to come up with a better strategy for how to do these conditional tests, but working within the constraints we have right now, this makes sense to me.

@benrr101 
Can you elaborate on what you think is wrong with the current strategy? The conditions are supposed to define test environment config requirements and skip the tests if the config is not satisfied.

In the case of this test, I could see from the history how it was "broken" by a recent change. Before the recent change, the condition excluded the test from .NET, which meant .NET Framework only, which meant Windows-only, which meant a the Windows Cert Store (any key store is the test requirement) was always available when the test ran. The recent change enabled the test for .NET, cross-platform. Most of our pipelines are set with most options configured to maximize coverage, which meant .NET and Linux jobs always had AKV configured. The Kerberos pipeline is more optimized to just those configs required for Kerberos testing, which resulted in this test failing because its condition wasn't specifically looking for a key store to be configured.

benrr101:
@David-Engel it's more a general purpose comment not specific to this test. One big problem we have currently is that checking conditions has to happen for every test, which means for tests that eg, require a connection to not be to synapse, it has to eg, connect and execute a query to the db to validate. If that fails, the test shows as failing, but it's incredibly unclear why. And of course all that overhead of hitting the server for each test slows down the tests.

One proposal @paulmedynski and I have discussed is having different connection string settings for the the different kinds of tests we want to run (eg, a synapse connection string, an ae connection string, a regular connection string), and operate under the assumption that if the string exists, it's connecting to the right kind of server for the test.

One other things we've been discussing is that skipping and categorizing tests by platform isn't ideal. The ideal would be to just not even compile them if they're not compatible with the platform, so wrapping them in #if for the platform (supported for .net platform but will need to wait until the common project for os platform) would be better.

Issue #1044, opened 2021-04-22
System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values.


moxixuan:
### Describe the bug
Hi, i hava a sql query the data.
I check my log and find some error in it，
it is not always happen, it happen something.
the data is about 3000 row, and 20 clomun return.

i add polly to retry.
and it may be success after retry

```
Exception message:
Stack trace:
```

System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values.    at Microsoft.Data.SqlClient.TdsParserStateObject.TryReadByteArray(Span`1 buff, Int32 len, Int32& totalRead)    at Microsoft.Data.SqlClient.TdsParserStateObject.TryReadUInt16(UInt16& value)    at Microsoft.Data.SqlClient.TdsParser.TryGetTokenLength(Byte token, TdsParserStateObject stateObj, Int32& tokenLength)    at Microsoft.Data.SqlClient.TdsParser.TrySkipValue(SqlMetaDataPriv md, Int32 columnOrdinal, TdsParserStateObject stateObj)    at Microsoft.Data.SqlClient.TdsParser.TrySkipRow(_SqlMetaDataSet columns, Int32 startCol, TdsParserStateObject stateObj)    at Microsoft.Data.SqlClient.SqlDataReader.TryCleanPartialRead()    at Microsoft.Data.SqlClient.SqlDataReader.TryCloseInternal(Boolean closeReader)    at Microsoft.Data.SqlClient.SqlDataReader.Close()    at Microsoft.Data.SqlClient.SqlDataReader.Dispose(Boolean disposing)    at Dapper.SqlMapper.QueryAsync[T](IDbConnection cnn, Type effectiveType, CommandDefinition command) in /_/Dapper/SqlMapper.Async.cs:line 456    at MeiYiJia.TakeOut.TaskProcessor.Repository.DapperExtensions.<>c__DisplayClass3_0`1.<<QueryAsyncWithRetry>b__0>d.MoveNext() in /src/src/MeiYiJia.TakeOut.TaskProcessor/Repository/DapperExtensions.cs:line 50 --- End of stack trace from previous location ---
--

```c#
var sql = $@"
               select  *
               from    tbGoods g
                       inner join tbPrice p (nolock)
                           on g.Gid = p.Gid
                       inner join tbGood sg (nolock)
                           on p.Gid = sg.Gid
                       inner join tbCategory cg (nolock)
[...]

PR #3534, opened 2025-07-31
Async multi packet fixes for 6.1.0

### LINKED ISSUE #3519: 6.1.0:  Errors while executing the query
### Describe the bug
After upgrading from 6.0.1 to 6.1.0, some of my queries started to fail with an error.
If you roll back to 6.0.1 the error disappears

Sometimes the request hangs for a long time (I stopped waiting after 3 minutes) and sometimes I get one of the following errors:
```
1:
Specified argument was out of the range of valid values., System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values.
   at Microsoft.Data.SqlClient.TdsParserStateObject.TryReadByteArray(Span`1 buff, Int32 len, Int32& totalRead, Int32 startOffset, Boolean writeDataSizeToSnapshot)
   at Microsoft.Data.SqlClient.TdsParserStateObject.TryReadByteArray(Span`1 buff, Int32 len)
   at Microsoft.Data.SqlClient.TdsParser.TryReadSqlValueInternal(SqlBuffer value, Byte tdsType, Int32 length, TdsParserStateObject stateObj)
   at Microsoft.Data.SqlClient.TdsParser.TryReadSqlValue(SqlBuffer value, SqlMetaDataPriv md, Int32 length, TdsParserStateObject stateObj, SqlCommandColumnEncryptionSetting columnEncryptionOverride, String columnName, SqlCommand command)
   at Microsoft.Data.SqlClient.SqlDataReader.TryReadColumnInternal(Int32 i, Boolean readHeaderOnly, Boolean forStreaming)
   at Microsoft.Data.SqlClient.SqlDataReader.TryReadColumn(Int32 i, Boolean setTimeout, Boolean allowPartiallyReadColumn, Boolean forStreaming)
   at Microsoft.Data.SqlClient.SqlDataReader.ReadAsyncExecute(Task task, Object state)
   at Microsoft.Data.SqlClient.SqlDataReader.ContinueAsyncCall[T](Task task, SqlDataReaderBaseAsyncCallContext`1 context)
   at Microsoft.Data.SqlClient.SqlDataReader.SqlDataReaderBaseAsyncCallContext`1.ExecuteAsyncCallCallback(Task task, Object state)
   at System.Threading.Tasks.ContinuationResultTaskFromResultTask`2.InnerInvoke()
   at System.Threading.ExecutionContext.RunFromThreadPoolDispatchLoop(Thread threadPoolThread, ExecutionContext executionContext, ContextCallback callback, Object state)   
   at System.Threading.ExecutionContext.RunFromThreadPoolDispatchLoop(Thread threadPoolThread, ExecutionContext executionContext, ContextCallback callback, Object state)   
   at System.Threading.Tasks.Task.ExecuteWithThreadLocal(Task& currentTaskSlot, Thread threadPoolThread)

2:
Offset and length were out of bounds for the array or count is greater than the number of elements from index to the end of the source collection., System.ArgumentException: Offset and length were out of bounds for the array or count is greater than the number of elements from index to the end of the source collection.
   at System.Buffer.BlockCopy(Array src, Int32 srcOffset, Array dst, Int32 dstOffset, Int32 count)
   at Microsoft.Data.SqlClient.TdsParserStateObject.TryReadPlpBytes(Byte[]& buff, Int32 offset, Int32 len, Int32& totalBytesRead, Boolean canContinue, Boolean writeDataSizeToSnapshot, Boolean compatibilityMode)
   at Microsoft.Data.SqlClient.TdsParserStateObject.TryReadPlpBytes(Byte[]& buff, Int32 offset, Int32 len, Int32& totalBytesRead)
   at Microsoft.Data.SqlClient.TdsParserStateObject.TryReadStringWithEncoding(Int32 length, Encoding encoding, Boolean isPlp, String& value)
   at Microsoft.Data.SqlClient.TdsParser.TryReadSqlStringValue(SqlBuffer value, Byte type, Int32 length, Encoding encoding, Boolean isPlp, TdsParserStateObject stateObj)
   at Microsoft.Data.SqlClient.TdsParser.TryReadSqlValue(SqlBuffer value, SqlMetaDataPriv md, Int32 length, TdsParserStateObject stateObj, SqlCommandColumnEncryptionSetting columnEncryptionOverride, String columnName, SqlCommand command)
   at Microsoft.Data.SqlClient.SqlDataReader.TryReadColumnInternal(Int32 i, Boolean readHeaderOnly, Boolean forStreaming)
   at Microsoft.Data.SqlClient.SqlDataReader.TryReadColumn(Int32 i, Boolean setTimeout, Boolean allowPartiallyReadColumn, Boolean forStreaming)
   at Microsoft.Data.SqlClient.SqlDataReader.ReadAsyncExecute(Task task, Object state)
   at Microsoft.Data.SqlClient.SqlDataReader.ContinueAsyncCall[T](Task task, SqlDataReaderBaseAsyncCallContext`1 context)
   at Microsoft.Data.SqlClient.SqlDataReader.SqlDataReaderBaseAsyncCallContext`1.ExecuteAsyncCallCallback(Task task, Object state)
   at System.Threading.Tasks.ContinuationResultTaskFromResultTask`2.InnerInvoke()
   at System.Threading.ExecutionContext.RunFromThreadPoolDispatchLoop(Thread threadPoolThread, ExecutionContext executionContext, ContextCallback callback, Object state)       
   at System.Threading.ExecutionContext.RunFromThreadPoolDispatchLoop(Thread threadPoolThread, ExecutionContext executionContext, ContextCallback callback, Object state)       
   at System.Threading.Tasks.Task.ExecuteWithThreadLocal(Task& currentTaskSlot, Thread threadPoolThread)


3:
Arithmetic operation resulted in an overflow., System.OverflowException: Arithmetic operation resulted in an overflow.
   at Microsoft.Data.SqlClient.TdsParserStateObject.TryReadPlpBytes(Byte[]& buff, Int32 offset, Int32 len, Int32& totalBytesRead, Boolean canContinue, Boolean writeDataSizeToSnapshot, Boolean compatibilityMode)
   at Microsoft.Data.SqlClient.TdsParserStateObject.TryReadPlpBytes(Byte[]& buff, Int32 offset, Int32 len, Int32& totalBytesRead)
   at Microsoft.Data.SqlClient.TdsParserStateObject.TryReadStringWithEncoding(Int32 length, Encoding encoding, Boolean isPlp, String& value)
   at Microsoft.Data.SqlClient.TdsParser.TryReadSqlStringValue(SqlBuffer value, Byte type, Int32 length, Encoding encoding, Boolean isPlp, TdsParserStateObject stateObj)
   at Microsoft.Data.SqlClient.TdsParser.TryReadSqlValue(SqlBuffer value, SqlMetaDataPriv md, Int32 length, TdsParserStateObject stateObj, SqlCommandColumnEncryptionSetting columnEncryptionOverride, String columnName, SqlCommand command)
   at Microsoft.Data.SqlClient.SqlDataReader.TryReadColumnInternal(Int32 i, Boolean readHeaderOnly, Boolean forStreaming)
   at Microsoft.Data.SqlClient.SqlDataReader.TryReadColumn(Int32 i, Boolean setTimeout, Boolean allowPartiallyReadColumn, Boolean forStreaming)
   at Microsoft.Data.SqlClient.SqlDataReader.ReadAsyncExecute(Task task, Object state)
   at Microsoft.Data.SqlClient.SqlDataReader.ContinueAsyncCall[T](Task task, SqlDataReaderBaseAsyncCallContext`1 context)
   at Microsoft.Data.SqlClient.SqlDataReader.SqlDataReaderBaseAsyncCallContext`1.ExecuteAsyncCallCallback(Task task, Object state)
   at System.Threading.Tasks.ContinuationResultTaskFromResultTask`2.InnerInvoke()
   at System.Threading.ExecutionContext.RunFromThreadPoolDispatchLoop(Thread threadPoolThread, ExecutionContext executionContext, ContextCallback callback, Object state)       
   at System.Threading.ExecutionContext.RunFromThreadPoolDispatchLoop(Thread threadPoolThread, ExecutionContext executionContext, ContextCallback callback, Object state)       
   at System.Threading.Tasks.Task.ExecuteWithThreadLocal(Task& currentTaskSlot, Thread threadPoolThread)

```

If you remove xml columns from the read query
```
Unexpected Unrecoverable Error., System.InvalidOperationException: Unexpected Unrecoverable Error.
   at Microsoft.Data.SqlClient.SqlDataReader.TryCloseInternal(Boolean closeReader)
   at Microsoft.Data.SqlClient.SqlDataReader.Close()
   at Microsoft.Data.SqlClient.SqlDataReader.Dispose(Boolean disposing)
```

### To reproduce
I created a small project on which I now consistently receive errors:

```c#

// I was able to reproduce the error only when connecting to a remote server with a login and password
string connectionString = "Data Source=remote-server; Initial Catalog=test_db; User Id= test_user; Password=password;Encrypt=False";
        
var dbTester = new DatabaseTester(connectionString);
        
try
{
    Console.WriteLine("Creating table...");
    dbTester.CreateTable();
            
    Console.WriteLine("Filling table...");
    dbTester.FillTable();
            
    Console.WriteLine($"Table have {dbTester.GetRecordCount()} records");
    
    Console.WriteLine("Getting records GetDataAsync...");
   var data2 = await dbTester.GetDataAsync();
   Console.WriteLine("Getting records GetDataAsync(int, int)...");
    await dbTester.GetDataAsync(100, 1000);
    
    foreach (var row in data2)
    {
        Console.WriteLine(row.ToString());
    }

    for (int i = 0; i < 100; i++)
    {
        Console.WriteLine($"TEST RUN {i}");
        var data22 = await dbTester.GetDataAsync();
        await dbTester.GetDataAsync(100, 1000);
    }
}
catch (Exception ex)
{
    Console.WriteLine($"Error: {ex.Message}, {ex.ToString()}");
}
finally
{
     dbTester.DropTable();
}


using System.Data;
using System.Text;
using System.Xml;

public class DatabaseTester
{
    private readonly string _connectionString;
    private const string TableName = "TestTable";

    public DatabaseTester(string connectionString)
    {
        _connectionString = connectionString;
    }
    
    public void CreateTable()
    {
        string createTableQuery = $@"
            IF OBJECT_ID('{TableName}', 'U') IS NOT NULL
                DROP TABLE {TableName};

            CREATE TABLE {TableName} (
                Id INT IDENTITY(1,1) PRIMARY KEY,
                Column1 VARCHAR(MAX),
                Column2 VARCHAR(MAX),
                Column3 VARCHAR(MAX),
                Column4 VARCHAR(MAX),
                Column5 VARCHAR(MAX),
                Column6 INT,
                Column7 DATETIME,
                Column8 DECIMAL(18,2),
                Column9 BIT,
                Column10 UNIQUEIDENTIFIER,
                Column11 XML,
                Column12 XML,
                Column13 BIGINT,
                Column14 SMALLINT,
                Column15 NVARCHAR(255)
            )";

        using (var connection = new SqlConnection(_connectionString))
        {
            connection.Open();
            using (var command = new SqlCommand(createTableQuery, connection))
            {
                command.ExecuteNonQuery();
            }
        }
    }
    
    public void FillTable()
    {
        using (var connection = new SqlConnection(_connectionString))
        {
            connection.Open();
            
            var random = new Random();
            var insertQuery = $@"
                INSERT INTO {TableName} (
                    Column1, Column2, Column3, Column4, Column5, 
                    Column6, Column7, Column8, Column9, Column10, 
                    Column11, Column12, Column13, Column14, Column15
                ) VALUES (
                    @Column1, @Column2, @Column3, @Column4, @Column5,
                    @Column6, @Column7, @Column8, @Column9, @Column10,
                    @Column11, @Column12, @Column13, @Column14, @Column15
                )";

            using (var command = new SqlCommand(insertQuery, connection))
            {
                command.Parameters.Add("@Column1", SqlDbType.VarChar, -1);
                command.Parameters.Add("@Column2", SqlDbType.VarChar, -1);
                command.Parameters.Add("@Column3", SqlDbType.VarChar, -1);
                command.Parameters.Add("@Column4", SqlDbType.VarChar, -1);
                command.Parameters.Add("@Column5", SqlDbType.VarChar, -1);
                command.Parameters.Add("@Column6", SqlDbType.Int);
                command.Parameters.Add("@Column7", SqlDbType.DateTime);
                command.Parameters.Add("@Column8", SqlDbType.Decimal);
                command.Parameters.Add("@Column9", SqlDbType.Bit);
                command.Parameters.Add("@Column10", SqlDbType.UniqueIdentifier);
                command.Parameters.Add("@Column11", SqlDbType.Xml);
                command.Parameters.Add("@Column12", SqlDbType.Xml);
                command.Parameters.Add("@Column13", SqlDbType.BigInt);
                command.Parameters.Add("@Column14", SqlDbType.SmallInt);
                command.Parameters.Add("@Column15", SqlDbType.NVarChar, 255);
                
                for (int i = 0; i < 5000; i++)
                {
                    command.Parameters["@Column1"].Value = GenerateRandomString(random, 1000);
                    command.Parameters["@Column2"].Value = GenerateRandomString(random, 1500);
                    command.Parameters["@Column3"].Value = GenerateRandomString(random, 2000);
                    command.Parameters["@Column4"].Value = GenerateRandomString(random, 500);
                    command.Parameters["@Column5"].Value = GenerateRandomString(random, 800);
                    command.Parameters["@Column6"].Value = random.Next(1, 1000000);
                    command.Parameters["@Column7"].Value = DateTime.Now.AddDays(-random.Next(1, 3650));
                    command.Parameters["@Column8"].Value = Math.Round((decimal)(random.NextDouble() * 10000), 2);
                    command.Parameters["@Column9"].Value = random.Next(0, 2) == 1;
                    command.Parameters["@Column10"].Value = Guid.NewGuid();
                    command.Parameters["@Column11"].Value = Random.Shared.Next(0, 3) == 2 ? GenerateRandomXml(random, 5) : DBNull.Value;
                    command.Parameters["@Column12"].Value = GenerateRandomXml(random, 3);
                    command.Parameters["@Column13"].Value = (long)random.Next(1, int.MaxValue) * 1000L;
                    command.Parameters["@Column14"].Value = (short)random.Next(1, 1000);
                    command.Parameters["@Column15"].Value = $"Record_{i}_{GenerateRandomString(random, 20)}";

                    command.ExecuteNonQuery();
                }
            }
        }
    }
    
    public async Task<List<TestTableRow>> GetDataAsync()
    {
        var result = new List<TestTableRow>();
        string selectQuery = $"SELECT * FROM {TableName} ORDER BY Id";
        
        using (var connection = new SqlConnection(_connectionString))
        {
            await connection.OpenAsync();
            using (var command = new SqlCommand(selectQuery, connection))
            {
                {
                    while (await reader.ReadAsync())
                    {
                        var row = new TestTableRow
                        {
                            Id = reader.GetInt32("Id"),
                            Column1 = reader.IsDBNull("Column1") ? null : reader.GetString("Column1"),
                            Column2 = reader.IsDBNull("Column2") ? null : reader.GetString("Column2"),
                            Column3 = reader.IsDBNull("Column3") ? null : reader.GetString("Column3"),
                            Column4 = reader.IsDBNull("Column4") ? null : reader.GetString("Column4"),
                            Column5 = reader.IsDBNull("Column5") ? null : reader.GetString("Column5"),
                            Column6 = reader.IsDBNull("Column6") ? 0 : reader.GetInt32("Column6"),
                            Column7 = reader.IsDBNull("Column7") ? DateTime.MinValue : reader.GetDateTime("Column7"),
                            Column8 = reader.IsDBNull("Column8") ? 0 : reader.GetDecimal("Column8"),
                            Column9 = reader.IsDBNull("Column9") ? false : reader.GetBoolean("Column9"),
                            Column10 = reader.IsDBNull("Column10") ? Guid.Empty : reader.GetGuid("Column10"),
                            Column11 = reader.IsDBNull("Column11") ? null : reader.GetString("Column11"),
                            Column12 = reader.IsDBNull("Column12") ? null : reader.GetString("Column12"),
                            Column13 = reader.IsDBNull("Column13") ? 0 : reader.GetInt64("Column13"),
                            Column14 = reader.IsDBNull("Column14") ? (short)0 : reader.GetInt16("Column14"),
                            Column15 = reader.IsDBNull("Column15") ? null : reader.GetString("Column15")
                        };
                        
                        result.Add(row);
                    }
                }
            }
        }

        return result;
    }
    
    public async Task<List<TestTableRow>> GetDataAsync(int pageNumber, int pageSize)
    {
        var result = new List<TestTableRow>();
        int offset = (pageNumber - 1) * pageSize;
        
        string selectQuery = $@"
            SELECT * FROM {TableName}
            ORDER BY Id
            OFFSET {offset} ROWS
            FETCH NEXT {pageSize} ROWS ONLY";

        using (var connection = new SqlConnection(_connectionString))
        {
            await connection.OpenAsync();
            using (var command = new SqlCommand(selectQuery, connection))
            {
                {
                    while (await reader.ReadAsync())
                    {
                        var row = new TestTableRow
                        {
                            Id = reader.GetInt32("Id"),
                            Column1 = reader.IsDBNull("Column1") ? null : reader.GetString("Column1"),
                            Column2 = reader.IsDBNull("Column2") ? null : reader.GetString("Column2"),
                            Column3 = reader.IsDBNull("Column3") ? null : reader.GetString("Column3"),
                            Column4 = reader.IsDBNull("Column4") ? null : reader.GetString("Column4"),
                            Column5 = reader.IsDBNull("Column5") ? null : reader.GetString("Column5"),
                            Column6 = reader.IsDBNull("Column6") ? 0 : reader.GetInt32("Column6"),
                            Column7 = reader.IsDBNull("Column7") ? DateTime.MinValue : reader.GetDateTime("Column7"),
                            Column8 = reader.IsDBNull("Column8") ? 0 : reader.GetDecimal("Column8"),
                            Column9 = reader.IsDBNull("Column9") ? false : reader.GetBoolean("Column9"),
                            Column10 = reader.IsDBNull("Column10") ? Guid.Empty : reader.GetGuid("Column10"), 
                            Column11 = reader.IsDBNull("Column11") ? null : reader.GetString("Column11"), 
                            Column12 = reader.IsDBNull("Column12") ? null : reader.GetString("Column12"),
                            Column13 = reader.IsDBNull("Column13") ? 0 : reader.GetInt64("Column13"),
                            Column14 = reader.IsDBNull("Column14") ? (short)0 : reader.GetInt16("Column14"),
                            Column15 = reader.IsDBNull("Column15") ? null : reader.GetString("Column15")
                        };
                        
                        result.Add(row);
                    }
                }
            }
        }

        return result;
    }

    public class TestTableRow
    {
        public int Id { get; set; }
        public string Column1 { get; set; }
        public string Column2 { get; set; }
        public string Column3 { get; set; }
        public string Column4 { get; set; }
        public string Column5 { get; set; }
        public int Column6 { get; set; }
        public DateTime Column7 { get; set; }
        public decimal Column8 { get; set; }
        public bool Column9 { get; set; }
        public Guid Column10 { get; set; }
        public string Column11 { get; set; } // XML
        public string Column12 { get; set; } // XML
        public long Column13 { get; set; }
        public short Column14 { get; set; }
        public string Column15 { get; set; }

        public override string ToString()
        {
            return $"Id: {Id}, Column15: {Column15}";
        }
    }
    
    public int GetRecordCount()
    {
        string countQuery = $"SELECT COUNT(*) FROM {TableName}";
        
        using (var connection = new SqlConnection(_connectionString))
        {
            connection.Open();
            using (var command = new SqlCommand(countQuery, connection))
            {
                return (int)command.ExecuteScalar();
            }
        }
    }
    
    public void DropTable()
    {
        string dropTableQuery = $"IF OBJECT_ID('{TableName}', 'U') IS NOT NULL DROP TABLE {TableName}";

        using (var connection = new SqlConnection(_connectionString))
        {
            connection.Open();
            using (var command = new SqlCommand(dropTableQuery, connection))
            {
                command.ExecuteNonQuery();
            }
        }
    }

    private DataTable CreateDataTable()
    {
        var dataTable = new DataTable();
        dataTable.Columns.Add("Column1", typeof(string));
        dataTable.Columns.Add("Column2", typeof(string));
        dataTable.Columns.Add("Column3", typeof(string));
        dataTable.Columns.Add("Column4", typeof(string));
        dataTable.Columns.Add("Column5", typeof(string));
        dataTable.Columns.Add("Column6", typeof(int));
        dataTable.Columns.Add("Column7", typeof(DateTime));
        dataTable.Columns.Add("Column8", typeof(decimal));
        dataTable.Columns.Add("Column9", typeof(bool));
        dataTable.Columns.Add("Column10", typeof(Guid));
        dataTable.Columns.Add("Column11", typeof(string));
        dataTable.Columns.Add("Column12", typeof(string));
        dataTable.Columns.Add("Column13", typeof(long));
        dataTable.Columns.Add("Column14", typeof(short));
        dataTable.Columns.Add("Column15", typeof(string));
        
        return dataTable;
    }

    private string GenerateRandomString(Random s, int length)
    {
        const string chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789 ";
        var random = new Random();
        var stringBuilder = new StringBuilder(length);
        
        for (int i = 0; i < length; i++)
        {
            stringBuilder.Append(chars[random.Next(chars.Length)]);
        }
        
        return stringBuilder.ToString();
    }

    private string GenerateRandomXml(Random s, int nodeCount)
    {
        var random = new Random();
        nodeCount = random.Next(10, 25);
        var xml = new XmlDocument();
        var root = xml.CreateElement("root");
        xml.AppendChild(root);

        for (int i = 0; i < nodeCount; i++)
        {
            var element = xml.CreateElement($"element_{i}");
            element.InnerText = GenerateRandomString(random, 100);
            root.AppendChild(element);
        }

        return xml.OuterXml;
    }
}
```

### Expected behavior
The query was completed without errors.

### Further technical details
Microsoft.Data.SqlClient version: 6.1.0
.NET target: net 9.0
Operating system: Windows 10 Pro
SqlServer: Microsoft SQL Server 2019 (RTM-CU15) (KB5008996) - 15.0.4198.2 (X64)

Wraith2:
Fixes #3519 

**Fix 1:**
When reading multi packet strings it is possible for multiple strings to happen in a single row. When reading asynchronously a snapshot is used which contains a linked list of packets. The current codebase has logic which keeps a cleared spare linked list node when the snapshot is cleared. The logic to clear the spare packet was faulty and did not clear all the fields leaving the data length in the node. In specific circumstances it is possible to re-use the spare linked list node containing an old data value as the first packet in a new linked list of packets. When this happens in a read which reaches the continue stage (3 or more packets) the size calculation is incorrect and various errors can occur.

The spare packet functionality is not very useful because it can store a single node. It doesn't retain the byte[] buffer so the memory saving is tiny. I have removed it and changed the linked list node fields to be readonly. This resolves the bug.

**Fix 2:**
When reading a multi packet string the plp chunks are read from each packet and the end is signalled by a terminator. It is possible for the data to align such that the contents of a string complete exactly at the end of a packet and the terminator is in the next packet. In this case some pre-existing logic checks for a 0 chars remaining and exists early. 

This logic needed to be updated so that in when continuing it returns the entire existing length read and not a 0 value. 

**Fix 3:**
While debugging the first two issues the buffer sizes and calculations were confusing me. I eventually realised that the code was directly using _longlenleft which is measured in bytes to size a char array, meaning that all char arrays were twice as long as needed. I have updated the code to handle that and use smaller appropriately sized arrays.

I have updated the existing test to iterate from 512 (minimum packet size) to 2048 bytes in size. This can cause lots of interesting alignments in the data testing the paths through the string reading code more effectively. The range could be increased but I considered that the runtime needed to be low enough to not timeout CI runs, most higher packet size will be similar to lower sized runs due to factoring. 

Thanks to @erenes and @Suchiman for their help finding the reproduction that worked on my machine, without that I would have been unable to fix anything

@dotnet/sqlclientdevteam can I get a CI run please.

/cc @jakimar

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
@@ -13206,7 +13206,7 @@ namespace Microsoft.Data.SqlClient
             if (stateObj._longlen == 0)
             {
                 Debug.Assert(stateObj._longlenleft == 0);
-                totalCharsRead = 0;
+                totalCharsRead = startOffsetByteCount >> 1;
                 return TdsOperationStatus.Done;       // No data
             }
 
@@ -13226,14 +13226,14 @@ namespace Microsoft.Data.SqlClient
             //  later needing to repeatedly allocate new target buffers and copy data as we discover new data
             if (buff == null && stateObj._longlen != TdsEnums.SQL_PLP_UNKNOWNLEN && stateObj._longlen < (int.MaxValue >> 1))
             {
-                if (supportRentedBuff && stateObj._longlen < 1073741824) // 1 Gib
+                if (supportRentedBuff && stateObj._longlen >> 1 < 1073741824) // 1 Gib
                 {
-                    buff = ArrayPool<char>.Shared.Rent((int)Math.Min((int)stateObj._longlen, len));
+                    buff = ArrayPool<char>.Shared.Rent((int)Math.Min((int)stateObj._longlen >> 1, len));
                     rentedBuff = true;
                 }
                 else
                 {
-                    buff = new char[(int)Math.Min((int)stateObj._longlen, len)];
+                    buff = new char[(int)Math.Min((int)stateObj._longlen >> 1, len)];
                     rentedBuff = false;
                 }
             }
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
@@ -3773,21 +3773,6 @@ namespace Microsoft.Data.SqlClient
                     return Math.Max(RunningDataSize - previous, 0);
                 }
 
-                internal void Clear()
-                {
-                    Buffer = null;
-                    Read = 0;
-                    NextPacket = null;
-                    if (PrevPacket != null)
-                    {
-                        PrevPacket.NextPacket = null;
-                        PrevPacket = null;
-                    }
-                    SetDebugStackImpl(null);
-                    SetDebugPacketId(0);
-                    SetDebugDataHash();
-                }
-
                 internal void SetDebugStack(string value) => SetDebugStackImpl(value);
                 internal void SetDebugPacketId(int value) => SetDebugPacketIdImpl(value);
                 internal void SetDebugDataHash() => SetDebugDataHashImpl();
@@ -4084,7 +4069,6 @@ namespace Microsoft.Data.SqlClient
             private PacketData _firstPacket;
             private PacketData _current;
             private PacketData _continuePacket;
-            private PacketData _sparePacket;
 
 #if DEBUG
             private int _packetCounter;
@@ -4165,15 +4149,8 @@ namespace Microsoft.Data.SqlClient
                     }
                 }
 #endif
-                PacketData packetData = _sparePacket;
-                if (packetData is null)
-                {
-                    packetData = new PacketData();
-                }
-                else
-                {
-                    _sparePacket = null;
-                }
+                PacketData packetData =  new PacketData();
+
                 packetData.Buffer = buffer;
                 packetData.Read = read;
 #if DEBUG
@@ -4357,13 +4334,10 @@ namespace Microsoft.Data.SqlClient
 
             private void ClearPackets()
             {
-                PacketData packet = _firstPacket;
                 _firstPacket = null;
                 _lastPacket = null;
                 _continuePacket = null;
                 _current = null;
-                packet.Clear();
-                _sparePacket = packet;
             }
 
             private void ClearState()
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
@@ -3741,8 +3741,9 @@ namespace Microsoft.Data.SqlClient
         {
             private sealed partial class PacketData
             {
-                public byte[] Buffer;
-                public int Read;
+                public readonly byte[] Buffer;
+                public readonly int Read;
+
                 public PacketData NextPacket;
                 public PacketData PrevPacket;
 
@@ -3751,6 +3752,12 @@ namespace Microsoft.Data.SqlClient
                 /// to get the offset of the previous packet data in the stored buffer
                 /// </summary>
                 public int RunningDataSize;
+
+                public PacketData(byte[] buffer, int read)
+                {
+                    Buffer = buffer;
+                    Read = read;
+                }
 
                 public int PacketID => Packet.GetIDFromHeader(Buffer.AsSpan(0, TdsEnums.HEADER_LEN));
 
@@ -4149,10 +4156,7 @@ namespace Microsoft.Data.SqlClient
                     }
                 }
 #endif
-                PacketData packetData =  new PacketData();
-
-                packetData.Buffer = buffer;
-                packetData.Read = read;
+                PacketData packetData = new PacketData(buffer, read);
 #if DEBUG
                 packetData.SetDebugStack(_stateObj._lastStack);
                 packetData.SetDebugPacketId(Interlocked.Increment(ref _packetCounter));
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataReaderTest/DataReaderTest.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataReaderTest/DataReaderTest.cs
@@ -1035,11 +1035,18 @@ INSERT INTO [{tableName}] (Data) VALUES (@data);";
                         }
                     }
 
-                    int counter = 0;
-                    while (counter < 10)
+                    SqlConnectionStringBuilder builder = new SqlConnectionStringBuilder(DataTestUtility.TCPConnectionString);
+                    builder.PersistSecurityInfo = true;
+                    builder.Pooling = false;
+                    
+                    for (int packetSize = 512; packetSize<2048; packetSize++)
                     {
-                        using (var cmd = cn.CreateCommand())
+                        builder.PacketSize = packetSize;
+                        using (SqlConnection sizedConnection = new SqlConnection(builder.ToString()))
+                        using (SqlCommand cmd = sizedConnection.CreateCommand())
                         {
+                            sizedConnection.Open();
+
                             cmd.CommandText = $"SELECT [d].[Id], [d].[DocumentIdentificationId], [d].[Name], [d].[Value] FROM [{tableName}] AS [d]";
                             using (var reader = await cmd.ExecuteReaderAsync())
                             {
@@ -1063,7 +1070,6 @@ INSERT INTO [{tableName}] (Data) VALUES (@data);";
                                 }
                             }
                         }
-                        counter++;
                     }
                 }
                 finally
--- a/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
+++ b/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
@@ -13396,7 +13396,7 @@ namespace Microsoft.Data.SqlClient
             if (stateObj._longlen == 0)
             {
                 Debug.Assert(stateObj._longlenleft == 0);
-                totalCharsRead = 0;
+                totalCharsRead = startOffsetByteCount >> 1;
                 return TdsOperationStatus.Done;       // No data
             }
 
@@ -13416,14 +13416,14 @@ namespace Microsoft.Data.SqlClient
             // allocate the whole buffer in one shot instead of realloc'ing and copying over each time
             if (buff == null && stateObj._longlen != TdsEnums.SQL_PLP_UNKNOWNLEN && stateObj._longlen < (int.MaxValue >> 1))
             {
-                if (supportRentedBuff && stateObj._longlen < 1073741824) // 1 Gib
+                if (supportRentedBuff && stateObj._longlen >> 1 < 1073741824) // 1 Gib
                 {
-                    buff = ArrayPool<char>.Shared.Rent((int)Math.Min((int)stateObj._longlen, len));
+                    buff = ArrayPool<char>.Shared.Rent((int)Math.Min((int)stateObj._longlen >> 1, len));
                     rentedBuff = true;
                 }
                 else
                 {
-                    buff = new char[(int)Math.Min((int)stateObj._longlen, len)];
+                    buff = new char[(int)Math.Min((int)stateObj._longlen >> 1, len)];
                     rentedBuff = false;
                 }
             }


Wraith2:
@ErikEJ this might reduce memory usage for string reads. It might be worth benching the artifacts if the CI runs green.

Wraith2:
I've added an additional fix which is the same as the 0 length left in terminator case and which occurs on the varchar not nvarchar read path.

Wraith2:
@dotnet/sqlclientdevteam can I get a CI run on this please. 

I've added a new commit which forces process sni mode to compatibility mode (and by extension, disabled async-continue mode) and adds in a fix for the pending read counter imbalance that we discussed and that @rhuijben has been assisting with tracking down today. This is a possible stable current codebase state to evaluate.

Wraith2:
I've aligned the appcontext switch test with the new defaults. Can i get another run please @dotnet/sqlclientdevteam

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -13206,7 +13206,7 @@ bool writeDataSizeToSnapshot
             if (stateObj._longlen == 0)
             {
                 Debug.Assert(stateObj._longlenleft == 0);
-                totalCharsRead = 0;
+                totalCharsRead = startOffsetByteCount >> 1;
```
paulmedynski: Is this a division by 2 in disguise?  Are you using a special property of right-bit-shift that divide-by-2 doesn't have?  Something else?

If the former, please use `startOffsetByteCount / 2` for clarity.  If the either of the latter, please document why.

paulmedynski (changes requested):
Asking for some clarity on `>> 1` versus `/ 2`.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
Wraith2: No magic. Just using the same idiom as the containing methods. I've changed it to use division instead of shift.

I've also changed the multiplexer test detection of compatibility to match the library which should skip the multiplexer tests correctly now.

rhuijben:
@Wraith2 when I run the testcase from the other issue against this branch I get in DEBUG mode
```
 SearchDogCrash.SearchDogCrashTests.TestSearchDogCrash
   Source: Class1.cs line 11
   Duration: 947 ms

  Message: 
Microsoft.VisualStudio.TestPlatform.TestHost.DebugAssertException : Method Debug.Fail failed with 'Invalid token after performing CleanPartialRead: 04
', and was translated to Microsoft.VisualStudio.TestPlatform.TestHost.DebugAssertException to avoid terminating the process hosting the test.

  Stack Trace: 
SqlDataReader.TryCleanPartialRead() line 867
SqlDataReader.TryCloseInternal(Boolean closeReader) line 1058
SqlDataReader.Close() line 1009
SqlDataReader.Dispose(Boolean disposing) line 924
DbDataReader.DisposeAsync()
SearchDogCrashTests.Do() line 49
ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
AsyncStateMachineBox`1.MoveNext(Thread threadPoolThread)
AwaitTaskContinuation.RunOrScheduleAction(IAsyncStateMachineBox box, Boolean allowInlining)
Task.RunContinuations(Object continuationObject)
Task`1.TrySetResult(TResult result)
UnwrapPromise`1.TrySetFromTask(Task task, Boolean lookForOce)
UnwrapPromise`1.ProcessInnerTask(Task task)
Task.RunContinuations(Object continuationObject)
Task.ExecuteWithThreadLocal(Task& currentTaskSlot, Thread threadPoolThread)
ThreadPoolWorkQueue.Dispatch()
WorkerThread.WorkerThreadStart()
```

In release mode the test passes.
(Reverted to old version of the library for RepoDB and $dayjob)

Wraith2:
multitasking here, can you link me to the exact repro you're talking about?

rhuijben:
> multitasking here, can you link me to the exact repro you're talking about?

The testcase from
https://github.com/dotnet/SqlClient/issues/3519#issuecomment-3178329931

I'm currently trying to get things reproduced against a docker instance of sqlserver 2019 so we can look at the same thing (and maybe even test this on github actions, like I do in the RepoDB project)

rhuijben:
This case  https://github.com/rhuijben/SqlClient/commit/1964bc1c94e4865214b84b3913ffb66fa577b955
(Extracted from https://github.com/dotnet/SqlClient/issues/3519#issuecomment-3136947692)

fails for me on this docker setup.
(If you have docker, running `docker compose up -d` in the directory with the compose script will give you a local sqlserver instance. The testcase then adds the schema and runs)

Too bad it is not the error I'm seeing myself, but it is still a valid testcase. Trying to extend this to include my case.

It fails on the first (smallest) packetsize of 512.
```
 SearchDogCrash.SearchDogCrashTests.OtherRepro
   Source: CrashTest.cs line 822
   Duration: 3,6 min

  Message: 
Microsoft.VisualStudio.TestPlatform.TestHost.DebugAssertException : Method Debug.Fail failed with 'partially read packets cannot be appended to the snapshot
', and was translated to Microsoft.VisualStudio.TestPlatform.TestHost.DebugAssertException to avoid terminating the process hosting the test.

  Stack Trace: 
StateSnapshot.AppendPacketData(Byte[] buffer, Int32 read) line 4158
TdsParserStateObject.ProcessSniPacketCompat(PacketHandle packet, UInt32 error) line 529
TdsParserStateObject.ProcessSniPacket(PacketHandle packet, UInt32 error) line 19
TdsParserStateObject.ReadAsyncCallback(IntPtr key, PacketHandle packet, UInt32 error) line 353
TdsParserStateObject.ReadSni(TaskCompletionSource`1 completion) line 3236
TdsParserStateObject.TryReadNetworkPacket() line 2818
TdsParserStateObject.TryPrepareBuffer() line 1299
TdsParserStateObject.TryReadByteArray(Span`1 buff, Int32 len, Int32& totalRead, Int32 startOffset, Boolean writeDataSizeToSnapshot) line 1492
TdsParserStateObject.TryReadByteArray(Span`1 buff, Int32 len, Int32& totalRead) line 1453
TdsParserStateObject.TryReadInt64(Int64& value) line 1721
<13 more frames...>
AwaitTaskContinuation.RunOrScheduleAction(Action action, Boolean allowInlining)
Task.RunContinuations(Object continuationObject)
Task`1.TrySetResult(TResult result)
TaskCompletionSource`1.TrySetResult(TResult result)
SqlDataReader.CompleteAsyncCall[T](Task`1 task, SqlDataReaderBaseAsyncCallContext`1 context) line 6123
SqlDataReaderBaseAsyncCallContext`1.CompleteAsyncCallCallback(Task`1 task, Object state) line 5822
ExecutionContext.RunFromThreadPoolDispatchLoop(Thread threadPoolThread, ExecutionContext executionContext, ContextCallback callback, Object state)
Task.ExecuteWithThreadLocal(Task& currentTaskSlot, Thread threadPoolThread)
ThreadPoolWorkQueue.Dispatch()
WorkerThread.WorkerThreadStart()

```

rhuijben:
```csharp
Debug.Assert(TdsEnums.HEADER_LEN + Packet.GetDataLengthFromHeader(buffer) == read, "partially read packets cannot be appended to the snapshot");
```
read=512
buffer = byte[512], (first and last byte are 0x04)
TdsEnums.HEADER_LEN = 8.
Packet.GetDataLengthFromHeader(buffer) returns 503

503+8 = 511, so mismatch.

Looks like the first byte of the next package is already in the buffer here.

Wraith2:
That assert will fire periodically when packet multiplexing is disabled. We should add in the context switch to the assertion.

That might be correct. I saw something similar while looking at the multipart xml reads with a weird packet size. If the packet status does not include the last packet bit, and the requiredlength is less than the total packet as long as the transferred data amount is the same as the buffer size it's technically correct, I think. I'm referring to this as padded packets. I hadn't seen them before 2 weeks ago but the spec doesn't preclude them. When i saw them the remaining data space in the packet buffer was filled with FF. This is part of the reason that i added the DumpPackets and DumpInBuff functions to my debug branch.

rhuijben:
With packet size configured as 512 I see 511 byte packets (which fail these tests), but also one really large packet (>= 60 KB). Not sure if the debug assert does the right thing. It looks like the demultiplexer handles these cases just fine.

With this packet code you also always have to handle short-reads caused by network security and TCP packets. There are standard proxies for that last case so you can always get small (or large) packets from the network layer. The DotNet core project uses fuzzing with that to catch http errors, as do a lot of other libraries.

Looks like these asserts are on the wrong layer... as from the network you can have much smaller or larger packets than the TDS packets (smaller when processing really fast, and much longer when the network already delivered more data than a single packet... Which can also happen on slow networks when one packet got lost and is re-delivered, while others are already in the queue.

Wraith2:
With the multiplexer active every packet which is appended to the snapshot is required to have a correct header and be a complete packet. If you look at the locations where packets are assigned to _inBuff from the snapshot you'll see that it gets the status and data length from the packet, that requires a correct packet header.

With the multiplexer inactive, which is ProcessSni compatibility mode those requirements no longer hold. So the correct things to do is to change the asserts to be:
```
                Debug.Assert(buffer != null, "packet data cannot be null");
                Debug.Assert(LocalAppContextSwitches.UseCompatibilityProcessSni || read >= TdsEnums.HEADER_LEN, "minimum packet length is TdsEnums.HEADER_LEN");
                Debug.Assert(LocalAppContextSwitches.UseCompatibilityProcessSni || TdsEnums.HEADER_LEN + Packet.GetDataLengthFromHeader(buffer) == read, "partially read packets cannot be appended to the snapshot");

```

In terms of packet lengths I refer to the TDS definition of a packet which is that all packets are of the communicated packetsize unless they are the last packet in a stream. Even though it's possible in network terms to receive any length of bytes up to the buffer size that was passed to the function we will generally find that there are full packets unless your server is quite slow. Unfortunately the slow case is very hard to replicate in a reliable way so there have been holes in the code. 

The demultiplexer current does not handle padded packets (physical length > logical length) correctly. I need to get some example packets with padding and then write some new test infra for the multiplexer to make sure that CurrentLength and RequiredLength are used correctly.

rhuijben:
On https://github.com/rhuijben/SqlClient/tree/test/repro-1 (your branch + 1 commit) I now have (with `AppContext.SetSwitch("Switch.Microsoft.Data.SqlClient.UseCompatibilityProcessSni", false);`)

Running the tests against a docker with a clean sqlserver 2019 container. (Create script is in that commit)

```
 SearchDogCrash.SearchDogCrashTests.OtherRepro
   Source: CrashTest.cs line 824
   Duration: 40,4 sec

  Message: 
    Microsoft.VisualStudio.TestPlatform.TestHost.DebugAssertException : Method Debug.Fail failed with 'dumping buffer
    _inBytesRead = 21
    _inBytesUsed = 11
    used buffer:
    04 01 00 15 00 34 01 00 FD 20 00 
    unused buffer:
    FD 00 00 00 00 00 00 00 00 00 
    
    ', and was translated to Microsoft.VisualStudio.TestPlatform.TestHost.DebugAssertException to avoid terminating the process hosting the test.

  Stack Trace: 
    TdsParser.TryRun(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj, Boolean& dataReady) line 2043
    TdsParser.Run(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj) line 1923
    TdsParser.ProcessAttention(TdsParserStateObject stateObj) line 8178
    TdsParser.ProcessPendingAck(TdsParserStateObject stateObj) line 344
    TdsParserStateObject.ResetCancelAndProcessAttention() line 839
    TdsParserStateObject.CloseSession() line 919
    SqlDataReader.TryCloseInternal(Boolean closeReader) line 1078
    SqlDataReader.Close() line 1009
    SqlDataReader.Dispose(Boolean disposing) line 924
    DbDataReader.DisposeAsync()
    SearchDogCrashTests.OtherRepro() line 844
    ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
    AwaitTaskContinuation.RunOrScheduleAction(IAsyncStateMachineBox box, Boolean allowInlining)
    Task.RunContinuations(Object continuationObject)
    Task.FinishSlow(Boolean userDelegateExecute)
    Task.TrySetException(Object exceptionObject)
    TaskCompletionSource`1.TrySetException(Exception exception)
    SqlDataReader.CompleteAsyncCall[T](Task`1 task, SqlDataReaderBaseAsyncCallContext`1 context) line 6115
    SqlDataReaderBaseAsyncCallContext`1.CompleteAsyncCallCallback(Task`1 task, Object state) line 5822
    ExecutionContext.RunFromThreadPoolDispatchLoop(Thread threadPoolThread, ExecutionContext executionContext, ContextCallback callback, Object state)
    Task.ExecuteWithThreadLocal(Task& currentTaskSlot, Thread threadPoolThread)
    ThreadPoolWorkQueue.Dispatch()
    WorkerThread.WorkerThreadStart()
```
(Don't look at the timings. My laptop is in silent mode, which is very slow with todays weather)

Wraith2:
Token packet, last packet of stream. No idea what the content is supposed to represent. This would silently fail in release mode because all that diagnostic output is a #DEBUG block

rhuijben:
> 04 01 00 15 00 34 01 00 FD 20 00

Not sure if I read this correctly but looks like:
Packet of 0x15 length (byte 2 and 3 in the array), so 21 decimal. This happens to be the number of bytes available. But only 11 are actually marked as read...

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataReaderTest/DataReaderTest.cs
```diff
@@ -1035,11 +1035,18 @@ [Value] [nvarchar](max) NULL
                         }
                     }
 
-                    int counter = 0;
-                    while (counter < 10)
+                    SqlConnectionStringBuilder builder = new SqlConnectionStringBuilder(DataTestUtility.TCPConnectionString);
+                    builder.PersistSecurityInfo = true;
+                    builder.Pooling = false;
```
rhuijben, line 1040: I hade the most luck reproducing issues by using pooling for just 2-3 times. Disabling pooling hides some backend errors as the connection is closed anyway.

Wraith2:
> > 04 01 00 15 00 34 01 00 FD 20 00
> 
> Not sure if I read this correctly but looks like: Packet of 0x15 length (byte 2 and 3 in the array), so 21 decimal. This happens to be the number of bytes available. But only 11 are actually marked as read...

Yes. That was my interpretation. Since I've no idea what the bytes that have been read were a part of I have no way to tell if the extra bytes are meaningful or not. It seems likely that they should be. If you've got my additional logging enabled from the -dev branch then you could look at the output of DumpLog() and see what operation consumed those bytes.

Wraith2:
I've pushed a bunch of new fixes. Can I get a CI run @dotnet/sqlclientdevteam  and if that builds some testing by brave people who have reproduced known issues please?

Wraith2:
All green apart from a timeout. @rhuijben if you get opportunity could you have a try with this branch or PR artifacts and see if you can repro any problems?

rhuijben:
I'm away from my PC for a few days. Will follow up when I get back.

ErikEJ:
I will test the artifact with my repro in the coming week

ErikEJ:
@Wraith2 My repro code using the SQL 2019 VM that I gave you acces to with build `6.11.0-pull.122707`  and
`AppContext.SetSwitch("Switch.Microsoft.Data.SqlClient.UseCompatibilityProcessSni", false);`
throws:

```
Unhandled exception. System.Xml.XmlException: Unexpected end tag. Line 1, position 39.
   at System.Xml.XmlTextReaderImpl.Throw(Exception e)
   at System.Xml.XmlTextReaderImpl.Throw(String res, String arg)
   at System.Xml.XmlTextReaderImpl.Throw(Int32 pos, String res)
   at System.Xml.XmlTextReaderImpl.ParseDocumentContent()
   at System.Xml.XmlWriter.WriteNode(XmlReader reader, Boolean defattr)
   at System.Data.SqlTypes.SqlXml.get_Value()
   at Microsoft.Data.SqlClient.SqlCachedBuffer.ToString()
   at Microsoft.Data.SqlClient.SqlBuffer.get_String()
   at Microsoft.Data.SqlClient.SqlDataReader.GetString(Int32 i)
   at DatabaseTester.GetAndCompareDataAsync(List`1 originalData) in C:\Users\ErikEjlskovJensen(De\source\repos\SqlClientB
```

Wraith2:
Odd. I can't connect to your server at all, could you check and see if the IP address changed?

to confirm, are these the settings you're seeing errors with?
```
bool compat = false;
bool managed = false;
bool useContinue = false;
AppContext.SetSwitch("Switch.Microsoft.Data.SqlClient.UseManagedNetworkingOnWindows", managed);
AppContext.SetSwitch("Switch.Microsoft.Data.SqlClient.UseCompatibilityAsyncBehaviour", !useContinue);
AppContext.SetSwitch("Switch.Microsoft.Data.SqlClient.UseCompatibilityProcessSni", compat);
```

ErikEJ:
@Wraith2 I have just turned it back on 😄 

`AppContext.SetSwitch("Switch.Microsoft.Data.SqlClient.UseCompatibilityProcessSni", false);` is the only swith I have in my code. Not sure how it relates to all the others. And I am running the test on Windows 11 PC (so using unmanaged networking)

Wraith2:
Ok. Well I've run this branch against my server in sweden which is similar to yours for a while and no replication. Now i'm running directly against yours again and I'll let it go for a while but I'm not seeing any problems occurring.

ErikEJ:
@Wraith2 This is my code, and the issue repros right away!


```c#

AppContext.SetSwitch("Switch.Microsoft.Data.SqlClient.UseManagedNetworkingOnWindows", false);
AppContext.SetSwitch("Switch.Microsoft.Data.SqlClient.UseCompatibilityAsyncBehaviour", false);
AppContext.SetSwitch("Switch.Microsoft.Data.SqlClient.UseCompatibilityProcessSni", false);

var connectionString = "Data Source=20.x.x.x,1433;Initial Catalog=TestDB;User Id=wraith2;Password=x;Encrypt=False;Trust Server Certificate=False;Command Timeout=30";

var dbTester = new DatabaseTester(connectionString);
        
var connectionsStringBuilder = new SqlConnectionStringBuilder(connectionString);
connectionsStringBuilder.Encrypt = true;
connectionsStringBuilder.TrustServerCertificate = true;
    
var dataSetTester = new DatabaseTester(connectionsStringBuilder.ConnectionString);
var originalRecords = await dataSetTester.GetAndCompareDataAsync(null);
while(true)
{
    Console.WriteLine("RUNNING");
    await dbTester.GetAndCompareDataAsync(originalRecords);
}

```

ErikEJ:
@Wraith2 If I change above to this, the issue goes away

`AppContext.SetSwitch("Switch.Microsoft.Data.SqlClient.UseCompatibilityAsyncBehaviour", true);`

Wraith2:
I still can't replicate. I'm a few commits ahead locally though so It's possible i've fixed it. I didn't push over the weekend so that the artifact would be available for people to test but since we're seeing a problem i've pushed now. @dotnet/sqlclientdevteam can you run CI please.

Wraith2:
new artifacts are her https://sqlclientdrivers.visualstudio.com/904996cc-6198-4d39-8540-eca72bdf0b7b/_apis/build/builds/123164/artifacts?artifactName=Artifacts&api-version=7.1&%24format=zip if you could try them please.

ErikEJ:
@Wraith2 When using `6.11.0-pull.123002` my repro no longer crashes! 🎉

Wraith2:
That's a relief. Thanks.

Edit: Though, reviewing the commits since that build I'm suspicious because none of them went near xml. I suspect that the problem was around continue mode with xml where I know there is a bug (I have a fix on my dev branch) so adding in the RequestContinue functionality is causing it to take the working vs broken path.

edwardneal:
I've been running my own test cases against the branch prior to the most recent changes, the results are available [here](https://gist.github.com/edwardneal/8d479971cb145dbe75a309ec75c151f9). These covered most of the combinations I can think of:
* Servers: Local SQL 2019 instance, @Wraith2's server in Sweden, and two Azure SQL instances (UKSouth and AustraliaEast)
* Read types: read a blob as a byte array or as a stream; read a varchar as a string or as a TextReader, read an nvarchar as a string or as a TextReader
* Data sizes: 1MB, 5MB, 25MB
* Packet sizes: 512, 567, 8000
* SqlDataReader behaviour: Default and SequentialAccess
* Parallel threads: 1, 2, 8
* Type: sync and async
* AppContext switches: all at default

I wasn't able to reproduce any exceptions which were unique to this PR; I'll kick them off against the latest version later this evening.

ErikEJ:
@edwardneal 

> AppContext switches: all at default

I am a little confused about the current state of the switches - does "all at deafult" enable all the new async multi  packet features?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/LocalAppContextSwitches.cs
```diff
@@ -78,13 +78,13 @@ public static bool UseCompatibilityProcessSni
...
-                    if (AppContext.TryGetSwitch(UseCompatibilityProcessSniString, out bool returnedValue) && returnedValue)
+                    if (AppContext.TryGetSwitch(UseCompatibilityProcessSniString, out bool returnedValue) && !returnedValue)
                     {
-                        s_useCompatibilityProcessSni = Tristate.True;
+                        s_useCompatibilityProcessSni = Tristate.False;
                     }
                     else
                     {
-                        s_useCompatibilityProcessSni = Tristate.False;
+                        s_useCompatibilityProcessSni = Tristate.True;
```
Wraith2: @ErikEJ in this PR the default is changed to that UseCompatibilityProcessSni is true by default. That means that new async behaviour is false by default.

ErikEJ: Thanks, so in a repro context, we should use:

```c#
AppContext.SetSwitch("Switch.Microsoft.Data.SqlClient.UseCompatibilityAsyncBehaviour", false);
AppContext.SetSwitch("Switch.Microsoft.Data.SqlClient.UseCompatibilityProcessSni", false);

```

Wraith2: For public consumption the settings you give are the defaults. 

If you can find an issue I want to know about it so I can fix it, regardless of the settings but you will need to tell me what the settings are so I can try to repro.

ErikEJ: Got it. No issues found with latest build.

MichelZ:
No issues for my repro with the latest build of this PR where it previously failed almost immediately. Will let it run for a while longer, using:

```
AppContext.SetSwitch("Switch.Microsoft.Data.SqlClient.UseCompatibilityAsyncBehaviour", false);
AppContext.SetSwitch("Switch.Microsoft.Data.SqlClient.UseCompatibilityProcessSni", false);
```

Wraith2:
@dotnet/sqlclientdevteam what do I need to move this forwards? Based on external testing you can see this seems to solve all the issues that have been reported. All the CI failures I've looked at are timeouts.

ErikEJ:
One option could be to make this opt-in in 7.0 and then opt-out in 7.1?

Wraith2:
Yes. I assume that's the path forward. To do that we have to get this merged. 
I also have the reworked continue mode with better performance in a branch based on this one and that would be good to get in too.

Wraith2:
> @dotnet/sqlclientdevteam what do I need to move this forwards? Based on external testing you can see this seems to solve all the issues that have been reported. All the CI failures I've looked at are timeouts.

Hello? @dotnet/sqlclientdevteam ? If something is needed now is the time to tell me so I can work on it over a weekend.

paulmedynski:
Hi @Wraith2 - I'll take a look when I'm back at work on Tuesday.  I'll encourage others to have a look as they finish their workday today.

Can you resolve the conflicts?

Wraith2:
Ok. rebase is done. I've done it because it's the stress tests and that's a valid reason to do so. It's probably worth running the CI.

In general I have said in the past that I'm not going to keep spending time rebasing my changes to keep up with main unless there is some review feedback reason to do so. I've wasted far too much time in the past keeping my PR's up to date when no-one is looking at it.

edwardneal:
I've built the latest version of the PR and re-ran my stress tests under similar constraints, results are [here](https://gist.github.com/edwardneal/4cb73ea1e19f945ac3e94b8aafd525aa) as before.

All stress tests pass with varying sizes of data (1/5/25MB) to servers with varying latency. This confirms the fix for #3572, and (as expected) the same fix also fixed a similar type of issue with TextReader.

Since 3572 also appears in earlier versions, we should probably backport bd780e4f37a6eec24b63e884f8c7c59e71cd6a01 to 6.1/6.0/5.1. I can handle that quickly enough if the SqlClient team agree.

mdaigle:
I would like to take these changes in 7.0.0 with the feature disabled by default. I will advocate for this approach in the team, but I do need help from you all to make that possible. Our primary concern with these changes is reliability. We want to be certain (or as near as possible) that continuing with these changes in 7.0.0 will not impact reliability for customers who use the driver with the feature disabled (UseCompatibilityAsyncBehaviour and UseCompatibilityProcessSni set to true).

I've seen a lot of testing focus on the issues identified in 6.1.0 which is awesome. It looks like there is consensus that all identified issues have been addressed with this change set. What I'd like to see is additional verification that, after taking this change set, the driver will perform identically to 6.1.1 when the feature is disabled.

I'd like to see the following boxes checked before we proceed:
- [ ] Stress tests
  - the output between 6.1.1 and this+feature_disabled should be the same. 
- [ ] EFcore tests
  - the output between 6.1.1 and this+feature_disabled  should be the same.
- [ ] Identified issues from 6.1.0
  - The identified issues from 6.1.0 do not reproduce on this+feature_disabled .
  - Please run any personal repros you have and ensure they don't reproduce with the feature disabled

If you have additional ideas for ways to verify reliability, please let me know.

Wraith2:
That's good news. 

For testing if you look through the tests that @edwardneal posted https://github.com/dotnet/SqlClient/pull/3534#issuecomment-3218237906 they're very thorough. They covered the existing scenarios and found some new problems as well. 

For the larger question of how do we get coverage of things that we don't know about I don't have any quick answers. My only idea is to ask for help from the community as I have done before. To do that we need a preview package that they can easily try. If we ask people to be creative and try to find edge cases where they think it might struggle, low resources, slow connections etc, then that might show up some more weird edge cases.

ErikEJ:
I can look at checkbox 2 and 3.

ErikEJ:
@mdaigle EFcore tests - the output between 6.1.1 and this+feature_disabled should be the same. - **Confirmed** with `6.11.0-pull.123707`

ErikEJ:
@mdaigle Identified issues from 6.1.0 - The identified issues from 6.1.0 do not reproduce on this+feature_disabled .
Please run any personal repros you have and ensure they don't reproduce with the feature disabled - **Confirmed** with `6.11.0-pull.123707`

paulmedynski:
This PR needs a merge from main to pickup macOS test failure fixes.

Wraith2:
Sure, but that doesn't block review of the code changes. 

As I mentioned in https://github.com/dotnet/SqlClient/pull/3534#issuecomment-3215944991 I'm happy to rebase when there is clear code reason to do so but failing test infra does not affect the code changes that I've made. If The changes pass review then I'll rebase in readiness of a merge.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -13245,14 +13247,14 @@ bool writeDataSizeToSnapshot
             //  later needing to repeatedly allocate new target buffers and copy data as we discover new data
             if (buff == null && stateObj._longlen != TdsEnums.SQL_PLP_UNKNOWNLEN && stateObj._longlen < (int.MaxValue >> 1))
             {
-                if (supportRentedBuff && stateObj._longlen < 1073741824) // 1 Gib
+                if (supportRentedBuff && stateObj._longlen >> 1 < 1073741824) // 1 Gib
```
paulmedynski: I think it's worth making a temporary `stateLength` rather than doing the division 3 times.

Same in the netfx file.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.Multiplexer.cs
```diff
@@ -141,19 +148,28 @@ out recurse
                             // we don't process it
                             if (!bufferIsPartialCompleted)
                             {
-                                SetBuffer(_inBuff, 0, 0);
```
paulmedynski, line 144: Should this stay SetBuffer()?  Do we need to explicitly discard the old buffer even though we're seeking back to its beginning?

mdaigle:
> I've built the latest version of the PR and re-ran my stress tests under similar constraints, results are [here](https://gist.github.com/edwardneal/4cb73ea1e19f945ac3e94b8aafd525aa) as before.
> 
> All stress tests pass with varying sizes of data (1/5/25MB) to servers with varying latency. This confirms the fix for #3572, and (as expected) the same fix also fixed a similar type of issue with TextReader.
> 
> Since 3572 also appears in earlier versions, we should probably backport [bd780e4](https://github.com/dotnet/SqlClient/commit/bd780e4f37a6eec24b63e884f8c7c59e71cd6a01) to 6.1/6.0/5.1. I can handle that quickly enough if the SqlClient team agree.

@edwardneal thanks for your work on this. Did you have a chance to run your stress tests on this+feature_disabled?
And I may have missed it, but have you shared the source code for these tests anywhere? I'd like to look through them to see the types of scenarios that triggered the errors.
We're looking to strengthen our stress testing suite and the tests you've built may be a good addition. See #3558 for where those are living.

mdaigle:
> @mdaigle EFcore tests - the output between 6.1.1 and this+feature_disabled should be the same. - **Confirmed** with `6.11.0-pull.123707`

Thank you @ErikEJ !

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
```diff
@@ -3204,7 +3230,13 @@ internal void ReadSni(TaskCompletionSource<object> completion)
...
                         }
                         else
-                        {
+                        {   
+                            // this call to IncrementPendingCallbacks is required for balance
+                            // the _pendingCallbacks counter will be unconditionally decremented in ReadAsyncCallback
+                            //  so we must make sure that even though we are not making a network call that we do
+                            //  not cause an incorrect decrement which will cause disconnection from the native
+                            //  component
+                            IncrementPendingCallbacks();
```
mdaigle: Note to self, readFromNetwork is always true in compat mode because partial packets are not supported.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlDataReader.cs
```diff
@@ -5753,6 +5751,10 @@ private static Task<T> GetFieldValueAsyncExecute<T>(Task task, object state)
                     {
                         return Task.FromResult<T>(reader.GetFieldValueFromSqlBufferInternal<T>(reader._data[columnIndex], reader._metaData[columnIndex], isAsync: true));
                     }
+                    else
+                    {
+                        return reader.ExecuteAsyncCall(context);
```
mdaigle, line 5398: Can you help me better understand this change? It's not clear to me which issue this addresses.

Wraith2: from the edit I made to the OP.

Fixes 3572.
When reading data sequentially and calling GetFieldValueAsync it is possible for the call to TryReadColumnInternal to fail because there is not enough data. When this happens an async read has been started so we should wait for it to return but the code does not do this and instead falls through to attempting a call to TryReadColumnValue. Under a debugger it's clear that there is a problem because you hit an assert but at runtime this was silently failing which could cause data loss as an async read collided with a sync read.. The fix is to prevent the fallthrough and queue the wait for the next packet.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.Multiplexer.cs
Wraith2: It's safer to discard it. One of the problems with packet buffers is that the lifetimes are muddy. The buffer that we're discarding could be the only reference to it or it could be part of the snapshot. 
One of the things you can't do with a profiler that I'd really like to be able to do is to trace when the last reference to objects is lost and they become eligible for GC. With that information I'd be able to trace and re-use packets in the tds layer as I've tried to do in the managed network layer.

So while it seems that it was safe previously to reset to the beginning of the buffer it's also safe to allocate a new buffer.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
Wraith2: Rebased and done.

edwardneal:
Thanks @mdaigle. I've re-run the stress tests with both of those AppContext switches set, and the behaviour's consistent from end to end - #3572 is fixed in (almost) all cases, irrespective of the AppContext switch settings.

---

I've noted one bug which appears under specific circumstances. This bug appears when run against M.D.S 5.1.7, 5.2.3, 6.0.2, 6.1.1 and this PR (with the UseCompatibilityProcessSni switch set to true.)

The circumstances in question are:
* Server: localhost (bug doesn't appear for Azure SQL databases in UKSouth or AustraliaEast.)
* UseCompatibilityAsyncBehaviour: irrelevant
* UseCompatibilityProcessSni: set to true (doesn't appear with this unset.)
* Operation: read a varbinary(max) as a stream
* Packet size: 567 bytes (doesn't appear when the packet size is 8000.)
* Data size: 1MB or greater
* Type: async
* SqlDataReader CommandBehaviour: SequentialAccess

When this bug appears, disposing of the SqlDataReader throws the exception below:
```
System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values.
   at Microsoft.Data.SqlClient.TdsParserStateObject.TryReadByteArray(Span`1 buff, Int32 len, Int32& totalRead)
   at Microsoft.Data.SqlClient.TdsParserStateObject.TryReadUInt32(UInt32& value)
   at Microsoft.Data.SqlClient.TdsParserStateObject.TryReadPlpLength(Boolean returnPlpNullIfNull, UInt64& lengthLeft)
   at Microsoft.Data.SqlClient.TdsParser.TrySkipPlpValue(UInt64 cb, TdsParserStateObject stateObj, UInt64& totalBytesSkipped)
   at Microsoft.Data.SqlClient.SqlDataReader.TryResetBlobState()
   at Microsoft.Data.SqlClient.SqlDataReader.TryCleanPartialRead()
   at Microsoft.Data.SqlClient.SqlDataReader.TryCloseInternal(Boolean closeReader)
   at Microsoft.Data.SqlClient.SqlDataReader.Close()
   at Microsoft.Data.SqlClient.SqlDataReader.Dispose(Boolean disposing)
```

I think it might be a reproduction of #1044.

---

You're welcome to cannibalise my test harness - link is [here](https://gist.github.com/edwardneal/872950601472023dda6ae7dfe67c474b). A few notes on them though:
* BenchmarkDotNet almost certainly isn't the right way to represent these - it was just a little more convenient for dealing with the various combinations of parameters and measuring performance. In particular, sorting through the results to find combinations of parameters which produce failures is sluggish.
* There are 3,024 combinations of tests here, and by the time we test the presence/absence of the two AppContext switches that could spiral to 12,096. I pared down the number of tests (eliminating the Xml test) to cut down the expected time to execute them, and found that it took 1.5 days to run 2,592 combinations.
* The Xml test is very memory hungry; I couldn't run it without slowing down the benchmark by several days. If there are any undiscovered bugs, they're likely in this area.
* The `VarCharAsTextReader` and `BlobAsStream` operations take almost identical code paths within SqlClient - if there's a reliable bug in that area, it'll probably appear in both operations. I found it was a little easier to reproduce bugs with `BlobAsStream`.
* The `BlobAsByteArray` operation doesn't have an async counterpart.
* I've not tested the use of `SqlBytes` or `SqlBinary`, I expect that these are covered by the testing of `byte[]` and `Stream`.
* I determined whether or not each combination passed by looking for benchmarks which didn't complete, then checking logs to see whether this was a timeout-related exception. Thankfully this didn't happen often.

Wraith2:
> The Xml test is very memory hungry; I couldn't run it without slowing down the benchmark by several days. If there are any undiscovered bugs, they're likely in this area.

Results from my dev branch:
| Method | UseContinue | Mean        | Error     | StdDev    | Gen0         | Gen1         | Gen2        | Allocated   |
|------- |------------ |------------:|----------:|----------:|-------------:|-------------:|------------:|------------:|
| Sync   | False       |    72.29 ms |  0.269 ms |  0.252 ms |    2857.1429 |    2571.4286 |    857.1429 |    35.91 MB |
| Async  | False       | 2,220.91 ms | 20.277 ms | 15.831 ms | 1450000.0000 | 1355000.0000 | 141000.0000 | 20989.11 MB |
| Sync   | True        |    71.04 ms |  0.454 ms |  0.379 ms |    2875.0000 |    2625.0000 |    875.0000 |    35.91 MB |
| Async  | True        |    70.63 ms |  0.926 ms |  0.821 ms |    3125.0000 |    3000.0000 |   1125.0000 |    37.04 MB |

paulmedynski (approved):

mdaigle:
FYI, we're having issues with macos tests at the moment, separate from this PR

Wraith2:
Any specific thing blocking merging?

paulmedynski:
Waiting on #3597 to fix our macOS CI issues.  One that merges, you can re-merge main and we can kick off another CI run.

Wraith2:
that pr to fix the ci  merged, rebased this pr. please run the CI.

ErikEJ:
So to take it to 11, what do I need to do with the latest main artifact? (Planning to share the simplest possible approach in a blog post)

Wraith2:
This makes main stable. It doesn't contain the speed improvements because this code can never enter an async-continue state. If you want to use this PR artifact for the sake of being on main and getting stability then that's fine, but that seems like a small use-case given that we have 6.1.1 which is stable.

I have some work in a branch which will add the speed improvements back in. I'll get started on that this evening.

ErikEJ:
@Wraith2 will that be opt-in via a switch then?

Wraith2:
Yes.

PR #3535, opened 2025-07-31
Remove CER

benrr101:
## Description
This is a fairly major change. 

https://learn.microsoft.com/en-us/dotnet/framework/performance/constrained-execution-regions#code-not-permitted-in-cers

For the longest time, and especially as part of this common project merge, we have maintained these Constrained Execution Regions as part of our netfx implementation. However, they've been a major pain to work around, since they involve try/catch/finally structures that aren't needed in netcore, or catching exceptions that are never expected to be caught in netcore. There have been several approaches brought forward to conditionally include the CER structures - some better than others. While working on the SqlCommand merge, I suggested another approach that moves the CER structure to a helper (see #3514). This uncovered a major limitation on CER blocks - that function pointers/delegates cannot be used (see link above).

This prompted a bit more research, and it appears that our codebase is riddled with code that cannot be called from a CER block. Since .NET 2.0, CER has been a guideline instead of a hard rule, and CER execution has become best-effort rather than guaranteed, and as a result much more permissive in what can be prepared/executed. Even with obvious issues in the code, CI has passed for ages, and no users have complained of faults arising from CER failures. As such, the question arises - **do we even need CER anymore?** The answer to this, so far, seems to be **no**.

So, in this PR, I searched the codebase for instances of `PrepareConstrainedRegion` and just removed the entire try/catch/finally structure where possible.

Here's some examples of places where we're doing unapproved things inside CER blocks:

* `SqlCommand.InternalExecuteNonQuery`
  * Called from `SqlCommand.ExecuteNonQuery`
  * Calls `RunExecuteNonQueryTds` which contains a lambda being called on timeout (among others)
* `SqlCommand.ExecuteReader`
  * Explicitly allocates the reader inside the constrained execution region 
  * Uses lambdas all over
* `SqlConnection.InternalOpenAsync`
  * Called from `SqlConnection.OpenAsync`
  * Uses function pointer in retry callback
* `TdsParserStateObject.ReadSniError`
  * Calls `TdsParserStateObject.ProcessSniPacket` which explicitly allocates a byte array

This PR is best viewed with whitespace changes ignored

## Issues
Will make remaining work for #1261 easier

## Testing
Project still builds, CI will hopefully be sufficient for validation of behavior.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
@@ -1411,18 +1411,6 @@ namespace Microsoft.Data.SqlClient
                 CachedAsyncState.SetActiveConnectionAndResult(completion, nameof(EndExecuteNonQuery), _activeConnection);
                 _stateObj.ReadSni(completion);
             }
-            // Cause of a possible unstable runtime situation on facing with `OutOfMemoryException` and `StackOverflowException` exceptions,
-            // trying to call further functions in the catch of either may fail that should be considered on debuging!
-            catch (System.OutOfMemoryException e)
-            {
-                _activeConnection.Abort(e);
-                throw;
-            }
-            catch (System.StackOverflowException e)
-            {
-                _activeConnection.Abort(e);
-                throw;
-            }
             catch (Exception)
             {
                 // Similarly, if an exception occurs put the stateObj back into the pool.
@@ -1967,20 +1955,6 @@ namespace Microsoft.Data.SqlClient
                 // must finish caching information before ReadSni which can activate the callback before returning
                 CachedAsyncState.SetActiveConnectionAndResult(completion, nameof(EndExecuteXmlReader), _activeConnection);
                 _stateObj.ReadSni(completion);
-            }
-            // Cause of a possible unstable runtime situation on facing with `OutOfMemoryException` and `StackOverflowException` exceptions,
-            // trying to call further functions in the catch of either may fail that should be considered on debuging!
-            catch (System.OutOfMemoryException e)
-            {
-                _activeConnection.Abort(e);
-                completion.TrySetException(e);
-                throw;
-            }
-            catch (System.StackOverflowException e)
-            {
-                _activeConnection.Abort(e);
-                completion.TrySetException(e);
-                throw;
             }
             catch (Exception e)
             {
@@ -2632,20 +2606,6 @@ namespace Microsoft.Data.SqlClient
                 // must finish caching information before ReadSni which can activate the callback before returning
                 CachedAsyncState.SetActiveConnectionAndResult(completion, nameof(EndExecuteReader), _activeConnection);
                 _stateObj.ReadSni(completion);
-            }
-            // Cause of a possible unstable runtime situation on facing with `OutOfMemoryException` and `StackOverflowException` exceptions,
-            // trying to call further functions in the catch of either may fail that should be considered on debuging!
-            catch (System.OutOfMemoryException e)
-            {
-                _activeConnection.Abort(e);
-                completion.TrySetException(e);
-                throw;
-            }
-            catch (System.StackOverflowException e)
-            {
-                _activeConnection.Abort(e);
-                completion.TrySetException(e);
-                throw;
             }
             catch (Exception e)
             {
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
@@ -4114,9 +4114,7 @@ namespace Microsoft.Data.SqlClient
                     SqlCommand command = (SqlCommand)state;
                     bool processFinallyBlockAsync = true;
                     bool decrementAsyncCountInFinallyBlockAsync = true;
-#if NETFRAMEWORK 
-                    RuntimeHelpers.PrepareConstrainedRegions();
-#endif
+
                     try
                     {
                         // Check for any exceptions on network write, before reading.
@@ -4187,10 +4185,7 @@ namespace Microsoft.Data.SqlClient
             {
                 bool processFinallyBlockAsync = true;
                 bool decrementAsyncCountInFinallyBlockAsync = true;
-
-#if NETFRAMEWORK
-                RuntimeHelpers.PrepareConstrainedRegions();
-#endif
+                
                 try
                 {
                     // Check for any exceptions on network write, before reading.
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnection.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnection.cs
@@ -1237,34 +1237,11 @@ namespace Microsoft.Data.SqlClient
             SqlStatistics statistics = null;
             RepairInnerConnection();
             SqlClientEventSource.Log.TryCorrelationTraceEvent("SqlConnection.ChangeDatabase | API | Correlation | Object Id {0}, Activity Id {1}, Database {2}", ObjectID, ActivityCorrelator.Current, database);
-            TdsParser bestEffortCleanupTarget = null;
-
-#if NETFRAMEWORK
-            RuntimeHelpers.PrepareConstrainedRegions();
-#endif
+            
             try
             {
-                bestEffortCleanupTarget = SqlInternalConnection.GetBestEffortCleanupTarget(this);
                 statistics = SqlStatistics.StartTimer(Statistics);
                 InnerConnection.ChangeDatabase(database);
-            }
-            catch (System.OutOfMemoryException e)
-            {
-                Abort(e);
-                throw;
-            }
-            catch (System.StackOverflowException e)
-            {
-                Abort(e);
-                throw;
-            }
-            catch (System.Threading.ThreadAbortException e)
-            {
-                Abort(e);
-#if NETFRAMEWORK
-                SqlInternalConnection.BestEffortCleanup(bestEffortCleanupTarget);
-#endif
-                throw;
             }
             finally
             {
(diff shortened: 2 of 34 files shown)


benrr101:
@edwardneal @Wraith2 @ErikEJ Would love to get some feedback on this idea. It's definitely a Change :tm:

edwardneal:
I definitely agree that the CERs shouldn't stay as they are; we should either write code which meets the reliability contract (and include the use case in CI, review with the SQL Server behaviour in mind, etc.) or stop advertising it as a capability. There's additional [linked documentation](https://learn.microsoft.com/en-us/dotnet/framework/performance/reliability-best-practices) which highlights other areas where we might also fall short.

This could have an impact on situations where someone is running SqlClient within a SQLCLR environment, but not using the context connection (i.e. a custom CLR stored procedure which instantiates a SqlConnection to a separate SQL Server instance.) I wasn't able to load the netfx SqlClient assembly when I tested in #2838, and at the time it wasn't a supported scenario.

I'm in favour of removing the CERs.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
```diff
@@ -1411,18 +1411,6 @@ private void BeginExecuteNonQueryInternalReadStage(TaskCompletionSource<object>
...
-                throw;
-            }
-            catch (System.StackOverflowException e)
-            {
-                _activeConnection.Abort(e);
-                throw;
-            }
             catch (Exception)
             {
                 // Similarly, if an exception occurs put the stateObj back into the pool.
```
paulmedynski, line 1417: Should this catch be calling `_activeConnection.Abort(e)` like the other catches used to?

Same with the other 2 below.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnection.cs
```diff
@@ -1237,35 +1237,12 @@ public override void ChangeDatabase(string database)
...
+            
             try
             {
-                bestEffortCleanupTarget = SqlInternalConnection.GetBestEffortCleanupTarget(this);
                 statistics = SqlStatistics.StartTimer(Statistics);
                 InnerConnection.ChangeDatabase(database);
             }
-            catch (System.OutOfMemoryException e)
-            {
-                Abort(e);
```
paulmedynski, line 1253: Should we still be calling `Abort(e)`?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -539,9 +539,6 @@ internal SqlInternalConnectionTds(
             _parserLock.Wait(canReleaseFromAnyThread: false);
             ThreadHasParserLockForClose = true;   // In case of error, let ourselves know that we already own the parser lock
 
-#if NETFRAMEWORK
```
paulmedynski, line 542: Why did these netcore files have `#if NETFRAMEWORK` in them anyway?  Was it an effort to make them similar to the netfx files so the merge would be simpler?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -573,21 +570,6 @@ internal SqlInternalConnectionTds(
                     }
                 }
             }
-            catch (System.OutOfMemoryException)
-            {
-                DoomThisConnection();
```
paulmedynski, line 578: No more doom on an exception?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlTransaction.cs
```diff
@@ -248,40 +191,16 @@ public override void Rollback()
...
-                        throw;
-                    }
-                    catch (ThreadAbortException e)
-                    {
-                        _connection.Abort(e);
-                        SqlInternalConnection.BestEffortCleanup(bestEffortCleanupTarget);
-                        throw;
-                    }
-                    #else
+                    #if NET
```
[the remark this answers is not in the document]
paulmedynski: It's correct as-is.

paulmedynski (changes requested):
Some removed catch blocks were calling methods that are no longer called.  Should they be?  Can those methods be removed now?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
```diff
@@ -1411,18 +1411,6 @@ private void BeginExecuteNonQueryInternalReadStage(TaskCompletionSource<object>
                 CachedAsyncState.SetActiveConnectionAndResult(completion, nameof(EndExecuteNonQuery), _activeConnection);
                 _stateObj.ReadSni(completion);
             }
-            // Cause of a possible unstable runtime situation on facing with `OutOfMemoryException` and `StackOverflowException` exceptions,
-            // trying to call further functions in the catch of either may fail that should be considered on debuging!
-            catch (System.OutOfMemoryException e)
```
mdaigle, line 1416: We can't catch System.StackOverflowException and System.Threading.ThreadAbortException isn't a thing in net core so it makes sense to remove those, but System.OutOfMemoryException is catchable... Can you explain a bit more of your reasoning for removing OOM catches?

Wraith2: The ability to recover from an OOM is unlikely and even if you somehow manage it without corrupting internal state there isn't a much you can do without the ability to allocate.

Wraith2:
This is a topic i've discussed with David Engel in the past. My understanding is that M.D.SqlClient is not supported or usable within sql server. The only supported version is S.D.SqlClient. With the deprecation of S.D.SqlClient I'm not sure what this means for the SQLCLR feature but until there is official confirmation that M.D.SqlClient is supported in SQL Server I would not expect it to work and if someone tries it and it does work it may just be luck, not something to rely on.

Netcore and now NET is not and will never be a supported host in sql server because much of the work done to improve netcore over netfx was removal of things that SQLCLR needed. Things like being able to terminate at any arbitrary point in execution without corrupting the runtime are simply not possible and were a constant source of bugs. ThreadAbortException, CER and related work were part of this featureset and have been removed.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
benrr101: I'm pretty sure the answer to that is: no
We can have exceptions from stuff like syntax errors in the SQL, so I suspect that we wouldn't want to abort the connection everytime that happens. But feel free to correct me.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlConnection.cs
benrr101: This one, we don't have a general exception catch, so the only time we're currently aborting is if something critical happens. I don't think we want to abort in all exception conditions.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
benrr101: I'm gonna take a guess without looking at git blame and say @edwardneal put this in the netcore implementation to make merging easier. I'm like 90% sure that's the case :) Not trying to throw @edwardneal under the bus, tho!

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
benrr101: We were on dooming on critical exceptions, there's no general exception case here. It's another scenario where I don't think we want to doom on all exceptions.

priyankatiwari08 (approved):

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.cs
mdaigle: I'm generally suspicious of how we handle and recover from OOM. Let's say an application decides it does want to catch OOM exceptions, that connection that was in use when the exception was thrown might end up back in the pool and recycled even though it has an inconsistent state.
 
Likewise for thread abort and stack overflow. They are catchable in .NET framework and an application could choose to do so.
 
Typically, it's up to the application to handle these types of things at the end of the day, but we have a good number of pooled objects that need to be kept in a good state.

Wraith2: My choice would be to abandon any attempt to return objects to pools in the case of an OOM. If an OOM happens anywhere there is a high likelihood of the process quitting and if that happens maintaining pooled resources is useless work. pooling is also a likely cause of pressure leading to the OOM in the first place. 

IMO the _only_ thing we should be doing for exceptions we don't own outside networking layers is to make sure any unmanaged resources are released correctly. I think we're already using safehandles to do this wherever possible.

mdaigle: Most of these exception handlers are just setting a "doomed" flag on the connection to make sure they're _not_ pooled. So it's less of getting object back into pools and more of making a best effort _not_ to pool corrupted objects.

I've discussed with @paulmedynski and @benrr101 offline and we agreed to proceed with removing the try/catches in this PR. We'll come back and do an investigation/POC into our critical exception handling and add handling back if proven useful.

mdaigle (approved):

PR #3542, opened 2025-08-05
Add CodeQL suppression for DefaultAzureCredential use in Production

paulmedynski:
- Added CodeQL suppression to the same line as the code it is suppressing.


cheenamalhotra:
This will be back-ported to release branches as well?

paulmedynski:
Yes, we can backport for the ReceiveTimeout change.  The CodeQL part is benign on non-default branches AFAIK.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ManagedSni/SniTcpHandle.netcore.cs
```diff
@@ -953,9 +953,30 @@ public override uint Receive(out SniPacket packet, int timeoutInMilliseconds)
...
-                    // to avoid blocking the thread in case of a timeout error.
-                    _socket.ReceiveTimeout = Timeout.Infinite;
+                    const int resetTimeout = Timeout.Infinite;
+
+                    try
+                    {
+                        // Reset the socket timeout to Timeout.Infinite after
+                        // the receive operation is done to avoid blocking the
+                        // thread in case of a timeout error.
+                        _socket.ReceiveTimeout = resetTimeout;
```
mdaigle, line 963: I don't understand when this reset would be useful. Our receive on the socket here is synchronous, so either we hit the timeout during that call, or we don't.

We always reset the timeout when we call this method, so at least failing to reset shouldn't be an issue for subsequent receives...

paulmedynski: Yes it's suspicious indeed, but I didn't want to remove it outright today.

PR #3538, opened 2025-08-02
Update Azure.Identity and other dependencies

cheenamalhotra:
Addresses #3509 
_(needs to be back-ported to release/6.1, release/6.0, and release/5.1)_

PR #3553, opened 2025-08-08
[6.0] Update Azure.Identity and other dependencies (#3538)

cheenamalhotra:
Ports #3538 to release/6.0

### DIFF
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectivityTests/AADConnectionTest.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectivityTests/AADConnectionTest.cs
@@ -501,7 +501,7 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests
 
             SqlException e = Assert.Throws<SqlException>(() => ConnectAndDisconnect(connStrWithNoCred));
 
-            string expectedMessage = "ManagedIdentityCredential authentication unavailable";
+            string expectedMessage = "[Managed Identity] Authentication";
             Assert.Contains(expectedMessage, e.GetBaseException().Message);
         }
 


tools/specs/Microsoft.Data.SqlClient.nuspec
```diff
@@ -52,15 +56,17 @@ When using NuGet 3.x this package requires at least version 3.4.</description>
...
         <dependency id="Azure.Identity" version="1.11.4" />
         <dependency id="Microsoft.Extensions.Caching.Memory" version="9.0.4" exclude="Compile" />
         <dependency id="Microsoft.IdentityModel.JsonWebTokens" version="7.5.0" />
         <dependency id="Microsoft.IdentityModel.Protocols.OpenIdConnect" version="7.5.0" />
         <dependency id="Microsoft.SqlServer.Server" version="1.0.0"/>
         <dependency id="System.Configuration.ConfigurationManager" version="9.0.4" exclude="Compile" />
-        <dependency id="System.Security.Cryptography.Pkcs" version="9.0.4"/>
-        <dependency id="Microsoft.Bcl.Cryptography" version="9.0.4"/>
+        <dependency id="System.Security.Cryptography.Pkcs" version="9.0.4" />
+        <dependency id="System.Text.Json" version="9.0.5" />
```
Frulfump, line 64: I don't think System.Text.Json 9.x is necessary in the net9.0 group as it's already supplied by the runtime or is it added here due to something else pulling in a vulnerable version (and package pruning not being enabled so it causes a false positive)? Should clarify with a comment if so.

cheenamalhotra: This is done as to override the defaults and we directly reference its APIs.

paulmedynski (approved):

PR #3554, opened 2025-08-09
Performance | Introduce lower-allocation AlwaysEncrypted primitives

edwardneal:
## Description

I'm trying to improve query performance in AE scenarios, and part of this means a review of the column encryption providers. These were written prior to a number of cryptography improvements in .NET 5.0; as a result of these improvements, we can save a little memory.

Besides the performance improvements, each provider reproduces the same serialization logic. These primitives centralise that logic; they use the column master key parameters to perform encryption and decryption of a column encryption key with an RSA instance. The only thing that the three `SqlColumnEncryptionKeyStoreProvider` derivatives need to do is retrieve an RSA instance using the master key path.

I considered moving this logic into a protected method on `SqlColumnEncryptionKeyStoreProvider`, but a third-party provider isn't guaranteed to perform encryption/decryption of the CEK using an RSA instance (or indeed, on the same server - `SqlColumnEncryptionAzureKeyVaultProvider` passes that duty to Azure Key Vault.)

A subsequent PR will refactor the column encryption providers to use these primitives, and the benchmark results for this are below. To summarise: no significant variations in execution time, and around a 15% reduction in AlwaysEncrypted memory usage.

### `SqlColumnEncryptionCertificateStoreProvider`

| Method                     | Branch | Mean     | Error    | StdDev   | Ratio | Allocated | Alloc Ratio |
|--------------------------- |------- |---------:|---------:|---------:|------:|----------:|------------:|
| DecryptColumnEncryptionKey | main   | 13.79 ms | 0.256 ms | 0.227 ms |  1.00 |  18.17 KB |        1.00 |
| DecryptColumnEncryptionKey | PR     | 13.16 ms | 0.180 ms | 0.150 ms |  0.95 |  15.43 KB |        0.85 |
| EncryptColumnEncryptionKey | main   | 12.59 ms | 0.224 ms | 0.375 ms |  1.00 |  18.94 KB |        1.00 |
| EncryptColumnEncryptionKey | PR     | 12.58 ms | 0.246 ms | 0.450 ms |  1.00 |  16.01 KB |        0.85 |

## `SqlColumnEncryptionCngProvider`

| Method                     | Branch | Mean     | Error     | StdDev    | Ratio | Allocated | Alloc Ratio |
|--------------------------- |------- |---------:|----------:|----------:|------:|----------:|------------:|
| DecryptColumnEncryptionKey | main   | 3.303 ms | 0.0639 ms | 0.1661 ms |  1.00 |   2.51 KB |        1.00 |
| DecryptColumnEncryptionKey | PR     | 3.285 ms | 0.0648 ms | 0.0842 ms |  0.99 |     901 B |        0.36 |
| EncryptColumnEncryptionKey | main   | 3.214 ms | 0.0395 ms | 0.0330 ms |  1.00 |   3.11 KB |        1.00 |
| EncryptColumnEncryptionKey | PR     | 3.332 ms | 0.0645 ms | 0.0603 ms |  1.03 |   1.61 KB |        0.52 |

## `SqlColumnEncryptionCspProvider`

| Method                     | Branch | Mean     | Error     | StdDev    | Ratio | Allocated | Alloc Ratio |
|--------------------------- |------- |---------:|----------:|----------:|------:|----------:|------------:|
| DecryptColumnEncryptionKey | main   | 2.775 ms | 0.0411 ms | 0.0520 ms |  1.00 |   3.59 KB |        1.00 |
| DecryptColumnEncryptionKey | PR     | 2.737 ms | 0.0256 ms | 0.0200 ms |  0.99 |   3.01 KB |        0.84 |
| EncryptColumnEncryptionKey | main   | 2.815 ms | 0.0556 ms | 0.0571 ms |  1.00 |   4.01 KB |        1.00 |
| EncryptColumnEncryptionKey | PR     | 2.847 ms | 0.0464 ms | 0.0434 ms |  1.01 |   3.23 KB |        0.81 |

## Testing

These primitives aren't used anywhere yet, so there's no test coverage. The existing providers have this coverage though, so the new code will gain coverage from the follow-up PR. I have a local branch where the column encryption providers use this code, and almost all tests pass. The two failing tests are capitalisation errors in exception messages produced by the CSP provider, which will be fixed at the same time.


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AlwaysEncrypted/ColumnMasterKeyMetadata.cs
```diff
@@ -0,0 +1,114 @@
...
+            ? stackalloc char[CharStackAllocationThreshold].Slice(0, masterKeyMetadataLength)
+            : new char[masterKeyMetadataLength];
+        Span<char> masterKeyMetadataSpan = masterKeyMetadata;
+#else
+        char[] masterKeyMetadata = new char[masterKeyMetadataLength];
+        Span<char> masterKeyMetadataSpan = masterKeyMetadata.AsSpan();
+#endif
+
+        providerName.AsSpan().ToLowerInvariant(masterKeyMetadataSpan);
+        masterKeyPath.AsSpan().ToLowerInvariant(masterKeyMetadataSpan.Slice(providerName.Length));
```
Wraith2, line 107: Are the characters in the input variable restricted such that it is certain that the output lower invariant replacement version is certain to be of exactly the same length?

edwardneal: Yes, indirectly. This is due to be used by the certificate store provider, and the constructor takes an `RSA` instance. The only way to get this instance is by using the master key path to select a valid certificate, and the only valid master key paths are ASCII text in a few formats:
* `[LocalMachine|CurrentUser]/My/[40 character SHA1 thumbprint]`
* `My/[40 character SHA1 thumbprint]`
* `[40 character SHA1 thumbprint]`

Wraith2: Excellent. It might be worth leaving a comment in the code somewhere so anyone who reads the code in future knows that there are input restrictions.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AlwaysEncrypted/ColumnMasterKeyMetadata.cs
```diff
@@ -0,0 +1,121 @@
...
+namespace Microsoft.Data.SqlClient.AlwaysEncrypted;
+
+/// <summary>
+/// Represents metadata about the column master key, to be signed or verified by an enclave.
+/// </summary>
+/// <remarks>
+/// This metadata is a lower-case string which is laid out in the following format:
+/// <list type="number">
+/// <item>
+/// Provider name. This always <see cref="SqlColumnEncryptionCertificateStoreProvider.ProviderName"/>.
```
paulmedynski: This _is_ always

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AlwaysEncrypted/ColumnMasterKeyMetadata.cs
```diff
@@ -0,0 +1,121 @@
...
+            ? stackalloc byte[ByteStackAllocationThreshold].Slice(0, byteCount)
+            : new byte[byteCount];
+
+        Encoding.Unicode.GetBytes(masterKeyMetadata, masterKeyMetadataBytes);
+
+        // Compute hash
+        SHA256.HashData(masterKeyMetadataBytes, _hash);
+#else
+        byte[] masterKeyMetadataBytes = Encoding.Unicode.GetBytes(masterKeyMetadata);
+        using SHA256 sha256 = SHA256.Create();
```
paulmedynski: Can this be in a block scope, to avoid it being used by mistake outside the `#if`?  And it will be disposed of immediately.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AlwaysEncrypted/ColumnMasterKeyMetadata.cs
```diff
@@ -0,0 +1,121 @@
...
+        // Compute hash
+        sha256.TransformFinalBlock(masterKeyMetadataBytes, 0, masterKeyMetadataBytes.Length);
+        _hash = sha256.Hash;
+#endif
+
+        _rsa = rsa;
+    }
+
+    public byte[] Sign() =>
+        _rsa.SignHash(_hash, s_hashAlgorithm, RSASignaturePadding.Pkcs1);
```
paulmedynski, line 141: Since this is invariant, can we compute it during construction?  Or is it more efficient to compute it each time here rather than storing it?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AlwaysEncrypted/ColumnMasterKeyMetadata.cs
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    private readonly Sha256Hash _hash;
+#else
+    private readonly byte[] _hash;
+#endif
+    private readonly RSA _rsa;
+
+    // @TODO: SqlColumnEncryptionCertificateStoreProvider.SignMasterKeyMetadata and .VerifyMasterKeyMetadata should use this type.
+    public ColumnMasterKeyMetadata(RSA rsa, string masterKeyPath, string providerName, bool allowEnclaveComputations)
```
paulmedynski, line 72: Can we get XML docs for the public API throughout this class?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AlwaysEncrypted/ColumnMasterKeyMetadata.cs
```diff
@@ -0,0 +1,121 @@
...
+/// </item>
+/// <item>
+/// Boolean to indicate whether the CMK supports enclave computations. This is either <c>true</c> or <c>false</c>.
+/// </item>
+/// </list>
+/// <para>
+/// This takes ownership of the RSA instance supplied to it, disposing of it when Dispose is called.
+/// </para>
+/// </remarks>
+internal readonly ref struct ColumnMasterKeyMetadata // : IDisposable
```
paulmedynski, line 35: I'm going to change `<LangVersion>` from `latest` to `13` so we can avoid this `// : IDisposable` for ref structs.  I'll hunt these down at the same time.  Just FYI.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AlwaysEncrypted/EncryptedColumnEncryptionKeyParameters.cs
```diff
@@ -0,0 +1,220 @@
...
+/// Version: 1 byte, always 0x01
+/// </item>
+/// <item>
+/// Key path length: 2 bytes, length of the key path in bytes
+/// </item>
+/// <item>
+/// Ciphertext length: 2 bytes, length of the ciphertext in bytes
+/// </item>
+/// <item>
+/// Key path: variable length, Unicode-encoded string representing the key path
```
paulmedynski: I think you mean `UTF-16 encoded` here.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AlwaysEncrypted/ColumnMasterKeyMetadata.cs
```diff
@@ -0,0 +1,121 @@
...
+        Span<char> masterKeyMetadataSpan = masterKeyMetadata;
+#else
+        char[] masterKeyMetadata = new char[masterKeyMetadataLength];
+        Span<char> masterKeyMetadataSpan = masterKeyMetadata.AsSpan();
+#endif
+
+        providerName.AsSpan().ToLowerInvariant(masterKeyMetadataSpan);
+        masterKeyPath.AsSpan().ToLowerInvariant(masterKeyMetadataSpan.Slice(providerName.Length));
+        enclaveComputationSpan.ToLowerInvariant(masterKeyMetadataSpan.Slice(providerName.Length + masterKeyPath.Length));
+        byteCount = Encoding.Unicode.GetByteCount(masterKeyMetadata);
```
paulmedynski, line 109: Weird that they didn't call this `Encoding.UTF16`.  Unicode is a character set, not an encoding - shrug!

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AlwaysEncrypted/EncryptedColumnEncryptionKeyParameters.cs
```diff
@@ -0,0 +1,220 @@
...
+    private static readonly HashAlgorithmName s_hashAlgorithm = HashAlgorithmName.SHA256;
+
+    private readonly RSA _rsa;
+    private readonly int _rsaKeySize;
+    private readonly string _keyPath;
+    private readonly string _keyType;
+    private readonly string _keyPathReference;
+
+    // @TODO: SqlColumnEncryptionCertificateStoreProvider, SqlColumnEncryptionCngProvider and SqlColumnEncryptionCspProvider should use this type.
+    public EncryptedColumnEncryptionKeyParameters(RSA rsa, string keyPath, string keyType, string keyPathReference)
```
paulmedynski, line 85: XML docs for the public APIs of this class.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AlwaysEncrypted/EncryptedColumnEncryptionKeyParameters.cs
```diff
@@ -0,0 +1,220 @@
...
+        Debug.Assert(keyPathReference is SqlColumnEncryptionCertificateStoreProvider.KeyPathReference
+            or SqlColumnEncryptionCngProvider.KeyPathReference or SqlColumnEncryptionCspProvider.KeyPathReference);
+        _keyType = keyType;
+        _keyPathReference = keyPathReference;
+    }
+
+    public byte[] Encrypt(byte[] columnEncryptionKey)
+    {
+        ushort keyPathSize = (ushort)Encoding.Unicode.GetByteCount(_keyPath);
+        int cekSize = sizeof(byte) + sizeof(ushort) + sizeof(ushort) + keyPathSize + _rsaKeySize + _rsaKeySize;
```
paulmedynski: Is this saying that the signature size (bytes) is the same as the RSA key size (bytes)?

Would it be clearer to have a signatureSize variable with a comment?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AlwaysEncrypted/EncryptedColumnEncryptionKeyParameters.cs
```diff
@@ -0,0 +1,220 @@
...
+/// Represents the parameters used to construct an encrypted column encryption key, used to encrypt and decrypt data in SQL Server Always Encrypted columns.
+/// </summary>
+/// <remarks>
+/// An encrypted CEK is a byte array that contains the following structure:
+/// <list type="number">
+/// <item>
+/// Version: 1 byte, always 0x01
+/// </item>
+/// <item>
+/// Key path length: 2 bytes, length of the key path in bytes
```
paulmedynski: Should mention the endianness for these two multi-byte lengths.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AlwaysEncrypted/EncryptedColumnEncryptionKeyParameters.cs
```diff
@@ -0,0 +1,220 @@
...
+        
+        Debug.Assert(keyType is SqlColumnEncryptionCertificateStoreProvider.MasterKeyType
+            or SqlColumnEncryptionCngProvider.MasterKeyType or SqlColumnEncryptionCspProvider.MasterKeyType);
+        Debug.Assert(keyPathReference is SqlColumnEncryptionCertificateStoreProvider.KeyPathReference
+            or SqlColumnEncryptionCngProvider.KeyPathReference or SqlColumnEncryptionCspProvider.KeyPathReference);
+        _keyType = keyType;
+        _keyPathReference = keyPathReference;
+    }
+
+    public byte[] Encrypt(byte[] columnEncryptionKey)
```
paulmedynski, line 111: Why is the parameters struct performing encryption/decryption?  From the docs above, this struct sounds like a bucket that holds some values used for encryption/decryption, which is implemented elsewhere.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AlwaysEncrypted/EncryptedColumnEncryptionKeyParameters.cs
```diff
@@ -0,0 +1,220 @@
...
+        Buffer.BlockCopy(signedHash, 0, encryptedColumnEncryptionKey, signatureOffset, bytesWritten);
+#endif
+
+        return encryptedColumnEncryptionKey;
+    }
+
+    public byte[] Decrypt(byte[] encryptedCek)
+    {
+        // Validate the version byte
+        if (encryptedCek[0] != AlgorithmVersion)
```
paulmedynski: Use `AlgorithmOffset` instead of `0` ?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AlwaysEncrypted/ColumnMasterKeyMetadata.cs
edwardneal: It's more efficient to compute it. `ColumnMasterKeyMetadata` is only going to be used by SqlColumnEncryptionCertificateStoreProvider, and this type's `Sign` and `Verify` methods correspond to `SignMasterKeyMetadata` and `VerifyMasterKeyMetadata`.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AlwaysEncrypted/EncryptedColumnEncryptionKeyParameters.cs
edwardneal: I struggled to name this particular struct, so the documentation might not be as helpful as it needs to be. The main purpose of this struct is to lift all of the (practically) copy-pasted cryptographic logic out of the three `SqlColumnEncryptionKeyStoreProvider` implementations, improving performance as we do so.

I've used a dedicated struct for this because conceptually, we're operating on a specific kind of object (a CEK) - we're just encrypting or decrypting it. I then decided to keep those two operations tied to the type, rather than introducing another floating static class of helper methods, it seemed tidier.

I considered having an `EncryptedColumnEncryptionKey` struct, with a constructor accepting a "raw" byte array containing the encrypted CEK and the cryptographic parameters, a static `FromPlaintext` factory method accepting a plaintext CEK with the cryptographic parameters, a `Decrypt` method to get the CEK and an `EncryptedValue` field. It felt a little like overkill though.

Both ideas would be used roughly as below:

```cs
byte[] DecryptColumnEncryptionKey(string masterKeyPath, string encryptionAlgorithm,
    byte[] encryptedColumnEncryptionKey)
{
    RSA rsa = GetKey();
    // As-is:
    using EncryptedColumnEncryptionKeyParameters encryptedCekParams = new(rsa, masterKeyPath,
        MasterKeyType, KeyPathReference);
    return encryptedCekParams.Decrypt(encryptedColumnEncryptionKey);

    // Alternatively:
    using EncryptedColumnEncryptionKey encryptedCek = new(encryptedColumnEncryptionKey, rsa, masterKeyPath,
        MasterKeyType, KeyPathReference);
    return encryptedCek.Decrypt();
}

byte[] EncryptColumnEncryptionKey(string masterKeyPath, string encryptionAlgorithm,
    byte[] columnEncryptionKey)
{
    RSA rsa = GetKey();
    // As-is:
    using EncryptedColumnEncryptionKeyParameters encryptedCekParams = new(rsa, masterKeyPath,
        MasterKeyType, KeyPathReference);
    return encryptedCekParams.Encrypt(columnEncryptionKey);

    // Alternatively:
    using EncryptedColumnEncryptionKey encryptedCek = EncryptedColumnEncryptionKey.FromPlaintext(
        columnEncryptionKey, rsa, masterKeyPath, MasterKeyType, KeyPathReference);
    return encryptedCek.EncryptedValue;
}
```

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AlwaysEncrypted/ColumnMasterKeyMetadata.cs
```diff
@@ -100,22 +119,42 @@ public ColumnMasterKeyMetadata(RSA rsa, string masterKeyPath, string providerNam
...
     }
 
+    /// <summary>
+    /// Signs the current master key metadata using the RSA key associated with this instance.
+    /// </summary>
+    /// <returns>
+    /// A byte array containing the digital signature of the master key metadata.
+    /// </returns>
     public byte[] Sign() =>
         _rsa.SignHash(_hash, s_hashAlgorithm, RSASignaturePadding.Pkcs1);
```
paulmedynski, line 141: It looks like this could throw at least CryptographicException, so that should be documented above.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AlwaysEncrypted/ColumnMasterKeyMetadata.cs
```diff
@@ -50,6 +50,25 @@ private struct Sha256Hash
...
+    /// <param name="masterKeyPath">The path to the column master key. This must be a valid path in one of the following formats:
+    /// <list type="bullet">
+    /// <item>[LocalMachine|CurrentUser]/My/[40-character SHA1 thumbprint]</item>
+    /// <item>My/[40-character SHA1 thumbprint]</item>
+    /// <item>[40-character SHA1 thumbprint]</item>
+    /// </list>
+    /// The path is case-insensitive and will be converted to lowercase for processing.</param>
+    /// <param name="providerName">The name of the provider associated with the column master key.</param>
+    /// <param name="allowEnclaveComputations">A value indicating whether enclave computations are allowed for this column master key.</param>
     public ColumnMasterKeyMetadata(RSA rsa, string masterKeyPath, string providerName, bool allowEnclaveComputations)
```
paulmedynski, line 72: Does this constructor throw any exceptions?  There are several method calls that _could_ throw, but perhaps you have setup their inputs to avoid that.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AlwaysEncrypted/EncryptedColumnEncryptionKeyParameters.cs
```diff
@@ -79,10 +96,20 @@ public EncryptedColumnEncryptionKeyParameters(RSA rsa, string keyPath, string ke
...
 
+    /// <summary>
+    /// Encrypts the specified column encryption key using the RSA key associated with this instance.
+    /// </summary>
+    /// <param name="columnEncryptionKey">The plaintext column encryption key to encrypt.</param>
+    /// <returns>
+    /// The encrypted column encryption key, including metadata such as the key path, ciphertext, and a digital signature
+    /// for integrity verification.
+    /// </returns>
     public byte[] Encrypt(byte[] columnEncryptionKey)
```
paulmedynski, line 111: RSA.Encrypt() can throw at least CryptographicException.  Other method calls may also throw.  Please document these.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AlwaysEncrypted/ColumnMasterKeyMetadata.cs
```diff
@@ -0,0 +1,160 @@
...
+        _rsa.VerifyHash(_hash, signature, s_hashAlgorithm, RSASignaturePadding.Pkcs1);
+
+    /// <summary>
+    /// Releases all resources used by this <see cref="ColumnMasterKeyMetadata"/>.
+    /// </summary>
+    /// <remarks>
+    /// This method disposes the <see cref="RSA"/> instance used to construct this <see cref="ColumnMasterKeyMetadata" /> instance.
+    /// </remarks>
+    public void Dispose() =>
+        _rsa.Dispose();
```
paulmedynski, line 161: Will the same RSA instance be passed to EncryptedColumnEncryptionKeyParameters as well?  That would double-dispose it.

paulmedynski (changes requested):
Thanks for the previous updates.  Just a couple more and one question about disposal.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AlwaysEncrypted/ColumnMasterKeyMetadata.cs
edwardneal: Most of the inputs have been arranged to avoid throwing, I've added the exception details to the XML documentation where this isn't the case.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AlwaysEncrypted/ColumnMasterKeyMetadata.cs
edwardneal: No - a unique RSA instance is retrieved every time by the current AlwaysEncrypted providers, and this will continue.

edwardneal:
Thanks. Responded, and merged following the merge of #3014.

edwardneal:
I'd missed that - added to `ColumnMasterKeyMetadata.Sign` and `EncryptedColumnEncryptionKeyParameters.Encrypt`.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AlwaysEncrypted/EncryptedColumnEncryptionKeyParameters.cs
```diff
@@ -0,0 +1,272 @@
...
+        using (SHA256 sha256 = SHA256.Create())
+        {
+            sha256.TransformFinalBlock(encryptedColumnEncryptionKey, 0, signatureOffset);
+            hash = sha256.Hash;
+        }
+        bytesWritten = hash.Length;
+        Debug.Assert(bytesWritten == HashSize, @"Hash size does not match the expected size.");
+
+        byte[] signedHash = _keyType == SqlColumnEncryptionCertificateStoreProvider.MasterKeyType
+            ? _rsa.SignHash(hash, s_hashAlgorithm, RSASignaturePadding.Pkcs1)
```
benrr101, line 170: Since this is new code (and I'm still pretty fuzzy after a couple weeks off), I just want to make triply sure that this is the same signature padding we were using before - since we can't change the padding without breaking at-rest/cross-driver compatibility.

benrr101 (changes requested):
Overall, looks quite good - one BIG question regarding padding schemes, but if you can answer that one, I'll take it right away

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/AlwaysEncrypted/EncryptedColumnEncryptionKeyParameters.cs
edwardneal: It's definitely worth checking. Part of the reason why I split the refactor in half is to make sure that any security review can be split away from the refactor itself. This is going to replace methods in a few places.

[`SqlColumnEncryptionCngProvider.RSASignHashedData`](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlColumnEncryptionCngProvider.Windows.cs#L299-L306): easy enough, a 1:1 confirmation.

[`SqlColumnEncryptionCertificateStoreProvider.RSASignHashedData`](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlColumnEncryptionCertificateStoreProvider.cs#L509-L524): this calls [`RSAPKCS1SignatureFormatter.CreateSignature`](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/RSAPKCS1SignatureFormatter.cs#L46-L56), and we can substitute parameter values to prove this.

[`SqlColumnEncryptionCspProvider.RSASignHashedData`](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlColumnEncryptionCspProvider.Windows.cs#L308-L314): this calls RSACryptoServiceProvider.SignData without explicitly specifying a padding scheme. That follows through as:
* [`SignData(byte[], object)`](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/RSACryptoServiceProvider.Windows.cs#L453-L459)
* [`SignHash(byte[], int)`](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/RSACryptoServiceProvider.Windows.cs#L502-L516)
* [`CapiHelper.SignValue`](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/CapiHelper.Windows.cs#L1159-L1191)
* [`Advapi32.CryptSignHash`](https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-cryptsignhasha)

CryptSignHash is called with `dwFlags` == 0, and the remarks for this API call say:
> By default, the Microsoft RSA providers use the PKCS #1 padding method for the signature.
This also aligns with the comment on SqlColumnEncryptionCspProvider.

benrr101: That sounds good to me. Thanks for the verification!

paulmedynski:
@edwardneal - A merge from main will increase the test timeout and should help with the CI failures.  We also increased the Azure SQL resources to support more parallel PR CI runs to help avoid timeouts.

edwardneal:
Thanks @paulmedynski - I've just merged. Could you re-run CI please?

edwardneal:
Thanks; it looks like there are some lingering CI issues, this time with the NuGet feed.

PR #3555, opened 2025-08-10
Merge | TdsParser functional changes

edwardneal:
## Description

This should cover all necessary functional changes to TdsParser. With these changes merged, the class as a whole should be ready to merge after TdsParserStateObject is dealt with.

Not all of these changes are functional - some are simply necessary to see the real ones clearly. Possibly impactful commits here are:
* 878fcd46225f6ad67fc5402114c9c4bbcae2d557: when TdsParser was writing `int`s and there was space in the packet buffer, `WriteInt` would write the bytes to a stackalloc'd `Span<byte>` and then copy into the buffer. I've eliminated the unnecessary intermediary span, and we now write directly into the buffer.
* a8be2f1d5c9ccc790801494d37954aee02f1257e: netfx had more detailed error messages when the parsing of a TDS stream failed. This ports those messages over to netcore.
* 01b02dbad59a5b7587d568b8ba00182e5cb6f07c: netfx will currently enforce the validation of a server's SSL certificate if Trust Server Certificate is false and the connection's AccessToken or AccessTokenCallback is set, but netcore doesn't currently check the AccessTokenCallback property. This commit changes the behaviour of netcore to match netfx.

There's also one other commit (6d98357ce2df0cfdc85e54c82f8cab086d1aca14) which looks more impactful than it is. `s_xmlMetadataSubstituteSequence` and `s_jsonMetadataSubstituteSequence` are the same length, so there's no runtime impact here.

## Issues

Relates to #1261.

## Testing

Unit tests passed locally.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
+++ b/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
@@ -13309,6 +13309,8 @@ namespace Microsoft.Data.SqlClient
             char[] temp = null;
             bool buffIsRented = false;
             int startOffset = 0;
+
+            stateObj.RequestContinue(true);
             (bool canContinue, bool isStarting, bool isContinuing) = stateObj.GetSnapshotStatuses();
 
             if (canContinue)


src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -10342,27 +10377,23 @@ private Task TDSExecuteRPCAddParameter(TdsParserStateObject stateObj, SqlParamet
         // This is in its own method to avoid always allocating the lambda in TDSExecuteRPCParameter
         private void TDSExecuteRPCParameterSetupWriteCompletion(SqlCommand cmd, IList<_SqlRPC> rpcArray, int timeout, bool inSchema, SqlNotificationRequest notificationRequest, TdsParserStateObject stateObj, bool isCommandProc, bool sync, TaskCompletionSource<object> completion, int startRpc, int startParam, Task writeParamTask)
         {
-            AsyncHelper.ContinueTaskWithState(
+            AsyncHelper.ContinueTask(
```
Wraith2, line 10432: Why don't we want the state parameter anymore?

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -11160,7 +11191,7 @@ internal void WriteBulkCopyMetaData(_SqlMetaDataSet metadataCollection, int coun
                             stateObj.WriteByte(md.scale);
                             break;
                         case SqlDbTypeExtensions.Json:
-                            stateObj.WriteByteArray(s_jsonMetadataSubstituteSequence, s_xmlMetadataSubstituteSequence.Length, 0);
+                            stateObj.WriteByteArray(s_jsonMetadataSubstituteSequence, s_jsonMetadataSubstituteSequence.Length, 0);
```
Wraith2, line 11209: Good catch.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -7302,11 +7341,8 @@ internal Task WriteSqlVariantValue(object value, int length, int offset, TdsPars
 
                 case TdsEnums.SQLUNIQUEID:
                     {
-                        System.Guid guid = (System.Guid)value;
-                        byte[] b = guid.ToByteArray();
-
-                        Debug.Assert((length == b.Length) && (length == 16), "Invalid length for guid type in com+ object");
-                        stateObj.WriteByteArray(b, length, 0);
+                        WriteGuid((Guid)value, stateObj);
+                        Debug.Assert(length == 16, "Invalid length for guid type in com+ object");
```
Wraith2: You should probably invert these lines. Generally asserts come before the work they are preconditions for.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
edwardneal: I'd switched away from using a state parameter because it wasn't doing anything and it wasn't present in netcore. We pass `this` as the state, cast it back to a `TdsParser`, then don't use it anywhere in the lambda expression (which closes over another 11 variables); it's used once in the failure path (which closes over another variable, so can't currently be made static either.)

The halfway house doesn't really help anyone, but I'm happy enough to reintroduce the state parameter and use a tuple in both places if you'd prefer?

Wraith2: Ok, if it's not used downstream that's fine. The netore version is typically better than the netfx.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -1721,6 +1722,48 @@ internal void WriteUnsignedShort(ushort us, TdsParserStateObject stateObj)
             WriteShort((short)us, stateObj);
         }
 
+        //
+        // Writes a guid, either to a specific buffer or to the wire.
```
paulmedynski: I think this function exclusively writes to the given buffer.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -1721,6 +1722,48 @@ internal void WriteUnsignedShort(ushort us, TdsParserStateObject stateObj)
...
+            Debug.Assert(buffer.Length >= GUID_SIZE);
+#if NET
+            v.TryWriteBytes(buffer, bigEndian: false, out _);
+#else
+            byte[] guidBytes = v.ToByteArray();
+            guidBytes.AsSpan().CopyTo(buffer);
+#endif
+        }
+
+        private static void WriteGuid(in SqlGuid v, TdsParserStateObject stateObj)
```
paulmedynski, line 1760: And this one writes to the wire?

I think I'm confused because each of these 3 methods should have its own comment.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -5319,6 +5363,85 @@ private TdsOperationStatus TryCommonProcessMetaData(TdsParserStateObject stateOb
...
+            ushort shortLength;
+            byte byteLength;
+
+            // max byte size
+            TdsOperationStatus result = stateObj.TryReadUInt16(out shortLength);
+            if (result != TdsOperationStatus.Done)
+            {
+                return result;
+            }
+            metaData.length = shortLength;
```
paulmedynski, line 5432: Is it ok to partially update the metaData fields, but then return a non-Done result?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -6786,13 +6908,24 @@ internal TdsOperationStatus TryReadSqlValueInternal(SqlBuffer value, byte tdsTyp
...
+#if NET
                         Span<byte> b = stackalloc byte[GUID_SIZE];
+#else
+                        byte[] b = _tempGuidBytes;
+                        if (b is null)
+                        {
+                            b = new byte[GUID_SIZE];
+                        }
+#endif
                         result = stateObj.TryReadByteArray(b, length);
```
paulmedynski, line 6971: What happens if length > GUID_SIZE here? Should we explicitly check for that and throw something appropriate?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -6786,13 +6908,24 @@ internal TdsOperationStatus TryReadSqlValueInternal(SqlBuffer value, byte tdsTyp
...
                         if (result != TdsOperationStatus.Done)
                         {
                             return result;
                         }
-                        value.Guid = ConstructGuid(b);
+                        value.Guid = new Guid(b);
+#if NETFRAMEWORK
+                        _tempGuidBytes = b;
```
paulmedynski: You can avoid this `#if` by constructing into `_tempGuidBytes` above:

```
if (_tempGuidBytes is null)
{
  _tempGuidBytes = new byte[GUID_SIZE];
}
byte[] b = _tempGuidBytes;
```

Or just declare and pre-allocate `_tempGuidBytes`  inside an `#if NETFRAMEWORK` and take the 16-byte hit.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -6945,18 +6978,24 @@ internal TdsOperationStatus TryReadSqlValueInternal(SqlBuffer value, byte tdsTyp
                     {
                         Debug.Assert(length == GUID_SIZE, "invalid length for SqlGuid type!");
 
+#if NET
+                        Span<byte> b = stackalloc byte[GUID_SIZE];
+#else
                         byte[] b = _tempGuidBytes;
```
paulmedynski: Does it make sense to also guard the declaration of `_tempGuidBytes` with `#if NETFRAMEWORK`?  No sense in wasting the bytes on the unsued array pointer for .NET.  In both netfx and netcore files.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
Wraith2: I believe the intent of the comment is to say that the buffer parameter could either be an internal buffer such as one being used to buffer data before it is encrypted, or it could be _outBuff which is the "wire" buffer.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
Wraith2: It doesn't happen, see tds spec 2.2.5.5.1.7 and 2.2.5.4.3.
In practical terms the if you pass an array larger than 16 bytes to the Guid ctor the length is checked internally and it will throw.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
Wraith2: Yes. 

Or more precisely:
in compatibility mode it cannot occur 
in multiplexer mode until we re-enable async-continue it cannot occur 
in multiplexer mode when we _do_ enable async-continue (which I have a branch with the work done) it will not occur.

if an async operation is being attempted and there is not enough data the result of the attempt is a failure and the snapshot will be reset to the start point, a new packet will be awaited. Once the new packet arrives the operation will be attempted again from the start point and  will now succeed (because minimum packet size is 512 there is no way that 2 packets will not contain the metadata for a single field).
In between the first packet being processed and the second arriving there is an ongoing async operation and it is not legal to start a second operation (sync or async) at that time.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
edwardneal: The comments in this section of the file describe sets of methods, not individual methods. The two methods above this (`WriteShort` on L1703 and `WriteUnsignedShort` on L1720) are described on lines 1700-1702: "Takes a 16 bit short and writes it.". This is confusing and doesn't show up well on the diff, I'll change it.

In this context, this comment describes all three methods below: one writes a `Guid` to a specific buffer, one writes a `SqlGuid` to the wire, another writes a `Guid` to the wire.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
edwardneal: It doesn't actually exist in the netcore version, but when we're ready to merge TdsParser, the merge PR will that.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
edwardneal: I've added comments for each of these methods.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
paulmedynski: I guess I was concerned about what TryReadByteArray() does when length > b.Length.  Looking at the implementation, it appears that the call to b.Slice() (TdsParserStateObject.cs:1504) would fail if length > b.Length AND there are > length bytes to read from the packet.  It's hard to tell for sure though.  There are no unit tests for TryReadSqlValueInternal() that could verify this.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
paulmedynski: I'm concerned about the API promise of this method.  It has no documentation, so I don't know if callers should expect their SqlMetaDataPriv instance to be mutated when the return value != Done.

The call on line 4362 doesn't seem to care - the SqlMetaDataPriv instance is discarded when the return value != Done.

The call on line 5182 will leak the mutated fields back to the caller of TryProcessTypeInfo(), which fans out to several more call points.

I would like to see the expected/desired behaviour documented up the call chain so callers know what to expect.  I think the expectation is that mutation can occur regardless of return value, so adding docs to that effect is probably easiest.  The alternative to avoid mutations unless Done is returned would involve code changes and doesn't seem necessary.

Wraith2: My understanding of this code is that it should not be possible for anything to observe a partially read metadata class. History suggests that since this code has not changed in a long time and has not been observed as a source of bugs that whatever it is doing is "safe" within the limits of the current call sites.

This is all existing code and TdsParser is highly conserved because everyone is scared of it. The original authors may know their intent and why it was permissible to do some operations but we can't ask them. All we can do is maintain the contracts that exist. Expecting anyone to be able to completely justify the existing contracts doesn't seem reasonable.

I have the same problem with the expectation that existing methods need to have documentation written for them. It's a 20 year old codebase and at some time someone who was getting paid to work on it should have documented the method. I'm not going to take responsibility for the history of a method just because I need to tweak the code, that's the worst sort of git blame = ownership game.

paulmedynski: No problem, I've opened #3584 for myself to add the docs while they're fresh in my head.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -1749,6 +1746,48 @@ internal void WriteUnsignedShort(ushort us, TdsParserStateObject stateObj)
             WriteShort((short)us, stateObj);
         }
 
+        //
+        // Writes a guid, either to a specific buffer or to the wire.
+        //
+        private static void SerializeGuid(in Guid v, Span<byte> buffer)
```
mdaigle, line 1770: Would like to see some unit tests for these to ensure they behave the same between netfx and netcore.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -13841,8 +13841,7 @@ internal string TraceString()
                             _statistics == null ? bool.TrueString : bool.FalseString,
                             _statisticsIsInTransaction ? bool.TrueString : bool.FalseString,
                             _fPreserveTransaction ? bool.TrueString : bool.FalseString,
-                            _connHandler == null ? "(null)" : _connHandler.ConnectionOptions.MultiSubnetFailover.ToString((IFormatProvider)null),
-                            _connHandler == null ? "(null)" : _connHandler.ConnectionOptions.TransparentNetworkIPResolution.ToString((IFormatProvider)null));
```
mdaigle: Does this get reintroduced under an #if NETFRAMEWORK somewhere? This option is still valid for netfx

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -1156,7 +1156,8 @@ private PreLoginHandshakeStatus ConsumePreLoginHandshake(
 
                 // Validate Certificate if Trust Server Certificate=false and Encryption forced (EncryptionOptions.ON) from Server.
                 bool shouldValidateServerCert = (_encryptionOption == EncryptionOptions.ON && !trustServerCert) ||
-                    (_connHandler._accessTokenInBytes != null && !trustServerCert);
+                    ((_connHandler._accessTokenInBytes != null || _connHandler._accessTokenCallback != null) && !trustServerCert);
```
mdaigle, line 1163: This will be important to highlight in the release notes. Can you file an issue for this and I'll tag it for 7.0 and link this PR?

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
mdaigle: Is there any test coverage for this?

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
edwardneal: The option's still valid, but the format string doesn't actually include an identifier referencing this argument. We could either remove the unused argument (current option) or change the format string so that it uses the argument - I've got no strong preference either way.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
edwardneal: I'd expect it to be implicitly covered by the SqlBulkCopy tests which handle JSON data. The variables in question are on lines 137-141 ([link](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs#L137-L141)), and we can verify that they're the same length directly.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
edwardneal: I've just raised #3586.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
mdaigle: Let's add it to the format string. The trace message is misleading as-is because the parallel connectivity behavior can still be enabled (via TNIR) even if multisubnetfailover is disabled.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
edwardneal: The methods which write/serialize the numeric types are covered by the existing tests, but I've found and widened `ParametersTest.Test_WithGuidValue_ShouldReturnGuid` to handle SqlGuids (with and without values) to ensure these are covered.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs
edwardneal: Sounds good. I've changed the format string to clear up whether the setting is for MSF or TNIR, although netcore will just print a hardcoded `false` for the latter.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -1013,7 +1013,7 @@ private PreLoginHandshakeStatus ConsumePreLoginHandshake(
                         payloadOffset = payload[offset++] << 8 | payload[offset++];
                         payloadLength = payload[offset++] << 8 | payload[offset++];
 
-                        EncryptionOptions serverOption = (EncryptionOptions)payload[payloadOffset];
+                        EncryptionOptions serverOption = ((EncryptionOptions)payload[payloadOffset]) & EncryptionOptions.OPTIONS_MASK;
```
mdaigle, line 1020: I think I missed this before. Is the goal of the options mask to disable the CTAIP and CLIENT_CERT options?

mdaigle (dismissed):

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -6865,13 +6907,24 @@ internal TdsOperationStatus TryReadSqlValueInternal(SqlBuffer value, byte tdsTyp
                     {
                         Debug.Assert(length == GUID_SIZE, "invalid length for SqlGuid type!");
 
+#if NET
                         Span<byte> b = stackalloc byte[GUID_SIZE];
+#else
+                        byte[] b = _tempGuidBytes;
+                        if (b is null)
+                        {
+                            b = new byte[GUID_SIZE];
```
benrr101: nit: could be one line of `byte[] b = _tempGuildBytes ?? new byte[GUID_SIZE]` but if this is meant to match what exists in netfx, then that's ok.

benrr101 (dismissed):
All looks good to me!

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
edwardneal: Yes. netfx has these, netcore doesn't. We removed support for this in an earlier PR, but netfx retained a load of masking.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
edwardneal: This looks like an earlier version of TdsParser? netfx and netcore have the same implementation now:
```cs
#if NET
    Span<byte> b = stackalloc byte[GUID_SIZE];
#else
    byte[] b = (_tempGuidBytes ??= new byte[GUID_SIZE]);
#endif
    result = stateObj.TryReadByteArray(b, length);
    if (result != TdsOperationStatus.Done)
    {
        return result;
    }
    value.Guid = new Guid(b);
    break;
```

paulmedynski:
@edwardneal - I think a merge from main will help clear up some of these CI failures.

edwardneal:
Thanks @paulmedynski - I've just merged, can you re-run CI please?

paulmedynski:
Please wait for #3607 to merge, and then re-merge main.  That should fix the intermittent macOS failures.

PR #3567, opened 2025-08-14
[WIP] Create Abstractions package

paulmedynski:
_**Work in Progress - This PR is a draft while I get the CI machinery working.**_

## Overview

As part of the Azure split work, a new Abstractions package is being created that contains types shared between MDS and its future extensions (Azure will be the first). This PR creates that package (with no meaningful content) to setup the MDS dependency, and get testing and CI setup accordingly.

I'm also experimenting with simplified .slnx files and some project directory structure changes.

## Issues

The first step towards #1108.

## Testing

- Unit tests for the sample class to prove that CI can run them successfully.
- There are no MDS tests to prove out the new dependency yet. Those will naturally happen when the first real bit of Azure functionality is moved out into the new Azure extension package.


eng/pipelines/dotnet-sqlclient-ci-core.yml
```diff
@@ -356,47 +391,15 @@ stages:
...
               AADServicePrincipalSecret: $(AADServicePrincipalSecret)
+            AADServicePrincipalId: $(AADServicePrincipalId)
             AzureKeyVaultUrl: $(AzureKeyVaultUrl)
             AzureKeyVaultTenantId: $(AzureKeyVaultTenantId)
             SupportsIntegratedSecurity: false
             UserManagedIdentityClientId: $(UserManagedIdentityClientId_eastus)
             LocalDbAppName: $(LocalDbAppName)
             LocalDbSharedInstanceName: $(LocalDbSharedInstanceName)
 
-        ${{ if eq(variables['system.pullRequest.isFork'], 'False') }}: # only run enclave jobs if the password is available
```
paulmedynski, line 396: My YAML linter was complaining about duplicate keys because these ${{if ...} lines are the same. Similar changes below for the enclave test configurations.

tools/props/Versions.props
```diff
@@ -1,29 +1,61 @@
...
+    -->
+    <NugetPackageVersion Condition="'$(NugetPackageVersion)' == ''">$(MdsVersionDefault).$(BuildNumber)-dev</NugetPackageVersion>
+
+    <!--
+      GOTCHA: These two Assembly properties must appear in the first
+      <PropertyGroup> in this file due to hardcoded XML processing in:
+
+      eng/pipelines/common/templates/jobs/validate-signed-package-job.yml
+    -->
+    <AssemblyFileVersion Condition="'$(AssemblyFileVersion)' == ''">$(MdsVersionDefault).$(BuildNumber)</AssemblyFileVersion>
```
paulmedynski, line 28: Minor tweak here - I added $(BuildNumber) - not sure if that's helpful, but it matches how the Extensions package does things.

It used to append `-dev` instead of the build number.  I think this will only affect manual builds.

paulmedynski (commented):
Adding comentary for reviewers.

I will be renaming the Extensions package to Abstractions.  That round of changes is next.

tools/props/Versions.props
```diff
@@ -1,29 +1,61 @@
...
-      number.  All other version parts must be set to 0.
+      We pin the AssemblyVersion of our package to its current Major version
+      number, with all remaining parts set to 0.
     -->
+    <ExtensionsPackageAssemblyVersion>$(ExtensionsPackageMajorVersion).0.0.0</ExtensionsPackageAssemblyVersion>
+  </PropertyGroup>
+
+  <!-- MDS Package Properties -->
+  <PropertyGroup>
+    <MdsVersionDefault>6.1.0</MdsVersionDefault>
```
cheenamalhotra: Also change this to 7.0.0

paulmedynski: I created #3570 , since we probably want it on main ASAP.

eng/pipelines/dotnet-sqlclient-ci-package-reference-pipeline.yml
```diff
@@ -62,6 +62,11 @@ parameters: # parameters are shown up in ADO UI in a build queue time
   type: object
   default: [1, 2, 3]
 
+- name: abstractionsPackageVersion
+  displayName: Abstractions Package Version Override
+  type: string
+  default: 7.0.0
```
paulmedynski: Should be 1.0.0, here and the next file.

src/Microsoft.Data.SqlClient.Extensions/Abstractions/README.md
```diff
@@ -0,0 +1,285 @@
...
+```
+
+Additionally, applications will need to instruct MDS to use the external Azure
+feature implementations via the appropriate APIs at runtime:
+
+- Authentication: [SqlAuthenticationProvider](https://learn.microsoft.com/en-us/dotnet/api/microsoft.data.sqlclient.sqlauthenticationprovider?view=sqlclient-dotnet-core-6.0)
+- Attestation: _**New API will be exposed.**_
+- Key Valut: [SqlColumnEncryptionKeyStoreProvider](https://learn.microsoft.com/en-us/dotnet/api/microsoft.data.sqlclient.sqlcolumnencryptionkeystoreprovider?view=sqlclient-dotnet-core-6.0)
+
+## Versioning Strategy
```
paulmedynski, line 154: I believe this aligns with our general extension strategy.  Let's discuss if anyone has concerns.

src/Microsoft.Data.SqlClient/NuGet.config
```diff
@@ -0,0 +1,13 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <!--
+      During development, the Extensions package that MDS depends on may only
```
paulmedynski: Extensions -> Abstractions; here and below.

src/Directory.Build.props
```diff
@@ -20,6 +20,21 @@
       > msbuild -p:configuration=Release
     -->
     <ReferenceType Condition="$(ReferenceType) == ''">Project</ReferenceType>
+
+    <!--
+      We embed debug symbols in some of our application (i.e. non-symbol) NuGet
+      packages.
+
+      TODO: Why are we doing this?
```
paulmedynski, line 28: Thoughts?  Why do we want symbols in our application packages, when we're publishing separate symbols packages anyway?

BUILDGUIDE.md
```diff
@@ -16,40 +16,47 @@ Once the environment is setup properly, execute the desired set of commands belo
 
 ### Targets
 
+The following build targets are defined in `build.proj`:
+
 |Target|Description|
 |-|-|
 |`BuildAllConfigurations`|Default target. Builds the .NET Framework and .NET drivers for all target frameworks and operating systems.|
+|`BuildExtensionsPackage`|Restore, build, and pack the Extensions package, publishing the resulting NuGet into `packages/`.|
```
paulmedynski: Extensions -> Abstrations

Here and throughout this file.

NuGet.config
```diff
@@ -2,10 +2,13 @@
 <configuration>
   <packageSources>
     <clear />
-    <add key="sqlclient" value="https://sqlclientdrivers.pkgs.visualstudio.com/public/_packaging/sqlclient/nuget/v3/index.json" />
+    <!-- We use a curated, vetted ADO feed for our dependencies. -->
+    <add key="curated" value="https://sqlclientdrivers.pkgs.visualstudio.com/public/_packaging/sqlclient/nuget/v3/index.json" />
```
benrr101: "governed" might be the more accurate name here ... but it made just as much sense to me to call it "sqlclient" since it comes from the sqlclient feed in ado

build.proj
```diff
@@ -91,6 +92,32 @@
...
+    The Abstractions package must be built and packed into the packages/
+    directory before the MDS projects can be restored.
+
+    We cannot make the driver targets depend on this one because of the way
+    MSBuild restores projects in "solution mode".  It will attempt to restore
+    all targets in the graph before building any of them, which fails because
+    MDS depends on the Abstractions package being built and packed into packages/
+    first.
+  -->
+  <Target Name="BuildAbstractionsPackage">
```
benrr101: I'm not confident about this change. The structure we currently have is set up more for separate build, test, and package steps that are meant to be run independently (primarily by pipelines). Further, I'm not sure about the explanation about about msbuild not being able to restore mds b/c of dependencies on abstractions. Afaik, eg, the tests use the ReferenceType parameter to switch between a reference to MDS as a package and a reference to MDS as the project, and defaulting to project reference. Is that not something we want to continue to use here?

I just don't want to deviate too far from existing build.proj constructs right now. I think we *need* to make changes to it, but I'm not sure this is the right time to introduce a new/another paradigm in the same file.

build.proj
```diff
@@ -346,13 +373,17 @@
...
-    <RemoveDir Directories='$([System.IO.Directory]::GetDirectories(".","packages", SearchOption.AllDirectories))' />
-    <RemoveDir Directories='$([System.IO.Directory]::GetDirectories(".",".nuget", SearchOption.AllDirectories))' />
+    <RemoveDir Directories='$([System.IO.Directory]::GetDirectories(".", "artifacts", SearchOption.AllDirectories))' />
+    <RemoveDir Directories='$([System.IO.Directory]::GetDirectories(".", "bin", SearchOption.AllDirectories))' />
+    <RemoveDir Directories='$([System.IO.Directory]::GetDirectories(".", "obj", SearchOption.AllDirectories))' />
+    <RemoveDir Directories='$([System.IO.Directory]::GetDirectories(".", ".nuget", SearchOption.AllDirectories))' />
+  </Target>
+
+  <!-- Clean all outputs, including NuGet packages. -->
+  <Target Name="CleanAll" DependsOnTargets="Clean">
```
benrr101: My understanding of the packages folder (as per your recent changes for the stress tests) is that this is a folder for local nuget package overrides. I suppose the question is who the target of this target is:
* if Developers, I'd ask - how often would a developer want to obliterate their local cache of nuget packages and not just do it manually? (I've never even touched the clean target because I've been taught that clean doesn't do anything useful, so if I need a clean I just resort to `git clean -xfd -e .idea`) My ask would be to probably just remove this "cleanall" target. Or at least not chain them together. It just seems to add more tangles to build.proj that we don't need, imo.
* if Pipelines, I'd ask - do we actually run these targets in the pipelines? And do we touch the local packages directory/use it during pipelines? I think my ask here would be to remove all the clean targets.

eng/pipelines/common/templates/jobs/ci-build-nugets-job.yml
```diff
@@ -28,6 +32,11 @@ parameters:
     type: stepList
     default: []
 
+  - name: abstractionsPackageVersion
```
benrr101, line 42: I'm not sure about this. Package versions are constructed from the major.minor.patch(-preview) variables, but the assembly versions are constructed from major.minorpatch.buildnumber (or something like that) variables. I suspect having a single parameter for this will not be sufficient.

The idea I've had is that versions for the package/assembly should come from some centralized location in the code. My ideal would be to store the components in the csproj. This would mostly be because it's easier to get the pipeline to read an xml file than it would be to get the csproj to pull in info from some other file.

Additionally, this parameter only applies to the CI pipeline. In that situation, it seems unnecessary to have an override parameter since we generally only ever run this pipeline via automation (ie, no overrides). And we shouldn't be trying to force a version if we ever ran the CI pipeline manually (ie, does it matter what version it spits out?).

eng/pipelines/common/templates/jobs/ci-build-nugets-job.yml
```diff
@@ -12,9 +12,13 @@ parameters:
     type: string
     default: ADO-MMS22-SQL19
 
-  - name: artifactName
+  - name: abstractionsArtifactName
     type: string
-    default: Artifacts
+    default: Abstractions.Artifact
+
+  - name: mdsArtifactName
```
benrr101, line 26: Gonna nit pick here, but can we please stick to alphabetical order for parameters? Unless we still plan to throw this pipeline away (which I support), let's at least try to keep it in order.

eng/pipelines/common/templates/steps/build-all-tests-step.yml
```diff
@@ -32,6 +32,11 @@ parameters:
   - name: testSet
     type: string
 
+  - name: abstractionsPackageVersion
```
benrr101: At this point, I see more of the picture of how this parameter feeds in here. I think I'd have to ask that we back this up to a higher level, like we do for the official builds. Ie, determine the complete package version at the root of the pipeline and feed it further down anywhere it's needed. At this level, it's not an override per se, it's *the* package version that needs to be referenced. But it's not the *complete* package version because we're still tacking on the build number at the end. To me this is quite confusing 😬

eng/pipelines/common/templates/steps/build-all-tests-step.yml
```diff
@@ -40,7 +45,7 @@ steps:
       solution: build.proj
       platform: '${{parameters.platform }}'
       configuration: '${{parameters.configuration }}'
-      msbuildArguments: '-t:BuildTestsNetFx -p:TF=${{parameters.targetFramework }} -p:TestSet=${{parameters.testSet }} -p:ReferenceType=${{parameters.referenceType }} -p:TestMicrosoftDataSqlClientVersion=${{parameters.nugetPackageVersion }}'
+      msbuildArguments: '-t:BuildTestsNetFx -p:TF=${{parameters.targetFramework }} -p:TestSet=${{parameters.testSet }} -p:ReferenceType=${{parameters.referenceType }} -p:TestMicrosoftDataSqlClientVersion=${{parameters.nugetPackageVersion }} -p:AbstractionsPackageVersion=${{ parameters.abstractionsPackageVersion }}.$(Build.BuildNumber)'
```
benrr101: Probably a bigger question than for just this PR, but why do we need the package version if we're build in in Project mode?

eng/pipelines/common/templates/steps/update-nuget-config-local-feed-step.yml
```diff
@@ -51,7 +56,7 @@ steps:
   inputs:
     command: 'custom'
     custom: 'msbuild'
-    arguments: 'build.proj -t:restore'
+    arguments: 'build.proj -t:restore -p:AbstractionsPackageVersion=${{parameters.abstractionsPackageVersion}}.$(buildNumber)'
```
benrr101: This is another case where we're constructing the package version on the fly, and we should really be doing it at the beginning of the pipeline.

eng/pipelines/dotnet-sqlclient-ci-core.yml
```diff
@@ -89,20 +94,39 @@ parameters:
 variables:
   - template: libraries/ci-build-variables.yml@self
   
-  - name: artifactName
-    value: Artifacts
+  - name: abstractionsArtifactName
```
benrr101, line 95: I'm kinda thinking these should be in ci-build-variables

eng/pipelines/dotnet-sqlclient-ci-core.yml
```diff
@@ -89,20 +94,39 @@ parameters:
...
+
+  - name: mdsArtifactName
+    value: MDS.Artifact
 
   - name: defaultHostedPoolName
     value: 'Azure Pipelines'
 
 stages:
+
+  # Build the Abstractions package, and publish it to the pipeline artifacts
```
benrr101, line 106: Should we be doing this for both project and package CI builds? It seems unnecessary in project builds

build.proj
benrr101: Secretly I was hoping we could have a nice little acronym for all these packages we release.
* MDS - Microsoft.Data.SqlClient
* MSS - Microsoft.SqlServer.Server
* AKV - ...AzureKeyVaultProvider
* _MDSEA - Microsoft.Data.SqlClient.Extensions.Abstractions?_

Then this becomes `BuildMdsea` and fits nicely with `BuildAkv` and eventual `BuildMds` `BuildMss` etc. Idk, I just kinda feel like "AbstractionsPackage" is redundant and even "Abstractions" is long. But that's just me trying to neatly group things together into nice little boxes.

benrr101: Heck, I'd take TAP - The Abstractions Package, mostly for the memes

eng/pipelines/stages/build-abstractions-package-ci-stage.yml
```diff
@@ -0,0 +1,351 @@
...
+#
+# The packages are published to pipeline artifacts with the name specified by
+# the 'artifactName' parameter.
+#
+# This template defines a stage named 'build_abstractions_package_stage' that
+# can be depended on by downstream stages.
+
+parameters:
+  # The type of build to produce (Release or Debug)
+  - name: buildConfiguration
```
benrr101, line 38: for new things, can we definitely have alphabetical naming?

eng/pipelines/stages/build-abstractions-package-ci-stage.yml
```diff
@@ -0,0 +1,351 @@
...
+    default: Abstractions.Artifact
+
+  # The list of .NET runtimes to test against.
+  - name: netTestRuntimes
+    displayName: .NET Test Runtimes
+    type: object
+    default: [net8.0, net9.0]
+
+  # The list of .NET Framework runtimes to test against.
+  - name: netFrameworkTestRuntimes
```
benrr101: Not expecting to solve this in this PR, but we should have this list centralized somewhere rather than here, in a directory.build.props, in some csproj files, etc...

eng/pipelines/stages/build-abstractions-package-ci-stage.yml
```diff
@@ -0,0 +1,351 @@
...
+    type: string
+    default: $(Build.BuildNumber)
+
+  # The name of the pipeline artifact to publish.
+  - name: artifactName
+    displayName: Pipeline Artifact Name
+    type: string
+    default: Abstractions.Artifact
+
+  # The list of .NET runtimes to test against.
```
benrr101: It should probably be noted that both this and the next one are really only being made parameters is not because we want to override them, but because variables cannot be object type.

eng/pipelines/stages/build-abstractions-package-ci-stage.yml
```diff
@@ -0,0 +1,351 @@
...
+      # dotnet CLI arguments common to all commands.
+      - name: commonArguments
+        value: >-
+          --verbosity ${{parameters.verbosity}}
+          --artifacts-path $(dotnetArtifactsDir)
+
+      # dotnet CLI arguments for build/test/pack commands
+      - name: buildArguments
+        value: >-
+          $(commonArguments)
```
benrr101: it would be really great if we can refer to variables using the same ${{ variables.commonArguments }} format, so it becomes obvious where they're coming from.

eng/pipelines/stages/build-abstractions-package-ci-stage.yml
```diff
@@ -0,0 +1,351 @@
...
+          --configuration ${{parameters.buildConfiguration}}
```
benrr101: I know it works to omit the spaces, but the docs for ado pipeline yaml always includes the spaces
ie `${{ parameters.buildConfiguration }}`

eng/pipelines/stages/build-abstractions-package-ci-stage.yml
```diff
@@ -0,0 +1,351 @@
...
+
+    - job: build_abstractions_package_job_linux
+      displayName: '[Linux] Build Abstractions Package'
+      pool:
+        name: Azure Pipelines
+        vmImage: ubuntu-latest
+
+      steps:
+
+      - task: UseDotNet@2
```
benrr101: can we indent these one more step over

src/Microsoft.Data.SqlClient.Extensions/Abstractions/Abstractions.slnx
```diff
@@ -0,0 +1,8 @@
+<Solution>
```
benrr101, line 1: I'm just gonna say we really shouldn't have >1 solution in this repository. The solution is the developer entry point into all the interlinked projects. Creating another one really confuses things and will likely turn into technical debt. I think we should either 1) work our automated builds through msbuild and targets in build.proj (at least until we can come up with a better alternative to it), or 2) make the dotnet commands work with the csproj files (I'm certain they can be made to do so!)

src/Microsoft.Data.SqlClient/NuGet.config
```diff
@@ -0,0 +1,13 @@
...
+<configuration>
+  <packageSources>
+    <!--
+      During development, the Abstractions package that MDS depends on may only
+      be available locally.
+
+      This directory must exist before MDS restore will succeed.  Building and
+      packing the Abstractions package will create it.
+    -->
+    <add key="local" value="../../packages/" />
```
benrr101: If you're adding this make sure to add an empty packages folder and put a .gitkeep file in there. Otherwise, as you noted restore fails.

However, I'm a bit skeptical of forcing this into the repo. Imo, adding more nuget.config files complicates the tree and can cause difficult to diagnose/silent errors during restore. This seems to me like something a developer should be able to do on their own in the handful of instances where its necessary, but not something that needs to be baked into the repo.

tools/props/Versions.props
```diff
@@ -1,29 +1,62 @@
...
+    <AbstractionsPackageMajorVersion>1</AbstractionsPackageMajorVersion>
 
     <!--
-      GOTCHA: These two properties must appear in the first <PropertyGroup> in
-      this file due to hardcoded XML processing in:
+      The default Abstractions package version.
 
-      eng/pipelines/common/templates/jobs/validate-signed-package-job.yml
+      This may be overridden by specifying a value on the command-line:
+
```
benrr101: super nit: sparsely populated comments like these seems ... idk, um, unnecessarily un-dense. We could get rid of this extra line and it becomes just a bit more readable imo.

benrr101 (changes requested):
I'm omitting all my comments about directory structure as per offline discussion. But I think we want to be really careful and thoughtful with how we set up pipelines, build targets, and project structure.

src/Microsoft.Data.SqlClient.Extensions/Abstractions/README.md
cheenamalhotra: I see spelling error in Vault above.

src/Directory.Build.props
cheenamalhotra: We should be publishing symbols to NuGet.org (in .snupkg) and to Microsoft internal/public symbols servers. I don't think we need anywhere else. Are there still symbols included elsewhere?

paulmedynski: It looks like we're telling dotnet/MSBuild to generate pdbs for our .nuget packages, but then the .nuspec excludes them.  I think we can just stop the former, let the latter continue as-is until we get rid of .nuspec files altogether.

NuGet.config
paulmedynski: The term "sqlclient" is already too overloaded and doesn't explain the purpose of the feed.  I'll expand the comment to describe why we're not just using the public nuget.org feed like one would expect.

build.proj
paulmedynski: I can introduce the concept of package vs project reference building of Abstractions.  I was hoping to avoid it entirely, since IMO that conecpt needs to be removed entirely from our builds.  But if you'd rather see that latter change as a separate body of work, I can create the old ReferenceType style here for now.  I have my doubts that build.proj will successfully build all target frameworks at once though - it currently fails to do that for our tests and AKV.

build.proj
paulmedynski: I expect my build tooling to be able to remove all files it creates.  That's what the existing "Clean" target did, and it is used in our YAML pipelines.  With the new Abstractions package dependency, the old "Clean" behaviour would nuke the pipeline artifacts and make pipeline building fail.  I removed deletion of packages from "Clean" to keep the YAML happy, and added "CleanAll" for developers.

If I revert to using the ReferenceType stuff, as noted above, then maybe this is all moot.

eng/pipelines/common/templates/jobs/ci-build-nugets-job.yml
paulmedynski: Yeah, my brain thinks of these like function arguments, which you wouldn't typically insert one into the middle.  Since these are all passed by name, the declared order doesn't matter, so I will fix.

eng/pipelines/common/templates/steps/build-all-tests-step.yml
paulmedynski: We don't, but that would be messy to if-statement out here.  Specifying a parameter that isn't used half of the time was probably the lesser evil.

eng/pipelines/dotnet-sqlclient-ci-core.yml
paulmedynski: I might like using variables if they were always scoped to identify where they're defined.  Our current approach to have variables scattered across at least 3 different definition spaces, and then just used like `$(Major)` 5 levels deep in a template file doesn't fly for me.  I think I'd rather a template declare all of the varying values it supports as formal parameters.

eng/pipelines/dotnet-sqlclient-ci-core.yml
paulmedynski: I purposely didn't want to support having the Abstractions project included as a project reference anywhere but its own unit tests project.  I think only unit tests should be using their sibling project as a project reference.  All other references should be package.  As discussed earlier, maybe now isn't the time to test this approach.  But I think we're wasting a lot of CI time building both ways, and it doesn't really test anything.

build.proj
paulmedynski: My goal is to be able to type `dotnet build` at the repo root and everything gets built.  I added this target to build.proj to satisfy folks to use it.  I won't use it, and neither will any CI or pipelines.  Maybe we don't need it?

eng/pipelines/stages/build-abstractions-package-ci-stage.yml
paulmedynski: And because I don't like variables that happen to be defined just because of how/where the template was called from.

eng/pipelines/stages/build-abstractions-package-ci-stage.yml
paulmedynski: I'm not sure that would work here, since `${{ }}` expansion happens before `$()` variables are defined.

eng/pipelines/stages/build-abstractions-package-ci-stage.yml
paulmedynski: Huh you're right, that is strange indeed.  Seems like someone decided that `${{foo}}` was too hard to read, but `$(foo)` was not.  Sigh

src/Microsoft.Data.SqlClient.Extensions/Abstractions/Abstractions.slnx
paulmedynski: I want to be able to go into `src/Microsoft.Data.SqlClient.Extensions.Abstractions/` and type `dotnet build` and have it build all of the projects.  Maybe there's another way to make that work without a solution?

build.proj
benrr101: Package vs Project build stuff was decided to be deferred for later offline

eng/pipelines/stages/build-abstractions-package-ci-stage.yml
benrr101: Agreed, didn't realize this variable was built on runtime-only variables

eng/pipelines/dotnet-sqlclient-ci-core.yml
benrr101: Decided offline that these should be cleaned up in a separate bundle of work

build.proj
benrr101: Decided names might want to be standardized via team meeting

benrr101 (changes requested):
Awaiting the rest of the changes, but we're definitely making progress

eng/pipelines/common/templates/jobs/ci-build-nugets-job.yml
```diff
@@ -44,34 +59,44 @@ jobs:
...
       nuspecPath: 'tools/specs/Microsoft.Data.SqlClient.nuspec'
       OutputDirectory: $(packagePath)
       generateSymbolsPackage: false
-      displayName: 'Generate NuGet package M.D.SqlClient'
+      displayName: 'Create MDS NuGet Package'
 
   - template: ../steps/generate-nuget-package-step.yml@self
     parameters:
-      NugetPackageVersion: $(NugetPackageVersion)
+      mdsPackageVersion: $(mdsPackageVersion)
```
paulmedynski: Should this be/have been specifying the AKV package version?

eng/pipelines/common/templates/jobs/build-signed-package-job.yml
```diff
@@ -27,8 +27,8 @@ jobs:
   variables:
   - template: ../../../libraries/variables.yml@self
   - ${{ if parameters.isPreview }}:
-    - name: NugetPackageVersion
```
paulmedynski, line 30: I found NugetPackageVersion confusing - which nuget package?  So I gave them names specific to the package.  I didn't rename any of these yml files to match such changes though.  I figure that's more trouble than it's worth, and we intend to re-factor the whole works here soon anyway.

eng/pipelines/common/templates/jobs/ci-run-tests-job.yml
```diff
@@ -4,12 +4,37 @@
 # See the LICENSE file in the project root for more information.                #
 #################################################################################
 parameters:
+  - name: abstractionsArtifactName
+    type: string
+
+  - name: abstractionsPackageVersion
+    type: string
+
+  - name: referenceType
```
paulmedynski: This was called "buildType", and I moved it into alphabetical order.  But then I renamed it - doh!

eng/pipelines/common/templates/steps/update-nuget-config-local-feed-step.yml
```diff
@@ -45,39 +41,3 @@ steps:
       # Display the content of the NuGet.config file
       Get-Content -Path "NuGet.config"
     displayName: 'Read NuGet.config [debug]'
-
-- task: DotNetCoreCLI@2
```
paulmedynski, line 49: Not sure why this was here, so I removed it.

eng/pipelines/common/templates/steps/update-nuget-config-local-feed-step.yml
```diff
@@ -45,39 +41,3 @@ steps:
...
-  displayName: 'Restore NuGets'
-  inputs:
-    command: 'custom'
-    custom: 'msbuild'
-    arguments: 'build.proj -t:restore'
-    feedsToUse: 'select'
-
-- powershell: |
```
paulmedynski, line 57: AKV .csproj already has conditional references depending on Project vs Package, so this wasn't necessary.

eng/pipelines/jobs/build-akv-official-job.yml
```diff
@@ -90,7 +87,7 @@ jobs:
             parameters:
                 assemblyFileVersion: '${{ parameters.assemblyFileVersion }}'
                 buildConfiguration: '${{ parameters.buildConfiguration }}'
-                mdsPackageVersion: '${{ parameters.mdsPackageVersion }}'
```
paulmedynski, line 93: I'm not sure about these 2 changes (here and line 187).  I will verify...

eng/pipelines/libraries/ci-build-variables.yml
```diff
@@ -15,8 +14,12 @@ variables:
     value: 'net8.0'
   - name: SQLTarget
     value: 'localhost'
-  - name: NugetPackageVersion
-    value: $(Major).$(Minor)$(Patch)-pull.1$(buildnumber)
```
paulmedynski, line 19: Moved these CI versioning variables from the ADO CI Packaging library to here.  It was confusing to have them separate like that.  We can decide where they will ultimately reside during the re-factoring work.

eng/pipelines/libraries/ci-build-variables.yml
```diff
@@ -15,8 +14,12 @@ variables:
...
+  - name: abstractionsPackageVersion
+    value: 1.0.0.$(buildNumber)-ci
+  - name: akvPackageVersion
```
paulmedynski, line 19: I created an AKV version variable for CI pipelines, which seemed to be missing.

I also added a "-ci" suffix to all of the package versions to help identify their provenance and differentiate them from developer and official builds.  The old MDS suffix of "-pull" didn't really identify _where_ the package came from, and wasn't necessarily accurate since you can manually trigger CI pipeline runs without a pull request involved.

eng/pipelines/steps/compound-build-akv-step.yml
```diff
@@ -46,7 +46,7 @@ steps:
           msbuildArguments: >-
               -t:BuildAkv
               -p:AssemblyFileVersion=${{ parameters.assemblyFileVersion }}
-              -p:NugetPackageVersion=${{ parameters.mdsPackageVersion }}
```
paulmedynski, line 49: Verify this is the correct version to specify.

eng/pipelines/steps/roslyn-analyzers-akv-step.yml
```diff
@@ -25,7 +25,7 @@ steps:
               $(REPO_ROOT)/build.proj
               -t:BuildAkv
               -p:Configuration=${{ parameters.buildConfiguration }}
-              -p:NugetPackageVersion=${{ parameters.mdsPackageVersion }}
```
paulmedynski, line 28: Verify

eng/pipelines/dotnet-sqlclient-ci-core.yml
```diff
@@ -89,65 +92,94 @@ parameters:
...
+      ${{if eq(parameters.debug, 'true')}}:
+        verbosity: diagnostic
+
+  # Build MDS and its NuGet packages.
+  - stage: build_mds_akv_packages_stage
+    displayName: 'Build MDS & AKV Packages'
+
+    # When building MDS via packages, we must depend on the Abstractions 
+    # package.
+    ${{ if eq(parameters.referenceType, 'Package') }}:
```
paulmedynski, line 122: I'm starting to think that the Reference Types are sufficiently different that maybe they shouldn't share stages/jobs/steps, as evidenced by needing to pass the referenceType and abstractionsPackageVersion values around everywhere, and only use them some of the time.

eng/pipelines/dotnet-sqlclient-ci-core.yml
```diff
@@ -89,65 +92,94 @@ parameters:
...
                 SNIVersion: ${{parameters.SNIVersion}}
                 SNIValidationFeed: ${{parameters.SNIValidationFeed}}
 
+  # Run the stress tests, if desired.
   - ${{ if eq(parameters.enableStressTests, true) }}:
     - template: stages/stress-tests-ci-stage.yml@self
       parameters:
         buildConfiguration: ${{ parameters.buildConfiguration }}
-        dependsOn: [build_nugets]
+        dependsOn: [build_mds_packages_job]
```
paulmedynski: Is this the correct job name?

eng/pipelines/dotnet-sqlclient-ci-core.yml
```diff
@@ -89,65 +92,94 @@ parameters:
...
+          - build_abstractions_package_stage
+          - build_mds_akv_packages_stage
+
+      prebuildSteps: # steps to run prior to building and running tests on each job
+        - ${{if ne(parameters.SNIVersion, '')}}:
           - template: common/templates/steps/override-sni-version.yml@self
             parameters:
               SNIVersion: ${{parameters.SNIVersion}}
               SNIValidationFeed: ${{parameters.SNIValidationFeed}}
-          - template: common/templates/steps/ci-prebuild-step.yml@self
```
paulmedynski, line 137: This and the else block are identical, so I pulled them out.

src/Microsoft.Data.SqlClient.Extensions/Abstractions/src/Abstractions.csproj
```diff
@@ -0,0 +1,114 @@
...
+    <!--
+      We accept two external properties to control versioning:
+
+        AbstractionsPackageVersion:
+          - Used as the NuGet package version, both in the package file name
+            and in the package metadata.
+          - Used as the base value of the assembly attribute
+            'AssemblyInformationalVersion' generated by MSBuild.
+          - If not specified, defaults to:
+              $(AbstractionsPackageMajorVersion).0.0.$(BuildNumber)-dev.
```
paulmedynski: DefaultMajorVersion

tools/targets/add-ons/GenerateAKVProviderNugetPackage.targets
```diff
@@ -2,11 +2,11 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Target Name="GenerateAKVProviderNugetPackage" Condition="'$(GenerateNuget)' == 'true'">
```
paulmedynski: Rename this target to GenerateAkvPackage.

NuGet.config
```diff
@@ -1,11 +1,30 @@
...
+
+    <!--
+      We do not use the public nuget.org feed.  Instead, we use a governed
+      feed maintained by our SqlClientDrivers ADO organization.  Packages must
+      be individually vetted and added to this feed manually before they will
+      be available for consumption by our builds.
+
+      https://sqlclientdrivers.visualstudio.com/public/_artifacts/feed/sqlclient
+    -->
+    <add key="curated" value="https://sqlclientdrivers.pkgs.visualstudio.com/public/_packaging/sqlclient/nuget/v3/index.json" />
```
paulmedynski: Rename to governed.

eng/pipelines/common/templates/jobs/build-signed-package-job.yml
benrr101: Makes sense to me. I appreciate that kind of verbosity, and it was something I aimed to resolve in the AKV provider pipeline. Needs to be spread to other pipelines :D

benrr101 (dismissed):
Yeah, I think this is a good improvement over the OG PR. We'll mess with it some more as we get time for it 👍

paulmedynski:
Abandoning in favour of #3626 and #3628.

PR #3587, opened 2025-08-29
Fix Debug Assertion in Connection Pool Due to Double Deactivation

ShreyaLaxminarayan:
Problem:
During execution of the test TestAutoEnlistment_TxScopeComplete and TestAutoEnlistment_TxScopeNonComplete in Debug mode,a debug assertion fails on CoreCLR: 
```Method Debug.Fail failed with 'activated multiple times?```

This occurs inside the DeactivateConnection method when returning a connection to the pool after a transaction completes.This assertion failure indicates that the internal _activateCount was already 0 when decrement was attempted again, deactivating the connection more than once.

Root Cause:
In the TransactionEnded function, after a transaction completes, the connection involved is returned to the general pool via:

```Pool.PutObjectFromTransactedPool(transactedObject);```

During this process, DeactivateConnection() is called again. However, the connection had already been deactivated earlier as part of the transaction's enlistment cleanup.This results in DeactivateConnection() being called twice without a corresponding call to ActivateConnection() in between, which causes _activateCount to decrement to -1, triggering the debug assertion.

Fix:
We prevent the second DeactivateConnection to decrement the _activateCount counter, by passing the parameter "false" to DeactivateConnection.

cc: @uweigand @giritrivedi @saitama951

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbConnectionInternal.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbConnectionInternal.cs
@@ -513,14 +513,16 @@ namespace Microsoft.Data.ProviderBase
             }
         }
 
-        internal void DeactivateConnection()
+        internal void DeactivateConnection(bool callDecrement=true)
         {
             // Internal method called from the connection pooler so we don't expose
             // the Deactivate method publicly.
             SqlClientEventSource.Log.TryPoolerTraceEvent("<prov.DbConnectionInternal.DeactivateConnection|RES|INFO|CPOOL> {0}, Deactivating", ObjectID);
 
             #if DEBUG
-            int activateCount = Interlocked.Decrement(ref _activateCount);
+            int activateCount = _activateCount;
+            if(callDecrement)
+                activateCount = Interlocked.Decrement(ref _activateCount);
             Debug.Assert(activateCount == 0, "activated multiple times?");
             #endif
 
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbConnectionInternal.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbConnectionInternal.cs
@@ -513,17 +513,26 @@ namespace Microsoft.Data.ProviderBase
             }
         }
 
-        internal void DeactivateConnection(bool callDecrement=true)
+        internal void DeactivateConnection()
         {
             // Internal method called from the connection pooler so we don't expose
             // the Deactivate method publicly.
             SqlClientEventSource.Log.TryPoolerTraceEvent("<prov.DbConnectionInternal.DeactivateConnection|RES|INFO|CPOOL> {0}, Deactivating", ObjectID);
 
             #if DEBUG
-            int activateCount = _activateCount;
-            if(callDecrement)
-                activateCount = Interlocked.Decrement(ref _activateCount);
-            Debug.Assert(activateCount == 0, "activated multiple times?");
+            int origCount, newCount;
+            do
+            {
+                origCount = _activateCount;
+
+                if (origCount == 0)
+                {
+                  break;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by shreyarao4** [comment]
@dotnet-policy-service agree company="IBM"

**Comment by ShreyaLaxminarayan** [comment]
@mdaigle does this look reasonable?

File: src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbConnectionInternal.cs [hunk 1]
```diff
@@ -513,14 +513,16 @@ internal virtual void CloseConnection(DbConnection owningObject, SqlConnectionFa
             }
         }
 
-        internal void DeactivateConnection()
+        internal void DeactivateConnection(bool callDecrement=true)
```
paulmedynski: We should avoid adding a parameter that is only used in Debug mode.  See my comment below on why this argument won't be necessary.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbConnectionInternal.cs
```diff
@@ -513,14 +513,16 @@ internal virtual void CloseConnection(DbConnection owningObject, SqlConnectionFa
...
         {
             // Internal method called from the connection pooler so we don't expose
             // the Deactivate method publicly.
             SqlClientEventSource.Log.TryPoolerTraceEvent("<prov.DbConnectionInternal.DeactivateConnection|RES|INFO|CPOOL> {0}, Deactivating", ObjectID);
 
             #if DEBUG
-            int activateCount = Interlocked.Decrement(ref _activateCount);
+            int activateCount = _activateCount;
```
paulmedynski: It is obvious from the calls to `Deactivate()` that it should be expected to be called more times than `Activate()` for the same connection instance.  This pattern is already happening in Release mode and we aren't seeing any ill effects.  We can debate whether or not it _should_ be a pattern, but that's out of scope for this work.

I suggest using `Interlocked.CompareExchange()` rather than `Decrement()`.  That way you can avoid the decrement if the count is already zero.  Unfortunately, this is ugly because there is no `CompareExchangeIfNotEqual()`:

```
int origCount, newCount;
do
{
  origCount = _activateCount;

  if (origCount == 0)
  {
    break;
  }

  newCount = origCount - 1;
}
while (Interlocked.CompareExchange(ref _activateCount, newCount, origCount) != origCount);
```

This is thread-safe and ensures you never decrement below zero.

Using this also means you don't need the `Debug.Assert()` at all, which is the ptoblem we're really trying to solve here.

paulmedynski (changes requested):
You can re-work things to eliminate the `Debug.Assert()` completely.

paulmedynski:
We've also been having some issues with macOS CI, which should be solved if you rebase or merge main.

paulmedynski (approved):

paulmedynski:
I'm also working on getting Debug mode PR pipelines to run successfully, and these Asserts() are causing trouble there as well.  We will take your improvement, and I will decide whether or not these checks are valuable in my PR.

shreyarao4:
Can we merge this?

apoorvdeshmukh (approved):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbConnectionInternal.cs
```diff
@@ -520,8 +520,19 @@ internal void DeactivateConnection()
...
+                origCount = _activateCount;
+
+                if (origCount == 0)
+                {
+                  break;
+                }
+
+                newCount = origCount - 1;
+            }
+            while (Interlocked.CompareExchange(ref _activateCount, newCount, origCount) != origCount);
```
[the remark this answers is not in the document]
paulmedynski: True, but this is `#if DEBUG` code and has no side effects on purpose.

PR #3588, opened 2025-08-29
Set LangVersion to 13 for all targets

paulmedynski:
## Description

- Explicitly setting C# language version to 13 for all targets.
- Adding IDisposable to ref structs where appropriate.

## Testing

Existing CI will cover everything.


mdaigle (approved):

apoorvdeshmukh (approved):

Wraith2:
Why? Is there some evidence that the ref structs weren't being handled correctly by the compiler? Even in the absence of the marker interface it was able to detect the dispose pattern and correctly dispose them.

It's not supported to use later langauge versions on old runtimes like netfx because the compiler requires runtime features that don't exist. If it's certain that this will fail correctly at compile time then it's probably OK but I didn't think that it was certain.

paulmedynski:
This is a refinement of #3240.  It wasn't helpful to have .NET 8 targets use C# 12 while the other targets use C# 13.

Ref structs implementing an interface is a compiler feature and works when running against older .NET runtimes.

benrr101 (approved):

PR #3590, opened 2025-08-30
Fix | Allow SqlBulkCopy to operate on hidden columns

### LINKED ISSUE #1854: SqlBulkCopy - throws error when Hidden columns are specified in the ColumnMappings
### Describe the bug
As the title describes, Hidden columns are always ignored by SqlBulkCopy, even when they are targets in SqlBulkCopy.ColumnMappings.

```
ColumnMapping does not exist in destination...
```

### To reproduce
I will create a repro of this in my fork, but wanted to log this first and get some feedback if this would be a candidate to even be fixed because the behavior is so long-standing.

### Expected behavior
Hidden columns should be accounted for when they are explicitly requested

### Further technical details
Microsoft.Data.SqlClient version: 5.0.1
.NET target: net6
SQL Server version: SQL Server 2019
Operating system: Windows 10

**Additional context**
It seems like this behavior has always been there, so maybe this is a feature request (it seems like a bug at first glance to me though)

edwardneal:
## Description

This deals with an older bug: if we try to use `SqlBulkCopy` to import data into a SQL Server table with hidden columns, we can't use those hidden columns as destinations - SqlBulkCopy won't acknowledge that the column exists.

The root cause of this is that we discover the columns in the target table by running the command below and looking at the TDS-level metadata:
```sql
SET FMTONLY ON;
SELECT * FROM {table}
SET FMTONLY OFF;
```
Hidden columns don't appear when we run a `SELECT *` query. This PR overcomes the problem by querying `sys.all_columns` for the column list, building a `SELECT` query dynamically.

It's worth noting that this PR doesn't result in a successful bulk copy. As far as I can tell, the only way to create a hidden column is to create a temporal table - specific columns can be marked `GENERATED ALWAYS AS ROW START HIDDEN`. Trying to insert data into these columns will result in a SqlException containing error 13536 from SQL Server:

> Cannot insert an explicit value into a GENERATED ALWAYS column in table '{database}.{schema}.{table}'.

This is a better failure mode than the existing behaviour (an InvalidOperationException) though - it makes it clear that SqlBulkCopy found the column and that SQL Server is rejecting it.

## Issues

Fixes #1854.

## Testing

New unit test added, all SqlBulkCopy unit tests pass.

The original issue mentioned that running DROP PERIOD FOR SYSTEM_TIME would disable versioning and leave the columns hidden. I wasn't able to reproduce this - once the table has been altered, the columns are no longer hidden.


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
```diff
@@ -456,27 +453,41 @@ private string CreateInitialQuery()
...
+            // Specify the column names explicitly. This is to ensure that we can map to hidden columns (e.g. columns in temporal tables.)
+            // If the target table doesn't exist, OBJECT_ID will return NULL and @Column_Names will remain non-null. The subsequent SELECT *
+            // query will then continue to fail with "Invalid object name" rather than with an unusual error because the query being executed
+            // is NULL.
+            // Some hidden columns (e.g. SQL Graph columns) cannot be selected, so we need to exclude them explicitly.
+            return $"""
+SELECT @@TRANCOUNT;
+
+DECLARE @Column_Names NVARCHAR(MAX) = NULL;
+IF EXISTS (SELECT TOP 1 * FROM sys.all_columns where object_id = object_id('sys.all_columns') AND [name] = 'graph_type')
```
benrr101: Nit: Make sure case for SQL is the same throughout (and preferably all UPPER CASE). Maybe also wrap all the identifiers in `[]` to be consistent, too

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/SqlBulkCopyTest/HiddenTargetColumn.cs
```diff
@@ -0,0 +1,71 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Data.Common;
+using Xunit;
+
+namespace Microsoft.Data.SqlClient.ManualTesting.Tests
+{
+    public class HiddenTargetColumn
```
benrr101, line 11: 🤔 Why does this need to be its own class? I'd say either add this code to the SqlBulkCopyTest.cs or move the tests into their own class of tests. To me it's counterintuitive to have a class in a test project that doesn't have any test entrypoints, nor utility code.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/SqlBulkCopyTest/SqlGraphTables.cs
```diff
@@ -0,0 +1,45 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Data;
+using System.Data.Common;
+using Xunit;
+
+namespace Microsoft.Data.SqlClient.ManualTesting.Tests
+{
+    public class SqlGraphTables
```
benrr101, line 12: Same argument as above with the separate class applies here.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
benrr101: I suppose since the results of this SELECT are only used to determine if anything exists, then it doesn't matter that we're using * here.

benrr101 (approved):
After working on sqltoolsservice and DSCT, I learned how it's pretty much always frowned upon to use * for "official" queries. So, I'm kinda surprised we were using * to discover the columns in a table.

A few changes that could be made, but I won't hold up the PR over them.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
edwardneal: I completely missed that! Yes, fixed now thanks.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/SqlBulkCopyTest/HiddenTargetColumn.cs
edwardneal: In principle, I agree. I'm unhappy with most of the SqlBulkCopyTest.cs file - it's ignoring the DataTestUtility methods to generate object names, it appears to be supressing tests on Azure SQL, and it's ultimately a class which should be a namespace.

Maintaining consistency means that I'd need to move 40-50 tests though, so I'd prefer to set aside a dedicated cleanup PR for those. The two new ones can set the standard for the others.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
edwardneal: That's correct - we're just testing for the existence of rows, not actually reading any values back.

paulmedynski (approved):

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/SqlBulkCopyTest/HiddenTargetColumn.cs
benrr101: Understandable, we can target cleaning this up more later...

benrr101:
@edwardneal so looking at the failures, I think the issue is because in main we've dropped DataTestUtility.GetUniqueNameForSqlServer and split it into GetShortName and GetLongName. Even though it's not a conflict, I guess we're doing the merge before running the CI builds? Anyways, take a look at the notes for those methods, and we'll get the PR going.

edwardneal:
Thanks @benrr101, I've just merged to main.

paulmedynski (dismissed):

benrr101:
@edwardneal looks like there's a test related to hidden columns failing :(
```
Failed Microsoft.Data.SqlClient.ManualTesting.Tests.SqlBulkCopyTests.HiddenTargetColumn.WriteToServer_CopyToHiddenTargetColumn_ThrowsSqlException [1 s]
     Assert.StartsWith() Failure: String start does not match
  String:         "Cannot insert an explicit value into a GE"···
  Expected start: "Cannot insert an explicit value into a GE"···
    Stack Trace:
       at Microsoft.Data.SqlClient.ManualTesting.Tests.SqlBulkCopyTests.HiddenTargetColumn.WriteToServer_CopyToHiddenTargetColumn_ThrowsSqlException() in /_/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/SqlBulkCopyTest/HiddenTargetColumn.cs:line 59
     at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
     at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
```

edwardneal:
That's frustrating, but thanks. Our test case is covered by the SqlException.Number assertion on the previous line, so I've simplified the comparison.

It's odd to see this fail though - I've tested against SQL 2022, 2025 and Azure...

benrr101 (approved):
Looks good to me! Thanks for hanging in there on the test failures!

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
```diff
@@ -456,27 +453,41 @@ private string CreateInitialQuery()
...
+            // Specify the column names explicitly. This is to ensure that we can map to hidden columns (e.g. columns in temporal tables.)
+            // If the target table doesn't exist, OBJECT_ID will return NULL and @Column_Names will remain non-null. The subsequent SELECT *
+            // query will then continue to fail with "Invalid object name" rather than with an unusual error because the query being executed
+            // is NULL.
+            // Some hidden columns (e.g. SQL Graph columns) cannot be selected, so we need to exclude them explicitly.
+            return $"""
+SELECT @@TRANCOUNT;
+
+DECLARE @Column_Names NVARCHAR(MAX) = NULL;
+IF EXISTS (SELECT TOP 1 * FROM sys.all_columns WHERE [object_id] = OBJECT_ID('sys.all_columns') AND [name] = 'graph_type')
```
0xfeeddeadbeef, line 474: Is this query missing a `{CatalogName}.` before `sys.all_columns`?

```diff
-IF EXISTS (SELECT TOP 1 * FROM sys.all_columns WHERE [object_id] = OBJECT_ID('sys.all_columns') AND [name] = 'graph_type')
+IF EXISTS (SELECT TOP 1 * FROM {CatalogName}.sys.all_columns WHERE [object_id] = OBJECT_ID('{CatalogName}.sys.all_columns') AND [name] = 'graph_type')
```

edwardneal: No - the only purpose of this query is to detect whether or not a `graph_type` column exists in `sys.all_columns`. This column is present when querying the DMV from SQL Server 2017 upwards, even if the database itself is in an earlier compatibility mode.

PR #3593, opened 2025-09-02
Merge | SqlCommand Prepare/Unprepare

benrr101:
## Description
This PR merges the methods used for preparing and unpreparing SqlCommand objects. This PR introduces the new merge style that I'm adopting for this class. 

### Merge Partials
Rather than try to do the merge as one big bang as I've done in the past, with simpler classes, this class needs some special attention. One particular issue that causes great headache is the disorder of the members of the class and the sheer size of the class. I would like to make improvements here while also making it clear what how things move from the separate projects to the common project. So, I've made the SqlCommand files in the separate projects partials, and added a partial to the common project. This allows me to move each member to the partial one at a time, give it access to what's in the separate project files, give the separate project files access to whats been merged, as well as sort the merged members and apply light cleanup in each merge.

Please note, there are a lot of "TODO" items added to this code, mostly as notes for myself on how to clean up the class once merging is complete.

**Each commit has been tested to build - it is encouraged to step through this PR commit-by-commit**

## Issues
Continuation of work for #1261 

## Testing
Code still builds, CI should confirm


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.cs
```diff
@@ -0,0 +1,294 @@
...
+        /// Current state of preparation of the command.
+        /// By default, assume the user is not sharing a connection so the command has not been prepared.
+        /// </summary>
+        private EXECTYPE _execType = EXECTYPE.UNPREPARED;
+
+        /// <summary>
+        /// True if the user changes the command text or number of parameters after the command has
+        /// already prepared.
+        /// </summary>
+        // @TODO: Consider renaming "_IsUserDirty"
```
paulmedynski, line 59: Maybe `_isPrepareStale` ?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.cs
```diff
@@ -0,0 +1,294 @@
...
+
+            _execType = EXECTYPE.PREPAREPENDING;
+
+            // Don't zero out the handle because we'll pass it in to sp_prepexec on the next prepare
+            // Unless the close count isn't the same as when we last prepared
+            if (_activeConnection.CloseCount != _preparedConnectionCloseCount ||
+                _activeConnection.ReconnectCount != _preparedConnectionReconnectCount)
+            {
+                // Reset our handle
+                _prepareHandle = s_cachedInvalidPrepareHandle;
```
paulmedynski, line 299: When did this get set?  Should it have moved into this new file?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.cs
```diff
@@ -0,0 +1,294 @@
...
+        /// function will test this flag to determine whether the returned value is a
+        /// _prepareHandle or something else.
+        /// </summary>
+        // @TODO: Make auto-property
+        private bool _inPrepare = false;
+
+        /// <summary>
+        /// TDS session the current instance is using.
+        /// </summary>
+        private TdsParserStateObject _stateObj;
```
paulmedynski, line 87: Maybe we can rename this TdsSession some day...

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.cs
```diff
@@ -0,0 +1,294 @@
...
+        private bool IsDirty
+        {
+            get
+            {
+                // @TODO: Factor out closeCount/reconnectCount checks to properties and clean up.
+                // To wit: closeCount checks whether the connection has been closed after preparation,
+                //    reconnectCount, the same only with reconnections.
+                
+                // only dirty if prepared
+                var activeConnection = _activeConnection;
```
paulmedynski, line 131: Why not just use `_activeConnection` below?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.cs
```diff
@@ -0,0 +1,294 @@
...
+
+            SqlStatistics statistics = null;
+            try
+            {
+                statistics = SqlStatistics.StartTimer(Statistics);
+
+                // Only prepare batch has parameters
+                // @TODO: Factor out stored proc/text+parameter count to property
+                // @TODO: Not using parentheses is confusing here b/c it relies on order of operations knowledge
+                if (IsPrepared && !IsDirty || CommandType == CommandType.StoredProcedure
```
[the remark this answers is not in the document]
paulmedynski: Agreed.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.cs
benrr101: Probably not that option... The problem is that there's multiple components that go into whether the command is dirty (see `IsDirty`), so this flag should indicate specifically that the user made a change to it that causes it to be dirty. `PrepareStale` is kind of just a synonym for `Dirty` since both indicate that the command needs to be re-prepared.

I'm leaving this as a todo since I don't want to make large refactor changes right now. So we can open it to suggestions later. Unless it's really offensive to leave this comment in today (and I'd kinda hope policheck would catch it if it was)

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.cs
benrr101: It probably makes sense to move _prepareHandle in this PR 👍 

Don't be surprised if there are things missing during the merge, I'm trying to make the PRs as managable a possible, so I kinda draw seemingly random lines around what items to move in one PR.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.cs
benrr101: Oh .... maybe. 😓

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.cs
benrr101: I suppose *theoretically* _activeConnection could change in the middle of the if statement and we'd have a race condition. Though I think if that happens, we've got much larger issues...

I made an attempt to clean this up more at one point, but decided merging isn't the time to make more than stylistic changes (unless it's necessary). I didn't trust myself enough with that change so I left it as-is with lots of comments.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.cs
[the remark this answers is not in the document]
benrr101: Alright - I made an update I'm happy with :)

paulmedynski (dismissed):
Looks good - thanks for the explanations!

cheenamalhotra:
It seems there's a test failure corresponding to changes in the PR, @benrr101 
Could you address them asap?

paulmedynski (approved):

samsharma2700 (approved):
LGTM

PR #3597, opened 2025-09-03
Fix unique db object name issues

paulmedynski:
## Description

- Fixed the unique name generators to:
  - Keep max lengths to 30 and 96 characters respectively.
  - Ensure sufficient uniqueness even when truncation occurs.

### DIFF
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
@@ -660,7 +660,7 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests
         /// 
         /// <param name="suffix">
         /// The suffix to use when generating the unique name, truncated to at
-        /// most 18 characters.
+        /// most 32 characters.
         /// </param>
         /// 
         /// <param name="withBracket">
@@ -685,9 +685,9 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests
             name.Append(GetGuidParts());
             name.Append('-');
 
-            if (suffix.Length > 18)
+            if (suffix.Length > 32)
             {
-                suffix = suffix[0..18];
+                suffix = suffix[0..32];
             }
 
             suffix =
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
@@ -578,7 +578,7 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests
         /// Generate a short unique database object name, whose maximum length
         /// is 30 characters, with the format:
         ///
-        ///   <GUID-Parts>-<Suffix>
+        ///   <GUID-Parts>_<Suffix>
         ///
         /// The GUID Parts will be the characters from the 1st and 4th blocks
         /// from a traditional string representation, as shown here:
@@ -604,7 +604,7 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests
         /// When true, the entire generated name will be enclosed in square
         /// brackets, for example:
         /// 
-        ///   [7ff01cb811f0-MySuffix]
+        ///   [7ff01cb811f0_MySuffix]
         /// </param>
         /// 
         /// <returns>
@@ -620,7 +620,7 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests
             }
 
             name.Append(GetGuidParts());
-            name.Append('-');
+            name.Append('_');
 
             int maxSuffixLength = withBracket ? 16 : 18;
             if (suffix.Length > maxSuffixLength)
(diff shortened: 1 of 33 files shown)


src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
```diff
@@ -556,59 +556,159 @@ public static bool DoesHostAddressContainBothIPv4AndIPv6()
...
+        {
+            StringBuilder name = new(30);
+
+            if (withBracket)
+            {
+                name.Append('[');
+            }
+
+            name.Append(GetGuidParts());
+            name.Append('-');
```
mdaigle: I think it's mad at this dash. We can try replacing it with an underscore.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
```diff
@@ -556,59 +556,159 @@ public static bool DoesHostAddressContainBothIPv4AndIPv6()
...
+        {
+            StringBuilder name = new(96);
+
+            if (withBracket)
+            {
+                name.Append('[');
+            }
+
+            name.Append(GetGuidParts());
+            name.Append('-');
```
mdaigle: same here

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
```diff
@@ -556,59 +556,159 @@ public static bool DoesHostAddressContainBothIPv4AndIPv6()
...
+        /// When true, the entire generated name will be enclosed in square
+        /// brackets, for example:
+        /// 
+        ///   [7ff01cb811f0_MySuffix]
+        /// </param>
+        /// 
+        /// <returns>
+        /// A unique database object name, no more than 30 characters long.
+        /// </returns>
+        public static string GetShortName(string suffix, bool withBracket = true)
```
mdaigle: Keeping the test provided string as a prefix is helpful for manually finding the generated tables in the database if you ever need to inspect test data locally when debugging.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
paulmedynski: Yes, underscores work when the name is unquoted (i.e. no enclosing [ ]).  Alredy updated.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
paulmedynski: So:

`<prefix>_<GUID-Parts>`

and

`<prefix>_<GUID-Parts>_<UserName>_<MachineName>`

No problem - I'll wait for the current CI run to pass, and then flip them.  I just want to guarantee that GUID Parts ends up in the name without being truncated.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
```diff
@@ -556,59 +556,176 @@ public static bool DoesHostAddressContainBothIPv4AndIPv6()
...
+            name.Append(prefix);
+            name.Append('_');
+            name.Append(GetGuidParts());
+
+            if (withBracket)
+            {
+                name.Append(']');
+            }
+
+            System.Console.WriteLine($"Generated short name \"{name.ToString()}\"");
```
paulmedynski: Remove these once we figure out what SQL Server doesn't like about these names.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
```diff
@@ -556,59 +556,176 @@ public static bool DoesHostAddressContainBothIPv4AndIPv6()
...
+            }
+
+            name.Append(suffix);
+
+            if (withBracket)
+            {
+                name.Append(']');
+            }
+
+            System.Console.WriteLine($"Generated long name \"{name.ToString()}\"");
```
paulmedynski: Remove once test failures are fixed.

paulmedynski (commented):
Commentary for myself.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
```diff
@@ -556,59 +556,176 @@ public static bool DoesHostAddressContainBothIPv4AndIPv6()
...
+        /// is 30 characters, with the format:
+        ///
+        ///   <Prefix>_<GuidParts>
+        ///
+        /// The Prefix will be truncated to satisfy the overall maximum length.
+        ///
+        /// The GUID parts will be the characters from the 1st and 4th blocks
+        /// from a traditional string representation, as shown here:
+        ///
+        ///   7ff01cb8-88c7-11f0-b433-00155d7e531e
```
benrr101, line 588: Alternatively, we could just turn this hex string into, idk, base64? Although then we have to consider upper/lower case ... etc ... so, this is probably fine.

benrr101 (approved):

apoorvdeshmukh (approved):

PR #3605, opened 2025-09-09
Add continue request usage

Wraith2:
Builds on the stable base provided by https://github.com/dotnet/SqlClient/pull/3534 and adds the optional async-continue capability back in. It is optional, it is not on by default in this build. To use it you must set a switch: `AppContext.SetSwitch("Switch.Microsoft.Data.SqlClient.UseCompatibilityAsyncBehaviour", false);` this will use the new multiplexer and async-continue support already defaults to enabled.

This changes the way that continue is used from the original implementation. Originally continue was always available which proved to cause problems with routes through the code that did not expect to be able to fail causing lost context. This new approach runs as if continue is disabled until a function is called which explicitly requests continue mode to be enabled. The only functions that can do this are the ones that read multi-packet string or binary data. The request is cleared when the read completes so it cannot affect other reads from the same packet.

## Benchmark results:
All Reads were 10 Mib of data.
### BinaryRead
| Method      | UseContinue | Mean        | Error     | StdDev    | Gen0      | Gen1      | Gen2      | Allocated |
|------------ |------------ |------------:|----------:|----------:|----------:|----------:|----------:|----------:|
| **Async**       | **False**       | **5,840.26 ms** | **94.748 ms** | **88.627 ms** | **2000.0000** | **1000.0000** | **1000.0000** |  **40.81 MB** |
| StreamAsync | False       |    46.91 ms |  0.925 ms |  1.692 ms | 3083.3333 | 1833.3333 | 1833.3333 |   51.8 MB |
| Sync        | False       |    26.87 ms |  0.534 ms |  1.042 ms | 1562.5000 |  906.2500 |  906.2500 |  30.73 MB |
| **Async**       | **True**        |    **50.78 ms** |  **1.003 ms** |  **1.561 ms** | **2181.8182** | **2090.9091** |  **909.0909** |  **40.82 MB** |
| StreamAsync | True        |    58.23 ms |  1.114 ms |  2.469 ms | 2888.8889 | 1666.6667 | 1666.6667 |   51.8 MB |
| Sync        | True        |    26.50 ms |  0.509 ms |  0.522 ms | 1562.5000 |  906.2500 |  906.2500 |  30.73 MB |

### StringRead

| Method | UseContinue | Mean        | Error      | StdDev     | Gen0      | Gen1      | Gen2     | Allocated |
|------- |------------ |------------:|-----------:|-----------:|----------:|----------:|---------:|----------:|
| **Async**  | **False**       | **5,832.31 ms** | **114.512 ms** | **191.324 ms** | **1000.0000** |         **-** |        **-** |   **50.8 MB** |
| Sync   | False       |    34.32 ms |   0.686 ms |   1.026 ms | 1000.0000 |  400.0000 | 400.0000 |  40.72 MB |
| **Async**  | **True**        |    **66.26 ms** |   **0.960 ms** |   **1.895 ms** | **1625.0000** | **1500.0000** | **375.0000** |  **50.81 MB** |
| Sync   | True        |    35.14 ms |   0.694 ms |   1.215 ms | 1000.0000 |  400.0000 | 400.0000 |  40.72 MB |

### XmlRead
 Method | UseContinue | Mean          | Error        | StdDev       | Gen0          | Gen1          | Gen2         | Allocated    |
------- |------------ |--------------:|-------------:|-------------:|--------------:|--------------:|-------------:|-------------:|
 **Sync**   | **False**       |      **92.20 ms** |     **1.334 ms** |     **1.183 ms** |     **3166.6667** |     **3000.0000** |    **1000.0000** |     **38.57 MB** |
 Async  | False       | 182,897.65 ms | 3,447.733 ms | 5,053.643 ms | 31360000.0000 | 31012000.0000 | 9640000.0000 | 360347.15 MB |
 **Sync**   | **True**        |      **89.87 ms** |     **0.803 ms** |     **0.712 ms** |     **3166.6667** |     **3000.0000** |    **1000.0000** |     **38.57 MB** |
 Async  | True        |     128.25 ms |     2.192 ms |     3.213 ms |     4800.0000 |     4600.0000 |    1400.0000 |     56.39 MB |

Here are the benchmark files that I used to get these numbers. You'll need connection strings and to generate a random 10Mib xml files from somewhere on the web. [Benchmarks.zip](https://github.com/user-attachments/files/22232851/Benchmarks.zip)
The async xml numbers are pretty unbelievable. If you replicate or improve the benches it would be good to get perf corroboration.

@ErikEJ this will be the build you'll want to test and then possible make available to people, as discussed on the previous PR.
@edwardneal if you have the resources could you run your fantastic test suite against this version and see if you can identify any new bugs?
@dotnet/sqlclientdevteam can you run CI please.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
@@ -6447,7 +6447,7 @@ namespace Microsoft.Data.SqlClient
                 case TdsEnums.SQLVECTOR:
                     // Vector data is read as non-plp binary value.
                     // This is same as reading varbinary(8000).
-                    result = stateObj.TryReadByteArrayWithContinue(length, out b);
+                    result = stateObj.TryReadByteArrayWithContinue(length, isPlp: false, out b);
                     if (result != TdsOperationStatus.Done)
                     {
                         return result;
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
@@ -6377,25 +6377,10 @@ namespace Microsoft.Data.SqlClient
                 case TdsEnums.SQLVARBINARY:
                 case TdsEnums.SQLIMAGE:
                     byte[] b = null;
-
-                    // If varbinary(max), we only read the first chunk here, expecting the caller to read the rest
-                    if (isPlp)
+                    result = stateObj.TryReadByteArrayWithContinue(length, isPlp, out b);
+                    if (result != TdsOperationStatus.Done)
                     {
-                        // If we are given -1 for length, then we read the entire value,
-                        // otherwise only the requested amount, usually first chunk.
-                        result = stateObj.TryReadPlpBytes(ref b, 0, length, out _);
-                        if (result != TdsOperationStatus.Done)
-                        {
-                            return result;
-                        }
-                    }
-                    else
-                    {
-                        result = stateObj.TryReadByteArrayWithContinue(length, out b);
-                        if (result != TdsOperationStatus.Done)
-                        {
-                            return result;
-                        }
+                        return result;
                     }
 
                     if (md.isEncrypted
@@ -6790,7 +6775,7 @@ namespace Microsoft.Data.SqlClient
                         // Note: Better not come here with plp data!!
                         Debug.Assert(length <= TdsEnums.MAXSIZE);
                         byte[] b = new byte[length];
-                        result = stateObj.TryReadByteArray(b, length);
+                        result = stateObj.TryReadByteArrayWithContinue(length, isPlp: false, out b);
                         if (result != TdsOperationStatus.Done)
                         {
                             return result;
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
@@ -1560,45 +1560,71 @@ namespace Microsoft.Data.SqlClient
             return TdsOperationStatus.Done;
         }
 
-        public TdsOperationStatus TryReadByteArrayWithContinue(int length, out byte[] bytes)
+        internal TdsOperationStatus TryReadByteArrayWithContinue(int length, out byte[] value)
         {
-            bytes = null;
-            int offset = 0;
-            byte[] temp = null;
+            return TryReadByteArrayWithContinue(length, false, out value);
+        }
+        internal TdsOperationStatus TryReadByteArrayWithContinue(int length, bool isPlp, out byte[] value)
+        {
+            Debug.Assert(_syncOverAsync || !_asyncReadWithoutSnapshot, "This method is not safe to call when doing sync over async");
+
+            byte[] buf = null;
+            RequestContinue(true);
             (bool canContinue, bool isStarting, bool isContinuing) = GetSnapshotStatuses();
-            if (canContinue)
+
+            if (isPlp)
             {
-                temp = TryTakeSnapshotStorage() as byte[];
-                Debug.Assert(temp != null || !isContinuing, "if continuing stored buffer must be present to contain previous data to continue from");
-                Debug.Assert(bytes == null || bytes.Length == length, "stored buffer length must be null or must have been created with the correct length");
-                
-                if (temp != null)
+                bool compatibilityMode = LocalAppContextSwitches.UseCompatibilityAsyncBehaviour;
+                TdsOperationStatus result = TryReadPlpBytes(ref buf, 0, int.MaxValue, out length, canContinue && !compatibilityMode, canContinue && !compatibilityMode, compatibilityMode);
+                if (result != TdsOperationStatus.Done)
                 {
-                    offset = GetSnapshotTotalSize();
+                    value = null;
+                    return result;
                 }
+
+                AssertValidState();
             }
-
-
-            if (temp == null)
+            else
             {
-                temp = new byte[length];
-            }
-
-            TdsOperationStatus result = TryReadByteArray(temp, length, out _, offset, isStarting || isContinuing);
-
-            if (result == TdsOperationStatus.Done)
-            {
-                bytes = temp;
-            }
-            else if (result == TdsOperationStatus.NeedMoreData)
-            {
+                int startOffset = 0;
                 if (canContinue)
                 {
-                    SetSnapshotStorage(temp);
+                    buf = TryTakeSnapshotStorage() as byte[];
+                    Debug.Assert(buf != null || !isContinuing, "if continuing stored buffer must be present to contain previous data to continue from");
+                    Debug.Assert(buf == null || buf.Length == length, "stored buffer length must be null or must have been created with the correct length");
+
+                    if (buf != null)
+                    {
+                        startOffset = GetSnapshotTotalSize();
+                    }
                 }
+
+                if (buf == null || buf.Length < length)
+                {
+                    buf = new byte[length];
+                }
+
+                TdsOperationStatus result = TryReadByteArray(buf, length, out _, startOffset, canContinue);
+
+                if (result != TdsOperationStatus.Done)
+                {
+                    if (result == TdsOperationStatus.NeedMoreData)
+                    {
+                        if (canContinue)
+                        {
+                            SetSnapshotStorage(buf);
+                        }
+                    }
+                    value = null;
+                    return result;
+                }
+
+                AssertValidState();
+
             }
 
-            return result;
+            value = buf;
+            return TdsOperationStatus.Done;
         }
 
         // Takes no arguments and returns a byte from the buffer.  If the buffer is empty, it is filled
@@ -2841,7 +2867,7 @@ namespace Microsoft.Data.SqlClient
                 }
 
                 // previous buffer is in snapshot
-                _inBuff = new byte[_inBuff.Length];
+                NewBuffer(_inBuff.Length);
                 result = TdsOperationStatus.NeedMoreData;
             }
 
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
@@ -1560,17 +1560,13 @@ namespace Microsoft.Data.SqlClient
             return TdsOperationStatus.Done;
         }
 
-        internal TdsOperationStatus TryReadByteArrayWithContinue(int length, out byte[] value)
-        {
-            return TryReadByteArrayWithContinue(length, false, out value);
-        }
         internal TdsOperationStatus TryReadByteArrayWithContinue(int length, bool isPlp, out byte[] value)
         {
             Debug.Assert(_syncOverAsync || !_asyncReadWithoutSnapshot, "This method is not safe to call when doing sync over async");
 
             byte[] buf = null;
             RequestContinue(true);
-            (bool canContinue, bool isStarting, bool isContinuing) = GetSnapshotStatuses();
+            (bool canContinue, bool _, bool isContinuing) = GetSnapshotStatuses();
 
             if (isPlp)
             {
@@ -1973,7 +1969,7 @@ namespace Microsoft.Data.SqlClient
 
             if (((_inBytesUsed + cBytes) > _inBytesRead) || (_inBytesPacket < cBytes))
             {
-                TdsOperationStatus result = TryReadByteArrayWithContinue(cBytes, out buf);
+                TdsOperationStatus result = TryReadByteArrayWithContinue(cBytes, isPlp: false, out buf);
                 if (result != TdsOperationStatus.Done)
                 {
                     value = null;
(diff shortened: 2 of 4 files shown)


ErikEJ:
@Wraith2 This looks awesome - looks like you also fixed the issue with increased allocations..

I will give this a try against the repro VM asap

Wraith2:
> @Wraith2 This looks awesome - looks like you also fixed the issue with increased allocations..

That was fixed in the original fix so it's already merged to main. https://github.com/dotnet/SqlClient/pull/3534/commits/4b891d8ab914a8c2be889d7916e4c7b00597b7c1

Since i've just reviewed the other PR i'll cc a couple of people who were interested in it and may want to test this if it generates artifacts, @MichelZ @rhuijben

edwardneal:
Excellent! I've built the PR locally and kicked off my stress tests, including the XML case. These will likely take a few days to fully run, but I'll post the results when available.

Wraith2:
> I've built the PR locally and kicked off my stress tests, including the XML case

The performance on the xml is shockingly bad. When I realised just how slow it was going on the current main branch I thought I must have either written the test to loop 1000 times or somehow broken SqlCachedReader.

ErikEJ:
@backstromjoel @valentiniliescu @nilzzzzzz @CSharpFiasco @luca-domenichini @warappa @stevendarby @wjrogers @AnderssonPeter @p10tyr @Eli-Black-Work @BradBarnich @igbenic 

You have all expressed an interest in this performance issue - please test this PR build!

gnegno84:
> Builds on the stable base provided by #3534 and adds the optional async-continue capability back in. It is optional, it is not on by default in this build. To use it you must set a switch: `AppContext.SetSwitch("Switch.Microsoft.Data.SqlClient.UseCompatibilityAsyncBehaviour", false);` this will use the new multiplexer and async-continue support already defaults to enabled.

On our side, we tried to repro one issue that was happening with 6.1.0 (randomly receiving null values for a json serialized string field). 
I've tested with the artifact in this PR and I can confirm that the issue IS NOT happening anymore with both approaches:
- UseCompatibilityAsyncBehaviour set to true (expected)
- UseCompatibilityAsyncBehaviour set to false 

From a performance perspective, I cannot add valuable info as our test was quite simple and not benchmarked. 
Thanks for the updates.

ErikEJ:
@Wraith2 The `6.11.0-pull.125205` artifcat verified OK against my Sweden repro VM.

ErikEJ:
@Wraith2 How do I set an appcontext swith for an entire xUnit test assembly?

Wraith2:
Probably in the ctor but because of the caching in the library once you've set it and the value has been read you can't change the value. To do that you'll need to use private reflection as we do in the tests here.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
```diff
@@ -1560,45 +1560,71 @@ public TdsOperationStatus TryReadByteArray(Span<byte> buff, int len, out int tot
             return TdsOperationStatus.Done;
         }
 
-        public TdsOperationStatus TryReadByteArrayWithContinue(int length, out byte[] bytes)
+        internal TdsOperationStatus TryReadByteArrayWithContinue(int length, out byte[] value)
```
paulmedynski: I think you can eliminate this method if you change the only remaining call to it (further down in this file) to use the new one below with `isPlp = false`.

Wraith2: Ok, unified. Please kick the CI.

paulmedynski (approved):

ErikEJ:
@Wraith2 The `6.11.0-pull.125205` artifact verified OK against the EF Core tests

ErikEJ:
@Wraith2 FYI, I added this class to the test project:

```c#
using Xunit.Sdk;
[assembly: TestFramework("Microsoft.EntityFrameworkCore.AssemblyFixture", "Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests")]

namespace Microsoft.EntityFrameworkCore;

public sealed class AssemblyFixture : XunitTestFramework
{
    public AssemblyFixture(IMessageSink messageSink)
        :base(messageSink)
    {
        AppContext.SetSwitch("Switch.Microsoft.Data.SqlClient.UseCompatibilityAsyncBehaviour", false);
    }
}

```

edwardneal:
The stress-testing passed as expected, but it's worth noting that we currently need two AppContext switches to enable this:
```cs
AppContext.SetSwitch("Switch.Microsoft.Data.SqlClient.UseCompatibilityAsyncBehaviour", false);
AppContext.SetSwitch("Switch.Microsoft.Data.SqlClient.UseCompatibilityProcessSni", false);
```
If the latter switch is not explicitly set, we see increased memory usage but not the performance improvements.

Async and sync are very close to parity in normal circumstances, remote SQL Server instances saw nearly identical results in my benchmarks. I noticed that in situations where the SQL Server is able to keep up with the supply of PLP data, the speed improvements here mean that SqlClient is able to call code paths in TryReadByteArray which copy packet buffers around when building the snapshot. This is the current speed bottleneck as far as I can tell.

While nothing's needed for this PR, it looks like SQL Server will try to adjust the size of the PLP data blocks it sends to align with the number of bytes left in each TDS packet. With that in mind, we could avoid copying memory; if a read will take us precisely to the end of the current TDS packet, we could construct a correctly-sized `ReadOnlyMemory<byte>` over the TDS packet buffer, pass that up to the snapshot directly and allocate a new buffer for the next packet.

I also saw a minor (~5%) performance bump in XML data when I changed the `SqlCachedBuffer.MaxChunkSize` constant to 8192 (a power of two which is greater than the default packet size.)

ErikEJ:
@edwardneal Thanks for the info about the switches, I was not aware of that.

Let me try to add both to the EF Core test suite

Wraith2:
The `"Switch.Microsoft.Data.SqlClient.UseCompatibilityAsyncBehaviour"` switch code is currently:

```csharp
        public static bool UseCompatibilityAsyncBehaviour
        {
            get
            {
                if (UseCompatibilityProcessSni)
                {
                    // If ProcessSni compatibility mode has been enabled then the packet
                    // multiplexer has been disabled. The new async behaviour using continue
                    // point capture is only stable if the multiplexer is enabled so we must
                    // return true to enable compatibility async behaviour using only restarts.
                    return true;
                }

                if (s_useCompatibilityAsyncBehaviour == Tristate.NotInitialized)
                {
                    if (!AppContext.TryGetSwitch(UseCompatibilityAsyncBehaviourString, out bool returnedValue) || returnedValue)
                    {
                        s_useCompatibilityAsyncBehaviour = Tristate.True;
                    }
                    else
                    {
                        s_useCompatibilityAsyncBehaviour = Tristate.False;
                    }
                }
                return s_useCompatibilityAsyncBehaviour == Tristate.True;
            }
        }
```

which unless I can't mentally process boolean logic (it's happened) means that it defaults to true and you have to turn it off. However because `"Switch.Microsoft.Data.SqlClient.UseCompatibilityProcessSni"` is currently false the short circuit is active. 
So as I stated in the first post If you turn OFF compatibility sni you should automatically get the new continue-async behaviour.

If you're settings both flags explicitly you'll have to manage your combinations and that's likely what's happening in our tests but for casual users it should only require one settings.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -6790,7 +6775,7 @@ internal TdsOperationStatus TryReadSqlValueInternal(SqlBuffer value, byte tdsTyp
                         // Note: Better not come here with plp data!!
                         Debug.Assert(length <= TdsEnums.MAXSIZE);
                         byte[] b = new byte[length];
-                        result = stateObj.TryReadByteArray(b, length);
+                        result = stateObj.TryReadByteArrayWithContinue(length, isPlp: false, out b);
```
ericsampson, line 6778: I might be dumb today, but how can this `case TdsEnums.SQLVECTOR` get hit? (and for that matter, the same question for all the types that fall into here)

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
```diff
@@ -1560,45 +1560,67 @@ public TdsOperationStatus TryReadByteArray(Span<byte> buff, int len, out int tot
...
+
+            if (isPlp)
             {
-                temp = TryTakeSnapshotStorage() as byte[];
-                Debug.Assert(temp != null || !isContinuing, "if continuing stored buffer must be present to contain previous data to continue from");
-                Debug.Assert(bytes == null || bytes.Length == length, "stored buffer length must be null or must have been created with the correct length");
-                
-                if (temp != null)
+                bool compatibilityMode = LocalAppContextSwitches.UseCompatibilityAsyncBehaviour;
+                TdsOperationStatus result = TryReadPlpBytes(ref buf, 0, int.MaxValue, out length, canContinue && !compatibilityMode, canContinue && !compatibilityMode, compatibilityMode);
```
ericsampson, line 1574: worth using named arguments here for non-IDE readability?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
Wraith2: This code switches on the tdsType value which is a parameter. So it's going to get called with the value. Other than that you'll have to explain what you think the problem is.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
Wraith2: possibly. But if you're using this codebase without an IDE you're going to find a lot of confusing stuff. I'd prefer to make such changes in a modernisation/syntax overhaul PR rather than mix pure style changes together with logic alterations because it makes it easier to review the important parts.

Wraith2:
Is this waiting for anything other than a second review @dotnet/sqlclientdevteam ?

mdaigle:
@Wraith2 I've been out for a couple weeks. I'll start taking a look tomorrow!

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
ericsampson: Makes sense, can mark as resolved.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
ericsampson: in the TryReadSqlValue switch these same cases will get caught/breaked first instead of falling through to the default case where TryReadSqlValueInternal is called, right?

Wraith2: `TryReadSqlValueInternal` can also be called from `TryReadSqlVariant`

cremor:
Thanks for all your work on this!

Assuming no problems are found, will this PR be included in a 6.1.x release?
If yes, will it stay opt-in as stated in the PR description?

Those questions are especially relevant for all users who don't know about this problem and just get SqlClient as an EF dependency. EF 10 uses SqlClient 6.1 and won't update to any non-LTS version.

And a small PS: You bolded the wrong lines in the "XmlRead" performance result table in the PR description.

Wraith2:
The only important change is the addition of the calls to RequestContinue. Everything needed by those calls was already present but could not be activated. The request enables continue mode _only_ for the read of the data that we know supports it. The revert of the flag, the conditional logic around ContinueEnabled, it was all already present but was previously only capable of working in async-retry mode.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObject.cs
mdaigle, line 1574: canContinue from GetSnapshotStatuses should already take into account the state of the compatibility switch. I think we can exclude checking its state here.

Wraith2: Possibly, but I'm not sure that using the composite value of canContinue instead of compatibilityMode is correct. It could be but it'll take a lot of reviewing to have any confidence in that. 
Note that the checking of the switch is cached so it's only an int comparison not a full AppContext lookup. 
If you want this change I'd prefer to do it as a future PR.

mdaigle: I'd like us to try to consolidate the bool/flag evaluations. There are a lot of complex bool evaluations involved in all of this code and it makes it very challenging to track. My suggestion here was motivated by that, but what you have now is functionally the same, so I won't block on it if there's concern.

rizi:
@Wraith2 where should I add this switch call in my code (not tests)?
E.g in my DB Context constructor or my Program.cs (Asp.net core).

Which version of SqlClient will take this switch into account? Any timeline when it will be released?

Thx for your hard work!

Br

Wraith2:
> @Wraith2 where should I add this switch call in my code (not tests)? E.g in my DB Context constructor or my Program.cs (Asp.net core).

That's your choice and will depend on how you structure your code. As long as the switch value is set before the first read is called then it will be active, the value is cached so once set it cannot be changed.

> Which version of SqlClient will take this switch into account? Any timeline when it will be released?

The version released from the `main` branch. I have no details on timelines or version numbers. I'm just a contributor.

rizi:
> @Wraith2 where should I add this switch call in my code (not tests)? E.g in my DB Context constructor or my Program.cs (Asp.net core).
> 
> Which version of SqlClient will take this switch into account? Any timeline when it will be released?
> 
> Thx for your hard work!
> 
> Br

Thx for the clarification.

paulmedynski:
@rizi - I've added this to the 7.0.0 Preview 2 milestone, release expected mid/late October.

rizi:
> @rizi - I've added this to the 7.0.0 Preview 2 milestone, release expected mid/late October.

Thx for the info

PR #3606, opened 2025-09-10
Dev/samsharma2700/user agent fe

samsharma2700:
## Description

Part 3 for User Agent Feature Extension work. 
Part 2 : https://github.com/dotnet/SqlClient/pull/3451
Part 1 : https://github.com/dotnet/SqlClient/pull/3582

Also adding two tests to verify changes. One test forces the test server to send an Ack(we do not expect this in reality)

Unit test and functional test are passing


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -209,7 +209,7 @@ internal bool IsDNSCachingBeforeRedirectSupported
         internal bool IsJsonSupportEnabled = false;
 
         // User Agent Flag
-        internal bool IsUserAgentEnabled = true;
+        internal bool IsUserAgentSupportEnabled = true;
```
paulmedynski: This field isn't used and can be removed.  Same in the netfx file.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -1417,7 +1417,7 @@ private void Login(ServerInfo server, TimeoutTimer timeout, string newPassword,
             requestedFeatures |= TdsEnums.FeatureExtension.SQLDNSCaching;
             requestedFeatures |= TdsEnums.FeatureExtension.JsonSupport;
             requestedFeatures |= TdsEnums.FeatureExtension.VectorSupport;
-
+        
         #if DEBUG
```
paulmedynski: At this point, we should always be sending the UserAgent.  This `#if` can be removed.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.EndPoint/ITDSServerSession.cs
```diff
@@ -93,5 +93,10 @@ public interface ITDSServerSession
         /// Indicates whether the client supports Vector column type
         /// </summary>
         bool IsVectorSupportEnabled { get; set; }
+
+        /// <summary>
+        /// Indicates whether the client supports UserAgent Feature Extension
+        /// </summary>
+        bool IsUserAgentSupportEnabled { get; set; }
```
paulmedynski, line 100: Nothing is setting or getting this field.  Either remove it, or check it somewhere.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/GenericTDSServerSession.cs
```diff
@@ -124,6 +124,11 @@ public class GenericTDSServerSession : ITDSServerSession
         /// </summary>
         public bool IsVectorSupportEnabled { get; set; }
 
+        /// <summary>
+        /// Indicates whether this session supports User Agent Feature Extension
+        /// </summary>
+        public bool IsUserAgentSupportEnabled { get; set; }
```
paulmedynski, line 130: Nothing is setting or getting this field. Either remove it, or check it somewhere.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -192,7 +193,14 @@ internal void TdsLogin(
 
                 int feOffset = length;
                 // calculate and reserve the required bytes for the featureEx
-                length = ApplyFeatureExData(requestedFeatures, recoverySessionData, fedAuthFeatureExtensionData, useFeatureExt, length);
+                length = ApplyFeatureExData(
```
paulmedynski: Just an observation - nothing for you to address here right now.

This approach of pre-calculating the packet length is inefficient.  We're making 2 passes over the data to be written.  Instead, we should be writing everything to the buffer once, leaving a hole where the header length goes, calculating the length as we write the data, and then going back at the end to fill in the header's length field.

I wonder if we're similarly inefficient for the other TDS packets/messages we write?

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlConnectionBasicTests.cs
```diff
@@ -620,5 +620,116 @@ public void TestConnWithVectorFeatExtVersionNegotiation(bool expectedConnectionR
...
+            bool loginFound = false;
+            bool responseFound = false;
+
+            // Inspect what the client sends in the LOGIN7 packet
+            server.OnLogin7Validated = loginToken =>
+            {
+                var token = loginToken.FeatureExt
+                               .OfType<TDSLogin7GenericOptionToken>()
+                               .FirstOrDefault(t => t.FeatureID == TDSFeatureID.UserAgentSupport);
+                if (token != null)
```
paulmedynski: The test must fail if no UserAgent feature extension token is found:

`Assert.NotNull(token);`

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlConnectionBasicTests.cs
```diff
@@ -620,5 +620,116 @@ public void TestConnWithVectorFeatExtVersionNegotiation(bool expectedConnectionR
...
+                {
+                    Assert.Equal((byte)TDSFeatureID.UserAgentSupport, (byte)token.FeatureID);
+                    Assert.Equal(0x1, token.Data[0]);
+                    loginFound = true;
+                }
+            };
+
+            // Inspect whether the server ever sends back an ACK
+            server.OnAuthenticationResponseCompleted = response =>
```
paulmedynski: This is not necessary.  It is just testing that our test TDS Server sends an ack, which we know because we implemented it.

If you want to test anything related to a server response ack, you should confirm that the log message from `SqlInternalConnectionTds.cs:3006` is emitted.

paulmedynski: Also, this test claims to expect no ack from the server, so why is this here at all?

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlConnectionBasicTests.cs
```diff
@@ -620,5 +620,116 @@ public void TestConnWithVectorFeatExtVersionNegotiation(bool expectedConnectionR
...
+            bool loginFound = false;
+            bool responseFound = false;
+
+            // Observe what the client sends in LOGIN7
+            server.OnLogin7Validated = loginToken =>
+            {
+                var token = loginToken.FeatureExt
+                                      .OfType<TDSLogin7GenericOptionToken>()
+                                      .FirstOrDefault(t => t.FeatureID == TDSFeatureID.UserAgentSupport);
+                if (token != null)
```
paulmedynski: Again, the test must fail if the driver didn't send a UserAgent feature extension.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlConnectionBasicTests.cs
```diff
@@ -620,5 +620,116 @@ public void TestConnWithVectorFeatExtVersionNegotiation(bool expectedConnectionR
...
+                {
+                    Assert.Equal((byte)TDSFeatureID.UserAgentSupport, (byte)token.FeatureID);
+                    loginFound = true;
+                }
+            };
+
+            // Verify the server sent back an ACK for UserAgentSupport
+            server.OnAuthenticationResponseCompleted = response =>
```
paulmedynski: As above, this is unnecessary.  If you want to verify if/how the driver handles a UserAgent ack, you must check for side-effects, which in this case is the log message.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/GenericTDSServer.cs
```diff
@@ -54,16 +54,41 @@ public delegate void OnAuthenticationCompletedDelegate(
         /// </summary>
         public bool EnableVectorFeatureExt { get; set; } = false;
 
+        /// <summary>
+        /// Property for enabling user agent feature extension.
+        /// </summary>
+        public bool EnableUserAgentFeatureExt { get; set; } = true;
```
paulmedynski: This flag isn't necessary.  Our test server should always capture the UserAgent version and data blob.  The test should then check those captured values if it wants to.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/GenericTDSServer.cs
```diff
@@ -287,7 +312,14 @@ public virtual TDSMessageCollection OnLogin7Request(ITDSServerSession session, T
                                 }
                                 break;
                             }
-
+                        case TDSFeatureID.UserAgentSupport:
+                            {
+                                if (EnableUserAgentFeatureExt)
+                                {
+                                    _clientSupportedUserAgentFeatureExtVersion = ((TDSLogin7GenericOptionToken)option).Data[0];
```
paulmedynski: We should be capturing the JSON data blob as well.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlConnectionBasicTests.cs
```diff
@@ -620,5 +620,116 @@ public void TestConnWithVectorFeatExtVersionNegotiation(bool expectedConnectionR
...
+            server.OnLogin7Validated = loginToken =>
+            {
+                var token = loginToken.FeatureExt
+                               .OfType<TDSLogin7GenericOptionToken>()
+                               .FirstOrDefault(t => t.FeatureID == TDSFeatureID.UserAgentSupport);
+                if (token != null)
+                {
+                    Assert.Equal((byte)TDSFeatureID.UserAgentSupport, (byte)token.FeatureID);
+                    Assert.Equal(0x1, token.Data[0]);
+                    loginFound = true;
```
paulmedynski: Need to verify that the UserAgent version and JSON data blob captured by the test server are identical to the one we (should have) sent.  Here and the next test.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParser.cs
samsharma2700: I will leave a comment here so that we can follow up and investigate a bit later.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlConnectionBasicTests.cs
samsharma2700: Updated the test as per our discussion and also left a comment to verify this via log message as well.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS/TDSFeatureID.cs
```diff
@@ -29,6 +29,11 @@ public enum TDSFeatureID : byte
         /// </summary>
         VectorSupport = 0x0E,
 
+        /// <summary>
+        /// User Agent Support
+        /// </summary>
+        UserAgentSupport = 0x0F,
```
apoorvdeshmukh: The design spec mentions that this identifier needs to be confirmed. 
Is this the final identifier? TDS spec should be updated for the same.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlConnectionBasicTests.cs
```diff
@@ -620,5 +618,117 @@ public void TestConnWithVectorFeatExtVersionNegotiation(bool expectedConnectionR
...
+
+            // Note: Accessing UserAgentInfo via Reflection.
+            // We cannot use InternalsVisibleTo here because making internals visible to FunctionalTests
+            // causes the *.TestHarness.cs stubs to clash with the real internal types in SqlClient.
+            var asm = typeof(SqlConnection).Assembly;
+            var userAgentInfoType =
+                asm.GetTypes().FirstOrDefault(t => string.Equals(t.Name, "UserAgentInfo", StringComparison.Ordinal)) ??
+                asm.GetTypes().FirstOrDefault(t => t.FullName?.EndsWith(".UserAgentInfo", StringComparison.Ordinal) == true);
+
+            Assert.True(userAgentInfoType != null,
```
paulmedynski: `Assert.NotNull(userAgentInfoType)`

Same with the prop below.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlConnectionBasicTests.cs
paulmedynski: Please remove this entire block.  We shouldn't be testing our mock TDS Server behaviour.  Leave the TODO on line 667, but remove lines 668-687.

You also don't need the `expectAck` argument, the `responseFound` variable, or any checks related to those.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlConnectionBasicTests.cs
```diff
@@ -620,5 +618,117 @@ public void TestConnWithVectorFeatExtVersionNegotiation(bool expectedConnectionR
...
+            // Try to get the property
+            var prop = userAgentInfoType.GetProperty("UserAgentCachedJsonPayload",
+                BindingFlags.Static | BindingFlags.Public | BindingFlags.NonPublic);
+
+            Assert.True(prop != null,
+                "Unable to find property 'UserAgentCachedJsonPayload' on UserAgentInfo");
+
+            ReadOnlyMemory<byte> cachedPayload = (ReadOnlyMemory<byte>)prop.GetValue(null)!;
+
+            Assert.False(cachedPayload.IsEmpty);
```
paulmedynski: You can just:

`Assert.Equal(cachedPayload, observedJsonBytes);`

src/Microsoft.Data.SqlClient/tests/FunctionalTests/Microsoft.Data.SqlClient.FunctionalTests.csproj
```diff
@@ -92,6 +92,8 @@
     <PackageReference Include="Microsoft.SqlServer.Types" />
     <PackageReference Include="Newtonsoft.Json" />
     <PackageReference Include="System.Security.Cryptography.Pkcs" />
+    <PackageReference Include="System.Configuration.ConfigurationManager" />
```
paulmedynski, line 96: Why were these added?

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/GenericTDSServer.cs
```diff
@@ -314,10 +309,7 @@ public virtual TDSMessageCollection OnLogin7Request(ITDSServerSession session, T
                             }
                         case TDSFeatureID.UserAgentSupport:
                             {
-                                if (EnableUserAgentFeatureExt)
-                                {
-                                    _clientSupportedUserAgentFeatureExtVersion = ((TDSLogin7GenericOptionToken)option).Data[0];
-                                }
+                                _clientSupportedUserAgentFeatureExtVersion = ((TDSLogin7GenericOptionToken)option).Data[0];
```
paulmedynski: I don't think this variable is used anywhere.  Do you need this case at all?

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS/TDSFeatureID.cs
paulmedynski: This value should be updated to 0x10.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/GenericTDSServer.cs
paulmedynski: We have 2 options for inspecting the UserAgent feature extension data:

1. Add 2 public variables to this TDS Server, save the version and paylod to them here, remove the OnLogin7Validated check, and just assert the public variables directly.
2. Remove this case entirely, don't save any User Agent values in TDS Server, and keep the OnLogin7Validated check.

I think # 2 is my preferred, since it doesn't add any extra work to TDS Server that nobody else will care about.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -3038,7 +3035,14 @@ internal void OnFeatureExtAck(int featureId, byte[] data)
                         IsVectorSupportEnabled = true;
                         break;
                     }
-
+                case TdsEnums.FEATUREEXT_USERAGENT:
+                    {
+                        // TODO: define comment, TDS spec doesnot define an ack 
+                        // Unexpected ack from server but we ignore it entirely
+                        // TODO for tfuture if we can find and verify this log message
```
samsharma2700: Typos. Will also update the comment for more details.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/Microsoft.Data.SqlClient.FunctionalTests.csproj
samsharma2700: Build issues. Will double check if the need for this package was a side-effect of trying to make internals visible to functional tests.

samsharma2700: Yep it was, will remove these.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlConnectionBasicTests.cs
samsharma2700: Will update this.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlConnectionBasicTests.cs
samsharma2700: Can't leave a custom failure message with `Assert.NotNull()`.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlConnectionBasicTests.cs
samsharma2700: `Assert.True(...SequenceEqual(...))` is more performant than `Assert.Equal(....ToArray(), ...ToArray())` but shouldn't matter much in a test. Will change.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/GenericTDSServer.cs
samsharma2700: Yes, will remove this, no use for it now.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS/TDSFeatureID.cs
samsharma2700: Nice catch!

paulmedynski (dismissed):
Yay - all good!

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlConnectionBasicTests.cs
paulmedynski: True, but it's pretty obvious what the assert is asserting for a NotNull() check.  I generally prefer _not_ to include failure messages for assertions.  The code probably has better commentary near the assertion, and the failure will tell you the exact line of code.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlConnectionBasicTests.cs
paulmedynski: I prefer the tests to be obviously correct, and Assert.Equal() is easier for a reader to understand IMO.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParser.cs
cheenamalhotra: Has this been addressed?

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/GenericTdsServer.cs
```diff
@@ -654,6 +669,30 @@ protected virtual TDSMessageCollection OnAuthenticationCompleted(ITDSServerSessi
                 }
             }
 
+            // If tests request it, force an ACK for UserAgentSupport with no negotiation
+            if (EmitUserAgentFeatureExtAck)
```
cheenamalhotra: Can you please rewrite this large method to split these conditional blocks into individual methods?

cheenamalhotra: Should you not be reading session instance value here as done above for other feature ext tokens?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -9306,6 +9317,11 @@ private int ApplyFeatureExData(TdsEnums.FeatureExtension requestedFeatures,
                         length += WriteVectorSupportFeatureRequest(write);
                     }
 
+                    if ((requestedFeatures & TdsEnums.FeatureExtension.UserAgent) != 0)
```
paulmedynski: We need to write the User Agent FE first.  Move this up to line 9274 and add a comment explaining why it must appear first.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/GenericTdsServer.cs
paulmedynski: The `EmitUserAgentFeatureExtAck` flag is currently server-wide.  If there are other flags that are per session, then perhaps this flag should be per session as well.

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionTests.cs
```diff
@@ -828,5 +828,86 @@ public void TestConnWithVectorFeatExtVersionNegotiation(bool expectedConnectionR
...
+            byte observedVersion = 0;
+            byte[] observedJsonBytes = Array.Empty<byte>();
+
+            // Inspect what the client sends in the LOGIN7 packet
+            server.OnLogin7Validated = loginToken =>
+            {
+                var token = loginToken.FeatureExt
+                                      .OfType<TDSLogin7GenericOptionToken>()
+                                      .FirstOrDefault(t => t.FeatureID == TDSFeatureID.UserAgentSupport);
+                if (token == null)
```
paulmedynski: This should be:
```
Assert.NotNull(token);
```
The driver is supposed to always send the User Agent FE.

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionTests.cs
```diff
@@ -828,5 +828,86 @@ public void TestConnWithVectorFeatExtVersionNegotiation(bool expectedConnectionR
...
+                {
+                    return;
+                }
+
+                var data = token.Data;
+                if (data == null || data.Length < 2)
```
paulmedynski: As above, this must fail the test:
```
Assert.NotNull(data);
Assert.GreaterThanOrEqual(2, data.Length);
```

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionTests.cs
paulmedynski: I realize that you're checking `loginFound` below, but failing here will allow you to pinpoint why the test failed rather than trying to work backwards from line 891.

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionTests.cs
```diff
@@ -828,5 +828,86 @@ public void TestConnWithVectorFeatExtVersionNegotiation(bool expectedConnectionR
...
+            // Captured from LOGIN7 as parsed by the test server
+            byte observedVersion = 0;
+            byte[] observedJsonBytes = Array.Empty<byte>();
+
+            // Inspect what the client sends in the LOGIN7 packet
+            server.OnLogin7Validated = loginToken =>
+            {
+                var token = loginToken.FeatureExt
+                                      .OfType<TDSLogin7GenericOptionToken>()
+                                      .FirstOrDefault(t => t.FeatureID == TDSFeatureID.UserAgentSupport);
```
paulmedynski: We should confirm that the User Agent FE is the first FE in the LOGIN7 packet.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsEnums.cs
```diff
@@ -242,7 +242,7 @@ public enum EnvChangeType : byte
         public const byte FEATUREEXT_JSONSUPPORT = 0x0D;
         public const byte FEATUREEXT_VECTORSUPPORT = 0x0E;
         // TODO: re-verify if this byte competes with another feature
```
benrr101: Is it still necessary to verify if it competes with another feature?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParser.SSPI.cs
```diff
@@ -192,7 +193,14 @@ internal void TdsLogin(
 
                 int feOffset = length;
                 // calculate and reserve the required bytes for the featureEx
-                length = ApplyFeatureExData(requestedFeatures, recoverySessionData, fedAuthFeatureExtensionData, useFeatureExt, length);
+                length = ApplyFeatureExData(
+                    requestedFeatures, 
+                    recoverySessionData, 
+                    fedAuthFeatureExtensionData,
+                    UserAgentInfo.UserAgentCachedJsonPayload.ToArray(),
```
benrr101: Be careful with .ToArray() - this is likely duplicating the array

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParser.cs
samsharma2700: Done

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionTests.cs
samsharma2700: Done

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionTests.cs
samsharma2700: Done

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionTests.cs
samsharma2700: Done

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsEnums.cs
samsharma2700: Removed the stray comment.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParser.SSPI.cs
samsharma2700: This is how the existing function consumes it. Leaving it as it is for now.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParser.cs
samsharma2700: Yes, for now added a note since this is not related to this change.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/GenericTdsServer.cs
samsharma2700: Done the refactoring

samsharma2700: Done

paulmedynski (approved):

PR #3607, opened 2025-09-10
Increase macOS SQL Server setup timeout

paulmedynski:
## Description

- Increased macOS SQL Server setup timeout to 2 minutes to avoid unnecessary test failures when the host VM is just slow.


benrr101 (approved):
🤷‍♂️ eh sounds good

Issue #3029, opened 2024-11-20
InternalsVisibleTo In Nonstandard Location


benrr101:
**Description**: In SqlConnection, there is an `[assembly:InternalsVisibleTo...]` attribute tucked at the beginning of the class. This is more/less in a non-standard location and makes was an unexpected surprise when I spotted it.

**Fix**: Move all InternalsVisibleTo attributes to AssemblyInfo.cs

edwardneal:
I agree. The current InternalsVisibleTo attribute is:
```csharp
[assembly: InternalsVisibleTo("System.Data.DataSetExtensions, PublicKey=" + Microsoft.Data.SqlClient.AssemblyRef.EcmaPublicKeyFull)]
// NOTE: The current Microsoft.VSDesigner editor attributes are implemented for System.Data.SqlClient, and are not publicly available.
// New attributes that are designed to work with Microsoft.Data.SqlClient and are publicly documented should be included in future.
```

Which internals does System.Data.DataSetExtensions need to use though? I read the comment as an indicator that the VS designer is currently linked against System.Data.SqlClient, so the InternalsVisibleTo attribute isn't in use. It might be worth reviewing the internal classes that System.Data.DataSetExtensions needs to reference and considering what public API surface could replace those references; without that, we can't make any code style changes to internal classes without risking breaking Visual Studio.

edwardneal:
After a little investigation, I think we can remove this completely.

The exact fully-qualified assembly name which is actually supplied to `InternalsVisibleTo` is
`System.Data.DataSetExtensions, PublicKey=00000000000000000400000000000000`

This public key is the [ECMA key](https://learn.microsoft.com/en-gb/archive/blogs/shawnfa/whats-the-deal-with-the-ecma-key), and it confirms that we're dealing with the version of System.Data.DataSetExtensions which is part of the .NET Framework, rather than a unique NuGet package. We have the reference source for this [here](https://github.com/microsoft/referencesource/tree/main/System.Data.DataSetExtensions).

[...]

PR #378, opened 2020-01-12
perf: Cache common callbacks with a single state parameter

Wraith2:
In a couple of common paths when reading data there are async context captures generated which use a single variable from the enclosing scope. These can be optimized so that a static delegate to the implementation is stored and the variable passed as state, this removes the closure allocation.

PR #3608, opened 2025-09-11
Merge | SqlConnection.cs

edwardneal:
## Description

This merges the SqlConnection class. I recommend reviewing this commit-by-commit with whitespace changes disabled; the underlying changes are pretty minor.

Earlier PRs have already addressed all of the breaking changes, so each commit handles a fairly small change needed to render the files identical. By b966f7a6876cc2e23346cfaa06647411aee6defd, these files are ready to merge.

I've tried to avoid functional changes. As a result, I noticed a handful of sections which enabled WindowsIdentity impersonation and have conditionally compiled them for netfx. These would probably work for netcore, but I don't have ready access to a domain from Windows and Linux and so can't test them.

On one final point, this doesn't deal with the SqlConnectionHelper.cs file. That file remains target-specific, it's too difficult to merge a sensible without dealing with the main SqlConnection.cs file first.

## Issues

Relates to #1261.
Ports #378 and #379 from netcore to netfx.

## Testing

Automated tests run.

Could someone run CI please?

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnection.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnection.cs
@@ -24,12 +24,13 @@ using Microsoft.Data.SqlClient.ConnectionPool;
 using Microsoft.Data.SqlClient.Diagnostics;
 using Microsoft.SqlServer.Server;
 #if NETFRAMEWORK
+using System.Runtime.CompilerServices;
 using System.Security.Permissions;
 using System.Security.Principal;
 #endif
 
 #if NETFRAMEWORK
-[assembly: System.Runtime.CompilerServices.InternalsVisibleTo("System.Data.DataSetExtensions, PublicKey=" + Microsoft.Data.SqlClient.AssemblyRef.EcmaPublicKeyFull)] // DevDiv Bugs 92166
+[assembly: InternalsVisibleTo("System.Data.DataSetExtensions, PublicKey=" + Microsoft.Data.SqlClient.AssemblyRef.EcmaPublicKeyFull)] // DevDiv Bugs 92166
 // NOTE: The current Microsoft.VSDesigner editor attributes are implemented for System.Data.SqlClient, and are not publicly available.
 // New attributes that are designed to work with Microsoft.Data.SqlClient and are publicly documented should be included in future.
 #endif
@@ -930,7 +931,7 @@ namespace Microsoft.Data.SqlClient
         {
             get
             {
-                if ((State & (ConnectionState.Open | ConnectionState.Executing | ConnectionState.Fetching)) > 0)
+                if ((State & (ConnectionState.Open | ConnectionState.Executing | ConnectionState.Fetching)) != 0)
                 {
                     return GetOpenTdsConnection().ServerProcessId;
                 }


edwardneal:
Thanks - I've just spotted the problem, could someone re-run CI please?

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnection.cs
```diff
@@ -279,9 +279,9 @@ internal static List<string> GetColumnEncryptionSystemKeyStoreProvidersNames()
         {
             if (s_systemColumnEncryptionKeyStoreProviders.Count > 0)
             {
-                return new List<string>(s_systemColumnEncryptionKeyStoreProviders.Keys);
+                return [.. s_systemColumnEncryptionKeyStoreProviders.Keys];
```
benrr101: 👀 👀 👀 I didn't know C# had an operator like that now

[no action required]

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnection.cs
```diff
@@ -921,8 +919,14 @@ public override string ServerVersion
         [DesignerSerializationVisibility(DesignerSerializationVisibility.Hidden)]
         public int ServerProcessId
         {
-            get => State.Equals(ConnectionState.Open) | State.Equals(ConnectionState.Executing) | State.Equals(ConnectionState.Fetching) ?
-                GetOpenTdsConnection().ServerProcessId : 0;
+            get
+            {
+                if ((State & (ConnectionState.Open | ConnectionState.Executing | ConnectionState.Fetching)) > 0)
```
benrr101: Is this the same as how it's written in netcore?
I'd say either use the `x is a or b or c` (which is equivalent since the original used x._equals_ for each one), or make sure to use `!= 0` (since enum is backed by an int and could overflow to a negative - even if it's unlikely based on the values currently in the enum, it's good practice)

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnection.cs
```diff
@@ -1226,7 +1230,7 @@ public SqlTransaction BeginTransaction(IsolationLevel iso, string transactionNam
                 bool isFirstAttempt = true;
                 do
                 {
-                    transaction = GetOpenConnection().BeginSqlTransaction(iso, transactionName, isFirstAttempt); // do not reconnect twice
+                    transaction = GetOpenTdsConnection().BeginSqlTransaction(iso, transactionName, isFirstAttempt); // do not reconnect twice
```
benrr101: [no action required] Since this is the only kind of internal connection, it'd be good to drop Tds from this once the types are consolidated.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnection.cs
```diff
@@ -1768,7 +1769,7 @@ private Task InternalOpenAsync(SqlConnectionOverrides overrides, CancellationTok
                         CancellationTokenRegistration registration = new CancellationTokenRegistration();
                         if (cancellationToken.CanBeCanceled)
                         {
-                            registration = cancellationToken.Register(() => completion.TrySetCanceled());
+                            registration = cancellationToken.Register(s_openAsyncCancel, completion);
```
benrr101: [no action required] Wouldn't it be as performant or more if we omit the Action<object> and just directly use the method?

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnection.cs
```diff
@@ -27,9 +24,12 @@
 using Microsoft.Data.SqlClient.ConnectionPool;
 using Microsoft.Data.SqlClient.Diagnostics;
 using Microsoft.SqlServer.Server;
+#if NETFRAMEWORK
+using System.Security.Permissions;
+#endif
 
 #if NETFRAMEWORK
-[assembly: InternalsVisibleTo("System.Data.DataSetExtensions, PublicKey=" + Microsoft.Data.SqlClient.AssemblyRef.EcmaPublicKeyFull)] // DevDiv Bugs 92166
+[assembly: System.Runtime.CompilerServices.InternalsVisibleTo("System.Data.DataSetExtensions, PublicKey=" + Microsoft.Data.SqlClient.AssemblyRef.EcmaPublicKeyFull)] // DevDiv Bugs 92166
```
benrr101: ... Why make this line more verbose?

benrr101 (approved):
Overall looks good, got a couple questions, but glad to move forward with it as-is

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnection.cs
edwardneal: It's not exactly the same, but the documentation here is confusing. The [`ConnectionState`](https://learn.microsoft.com/en-us/dotnet/api/system.data.connectionstate?view=net-9.0) enum is marked as a set of flags, but the remarks indicate that it's not designed to be used as a set of flags. One target treated it as an enumeration, the other as a set of fixed values - I've merged by picking the version which would cover both bases.

I'll correct the `> 0` to `!= 0`.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnection.cs
edwardneal: Fixed and committed!

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlConnection.cs
benrr101: Sounds good to me 👍 
Yeah it seems like the connection state shouldn't be more than one state at a time - so [Flags] doesn't seem right. But that's outside the scope of this.

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -919,6 +919,9 @@ public void RegisterColumnEncryptionKeyStoreProvidersOnConnection(System.Collect
         public int CommandTimeout { get { throw null; } }
         /// <include file='../../../../doc/snippets/Microsoft.Data.SqlClient/SqlConnection.xml' path='docs/members[@name="SqlConnection"]/ConnectionString/*'/>
         [System.ComponentModel.DefaultValueAttribute("")]
+#pragma warning disable CS0618
+        [System.ComponentModel.RecommendedAsConfigurableAttribute(true)]
```
paulmedynski, line 923: Why was this added?  It is marked as obsolete:

https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.recommendedasconfigurableattribute

src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
```diff
@@ -812,7 +815,6 @@
     </Compile>
     
     <Compile Include="Microsoft\Data\SqlClient\SqlCommand.netcore.cs" />
-    <Compile Include="Microsoft\Data\SqlClient\SqlConnection.cs" />
```
paulmedynski, line 815: This file didn't exist before its removal here.  How was this building?

tools/targets/NotSupported.targets
```diff
@@ -8,6 +8,7 @@
     <!-- Not supported sources are created from the ref assembly, we currently don't produce finalizers in dummmy assemblies, so we disable ApiCompat to not fail. -->
     <RunApiCompat>false</RunApiCompat>
     <GenerateDocumentationFile Condition="'$(OSGroup)' != 'Windows_NT'">false</GenerateDocumentationFile>
+    <NoWarn>$(NoWarn);CS0618</NoWarn>
```
paulmedynski, line 11: Why are we generating C# code that uses obsolete APIs?  Is this because we added the obsolete `System.ComponentModel.RecommendedAsConfigurableAttribute` to the ref APIs?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnection.cs
```diff
@@ -23,14 +23,28 @@
...
 using Microsoft.Data.SqlClient.Diagnostics;
 using Microsoft.SqlServer.Server;
+#if NETFRAMEWORK
+using System.Runtime.CompilerServices;
+using System.Security.Permissions;
+using System.Security.Principal;
+#endif
 
+#if NETFRAMEWORK
+[assembly: InternalsVisibleTo("System.Data.DataSetExtensions, PublicKey=" + Microsoft.Data.SqlClient.AssemblyRef.EcmaPublicKeyFull)] // DevDiv Bugs 92166
```
paulmedynski, line 33: Yikes - does this mean that System.Data.DataSetExtensions has full access to all MDS internals?  Is this still necessary?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnection.cs
```diff
@@ -455,41 +474,39 @@ public bool StatisticsEnabled
             }
             set
             {
+                if (value)
                 {
```
paulmedynski, line 478: I think this whole diff chunk is just you removing this unnecessary { } block, with no other changes.  Is that correct?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnection.cs
```diff
@@ -1081,7 +1104,11 @@ private void CheckAndThrowOnInvalidCombinationOfConnectionOptionAndAccessToken(S
             // Check if the usage of AccessToken has the conflict with credential
             if (_credential != null)
             {
+#if NET
                 throw ADP.InvalidMixedUsageOfCredentialAndAccessToken();
+#else
+                throw ADP.InvalidMixedUsageOfAccessTokenAndCredential();
```
paulmedynski, line 1110: Do these functions really produce a different exception?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
edwardneal: I'm not sure what you mean - this was the target-specific SqlConnection.cs file. We remove the target-specific files from both projects, then add the version under $(CommonSourceRoot) above.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnection.cs
edwardneal: It does, but only on netfx.

I don't personally think it's necessary - I think it was ported across from the netfx version of System.Data.SqlClient. A check of System.Data.DataSetExtensions' dependencies in netfx shows that it doesn't reference MDS, and given how high the bar for netfx servicing is, it seems unlikely that that will ever change.

Maybe the DevDiv bug (92166) will have some context here? If there's confirmation that it's unnecessary then I'll remove it and we can close #3029 at the same time.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnection.cs
edwardneal: Yes - it was changed in 26c732a6994c15d37a594a46f473a8bf7a4162b7.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnection.cs
edwardneal: Sadly. It's just the exception message though, so might be okay for a major version release.

https://github.com/dotnet/SqlClient/blob/2655083e3ed14645d0190cad25fed8dae4881134/src/Microsoft.Data.SqlClient/src/Resources/Strings.resx#L240-L245

tools/targets/NotSupported.targets
edwardneal: It is. I added it to netcore to bring the public API surface into sync with netfx. If you'd prefer, I can change both API surfaces to reference `SettingsBindableAttribute` instead?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
paulmedynski: Ahh I see - this file was git moved to the common one.

tools/targets/NotSupported.targets
paulmedynski: No let's do the minimal to make .NET and .NET Framework similar.  We can address this after the merge work is complete.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnection.cs
paulmedynski: Punt to the post-merge world.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnection.cs
paulmedynski: Wild.  Exception text messages aren't (or certainly shouldn't be) part of the public API, so we can change them as we see fit.  There will be lots of post-merge auditing for this kind of thing.

PR #3609, opened 2025-09-11
Optimization: Use Environment.TickCount for SqlStatistics execution timing

SAY14489:
## Description

This change optimizes the performance of SQL execution timing by replacing expensive DateTime operations with lightweight Environment.TickCount calls in SqlStatistics. The optimization targets the RequestExecutionTimer() and ReleaseAndUpdateExecutionTimer() methods, which are frequently called during database operations. This issue was exposed by profilers which showed that these two calling methods made up a significant portion of DateTime.UtcNow CPU usage.

### Performance improvement: 
| Method                      | Mean      | Error     | StdDev    | Ratio | RatioSD | Allocated | Alloc Ratio |
|---------------------------- |----------:|----------:|----------:|------:|--------:|----------:|------------:|
| DateTimeUtcNowToFileTimeUtc | 57.243 ns | 0.9307 ns | 0.8250 ns |  1.00 |    0.02 |         - |          NA |
| EnvironmentTickCount        |  1.272 ns | 0.0533 ns | 0.1328 ns |  0.02 |    0.00 |         - |          NA |

^ Done Through BenchmarkDotNet
Benchmarking shows an approximately 55x performance improvement (from ~57ns to ~1ns per timing operation) while maintaining adequate accuracy for database operation monitoring.

### Changes made:

- Added ADP.FastTimerCurrent() method using Environment.TickCount
- Modified SqlStatistics execution timing methods to use the optimized timer
- Implemented wraparound handling for 32-bit integer overflow
- Updated timing units to maintain consistency with existing millisecond reporting

### Backwards compatibility: 
Fully maintained. All existing APIs, return values, and units remain unchanged. The optimization is internal implementation only.

### Scope: 
This change specifically optimizes execution timing measurement without affecting timeout logic or other timing-dependent functionality in the codebase.

## Issues

Performance optimization. Addresses inefficient use of DateTime.UtcNow.ToFileTimeUtc() for high-frequency elapsed time measurements.

### Manual testing performed:

- Created test application measuring timing accuracy across different durations (100ms to 2000ms)
- Verified timing accuracy using Stopwatch as baseline for actual elapsed time measurement and compared against DateTime.UtcNow.ToFileTimeUtc()
- Confirmed proper accumulation of execution times across multiple measurement periods

### Accuracy results: 
Testing showed the TickCount optimization maintains acceptable accuracy for duration timing, with error rates under 3% compared to StopWatch baseline for operations lasting 100ms or longer. It is a slight loss in accuracy compared to DateTime.UtcNow.ToFileTimeUtc(), which I believe is due to the lower precision, but for the use case and performance improvement of 55x, it seems like a worthwhile switch.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -158,23 +158,13 @@ dotnet_code_quality.CA2100.excluded_type_names_with_derived_types = Microsoft.Da
 dotnet_diagnostic.xUnit1031.severity=none
 dotnet_diagnostic.xUnit1030.severity=none
 
-[*.{csproj,vcxproj,vcxproj.filters,proj,nativeproj,locproj}]
-indent_size = 2
-
-# Xml build files
-[*.builds]
-indent_size = 2
-
 # Xml files
-[*.{xml,stylecop,resx,ruleset}]
-indent_size = 2
-
-# Xml config files
-[*.{props,targets,config,nuspec}]
+[*.{xml,csproj,stylecop,resx,ruleset,props,targets,config,nuspec}]
 indent_size = 2
 
 # Shell scripts
 [*.sh]
 end_of_line = lf
+
 [*.{cmd, bat}]
 end_of_line = crlf
--- a/eng/pipelines/common/templates/steps/publish-symbols-step.yml
+++ b/eng/pipelines/common/templates/steps/publish-symbols-step.yml
@@ -12,7 +12,6 @@ parameters:
 
   - name: publishSymbols
     type: string
-    default: '$(PublishSymbols)'
 
   - name: symbolsVersion
     type: string
--- a/eng/pipelines/dotnet-sqlclient-signing-pipeline.yml
+++ b/eng/pipelines/dotnet-sqlclient-signing-pipeline.yml
@@ -38,7 +38,7 @@ parameters: # parameters are shown up in ADO UI in a build queue time
 - name: 'debug'
   displayName: 'Enable debug output'
   type: boolean
-  default: true
+  default: false
 
 - name: publishSymbols
   displayName: 'Publish symbols'
--- a/eng/pipelines/steps/compound-nuget-pack-step.yml
+++ b/eng/pipelines/steps/compound-nuget-pack-step.yml
@@ -56,4 +56,3 @@ steps:
                     -Version ${{ parameters.packageVersion }}
                     -OutputDirectory ${{ parameters.outputDirectory }}
                     -Properties "COMMITID=$(Build.SourceVersion);Configuration=${{ parameters.buildConfiguration }};ReferenceType=${{ parameters.referenceType }}"
-
--- a/src/Directory.Packages.props
+++ b/src/Directory.Packages.props
@@ -1,71 +1,112 @@
 <Project>
-    <PropertyGroup>
-        <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
-        <CentralPackageTransitivePinningEnabled>true</CentralPackageTransitivePinningEnabled>
-    </PropertyGroup>
-    <!-- NetFx project dependencies -->
-    <ItemGroup>
-        <PackageVersion Include="Microsoft.Data.SqlClient.SNI" Version="6.0.2" />
-        <PackageVersion Include="System.Buffers" Version="4.5.1" />
-        <PackageVersion Include="System.Diagnostics.DiagnosticSource" Version="8.0.1" />
-        <PackageVersion Include="System.Memory" Version="4.5.5" />
-        <PackageVersion Include="System.Data.Common" Version="4.3.0" />
-        <PackageVersion Include="System.Text.Encodings.Web" Version="8.0.0" />
-        <PackageVersion Include="System.ValueTuple" Version="4.6.1" />
-    </ItemGroup>
-    <!-- NetFx and NetCore project dependencies -->
-    <ItemGroup>
-        <PackageVersion Include="Azure.Identity" Version="1.14.2" />
-        <PackageVersion Include="Microsoft.IdentityModel.JsonWebTokens" Version="7.7.1" />
-        <PackageVersion Include="Microsoft.IdentityModel.Protocols.OpenIdConnect" Version="7.7.1" />
-        <PackageVersion Include="System.Runtime.InteropServices.RuntimeInformation" Version="4.3.0" />
-    </ItemGroup>
-    <!-- NetCore project dependencies -->
-    <ItemGroup>
-        <PackageVersion Include="Microsoft.Data.SqlClient.SNI.runtime" Version="6.0.2" />
-        <PackageVersion Include="Microsoft.SqlServer.Server" Version="1.0.0" />
-    </ItemGroup>
-    <!-- AKV Provider project dependencies -->
-    <ItemGroup>
-        <PackageVersion Include="Azure.Core" Version="[1.47.1,2.0.0)" />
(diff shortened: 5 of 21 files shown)


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlStatistics.cs
```diff
@@ -137,10 +137,16 @@ internal void RequestNetworkServerTimer()
...
-            if (_startExecutionTimestamp > 0)
+            if (_startExecutionTimestamp.HasValue)
             {
-                _executionTime += (ADP.TimerCurrent() - _startExecutionTimestamp);
-                _startExecutionTimestamp = 0;
+                long elapsed = ADP.FastTimerCurrent() - _startExecutionTimestamp.Value;
+                // Wraparound logic for timer
+                if (elapsed < 0)
+                {
+                    elapsed += (long)uint.MaxValue + 1;
```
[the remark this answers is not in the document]
SAY14489: 0x100000000L covers the full range of Environment.TickCount and is equivalent to what I have, which is (long)uint.MaxValue + 1. (long)int.MaxValue + 1 would only cover half this range and would not be a proper wrap around.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlStatistics.cs
```diff
@@ -108,7 +108,7 @@ internal IDictionary GetDictionary()
                 { "UnpreparedExecs", _unpreparedExecs },
 
                 { "ConnectionTime", ADP.TimerToMilliseconds(_connectionTime) },
-                { "ExecutionTime", ADP.TimerToMilliseconds(_executionTime) },
+                { "ExecutionTime", _executionTime },
```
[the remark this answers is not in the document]
SAY14489: The ExecutionTime entry is correctly implemented. _executionTime stores values in milliseconds from Environment.TickCount, so no conversion is needed, it's already in the correct units for reporting.
The other timing values (ConnectionTime, NetworkServerTime) use ADP.TimerToMilliseconds() because they store values in 100-nanosecond ticks from ADP.TimerCurrent() and need conversion to milliseconds.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlStatistics.cs
```diff
@@ -137,10 +137,16 @@ internal void RequestNetworkServerTimer()
 
         internal void ReleaseAndUpdateExecutionTimer()
         {
-            if (_startExecutionTimestamp > 0)
+            if (_startExecutionTimestamp.HasValue)
             {
-                _executionTime += (ADP.TimerCurrent() - _startExecutionTimestamp);
-                _startExecutionTimestamp = 0;
+                long elapsed = ADP.FastTimerCurrent() - _startExecutionTimestamp.Value;
```
paulmedynski: Can this be simplified to:
```
uint elapsed = (uint)(ADP.FastTimerCurrent() - _startExecutionTimestamp.Value);
_executionTime += elapsed;
_startExecutionTimestamp = null
```

And also we need to account for `_executionTime` wrapping:
```
if (long.MaxValue - elapsed < _executionTime)
{
  _executionTime = long.MaxValue;
}
else
{
  _executionTime += elapsed;
}
```
I assume we never want `_executionTime` to be negative - that doesn't make sense.

paulmedynski: It might also be nice to put this algorithm into an internal function, and then add tests for it in the UnitTests project.  Seems like it's worthwhile testing the edge cases.

SAY14489:
> @SAY14489 please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @dotnet-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @dotnet-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @dotnet-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlStatistics.cs
SAY14489: Hey, thanks for picking this up. 
I committed a new change which follows your advice and uses the (uint) cast as its wrap around logic. I also moved the algorithm to an internal function and added a unit testing file for it which checks different cases.
For the _executionTime wrapping, I was wondering if it would be truly necessary for the use case. To reach long.MaxValue would require approximately 292 million years of accumulated execution time (9.22 × 10^18 milliseconds).

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/SqlBulkCopyTest/CopyAllFromReader.cs
```diff
@@ -52,8 +52,6 @@ public static void Test(string srcConstr, string dstConstr, string dstTable)
 
                             Assert.True(0 < (long)stats["BytesReceived"], "BytesReceived is non-positive.");
                             Assert.True(0 < (long)stats["BytesSent"], "BytesSent is non-positive.");
-                            Assert.True((long)stats["ConnectionTime"] >= (long)stats["ExecutionTime"], "Connection Time is less than Execution Time.");
```
SAY14489, line 55: I had a question on this manual test that was previously flaky. My TickCount optimization creates a timing precision mismatch that occasionally causes the expected hierarchy (ConnectionTime ≥ ExecutionTime ≥ NetworkServerTime) to differ by a few milliseconds. Since ExecutionTime now uses a different timing source than the others, it can measure slightly higher values in some cases, typically within 3-5ms variance.
I'd like to remove these specific timing relationship assertions because the test's core purpose is validating SqlBulkCopy functionality: row counts, data integrity, and operation success. The timing hierarchy checks are implementation details that don't reflect the actual business logic being tested.
The 50x performance improvement justifies this small timing precision trade-off, and I've added dedicated unit tests to validate the TickCount logic itself. Would it be acceptable to remove just these two timing assertions while keeping all the functional validation intact?

paulmedynski: Tests that check timing are generally a bad idea, since they become sensitive to the test environment.  I'm fine with you removing this check.

paulmedynski:
@SAY14489 - Thank you for the great PR, especially your first contribution.  Performance improvements like this that are isolated and easy to verify are always welcome!

uc-msft (approved):

SAY14489:
> Great work!

I’m trying to complete this PR. I just synced this branch with the up to date main since I couldn’t see the merge option, could you please start a pipeline run to fulfill the requirements? Thank You!

PR #3612, opened 2025-09-11
Performance | Use lower-allocation AlwaysEncrypted primitives

edwardneal:
## Description

This builds on #3554, using the lower-allocation AE primitives in `SqlColumnEncryptionCspProvider` and `SqlColumnEncryptionCngProvider`. Benchmarks for this are located in the original PR.

I'm separating these two changes from `SqlColumnEncryptionCertificateStoreProvider` due to the latter's size, but it'll be fairly similar once the use of `X509Certificate2` has been refactored into an `RSA` instance. I'll incorporate the feedback from this PR into the second one.

We make some minor code style changes, largely to remove a few Yoda conditions, replace `== null` with `is null` comparisons and clean up the comments. Besides this though, almost all of the `DecryptColumnEncryptionKey` and `EncryptColumnEncryptionKey` methods are replaced with references to the new AE primitives.

While not the main goal, I also noticed and fixed one case where we were opening a `CngKey` and never disposing it.

## Issues

Builds on #3554.

## Testing

Automated tests pass.

I also created an AE-enabled table in SSMS which used a CNG-based master key, and confirmed that I was able to query values from it. This should confirm that data created outside of SqlClient can now be read successfully.

Could someone run CI please?

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlColumnEncryptionCngProvider.Windows.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlColumnEncryptionCngProvider.Windows.cs
@@ -33,70 +33,6 @@ namespace Microsoft.Data.SqlClient
         /// </summary>
         private const string RSAEncryptionAlgorithmWithOAEP = @"RSA_OAEP";
 
-        /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionCngProvider.xml' path='docs/members[@name="SqlColumnEncryptionCngProvider"]/DecryptColumnEncryptionKey/*' />
-        public override byte[] DecryptColumnEncryptionKey(string? masterKeyPath, string? encryptionAlgorithm, byte[]? encryptedColumnEncryptionKey)
-        {
-            // Validate the input parameters
-            ValidateNonEmptyKeyPath(masterKeyPath, isSystemOp: true);
-
-            if (encryptedColumnEncryptionKey is null)
-            {
-                throw SQL.NullEncryptedColumnEncryptionKey();
-            }
-
-            if (encryptedColumnEncryptionKey.Length == 0)
-            {
-                throw SQL.EmptyEncryptedColumnEncryptionKey();
-            }
-
-            // Validate encryptionAlgorithm
-            ValidateEncryptionAlgorithm(encryptionAlgorithm, isSystemOp: true);
-
-            // Create RSA Provider with the given CNG provider name and key name
-            RSA rsaProvider = CreateRSACngProvider(masterKeyPath, isSystemOp: true);
-            using EncryptedColumnEncryptionKeyParameters cekDecryptionParameters = new(rsaProvider, masterKeyPath, MasterKeyType, KeyPathReference);
-
-            return cekDecryptionParameters.Decrypt(encryptedColumnEncryptionKey);
-        }
-
-        /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionCngProvider.xml' path='docs/members[@name="SqlColumnEncryptionCngProvider"]/EncryptColumnEncryptionKey/*' />
-        public override byte[] EncryptColumnEncryptionKey(string? masterKeyPath, string? encryptionAlgorithm, byte[]? columnEncryptionKey)
-        {
-            // Validate the input parameters
-            ValidateNonEmptyKeyPath(masterKeyPath, isSystemOp: false);
-
-            if (columnEncryptionKey is null)
-            {
-                throw SQL.NullColumnEncryptionKey();
-            }
-
-            if (columnEncryptionKey.Length == 0)
-            {
-                throw SQL.EmptyColumnEncryptionKey();
-            }
-
-            // Validate encryptionAlgorithm
-            ValidateEncryptionAlgorithm(encryptionAlgorithm, isSystemOp: false);
-
-            // Create RSA Provider with the given CNG provider name and key name
-            RSA rsaProvider = CreateRSACngProvider(masterKeyPath, isSystemOp: false);
-            using EncryptedColumnEncryptionKeyParameters cekEncryptionParameters = new(rsaProvider, masterKeyPath, MasterKeyType, KeyPathReference);
-
-            return cekEncryptionParameters.Encrypt(columnEncryptionKey);
-        }
-
-        /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionCngProvider.xml' path='docs/members[@name="SqlColumnEncryptionCngProvider"]/SignColumnMasterKeyMetadata/*' />
-        public override byte[] SignColumnMasterKeyMetadata(string? masterKeyPath, bool allowEnclaveComputations)
-        {
-            throw new NotSupportedException();
-        }
-
-        /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionCngProvider.xml' path='docs/members[@name="SqlColumnEncryptionCngProvider"]/VerifyColumnMasterKeyMetadata/*' />
-        public override bool VerifyColumnMasterKeyMetadata(string? masterKeyPath, bool allowEnclaveComputations, byte[]? signature)
-        {
-            throw new NotSupportedException();
-        }
-
         /// <summary>
         /// This function validates that the encryption algorithm is RSA_OAEP and if it is not,
         /// then throws an exception.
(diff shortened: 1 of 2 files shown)


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlColumnEncryptionCngProvider.Windows.cs
```diff
@@ -255,109 +120,33 @@ private void ValidateEncryptionAlgorithm(string encryptionAlgorithm, bool isSyst
...
         {
             // Get CNGProvider and the KeyID
-            string cngProviderName;
-            string keyIdentifier;
-            GetCngProviderAndKeyId(keyPath, isSystemOp, out cngProviderName, out keyIdentifier);
+            GetCngProviderAndKeyId(keyPath, isSystemOp, out string cngProviderName, out string keyIdentifier);
 
-            CngProvider cngProvider = new CngProvider(cngProviderName);
+            CngProvider cngProvider = new(cngProviderName);
             CngKey cngKey;
```
paulmedynski: Could this be simplified to:

```
try
{
    using CngKey cngKey = CngKey.Open(keyIdentifier, cngProvider);

    // RSACng() makes a copy of the key, so we can safely dispose of our copy.
    return new RSACng(cngKey);
}
catch (CryptographicException)
{
    throw SQL.InvalidCngKey(keyPath, cngProviderName, keyIdentifier, isSystemOp);
}
```

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlColumnEncryptionCngProvider.Windows.cs
```diff
@@ -255,109 +120,33 @@ private void ValidateEncryptionAlgorithm(string encryptionAlgorithm, bool isSyst
...
-        /// Gets the public Key size in bytes
-        /// </summary>
-        /// <param name="rsaCngProvider">RSA CNG Provider.</param>
-        /// <returns>Key size in bytes</returns>
-        private int GetKeySize(RSACng rsaCngProvider)
-        {
-            Debug.Assert(rsaCngProvider != null);
-
-            return rsaCngProvider.KeySize / 8; // Convert from bits to byte
+            if (string.IsNullOrWhiteSpace(masterKeyPath))
```
paulmedynski, line 68: Unfortunate that there is no `string.IsWhiteSpace()` method, since you've already checked for null so why check again?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlColumnEncryptionCngProvider.Windows.cs
```diff
@@ -31,217 +33,80 @@ public class SqlColumnEncryptionCngProvider : SqlColumnEncryptionKeyStoreProvide
...
         /// This function validates that the encryption algorithm is RSA_OAEP and if it is not,
-        /// then throws an exception
+        /// then throws an exception.
         /// </summary>
-        /// <param name="encryptionAlgorithm">Asymmetric key encryptio algorithm</param>
-        /// <param name="isSystemOp">Indicates if ADO.NET calls or the customer calls the API</param>
-        private void ValidateEncryptionAlgorithm(string encryptionAlgorithm, bool isSystemOp)
+        /// <param name="encryptionAlgorithm">Asymmetric key encryption algorithm.</param>
+        /// <param name="isSystemOp">Indicates if ADO.NET calls or the customer calls the API.</param>
+        private static void ValidateEncryptionAlgorithm([NotNull] string? encryptionAlgorithm, bool isSystemOp)
```
paulmedynski, line 42: Do we need this [NotNull] since we're explicitly checking for that condition?  What problem is this solving?  Would this suppress an IDE/compiler warning if the null check on line 109 is removed?

Here and the next function.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlColumnEncryptionCspProvider.Windows.cs
```diff
@@ -157,294 +82,134 @@ public override byte[] EncryptColumnEncryptionKey(string masterKeyPath, string e
...
         /// This function validates that the encryption algorithm is RSA_OAEP and if it is not,
-        /// then throws an exception
+        /// then throws an exception.
         /// </summary>
-        /// <param name="encryptionAlgorithm">Asymmetric key encryption algorithm</param>
-        /// <param name="isSystemOp">Indicates if ADO.NET calls or the customer calls the API</param>
-        private void ValidateEncryptionAlgorithm(string encryptionAlgorithm, bool isSystemOp)
+        /// <param name="encryptionAlgorithm">Asymmetric key encryption algorithm.</param>
+        /// <param name="isSystemOp">Indicates if ADO.NET calls or the customer calls the API.</param>
+        private static void ValidateEncryptionAlgorithm([NotNull] string? encryptionAlgorithm, bool isSystemOp)
```
paulmedynski, line 45: Same comments here as in the Cng file.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlColumnEncryptionCngProvider.Windows.cs
edwardneal: It can. I'd initially ruled it out in case the `RSACng` constructor threw, but that only throws `ArgumentException` and `ArgumentNullException`.

In fact - I've also just noticed that I can reuse the static `CngProvider` instances in some cases. I'll roll that up as well.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlColumnEncryptionCngProvider.Windows.cs
edwardneal: The key method is `ValidateNonEmptyKeyPath`. In that method, applying [NotNull] tells the compiler that the value passed as the `masterKeyPath` parameter is non-null; this means that `DecryptColumnEncryptionKey` and `EncryptColumnEncryptionKey` can use it without needing to null-check.

I've added the annotation here because it matches the contract of the method: we know that if it returns without throwing an exception, the parameter is non-null.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlColumnEncryptionCngProvider.Windows.cs
paulmedynski: We must always dispose of our CngKey, regardless of the outcome of `RSACng()`.  I think my suggestion makes it clear that will always happen no matter how the `try { }` block is exited.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlColumnEncryptionCngProvider.Windows.cs
paulmedynski: Ok thanks - I misunderstood the intention of `[NotNull]`.  It's how methods like `string.IsNullOrWhitespace()` tell the compiler that, if they return (or return a certain value) that the input argument cannot be null after that point.  Neato!

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlColumnEncryptionCngProvider.Windows.cs
edwardneal: I agree. I've pushed the change, it's much clearer now thanks

paulmedynski (dismissed):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlColumnEncryptionCngProvider.Windows.cs
benrr101: @paulmedynski it also provided annotations to fancier IDEs so they can provide warnings like "hey, you might be sending null to an argument that shouldn't be null"

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlColumnEncryptionCngProvider.Windows.cs
```diff
@@ -31,217 +33,80 @@ public class SqlColumnEncryptionCngProvider : SqlColumnEncryptionKeyStoreProvide
...
```
benrr101, line 42: My followup question is - if we don't want null, why do we have the `string?` here? Why not make it `string` and it implicitly becomes not-null. If it might be not-null somewhere else, we can use `variableName!` to assert it's not null, right? Since it's private the scope of it should be pretty small

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlColumnEncryptionCngProvider.Windows.cs
```diff
@@ -255,150 +120,93 @@ private void ValidateEncryptionAlgorithm(string encryptionAlgorithm, bool isSyst
         /// <summary>
         /// Checks if the CNG key path is Empty or Null (and raises exception if they are).
         /// </summary>
-        /// <param name="masterKeyPath">keypath containing the CNG provider name and key name</param>
-        /// <param name="isSystemOp">Indicates if ADO.NET calls or the customer calls the API</param>
-        private void ValidateNonEmptyKeyPath(string masterKeyPath, bool isSystemOp)
+        /// <param name="masterKeyPath">Key path containing the CNG provider name and key name.</param>
+        /// <param name="isSystemOp">Indicates if ADO.NET calls or the customer calls the API.</param>
+        private static void ValidateNonEmptyKeyPath([NotNull] string? masterKeyPath, bool isSystemOp)
```
benrr101, line 61: _technically_ static methods should go be fore instance methods

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlColumnEncryptionCngProvider.Windows.cs
```diff
@@ -255,150 +120,93 @@ private void ValidateEncryptionAlgorithm(string encryptionAlgorithm, bool isSyst
...
             }
+
+            ReadOnlySpan<char> cngProviderName = keyPathSpan.Slice(0, indexOfSlash);
+
+            // If the provider is one of the well-known providers, use the static instance instead of allocating a new string and CngProvider.
+            if (cngProviderName.Equals(CngProvider.MicrosoftSoftwareKeyStorageProvider.Provider.AsSpan(), StringComparison.OrdinalIgnoreCase))
+            {
+                cngProvider = CngProvider.MicrosoftSoftwareKeyStorageProvider;
+            }
+            else if (cngProviderName.Equals(CngProvider.MicrosoftSmartCardKeyStorageProvider.Provider.AsSpan(), StringComparison.OrdinalIgnoreCase))
```
benrr101, line 130: It seems a bit redundant to repeatedly convert the string to span when there's an overload that directly takes the string. It's just checking for equality so I don't think it will be doing any allocations based on the string.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlColumnEncryptionCngProvider.Windows.cs
edwardneal: We could, but `variableName!` is an escape hatch. The point of `ValidateEncryptionAlgorithm` is that it receives a string which might be null, then asserts that by the time control flow leaves the method successfully, the string it receives is guaranteed to be non-null. The NRT annotations are set up as they are to reflect that, and the analysis tooling then feeds that awareness about the parameter passed to `ValidateEncryptionAlgorithm` into the calling method - so there are no warnings in the subsequent usage of that parameter.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlColumnEncryptionCngProvider.Windows.cs
edwardneal: That's correct for the RHS, but `cngProviderName` is a ROS<char>. If I compare two strings, I've got to allocate a string and materialise `cngProviderName` into it.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlColumnEncryptionCngProvider.Windows.cs
edwardneal: Yes - done!

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlColumnEncryptionCngProvider.Windows.cs
benrr101: ah, ok, that's fair

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlColumnEncryptionCngProvider.Windows.cs
benrr101: Ah, I think I misunderstood this as well. ReSharper provides their own flavor of annotations, including a `[NotNull]` that has different implications than CodeAnalysis package's `[NotNull]` (https://www.jetbrains.com/help/resharper/Reference__Code_Annotation_Attributes.html#ui8qor_5)
I see how the intention is to signal that this method validates the encryptionAlgorithm argument is not null so that the code after calling ValidateEncryptionAlgorithm can use encryptionAlgorithm without checking for null again.

ReSharper's `[NotNull]` signals that null should not be passed into the method (similar to CodeAnalysis's `[DisallowNull]`). So hopefully you can see how I was confused - why explicitly allow null when we don't want null to be passed to it?! It would be like calling `string.IsNull(null)` and then yelling at compile time that you passed null to it! But that's not what we're doing here, so I get it now.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlColumnEncryptionCngProvider.Windows.cs
benrr101: @paulmedynski forget what I said - that behavior applies to ReSharper's code annotations, not CodeAnalysis's. https://www.jetbrains.com/help/resharper/Reference__Code_Annotation_Attributes.html#ItemNotNullAttribute

benrr101 (approved):
I get it now - and like Paul said, neato!

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlColumnEncryptionCngProvider.Windows.cs
edwardneal: I see what you mean! I'm working from the compiler documentation [here](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/attributes/nullable-analysis), but agree - the ReSharper documentation has a very different description. It looks like ReSharper will respect the results of the compiler's nullability analysis though, so we shouldn't find any conflicts between VS' and ReSharper's results.

PR #3617, opened 2025-09-13
[5.1] Use UTF8 instance that doesn't emit BOM

apoorvdeshmukh:
## Description

Ports #3399 to release/5.1

## Issues

#3397

## Testing

Port contains the testcase to validate the fix


benrr101:
Afaik, this will need to have the branch merged into it to pick up CI changes once they go through.

Issue #3459, opened 2025-06-30
[PLACEHOLDER] Improve Async Pathways in Connectivity APIs


mdaigle:
### Is your feature request related to a problem? Please describe.
This is a placeholder issue, more details will be added over time.

SqlConnection.OpenAsync() violates async best practices in numerous locations. This issue will track those locations and how they should be fixed.

- connection resiliency reconnect uses Task.Run. Introduce a sync api.
internal Task ValidateAndReconnect(Action beforeDisconnect, int timeout)
runningReconnect = Task.Run(() => ReconnectAsync(timeout));
- SqlInternalConnectionTds.cs - fedauth token acquisition
    - _fedAuthToken = new(Task.Run(async () => await authProvider!.AcquireTokenAsync(authParamsBuilder)).GetAwaiter().GetResult());

### Additional context
Add any other context about the feature request here.

PR #3619, opened 2025-09-17
Merge | SqlCommand Scalar Methods

benrr101:
## Description
Welcome to the next installment of SqlCommand merge PRs. This PR addresses merging of ExecuteScalar and related methods. This begins another organizational step of creating separate partial classes for different large blocks of code. This introduces the .Scalar partial that contains methods that only pertain to executing scalars. Code that is shared across types of execution will remain in the main partial, as will all member variables and properties. Without some division like this, the SqlCommand file will be over 10k lines long and become incredibly difficult to navigate. Feel free to propose alternatives.

This PR merges:
* ExecuteScalar
* ExecuteScalarAsync
* ExecuteScalarBatchAsync
* ExecuteScalarUntilEndAsync
* InternalExecuteScalarAsync (renamed to ExecuteScalarAsyncInternal)
* CompleteExecuteScalar

## Issues
Continuing work on #1261 

## Testing
Projects continue to build, and SqlCommandTests pass locally. CI will complete the validation.


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Scalar.cs
```diff
@@ -106,6 +105,137 @@ private static object CompleteExecuteScalar(SqlDataReader reader, bool returnLas
...
+                "SqlCommand.ExecuteScalarAsyncInternal | API " +
+                $"Object Id {ObjectID}, " +
+                $"Client Connection Id {_activeConnection?.ClientConnectionId}, " +
+                $"Command Text '{CommandText}'");
+            SqlClientEventSource.Log.TryCorrelationTraceEvent(
+                "SqlCommand.InternalExecuteScalarAsync | API | Correlation | " +
+                $"Object Id {ObjectID}, " +
+                $"Activity Id {ActivityCorrelator.Current}, " +
+                $"Client Connection Id {_activeConnection?.ClientConnectionId}, " +
+                $"Command Text '{CommandText}'");
```
edwardneal: These have switched order when compared to netcore.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Scalar.cs
```diff
@@ -106,6 +105,137 @@ private static object CompleteExecuteScalar(SqlDataReader reader, bool returnLas
...
+
+            #if NET
+            Guid operationId = s_diagnosticListener.WriteCommandBefore(this, _transaction);
+            _parentOperationStarted = true;
+            #endif
+
+            // @TODO: Use continue with state? This would be a good candidate for rewriting async/await
```
edwardneal, line 241: Completely agreed. SqlBulkCopy is in a similar position, it'd be good to start working towards async/await in places like this.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Scalar.cs
```diff
@@ -0,0 +1,348 @@
...
+                                finally
+                                {
+                                    reader.Dispose();
+                                }
+
+                                if (exception is not null)
+                                {
+                                    s_diagnosticListener.WriteCommandError(operationId, this, _transaction, exception);
+                                    source.SetException(exception);
+                                }
```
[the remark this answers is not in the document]
edwardneal: Agreed - this looks like a bug in the netcore version. I'll deal with this when we port the diagnostics tracing to SqlCommand.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Scalar.cs
```diff
@@ -193,6 +193,25 @@ private static object CompleteExecuteScalar(SqlDataReader reader, bool returnLas
...
+            SqlDataReader reader,
+            CancellationToken cancellationToken)
+        {
+            // @TODO: This is doing almost the same thing as CompleteExecuteScalar when returnLastResult is true
+            object result = null;
+            do
+            {
+                if (await reader.ReadAsync(cancellationToken).ConfigureAwait(false) && reader.FieldCount > 0)
+                {
+                    result = reader.GetValue(0);
```
edwardneal, line 213: We've lost the comment here. The comment seems to be talking about a real problem - this could technically be an async-over-sync operation (for a sufficiently large value.)

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Scalar.cs
```diff
@@ -0,0 +1,348 @@
...
+                statistics = SqlStatistics.StartTimer(Statistics);
+                WriteBeginExecuteEvent();
+
+                // @TODO: Rename ds to reader
+                SqlDataReader ds = IsProviderRetriable
+                    ? RunExecuteReaderWithRetry(CommandBehavior.Default, RunBehavior.ReturnImmediately, returnStream: true)
+                    : RunExecuteReader(CommandBehavior.Default, RunBehavior.ReturnImmediately, returnStream: true);
+                success = true;
+
+                return CompleteExecuteScalar(ds, _batchRPCMode);
```
edwardneal: This has ported a bug from netcore: it sets `success` before it's run `CompleteExecuteScalar` (which can throw an exception.) The netfx behaviour was correct, we should only set `success` after the call to `CompleteExecuteScalar` has run without throwing an exception.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Scalar.cs
```diff
@@ -0,0 +1,348 @@
...
+        }
+        
+        #if NET
+        internal Task<object> ExecuteScalarBatchAsync(CancellationToken cancellationToken)
+        {
+            Guid operationId = s_diagnosticListener.WriteCommandBefore(this, _transaction);
+            _parentOperationStarted = true;
+            
+            // @TODO: This code is almost identical to ExecuteScalarAsyncInternal - we can definitely refactor it!
+            return ExecuteReaderAsync(cancellationToken).ContinueWith(executeTask =>
```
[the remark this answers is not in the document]
benrr101: THAT'S EXACTLY WHAT I SAID, YOU CAN'T JUST COPY MY TODO's AND PRETEND THEY'RE YOURS!

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Scalar.cs
benrr101: Yeah, these TODOs are to capture things that I hope to change in later PRs.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Scalar.cs
benrr101: Ah yeah, I tried to find those and follow the seemingly correct behavior. I must've missed this one.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Scalar.cs
benrr101: I guess that's true ... Is it expected that correlation happens before trace events? I don't really know the tracing expectations aside from them being kinda inconsistent. (Like personally, I'd like it to be the first that happens in each public method)

I'll make the change, but if you have any insights, let me know.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Scalar.cs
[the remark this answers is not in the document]
benrr101: Hmm, yeah even I've got the "expression is always false" warning from my IDE

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Scalar.cs
edwardneal: I've not got any insights unfortunately - I was simply flagging the change in case it was done deliberately because there'd been a change in the way the team wants to use tracing. If it's just a coincidence then no action's needed.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Scalar.cs
```diff
@@ -0,0 +1,352 @@
...
+
+using System;
+using System.Data;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Data.Common;
+
+namespace Microsoft.Data.SqlClient
+{
+    public partial class SqlCommand
```
paulmedynski, line 16: With these partials, I would like to see a class comment for each one that describes what part of the declaration and definitions it contains, and rationale for its existence.

benrr101: I *think* I can do that - I think using the <content> tag will work. The only thing I'm uncertain about is how it plays with the public API docs.

paulmedynski: Normal `//` comments would be fine by me.  If you can use `///` that's cool, but not required IMO.

paulmedynski (approved):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Scalar.cs
mdaigle: Please feel free to add/comment any async issues you find to #3459. I'm looking to use that as the tracking issue for all of these. Or we can split non-connectivity off into a separate issue.

mdaigle (approved):

PR #3620, opened 2025-09-18
Fix SetProvider to return immediately if user-defined provider found

### LINKED ISSUE #3618: Custom SqlClientAuthenticationProvider being overwritten by default one
### Describe the bug

When a custom provider is registered in MDS 5.2+ it will be overwritten by the default provider. This is from the verbose logs for our app : 

```
2025-09-13T16:33:21.7520936Z [78]: <sc|SqlAuthenticationProviderManager|SetProvider|Error>Failed to add provider Microsoft.Data.SqlClient.ActiveDirectoryAuthenticationProvider because a user-defined provider with type Microsoft.SqlServer.Management.UI.AadInteractiveAuthProvider.AadAuthenticationProvider already existed for authentication ActiveDirectoryInteractive.
2025-09-13T16:33:21.7531099Z [78]: <sc|ActiveDirectoryAuthenticationProvider|BeforeLoad|Info>being loaded into SqlAuthProviders for ActiveDirectoryInteractive.
2025-09-13T16:33:21.7531099Z [78]: <sc|SqlAuthenticationProviderManager|SetProvider|Info>Added auth provider Microsoft.Data.SqlClient.ActiveDirectoryAuthenticationProvider, overriding existed provider Microsoft.SqlServer.Management.UI.AadInteractiveAuthProvider.AadAuthenticationProvider for authentication ActiveDirectoryInteractive.
```

### To reproduce

1. Create and register a custom auth provider as documented here https://learn.microsoft.com/en-us/sql/connect/ado-net/sql/azure-active-directory-authentication?view=sql-server-ver17#support-for-a-custom-sql-authentication-provider
2. Start up the application

### Expected behavior

The custom auth provider is registered and called when Entra ID tokens are needed for authentication.

### Further technical details

N/A

**Additional context**

This appears to have been introduced in https://github.com/dotnet/SqlClient/pull/1949/files#diff-d95a8fbf13f839abdfbc99fa02b0b3c2170eb8ddcb8747c729842f52854e56baL172

FYI @Wraith2 

Previously, the method would return if a matching provider was already registered. But after the change the break just causes it to exit the loop, but then continue on to the following code which registers the default provider.

This was first introduced in MDS 5.2, but still seems to be an issue in 6.0 (I haven't verified this yet though except by looking at the code)

https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlAuthenticationProviderManager.cs#L189

cheenamalhotra:
Fixes #3618 

The issue is reproducible when config file is configured with a custom authentication provider.
Using SetProvider (as in tests) to override default providers does not reproduce issue, hence missed.

The fix will be back-ported to other affected branches (release/6.0 and 6.1).


paulmedynski:
Should we add a test that confirms the broken behaviour when using the config-file based auth provider registration?  And then see it pass with your fix.

ErikEJ:
Wonder if it would be possible to add a test to prevent regression?

src/Microsoft.Data.SqlClient/tests/FunctionalTests/app.config
paulmedynski, line 1: How does this file end up influencing the runtime?  Is it published somewhere the tests know to read it?  Some comments in the XML to explain how it all works together would be helpful.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/AADAuthenticationTests.cs
```diff
@@ -50,6 +50,18 @@ private void InvalidCombinationCheck(SqlCredential credential)
             }
         }
 
+        [Fact]
```
paulmedynski, line 61: A comment here describing the interaction between MDS and the App.config would be helpful, and how/why we expect the dummy provider to be registered for ActiveDirectoryInteractive here.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/app.config
cheenamalhotra: App.config file is loaded by .NET framework runtime before loading managed types. Added note in file.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/AADAuthenticationTests.cs
```diff
@@ -50,6 +50,24 @@ private void InvalidCombinationCheck(SqlCredential credential)
...
         }
 
+        /// <summary>
+        /// Tests whether SQL Auth provider is overridden using app.config file.
+        /// This use case is only supported for .NET Framework applications, as driver doesn't support reading configuration from appsettings.json file.
+        /// In future if need be, appsettings.json support can be added.
+        /// </summary>
+        /// <returns></returns>
+        [ConditionalFact(typeof(TestUtility), nameof(TestUtility.IsFullFramework))]
+        public async Task IsDummySqlAuthenticationProviderSetByDefault()
```
cheenamalhotra, line 62: This test seems to run fine locally but I haven't been able to run it in pipelines for some reason. It seems the app.config is not being considered by dotnet test even though we're not rebuilding it. 

I will disable it in pipelines for now, with a comment to address it in sometime future to unblock releasing the fix.

cheenamalhotra: No-op. Resolved.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlAuthenticationProviderTest.cs
```diff
@@ -22,5 +22,18 @@ public void DefaultAuthenticationProviders(SqlAuthenticationMethod method)
...
+
+        #if NETFRAMEWORK
+        // This test is only valid for .NET Framework
+
+        // Overridden by app.config in this project
+        [Theory]
+        [InlineData(SqlAuthenticationMethod.ActiveDirectoryInteractive)]
+        public void DefaultAuthenticationProviders_Interactive(SqlAuthenticationMethod method)
+        {
+            Assert.IsType<DummySqlAuthenticationProvider>(SqlAuthenticationProvider.GetProvider(method));
```
mdaigle, line 34: Does this test add anything over the other one? It seems we could remove this.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/AADAuthenticationTests.cs
```diff
@@ -49,6 +49,27 @@ private void InvalidCombinationCheck(SqlCredential credential)
...
+
+        /// <summary>
+        /// Tests whether SQL Auth provider is overridden using app.config file.
+        /// This use case is only supported for .NET Framework applications, as driver doesn't support reading configuration from appsettings.json file.
+        /// In future if need be, appsettings.json support can be added.
+        /// </summary>
+        [Fact]
+        public async Task IsDummySqlAuthenticationProviderSetByDefault()
+        {
+            var provider = SqlAuthenticationProvider.GetProvider(SqlAuthenticationMethod.ActiveDirectoryInteractive);
```
mdaigle, line 64: It would be nice to get the config injected into the SqlAuthenticationProviderManager (maybe via an internal method). I see people doing facade patterns over ConfigurationManager which lets you set config without having to modify a global app.config that is visible to all tests.

We have the same basic issue with AppContext switches. We need to move toward more dependency injection so that we can test our flags/config/switches without impacting test parallelization.

mdaigle (approved):

src/Microsoft.Data.SqlClient/tests/FunctionalTests/SqlAuthenticationProviderTest.cs
cheenamalhotra: I added this for future reference, to capture all the expected providers wherever default providers are tested. Doesn't hurt to test more than once.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/AADAuthenticationTests.cs
cheenamalhotra: That is possible, but that scenario doesn't help us with reproducing the issue. Since we are setting the default providers in constructors (which is what we want to test here), the issue is reproduced by loading custom providers via external config file.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/AADAuthenticationTests.cs
```diff
@@ -49,24 +49,27 @@ private void InvalidCombinationCheck(SqlCredential credential)
                 Assert.Throws<InvalidOperationException>(() => connection.AccessToken = "SampleAccessToken");
             }
         }
+        
+        #if NETFRAMEWORK
```
paulmedynski, line 53: Does this test compile for .NET?  If so I would prefer to keep it compiling there, and then have the test conditionals skip it.

The less code we have that only compiles conditionally, the better.

Won't hold up this PR for it, but something to keep in mind after the code merge and we start cleaning up how the projects build/run.

paulmedynski (approved):

src/Microsoft.Data.SqlClient/tests/FunctionalTests/AADAuthenticationTests.cs
cheenamalhotra: I'm not sure if we have functional attributes for conditional skipping of tests on targeted framework. That itself needs to be investigated.

Alternatively, using if-defs gives us 2 benefits:
1. Prevents compilation on other frameworks where sometimes APIs could be missing.
2. Helps with testing in Visual Studio, as they don't compile in the list of tests against frameworks, whereas conditional skipping still runs the tests (even when they are meant to skip for the condition) (e.g. we have seen that for OS conditions).

PR #3304, opened 2025-04-24
Tests | Activate "ActiveIssue" tests

benrr101:
_This is a recreation of #3012 to resolve merges, rebase it off of main, and enable full CI builds. Original description is below_

This PR activates a couple of tests that were marked with "ActiveIssue"
Upon trying them out, they seem to run fine.

Also does some minor improvements on a few tests or disable some tests on specific platform

Fixes #3035 

This also fixes and allows to close the following ActiveIssue's in AzDO:

- 5531
- 5533
- 5535
- 5536
- 5541
- 5540
- 6643
- 9196
- 12161
- 12161
- 14588
- 14325
- 14590
- 20245
- 21707
- 25147
- 25421
- 31754 (Partly - 2 tests activated, at least one remains)

PR #3624, opened 2025-09-22
Tests | Active Issues (Part 2)

benrr101:
## Description
This is a continuation of submitting the individual active issue re-enables from #3304. This submission re-enables 2 tests and changes how they're included slightly:

Tests Enabled:
* ManualTests - CertificateTest.OpeningConnectionWithGoodCertificateTest
* ManualTests - CertificateTest.OpeningConnectionWithNicTest

Other Changes:
* Rather than Assert.True the tests are ran as Admin, the tests are now skipped if not executed in Admin mode
* Making the entire class platform-specific to Windows rather than individual tests

## Issues
31754, which belongs to some ticketing system forgotten to time

## Testing
Tests build and are successfully skipped if not ran in Admin mode. CI should validate it further.


mdaigle (approved):
Do we run any stages as admin? Or will these still be skipped in the pipeline?

Issue #2610, opened 2024-06-26
Fail to connect after database failover


brunoblank:
### Describe the bug
We have a database setup using mirroring. 

Connection-String: `server=tcp:172.21.6.8;user id=<user>;password=<password>;database=<database>;Failover Partner=tcp:172.21.6.7;TrustServerCertificate=true`

When the database fails over from the "secondary" then everything works instantly.
When the database fails over from the "primary" (the one that _server_ points to) then we see the exception below. The connection works if the application is restarted or if there has been no requests for a couple of minutes.

The exception indicates that `Named pipes` are being used even though we are specifying `tcp:`

```
A network-related or instance-specific error occurred while establishing a connection to SQL Server. The server was not found or was not accessible. Verify that the instance name is correct and that SQL Server is configured to allow remote connections. (provider: Named Pipes Provider, error: 40 - Could not open a connection to SQL Server)
   at Microsoft.Data.SqlClient.SqlInternalConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   at Microsoft.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObject stateObj, SqlCommand command, Boolean callerHasConnectionLock, Boolean asyncClose)
   at Microsoft.Data.SqlClient.TdsParser.Connect(ServerInfo serverInfo, SqlInternalConnectionTds connHandler, TimeoutTimer timeout, SqlConnectionString connectionOptions, Boolean withFailover)
   at Microsoft.Data.SqlClient.SqlInternalConnectionTds.AttemptOneLogin(ServerInfo serverInfo, String newPassword, SecureString newSecurePassword, TimeoutTimer timeout, Boolean withFailover)
   at Microsoft.Data.SqlClient.SqlInternalConnectionTds.LoginWithFailover(Boolean useFailoverHost, ServerInfo primaryServerInfo, String failoverHost, String newPassword, SecureString newSecurePassword, Boolean redirectedUserInstance, SqlConnectionString connectionOptions, SqlCredential credential, TimeoutTimer timeout)
   at Microsoft.Data.SqlClient.SqlInternalConnectionTds.OpenLoginEnlist(TimeoutTimer timeout, SqlConnectionString connectionOptions, SqlCredential credential, String newPassword, SecureString newSecurePassword, Boolean redirectedUserInstance)
[...]

Issue #2365, opened 2024-02-21
Add option to ignore server provided failover partner


David-Engel:
### Is your feature request related to a problem? Please describe.
Due to unique networking configurations for some clients of an application, the failover partner information defined on the server can't be reached over the network without using a custom server name for the failover partner for those clients.

### Describe the solution you'd like
Customers would like to be able to specify a failover partner on the client that will not get overridden by the failover partner information provided by the server. Ideally, this would be a connection string option like `ignoreServerFailoverPartner=true`.

### Describe alternatives you've considered
Disabling the failover partner information sent by the server via a server-side trace flag. However, this is not an ideal solution.

edwardneal:
I've had this problem before, and something to work around it would be useful. The situation I encountered it in related primarily to multiple domains. The SQL Server instances would be in the `resources.contoso.com` domain, and mirroring was configured to use single-label names. This caused a problem when the clients are in the `ad.contoso.com` domain - the connection string's server name and failover partner would use FQDNs, but the failover partner was overwritten with a single-label name (which would never resolve.)

PR #3625, opened 2025-09-22
Fix #3400, ignore server-provided failover partner

### LINKED ISSUE #3400: Connection fails and ignores FailoverPartner port after Linux SQL Server BAG failover, defaults to port 1433
## Description

When using `Microsoft.Data.SqlClient` with **Linux-based Basic Availability Groups** (2 synchronous replicas) and specifying a **custom port** (e.g., `1066`) in the connection string, **automatic failover** to the secondary server **fails**.

After a failover, the client **attempts to connect to port 1433**, **ignoring the explicitly configured `FailoverPartner` port and actual server port**.

✅ This issue **does not occur** with **Windows-based SQL Server** using named instances, even with custom ports, since the client uses the named instance to connect ot the failover (`server\instance` like `my-failover-partner\sql64std`).

### Behavior After Failover for this connection string
```
DataSource = "tcp:my-master-server.mycompany.com,1066",
FailoverPartner = "tcp:my-failover-partner.mycompany.com,1066",
```

- ❌ If a failover is performed after successfully connecting to the primary instance, the client attempts to connect to `my-failover-partner` **without FQDN or port**, defaults to port **1433**, and fails.
- ✅ If the **application is restarted**, it correctly connects to the failover partner using the specified port.
- ✅ If failback occurs, connections resume correctly.

### Potential Cause

From source inspection, the `FailoverPartner` may be overridden by the environment change `ENV_LOGSHIPNODE` without retrieving the port:  
🔗 https://github.com/search?q=repo%3Adotnet%2FSqlClient+ENV_LOGSHIPNODE&type=code

### Exception Message
```
A network-related or instance-specific error occurred while establishing a connection to SQL Server.
The server was not found or was not accessible.
(provider: TCP Provider, error: 40 - Could not open a connection to SQL Server)
```

## To Reproduce

1. Deploy a **Basic Availability Group** on SQL Server **on Linux** (e.g., Ubuntu 20.04) with a **custom port** (e.g., 1066).
2. Use this client code:
```
private static async Task SimpleFailoverTest()
{
    string connectionString = new SqlConnectionStringBuilder()
    {
        InitialCatalog = "DowntimeTestDB",
        DataSource = "tcp:my-master-server.mycompany.com,1066",
        FailoverPartner = "tcp:my-failover-partner.mycompany.com,1066",
        IntegratedSecurity = true,
        TrustServerCertificate = true,
    }.ConnectionString;

    while (true)
    {
        try
        {
            using var conn = new SqlConnection(connectionString);
            await conn.OpenAsync();
            var cmd = new SqlCommand("SELECT * from [DowntimeTestDB].[dbo].[OperationsLog]", conn);
            await cmd.ExecuteNonQueryAsync();
            Console.WriteLine($"Success");
        }
        catch (Exception ex)
        {
            Console.WriteLine($"Error: {ex.Message}");
        }
        await Task.Delay(1000);
    }
}
```
3. Trigger a **manual failover** to the secondary server.
4. Observe client fails to connect post-failover.

## Expected Behavior

Client should:

- Dynamically retrieve the actual custom port from the sql server (e.g., port 1066)
- **Or**, provide a **configurable option** to **prevent overriding** the `FailoverPartner` in the connection string.

## Additional Context

- ✅ **Wireshark confirmed:** Client attempts to connect to port `1433` post-failover, ignoring custom port `1066`.
- ⚠️ Failover **only works when using port 1433**, which is **not feasible** for us.
- I used the following snippet to setup the availability groups, I think that the instance names in this query are used by the client which is why the connection fails (`my-failover-partner` was used for the connection) and the availability group fails to setup with including the custom port like `my-failover-partner,1066` and `my-master-server,1066`)
```
CREATE AVAILABILITY GROUP [ag_DowntimeTestDB]
WITH (
    BASIC,
    CLUSTER_TYPE = NONE,
    REQUIRED_SYNCHRONIZED_SECONDARIES_TO_COMMIT = 0
)
FOR DATABASE [DowntimeTestDB]
REPLICA ON 
    N'my-master-server' WITH (
        ENDPOINT_URL = N'tcp://my-master-server.mycompany.com:5022',
        FAILOVER_MODE = MANUAL,
        AVAILABILITY_MODE = SYNCHRONOUS_COMMIT,
        SEEDING_MODE = AUTOMATIC,
        SECONDARY_ROLE(ALLOW_CONNECTIONS = NO)
    ),
    N'my-failover-partner' WITH (
        ENDPOINT_URL = N'tcp://my-failover-partner.mycompany.com:5022',
        FAILOVER_MODE = MANUAL,
        AVAILABILITY_MODE = SYNCHRONOUS_COMMIT,
        SEEDING_MODE = AUTOMATIC,
        SECONDARY_ROLE(ALLOW_CONNECTIONS = NO)
    )
GO
```

## Logs / Traces from SqlClient

- ✅ **Working Failover Trace:** Windows SQL Server with named instance failover succeeded on non-default port `1066` (it can be seen that the named instance was used for the connection)
[trace_wsl_targets_windows_failover_success.txt](https://github.com/user-attachments/files/20634008/trace_wsl_targets_windows_failover_success.txt)
- ❌ **Failing Failover Trace:** Linux SQL Server failover, client doesn’t connect to custom port `1066` (we can see that the instance name with no port was used)
[trace_wsl_targets_linux_failover_failure.txt](https://github.com/user-attachments/files/20634009/trace_wsl_targets_linux_failover_failure.txt)

## Current Workarounds (Undesirable)

1. External port redirection from `1433 → 1066`  
   ⛔ Not sure how feasable and complex it is

2. Standardize to default port `1433` across all clients and servers  
   ⛔ Not feasible due to massive deployment and client base

3. Fork and patch `SqlClient` to use `1066` as the default  
   ⛔ Maintenance overhead and unclear reliability

## Technical Details

- **Microsoft.Data.SqlClient version:** `5.2.2`, `6.0.2`
- **.NET Target Framework:** `.NET 8.0`
- **SQL Server Version:**
  - Linux: SQL Server 2022 CU19 with non-default port
  - Windows: SQL Server 2022 CU11 with non-default port
- **OS Configuration:**
  - ❌ Failing: WSL Linux client / Windows client → Linux SQL Server
  - ✅ Working: WSL Linux client / Windows client → Windows SQL Server (named instances)

---

Please advise on a potential fix or workaround that would allow failover to connect to the custom SQL Server port or provide a way to prevent overriding FailoverPartner provided in the connection string.

mdaigle:
## Description

A Basic Availability Group (BAG), is a basic high availability configuration for standard edition, on-prem servers (not applicable to azure). It involves two servers, one a primary replica and the other a secondary replica (a.k.a. failover partner / mirroring partner). When the primary is unavailable, clients will connect to the secondary. When connecting to the primary in a BAG, the server specifies the domain name of the secondary to the client via a Real Time Log Shipping ENVCHANGE token. This allows the BAG's secondary configuration to be updated dynamically without a code change. We call this the server-provided failover partner. Clients can also specify a failover partner directly in the connection string. The client-provided failover partner is used only if the initial connection to the primary replica fails (no server-side failover partner available). Otherwise, the server-provided failover partner supersedes the client-provided failover partner.

When configuring a BAG, replicas are added to the group by specifying their NetBIOS or network name ([CREATE AVAILABILITY GROUP (Transact-SQL) - SQL Server | Microsoft Learn](https://learn.microsoft.com/en-us/sql/t-sql/statements/create-availability-group-transact-sql?view=sql-server-ver17#server_instance)), NOT their protocol, fully qualified domain name, and port. This network name is the value returned to the client as the server-provided failover partner. This causes an issue when using TCP for either replica because the server-provided failover partner will not contain the protocol and FQDN.

The goal of this PR is to provide support for TCP in BAGs while respecting the following constraints:
- We must continue to prefer server-provided failover partners by default. This behavior is useful and changing it would not be backwards compatible.
- Low cost. Avoid server-side or TSQL changes.

We introduce a client-side app context switch that, when enabled, ignores the server-provided failover partner and only uses the failover partner provided in the connection string.

### Example usage
Connecting to a BAG via TCP:
```c#
AppContext.SetSwitch("Switch.Microsoft.Data.SqlClient.IgnoreServerProvidedFailoverPartner", true);

SqlConnection connection = new SqlConnection("Data Source=tcp:localhost,1234;Failover Partner=tcp:localhost,5678;...");
connection.Open();
```

## Issues

#3400

## Testing

Testing capabilities are currently limited in this area. 
#3488 will add basic integration level test coverage for this area.

### DIFF
--- a/src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionFailoverTests.cs
+++ b/src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionFailoverTests.cs
@@ -4,6 +4,8 @@
 
 using System;
 using System.Data;
+using Microsoft.Data.SqlClient;
+using Microsoft.Data.SqlClient.Tests.Common;
 using Microsoft.SqlServer.TDS.Servers;
 using Xunit;
 
@@ -345,7 +347,7 @@ namespace Microsoft.Data.SqlClient.UnitTests.SimulatedServerTests
                 new TdsServerArguments
                 {
                     // Doesn't need to point to a real endpoint, just needs a value specified
-                    FailoverPartner = "localhost:1234",
+                    FailoverPartner = "localhost,1234",
                 });
             failoverServer.Start();
 
@@ -354,7 +356,7 @@ namespace Microsoft.Data.SqlClient.UnitTests.SimulatedServerTests
                 {
                     IsEnabledTransientError = true,
                     Number = errorCode,
-                    FailoverPartner = $"localhost:{failoverServer.EndPoint.Port}",
+                    FailoverPartner = $"localhost,{failoverServer.EndPoint.Port}",
                 });
             server.Start();
 
@@ -520,6 +522,69 @@ namespace Microsoft.Data.SqlClient.UnitTests.SimulatedServerTests
             }
 
             Assert.Fail();
+        }
+
+        [Fact]
+        public void TransientFault_IgnoreServerProvidedFailoverPartner_ShouldConnectToUserProvidedPartner()
+        {
+            // Arrange
+            using LocalAppContextSwitchesHelper switchesHelper = new();
+            switchesHelper.IgnoreServerProvidedFailoverPartnerField = LocalAppContextSwitchesHelper.Tristate.True;
+
+            using TdsServer failoverServer = new(
+                new TdsServerArguments
+                {
+                    // Doesn't need to point to a real endpoint, just needs a value specified
+                    FailoverPartner = "localhost,1234",
+                });
+            failoverServer.Start();
+
+            using TdsServer server = new(
+                new TdsServerArguments()
+                {
+                    // Set an invalid failover partner to ensure that the connection fails if the
+                    // server provided failover partner is used.
+                    FailoverPartner = $"invalidhost",
+                });
+            server.Start();
+
+            SqlConnectionStringBuilder builder = new()
+            {
+                DataSource = $"localhost,{server.EndPoint.Port}",
+                InitialCatalog = "master",
+                Encrypt = false,
+                FailoverPartner = $"localhost,{failoverServer.EndPoint.Port}",
+                // Ensure pooling is enabled so that the failover partner information
+                // is persisted in the pool group. If pooling is disabled, the server
+                // provided failover partner will never be used.
+                Pooling = true
+            };
+            SqlConnection connection = new(builder.ConnectionString);
+
+            // Connect once to the primary to trigger it to send the failover partner
+            connection.Open();
+            Assert.Equal("invalidhost", (connection.InnerConnection as SqlInternalConnectionTds)!.ServerProvidedFailoverPartner);
+
+            // Close the connection to return it to the pool
+            connection.Close();
+
+
+            // Act
+            // Dispose of the server to trigger a failover
+            server.Dispose();
+
+            // Opening a new connection will use the failover partner stored in the pool group.
+            // This will fail if the server provided failover partner was stored to the pool group.
+            using SqlConnection failoverConnection = new(builder.ConnectionString);
+            failoverConnection.Open();
+
+            // Assert
+            Assert.Equal(ConnectionState.Open, failoverConnection.State);
+            Assert.Equal($"localhost,{failoverServer.EndPoint.Port}", failoverConnection.DataSource);
+            // 1 for the initial connection
+            Assert.Equal(1, server.PreLoginCount);
+            // 1 for the failover connection
+            Assert.Equal(1, failoverServer.PreLoginCount);
         }
     }
 }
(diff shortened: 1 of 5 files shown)


cheenamalhotra:
Can you add an example here on how should a user specify secondary failover replica info with this change and enable app context switch alongside?

edwardneal:
This might also be the root cause of #2610.

Besides this, another issue (#2365) has suggested a similar idea, but controlled using a connection string option.

mdaigle:
> This might also be the root cause of #2610.
> 
> Besides this, another issue (#2365) has suggested a similar idea, but controlled using a connection string option.

Yes, looks like this is the same issue. The network name used to configure a BAG doesn't square with most people's setups these days.

cheenamalhotra (dismissed):

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionFailoverTests.cs
```diff
@@ -521,5 +523,68 @@ public void TransientFault_WithUserProvidedPartner_RetryDisabled_ShouldFail(uint
 
             Assert.Fail();
         }
+
+        [Fact]
+        public void TransientFault_IgnoreServerProvidedFailoverPartner_ShouldConnectToUserProvidedPartner()
```
samsharma2700, line 528: Can we do some manual test with Linux and Windows SQL server AG setup since that was the bifurcation point of original issue?

samsharma2700 (approved):

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionFailoverTests.cs
mdaigle: If you take a look at #2610, you'll see that the issue is present for windows as well. The logic that was changed is agnostic of platform, so this test should show that the behavior has been corrected for all platforms.

PR #3603, opened 2025-09-05
Merge | TdsParserStateObject packet handling

edwardneal:
## Description

This follows up #3394, merging a number of methods relating to packet handling and some final errata. This is split into multiple commits for ease of review.

I'm moving a few members from TdsParserStateObject to TdsParserStateObjectNative:
* SetPacketData
* CreateAndSetAttentionPacket
* CheckConnection
* EnableSSL
* Handle

SetPacketData is the only truly standout method here. The netfx implementation was structured slightly differently to the netcore implementation, and it also cleared _outBuff after writing a SecureString to the packet. I've added this feature in dcd231eb41573c12640b5d7f72c8c85abc9034d9.

This is the final PR prior to the merge of TdsParserStateObject and TdsParserStateObjectNative.

## Issues

Relates to #1261.

## Testing

Automated tests pass. Can someone run CI please?

PR #591, opened 2020-06-05
Raise warning message for insecure protocols

DavoudEshtehari:
For protocols, lower than Tls 1.2 raises a warning message by `Console.Out` in addition to `EventSource` at `Netfx` and `NetCore`.

PR #3631, opened 2025-09-23
Merge | TdsParserStateObjectNative

edwardneal:
## Description

This builds on #3603, merging TdsParserStateObjectNative. The changes here aren't particularly controversial, although I'd advise reviewing b9e2beaf3fb35dcacae693ef694c955067ef70a0 with a little care - this change expects each SSL protocol's value in SslProtocols to be equivalent to the OR of its client and server values in `NativeProtocols`.

## Issues

Contributes to #1261.

## Testing

Automated tests pass locally, could someone run CI please?


src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParserStateObjectNative.cs
```diff
@@ -226,11 +229,6 @@ internal override void Dispose()
 
             if (sessionHandle != null || packetHandle != null)
             {
-                // Comment CloseMARSSession
-                // UNDONE - if there are pending reads or writes on logical connections, we need to block
```
mdaigle, line 230: Can we maintain this comment? Unless we know for sure it's no longer applicable.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObjectNative.cs
```diff
@@ -440,42 +421,8 @@ internal override uint SetConnectionBufferSize(ref uint unsignedPacketSize)
...
-            {
-                protocolVersion = SslProtocols.Ssl2;
-#pragma warning restore CS0618, CA5397
-            }
-            else //if (nativeProtocol.HasFlag(NativeProtocols.SP_PROT_NONE))
-            {
-                protocolVersion = SslProtocols.None;
-            }
-#pragma warning restore CA5398 // Avoid hardcoded SslProtocols values 
+            protocolVersion = (SslProtocols)nativeProtocolVersion;
```
mdaigle: I don't know if this is true? Maybe my hex/bit conversions are off.

Thinking through the first example if native protocol has flag SP_PROT_TLS1_2_CLIENT or SP_PROT_TLS1_2_SERVER, then it had hex value 0x00000400 or 0x00000800, 1024 of 2048 in decimal. This is mapped to SslProtocols.Tls12 which has decimal value 3072, so they wouldn't seem to match.

edwardneal: On its own, it wouldn't precisely match. However, this is only called in one place: `TdsParser.EnableSsl` ([netfx](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParser.cs#L964), [netcore](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs#L910)). In both cases, the return value is passed to the extension method [`SslProtocolsHelper.GetProtocolWarning`](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserHelperClasses.cs#L815) (and by extension, [`ToFriendlyName`](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserHelperClasses.cs#L756).) Both of these methods only operate by masking against `SslProtocols.Tls12` / etc. and checking for a non-zero return value.

In this case, either or both of the SP_PROT_TLS1_2_CLIENT and SP_PROT_TLS_1_2_SERVER bits being set would trigger the criteria in the only place the value's used.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/TdsParserStateObjectNative.cs
edwardneal: Yes - re-added.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserStateObjectNative.Windows.cs
```diff
@@ -410,10 +414,6 @@ internal override uint EnableSsl(ref uint info, bool tlsFirst, string serverCert
             AuthProviderInfo authInfo = new AuthProviderInfo();
             authInfo.flags = info;
             authInfo.tlsFirst = tlsFirst;
-            authInfo.certId = null;
-            authInfo.certHash = false;
-            authInfo.clientCertificateCallbackContext = IntPtr.Zero;
```
paulmedynski, line 415: `clientCertificateCallbackContext` will now be `null`.  Is that ok?

edwardneal: Yes. The existing netcore version of [`EnableSsl`](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObjectNative.cs#L420-L429) doesn't set this at all, at which point its default value is `null`.

paulmedynski (approved):

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObjectNative.cs
mdaigle: I want to make sure that what we have is robust to future changes as well. It still doesn't seems to me that it's safe to directly cast the value we're getting out of sni to SslProtocols, as the values don't match. I would rather maintain the NativeProtocols enum and do the mapping explicitly.

edwardneal: Fair enough. The mapping between SNI enum values and `SslProtocols` wasn't really in the right place, so I've moved it into the `SniNativeWrapper` layer and tweaked the `SniNativeMethods` implementations' return signatures to specify the re-added `SniSslProtocols` enum.

mdaigle (approved):

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObjectNative.cs
mdaigle: Thanks, looks good to me!

src/Microsoft.Data.SqlClient/src/Interop/Windows/Sni/SniNativeWrapper.cs
```diff
@@ -376,19 +376,67 @@ internal static uint SniSetInfo(SNIHandle pConn, QueryType qType, ref uint pbQIn
...
+            out System.Security.Authentication.SslProtocols pProtocolVersion)
+        {
+            uint returnValue = s_nativeMethods.SniWaitForSslHandshakeToComplete(pConn, dwMilliseconds, out SniSslProtocols nativeProtocolVersion);
+
+#pragma warning disable CA5398 // Avoid hardcoded SslProtocols values
+            if ((nativeProtocolVersion & SniSslProtocols.SP_PROT_TLS1_2) != 0)
+            {
+                pProtocolVersion = System.Security.Authentication.SslProtocols.Tls12;
+            }
+            else if ((nativeProtocolVersion & SniSslProtocols.SP_PROT_TLS1_3) != 0)
```
paulmedynski, line 392: Why do we prefer TLS 1.2 over TLS 1.3?  If both flags are set in `nativeProtocolVersion` we choose TLS 1.2.

src/Microsoft.Data.SqlClient/src/Interop/Windows/Sni/SniSslProtocols.cs
```diff
@@ -0,0 +1,32 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Interop.Windows.Sni
+{
+    internal enum SniSslProtocols : uint
+    {
+        // Protocol versions from native SNI
+        SP_PROT_SSL2_SERVER = 0x00000004,
```
paulmedynski, line 10: What's the significance of the naming convention here? It doesn't seem very C#-ish.

edwardneal: It's designed to align with the native SNI's names. In turn, I think the native SNI passes along the values from SChannel, which references `SP_PROT_[Protocol]_[CLIENT|SERVER]` as documented [here](https://learn.microsoft.com/en-us/windows/win32/api/schannel/ns-schannel-schannel_cred).

src/Microsoft.Data.SqlClient/src/Interop/Windows/Sni/SniNativeWrapper.cs
edwardneal: This aligns with the existing logic in [netcore](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParserStateObjectNative.cs#L434) and ports the translation logic over to netfx, but without knowing what the native SNI is doing in `SniWaitForSslHandshakeToComplete` I don't know.

One possible cause is documented in [SslProtocolsHelper](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserHelperClasses.cs#L760) (the only place the output of this mapping logic is used.) In 2020, #591 added a comment indicating that the native SNI didn't yet support TLS 1.3, so this change would make sure that our managed wrapper aligns with it. This support might have been added since then though.

PR #3632, opened 2025-09-23
Fix connection pool concurrency issue

cheenamalhotra:
Fixes connection pool concurrency issue caused due to waiting on multiple semaphore handles, instead of only creation handle during pool create request.

Issue was first introduced in PR https://github.com/dotnet/SqlClient/pull/2390 (https://github.com/dotnet/SqlClient/pull/2390/files#diff-d7c1dbddf2a7e205d7a85cbafbef036899799672dd5d1c486a64a05979e92230R1627) and has affected MDS v6 onwards.

Needs to be backported to v6.0 and v6.1 branches as well.

### DIFF
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectionPoolTest/ConnectionPoolStressTest.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectionPoolTest/ConnectionPoolStressTest.cs
@@ -55,7 +55,7 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests
         // Reflection fields for accessing internal connection properties
         private readonly FieldInfo? _internalConnectionField;
 
-        internal ConnectionPoolStressTest()
+        public ConnectionPoolStressTest()
         {
             try
             {


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/WaitHandleDbConnectionPool.cs
```diff
@@ -1506,7 +1506,9 @@ private void PoolCreateRequest(object state)
                             {
                                 // Obtain creation mutex so we're the only one creating objects
                                 // and we must have the wait result
-                                waitResult = WaitHandle.WaitAny(_waitHandles.GetHandles(withCreate: true), CreationTimeout);
+                                bool creationHandleObtained = _waitHandles.CreationSemaphore.WaitOne(CreationTimeout);
+                                waitResult = creationHandleObtained ? CREATION_HANDLE : WaitHandle.WaitTimeout;
```
paulmedynski: There's no need to assign `WaitHandle.WaitTimeout` here.  `waitResult` is only compared to `CREATION_HANDLE` for the remainder of its scope.  You can simplify these 2 lines to:
```
if (_waitHandles.CreationSemaphore.WaitOne(CreationTimeout))
{
    waitResult = CREATION_HANDLE;
}
```

paulmedynski: Looking closer, I don't think we need `waitResult` at all.  It's being used as a stand-in for `bool obtainedSemaphore`.  I think this would be clearer:
```
bool obtainedSemaphore = false;

try
{
    obtainedSemaphore = _waitHandles.CreationSemaphore.WaitOne(CreationTimeout);

    if (obtainedSemaphore)
    {
        ...
    }

    ...
}
finally
{
    if (obtainedSemaphore)
    {
        _waitHandles.CreationSemaphore.Release(1);
    }
}
```

paulmedynski: Bonus points if you want to create a SemaphoreScope RAII helper that obtains the semaphore on construction, and releases it on disposal.  That would let the compiler write the finally block for you, and would keep the semaphore scoped inside the try block.

edwardneal:
This might also need to be backported to v5.1. #2390 ported the change from netcore, which has exactly the same behaviour:

https://github.com/dotnet/SqlClient/blob/ee2d196f1dc12c80f4b2afb70b5d895bd7fac4b8/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/ProviderBase/DbConnectionPool.cs#L1482

v6.0 is the first point that it impacted netfx, but this has been the netcore behaviour from the outset of the driver (even when going back to the PR which added System.Data.SqlClient to its first version of netcore.)

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/WaitHandleDbConnectionPool.cs
mdaigle: Agreed, the wait result pattern is only really meaningful when doing a WaitAny to see which semaphore you obtained.

cheenamalhotra: I started with bare minimum changes to make it work :P

cheenamalhotra:
> This might also need to be backported to v5.1. #2390 ported the change from netcore, which has exactly the same behaviour:
> 
> https://github.com/dotnet/SqlClient/blob/ee2d196f1dc12c80f4b2afb70b5d895bd7fac4b8/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/ProviderBase/DbConnectionPool.cs#L1482
> 
> v6.0 is the first point that it impacted netfx, but this has been the netcore behaviour from the outset of the driver (even when going back to the PR which added System.Data.SqlClient to its first version of netcore.)

I've verified this directly as well as in the test suite, this doesn't seem to cause any concerns with the repro I have, also added as a test case in the PR.

The same test case fails with NetFx (without fix), but passes in NetCore without any changes.

I'm going to give that a pass unless there are any reported concerns. It maybe possible that there's a different difference between NetFx and NetCore source that is part of the problem. But I'll leave that until code merge happens for all the layers.

The test case will fail if after merge there is inconsistency in functionality and we'll know!

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectionPoolTest/ConnectionPoolStressTest.cs
```diff
@@ -0,0 +1,423 @@
...
+        /// <summary>
+        /// Generic stress test method that works with both SQL client libraries using DbConnection/DbCommand
+        /// </summary>
+        private void RunStressTest(
+            string connectionString,
+            Func<DbConnection, bool> doomAction,
+            bool async = false)
+        {
+            var threads = new Thread[ConcurrentConnections];
+            var barrier = new Barrier(ConcurrentConnections);
```
mdaigle:
```suggestion
            using var barrier = new Barrier(ConcurrentConnections);
```

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectionPoolTest/ConnectionPoolStressTest.cs
```diff
@@ -0,0 +1,423 @@
...
+            var countdown = new CountdownEvent(ConcurrentConnections);
```
mdaigle:
```suggestion
            using var countdown = new CountdownEvent(ConcurrentConnections);
```

mdaigle (dismissed):

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectionPoolTest/ConnectionPoolStressTest.cs
```diff
@@ -0,0 +1,414 @@
...
+        // Reflection fields for accessing internal connection properties
+        private static readonly FieldInfo _internalConnectionField;
+
+        static ConnectionPoolStressTest()
+        {
+            try
+            {
+                // Cache reflection info for Microsoft.Data.SqlClient
+                Type msDataConnectionType = typeof(SqlConnection);
+                _internalConnectionField = msDataConnectionType.GetField("_innerConnection", BindingFlags.NonPublic | BindingFlags.Instance);
```
paulmedynski, line 64: GetField returns null if it can't find the field.  I would suggest enabling nullable context for new files like this.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectionPoolTest/ConnectionPoolStressTest.cs
```diff
@@ -0,0 +1,414 @@
...
+            }
+            catch (Exception ex)
+            {
+                Console.WriteLine($"Warning: Failed to initialize reflection for connection dooming: {ex.Message}");
```
paulmedynski, line 68: I'm not sure why we would want the tests to run if we can't find the inner connection field.

Maybe this setup belongs in a constructor instead, and then failures here can properly fail the tests without blowing up the entire test process.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectionPoolTest/ConnectionPoolStressTest.cs
```diff
@@ -0,0 +1,414 @@
...
+        }
+
+        /// <summary>
+        /// Dooms a Microsoft.Data.SqlClient connection by calling DoomThisConnection on its internal connection
+        /// </summary>
+        private static bool DoomMicrosoftDataConnection(SqlConnection connection)
+        {
+            try
+            {
+                if (_internalConnectionField?.GetValue(connection) is object internalConnection)
```
paulmedynski: Return early on false cases - avoid nesting and fragile else conditions?

You won't need the `?` if acquiring the field instance becomes a class invariant.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectionPoolTest/ConnectionPoolStressTest.cs
```diff
@@ -0,0 +1,414 @@
...
+                return false;
+            }
+        }
+
+        #endregion
+
+        #region Configuration
+
+        /// <summary>
+        /// Sets up connection strings for both SQL client libraries with optimized settings
```
paulmedynski: Not sure what 'both' means here.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectionPoolTest/ConnectionPoolStressTest.cs
```diff
@@ -0,0 +1,414 @@
...
+                    {
+                        if (doomConnections && doomAction != null)
+                        {
+                            // Dooming thread - barriers inside using block to doom before disposal
+                            using var conn = new SqlConnection(connectionString);
+                            if (async)
+                            {
+                                await conn.OpenAsync();
+                            }
+                            else
```
paulmedynski, line 240: Missing `{ }`.  Here and several places below.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectionPoolTest/ConnectionPoolStressTest.cs
```diff
@@ -0,0 +1,414 @@
...
+                Assert.Equal(maxPoolSize, connections.Count);
+            }
+            catch
+            {
+                return false;
+            }
+            finally
+            {
+                // Clean up all connections
+                foreach (SqlConnection conn in connections)
```
paulmedynski, line 369: Too bad there isn't a DisposingList<> that would do this for you.

paulmedynski (changes requested):

PR #3639, opened 2025-09-26
Cleanup | Remove GenerateResourceStringsSource MSBuild target

edwardneal:
## Description

At the moment, the build has a custom target, `GenerateResourceStringsSource`. This target has been around since the start of the repo; it runs prior to compile and generates a file which contains one const for each resource string. An example file is

```cs
internal partial class StringsHelper : Strings
{
    internal class ResourceNames
    {
        internal const string ADP_Ascending = @"Ascending";
        // ..etc.
    }
}
```

This `ResourceNames` class is only used in the `ResCategory` and `ResDescription` attributes applied to certain properties within the implementation assemblies for netfx and netcore.

This is redundant because the [`ResCategoryAttribute`](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/src/Resources/ResCategoryAttribute.cs) and [`ResDescriptionAttribute`](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/src/Resources/ResDescriptionAttribute.cs) classes will already look up the parameter passed to them as a resource string. Switching to using this functionality (with associated test) means that we don't need the `StringsHelper` class, which means that we don't need the target which generates it.

## Issues

None.

## Testing

Two automated tests added to prove that the descriptions contain the localized values as we expect. One manual check by binding a PropertyGrid to a `SqlConnectionStringBuilder` (in netcore and netfx) to make sure that WinForms isn't performing something unusual to get the description in a way which bypasses the custom logic in `ResCategoryAttribute` and `ResDescriptionAttribute`.

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlParameter.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlParameter.cs
@@ -899,7 +899,7 @@ namespace Microsoft.Data.SqlClient
         }
 
         /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlParameter.xml' path='docs/members[@name="SqlParameter"]/ToString/*' />
-        [ResCategory("Data")]
+        [ResCategory(nameof(Strings.DataCategory_Data))]
         public override string ToString() => ParameterName;
 
         /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlParameter.xml' path='docs/members[@name="SqlParameter"]/SourceVersion/*' />


cheenamalhotra:
This is actually used by Localization CI for generation of resource strings. Removing this target might break things. Would recommend taking a closer look for safety or adding necessary documentation around it to define its purpose.

@paulmedynski @mdaigle fyi for ref: [Link to usage](https://sqlclientdrivers.visualstudio.com/ADO.Net/_build/results?buildId=126197&view=logs&j=275f1d19-1bd8-5591-b06b-07d489ea915a&t=4096face-155a-5c7d-a923-db1f83fd19d1&l=702)

benrr101:
I also love these changes, especially since my IDE complains about ResourceNames not existing a lot.
@cheenamalhotra is there any way we can validate the localization workflow prior to taking this PR?

cheenamalhotra:
I don't think there's an easy way to validate this before the merge - you can carefully scan for the targets being removed and what its doing from the logs.

or

Fail Fast: Go ahead and merge, be diligent to observe the Loc CI and be ready to revert this change if Localization CI is not happy after few days.

edwardneal:
Can the Localization CI run against a branch other than main? I'm happy to resubmit a PR which merges to another branch if that'd help validate the behaviour.

edwardneal:
Updating to hold this open for now.

benrr101:
@cheenamalhotra we can't fail fast and merge (and back out on failure) until you approve - your requested changes are blocking :)

edwardneal:
Thanks for merging this. I've not seen any build failures related to the localization CI, have there been any non-public failures?

PR #3644, opened 2025-09-29
Merge | SqlCommand Execute Reader

benrr101:
## Description
In this installment of merging SqlCommand, the focus is on the ExecuteReader methods and supporting structures. As with previous PRs in this series, a new partial has been introduced for the Reader methods. Every commit is a bite-sized merge of a single method or so. This is probably the largest of the partials due to the many methods that make up execution of readers. It can definitely be simplified by using proper async primitives, and I have plans to do so. Please also take special look at this PR since all the other execute partials call into this partial.

The following methods were merged:
* BeginExecuteReader
* BeginExecuteReaderInternal
* BeginExecuteReaderInternalReadStage
* BuildExecute
* BuildExecuteSql
* BuildPrpeExec
* BuildRPC
* CleanupExecuteReaderAsync
* CompleteAsyncExecuteReader - both arguments were made required to clear up usages of the method
* EndExecuteReader
* EndExecuteReaderAsync
* EndExecuteReaderInternal
* ExecuteDbDataReader
* ExecuteDbDataReaderAsync
* ExecuteReader
* ExecuteReaderAsync
* ExecuteReaderWithRetry
* ExecuteReaderWithRetryAsync
* InternalEndExecuteReader
* InternalExecuteReaderAsync
* InternalExecuteReaderWithRetryAsync
* GenerateEnclavePackage
* RunExecuteReader
* RunExecuteReaderTds
* RunExecuteReaderTdsSetupContinuation
* RunExecuteReaderTdsSetupReconnectionContinuation - This was factored out of the netfx implementation to match the netcore implementation
* RunExecuteReaderTdsWithTransparentParameterEncryption
* SetCachedCommandExecuteReaderAsyncContext

And also the ExecuteReaderAsyncCallContext internal class

## Issues
Continuation of work for #1261 

## Testing
Code still compiles, CI suite will validate


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Reader.cs
```diff
@@ -0,0 +1,1825 @@
...
+                        out usedCache,
+                        asyncWrite: isAsync,
+                        isRetry: true,
+                        method);
+                }
+            }
+            else
+            {
+                // @TODO: Reminder, this is where we execute if transparent parameter encryption is not needed.
+                usedCache = false;
```
edwardneal: This line is new - is it necessary? We already set it earlier in the procedure.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Reader.cs
```diff
@@ -0,0 +1,1825 @@
...
+            CheckThrowSNIException();
+
+            SqlDataReader reader = CompleteAsyncExecuteReader(isInternal, forDescribeParameterEncryption: false);
+            Debug.Assert(_stateObj is null, "non-null state object in InternalEndExecuteReader");
+            return reader;
+            // @TODO: CER Exception Handling was removed here (see GH#3581)
+        }
+
+        // @TODO: We're passing way too many arguments around here... can we simplify this some?
+        private SqlDataReader RunExecuteReader(
```
edwardneal, line 1071: There was a comment here:
```cs
// task is created in case of pending asynchronous write, returned SqlDataReader should not be utilized until that task is complete
```
Was its removal deliberate? It seemed useful in isolation, but maybe other methods make enough use of the same pattern that it's already documented elsewhere.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Reader.cs
```diff
@@ -0,0 +1,1825 @@
...
+            SetUpRPCParameters(rpc, inSchema, _parameters);
+            return rpc;
+        }
+
+        /// <summary>
+        /// Build the RPC record header for sp_executesql and add the parameters.
+        /// </summary>
+        /// <remarks>
+        /// Prototype for sp_executesql is:
+        /// sqp_executesql(@batch_text nvarchar(4000), @batch_params nvarchar(4000), param1, param2, ...)
```
edwardneal:
```suggestion
        /// sp_executesql(@batch_text nvarchar(4000), @batch_params nvarchar(4000), param1, param2, ...)
```

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Reader.cs
```diff
@@ -0,0 +1,1825 @@
...
+
+        private SqlDataReader EndExecuteReaderAsync(IAsyncResult asyncResult)
+        {
+            Debug.Assert(!_internalEndExecuteInitiated || _stateObj is null);
+
+            SqlClientEventSource.Log.TryCorrelationTraceEvent(
+                "SqlCommand.EndExecuteReaderAsync | API | Correlation | " +
+                $"Object Id {ObjectID}, " +
+                $"Activity Id {ActivityCorrelator.Current}, " +
+                $"Client Connection Id {_activeConnection.ClientConnectionId}, " +
```
edwardneal: Small difference here - the previous version embedded `Connection?.ClientConnectionId`, which handled nulls.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Reader.cs
```diff
@@ -815,6 +815,105 @@ private void GenerateEnclavePackage()
...
+                        context.Dispose();
+                        command.CleanupExecuteReaderAsync(task, source, operationId);
+                    },
+                    scheduler: TaskScheduler.Default);
+            }
+            catch (Exception e)
+            {
+                source.SetException(e);
+                context?.Dispose();
+            }
```
edwardneal, line 1027: This needs to have the logic from netcore L930-933, where it calls WriteCommandError if necessary.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Reader.cs
```diff
@@ -0,0 +1,1825 @@
...
+        }
+
+        #endregion
+
+        #region Private Methods
+
+        /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlCommand.xml' path='docs/members[@name="SqlCommand"]/ExecuteDbDataReader[@name="CommandBehavior"]/*'/>
+        protected override DbDataReader ExecuteDbDataReader(CommandBehavior behavior)
+        {
+            // @TODO: Yknow, we use this all over the place. It could be factored out.
```
edwardneal, line 215: I'd ultimately like to see us pushing these strings out via ILogger; this also exposes these logs to OpenTelemetry. When combined with the `[LoggerMessage]` source generator, we'd have one method per unique logging string.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Reader.cs
benrr101: 🤔 Hmm, I guess not.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Reader.cs
benrr101: I guess it's fair to keep - I think I was looking at it like a lot of those comments aren't super useful, and the task wasn't actually being created in that method. I can add it back, but I really want to rewrite most of this code to not be so awfully complex.

edwardneal: I agree in general, I've not commented on most of them. This one in particular seemed useful though, it's definitely not an intuitive pattern.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Reader.cs
benrr101: Maybe someday!

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Reader.cs
mdaigle: The best would be to return `Task<SqlDataReader>`. This whole method looks like it needs a refactor at some point.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Reader.cs
mdaigle: Would also like to see this removed

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlCommand.netfx.cs
```diff
@@ -2689,67 +2689,6 @@ private void ReadDescribeEncryptionParameterResults(
...
-                    {
-                        ((SqlCommand)state).CachedAsyncState?.ResetAsyncState();
-                        if (exception != null)
-                        {
-                            throw exception;
-                        }
-                    },
-                    onCancellation: static (object state) => ((SqlCommand)state).CachedAsyncState?.ResetAsyncState(),
-                    connectionToDoom: null,
-                    connectionToAbort: _activeConnection);
```
mdaigle, line 2990: These were dropped in the merged version. Can you comment on the impact of that?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Reader.cs
```diff
@@ -136,6 +136,143 @@ internal SqlDataReader RunExecuteReader(
...
+                    // for "forDescribeParameterEncryption" is risky, we introduced a new parameter
+                    // to determine whether we should cache a data reader's metadata or not.
+                    if (shouldCacheForAlwaysEncrypted)
+                    {
+                        _cachedMetaData = ds.MetaData;
+                    }
+                    else
+                    {
+                        // We need this call to ensure the data reader is properly initialized, the
+                        // getter is initializing state in SqlDataReader.
```
mdaigle, line 828: This is an interesting pattern...
Nothing to do here, just commenting.

mdaigle (approved):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Reader.cs
edwardneal: We'll eventually need an async version of this property's getter - it's a prerequisite of async transaction operations and SqlDataReader.GetSchemaAsync support.

mdaigle: Either the constructor should do the initialization, or we should have a dedicated initialization method. It's too weird to initialize an object by expecting side effects from some random getter.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Reader.cs
benrr101: Yeah. Very much so.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Reader.cs
benrr101: I makes sense from a perspective of "don't fetch stuff until it's needed" buuuut to rely on the *side effects* in a getter is bad practice.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlCommand.netfx.cs
benrr101: :/ Not really... Really the way all these components interact with each other is a complete enigma to me. In my mind, it's not one can be rectified by trying to picking it all apart, but rather needs to be rewritten to simplify. My suspicion is if everything else in the netcore implementation works appropriately, then we shouldn't need to bring this specific line over from netfx. But, since you bring it up, and since I can't say for sure, I'll conditionally add it on netfx.

edwardneal: These made a lot more sense before we removed the defunct CER blocks. [`ContinueTaskWithState`](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlUtil.cs#L222) doesn't meaningfully use the `connectionToDoom` or `connectionToAbort` parameters any more, so passing them is a nop.

There's room to untangle quite a bit of this logic in AsyncHelper.

benrr101: Ah, thanks for that. Yes, I'd been toying with an idea to strongly type the "with state" async helpers, got about waist deep into it and realized it was going to be more work than I had time for (specifically with some legit uses of typed CompletionTaskSource). This change (removing the defunct connectionToDoom/Abort from ContinueTaskWithState seems like a low hanging fruit to clean up.

mdaigle (approved):

PR #3645, opened 2025-09-30
Style | Enforce Braces via EditorConfig

benrr101:
## Description
As part of recent discussion, it's looking like it will be beneficial to enforce style guidelines we can all agree on as much as possible. This is the first in probably a series of a few PRs that aim do just that. This PR applies to using braces (sometimes called curly braces) `{ }` for all blocks of code that follow an `if`, `for`, `while`, etc statement.

**Example Bad:**
```
if (foo == bar)
    doSomething();
```

**Example Good:**
```
if (foo == bar)
{
    doSomething();
}
```

Why is this change being made? This is a well-know pathway for bugs to emerge. Take a look at this bug in OSX's SSL library that was caused by a mistake in not using braces around an if statement: https://www.imperialviolet.org/2014/02/22/applebug.html

The rule has been added to the .editorconfig file via the [`csharp_prefer_braces`](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0011#csharp_prefer_braces) rule. All files **except SqlCommand and RetryLogicTestHelper** (since they have pending PRs) have been updated to add braces where missing. There were ~750 instances (though can be cut in half since the files were used twice). I've held back as much as possible and _only_ added braces where missing, and newlines where appropriate (there was even a bug I noticed and I didn't fix it!).

## Issues
N/A

## Testing
Code still compiles, CI should validate.

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParser.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParser.cs
@@ -678,7 +678,9 @@ namespace Microsoft.Data.SqlClient
 
 #if NET
                 if (LocalAppContextSwitches.UseManagedNetworking)
+                {
                     _pMarsPhysicalConObj.IncrementPendingCallbacks();
+                }
 #endif
 
                 uint info = 0;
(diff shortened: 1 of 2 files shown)


mdaigle (approved):

ErikEJ:
So great you are finally in a place to do this.

cheenamalhotra (approved):

Issue #3641, opened 2025-09-29
Support routing to named read replicas


mdaigle:
## Summary
Azure SQL hyperscale is introducing a new gateway feature to load balance read requests between named reader replicas. Load balancing support is accessed by connecting to a "reader endpoint" FQDN at the gateway, separate from the regular gateway FQDN. The load balancing occurs without the client needing to specify the replica's database name (each replica may have its own unique name and may differ from the primary's database name). 

The current routing ENVCHANGE token supports protocol, port, and FQDN. To connect to a particular database at the routed location, we need to introduce a mechanism for the gateway to specify which database to use.

### TDS Updates
Update TDS to include a new FEATUREEXT 0x10 (EnhancedRouting) token, and an accompanying FEATUREEXTACK to indicate gateway support.

Update TDS to include a new ENVCHANGE 0x21 (EnhancedRouting) payload type.  The payload includes all of the data currently present in 0x20 (Routing) plus a required database name field. 

#### ENVCHANGE 0x21 (EnhancedRouting) payload shape:
```
TokenType: BYTE				 // 0xE3 = ENVCHANGE token 
Length:    USHORT                // byte length of EnvValueData 
Type:      BYTE                  // 0x15 = Type 21, EnhancedRouting 
 
EnvValueData: 
  Protocol:         BYTE        // 0x00 = TCP/IP 
  ProtocolProperty: USHORT      // Port 8080 (hex 1F90) 
  AlternateServer:  US_VARCHAR  // USHORT character stream 
  AlternateDatabase: US_VARCHAR // USHORT character stream, not to exceed 128 characters 
```

### Driver Behavior
New driver versions will always send the new FEATUREEXT 0x0F (EnhancedRouting) token on all connection requests.

If (and only if) EnhancedRouting support is successfully negotiated with the server (via the ENHANCEDROUTINGSUPPORT featureext ack), the driver will respect ENVCHANGE 0x21 (EnhancedRouting) tokens, otherwise the token is rejected.

This feature is separate from the ReadOnly intent feature and does not require ReadOnly intent to be set.

PR #3646, opened 2025-09-30
Enhanced routing envchange token support

mdaigle:
## Description
Adds enhanced routing envchange support in service of #3641 
Changes can be followed commit-by-commit for ease of review.

## Issues

#3641 

## Testing

Adds simulated server tests to verify enhanced routing behavior.


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
paulmedynski: This log could be misleading if the `data[0] != 1`.  It implies that enhanced routing is enabled.  Perhaps move it below line 3024 and include whether or not the feature is enabled as well.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -3152,10 +3145,19 @@ internal ServerInfo(SqlConnectionString userOptions, RoutingInfo routing, string
             {
                 UserServerName = string.Format(CultureInfo.InvariantCulture, "{0},{1}", routing.ServerName, routing.Port);
             }
+
+            if (routing == null || routing.DatabaseName == null)
```
paulmedynski: Would this statement be easier to understand if it were flipped?
```
// Always prefer a routing database name from the server, if supplied.
if (routing is not null && routing.DatabaseName is not null)
{
    ResolvedDatabaseName = routing.DatabaseName;
}
else
{
    ResolvedDatabaseName = userOptions.InitialCatalog;
}
```

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/FeatureExtensionEnablementTriState.cs
paulmedynski: Maybe `FeatureExtensionBehaviour` ?  It would be funny if you add a 4th member later :)

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/GenericTdsServer.cs
paulmedynski: Shouldn't this depend on how the server intends to respond?  If `EnableEnhancedRouting` is `Disabled` or `DoNotAcknowledge`, then support isn't enabled.

Maybe this flag should be called `DidClientAskForEnhancedRouting` or something more succinct.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/RoutingTdsServerArguments.cs
```diff
@@ -24,6 +24,12 @@ public class RoutingTdsServerArguments : TdsServerArguments
         /// </summary>
         public string RoutingTCPHost { get; set; } = string.Empty;
 
+        /// <summary>
+        /// Database name to use at the routed location.
+        /// Should only be set when doing enhanced routing.
```
paulmedynski: "Setting this to a non-empty value will cause the server to include an enhanced routing ENVCHANGE token in the Login Response message.  An empty value will include a legacy routing ENVCHANGE token."

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionEnhancedRoutingTests.cs
```diff
@@ -0,0 +1,118 @@
...
+[Collection("SimulatedServerTests")]
+public class ConnectionEnhancedRoutingTests
+{
+    [Fact]
+    public void RoutedConnection()
+    {
+        // Arrange
+        using TdsServer server = new(new()
+        {
+            Log = Console.Out
```
paulmedynski: Our tests should probably be using xUnit's [ITestOutputHelper](https://xunit.net/docs/capturing-output).  Just an FYI.

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionEnhancedRoutingTests.cs
```diff
@@ -0,0 +1,118 @@
...
+            });
+        router.Start();
+        router.EnableEnhancedRouting = FeatureExtensionEnablementTriState.Enabled;
+
+        string connectionString = (new SqlConnectionStringBuilder()
+        {
+            DataSource = $"localhost,{router.EndPoint.Port}",
+            Encrypt = false,
+            ConnectTimeout = 10000
+        }).ConnectionString;
```
paulmedynski, line 49: This arrange stuff is common to both tests.  Can you pull it out?

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionEnhancedRoutingTests.cs
```diff
@@ -0,0 +1,118 @@
...
+                RequireReadOnly = false
+            });
+        router.Start();
+        router.EnableEnhancedRouting = FeatureExtensionEnablementTriState.Enabled;
+
+        string connectionString = (new SqlConnectionStringBuilder()
+        {
+            DataSource = $"localhost,{router.EndPoint.Port}",
+            Encrypt = false,
+            ConnectTimeout = 10000
```
paulmedynski, line 48: Do we need to test the case where InitialCatalog is explicitly specified by the app, and ignored during routing?

What about the cases where the server ignores or explicitly denies enhanced routing?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -2277,6 +2259,21 @@ internal void OnEnvChange(SqlEnvChange rec)
                     RoutingInfo = rec._newRoutingInfo;
                     break;
 
+                case TdsEnums.ENV_ENHANCEDROUTING:
+                    SqlClientEventSource.Log.TryAdvancedTraceEvent("<sc.SqlInternalConnectionTds.OnEnvChange|ADV> {0}, Received enhanced routing info", ObjectID);
```
benrr101, line 2263: I think the tracing format we're trying to use going forward is the `xyz | abc | whatever` format. But if the rest of the strings in this file are using the `<abc...> asdfasdfads` format, then I won't complain.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -3134,10 +3145,19 @@ internal ServerInfo(SqlConnectionString userOptions, RoutingInfo routing, string
             {
                 UserServerName = string.Format(CultureInfo.InvariantCulture, "{0},{1}", routing.ServerName, routing.Port);
             }
+
+            if (routing == null || routing.DatabaseName == null)
+            {
+                ResolvedDatabaseName = userOptions.InitialCatalog;
```
benrr101: I can has ternary ? 👉👈🥺

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -3073,51 +3073,63 @@ private TdsOperationStatus TryProcessEnvChange(int tokenLength, TdsParserStateOb
...
                         {
-                            return result;
-                        }
-                        env._newRoutingInfo = new RoutingInfo(protocol, port, serverName);
-                        ushort oldLength;
-                        result = stateObj.TryReadUInt16(out oldLength);
-                        if (result != TdsOperationStatus.Done)
-                        {
-                            return result;
+                            ushort newLength;
```
benrr101, line 3077: So the whole deal with this big block o changes was you added scope `{}` to this case? ... If we can avoid doing that that'd be better. But, if you reallly reallllly want to use a new variable in two scopes, I'll let it slide

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -3130,6 +3142,91 @@ private TdsOperationStatus TryProcessEnvChange(int tokenLength, TdsParserStateOb
...
 
+        /// <summary>
+        /// Reads an enhanced routing token from the stream and populates the provided <see cref="SqlEnvChange"/> object.
+        /// </summary>
+        /// <param name="stateObj">The state object containing the stream.</param>
+        /// <param name="env">The object to populate.</param>
+        /// <returns>Returns a <see cref="TdsOperationStatus"/> to indicate the status of the operation.</returns>
+        private TdsOperationStatus TryProcessEnhancedRoutingToken(TdsParserStateObject stateObj, SqlEnvChange env)
+        {
+            ushort newLength;
```
benrr101: Can we inline the declarations here?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -3130,6 +3142,91 @@ private TdsOperationStatus TryProcessEnvChange(int tokenLength, TdsParserStateOb
             return TdsOperationStatus.Done;
         }
 
+        /// <summary>
+        /// Reads an enhanced routing token from the stream and populates the provided <see cref="SqlEnvChange"/> object.
```
benrr101, line 3146: Not required, but I think it'd be really nice to have a layout of the token structure in a comment either a remarks or comment in the body, maybe?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsEnums.cs
```diff
@@ -190,29 +190,7 @@ internal static class TdsEnums
...
-            ENVCHANGE_DEFECTDTC = ENV_DEFECTDTC,
-            ENVCHANGE_LOGSHIPNODE = ENV_LOGSHIPNODE,
-            ENVCHANGE_PROMOTETRANSACTION = ENV_PROMOTETRANSACTION,
-            ENVCHANGE_TRANSACTIONMANAGERADDRESS = ENV_TRANSACTIONMANAGERADDRESS,
-            ENVCHANGE_TRANSACTIONENDED = ENV_TRANSACTIONENDED,
-            ENVCHANGE_SPRESETCONNECTIONACK = ENV_SPRESETCONNECTIONACK,
-            ENVCHANGE_USERINSTANCE = ENV_USERINSTANCE,
-            ENVCHANGE_ROUTING = ENV_ROUTING
-        }
+        public const byte ENV_ENHANCEDROUTING = 21; // Enhanced Routing (ROR) information
```
benrr101, line 193: In terms of staying consistent and eliminating cruft, this is good. Not really a fan of having these be arbitrary constants rather than an enum, but we can tackle that some other day. TdsEnums needs to be burned to the ground, imo 🔥🔥🚒

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsEnums.cs
```diff
@@ -241,8 +219,9 @@ public enum EnvChangeType : byte
         public const byte FEATUREEXT_SQLDNSCACHING = 0x0B;
         public const byte FEATUREEXT_JSONSUPPORT = 0x0D;
         public const byte FEATUREEXT_VECTORSUPPORT = 0x0E;
+        public const byte FEATUREEXT_ENHANCEDROUTINGSUPPORT = 0x0F;
         // TODO: re-verify if this byte competes with another feature
-        public const byte FEATUREEXT_USERAGENT = 0x0F;
+        public const byte FEATUREEXT_USERAGENT = 0x10;
```
benrr101, line 224: ??? This seems like a big change - I trust this is accurate?

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionEnhancedRoutingTests.cs
benrr101: More so, I think we should be avoiding outputting random text during a test anyhow. But I think that's more of an existing issue that we're just not touching today.

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionEnhancedRoutingTests.cs
benrr101: I'll counter that this doesn't meet the rule of 3 for refactoring :)

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/FeatureExtensionNegotiationTests.cs
```diff
@@ -0,0 +1,112 @@
...
+            Assert.Equal((byte)TDSFeatureID.EnhancedRoutingSupport, (byte)token.FeatureID);
+
+            clientFeatureExtensionFound = true;
+        };
+
+        using SqlConnection sqlConnection = new(_connectionString);
+
+        // Act
+        sqlConnection.Open();
+
```
benrr101, line 58: nit: extra line here and in below instances

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/FeatureExtensionNegotiationTests.cs
```diff
@@ -0,0 +1,112 @@
...
+
+        // Assert
+        Assert.False(((SqlInternalConnectionTds)sqlConnection.InnerConnection).IsEnhancedRoutingSupportEnabled);
+    }
+}
+
+public class SimulatedServerFixture : IDisposable
```
benrr101: Please put top level classes in their own files

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/FeatureExtensionEnablementTriState.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+
+namespace Microsoft.SqlServer.TDS.Servers
+{
+    public enum FeatureExtensionEnablementTriState
```
benrr101: Can we do this as effectively with a `bool?` ?

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/GenericTdsServer.cs
```diff
@@ -681,6 +689,40 @@ protected virtual TDSMessageCollection OnAuthenticationCompleted(ITDSServerSessi
                 }
             }
 
+            if (session.IsEnhancedRoutingSupportEnabled &&
+                EnableEnhancedRouting != FeatureExtensionEnablementTriState.DoNotAcknowledge)
+            {
+                // Create ack data (1 byte: IsEnabled)
+                byte[] data = new byte[1];
+
+                if (EnableEnhancedRouting == FeatureExtensionEnablementTriState.Enabled)
```
benrr101: Can has ternary again?

benrr101: In fact, can just do
```
byte[] data = EnableEnhancedRouting is FeatureExtensionEnablementTristate.Enabled
    ? [1]
    : [0];
```

(and if we switch from tristate to bool? we could condense even more to)
```
byte[] data = IsEnhancedRoutingEnabled ? [1] : [0]
```

benrr101 (approved):
Generally speaking it's pretty good. I have a handful of style comments and a couple general questions. But I don't think they're worth holding up the PR if you're not interested (I'm kinda breaking my own rules by not taking a super hard stance on it, tho)

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
mdaigle: agreed

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/GenericTdsServer.cs
mdaigle: agreed

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionEnhancedRoutingTests.cs
mdaigle: This is here as a convenience when working with these tests. It's hard to debug directly because the "server" runs on a separate thread. These were left over from when I was writing the tests. I'll remove them and add a todo to look into test output helper.

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionEnhancedRoutingTests.cs
mdaigle: I'm kind of anti-abstraction for tests because I think it hides what's actually involved to make a test work. It's easy to reuse test abstractions and then you end up with tests that "depend" on a lot of things they don't actually need or that pass because they're not actually testing what you think they're testing.

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionEnhancedRoutingTests.cs
mdaigle: added

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
mdaigle: They are. If they're standardized on the format you mentioned as part of the merge work, then I'll bring these new messages in line when we bring this branch to main.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/TdsParser.cs
mdaigle: It feels silly to rely on a variable from another case. It also feels silly to use a new variable name just to avoid a collision. Using the {} keeps things well scoped.

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionEnhancedRoutingTests.cs
paulmedynski: I think ITestOutputHelper gathers output up from whatever threads they are emitted by, and then outputs them (on failure by default?) nicely.  I don't think Console.Out prevents threads from mixing lines of output together, so we could get a mess.  Agreed that this is an enhancement for another PR.

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/ConnectionEnhancedRoutingTests.cs
paulmedynski: Did you forget to push some commits?  I see responses to my comments, but no corresponding commits making any changes.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsEnums.cs
mdaigle: Yes, the same change is present in #3606

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/FeatureExtensionEnablementTriState.cs
mdaigle: I guess we could? But I'm not sure it would convey the intent as well.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
mdaigle: I'll do you one better and use null coalescing.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
benrr101: No worries. I didn't know these intentions when I started on merge so some of my older merged files have #if/#else for the two different formats 🤦‍♂️

PR #3660, opened 2025-10-04
Performance | Use lower-allocation AE primitives in SqlColumnEncryptionCertificateStoreProvider

edwardneal:
## Description

This is my final PR in this set of changes to this area, picking up where #3612 left off. Benchmarks for this PR are located in #3554.

The previous PR used the AlwaysEncrypted primitives in `SqlColumnEncryptionCspProvider` and `SqlColumnEncryptionCngProvider`. This one introduces them to `SqlColumnEncryptionCertificateStoreProvider`.

While verifying this, I also noticed that the `TestRsaCryptoWithNativeBaseline` test was calling private methods using reflection, but that there was a TODO item in the test to use the public API surface instead. I've done this.

## Issues

PR 1/3: #3554.
PR 2/3: #3612.
PR 3/3: this one.

## Testing

As in the previous PR, automated tests pass. I also created an AE-enabled table in SSMS which used a certificate store-based master key and confirmed that I was able to query it and insert records into it.

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlColumnEncryptionCertificateStoreProvider.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlColumnEncryptionCertificateStoreProvider.cs
@@ -3,10 +3,12 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
-using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
 using System.Security.Cryptography;
 using System.Security.Cryptography.X509Certificates;
-using System.Text;
+using Microsoft.Data.SqlClient.AlwaysEncrypted;
+
+#nullable enable
 
 namespace Microsoft.Data.SqlClient
 {
@@ -16,7 +18,7 @@ namespace Microsoft.Data.SqlClient
         // Constants
         //
         // Assumption: Certificate Locations (LocalMachine & CurrentUser), Certificate Store name "My"
-        // Certificate provider name (CertificateStore) dont need to be localized.
+        // Certificate provider name (CertificateStore) don't need to be localized.
 
         /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionCertificateStoreProvider.xml' path='docs/members[@name="SqlColumnEncryptionCertificateStoreProvider"]/ProviderName/*' />
         public const string ProviderName = @"MSSQL_CERTIFICATE_STORE";
@@ -37,12 +39,12 @@ namespace Microsoft.Data.SqlClient
         internal const string RSAEncryptionAlgorithmWithOAEP = @"RSA_OAEP";
 
         /// <summary>
-        /// LocalMachine certificate store location. Valid certificate locations are LocalMachine and CurrentUser.
+        /// LocalMachine certificate store location. Valid certificate locations are LocalMachine (on Windows) and CurrentUser.
         /// </summary>
         private const string CertLocationLocalMachine = @"LocalMachine";
 
         /// <summary>
-        /// CurrentUser certificate store location. Valid certificate locations are LocalMachine and CurrentUser.
+        /// CurrentUser certificate store location. Valid certificate locations are LocalMachine (on Windows) and CurrentUser.
         /// </summary>
         private const string CertLocationCurrentUser = @"CurrentUser";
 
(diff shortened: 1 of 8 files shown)


src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/CoreCryptoTests.cs
```diff
@@ -64,24 +67,49 @@ public void TestRsaCryptoWithNativeBaseline()
             byte[] rsaKeyPair = cryptoParametersListForTest[0].RsaKeyPair;
             byte[] rsaPfx = cryptoParametersListForTest[1].RsaKeyPair;
 
-            // For each crypto vector, run the test to compare the output generated through sqlclient's code and the native code.
-            foreach (CryptoVector cryptoParameter in cryptoParametersListForTest)
+            // Convert the PFX into a certificate and install it into the local user's certificate store.
+            // We can only do this cross-platform on the CurrentUser store, which matches the baseline data we have.
+            Debug.Assert(rsaPfx != null && rsaPfx.Length > 0);
```
paulmedynski: Should this be an xUnit Assert() ? Here and below.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlColumnEncryptionCertificateStoreProvider.cs
```diff
@@ -52,511 +54,272 @@ public class SqlColumnEncryptionCertificateStoreProvider : SqlColumnEncryptionKe
...
-            Debug.Assert(!string.IsNullOrEmpty(keyPath));
-
-            // Assign default values for omitted fields
-            StoreLocation storeLocation = StoreLocation.LocalMachine; // Default to Local Machine
-            StoreName storeName = StoreName.My;
-            string[] certParts = keyPath.Split('/');
+            StoreLocation storeLocation;
+            StoreName storeName;
+            ReadOnlySpan<char> keyPathSpan = keyPath.AsSpan();
+            int firstSeparator = keyPathSpan.IndexOf('/');
```
paulmedynski, line 121: Is all of this new code to avoid string.Split() ?  For performance reasons?  It's worth a comment since Split() seems like an obvious choice.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlColumnEncryptionCertificateStoreProvider.cs
```diff
@@ -52,511 +54,272 @@ public class SqlColumnEncryptionCertificateStoreProvider : SqlColumnEncryptionKe
...
+            // Get thumbprint
+            if (thumbprintSpan.IsEmpty)
             {
                 // An empty thumbprint specified
                 throw SQL.EmptyCertificateThumbprint(keyPath, isSystemOp);
             }
-
-            // Find the certificate and return
-            return GetCertificate(storeLocation, storeName, keyPath, thumbprint, isSystemOp);
+            else
```
paulmedynski: Unnecessary `else` here.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlColumnEncryptionCertificateStoreProvider.cs
```diff
@@ -52,511 +54,272 @@ public class SqlColumnEncryptionCertificateStoreProvider : SqlColumnEncryptionKe
...
-                return certificate;
+                throw SQL.CertificateNotFound(thumbprint.ToString(), storeName.ToString(), storeLocation.ToString(), isSystemOp);
             }
-            finally
+
+            using X509Certificate2 certificate = matchingCertificates[0];
+            if (!certificate.HasPrivateKey)
             {
-                // Close the certificate store
-                certificateStore?.Close();
```
paulmedynski, line 473: Do we still need to Close()?  The X509Store docs don't say whether or not Dispose() includes whatever Close() does.

The source code clearly does though:

https://github.com/dotnet/runtime/blob/1d1bf92fcf43aa6981804dc53c5174445069c9e4/src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/X509Certificates/X509Store.cs#L203

So you're good, despite the docs :(

paulmedynski (changes requested):

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/CoreCryptoTests.cs
edwardneal: Yes. There were a few other existing instances in the same test, I've changed those too.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlColumnEncryptionCertificateStoreProvider.cs
edwardneal: That's correct - string.Split allocates a new array and up to three strings inside it. I've added a comment.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlColumnEncryptionCertificateStoreProvider.cs
edwardneal: Yes - backed up by the netfx [reference source](https://github.com/microsoft/referencesource/blob/main/System/security/system/security/cryptography/x509/x509store.cs#L176).

It looks like this was new in net46 though, which might explain the old usage.

paulmedynski (approved):
Looks great!

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/CoreCryptoTests.cs
```diff
@@ -49,6 +50,8 @@ public void TestAeadCryptoWithNativeBaseline()
         [ConditionalFact(typeof(DataTestUtility), nameof(DataTestUtility.AreConnStringsSetup), nameof(DataTestUtility.IsNotAzureSynapse))]
         public void TestRsaCryptoWithNativeBaseline()
```
mdaigle, line 49: I just refactored these tests slightly on main to remove the class fixture and to make them run unconditionally. Can you pull the changes?
Now that you've removed the reflection (thank you!) can we also move these tests to the unit tests project?

edwardneal: I've just merged main and pushed. I also started moving the tests, but they drag quite a few dependencies. Do you mind if I submit a separate PR which removes the reflection against `SqlAeadAes256CbcHmac256Factory` and converts them to use `TestData` rather than manually parsing the two `TCECryptoNativeBaseline*.txt` files?

mdaigle: Yes, sounds good!

mdaigle (approved):

PR #3663, opened 2025-10-06
Cleanup disposable semaphore holder

cheenamalhotra:
Managing Semaphores with Disposable pattern may not be a good idea for .NET Framework applications. Reverting this change for now, to prevent any side effects.

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/WaitHandleDbConnectionPool.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/WaitHandleDbConnectionPool.cs
@@ -383,41 +383,6 @@ namespace Microsoft.Data.SqlClient.ConnectionPool
             }
         }
 
-        /// <summary>
-        /// Helper class to obtain and release a semaphore.
-        /// </summary>
-        internal class SemaphoreHolder : IDisposable
-        {
-            private readonly Semaphore _semaphore;
-
-            /// <summary>
-            /// Whether the semaphore was successfully obtained within the timeout.
-            /// </summary>
-            internal bool Obtained { get; private set; }
-
-            /// <summary>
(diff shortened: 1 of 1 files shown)


paulmedynski (approved):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/WaitHandleDbConnectionPool.cs
```diff
@@ -1316,18 +1281,32 @@ private bool TryGetConnection(DbConnection owningObject, uint waitForMultipleObj
...
                                     if (onlyOneCheckConnection)
                                     {
-                                        using SemaphoreHolder semaphoreHolder = new(_waitHandles.CreationSemaphore, unchecked((int)waitForMultipleObjectsTimeout));
-                                        if (semaphoreHolder.Obtained)
+                                        bool obtained = false;
+                                        try
                                         {
-                                            SqlClientEventSource.Log.TryPoolerTraceEvent("<prov.DbConnectionPool.GetConnection|RES|CPOOL> {0}, Creating new connection.", Id);
-                                            obj = UserCreateRequest(owningObject, userOptions);
+                                            obtained = _waitHandles.CreationSemaphore.WaitOne(unchecked((int)waitForMultipleObjectsTimeout));
```
paulmedynski, line 1287: Does this approach work if an async abort occurs after WaitOne() modifies the semaphore, but before `obtained` is assigned `true`?  I think all bets are off when the thread can be killed at any time.

cheenamalhotra: That's true, but since this is also the existing behavior I would expect the risk to be low with current design.

cheenamalhotra:
Since we no longer have CERs implemented, this change is a no-op in main branch.

PR #3673, opened 2025-10-08
Introduce Async API counterparts for AE base class

### LINKED ISSUE #3672: Async friendly Always Encrypted support
Always Encrypted types today don't support async APIs in base class as well as in implementations, leading to sync-over-async in Async workflows. This gap is important to fill for a true async workflow experience when executing commands where encryption/decryption is needed to be performed.

This will be covered in following tasks:
- [ ] #3673
- [ ] Implement async counterparts in all Key Store provider implementations:
- [ ] SqlColumnEncryptionCertificateStoreProvider
- [ ] SqlColumnEncryptionCngProvider
- [ ] SqlColumnEncryptionCspProvider
- [ ] SqlColumnEncryptionAzureKeyVaultProvider
- [ ] Define virtual async counterparts in base type 'SqlColumnEncryptionEnclaveProvider'
- [ ] Implement async counterparts in all Enclave provider implementations:
- [ ] NoneAttestationEnclaveProvider
- [ ] AzureAttestationEnclaveProvider
- [ ] VirtualizationBasedSecurityEnclaveProviderBase
- [ ] Implement Async API usage in Async code paths
- [ ] Testing and code coverage

cheenamalhotra:
## Summary

Introduces async API counterparts for the `SqlColumnEncryptionKeyStoreProvider` abstract base class (Always Encrypted). This enables custom key store providers to implement truly asynchronous key operations instead of being forced to block on synchronous calls during query execution.

Fixes #3672

## Public API

Four new `virtual` methods added to `SqlColumnEncryptionKeyStoreProvider`:

```csharp
public virtual Task<byte[]> DecryptColumnEncryptionKeyAsync(
    string masterKeyPath, string encryptionAlgorithm,
    byte[] encryptedColumnEncryptionKey,
    CancellationToken cancellationToken = default);

public virtual Task<byte[]> EncryptColumnEncryptionKeyAsync(
    string masterKeyPath, string encryptionAlgorithm,
    byte[] columnEncryptionKey,
    CancellationToken cancellationToken = default);

public virtual Task<byte[]> SignColumnMasterKeyMetadataAsync(
    string masterKeyPath, bool allowEnclaveComputations,
    CancellationToken cancellationToken = default);

public virtual Task<bool> VerifyColumnMasterKeyMetadataAsync(
    string masterKeyPath, bool allowEnclaveComputations,
    byte[] signature,
    CancellationToken cancellationToken = default);
```

### Design Decisions

- **Single overload with `CancellationToken = default`** — no separate convenience overloads
- **Default implementation wraps sync counterpart** via `Task.FromResult` / `Task.FromException` so existing providers work without changes
- **Fatal exceptions are not caught** — uses `ADP.IsCatchableExceptionType(e)` filter in catch blocks (consistent with runtime patterns for `StackOverflowException`, `OutOfMemoryException`, etc.)

## Changes

| File | Description |
|------|-------------|
| `src/.../SqlColumnEncryptionKeyStoreProvider.cs` | Implementation of 4 virtual async methods |
| `src/.../ref/Microsoft.Data.SqlClient.cs` | Reference assembly surface |
| `doc/snippets/.../SqlColumnEncryptionKeyStoreProvider.xml` | XML documentation |
| `specs/002-async-always-encrypted/spec.md` | Design specification |
| `tests/.../SqlColumnEncryptionKeyStoreProviderAsyncShould.cs` | 16 unit tests |

## Testing

- [x] 16 unit tests covering: default sync fallback, faulted task propagation, cancellation support, default parameter behavior
- [x] Builds clean on all target frameworks (net462, net8.0, net9.0)
- [x] No breaking changes — existing providers continue to work without modification

## Checklist

- [x] Tests added
- [x] Public API changes documented (ref assembly + XML docs)
- [x] Design spec included
- [x] No breaking changes introduced

### DIFF
--- a/doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionKeyStoreProvider.xml
+++ b/doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionKeyStoreProvider.xml
@@ -24,7 +24,7 @@
         Decrypts the specified encrypted value of a column encryption key. The encrypted value is expected to be encrypted using the column master key with the specified key path and using the specified algorithm.
       </summary>
       <returns>
-        Returns <see cref="T:System.Byte" /> array representing the decrypted column encryption key.
+        Returns <see cref="T:System.Byte[]" /> array representing the decrypted column encryption key.
       </returns>
     </DecryptColumnEncryptionKey>
     <DecryptColumnEncryptionKeyAsync>
@@ -41,7 +41,7 @@
         Decrypts the specified encrypted value of a column encryption key asynchronously. The encrypted value is expected to be encrypted using the column master key with the specified key path and using the specified algorithm.
       </summary>
       <returns>
-        Returns a task that returns <see cref="T:System.Byte" /> array representing the decrypted column encryption key on completion.
+        Returns a task that returns <see cref="T:System.Byte[]" /> array representing the decrypted column encryption key on completion.
       </returns>
     </DecryptColumnEncryptionKeyAsync>
     <EncryptColumnEncryptionKey>
@@ -58,7 +58,7 @@
         Encrypts a column encryption key using the column master key with the specified key path and using the specified algorithm.
       </summary>
       <returns>
-        Returns <see cref="T:System.Byte" /> array representing the encrypted column encryption key.
+        Returns <see cref="T:System.Byte[]" /> array representing the encrypted column encryption key.
       </returns>
     </EncryptColumnEncryptionKey>
     <EncryptColumnEncryptionKeyAsync>
@@ -75,7 +75,7 @@
         Encrypts a column encryption key asynchronously using the column master key with the specified key path and using the specified algorithm.
       </summary>
       <returns>
-        Returns a task that returns <see cref="T:System.Byte" /> array representing the encrypted column encryption key on completion.
+        Returns a task that returns <see cref="T:System.Byte[]" /> array representing the encrypted column encryption key on completion.
       </returns>
     </EncryptColumnEncryptionKeyAsync>
     <SignColumnMasterKeyMetadata>
--- a/doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionKeyStoreProvider.xml
+++ b/doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionKeyStoreProvider.xml
@@ -24,9 +24,26 @@
         Decrypts the specified encrypted value of a column encryption key. The encrypted value is expected to be encrypted using the column master key with the specified key path and using the specified algorithm.
       </summary>
       <returns>
-        Returns <see cref="T:System.Byte" />. The decrypted column encryption key.
+        Returns <see cref="T:System.Byte" /> array representing the decrypted column encryption key.
       </returns>
     </DecryptColumnEncryptionKey>
+    <DecryptColumnEncryptionKeyAsync>
+      <param name="masterKeyPath">
+        The master key path.
+      </param>
+      <param name="encryptionAlgorithm">
+        The encryption algorithm.
+      </param>
+      <param name="encryptedColumnEncryptionKey">
+        The encrypted column encryption key.
+      </param>
+      <summary>
+        Decrypts the specified encrypted value of a column encryption key asynchronously. The encrypted value is expected to be encrypted using the column master key with the specified key path and using the specified algorithm.
+      </summary>
+      <returns>
+        Returns a task that returns <see cref="T:System.Byte" /> array representing the decrypted column encryption key on completion.
+      </returns>
+    </DecryptColumnEncryptionKeyAsync>
     <EncryptColumnEncryptionKey>
       <param name="masterKeyPath">
         The master key path.
@@ -41,9 +58,26 @@
         Encrypts a column encryption key using the column master key with the specified key path and using the specified algorithm.
       </summary>
       <returns>
-        Returns <see cref="T:System.Byte" />. The encrypted column encryption key.
+        Returns <see cref="T:System.Byte" /> array representing the encrypted column encryption key.
       </returns>
     </EncryptColumnEncryptionKey>
+    <EncryptColumnEncryptionKeyAsync>
+      <param name="masterKeyPath">
+        The master key path.
+      </param>
+      <param name="encryptionAlgorithm">
+        The encryption algorithm.
+      </param>
+      <param name="columnEncryptionKey">
+        The plaintext column encryption key.
+      </param>
+      <summary>
+        Encrypts a column encryption key asynchronously using the column master key with the specified key path and using the specified algorithm.
+      </summary>
+      <returns>
+        Returns a task that returns <see cref="T:System.Byte" /> array representing the encrypted column encryption key on completion.
+      </returns>
+    </EncryptColumnEncryptionKeyAsync>
     <SignColumnMasterKeyMetadata>
       <param name="masterKeyPath">
         The column master key path.
@@ -55,7 +89,7 @@
         When implemented in a derived class, digitally signs the column master key metadata with the column master key referenced by the <paramref name="masterKeyPath" /> parameter. The input values used to generate the signature should be the specified values of the <paramref name="masterKeyPath" /> and <paramref name="allowEnclaveComputations" /> parameters.
       </summary>
       <returns>
-        The signature of the column master key metadata.
+        Returns the signature of the column master key metadata.
       </returns>
       <remarks>
         <para>
@@ -69,6 +103,31 @@
         In all cases.
       </exception>
     </SignColumnMasterKeyMetadata>
+    <SignColumnMasterKeyMetadataAsync>
+      <param name="masterKeyPath">
+        The column master key path.
+      </param>
+      <param name="allowEnclaveComputations">
+        <see langword="true" /> to indicate that the column master key supports enclave computations; otherwise, <see langword="false" />.
+      </param>
+      <summary>
+        When implemented in a derived class, asynchronously digitally signs the column master key metadata with the column master key referenced by the <paramref name="masterKeyPath" /> parameter. The input values used to generate the signature should be the specified values of the <paramref name="masterKeyPath" /> and <paramref name="allowEnclaveComputations" /> parameters.
+      </summary>
+      <returns>
+        Returns a task that returns the signature of the column master key metadata on completion.
+      </returns>
+      <remarks>
+        <para>
+          To ensure that the <see cref="M:Microsoft.Data.SqlClient.SqlColumnEncryptionKeyStoreProvider.SignColumnMasterKeyMetadata(System.String,System.Boolean)" /> method doesn't break applications that rely on an old API, it throws a <see cref="T:System.NotImplementedException" /> exception by default.
+        </para>
+        <para>
+          The <see cref="M:Microsoft.Data.SqlClient.SqlColumnEncryptionKeyStoreProvider.SignColumnMasterKeyMetadata(System.String,System.Boolean)" /> method will be used by client tools that generate Column Master Keys (CMK) for customers. <see cref="M:Microsoft.Data.SqlClient.SqlColumnEncryptionKeyStoreProvider.SignColumnMasterKeyMetadata(System.String,System.Boolean)" /> must be implemented by the corresponding key store providers that wish to use enclaves with <see href="https://learn.microsoft.com/sql/relational-databases/security/encryption/always-encrypted-database-engine">Always Encrypted</see>.
+        </para>
+      </remarks>
+      <exception cref="T:System.NotImplementedException">
+        In all cases.
+      </exception>
+    </SignColumnMasterKeyMetadataAsync>
     <VerifyColumnMasterKeyMetadata>
       <param name="masterKeyPath">
         The column master key path.
@@ -86,6 +145,23 @@
         When implemented in a derived class, the method is expected to return true if the specified signature is valid, or false if the specified signature is not valid. The default implementation throws `NotImplementedException`.
       </returns>
     </VerifyColumnMasterKeyMetadata>
+    <VerifyColumnMasterKeyMetadataAsync>
+      <param name="masterKeyPath">
+        The column master key path.
+      </param>
+      <param name="allowEnclaveComputations">
+        Indicates whether the column master key supports enclave computations.
+      </param>
+      <param name="signature">
+        The signature of the column master key metadata.
+      </param>
+      <summary>
+        When implemented in a derived class, this method is expected to verify the specified signature is valid for the column master key with the specified key path and the specified enclave behavior asynchronously. The default implementation throws `NotImplementedException`.
+      </summary>
+      <returns>
+        When implemented in a derived class, the method is expected to return true if the specified signature is valid, or false if the specified signature is not valid. The default implementation throws `NotImplementedException`.
+      </returns>
+    </VerifyColumnMasterKeyMetadataAsync>
     <ColumnEncryptionKeyCacheTtl>
       <summary>
         Gets or sets the lifespan of the decrypted column encryption key in the cache. Once the timespan has elapsed, the decrypted column encryption key is discarded and must be revalidated.
--- a/doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionKeyStoreProvider.xml
+++ b/doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionKeyStoreProvider.xml
@@ -30,7 +30,8 @@
         The encrypted column encryption key.
       </param>
       <returns>
-        Returns the decrypted column encryption key as a byte array.
+        Returns a <see cref="T:System.Byte[]" /> representing the decrypted
+        column encryption key.
       </returns>
     </DecryptColumnEncryptionKey>
     <DecryptColumnEncryptionKeyAsync>
@@ -56,26 +57,51 @@
         A token to cancel the asynchronous operation.
       </param>
       <returns>
-        A task that, when completed, returns the decrypted column encryption
-        key as a byte array.
+        A task that returns a <see cref="T:System.Byte[]" /> representing
+        the decrypted column encryption key on completion.
       </returns>
       <remarks>
         <para>
-          The default implementation calls the synchronous
+          The default implementation first checks the
+          <paramref name="cancellationToken" /> and returns a canceled task
+          if cancellation has been requested. Otherwise, it calls the
+          synchronous
           <see cref="M:Microsoft.Data.SqlClient.SqlColumnEncryptionKeyStoreProvider.DecryptColumnEncryptionKey(System.String,System.String,System.Byte[])" />
           method and wraps the result in a completed task. If the synchronous
           method throws, the exception is caught and a faulted task is
           returned rather than throwing synchronously.
         </para>
         <para>
-          The <paramref name="cancellationToken" /> parameter is not observed
-          by the default implementation. Derived classes that perform I/O
-          (such as calls to Azure Key Vault) should override this method with
-          a truly asynchronous implementation that honors the cancellation
-          token.
+          Derived classes that perform I/O (such as calls to Azure Key Vault)
+          should override this method with a truly asynchronous
+          implementation that honors the cancellation token.
         </para>
       </remarks>
     </DecryptColumnEncryptionKeyAsync>
+    <DecryptColumnEncryptionKeyAsyncNoCancellation>
+      <summary>
+        Asynchronously decrypts the specified encrypted value of a column
+        encryption key. The encrypted value is expected to be encrypted using
+        the column master key with the specified key path and using the
+        specified algorithm.
+      </summary>
+      <param name="masterKeyPath">
+        The column master key path. The path format is specific to the key
+        store provider implementation (e.g. a thumbprint for the certificate
+        store, or a key identifier URL for Azure Key Vault).
+      </param>
+      <param name="encryptionAlgorithm">
+        The encryption algorithm name. For example,
+        <c>RSA_OAEP</c>.
+      </param>
+      <param name="encryptedColumnEncryptionKey">
+        The encrypted column encryption key.
+      </param>
+      <returns>
+        A task that returns a <see cref="T:System.Byte[]" /> representing
+        the decrypted column encryption key on completion.
+      </returns>
+    </DecryptColumnEncryptionKeyAsyncNoCancellation>
     <EncryptColumnEncryptionKey>
       <summary>
         Encrypts a column encryption key using the column master key with
@@ -94,7 +120,8 @@
         The column encryption key to encrypt.
       </param>
       <returns>
-        Returns the encrypted column encryption key as a byte array.
+        Returns a <see cref="T:System.Byte[]" /> representing the encrypted
+        column encryption key.
       </returns>
     </EncryptColumnEncryptionKey>
     <EncryptColumnEncryptionKeyAsync>
@@ -119,26 +146,50 @@
         A token to cancel the asynchronous operation.
       </param>
       <returns>
-        A task that, when completed, returns the encrypted column encryption
-        key as a byte array.
+        A task that returns a <see cref="T:System.Byte[]" /> representing
+        the encrypted column encryption key on completion.
       </returns>
       <remarks>
         <para>
-          The default implementation calls the synchronous
+          The default implementation first checks the
+          <paramref name="cancellationToken" /> and returns a canceled task
+          if cancellation has been requested. Otherwise, it calls the
+          synchronous
           <see cref="M:Microsoft.Data.SqlClient.SqlColumnEncryptionKeyStoreProvider.EncryptColumnEncryptionKey(System.String,System.String,System.Byte[])" />
           method and wraps the result in a completed task. If the synchronous
           method throws, the exception is caught and a faulted task is
           returned rather than throwing synchronously.
         </para>
         <para>
-          The <paramref name="cancellationToken" /> parameter is not observed
-          by the default implementation. Derived classes that perform I/O
-          (such as calls to Azure Key Vault) should override this method with
-          a truly asynchronous implementation that honors the cancellation
-          token.
+          Derived classes that perform I/O (such as calls to Azure Key Vault)
+          should override this method with a truly asynchronous
+          implementation that honors the cancellation token.
         </para>
       </remarks>
     </EncryptColumnEncryptionKeyAsync>
+    <EncryptColumnEncryptionKeyAsyncNoCancellation>
+      <summary>
+        Asynchronously encrypts a column encryption key using the column
+        master key with the specified key path and using the specified
+        algorithm.
+      </summary>
+      <param name="masterKeyPath">
+        The column master key path. The path format is specific to the key
+        store provider implementation (e.g. a thumbprint for the certificate
+        store, or a key identifier URL for Azure Key Vault).
+      </param>
+      <param name="encryptionAlgorithm">
+        The encryption algorithm name. For example,
+        <c>RSA_OAEP</c>.
+      </param>
+      <param name="columnEncryptionKey">
+        The column encryption key to encrypt.
+      </param>
+      <returns>
+        A task that returns a <see cref="T:System.Byte[]" /> representing
+        the encrypted column encryption key on completion.
+      </returns>
+    </EncryptColumnEncryptionKeyAsyncNoCancellation>
     <SignColumnMasterKeyMetadata>
       <summary>
         When implemented in a derived class, signs the column master key
@@ -190,12 +241,15 @@
         A token to cancel the asynchronous operation.
       </param>
       <returns>
-        A task that, when completed, returns the signature of the column
-        master key metadata as a byte array.
+        A task that, when completed, returns a <see cref="T:System.Byte[]" />
+        representing the signature of the column master key metadata.
       </returns>
       <remarks>
         <para>
-          The default implementation calls the synchronous
+          The default implementation first checks the
+          <paramref name="cancellationToken" /> and returns a canceled task
+          if cancellation has been requested. Otherwise, it calls the
+          synchronous
           <see cref="M:Microsoft.Data.SqlClient.SqlColumnEncryptionKeyStoreProvider.SignColumnMasterKeyMetadata(System.String,System.Boolean)" />
           method, which throws a
           <see cref="T:System.NotImplementedException" /> by default. In
@@ -204,10 +258,9 @@
           throwing synchronously.
         </para>
         <para>
-          The <paramref name="cancellationToken" /> parameter is not observed
-          by the default implementation. Derived classes that perform I/O
-          should override this method with a truly asynchronous
-          implementation that honors the cancellation token.
+          Derived classes that perform I/O should override this method with
+          a truly asynchronous implementation that honors the cancellation
+          token.
         </para>
         <para>
           Key store providers that wish to use enclaves with
@@ -217,6 +270,27 @@
         </para>
       </remarks>
     </SignColumnMasterKeyMetadataAsync>
+    <SignColumnMasterKeyMetadataAsyncNoCancellation>
+      <summary>
+        When implemented in a derived class, asynchronously signs the column
+        master key metadata with the column master key referenced by the
+        <paramref name="masterKeyPath" /> parameter.
+      </summary>
+      <param name="masterKeyPath">
+        The column master key path. The path format is specific to the key
+        store provider implementation (e.g. a thumbprint for the certificate
+        store, or a key identifier URL for Azure Key Vault).
+      </param>
+      <param name="allowEnclaveComputations">
+        <see langword="true" /> to indicate that the column master key
+        supports enclave computations; otherwise,
+        <see langword="false" />.
+      </param>
+      <returns>
+        A task that, when completed, returns a <see cref="T:System.Byte[]" />
+        representing the signature of the column master key metadata.
+      </returns>
+    </SignColumnMasterKeyMetadataAsyncNoCancellation>
     <VerifyColumnMasterKeyMetadata>
       <summary>
         When implemented in a derived class, this method is expected to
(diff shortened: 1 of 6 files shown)


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlColumnEncryptionKeyStoreProvider.cs
```diff
@@ -16,19 +17,35 @@ public abstract class SqlColumnEncryptionKeyStoreProvider
         /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionKeyStoreProvider.xml' path='docs/members[@name="SqlColumnEncryptionKeyStoreProvider"]/DecryptColumnEncryptionKey/*'/>
         public abstract byte[] DecryptColumnEncryptionKey(string masterKeyPath, string encryptionAlgorithm, byte[] encryptedColumnEncryptionKey);
 
+        /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionKeyStoreProvider.xml' path='docs/members[@name="SqlColumnEncryptionKeyStoreProvider"]/DecryptColumnEncryptionKeyAsync/*'/>
+        public virtual Task<byte[]> DecryptColumnEncryptionKeyAsync(string masterKeyPath, string encryptionAlgorithm, byte[] encryptedColumnEncryptionKey)
```
edwardneal: Cryptographic tasks will normally complete synchronously; Azure Key Vault is the exception here. What about returning `ValueTask<byte[]>`, to remove an allocation in the synchronous case?

paulmedynski: Since these 4 new async methods are all implemented by calling their synchronous counterparts, I would agree that the operations will definitely complete synchronously.

Do we plan to re-implement these in terms of some other underlying async methods later?

paulmedynski (commented):
Asking for clarification on return types and synchronous implementation.

benrr101 (dismissed):
For new APIs we should have some discussion on the signatures. Definitely won't be able to get this in before preview2.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlColumnEncryptionKeyStoreProvider.cs
cheenamalhotra: Yes, these implementations will be re-implemented in all Key Vault providers, as issue #3672 covers. Both sync and async will be made truly sync and truly async flows - with external calls separated.

Also, in the codebase, we want to make sure we call sync v/s async APIs of AE that are separated based on caller.

This, here, is a fallback for any consumer of AE implementation without async implementation.

paulmedynski (dismissed):

cheenamalhotra:
Wait, wait 🙈

cheenamalhotra:
Stay

cheenamalhotra:
Getting back on it soon..

doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionKeyStoreProvider.xml
```diff
@@ -24,9 +24,29 @@
...
+      <param name="cancellationToken">
+        A token to cancel the asynchronous operation.
+      </param>
+      <summary>
+        Decrypts the specified encrypted value of a column encryption key asynchronously. The encrypted value is expected to be encrypted using the column master key with the specified key path and using the specified algorithm.
+      </summary>
+      <returns>
+        Returns a task that returns <see cref="T:System.Byte[]" /> array representing the decrypted column encryption key on completion.
+      </returns>
+    </DecryptColumnEncryptionKeyAsync>
```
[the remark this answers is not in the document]
cheenamalhotra: Addressed — added `<remarks>` sections to `DecryptColumnEncryptionKeyAsync`, `EncryptColumnEncryptionKeyAsync`, `SignColumnMasterKeyMetadataAsync`, and `VerifyColumnMasterKeyMetadataAsync` XML docs that now describe:
1. The default implementation delegates to the sync counterpart via `Task.FromResult`
2. If the sync method throws, a faulted task is returned (not a synchronous throw)
3. The `CancellationToken` parameter is not observed by the default implementation; derived classes performing I/O should override with a truly async implementation that honors the token

src/Microsoft.Data.SqlClient/tests/FunctionalTests/AlwaysEncryptedTests/SqlColumnEncryptionKeyStoreProviderAsyncShould.cs
```diff
@@ -0,0 +1,127 @@
...
+        [Fact]
+        public async Task SignColumnMasterKeyMetadataAsync_DefaultImplementation_ReturnsFaultedTask()
+        {
+            var provider = new TestKeyStoreProvider();
+            Task<byte[]> task = provider.SignColumnMasterKeyMetadataAsync("path", true);
+
+            // The base sync method throws NotImplementedException; the async version
+            // must surface this as a faulted Task, not a synchronous throw.
+            Assert.True(task.IsFaulted);
+            await Assert.ThrowsAsync<NotImplementedException>(() => task);
```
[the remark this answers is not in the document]
cheenamalhotra: The `IsFaulted` assertion is intentional — it validates the specific contract documented in FR-003: the default implementation MUST return a faulted Task (not throw synchronously). This is a distinct guarantee from "the exception is observable when awaited" (which `Assert.ThrowsAsync` validates). Both assertions together prove the full contract:

1. `task.IsFaulted == true` → no synchronous throw, exception is on the Task
2. `Assert.ThrowsAsync` → the exception is the expected type

If the implementation were later changed to `async/await`-based, both assertions would still pass since `async` methods always surface exceptions on the returned Task. Keeping `IsFaulted` makes the test fail fast and clearly if someone accidentally removes the `try/catch` wrapping.

doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionKeyStoreProvider.xml
```diff
@@ -24,9 +24,37 @@
         Decrypts the specified encrypted value of a column encryption key. The encrypted value is expected to be encrypted using the column master key with the specified key path and using the specified algorithm.
       </summary>
       <returns>
-        Returns <see cref="T:System.Byte" />. The decrypted column encryption key.
+        Returns <see cref="T:System.Byte[]" /> array representing the decrypted column encryption key.
       </returns>
     </DecryptColumnEncryptionKey>
+    <DecryptColumnEncryptionKeyAsync>
+      <param name="masterKeyPath">
+        The master key path.
```
paulmedynski: What kind of path is this expected to be?  Local filesystem, UNC, URL ... ?

For example:
```
The local filesystem path to the master key file.  This will be used by <.NET API>; see [here] for further information.
```

doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionKeyStoreProvider.xml
```diff
@@ -24,9 +24,37 @@
...
+      </param>
+      <param name="encryptionAlgorithm">
+        The encryption algorithm.
```
paulmedynski: What kind of values are expected/accepted here?  Is it an opaque string to us?  Examples and where they are documented would help.

src/Microsoft.Data.SqlClient/ref/Microsoft.Data.SqlClient.cs
```diff
@@ -704,12 +704,20 @@ public abstract class SqlColumnEncryptionKeyStoreProvider
     protected SqlColumnEncryptionKeyStoreProvider() { }
     /// <include file='../../../doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionKeyStoreProvider.xml' path='docs/members[@name="SqlColumnEncryptionKeyStoreProvider"]/DecryptColumnEncryptionKey/*'/>
     public abstract byte[] DecryptColumnEncryptionKey(string masterKeyPath, string encryptionAlgorithm, byte[] encryptedColumnEncryptionKey);
+    /// <include file='../../../doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionKeyStoreProvider.xml' path='docs/members[@name="SqlColumnEncryptionKeyStoreProvider"]/DecryptColumnEncryptionKeyAsync/*'/>
```
paulmedynski, line 707: Let's use nullable annotation for new APIs, and then for any nullable arguments, we must document what null means.

doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionKeyStoreProvider.xml
```diff
@@ -24,9 +24,37 @@
...
+      </param>
+      <param name="encryptedColumnEncryptionKey">
+        The encrypted column encryption key.
+      </param>
+      <param name="cancellationToken">
+        A token to cancel the asynchronous operation.
+      </param>
+      <summary>
```
paulmedynski: Seems weird to put the summary after the arguments.

doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionKeyStoreProvider.xml
```diff
@@ -24,9 +24,37 @@
...
+        Decrypts the specified encrypted value of a column encryption key asynchronously. The encrypted value is expected to be encrypted using the column master key with the specified key path and using the specified algorithm.
```
paulmedynski: IMO we should wrap long lines to our standard 100 chars.

doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionKeyStoreProvider.xml
```diff
@@ -41,9 +69,37 @@
         Encrypts a column encryption key using the column master key with the specified key path and using the specified algorithm.
       </summary>
       <returns>
-        Returns <see cref="T:System.Byte" />. The encrypted column encryption key.
+        Returns <see cref="T:System.Byte[]" /> array representing the encrypted column encryption key.
       </returns>
     </EncryptColumnEncryptionKey>
+    <EncryptColumnEncryptionKeyAsync>
```
paulmedynski, line 103: Same comments regarding arguments here as with Decrypt above.

doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionKeyStoreProvider.xml
```diff
@@ -41,9 +69,37 @@
...
+      <param name="masterKeyPath">
+        The master key path.
+      </param>
+      <param name="encryptionAlgorithm">
+        The encryption algorithm.
+      </param>
+      <param name="columnEncryptionKey">
+        The plaintext column encryption key.
```
paulmedynski: It's a byte array, so it isn't text, per se.  I would suggest: `The column encryption key to encrypt`

doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionKeyStoreProvider.xml
```diff
@@ -69,6 +125,34 @@
         In all cases.
       </exception>
     </SignColumnMasterKeyMetadata>
+    <SignColumnMasterKeyMetadataAsync>
+      <param name="masterKeyPath">
+        The column master key path.
```
paulmedynski: Same comment regarding paths.

doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionKeyStoreProvider.xml
```diff
@@ -69,6 +125,34 @@
...
+      </param>
+      <param name="allowEnclaveComputations">
+        <see langword="true" /> to indicate that the column master key supports enclave computations; otherwise, <see langword="false" />.
+      </param>
+      <param name="cancellationToken">
+        A token to cancel the asynchronous operation.
+      </param>
+      <summary>
+        When implemented in a derived class, asynchronously digitally signs the column master key metadata with the column master key referenced by the <paramref name="masterKeyPath" /> parameter. The input values used to generate the signature should be the specified values of the <paramref name="masterKeyPath" /> and <paramref name="allowEnclaveComputations" /> parameters.
```
paulmedynski: `digitally` isn't necessary.

I'm not sure what the second sentence is trying to say.  It sounds like "We will use the method arguments to do the work", which is obvious.

doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionKeyStoreProvider.xml
```diff
@@ -69,6 +125,34 @@
         In all cases.
       </exception>
     </SignColumnMasterKeyMetadata>
+    <SignColumnMasterKeyMetadataAsync>
+      <param name="masterKeyPath">
+        The column master key path.
+      </param>
+      <param name="allowEnclaveComputations">
+        <see langword="true" /> to indicate that the column master key supports enclave computations; otherwise, <see langword="false" />.
```
paulmedynski: But what actually happens when I specify true vs false?

doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionKeyStoreProvider.xml
```diff
@@ -86,6 +170,34 @@
         When implemented in a derived class, the method is expected to return true if the specified signature is valid, or false if the specified signature is not valid. The default implementation throws `NotImplementedException`.
       </returns>
     </VerifyColumnMasterKeyMetadata>
+    <VerifyColumnMasterKeyMetadataAsync>
+      <param name="masterKeyPath">
+        The column master key path.
```
paulmedynski: Same path and enclave comments.

doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionKeyStoreProvider.xml
```diff
@@ -86,6 +170,34 @@
...
+      </param>
+      <param name="allowEnclaveComputations">
+        Indicates whether the column master key supports enclave computations.
+      </param>
+      <param name="signature">
+        The signature of the column master key metadata.
```
paulmedynski: Is this expected to be a byte array returned by Sign() ?  Would this accept any format/encoding signature type?

doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionKeyStoreProvider.xml
```diff
@@ -69,6 +125,34 @@
...
+        <see langword="true" /> to indicate that the column master key supports enclave computations; otherwise, <see langword="false" />.
+      </param>
+      <param name="cancellationToken">
+        A token to cancel the asynchronous operation.
+      </param>
+      <summary>
+        When implemented in a derived class, asynchronously digitally signs the column master key metadata with the column master key referenced by the <paramref name="masterKeyPath" /> parameter. The input values used to generate the signature should be the specified values of the <paramref name="masterKeyPath" /> and <paramref name="allowEnclaveComputations" /> parameters.
+      </summary>
+      <returns>
+        Returns a task that returns the signature of the column master key metadata on completion.
```
paulmedynski: What is the format/encoding of the signature?  Is it HMAC, SHA256, etc?

src/Microsoft.Data.SqlClient/tests/FunctionalTests/AlwaysEncryptedTests/SqlColumnEncryptionKeyStoreProviderAsyncShould.cs
```diff
@@ -0,0 +1,127 @@
...
+        {
+            var provider = new TestKeyStoreProvider();
+            Task<bool> task = provider.VerifyColumnMasterKeyMetadataAsync("path", true, new byte[] { 1 });
+
+            Assert.True(task.IsFaulted);
+            await Assert.ThrowsAsync<NotImplementedException>(() => task);
+        }
+
+        [Fact]
+        public async Task DecryptColumnEncryptionKeyAsync_AcceptsCancellationToken()
```
paulmedynski: Should we also test that async encrypt and decrypt _ignore_ the cancellation token?
```
var task = EncryptAsync(..., token);
token.Cancel();
var result = await task;
Assert.NotNull(result);
```

paulmedynski (changes requested):

doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionKeyStoreProvider.xml
edwardneal: From SQL Server's perspective, `masterKeyPath` is an opaque provider-specific string. It's designed to uniquely identify the key material for a specific cryptographic key within the provider. For AKV, it'll be a URI; for certificates, it'll be `[Store Location]/[Store Name]/[Thumbprint]`. The rest of SqlClient doesn't know about its format.

doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionKeyStoreProvider.xml
edwardneal: It's a technically-opaque string returned by SQL Server - for the moment, it's always `RSA_OAEP`. This is documented [here](https://learn.microsoft.com/en-us/sql/t-sql/statements/create-column-encryption-key-transact-sql?view=sql-server-ver17), in the `algorithm_name` argument.

doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionKeyStoreProvider.xml
edwardneal: This is handled within SQL Server. It's only used during enclave-enabled Always Encrypted, and the flag indicates that the key can be passed into the secure enclave. From the view of this part of SqlClient, it's an opaque flag which is designed to be signed.

The idea here is that SQL Server's record of the column master key is little more than metadata. This method locates the key material identified by the key path, then signs the metadata that SQL Server holds. SQL Server keeps a record of that signature, and passes it to the client for validation. It forms a trust anchor for cases where SQL Server itself is compromised, but the enclave can't be tampered with: if someone modifies the CMK's key path (to try and force the client to encrypt data with an attacker-specified key) then this validation fails. The attacker can't just recalculate the correct signature because SQL Server doesn't have access to the key material - just its metadata.

doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionKeyStoreProvider.xml
edwardneal: Strictly speaking, it's opaque. For the builtin providers' purposes, it's an RSA signature of the SHA256 hash of `[Provider Name] + [Master Key Path] + [Enclave Computations Enabled]`. The only real API requirement that is that an untampered signature verifies successfully, and a tampered signature does not.

specs/002-async-always-encrypted/spec.md
```diff
@@ -0,0 +1,423 @@
...
+try { cache.Set(key, value); }
+finally { semaphore.Release(); }
+return value;
+```
+
+> **Important**: The first acquire MUST be wrapped in `try/finally` to guarantee semaphore release if `TryGetValue` or any cache-inspection code throws unexpectedly.
+
+Two threads may independently fetch the same key on a concurrent miss; the second write is idempotent (same decrypted value). This is acceptable and already implicitly assumed by the existing comment in `SqlSymmetricKeyCache`: "first one wins."
+
+### Decision 8: `IMemoryCache` Has No `GetOrCreateAsync` — Use Explicit `TryGetValue`/`Set`
```
edwardneal: While there's no method on `IMemoryCache`, there's an extension method in CacheExtensions ([link](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.caching.memory.cacheextensions.getorcreateasync?view=net-10.0-pp#microsoft-extensions-caching-memory-cacheextensions-getorcreateasync-1(microsoft-extensions-caching-memory-imemorycache-system-object-system-func((microsoft-extensions-caching-memory-icacheentry-system-threading-tasks-task((-0))))))). Would this be suitable?

specs/002-async-always-encrypted/spec.md
```diff
@@ -0,0 +1,423 @@
...
+
+**Rationale**: SqlClient is a library, not an application. If `ConfigureAwait(false)` is omitted, the `await` continuation is scheduled back onto the caller's `SynchronizationContext`. In app models that have a synchronization context (ASP.NET Framework, WPF, WinForms), any code that calls async methods with `.Result` or `.Wait()` (common in app code calling library APIs) will deadlock. This is a required invariant for all library async code in the .NET ecosystem. Applies to Phases 2, 3, 4, and 5.
+
+### Decision 7: Lock-During-I/O Must Use Check-Release-Fetch-Relock Pattern
+
+**Chosen**: In all async code paths, the `SemaphoreSlim` MUST be released before performing I/O (HTTP calls to AKV, Azure Attestation, HGS), then re-acquired to update the cache.
+
+**Rationale**: Three existing `SemaphoreSlim` sites hold their lock for the full duration of I/O:
+
+- `SqlSymmetricKeyCache._cacheLock` — static singleton, held during `DecryptColumnEncryptionKey`
```
edwardneal: This is centered on SqlSymmetricKeyCache but might be asked more broadly: do we need to retain the SemaphoreSlim instances at all?
MemoryCache is thread-safe, the documentation simply hasn't been updated yet (dotnet/AspNetCore.Docs#33103).
Post implementation, this pattern would cover:

* checks against trusted key paths, followed by the column encryption keystore provider lookup (not thread-safe - but this wouldn't change.)
* assignment of ColumnEncryptionKeyCacheTtl (if we say that the calls to DecryptColumnEncryptionKey/Async will no longer be serialized, we're already introducing a race condition if the provider sets that property. I'm not completely clear how serializing the assignment to the property would solve this.)
* wrapping of the plaintext key in a SqlClientSymmetricKey instance (thread-safe.)
* consecutive reads of static ColumnEncryptionKeyCacheTtl property (we could read this once and cache it.)
* writing the cache (thread-safe.)

At present, the only true effect of the SemaphoreSlim is to ensure that calls to the keystore provider are serialized, and that they only occur with ColumnEncryptionKeyCacheTtl set to zero. We're explicitly removing the former constraint, and I'm not convinced we'll be able to ensure the latter in all circumstances (unless we pass it as a parameter.)

specs/002-async-always-encrypted/spec.md
```diff
@@ -0,0 +1,423 @@
...
+
+**Unit Tests** (`src/Microsoft.Data.SqlClient/tests/UnitTests/`):
+
+- Default async fallback behavior
+- CancellationToken propagation
+- Faulted Task exception behavior
+
+**Static Analysis (CI gate)**:
+
+- Run a Roslyn analyzer or `grep`-based CI check to verify every `await` in `Microsoft.Data.SqlClient` and `Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider` assemblies uses `.ConfigureAwait(false)`. A single missed instance can cause production deadlocks in ASP.NET Framework / WPF / WinForms hosts. This MUST be enforced in CI, not left to manual review.
```
edwardneal: We can enforce this using code quality rule [CA2007](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2007).

This also impacts further consideration 1.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/AlwaysEncryptedTests/SqlColumnEncryptionKeyStoreProviderAsyncShould.cs
edwardneal: It's worth noting that the default implementation of [`DbConnection.GetSchemaAsync`](https://github.com/dotnet/dotnet/blob/b0f34d51fccc69fd334253924abd8d6853fad7aa/src/runtime/src/libraries/System.Data.Common/src/System/Data/Common/DbConnection.cs#L229-L244) will return a cancelled Task if it's passed a cancelled `CancellationToken`. This seems like a sensible approach to take.

doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionKeyStoreProvider.xml
```diff
@@ -41,9 +69,37 @@
         Encrypts a column encryption key using the column master key with the specified key path and using the specified algorithm.
       </summary>
       <returns>
-        Returns <see cref="T:System.Byte" />. The encrypted column encryption key.
+        Returns <see cref="T:System.Byte[]" /> array representing the encrypted column encryption key.
```
mdaigle:
```suggestion
        Returns a <see cref="T:System.Byte[]" /> representing the encrypted column encryption key.
```

doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionKeyStoreProvider.xml
```diff
@@ -24,9 +24,37 @@
         Decrypts the specified encrypted value of a column encryption key. The encrypted value is expected to be encrypted using the column master key with the specified key path and using the specified algorithm.
       </summary>
       <returns>
-        Returns <see cref="T:System.Byte" />. The decrypted column encryption key.
+        Returns <see cref="T:System.Byte[]" /> array representing the decrypted column encryption key.
```
mdaigle:
```suggestion
        Returns a <see cref="T:System.Byte[]" /> representing the decrypted column encryption key.
```

doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionKeyStoreProvider.xml
```diff
@@ -41,9 +69,37 @@
...
+        The plaintext column encryption key.
+      </param>
+      <param name="cancellationToken">
+        A token to cancel the asynchronous operation.
+      </param>
+      <summary>
+        Encrypts a column encryption key asynchronously using the column master key with the specified key path and using the specified algorithm.
+      </summary>
+      <returns>
+        Returns a task that returns <see cref="T:System.Byte[]" /> array representing the encrypted column encryption key on completion.
```
mdaigle:
```suggestion
        Returns a task that returns a <see cref="T:System.Byte[]" /> representing the encrypted column encryption key on completion.
```

doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionKeyStoreProvider.xml
```diff
@@ -24,9 +24,37 @@
...
+        The encrypted column encryption key.
+      </param>
+      <param name="cancellationToken">
+        A token to cancel the asynchronous operation.
+      </param>
+      <summary>
+        Decrypts the specified encrypted value of a column encryption key asynchronously. The encrypted value is expected to be encrypted using the column master key with the specified key path and using the specified algorithm.
+      </summary>
+      <returns>
+        Returns a task that returns <see cref="T:System.Byte[]" /> array representing the decrypted column encryption key on completion.
```
mdaigle:
```suggestion
        Returns a task that returns a <see cref="T:System.Byte[]" /> representing the decrypted column encryption key on completion.
```

specs/002-async-always-encrypted/spec.md
mdaigle: Agreed

src/Microsoft.Data.SqlClient/ref/Microsoft.Data.SqlClient.cs
mdaigle: The default for CancellationToken is CancellationToken.None. 

But I'd argue that all of these should follow the standard library pattern where async methods get multiple overloads. In our case, that would look like:
```c#
 public virtual System.Threading.Tasks.Task<byte[]> DecryptColumnEncryptionKeyAsync(string masterKeyPath, string encryptionAlgorithm, byte[] encryptedColumnEncryptionKey, System.Threading.CancellationToken cancellationToken)
```
and
```c#
 public System.Threading.Tasks.Task<byte[]> DecryptColumnEncryptionKeyAsync(string masterKeyPath, string encryptionAlgorithm, byte[] encryptedColumnEncryptionKey)
```

Where the second calls the first, passing CancellationToken.None.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/AlwaysEncryptedTests/SqlColumnEncryptionKeyStoreProviderAsyncShould.cs
mdaigle: We should respect cancellation tokens, especially ahead of a potentially time-consuming operation.

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft/Data/SqlClient/AlwaysEncrypted/SqlColumnEncryptionKeyStoreProviderAsyncShould.cs
```diff
@@ -0,0 +1,127 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
```
mdaigle, line 5: These can move to UnitTests.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlColumnEncryptionKeyStoreProvider.cs
```diff
@@ -16,19 +18,71 @@ public abstract class SqlColumnEncryptionKeyStoreProvider
         /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionKeyStoreProvider.xml' path='docs/members[@name="SqlColumnEncryptionKeyStoreProvider"]/DecryptColumnEncryptionKey/*'/>
         public abstract byte[] DecryptColumnEncryptionKey(string masterKeyPath, string encryptionAlgorithm, byte[] encryptedColumnEncryptionKey);
 
+        /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionKeyStoreProvider.xml' path='docs/members[@name="SqlColumnEncryptionKeyStoreProvider"]/DecryptColumnEncryptionKeyAsync/*'/>
+        public virtual Task<byte[]> DecryptColumnEncryptionKeyAsync(string masterKeyPath, string encryptionAlgorithm, byte[] encryptedColumnEncryptionKey, CancellationToken cancellationToken = default)
+        {
+            try
```
mdaigle, line 30: The cancellation token should be checked and a cancelled task returned if the token is cancelled.

mdaigle (changes requested):

doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionKeyStoreProvider.xml
benrr101: Agreed ... though good luck getting the 🤖 to actually cap it at 100 characters.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlColumnEncryptionKeyStoreProvider.cs
cheenamalhotra: Done

src/Microsoft.Data.SqlClient/ref/Microsoft.Data.SqlClient.cs
cheenamalhotra: I don't know which standard library pattern describes having multiple overloads in a base class?

What I'm concerned with multiple overloads is that it becomes difficult for implementers to identify which API to implement, and also uncertain to decide which API we call internally when performing async operations (encryption/decryption) while reading data.

Customers can implement either, and that causes confusions. There should be a single API for single purpose in a base class, and so I am reverting back to single virtual API.

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft/Data/SqlClient/AlwaysEncrypted/SqlColumnEncryptionKeyStoreProviderAsyncShould.cs
cheenamalhotra: Out of scope for this PR. We should do those in separate PR.

doc/snippets/Microsoft.Data.SqlClient/SqlColumnEncryptionKeyStoreProvider.xml
```diff
@@ -69,6 +103,31 @@
...
+        When implemented in a derived class, asynchronously digitally signs the column master key metadata with the column master key referenced by the <paramref name="masterKeyPath" /> parameter. The input values used to generate the signature should be the specified values of the <paramref name="masterKeyPath" /> and <paramref name="allowEnclaveComputations" /> parameters.
+      </summary>
+      <returns>
+        Returns a task that returns the signature of the column master key metadata on completion.
+      </returns>
+      <remarks>
+        <para>
+          To ensure that the <see cref="M:Microsoft.Data.SqlClient.SqlColumnEncryptionKeyStoreProvider.SignColumnMasterKeyMetadata(System.String,System.Boolean)" /> method doesn't break applications that rely on an old API, it throws a <see cref="T:System.NotImplementedException" /> exception by default.
+        </para>
+        <para>
```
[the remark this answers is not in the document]
paulmedynski: The default behaviour is clear.

paulmedynski (approved):

apoorvdeshmukh (approved):

src/Microsoft.Data.SqlClient/ref/Microsoft.Data.SqlClient.cs
mdaigle: The customer doesn't have to implement multiple. They implement the single virtual method that accepts a cancellation token. The overload that does not accept a cancellation token is concrete and calls directly into the other passing CancellationToken.None.

mdaigle (approved):

PR #3676, opened 2025-10-08
Merge | SqlCommand Encryption Methods

benrr101:
## Description
This is the second to last PR I have planned for merging SqlCommand. In this installment, the methods related to parameter encryption are merged, as well as some other blocks that don't have well defined organization. I'm not certain that these should have their own partial, but to avoid the base partial being too large to navigate, I made the executive decision to make a partial. Each commit is bite-sized and focuses on a single or a couple related methods at a time.

The following methods were merged in this PR:

* `BuildStoredProcedureStatementForColumnEncryption`
* `ClearDescribeParameterEncryptionRequests`
* `_currentlyExecutingDescribeParameterEncryptionRPC`
* `_customColumnEncryptionKeyStoreProviders`
* `customData`
* `customDataLength`
* `enclaveAttestationParameters`
* `GetColumnEncryptionCustomKeyProvidersNames`
* `GetEnclaveSessionParameters`
* `GetParameterEncryptionDataReader` - code from netcore taken since it is more concise
* `GetParameterEncryptionDataReaderAsync` - code from netcore taken since it is more concise
* `HasColumnEncryptionKeyStoreProviderRegistered`
* `InvalidateEnclaveSession`
* `IsDescribeParameterEncryptionRPCCurrentlyInProgress`
* `keysToBeSentToEnclave`
* `PrepareDescribeParameterEncryptionRequest`
* `PrepareTransparentEncryptionFinallyBlock`
* `ReadDescribeEncryptionParameterResults`
  * Factored this massive method into three additional methods (1-3) that read a specific result set from `sp_describe_parameter_encryption`
  * Removed debug-only `rowsAffected` member as it was only being used to check if the server returned the right number of rows from `sp_describe_parameter_encryption` in debug mode.
* `requiresEnclaveComputations`
* `ResetEncryptionState`
* `RowsAffectedByDescribeParameterEncryption`
* `_rowsAffectedBySpDescribeParameterEncryption`
* `_rpcForEncryption`
* `SetColumnEncryptionSetting`
* `ShouldCacheEncryptionMetadata`
* `ShouldUseEnclaveBasedWorkflow`
* `_sqlRPCParameterEncryptionRegArray`
* Test behavior overrides
  * `_forceInternalEndQuery`
  * `_forceRetryableEnclaveQueryExecutionExceptionDuringGenerateEnclavePackage`
  * `_sleepAfterReadDescribeEncryptionParameterResults`
  * `_sleepDuringRunExecuteReaderTdsForSpDescribeParameterEncryption`
  * `_sleepDuringTryFetchInputParameterEncryptionInfo`
* `TryFetchInputParameterEncryptionInfo`
* `TryGetColumnEncryptionKeyStoreProvider`
* `ValidateCustomProviders`
* `_wasBatchModeColumnEncryptionSettingsSetOnce`

## Issues
Continuation of work in https://github.com/dotnet/SqlClient/issues/1261

## Testing
Build passes, SqlCommandTests pass locally. CI should validate the rest of it.

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Encryption.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Encryption.cs
@@ -1064,7 +1064,7 @@ namespace Microsoft.Data.SqlClient
                     SqlSecurityUtility.VerifyColumnMasterKeySignature(
                         providerName,
                         keyPath,
-                        isEnclaveEnabled: true,
+                        isEnclaveEnabled: isRequestedByEnclave,
                         keySignature,
                         _activeConnection,
                         this);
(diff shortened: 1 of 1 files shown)


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Encryption.cs
```diff
@@ -164,5 +165,87 @@ private SqlDataReader GetParameterEncryptionDataReaderAsync(
...
+                // part of the array. Refer to GetRPCObject().
+                for (int index = 0; index < userParamCount; index++)
+                {
+                    SqlParameter sqlParameter = rpc.userParams[index];
+                    Debug.Assert(sqlParameter is not null, "sqlParameter should not be null.");
+
+                    // @TODO: And what happens if they're not in the same order?
+                    if (SqlParameter.ParameterNamesEqual(sqlParameter.ParameterName, parameterName))
+                    {
+                        Debug.Assert(sqlParameter.CipherMetadata is not null, "param.CipherMetaData should not be null.");
```
edwardneal: This assertion's logic has been inverted.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Encryption.cs
```diff
@@ -164,5 +165,87 @@ private SqlDataReader GetParameterEncryptionDataReaderAsync(
...
+                // When the RPC object gets reused, the parameter array has more parameters than
+                // the valid params for the command. Null is used to indicate the end of the valid
+                // part of the array. Refer to GetRPCObject().
+                for (int index = 0; index < userParamCount; index++)
+                {
+                    SqlParameter sqlParameter = rpc.userParams[index];
+                    Debug.Assert(sqlParameter is not null, "sqlParameter should not be null.");
+
+                    // @TODO: And what happens if they're not in the same order?
+                    if (SqlParameter.ParameterNamesEqual(sqlParameter.ParameterName, parameterName))
```
edwardneal, line 1126: This needs a third parameter, passing `StringComparison.Ordinal`.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Encryption.cs
```diff
@@ -219,6 +219,227 @@ private void InvalidateEnclaveSession()
...
+                "_activeConnection.Parser should not be null in PrepareForTransparentEncryption.");
+            Debug.Assert(_activeConnection.Parser.IsColumnEncryptionSupported,
+                "_activeConnection.Parser.IsColumnEncryptionSupported should be true in PrepareForTransparentEncryption.");
+            Debug.Assert(_columnEncryptionSetting == SqlCommandColumnEncryptionSetting.Enabled
+                         || (_columnEncryptionSetting == SqlCommandColumnEncryptionSetting.UseConnectionSetting && _activeConnection.IsColumnEncryptionSettingEnabled),
+                "ColumnEncryption setting should be enabled for input parameter encryption.");
+            Debug.Assert(isAsync == (completion != null),
+                "completion should can be null if and only if mode is async.");
+
+            // Fetch reader witn input params
```
edwardneal: nit: "witn" -> "with"

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Encryption.cs
```diff
@@ -259,6 +259,165 @@ private void InvalidateEnclaveSession()
...
+                    "_stateObj should be null at this time, in PrepareDescribeParameterEncryptionRequest.");
+                Debug.Assert(_activeConnection is not null,
+                    "_activeConnection should not be null at this time, in PrepareDescribeParameterEncryptionRequest.");
+
+                // @TODO: Shouldn't there be a way to do all this straight from the connection itself?
+                TdsParser tdsParser = null;
+                if (_activeConnection.Parser is not null)
+                {
+                    tdsParser = _activeConnection.Parser;
+                    if (tdsParser?.State is TdsParserState.Broken or TdsParserState.Closed)
```
edwardneal: We also want to throw this exception if `tdsParser` is null.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Encryption.cs
benrr101: Technically the default value of that parameter is `Ordinal`, and all the references to the method use Ordinal. I'm ok with forcing specification of the comparison, but since it's not necessary here, and removing it brings the line length within 120 ... :)

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Encryption.cs
benrr101: Hmm. Wonder how I ended up doing that.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Encryption.cs
benrr101: And right above that too "should can" 😂

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Encryption.cs
benrr101: *sigh* technically you're right - though the checks are being made in so many different places, I had to think through it for a solid 10min before I realized, yes it technically could still be null at that stage.

benrr101: But even so, as it is, it has a race condition in it and doesn't account for the case where the parser is null but connection.

How about I rewrite it like this (as per next commit)

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.netcore.cs
```diff
@@ -1790,47 +1690,15 @@ private void ReadDescribeEncryptionParameterResults(
...
-                            isRetry);
-                        enclaveAttestationParameters = null;
-                        attestationInfoRead = true;
-                    }
-
-                    if (!attestationInfoRead)
-                    {
-                        throw SQL.AttestationInfoNotReturnedFromSqlServer(this._activeConnection.Parser.EnclaveType, this._activeConnection.EnclaveAttestationUrl);
-                    }
+                    ReadDescribeEncryptionParameterResults3(ds, isRetry);
```
mdaigle: Would like to see these refactored to have more meaningful names.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Encryption.cs
edwardneal: Agreed - it'd be better to remove the parameter from ParameterNamesEqual then. We can do that later though...

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Encryption.cs
edwardneal: That looks good, thanks.  I agree there's a race condition - TdsParser instances being set to null is annoying!

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.netcore.cs
benrr101: Sure - I was going for the sequence of the result sets, since the main method is the one that gets the reader into position to read them. But I can rename to a very simple explanation of whats contained in the resultset.

mdaigle:
Looks like there are some genuine test failures in here:
image
image

benrr101:
Found all the bugs! Requesting new reviews from @mdaigle @paulmedynski :)

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Encryption.cs
```diff
@@ -198,7 +198,7 @@ private SqlParameter BuildStoredProcedureStatementForColumnEncryption(
                 // @TODO: Invert
                 if (parameter.Direction is not ParameterDirection.ReturnValue)
                 {
-                    execStatement.Append(' ');
+                    execStatement.Append(", ");
```
mdaigle, line 201: I'm curious how this didn't cause an issue before. Did this PR also refactor the building of the statement string?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Encryption.cs
```diff
@@ -1064,7 +1064,7 @@ private bool ReadDescribeEncryptionParameterResultsKeys(
                     SqlSecurityUtility.VerifyColumnMasterKeySignature(
                         providerName,
                         keyPath,
-                        isEnclaveEnabled: true,
+                        isEnclaveEnabled: isRequestedByEnclave,
```
mdaigle, line 1067: Can you explain the intention behind these changes? I'm finding it hard to follow with just the diff.

mdaigle (changes requested):
Great to see the build passing. Some commentary on the fixes will be nice.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Encryption.cs
```diff
@@ -140,7 +140,7 @@ private SqlParameter BuildStoredProcedureStatementForColumnEncryption(
             if (returnValueParameter is not null)
             {
                 SqlParameter.AppendPrefixedParameterName(execStatement, returnValueParameter.ParameterName);
-                execStatement.Append("=");
+                execStatement.Append('=');
```
benrr101, line 143: Replacing `"` with `'` to probably remove a string allocation.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Encryption.cs
benrr101, line 201: This one caused the "invalid syntax" error in one of the tests

benrr101: It did cause the "invalid syntax" error in one of the tests.
There was some slight changes made, mostly changing AppendFormats to Append when it was just a straight string.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Encryption.cs
benrr101: I'm not entirely sure how I made this mistake originally. In my attempt to figure out the problems the mistakes I made, I basically re-did the merge on another copy of the repo (but this time copy and paste only, no rewriting (which i probably should've done from the start, but oh well lesson learned)) which passed the tests, so I compared that merge with this merge, scrutinized every change, brought over all the trivial differences, then found the three I had to fix:
* The `", "` one
* This one
* Closing the if statement in the wrong place

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Encryption.cs
```diff
@@ -1276,92 +1276,95 @@ private SqlDataReader TryFetchInputParameterEncryptionInfo(
                         enclaveAttestationParameters,
                         enclaveType);
                 }
+            }
```
benrr101, line 1279: This caused the last issue in the tests - I had originally brought it over at the bottom of the following changes, but it should have been here. This meant we were only doing a whole bunch of other transforms to the RPC list if "enclave based workflow" was required.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Encryption.cs
mdaigle: I see, I thought this was a new change over what's currently present in the unmerged SqlCommand file.

PR #3628, opened 2025-09-23
Abstractions Package - Pipeline Changes

paulmedynski:
## Overview

As part of the Azure split work, a new Abstractions package is being created that contains types shared between MDS and its future extensions (Azure will be the first). This PR contains the YAML pipeline changes for CI and Official builds.

This supercedes the original PRs #3471 and #3567.

This depends on PR #3626.

## Issues

The first step towards #1108.

## Testing

- Manual inspection of all artifacts produced by the pipelines.

PR #3626, opened 2025-09-23
Abstractions Package - C# project changes

paulmedynski:
## Overview

As part of the Azure split work, a new Abstractions package is being created that contains types shared between MDS and its future extensions (Azure will be the first). This PR contains the C# project changes that create that package (with no meaningful content) to setup the MDS dependency.

This supercedes the original PRs #3471 and #3567.

This is depended on by PR #3628, which contains all of the pipeline changes to get the Abstractions package built and tested for CI, and packaged for official release.

## Issues

The first step towards #1108.

## Testing

- Unit tests for the sample class to prove that IDE and command-line building and testing work.
- There are no MDS tests to prove out the new dependency yet. Those will naturally happen when the first real bit of Azure functionality is moved out into the new Azure extension package.

PR #3680, opened 2025-10-10
Move AAD/Entra Authentication into new Azure package

paulmedynski:
## Overview

This PR contains changes to move the `ActiveDirectoryAuthenticationProvider` out of the MDS package and into the new Azure package.  Please refer to the design here:

[MDS Azure Extension Design](https://github.com/dotnet/SqlClient/discussions/3579)

This PR is only concerned with moving the provider code out of MDS and creating the necessary base classes and types in the Abstractions package, which includes:
- Creation of base classes/types in Abstractions related to the SQL authentication provider APIs.
- Move of the `ActiveDirectoryAuthenticationProvider.cs` file into the Azure package.
- Creation of the Azure package and related PR/CI pipelines.
- Targeted updates/improvements due to blast radius/fallout of the above.

Future PRs will address:
- Proper documentation (API, samples, reference).
- Dependency updates.
- Terminology modernization.
- String localization.
- Code refactoring/improvements.
- Addition of a synchronous `AcquireToken()` method.
- Separation of existing tests out of the MDS project and into the Azure project.
- Addition of unit tests and improved integration tests.
- New tests in the MDS project to confirm the authentication provider management logic.
- Any additional issues we come up with that aren't directly related to the move.

This builds on the previous 2 PRs:
- #3626 
- #3628

## Testing
All existing tests in the MDS project are passing.  This confirms that the move didn't break anything that is currently being tested.


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlAuthenticationParameters.cs
```diff
@@ -11,35 +11,8 @@ namespace Microsoft.Data.SqlClient
 {
 
     /// <include file='../../../../../../doc/snippets/Microsoft.Data.SqlClient/SqlAuthenticationParameters.xml' path='docs/members[@name="SqlAuthenticationParameters"]/SqlAuthenticationParameters/*'/>
-    public class SqlAuthenticationParameters
+    public class SqlAuthenticationParameters : SqlAuthenticationParametersBase
```
edwardneal: What about completely moving SqlAuthenticationParameters and SqlAuthenticationToken to MDS.Abstractions and using `[TypeForwardedTo]` instead?

paulmedynski: I think you're correct about SqlAuthenticationParameters.  I would just make a new internal class for the Builder and update any code that calls it currently.

For SqlAuthenticationToken, it's constructor makes a policy decision specific to MDS that I don't think we want to impose generally.  Because the MDS SqlAuthenticationToken is part of the public API, I can't rename it or change the constructor behaviour without breaking existing apps.  Maybe that would be OK for a major version bump - I will discuss internally.

I think the same goes for SqlAuthenticationProvider.  It contains 2 static functions that cannot move to Abstractions, and renaming it could similarily break apps calling those functions.

edwardneal: Thanks. I definitely agree with you on SqlAuthenticationProvider.

The policy decision on SqlAuthenticationToken seems to be that we can't hand an instance which represents an empty token to SqlClient. What do you think about keeping the policy decision in the constructor (so we guarantee that every instance represents a non-null, non-empty token value); allowing SqlAuthenticationProvider.AcquireTokenAsync to return a Task with a value of null if it wants to represent an empty token; and ensuring that the call sites in SqlClient reject null? This seems like it'd separate the policy reasonably well.

paulmedynski: You're correct - there doesn't seem to be a valid use case for a null or empty token, so I've moved that policy into Abstractions.  Providers can throw from AcquireTokenAsync() if they would otherwise return one.

eng/pipelines/common/templates/jobs/validate-signed-package-job.yml
```diff
@@ -305,17 +305,6 @@ jobs:
       Get-ChildItem *.dll -Path $(extractedNugetPath)  -Recurse | ForEach-Object VersionInfo | Format-List
     displayName: 'Verify "File Version" matches expected values for DLLs'
 
-  - powershell: |
```
paulmedynski, line 308: I removed the TestMicrosoftDataSqlClientVersion build property entirely.  It was made redundant by MdsPackageVersion.

src/Microsoft.Data.SqlClient.Extensions/Abstractions/src/Abstractions.csproj
```diff
@@ -23,17 +23,19 @@
...
       number, with all remaining parts set to 0.
     -->
-    <AssemblyVersion>$(_DefaultMajorVersion).0.0.0</AssemblyVersion>
+    <AssemblyVersion>$(AbstractionsDefaultMajorVersion).0.0.0</AssemblyVersion>
 
     <AssemblyFileVersion>$(AbstractionsAssemblyFileVersion)</AssemblyFileVersion>
     <FileVersion>$(AbstractionsAssemblyFileVersion)</FileVersion>
     <Version>$(AbstractionsPackageVersion)</Version>
+
+    <DocumentationFile>$(Artifacts)/doc/$(TargetFramework)/$(AssemblyName).xml</DocumentationFile>
```
paulmedynski, line 44: I will figure out where to put the XML docs in a later PR.  For now, they go somewhere ignored by git.

src/Microsoft.Data.SqlClient.Extensions/Abstractions/src/SqlAuthenticationMethod.cs
paulmedynski, line 1: This enum was moved out of TdsEnums.cs as-is.

src/Microsoft.Data.SqlClient.Extensions/Abstractions/src/SqlAuthenticationParameters.cs
```diff
@@ -0,0 +1,59 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.Data.SqlClient;
+
+/// <include file='../doc/SqlAuthenticationParameters.xml' path='docs/members[@name="SqlAuthenticationParameters"]/SqlAuthenticationParameters/*'/>
+public sealed class SqlAuthenticationParameters
```
paulmedynski, line 8: This and the token class are sealed.  There is no need for an app to derive from them.

src/Microsoft.Data.SqlClient.Extensions/Abstractions/src/SqlAuthenticationToken.cs
```diff
@@ -0,0 +1,37 @@
...
+
+    /// <include file='../doc/SqlAuthenticationToken.xml' path='docs/members[@name="SqlAuthenticationToken"]/ExpiresOn/*'/>
+    public DateTimeOffset ExpiresOn { get; }
+
+    /// <include file='../doc/SqlAuthenticationToken.xml' path='docs/members[@name="SqlAuthenticationToken"]/ctor/*'/>
+    public SqlAuthenticationToken(
+        string accessToken,
+        DateTimeOffset expiresOn)
+    {
+        if (string.IsNullOrEmpty(accessToken))
```
paulmedynski, line 21: A null or empty token is meaningless, so I moved this policy decision out of MDS and into Abstractions.  I think this is correct.

src/Microsoft.Data.SqlClient.Extensions/Azure/src/ActiveDirectoryAuthenticationProvider.cs
paulmedynski, line 1: This file was moved from MDS.  I'm not sure why it doesn't show up as such.  I will call out differences.

src/Microsoft.Data.SqlClient.Extensions/Azure/src/ActiveDirectoryAuthenticationProvider.cs
```diff
@@ -0,0 +1,868 @@
...
+                    // If no existing 'account' is found, we request user to sign in interactively.
+                    result = await AcquireTokenInteractiveDeviceFlowAsync(app, scopes, parameters.ConnectionId, parameters.UserId, parameters.AuthenticationMethod, cts, _customWebUI, _deviceCodeFlowCallback).ConfigureAwait(false);
+                    SqlClientEventSource.Log.TryTraceEvent("AcquireTokenAsync | Acquired access token (interactive) for {0} auth mode. Expiry Time: {1}", parameters.AuthenticationMethod, result?.ExpiresOn);
+                }
+            }
+            else
+            {
+                SqlClientEventSource.Log.TryTraceEvent("AcquireTokenAsync | {0} authentication mode not supported by ActiveDirectoryAuthenticationProvider class.", parameters.AuthenticationMethod);
+
+                throw new Extensions.Azure.AuthenticationException(
```
paulmedynski, line 343: Throwing a different exception here.

src/Microsoft.Data.SqlClient.Extensions/Azure/src/ActiveDirectoryAuthenticationProvider.cs
```diff
@@ -0,0 +1,868 @@
...
+            if (result is null)
+            {
+                throw new Extensions.Azure.AuthenticationException(
+                    parameters.AuthenticationMethod,
+                    "Internal error - authentication result is null");
+            }
+
+            return new SqlAuthenticationToken(result.AccessToken, result.ExpiresOn);
+        }
+        catch (MsalException ex)
```
paulmedynski, line 358: This whole exception handling block is new, and moved the interpretation of MSAL errors from MDS to here.

src/Microsoft.Data.SqlClient.Extensions/Azure/src/ActiveDirectoryAuthenticationProvider.cs
```diff
@@ -0,0 +1,868 @@
...
+
+        if (account != null)
+        {
+            // If 'account' is available in 'app', we use the same to acquire token silently.
+            // Read More on API docs: https://docs.microsoft.com/dotnet/api/microsoft.identity.client.clientapplicationbase.acquiretokensilent
+            result = await app.AcquireTokenSilent(scopes, account).ExecuteAsync(cancellationToken: cts.Token).ConfigureAwait(false);
+            SqlClientEventSource.Log.TryTraceEvent("AcquireTokenAsync | Acquired access token (silent) for {0} auth mode. Expiry Time: {1}", parameters.AuthenticationMethod, result?.ExpiresOn);
+        }
+
+        // TODO: Existing bug?  result may be null here.
```
paulmedynski: Another new null case that wasn't handled in the old code.

src/Microsoft.Data.SqlClient.Extensions/Azure/src/ActiveDirectoryAuthenticationProvider.cs
```diff
@@ -0,0 +1,868 @@
...
+            else
+            {
+                SqlClientEventSource.Log.TryTraceEvent("AcquireTokenAsync | {0} authentication mode not supported by ActiveDirectoryAuthenticationProvider class.", parameters.AuthenticationMethod);
+
+                throw new Extensions.Azure.AuthenticationException(
+                    parameters.AuthenticationMethod,
+                    $"Authentication method {parameters.AuthenticationMethod} not supported.");
+            }
+
+            // TODO: Existing bug?  result may be null here.
```
paulmedynski, line 348: An existing Null case exposed by nullable context.  This will get tests written to verify it in a future PR.

src/Microsoft.Data.SqlClient.Extensions/Azure/src/ActiveDirectoryAuthenticationProvider.cs
```diff
@@ -0,0 +1,868 @@
...
+            }
+            return false;
+        }
+
+        public override int GetHashCode() => Tuple.Create(_tokenCredentialType, _authority, _scope, _audience, _clientId).GetHashCode();
+    }
+
+}
+
+internal class ActiveDirectoryAuthentication
```
paulmedynski: Some temporary helper classes that will be removed in future PRs as I work through remaining tasks.

src/Microsoft.Data.SqlClient.Extensions/Azure/src/AuthenticationException.cs
```diff
@@ -0,0 +1,29 @@
+namespace Microsoft.Data.SqlClient.Extensions.Azure;
+
+internal class AuthenticationException : SqlAuthenticationProviderException
+{
+    public AuthenticationException(
```
paulmedynski: @paulmedynski Could use some comments here.

src/Microsoft.Data.SqlClient/add-ons/Directory.Build.props
```diff
@@ -8,7 +8,6 @@
     <OSGroup Condition="$(OSGroup) == ''">$(OS)</OSGroup>
     <TargetsWindows Condition="'$(OSGroup)'=='Windows_NT'">true</TargetsWindows>
     <TargetsUnix Condition="'$(OSGroup)'=='Unix'">true</TargetsUnix>
-    <ReferenceType Condition="'$(ReferenceType)'==''">Project</ReferenceType>
```
paulmedynski, line 11: There were a couple of redundant default ReferenceType properties lying about, so I removed them.

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.csproj
```diff
@@ -32,8 +32,6 @@
     <Compile Include="..\..\ref\Microsoft.Data.SqlClient.Batch.NetCoreApp.cs" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Azure.Core" />
```
paulmedynski, line 35: Hooray!  The whole point of this work was to remove these dependencies from MDS.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -2635,89 +2622,44 @@ internal SqlFedAuthToken GetFedAuthToken(SqlFedAuthInfo fedAuthInfo)
...
-                        throw new MsalServiceException(MsalError.UnknownError, "Force retry in GetFedAuthToken");
-                    }
-#endif
+
                     // Break out of the retry loop in successful case.
                     break;
                 }
-                // Deal with Msal service exceptions first, retry if 429 received.
-                catch (MsalServiceException serviceException)
+                catch (SqlAuthenticationProviderException ex)
```
paulmedynski, line 2627: There are some subtle differences to what we throw and when here.  I will document them in the public APIs in a later PR.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ActiveDirectoryAuthenticationTimeoutRetryHelper.cs
```diff
@@ -132,7 +132,8 @@ private static string GetTokenHash(SqlFedAuthToken token)
             }
 
             // Here we mimic how ADAL calculates hash for token. They use UTF8 instead of Unicode.
-            var originalTokenString = SqlAuthenticationToken.AccessTokenStringFromBytes(token.accessToken);
+            var originalTokenString = Encoding.Unicode.GetString(token.accessToken);
```
paulmedynski: This is what AccessTokenStringFromBytes() used to do.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlAuthenticationParametersBuilder.cs
paulmedynski, line 1: This is what remained in MDS from the SqlAuthenticationParamaters implementation.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlAuthenticationProviderManager.cs
```diff
@@ -6,6 +6,10 @@
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Configuration;
+using System.IO;
+using System.Reflection;
+
+#nullable enable
```
paulmedynski, line 12: You will see some new `?` and null checks below related to this.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserHelperClasses.cs
```diff
@@ -143,6 +144,22 @@ internal sealed class SqlFedAuthToken
...
         internal long expirationFileTime;
+
+        internal SqlFedAuthToken()
+        {
+        }
+
+        /// <summary>
+        /// Convert from a SqlAuthenticationToken.
+        /// </summary>
+        internal SqlFedAuthToken(SqlAuthenticationToken token)
```
paulmedynski, line 169: Moved from SqlAuthenticationToken, since conversion _to_ a SqlFedAuthToken belongs here anyway.

tools/targets/GenerateThisAssemblyCs.targets
```diff
@@ -4,18 +4,31 @@
     <Compile Include="$(IntermediateOutputPath)\ThisAssembly.cs"/>
   </ItemGroup>
 
+  <PropertyGroup>
+    <!--
+      Prepend an assembly CLS Compliance directive unless
+      $(ThisAssemblyClsCompliant) is false.
+    -->
+    <ThisAssemblyClsCompliantContent Condition="'$(ThisAssemblyClsCompliant)' != 'false'">
+[assembly: System.CLSCompliant(true)]
```
paulmedynski, line 13: These new config properties are necessary to avoid collisions now that MDS depends on Abstractions.

doc/snippets/Microsoft.Data.SqlClient/SqlAuthenticationProviderManager.xml
paulmedynski, line 1: I decided to see how things look if we move GetProvider() and SetProvider() from SqlAuthenticationProvider to SqlAuthenticationProviderManager.  This is a breaking change, but it separates the provider API definition from the act of managing providers.  It was a mistake to combine these two into the provider API originally.  I feel liks this is a worthwhile breaking change, and it will only affect apps that were managing providers explicitly via the get/set methods.

If we don't like this change, I can easily revert since it is all in a single commit.

edwardneal: In my opinion the API split itself makes sense (provided the breaking change is acceptable.) The class itself is currently halfway between being a completely static class and a class with a singleton instance, and if it's going to be exposed then I think it'd be best to settle on one approach or the other.

I've got one or two other comments, but they're not central to the public API; it's up to you whether they get pushed out to a later preview.
* `AuthenticationEnumFromString` overlaps with `SqlConnectionStringBuilder`'s parsing logic, and if the constants line up then we might be able to simplify by using that instead.
* `SetProvider` currently iterates over a HashSet to perform a search for the authentication method; it can just use `Contains` instead.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserHelperClasses.cs
```diff
@@ -156,7 +156,7 @@ internal sealed class SqlFedAuthToken
 
         internal SqlFedAuthToken(
             byte[] accessToken,
-            uint expirationFileTime)
+            long expirationFileTime)
```
ErikEJ, line 159: Will this also affect the current codebase - seem to have seen a few issues relating to to many token renewals?

paulmedynski: This is me fixing my own bug from a few commits ago.  I botched this argument type when I introduced the constructor and made the fields immutable.  The main codebase accesses/mutates the fields directly (which isn't great, but doesn't truncate the long).

eng/pipelines/common/templates/steps/build-all-tests-step.yml
```diff
@@ -42,7 +45,7 @@ steps:
       solution: build.proj
       platform: '${{parameters.platform }}'
       configuration: '${{parameters.configuration }}'
-      msbuildArguments: '-t:BuildTestsNetFx -p:TF=${{parameters.targetFramework }} -p:TestSet=${{parameters.testSet }} -p:ReferenceType=${{parameters.referenceType }} -p:TestMicrosoftDataSqlClientVersion=${{parameters.mdsPackageVersion }} -p:AbstractionsPackageVersion=${{ parameters.abstractionsPackageVersion }}'
+      msbuildArguments: '-t:BuildTestsNetFx -p:TF=${{parameters.targetFramework }} -p:TestSet=${{parameters.testSet }} -p:ReferenceType=${{parameters.referenceType }} -p:MdsPackageVersion=${{parameters.mdsPackageVersion }} -p:AbstractionsPackageVersion=${{ parameters.abstractionsPackageVersion }} -p:AzurePackageVersion=${{ parameters.azurePackageVersion }}'
```
paulmedynski, line 48: For now, the existing tests for AAD/Entra auth will remain within the MDS tests, so they need to depend on the new Azure package where the code was moved.

In a future PR, these tests will move over to the Azure package, and none of the MDS tests will depend on the Azure package.

eng/pipelines/jobs/pack-azure-package-ci-job.yml
```diff
@@ -0,0 +1,188 @@
...
+
+        # Download the Abstractions package artifacts into packages/.
+        - task: DownloadPipelineArtifact@2
+          displayName: Download Abstractions Package Artifact
+          inputs:
+            artifactName: ${{ parameters.abstractionsArtifactName }}
+            targetPath: $(Build.SourcesDirectory)/packages
+
+        # Use the local NuGet.config that references the packages/ directory.
+        - pwsh: cp $(Build.SourcesDirectory)/NuGet.config.local $(Build.SourcesDirectory)/NuGet.config
```
paulmedynski, line 142: This is much simpler than what the existing pipelines do to parse the JSON, update it, and then write it back out via PowerShell foo.  We should use this approach when we develop the new pipelines.

src/Microsoft.Data.SqlClient.Extensions/Abstractions/doc/SqlAuthenticationProviderException.xml
paulmedynski, line 1: Docs for the Abstractions and Azure packages are still considered works in progress.  There will be future PRs that provide complete documentation.

src/Microsoft.Data.SqlClient.Extensions/Abstractions/test/SqlAuthenticationProviderExceptionTest.cs
```diff
@@ -0,0 +1,117 @@
...
+        var causedBy = new Exception("causedBy");
+
+        Error ex = new(message, causedBy);
+
+        Assert.Equal(SqlAuthenticationMethod.NotSpecified, ex.Method);
+        Assert.Equal("Unknown", ex.FailureCode);
+        Assert.False(ex.ShouldRetry);
+        Assert.Equal(0, ex.RetryPeriod);
+        Assert.Equal(message, ex.Message);
+        Assert.Equal(causedBy, ex.InnerException);
```
paulmedynski: @paulmedynski - Use Assert.Is() instead.

src/Microsoft.Data.SqlClient.Extensions/Abstractions/test/SqlAuthenticationTokenTest.cs
```diff
@@ -0,0 +1,47 @@
+namespace Microsoft.Data.SqlClient.Extensions.Abstractions.Test;
```
paulmedynski, line 5: @paulmedynski - Missing typical file header.

src/Microsoft.Data.SqlClient.Extensions/Abstractions/test/SqlAuthenticationTokenTest.cs
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    // Verify that a null token is rejected.
+    [Fact]
+    public void Constructor_NullToken()
+    {
+        string? token = null;
+        var expiry = DateTimeOffset.UtcNow.AddHours(1);
+
+        var ex = Assert.Throws<SqlAuthenticationToken.TokenException>(() =>
```
paulmedynski: @paulmedynski - This should confirm that the exception is-a SqlAuthenticationProviderException, since that is the public type.

src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider.csproj
```diff
@@ -33,15 +33,22 @@
     <EmbeddedFiles Include="$(GeneratedAssemblyInfoFile)" />
     <EmbeddedFiles Include="$(TargetFrameworkMonikerAssemblyAttributesPath)" />
   </ItemGroup>
-  <ItemGroup>
-    <ProjectReference Condition="'$(TargetGroup)'=='netcoreapp' AND !$(ReferenceType.Contains('Package'))" Include="$(NetCoreSource)src\Microsoft.Data.SqlClient.csproj" />
-    <ProjectReference Condition="'$(TargetGroup)'=='netfx' AND !$(ReferenceType.Contains('Package'))" Include="$(NetFxSource)src\Microsoft.Data.SqlClient.csproj" />
-    <PackageReference Condition="$(ReferenceType.Contains('Package'))" Include="Microsoft.Data.SqlClient" />
+
+  <!-- Conditional Project References -->
```
paulmedynski, line 37: No functional changes here.  I just moved the conditions onto ItemGroups to avoid duplicating them on each Project/Package reference.  Similar changes to other project files below.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -232,17 +230,6 @@ internal bool IsDNSCachingBeforeRedirectSupported
...
-        // This is a test hook to enable testing of the retry paths for MSAL get access token.
-        // Sample code to enable:
-        //
-        //    Type type = typeof(SqlConnection).Assembly.GetType("Microsoft.Data.SqlClient.SQLInternalConnectionTds");
-        //    System.Reflection.FieldInfo field = type.GetField("_forceMsalRetry", System.Reflection.BindingFlags.NonPublic | System.Reflection.BindingFlags.Static);
-        //    if (field != null) {
-        //        field.SetValue(null, true);
-        //    }
-        //
-        internal static bool _forceMsalRetry = false;
```
paulmedynski, line 241: @paulmedynski - This probably breaks some existing tests that only exist in DEBUG.  Find them and make a note so they can be moved into the Azure package tests properly.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -2495,44 +2481,41 @@ internal bool TryGetFedAuthTokenLocked(SqlFedAuthInfo fedAuthInfo, DbConnectionP
...
-            // No:of attempts, for tracing purposes, if we underwent retries.
-            int numberOfAttempts = 0;
+            // Number of milliseconds to sleep for the initial back off, if a
+            // retry period is not specified by the provider.
+            const int defaultRetryPeriod = 100;
 
-            // Object that will be returned to the caller, containing all required data about the token.
-            _fedAuthToken = new SqlFedAuthToken();
+            // Number of retry attempts we are willing to perform.
+            const int maxRetryAttempts = 1;
```
paulmedynski: The old while loop was hardcoded to perform at most 2 attempts (1 retry), so I codified that here to make it obvious.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -19,7 +19,6 @@
 using Microsoft.Data.Common.ConnectionString;
 using Microsoft.Data.ProviderBase;
 using Microsoft.Data.SqlClient.ConnectionPool;
-using Microsoft.Identity.Client;
```
paulmedynski, line 22: I'll sure be happy when we don't have to make complex changes like this to 2 different files with possibly subtle differences!

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ActiveDirectoryAuthenticationTimeoutRetryHelper.cs
```diff
@@ -132,7 +132,8 @@ private static string GetTokenHash(SqlFedAuthToken token)
             }
 
             // Here we mimic how ADAL calculates hash for token. They use UTF8 instead of Unicode.
-            var originalTokenString = SqlAuthenticationToken.AccessTokenStringFromBytes(token.accessToken);
+            var originalTokenString = Encoding.Unicode.GetString(token.AccessToken);
```
paulmedynski, line 135: This is all that `AccessTokenStringFromBytes()` was doing.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlAuthenticationProviderManager.cs
```diff
@@ -46,49 +48,116 @@ static SqlAuthenticationProviderManager()
...
-        }
 
-        /// <summary>
-        /// Sets default supported Active Directory Authentication providers by the driver 
-        /// on the SqlAuthenticationProviderManager instance.
-        /// </summary>
-        private static void SetDefaultAuthProviders(SqlAuthenticationProviderManager instance)
-        {
-            if (instance != null)
+            // If our Azure extensions package is present, use its
```
paulmedynski, line 51: This is the new logic that looks for the Azure package assembly and loads the AAD Auth class from it.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlException.cs
```diff
@@ -205,8 +205,6 @@ internal static SqlException CreateException(SqlErrorCollection errorCollection,
 
         internal static SqlException CreateException(SqlErrorCollection errorCollection, string serverVersion, Guid conId, Exception innerException = null, SqlBatchCommand batchCommand = null)
         {
-            Debug.Assert(errorCollection != null && errorCollection.Count > 0, "no errorCollection?");
```
paulmedynski, line 208: Don't use debug asserts to check pre-conditions!

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlUtil.cs
```diff
@@ -1382,17 +1382,6 @@ internal static Exception UnsupportedFeatureAndToken(SqlInternalConnectionTds in
             return exc;
         }
 
-        internal static Exception Azure_ManagedIdentityException(string msg)
```
paulmedynski, line 1385: This was unused before any of these changes.  I removed it to avoid future confusion.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserHelperClasses.cs
```diff
@@ -128,23 +129,53 @@ internal sealed class SqlLoginAck
         internal uint tdsVersion;
     }
 
+    #nullable enable
+
     internal sealed class SqlFedAuthInfo
     {
-        internal string spn;
-        internal string stsurl;
+        internal string Spn { get; }
```
paulmedynski, line 136: These fields are now immutable, which provides a stronger class invariant.  Same changes to SqlFedAuthToken below.

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/DataClassificationTest/DataClassificationTest.cs
```diff
@@ -25,7 +25,6 @@ public static void TestDataClassificationResultSetRank()
                 try
                 {
                     sqlConnection.Open();
-                    Assert.True(DataTestUtility.IsSupportedDataClassification());
```
paulmedynski, line 28: This is already checked by the `[ConditionalFact]` above.

eng/pipelines/stages/build-azure-package-ci-stage.yml
```diff
@@ -0,0 +1,162 @@
...
+      # Build and test on Windows
+
+      - template: ../jobs/test-azure-package-ci-job.yml@self
+        parameters:
+          abstractionsArtifactName: ${{ parameters.abstractionsArtifactName }}
+          abstractionsPackageVersion: ${{ parameters.abstractionsPackageVersion }}
+          buildConfiguration: ${{ parameters.buildConfiguration }}
+          displayNamePrefix: Win
+          jobNameSuffix: windows
+          netFrameworkRuntimes: [net462, net47, net471, net472, net48, net481]
```
David-Engel: Do we really need to test all those combinations? Don't other jobs just target the lowest version we support?

src/Microsoft.Data.SqlClient.Extensions/Abstractions/doc/SqlAuthenticationProvider.xml
```diff
@@ -89,48 +89,36 @@
...
+      <remarks>This method must not throw.</remarks>
       <returns>
         <see langword="true" /> if the specified authentication method is supported; otherwise, <see langword="false" />.
       </returns>
+      <param name="authenticationMethod">The authentication method.</param>
     </IsSupported>
     <AcquireTokenAsync>
-      <param name="parameters">The Active Directory authentication parameters passed by the driver to authentication providers.</param>
       <summary>Acquires a security token from the authority.</summary>
+      <param name="parameters">The Active Directory authentication parameters passed by the driver to authentication providers.</param>
```
David-Engel: Perfect opportunity to fix this name (Azure AD to Entra ID rebranding a few years ago). In fact, I don't think we even need the second-fourth words:
```suggestion
      <param name="parameters">The parameters passed by the driver to authentication providers.</param>
```

src/Microsoft.Data.SqlClient.Extensions/Abstractions/doc/SqlAuthenticationParameters.xml
```diff
@@ -6,23 +6,31 @@
...
-      <param name="userId">The user login name/ID.</param>
-      <param name="password">The user password.</param>
+      <param name="userId">The user login name/ID, or null if not applicable.</param>
+      <param name="password">The user password, or null if not applicable.</param>
       <param name="connectionId">The connection ID.</param>
-      <param name="connectionTimeout">The connection timeout value in seconds.</param>
-      <summary>
-        Initializes a new instance of the <see cref="T:Microsoft.Data.SqlClient.SqlAuthenticationParameters" /> class using the specified authentication method, server name, database name, resource URI, authority URI, user login name/ID, user password, connection ID and connection timeout value.
-      </summary>
+      <param name="connectionTimeout">The connection timeout, in seconds.</param>
```
David-Engel: This has always bothered me. This should be the timeout for the authentication step within the overall connection establishment. We might even be within a command timeout during an idle connection recovery event. Calling it "connection timeout" seems inaccurate. I think "authentication timeout" would be better.

src/Directory.Packages.props
```diff
@@ -109,16 +124,22 @@
...
+  <!-- ===================================================================== -->
   <!-- MSS Dependencies -->
 
   <!-- None -->
 
+  <!-- ===================================================================== -->
+  <!-- Azure Dependencies -->
+
+  <ItemGroup>
+    <PackageVersion Include="Azure.Identity" Version="1.14.2" />
```
mdaigle: Is there a reason not to use the latest version (1.17.0)?

src/Microsoft.Data.SqlClient.Extensions/Azure/src/AuthenticationException.cs
```diff
@@ -0,0 +1,33 @@
+namespace Microsoft.Data.SqlClient.Extensions.Azure;
```
mdaigle, line 5: Needs standard header

src/Microsoft.Data.SqlClient.Extensions/Azure/src/Azure.csproj
```diff
@@ -0,0 +1,113 @@
...
+    <DocumentationFile>$(Artifacts)/doc/$(TargetFramework)/$(AssemblyName).xml</DocumentationFile>
+  </PropertyGroup>
+  <ItemGroup>
+    <!-- Our API is CLS Compliant. -->
+    <AssemblyAttribute Include="System.CLSCompliantAttribute">
+      <_Parameter1>true</_Parameter1>
+    </AssemblyAttribute>
+
+    <!-- Our unit tests have access to our internal types. -->
+    <InternalsVisibleTo Include="$(AssemblyName).Test" />
```
mdaigle, line 60: Consider limiting to only non-official builds so that the friend assembly config is not present in the released artifacts.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ActiveDirectoryAuthenticationProvider.cs
```diff
@@ -1,743 +0,0 @@
...
-                    tokenCredentialInstance = CreateTokenCredentialInstance(tokenCredentialKey, secret);
-                    s_tokenCredentialMap.TryAdd(tokenCredentialKey, tokenCredentialInstance);
-                }
-                finally
-                {
-                    s_tokenCredentialMapModifierSemaphore.Release();
-                }
-            }
-
-            return await tokenCredentialInstance._tokenCredential.GetTokenAsync(tokenRequestContext, cancellationToken);
```
mdaigle, line 521: ✏️  I think this task can be returned directly

src/Microsoft.Data.SqlClient.Extensions/Abstractions/src/SqlAuthenticationProvider.cs
```diff
@@ -0,0 +1,21 @@
...
+
+    /// <include file='../doc/SqlAuthenticationProvider.xml' path='docs/members[@name="SqlAuthenticationProvider"]/BeforeUnload/*'/>
+    public virtual void BeforeUnload(SqlAuthenticationMethod authenticationMethod) { }
+
+    /// <include file='../doc/SqlAuthenticationProvider.xml' path='docs/members[@name="SqlAuthenticationProvider"]/IsSupported/*'/>
+    public abstract bool IsSupported(SqlAuthenticationMethod authenticationMethod);
+
+    /// <include file='../doc/SqlAuthenticationProvider.xml' path='docs/members[@name="SqlAuthenticationProvider"]/AcquireTokenAsync/*'/>
+    public abstract Task<SqlAuthenticationToken> AcquireTokenAsync(SqlAuthenticationParameters parameters);
+}
```
David-Engel, line 49: Add new sync API? (Or maybe you planned that for a little later...)
```suggestion
}

    public virtual SqlAuthenticationToken AcquireToken(SqlAuthenticationParameters parameters) => Task.Run(async () => await AcquireTokenAsync(parameters)).GetAwaiter().GetResult();
```

src/Microsoft.Data.SqlClient.Extensions/Abstractions/src/SqlAuthenticationToken.cs
David-Engel: If you move it out, you need to localize the exception message...

src/Microsoft.Data.SqlClient.Extensions/Abstractions/test/SqlAuthenticationMethodTest.cs
```diff
@@ -0,0 +1,34 @@
+namespace Microsoft.Data.SqlClient.Extensions.Abstractions.Test;
```
David-Engel, line 5: Missing license header

src/Microsoft.Data.SqlClient.Extensions/Abstractions/test/SqlAuthenticationParametersTest.cs
```diff
@@ -0,0 +1,78 @@
+namespace Microsoft.Data.SqlClient.Extensions.Abstractions.Test;
```
David-Engel, line 5: Missing license header

src/Microsoft.Data.SqlClient.Extensions/Abstractions/test/SqlAuthenticationProviderExceptionTest.cs
```diff
@@ -0,0 +1,117 @@
+namespace Microsoft.Data.SqlClient.Extensions.Abstractions.Test;
```
David-Engel, line 5: Missing license header

src/Microsoft.Data.SqlClient.Extensions/Azure/src/Azure.csproj
```diff
@@ -0,0 +1,113 @@
...
+  <ItemGroup Condition="'$(ReferenceType)' == 'Project'">
+    <ProjectReference Include="../../Abstractions/src/Abstractions.csproj"/>
+  </ItemGroup>
+  <ItemGroup Condition="'$(ReferenceType)' == 'Package'">
+    <PackageReference Include="Microsoft.Data.SqlClient.Extensions.Abstractions" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Azure.Identity" />
+    <PackageReference Include="Microsoft.Extensions.Caching.Memory" />
```
mdaigle, line 99: We'll need a nuspec file for this package.

src/Microsoft.Data.SqlClient.Extensions/Azure/src/AuthenticationException.cs
```diff
@@ -0,0 +1,33 @@
...
+// This exception is used internally by authentication providers to signal
+// authentication failures. It is not exposed publicly.
+internal class AuthenticationException : SqlAuthenticationProviderException
+{
+    // Construct with just a method and message.  Other properties are set to
+    // defaults per the base class.
+    internal AuthenticationException(
+        SqlAuthenticationMethod method,
+        string message)
+    : base($"Failed to acquire access token for {method}: {message}", null)
```
David-Engel, line 22: Are the messages exposed publicly? (Do they need localization?)

src/Microsoft.Data.SqlClient.Extensions/Azure/test/Azure.Test.csproj
```diff
@@ -0,0 +1,26 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net462;net47;net471;net472;net48;net481;net8.0;net9.0</TargetFrameworks>
```
David-Engel: I think we really only need net462...

David-Engel (commented):
Partial 45/89 review

src/Microsoft.Data.SqlClient.Extensions/Azure/src/AuthenticationException.cs
paulmedynski: Yes, in a future PR.

src/Microsoft.Data.SqlClient.Extensions/Abstractions/src/SqlAuthenticationProvider.cs
paulmedynski: Yes, in a future PR.

src/Microsoft.Data.SqlClient.Extensions/Azure/src/Azure.csproj
paulmedynski: The project file contains everything required to generate the NuGet package.

mdaigle: great!

src/Microsoft.Data.SqlClient.Extensions/Abstractions/src/SqlAuthenticationToken.cs
paulmedynski: Will be localized in a future PR.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
paulmedynski: Nope, nothing was using this except the commented-out lines that were removed elsewhere in this file.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ActiveDirectoryAuthenticationProvider.cs
paulmedynski: GetTokenAsync() returns a ValueTask<AccessToken> which can't be implicitly converted to AccessToken.

src/Microsoft.Data.SqlClient.Extensions/Abstractions/src/SqlAuthenticationParameters.cs
```diff
@@ -44,7 +50,7 @@ public SqlAuthenticationParameters(
         string? userId,
         string? password,
         Guid connectionId,
-        int connectionTimeout)
+        int authenticationTimeout)
```
paulmedynski: @paulmedynski - Hmmm... is this a breaking change for callers using named arguments?

src/Microsoft.Data.SqlClient.Extensions/Abstractions/src/SqlAuthenticationProvider.cs
```diff
@@ -18,4 +18,28 @@ public virtual void BeforeUnload(SqlAuthenticationMethod authenticationMethod) {
...
     /// <include file='../doc/SqlAuthenticationProvider.xml' path='docs/members[@name="SqlAuthenticationProvider"]/AcquireTokenAsync/*'/>
     public abstract Task<SqlAuthenticationToken> AcquireTokenAsync(SqlAuthenticationParameters parameters);
+
+
+    /// <include file='../doc/SqlAuthenticationProvider.xml' path='docs/members[@name="SqlAuthenticationProvider"]/GetProvider/*'/>
+    [Obsolete(
+        "This method is deprecated and will be removed in future versions.  " +
+        "Use SqlAuthenticationProviderManager.GetProvider() from the " +
+        "Microsoft.Data.SqlClient package instead.")]
+    public static SqlAuthenticationProvider? GetProvider(
```
paulmedynski, line 30: @paulmedynski - Add some comments here for future devs suggesting when these methods may be removed.

src/Microsoft.Data.SqlClient.Extensions/Abstractions/src/SqlAuthenticationProviderInternal.cs
```diff
@@ -0,0 +1,150 @@
...
+
+/// <include file='../doc/SqlAuthenticationProvider.xml' path='docs/members[@name="SqlAuthenticationProvider"]/SqlAuthenticationProvider/*'/>
+public abstract partial class SqlAuthenticationProvider
+{
+    private static class Internal
+    {
+        private static MethodInfo? _getProvider = null;
+        private static MethodInfo? _setProvider = null;
+
+        static Internal()
```
paulmedynski, line 32: @paulmedynski - Add some comments explaining this Internal class and its methods.

src/Microsoft.Data.SqlClient.Extensions/Abstractions/test/SqlAuthenticationProviderTest.cs
```diff
@@ -0,0 +1,43 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.Data.SqlClient.Extensions.Abstractions.Test;
+
+public class SqlAuthenticationProviderTest
+{
+    [Fact]
+    public void GetProvider()
```
paulmedynski: @paulmedynski - Document what these tests are testing.

doc/snippets/Microsoft.Data.SqlClient/SqlAuthenticationProviderManager.xml
cheenamalhotra: This will be of lot of impact and will break many customers on upgrade as most customers use this API to register custom provider, I would say we keep it as-is for now if its not necessarily needed.

We need to keep the API impact nil for the Azure Split work, only library inclusion should be needed.

cheenamalhotra (changes requested):

src/Microsoft.Data.SqlClient.Extensions/Abstractions/src/SqlAuthenticationParameters.cs
```diff
@@ -0,0 +1,69 @@
...
+    public string DatabaseName { get; }
+
+    /// <include file='../doc/SqlAuthenticationParameters.xml' path='docs/members[@name="SqlAuthenticationParameters"]/AuthenticationTimeout/*'/>
+    public int AuthenticationTimeout { get; }
+
+    /// <include file='../doc/SqlAuthenticationParameters.xml' path='docs/members[@name="SqlAuthenticationParameters"]/ConnectionTimeout/*'/>
+    [Obsolete(
+        "This property is deprecated and will be removed in future versions.  " +
+        "Use AuthenticationTimeout instead.")]
+    // Obsoleted with MDS 7.0.0; to be removed at least 2 major versions later.
```
cheenamalhotra: Also a breaking change, this will affect all customer applications. Please avoid all other breaking changes for max backwards compatibility.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/AADAuthenticationTests.cs
```diff
@@ -50,103 +50,52 @@ private void InvalidCombinationCheck(SqlCredential credential)
...
+        /// runtime.
+        ///
+        /// See the app.config file in the same directory as this file.
+        /// 
+        /// .NET (Core) reads similar configuration from appsettings.json, but
+        /// our SqlAuthenticationProviderManager does not currently support
+        /// that configuration source.
         /// </summary>
-        [Fact]
+        [ConditionalFact(typeof(TestUtility), nameof(TestUtility.IsNetFramework))]
```
paulmedynski, line 65: Prefer conditional execution to conditional compilation.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/AADAuthenticationTests.cs
```diff
@@ -50,103 +50,52 @@ private void InvalidCombinationCheck(SqlCredential credential)
...
         }
-#endif
-
-        // Verify that we can get and set providers via both the Abstractions
-        // package and Manager class interchangeably.
-        //
-        // This tests the dynamic assembly loading code in the Abstractions
-        // package.
-        [Fact]
-        public void Abstractions_And_Manager_GetSetProvider_Equivalent()
```
paulmedynski: I moved this to the UnitTests project for easy access to the Manager internals.

src/Microsoft.Data.SqlClient/tests/FunctionalTests/DataCommon/TestUtility.cs
```diff
@@ -9,9 +9,18 @@ namespace Microsoft.Data.SqlClient.Tests
     public static class TestUtility
     {
         public static readonly bool IsNotArmProcess = RuntimeInformation.ProcessArchitecture != Architecture.Arm;
-        public static bool IsNet => RuntimeInformation.FrameworkDescription == ".NET";
-        public static bool IsNetCore => RuntimeInformation.FrameworkDescription == ".NET Core";
-        public static bool IsNetFramework => RuntimeInformation.FrameworkDescription == ".NET Framework";
-        public static bool IsNetNative => RuntimeInformation.FrameworkDescription == ".NET Native";
+        public static bool IsNet
```
mdaigle, line 12: does this mean .NET Standard?

ErikEJ: Is it not UWP?

paulmedynski: https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.runtimeinformation.frameworkdescription

It's called ".NET Native" so that's what I stuck with.  It seems to support a .NET Framework compilation mode for desktop apps that produces native instructions, and thus runs without any JIT stuff.

https://learn.microsoft.com/en-us/windows/uwp/dotnet-native/net-native-and-compilation

mdaigle: Interesting! I didn't know about that.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
mdaigle: I'm mainly seeing differences in the inner exception, but it looks like the outer exceptions that are ultimately thrown should be the same.

mdaigle (approved):

paulmedynski:
Perhaps - all documentation activities are deferred to future PRs.

src/Microsoft.Data.SqlClient.Extensions/Azure/test/Azure.Test.csproj
```diff
@@ -0,0 +1,27 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net462;net8.0;net9.0</TargetFrameworks>
```
cheenamalhotra, line 4: Are you planning to use common variables for target frameworks as done in MDS tests in another PR?

src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider.csproj
```diff
@@ -33,15 +33,22 @@
...
+  <ItemGroup Condition="'$(ReferenceType)' == 'Project'">
+    <ProjectReference Condition="'$(TargetGroup)' == 'netcoreapp'" Include="$(NetCoreSource)src\Microsoft.Data.SqlClient.csproj" />
+    <ProjectReference Condition="'$(TargetGroup)' == 'netfx'" Include="$(NetFxSource)src\Microsoft.Data.SqlClient.csproj" />
+  </ItemGroup>
+
+  <!-- Conditional Package References -->
+  <ItemGroup Condition="'$(ReferenceType)' == 'Package'">
+    <PackageReference Include="Microsoft.Data.SqlClient" />
```
cheenamalhotra, line 45: I see you removed the packages.props in above directory.. What version will be picked up here?

cheenamalhotra (dismissed):
Minor comments, generally we can take it forward and address remaining work in future PRs.

src/Microsoft.Data.SqlClient.Extensions/Abstractions/src/Abstractions.csproj
```diff
@@ -23,26 +23,35 @@
   <!-- Build Config -->
   <PropertyGroup>
     <AssemblyName>Microsoft.Data.SqlClient.Extensions.Abstractions</AssemblyName>
-    <RootNamespace>Microsoft.Data.SqlClient.Extensions.Abstractions</RootNamespace>
+    <RootNamespace>$(AssemblyName)</RootNamespace>
```
benrr101: Please make sure that this is accurate for the file path and the namespaces of the files. _Some_ IDEs will complain if the namespaces don't match the path, and I agree with that preference. That's why a while back I made a change to remove the root namespace from MDS projects (since the paths replicate the full namespace, and contain things outside the MDS namespace).

src/Microsoft.Data.SqlClient.Extensions/Abstractions/src/SqlAuthenticationMethod.cs
```diff
@@ -0,0 +1,44 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.Data.SqlClient;
```
benrr101, line 5: Calling back to the root namespace comment, since this is in the Microsoft.Data.SqlClient namespace, this will raise a warning in _some_ IDEs. Since the root namesapce is Microsoft.Data.SqlClient.Extensions.Abstractions, the expected namespace for this class would be Microsoft.Data.SqlClient.Extensions.Abstractions. But, since this is a public API, we can't change the namespace without it being a breaking change. Soooo, I'd recommend making the root namespace either blank (and putting the files in a Microsoft.Data.SqlClient folder) or making the root namespace Microsoft.Data.SqlClient (and all abstraction files in an Extensions.Abstractions folder)

src/Microsoft.Data.SqlClient.Extensions/Abstractions/src/SqlAuthenticationMethod.cs
```diff
@@ -0,0 +1,44 @@
...
+    /// <include file='../doc/SqlAuthenticationMethod.xml' path='docs/members[@name="SqlAuthenticationMethod"]/NotSpecified/*'/>
+    NotSpecified = 0,
+
+    /// <include file='../doc/SqlAuthenticationMethod.xml' path='docs/members[@name="SqlAuthenticationMethod"]/SqlPassword/*'/>
+    SqlPassword,
+
+    /// <include file='../doc/SqlAuthenticationMethod.xml' path='docs/members[@name="SqlAuthenticationMethod"]/ActiveDirectoryPassword/*'/>
+    [Obsolete("ActiveDirectoryPassword is deprecated, use a more secure authentication method. See https://aka.ms/SqlClientEntraIDAuthentication for more details.")]
+    // Obsoleted with MDS 7.0.0; to be removed at least 2 major versions later.
+    ActiveDirectoryPassword,
```
benrr101, line 19: Introducing aliases for "Entra*" might be a good idea at ... some point.

src/Microsoft.Data.SqlClient.Extensions/Abstractions/src/SqlAuthenticationParameters.cs
```diff
@@ -0,0 +1,67 @@
...
+    public Guid ConnectionId { get; }
+
+    /// <include file='../doc/SqlAuthenticationParameters.xml' path='docs/members[@name="SqlAuthenticationParameters"]/ServerName/*'/>
+    public string ServerName { get; }
+
+    /// <include file='../doc/SqlAuthenticationParameters.xml' path='docs/members[@name="SqlAuthenticationParameters"]/DatabaseName/*'/>
+    public string DatabaseName { get; }
+
+    /// <include file='../doc/SqlAuthenticationParameters.xml' path='docs/members[@name="SqlAuthenticationParameters"]/ConnectionTimeout/*'/>
+    //
```
benrr101, line 35: This will cause some grumbling from some IDEs... but I can't think of a good alternative that doesn't inject our internal comments into public docs.

src/Microsoft.Data.SqlClient.Extensions/Abstractions/src/SqlAuthenticationParameters.cs
```diff
@@ -0,0 +1,67 @@
...
+    // We would like to deprecate this method in favour of a new
+    // AuthenticationTimeout property.  See:
+    //
+    // https://microsoft.sharepoint-df.com/:fl:/g/contentstorage/CSP_e68c6b62-34b4-4eaa-b836-82e9cdaa0149/IQCyPmTP5HlYSpafY3DJ-8sQAbY4Ajjn2ztRZrM_eQZkyJQ?e=k1nHJd&nav=cz0lMkZjb250ZW50c3RvcmFnZSUyRkNTUF9lNjhjNmI2Mi0zNGI0LTRlYWEtYjgzNi04MmU5Y2RhYTAxNDkmZD1iJTIxWW11TTVyUTBxazY0Tm9McHphb0JTYXhVNHFkaEY5ZE9yS0ZkWTR0cDY3WU5rRUhKaHM0R1JJTjhQanNwcGliSyZmPTAxWklZTVRaNVNIWlNNN1pEWkxCRkpOSDNET0RFN1hTWVEmYz0lMkYmYT1Mb29wQXBwJnA9JTQwZmx1aWR4JTJGbG9vcC1wYWdlLWNvbnRhaW5lciZ4PSU3QiUyMnclMjIlM0ElMjJUMFJUVUh4dGFXTnliM052Wm5RdWMyaGhjbVZ3YjJsdWRDMWtaaTVqYjIxOFlpRlpiWFZOTlhKUk1IRnJOalJPYjB4d2VtRnZRbE5oZUZVMGNXUm9SamxrVDNKTFJtUlpOSFJ3TmpkWlRtdEZTRXBvY3pSSFVrbE9PRkJxYzNCd2FXSkxmREF4V2tsWlRWUmFXbE5DTlVVMFJrMVFSemRhUlROWlV6Vk9SVkZDTmxkRE1rRSUzRCUyMiUyQyUyMmklMjIlM0ElMjI1YzA2ZTE4OS03NWExLTRkNDktYjQyYi1iOTk2YmM4MDc4ZjklMjIlN0Q%3D
+    //
+    public int ConnectionTimeout { get; }
+
+    /// <include file='../doc/SqlAuthenticationParameters.xml' path='docs/members[@name="SqlAuthenticationParameters"]/ctor/*'/>
+    public SqlAuthenticationParameters(
```
benrr101, line 44: Constructors _then_ properties :)

src/Microsoft.Data.SqlClient.Extensions/Abstractions/src/SqlAuthenticationProviderInternal.cs
```diff
@@ -0,0 +1,166 @@
...
+            }
+            // All of these exceptions mean we couldn't find the get/set
+            // methods.
+            catch (Exception ex)
+            when (ex is AmbiguousMatchException ||
+                  ex is BadImageFormatException ||
+                  ex is FileLoadException ||
+                  ex is FileNotFoundException)
+            {
+                // SqlClientEventSource.Log.TryTraceEvent(
```
benrr101, line 100: Intentionally left commented out?

src/Microsoft.Data.SqlClient.Extensions/Abstractions/src/SqlAuthenticationProviderInternal.cs
```diff
@@ -0,0 +1,166 @@
...
+                // TODO: Logging
+                // SqlClientEventSource.Log.TryTraceEvent(
+                //     nameof(SqlAuthenticationProviderManager) +
+                //     $": Azure extension class={className} installed as " +
+                //     "provider for all Active Directory authentication methods");
+            }
+            // All of these exceptions mean we couldn't find the get/set
+            // methods.
+            catch (Exception ex)
+            when (ex is AmbiguousMatchException ||
```
benrr101, line 94: this would fit nicely on one line if you use `is AmbiguoutMatchException or BadImageFormatException or FileLoadException or FileNotFoundException` :)

src/Microsoft.Data.SqlClient.Extensions/Abstractions/test/SqlAuthenticationProviderExceptionTest.cs
```diff
@@ -0,0 +1,121 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.Data.SqlClient.Extensions.Abstractions.Test;
+
+public class SqlAuthenticationProviderExceptionTest
+{
+    // Derive from SqlAuthenticationProviderException to test the abstract
+    // class' functionality.
+    private class Error : SqlAuthenticationProviderException
```
benrr101: nit: I think private classes are supposed to go at the end of a file 🤷‍♂️

src/Microsoft.Data.SqlClient.Extensions/Abstractions/test/SqlAuthenticationProviderExceptionTest.cs
```diff
@@ -0,0 +1,121 @@
...
+    {
+        public Error(
```
benrr101: nit: can we put this on one line, pls?

src/Microsoft.Data.SqlClient.Extensions/Abstractions/test/SqlAuthenticationProviderExceptionTest.cs
```diff
@@ -0,0 +1,121 @@
...
+        var retryPeriod = 42;
+        var message = "message";
+
+        Error ex = new(
+            method,
+            failureCode,
+            shouldRetry,
+            retryPeriod,
+            message,
+            null);
```
benrr101: Nit: for constants like `null`, it would be very nice to use the argument label

src/Microsoft.Data.SqlClient.Extensions/Abstractions/test/SqlAuthenticationProviderTest.cs
```diff
@@ -0,0 +1,57 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Reflection;
+
+namespace Microsoft.Data.SqlClient.Extensions.Abstractions.Test;
+
+// Tests for the obsolete SqlAuthenticationProvider.GetProvider and SetProvider
```
benrr101: If you want to add comment blocks to test files, I won't complain, but please use xmldocs

src/Microsoft.Data.SqlClient.Extensions/Abstractions/test/SqlAuthenticationProviderTest.cs
```diff
@@ -0,0 +1,57 @@
...
+            SqlAuthenticationParameters parameters)
+        {
+            throw new NotImplementedException();
+        }
+    }
+
+    // Test that GetProvider fails predictably when the MDS assembly can't be
+    // found.
+    [Fact]
+    public void GetProvider()
```
benrr101: I'm gonna sound like a broken, pedantic but this is why it's important to layout the stages of a test, and use descriptive names for the tests. At first glance, I'd assume this test verifies that a provider can be gotten successfully. And at second glance, I'd assume that there's actually two test cases in here.

So, I'd recommend rewriting as

```
[Fact]
public void GetProvider_MdsMissing_ReturnsNull()
{
   // Precondition
   // - Ensure MDS really doesn't exist in the working path
   Assert.Throws<FileNotFoundException>(() => Assembly.Load("Microsoft.Data.SqlClient"));

   // Act
   var result = SqlAuthenticationProvider.GetProvider(SqlAuthenticationMethod.ActiveDirectoryInteractive);

   // Assert
   Assert.Null(result);
}
```

And with these changes, we know at a glance:
* What the test does (without resorting to reading comments - or writing them!)
* Which part is a pre-condition of the test passing
* Which part is the action
* Which part is the validation

I know it's more total lines, but it really does help make it easier to figure out tests at-a-glance. And considering most tests will sit for a loooong time without being looked at, we want future generations to have the easiest time understanding unit tests as possible.

src/Microsoft.Data.SqlClient.Extensions/Azure/test/Azure.Test.csproj
```diff
@@ -0,0 +1,27 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
benrr101, line 1: I'd still personally prefer it if test projects go in the root /test/ rather than under the project. But, we can argue about that later.

src/Microsoft.Data.SqlClient.sln
```diff
@@ -327,6 +327,16 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "test", "test", "{59667E4C-0
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Abstractions.Test", "Microsoft.Data.SqlClient.Extensions\Abstractions\test\Abstractions.Test.csproj", "{04ACBF75-CFF2-41AB-B652-776BC0533490}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Azure", "Azure", "{A20114E1-82D8-903A-C389-726EB4FD943F}"
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{0D2F834B-6D91-18D0-3F09-672D448751BD}"
```
benrr101, line 332: oof, yeah we're gonna have to argue about this

benrr101 (changes requested):
Overall I'm only really complaining about the structure tests to make it clearer with less words. But I will accept things like "oh we'll fix those later"

src/Microsoft.Data.SqlClient.Extensions/Azure/test/Azure.Test.csproj
paulmedynski: That sounds like a great idea after the .NET and .NET Framework MDS projects merge, and we have the same set of target frameworks for all of our projects.

src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider.csproj
paulmedynski: MSBuild walks up the directory tree until it finds a `Directory.Packages.props` file, so this project will find this props file:

`src/Directory.Packages.props`

That's the same one everything else in the `src` directory uses.  It specifies the MDS package version as:
```
    <PackageVersion
      Include="Microsoft.Data.SqlClient"
      Version="$(MdsPackageVersion)" />
```

The `$(MdsPackageVersion)` is explicitly set by all pipelines, so it's either a CI/PR build number, or the official values found in YAML or ADO Libraries.  For command-line/IDE builds, we use a default value chosen by `Version.props`:
```
    <MdsPackageVersion Condition="'$(MdsPackageVersion)' == ''">$(MdsVersionDefault).$(BuildNumber)-dev</MdsPackageVersion>
```

src/Microsoft.Data.SqlClient.Extensions/Abstractions/src/SqlAuthenticationMethod.cs
paulmedynski: Let's defer project directory structure discussions/changes until after the MDS project merge.  We already have a mishmash of approaches and we need to do a full sweep anyway.

src/Microsoft.Data.SqlClient.Extensions/Abstractions/src/SqlAuthenticationMethod.cs
paulmedynski: Agreed, but not today :)

src/Microsoft.Data.SqlClient.Extensions/Abstractions/src/SqlAuthenticationParameters.cs
paulmedynski: Shrug - I can't imagine why an IDE would grumble about XML and normal comments appearing together.  They serve different purposes.

src/Microsoft.Data.SqlClient.Extensions/Abstractions/src/SqlAuthenticationParameters.cs
paulmedynski: I avoided style changes like this on purpose to make diffs to the originals sane.

src/Microsoft.Data.SqlClient.Extensions/Abstractions/src/SqlAuthenticationProviderInternal.cs
paulmedynski: Missed the TODO comment here; will add.

src/Microsoft.Data.SqlClient.Extensions/Abstractions/src/SqlAuthenticationProviderInternal.cs
paulmedynski: I always regret putting lists of things like this on a single line.  At some point you add one more and then it's "too long" and you break it up anyway.  Also, future diffs where you add/remove/change one are harder to read.

src/Microsoft.Data.SqlClient.Extensions/Abstractions/test/SqlAuthenticationProviderTest.cs
paulmedynski: I'll move the precondition out into a constructor.

IMO, introducing variables just to follow a pattern is a recipe for a botched test.  There are certainly conditions where you need to do a few things and then check state or results.  This isn't one of those cases though.

src/Microsoft.Data.SqlClient.Extensions/Azure/test/Azure.Test.csproj
paulmedynski: We definitely need a consistent approach - TBD!

src/Microsoft.Data.SqlClient.sln
paulmedynski: That's the result of:
```
$ cd src
$ dotnet sln add Microsoft.Data.SqlClient.Extensions/Azure/src/Azure.csproj
$ dotnet sln add Microsoft.Data.SqlClient.Extensions/Azure/test/Azure.Test.csproj
```
I assume `dotnet` knows what it's doing.

apoorvdeshmukh (approved):

PR #3681, opened 2025-10-10
Switch arm64 jobs to westus2 pool

mdaigle:
## Description

The eus region is currently capacity constrained for arm64 vms. This PR moves arm64 jobs to 1ES hosted pool in the westus2 region.


mdaigle:
Ended up configuring backup pools with auto failover, so this change is not needed.

PR #3688, opened 2025-10-14
Remove audit sources from NuGet.config

akoeplinger:
This is no longer necessary now that AzDO provides the same data and it will make it easier to comply with the CFSClean network isolation policy required by 1ES.

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -4,8 +4,4 @@
     <clear />
     <add key="sqlclient" value="https://sqlclientdrivers.pkgs.visualstudio.com/public/_packaging/sqlclient/nuget/v3/index.json" />
   </packageSources>
-  <auditSources>
-    <clear />
-    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
-  </auditSources>
 </configuration>


paulmedynski:
@benrr101 - Do you remember if/how we verify that audits are still taking place?

cheenamalhotra:
> @benrr101 - Do you remember if/how we verify that audits are still taking place?

They happen in Official builds only and are disabled with `NuGetAudit` property everywhere else.

paulmedynski:
It looks like the opposite, from `src/Directory.Build.props`:

```
  <!-- NuGet Audit Settings -->
  <PropertyGroup Condition="'$(TF_BUILD)' == 'true'">
    <!--
      ADO does not support audit/vulnerability feeds, so the audit feed is specified (in
      nuget.config) as nuget.org. OneBranch default network isolation does not allow connections
      to nuget.org. To avoid this issue, we will disable auditing for official builds, but leave it
      enabled for local builds.
      @TODO: If/when auditing is enabled for central feeds services, this can be removed. 
    -->
    <NuGetAudit>false</NuGetAudit>
  </PropertyGroup>
```

The `TF_BUILD` environment variable is set to `true` for all Azure Pipelines runs, so we're explicitly disabling NuGet auditing everywhere except developer console builds currently.

There are several other instances of `<NuGetAudit>` in our project/props files.  I think they will all need to be explicitly  _enabled_ if we want auditing to occur everywhere.  I would suggest we put this wherever it needs to be for all code that we build (driver, tests, tools, etc):

```
<PropertyGroup>
  <NuGetAudit>true</NuGetAudit>
  <NuGetAuditMode>all</NuGetAuditMode>
  <NuGetAuditLevel>low</NuGetAuditLevel>
</PropertyGroup>
```

@akoeplinger - Please make these changes and our CI should verify that auditing is occurring and passing.

akoeplinger:
@paulmedynski we actually just hit an issue with the new AzDO vulnerability feed support which you can turn on via this option (and which you have turned on in your org already):
image

AzDO currently requires auth to access the vulnerability endpoint, even for public feeds. That means with this change you won't be able to restore unless you have nuget credential manager configured with a valid AzDO cred.

We have a discussion with AzDO folks about this but I think you should not merge this PR for now until this is fixed. Given your settings in src/Directory.Build.props you won't be hitting the CFS issue anyway in official builds (I just grepped across the whole dotnet org so missed that you already handled this)

PR #3689, opened 2025-10-14
[7.0.0-preview2] Version Bump

benrr101:
## Description
Bumping preview version in the MDS official pipeline.

## Testing
Just a one character change that affects pipeline.


benrr101:
Bypassing rules since this is just a change to the release pipeline

PR #3690, opened 2025-10-14
Change expected newline char when parsing crypto test baseline file.

mdaigle:
## Description
CoreCryptoTests fail when run on linux because line endings are normalized to LF when the repo is checked out. The tests are otherwise still applicable to linux and can pass when the "baseline" files containing the keypairs are properly parsed.

This PR:
- Adjusts keypair files to use LF in place of CRLF.
  - This doesn't show in the diff, unfortunately.
- Adjust file parsing to look for LF.

Successful run: https://sqlclientdrivers.visualstudio.com/ADO.Net/_build/results?buildId=127269&view=results

### DIFF
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/CoreCryptoTests.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/CoreCryptoTests.cs
@@ -9,10 +9,9 @@ using Xunit;
 
 namespace Microsoft.Data.SqlClient.ManualTesting.Tests.AlwaysEncrypted
 {
-    public class CoreCryptoTests : IClassFixture<SQLSetupStrategyCertStoreProvider>
+    public class CoreCryptoTests
     {
-        // Synapse: Always Encrypted not supported in Azure Synapse.
-        [ConditionalFact(typeof(DataTestUtility), nameof(DataTestUtility.AreConnStringsSetup), nameof(DataTestUtility.IsNotAzureSynapse))]
+        [Fact]
         public void TestAeadCryptoWithNativeBaseline()
         {
             // Initialize the reader for resource text file which has the native code generated baseline.
@@ -45,8 +44,7 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests.AlwaysEncrypted
             }
         }
 
-        // Synapse: Always Encrypted not supported in Azure Synapse.
-        [ConditionalFact(typeof(DataTestUtility), nameof(DataTestUtility.AreConnStringsSetup), nameof(DataTestUtility.IsNotAzureSynapse))]
+        [Fact]
         public void TestRsaCryptoWithNativeBaseline()
         {
             // Initialize the reader for resource text file which has the native code generated baseline.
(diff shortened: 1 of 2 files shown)


src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/TestFixtures/Setup/CryptoNativeBaselineReader.cs
```diff
@@ -174,7 +174,7 @@ internal class CryptoNativeBaselineReader
         /// <summary>
         /// Parameter Data End identifier, to search inside resource text file.
         /// </summary>
-        private static readonly Regex regexEndParameterIdentifier = new Regex(@"\r");
```
paulmedynski, line 177: `\r` isn't a valid line ending on any OS we support, so I'm not sure what this was trying to do.

git should be using LF as the text file line ending everywhere except Windows, where it should use CRLF.

Should we be using [Environment.NewLine](https://learn.microsoft.com/en-us/dotnet/api/system.environment.newline?view=net-9.0#system-environment-newline) instead of a regex?

paulmedynski (dismissed):

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/TestFixtures/Setup/CryptoNativeBaselineReader.cs
mdaigle: I made a ticket to convert these to unit tests. I'll add a note there to look into this change as well.

paulmedynski: Agreed - we have some housecleaning to do related to our text file encodings, line endings, etc.

mdaigle: I may need to make this change now... Expecting \n just flips the problem over to windows 🤦

src/Microsoft.Data.SqlClient/tests/ManualTests/AlwaysEncrypted/CoreCryptoTests.cs
```diff
@@ -9,10 +9,9 @@
 
 namespace Microsoft.Data.SqlClient.ManualTesting.Tests.AlwaysEncrypted
 {
-    public class CoreCryptoTests : IClassFixture<SQLSetupStrategyCertStoreProvider>
+    public class CoreCryptoTests
     {
-        // Synapse: Always Encrypted not supported in Azure Synapse.
-        [ConditionalFact(typeof(DataTestUtility), nameof(DataTestUtility.AreConnStringsSetup), nameof(DataTestUtility.IsNotAzureSynapse))]
```
cheenamalhotra, line 15: When and if we setup testing against Synapse / DW instance, wouldn't this test fail?

cheenamalhotra: If it doesn't change the outcome of Kerberos tests, I would avoid that change here.

mdaigle: These tests don’t actually talk to a server. The comment and class fixture were incorrectly included. What’s worse, the combination of conditionals meant that they actually never ran in our ci pipeline, only in the Kerberos pipeline, even though they’re not Kerberos specific.

cheenamalhotra: Ok, I guess we need to configure the Kerberos pipeline to test only Kerberos stuff in that case :)

cheenamalhotra:
I still see failures in Windows jobs in the linked build, is that expected?

priyankatiwari08 (approved):

PR #3705, opened 2025-10-20
Cleanup | AsyncHelpers Generic State

benrr101:
## Description
Hello all, in this latest installment, I'm rewriting majority of the AsyncHelper ContinueTask and CreateContinuationTask methods to have generic state. One thing that has bothered me a lot throughout the codebase is that many (if not all) of the ContinueTask or CreateContinuationTask callbacks have lines that decode the state object(s) passed into them. Something like this:

```C#
AsyncHelper.CreateContinuationTaskWithState(
    myTask,
    this,
    (object state) => {
        SqlCommand sqlCommand = (SqlCommand)state;
        sqlCommand.DoSomething();
    });
```

Not only is this challenging to read, it poses the potential for bugs where the type the state object is cast to could be anything, valid or invalid.

Instead, I've taken the liberty to:
1) Rewrite these methods with up to two generics for the stateful methods. This allows full IDE assistance when writing callbacks, improving readability, as well as helping eliminate runtime bugs from incorrect type casts.
2) Clean up the methods to follow a pattern
   1) Records for encapsulating the state object(s) internally to make all task continuations static lambdas
   2) Remove holes in the implemented methods - each method handles the entire continuation (please note that this *does* significantly increase redundant code, but I argue that this is acceptable because chaining the methods together introduces additional layers of call unwrapping and state wrapping, which decreases performance and makes debugging a nightmare - we should aim to remove these methods anyhow and move to async/await natives wherever possible)
3) Removed the "exception converter" from the methods - this was being used around two times and could be handled better by setting the converted exception in the onFailure handler.
4) Clean up calls to these methods
    1) If the method could be moved to using the stateful version, it has been
    2) If the method used a Tuple<x, y> before, it's been decomposed to use the two-generic overload
    3) There are still a handful that could not be migrated to static lambdas due to their excessive amount of closures being used.
5) Add a large number of **unit tests** that test just about every scenario that could happen in the async handler.
    1) This introduces a dependency on Moq

You can now safely rewrite the above example as (generic isn't necessary since it's obvious via object types):
```C#
AsyncHelper.CreateContinuationTaskWithState(
    taskToContinue: myTask,
    state: this,
    onSuccess: this2 => this2.DoSomething());
```

## Issues
N/A

## Testing
Introduced many unit tests for the new code, existing code can be validated via CI.


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.netcore.cs
```diff
@@ -1334,41 +1334,45 @@ private SqlDataReader GetParameterEncryptionDataReader(out Task returnTask, Task
             SqlDataReader describeParameterEncryptionDataReader,
             ReadOnlyDictionary<_SqlRPC, _SqlRPC> describeParameterEncryptionRpcOriginalRpcMap, bool describeParameterEncryptionNeeded, bool isRetry)
         {
-            returnTask = AsyncHelper.CreateContinuationTaskWithState(fetchInputParameterEncryptionInfoTask, this,
-                (object state) =>
+            // @TODO: Why use the with state version if we can't make the lambda static?
+            returnTask = AsyncHelper.CreateContinuationTaskWithState(
+                taskToContinue: fetchInputParameterEncryptionInfoTask,
+                state: this,
+                onSuccess: this2 =>
```
paulmedynski: I prefer to explicitly declare the argument types for lambdas.  It's hard to tell what type `this2` is:
```
onSuccess: (SqlCommand thisCommand) =>
{
  ...
  thisCommand.CheckThrowSNIException();
}
```

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Reader.cs
```diff
@@ -308,14 +308,12 @@ private IAsyncResult BeginExecuteReaderInternal(
...
-                        onSuccess: static state =>
-                        {
-                            var parameters = (Tuple<SqlCommand, TaskCompletionSource<object>>)state;
-                            parameters.Item1.BeginExecuteReaderInternalReadStage(parameters.Item2);
-                        });
+                        taskToContinue: writeTask,
+                        taskCompletionSource: localCompletion,
+                        state1: this,
+                        state2: localCompletion,
+                        onSuccess: static (this2, localCompletion2) =>
```
paulmedynski, line 318: Ahh yes, C#'s refusal to allow variable masking, so now we get `localCompletion` and `localCompletion2` - much clearer!

Not ragging on @benrr101 - there isn't really any naturally good name for the 2nd variable other than `localCompletion`, but it's not allowed :(

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Reader.cs
```diff
@@ -1582,21 +1580,19 @@ private Task RunExecuteReaderTdsSetupContinuation(
...
         {
-            // @TODO: Why use the state version if we can't make this a static helper?
             return AsyncHelper.CreateContinuationTaskWithState(
-                task: writeTask,
-                state: _activeConnection,
-                onSuccess: state =>
+                taskToContinue: writeTask,
+                state1: this,
+                state2: Tuple.Create(ds, runBehavior, optionSettings),
+                onSuccess: static (this2, parameters) =>
```
paulmedynski, line 1611: It seems weird to have a static method with a `this2` variable.  Static things don't have this.  Would this be clearer:
```
onSuccess: static (command, parameters) =>
```
Even better with argument types?  It isn't obvious which tuple items to pass as which arguments to `SetAsyncReaderState()` below.  Maybe `SetAsyncReaderState()` should take a struct/class/record and then you'd have a nice type to use for `state2` above rather than a tuple.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Reader.cs
```diff
@@ -1600,21 +1598,19 @@ private Task RunExecuteReaderTdsSetupContinuation(
...
-                    CachedAsyncState.SetAsyncReaderState(ds, runBehavior, optionSettings);
+                    this2._activeConnection.GetOpenTdsConnection();
+                    this2.CachedAsyncState.SetAsyncReaderState(parameters.Item1, parameters.Item2, parameters.Item3);
                 },
-                onFailure: static (exception, state) =>
-                {
-                    ((SqlConnection)state).GetOpenTdsConnection().DecrementAsyncCount();
-                });
+                onFailure: static (this2, _, exception) =>
+                    this2._activeConnection.GetOpenTdsConnection().DecrementAsyncCount());
```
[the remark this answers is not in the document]
paulmedynski: You can use a `_` discard for `exception` if you aren't going to use it:
```
onFailure: static (connection, _, _) => connection...();
```

src/Microsoft.Data.SqlClient/tests/UnitTests/Utilities/MockExtensions.cs
paulmedynski, line 1: Can you walk me/us through these extensions?  I'd like to understand why they're necessary and why we need a new dependency on Moq.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
```diff
@@ -2362,15 +2364,15 @@ private Task CopyColumnsAsync(int col, TaskCompletionSource<object> source = nul
...
-                source,
+                taskToContinue: task,
+                taskCompletionSource: source,
                 state: this,
-                onSuccess: (object state) =>
+                onSuccess: this2 =>
                 {
-                    SqlBulkCopy sqlBulkCopy = (SqlBulkCopy)state;
-                    if (i + 1 < sqlBulkCopy._sortedColumnMappings.Count)
+                    if (i + 1 < this2._sortedColumnMappings.Count)
```
paulmedynski: Interesting... we're willing to capture `i`, but not `this`?  Could this be:
```
state1: this,
state2: i,
onSuccess: static (connection, count) =>
{
  if (count + 1 < connection._sortedColumnMappings.Count)
   ...
}
```

Hmm, maybe not because of the use of `source` below?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
```diff
@@ -2505,53 +2507,53 @@ private Task CopyRowsAsync(int rowsSoFar, int totalRows, CancellationToken cts,
...
                                 {
                                     AsyncHelper.ContinueTaskWithState(
-                                        readTask,
-                                        source,
-                                        state: sqlBulkCopy,
-                                        onSuccess: (object state2) => ((SqlBulkCopy)state2).CopyRowsAsync(i + 1, totalRows, cts, source));
+                                        taskToContinue: readTask,
+                                        taskCompletionSource: source,
+                                        state: this2,
+                                        onSuccess: this3 => this3.CopyRowsAsync(i + 1, totalRows, cts, source));
```
paulmedynski, line 2646: Yay `this3`  :)

paulmedynski (commented):
Looked at everything except `AsyncHelper.cs` for now.  Maybe a quick call to explain this a bit more for me (and the team?) would help.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlCommand.netcore.cs
benrr101: I guess I prefer not to since it tidies things up in many cases to one line lambdas. (eg `state => state.Qqq()`)
And the way I look at it, calling it `this2` makes it clear it is `this` just inside the scope.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Reader.cs
benrr101: It think it makes sense - if it was allowed, it becomes more confusing whether it introduces a closure.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Reader.cs
benrr101: Hmm... I disagree. Previously it was using `this` via a closure, so now that we aren't using a closure, it should be obvious `this2` is just `this` inside the closure. Using `command` makes it less obvious that it's `this` command, to me at least.

Don't disagree with introducing a record or something for the params. Just that's outside the scope of this PR...
Honestly, rather than try to continue to improve these helpers, my goal is to get rid of them and use native async/await for task continuation. But, this PR at least addresses the inelegance of the existing helpers and clears up some confusion for further refactoring attempts.

src/Microsoft.Data.SqlClient/tests/UnitTests/Utilities/MockExtensions.cs
benrr101: I'm surprised we didn't have a mocking framework already 🤯 But then again we didn't really have actual unit tests before

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
benrr101: Doubly nested task completion via TaskCompletionSource objects whoo 🙃

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
benrr101: Yeah, I think this one was because it was using `i` and `source` and I didn't want to make a three-generic version of the helper without a really dang good reason. 

I can still make this a static lambda, it just gets a bit more clunky. See what you think of this version.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Reader.cs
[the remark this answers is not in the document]
benrr101: You are right!

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Utilities/AsyncHelper.cs
```diff
@@ -0,0 +1,519 @@
...
+
+using System;
+using System.Diagnostics;
+using System.Runtime.ExceptionServices;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Data.SqlClient.Utilities
+{
+    internal static class AsyncHelper
```
paulmedynski, line 26: What does this class do?  Some docs here and on the methods would be super!  Same for the test file.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Utilities/AsyncHelper.cs
```diff
@@ -0,0 +1,519 @@
...
+                    else
+                    {
+                        try
+                        {
+                            onSuccess();
+                        }
+                        // @TODO: CER Exception Handling was removed here (see GH#3581)
+                        catch (Exception e)
+                        {
+                            taskCompletionSource.SetException(e);
```
paulmedynski: Why are we using `SetException()` here and `TrySetException()` above?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Utilities/AsyncHelper.cs
```diff
@@ -0,0 +1,519 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Diagnostics;
+using System.Runtime.ExceptionServices;
+using System.Threading;
+using System.Threading.Tasks;
+
```
paulmedynski, line 10: New code should use `#nullable enable`.  Same for the test file.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Utilities/AsyncHelper.cs
```diff
@@ -0,0 +1,519 @@
...
+        internal static Task CreateContinuationTask(
+            Task taskToContinue,
+            Action onSuccess,
+            Action<Exception> onFailure = null,
+            Action onCancellation = null)
+        {
+            if (taskToContinue is null)
+            {
+                onSuccess();
```
paulmedynski, line 350: Interesting choice - if no task to continue was supplied, we consider it to have succeeded, even though it doesn't exist?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Utilities/AsyncHelper.cs
```diff
@@ -0,0 +1,519 @@
...
+
+        internal static void WaitForCompletion(
+            Task task,
+            int timeoutInSeconds,
+            Action onTimeout = null,
+            bool rethrowExceptions = true)
+        {
+            try
+            {
+                task.Wait(timeoutInSeconds > 0 ? 1000 * timeoutInSeconds : Timeout.Infinite);
```
paulmedynski: Need to protect against overflow here:
```
int timeout = Timeout.Infinite;
if (timeoutInSeconds > 0 && timeoutInSeconds < (int.MaxValue / 1000))
{
    timeout = timeoutInSeconds * 1000;
}
```

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft/Data/SqlClient/Utilities/AsyncHelperTest.cs
```diff
@@ -0,0 +1,1291 @@
...
+    public class AsyncHelperTest
+    {
+        #region ContinueTask
+
+        [Fact]
+        public async Task ContinueTask_TaskCompletes()
+        {
+            // Arrange
+            Task taskToContinue = Task.CompletedTask;
+            TaskCompletionSource<object> taskCompletionSource = new();
```
paulmedynski: Can you specialize with `int` here instead of `object`, since your result is an integer (one line 34)?

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft/Data/SqlClient/Utilities/AsyncHelperTest.cs
```diff
@@ -0,0 +1,1291 @@
...
+            using CancellationTokenSource cts = new();
+            cts.Cancel();
+
+            return Task.FromCanceled(cts.Token);
+        }
+
+        private static async Task RunWithTimeout(Task taskToRun, TimeSpan timeout)
+        {
+            Task winner = await Task.WhenAny(taskToRun, Task.Delay(timeout));
+            if (winner != taskToRun)
```
paulmedynski, line 1344: The `Task` class isn't sealed, so it's possible that `taskToRun` is a derived instance that has overridden the `!=` operator to do something other than reference equality.  It's safer here to use `Object.ReferenceEquals(winner, taskToRun)`.

paulmedynski (changes requested):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlBulkCopy.cs
mdaigle: 😞

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Utilities/AsyncHelper.cs
```diff
@@ -0,0 +1,519 @@
...
+                // Ensure the task does not leave an unobserved exception
+                task.ContinueWith(static t => { _ = t.Exception; });
+                onTimeout?.Invoke();
+            }
+        }
+
+        private static void ObserveContinuationException(Task continuationTask)
+        {
+            continuationTask.ContinueWith(
+                static task => _ = task.Exception,
```
mdaigle: This makes me feel a bit uneasy. We're swallowing the exception to avoid raising an event: https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.taskscheduler.unobservedtaskexception?view=net-9.0

Why not just keep raising the event until our code correctly handles exceptions? It's a legitimate issue and the exception is still unhandled even if we observe it.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Utilities/AsyncHelper.cs
```diff
@@ -0,0 +1,519 @@
...
+            ContinuationState<TState1, TState2> continuationState = new(
+                OnCancellation: onCancellation,
+                OnFailure: onFailure,
+                OnSuccess: onSuccess,
+                State1: state1,
+                State2: state2,
+                TaskCompletionSource: taskCompletionSource);
+
+            Task continuationTask = taskToContinue.ContinueWith(
+                static (task, state2) =>
```
mdaigle: `state2` collides with the method parameter name, which makes this confusing to read

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Utilities/AsyncHelper.cs
mdaigle: Can use TimeSpan instead.

mdaigle (changes requested):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Utilities/AsyncHelper.cs
benrr101: Alrighty, you asked for it - you get comments :)

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Utilities/AsyncHelper.cs
benrr101: ... because that's how it was originally written
I think the reason for that was the expectation that the onSuccess handler would never set an exception.

But I do agree, it's likely safer to try to set the exception.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Utilities/AsyncHelper.cs
benrr101: Alrighty... this complicates things a bit, but it'll probably be ok.

Test file has it enabled in the project file

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Utilities/AsyncHelper.cs
benrr101: This is a remnant of ye olde async/sync code that returns null tasks when executing synchronously. So it's still desired that the onSuccess executes regardless of whether the preceding action was synchronous or asynchronous.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Utilities/AsyncHelper.cs
benrr101: Yep, I wanted to use TimeSpan originally but didn't see the Timeout.InfiniteTimeSpan option.

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft/Data/SqlClient/Utilities/AsyncHelperTest.cs
benrr101: No ... not without introducing a generic for the type of the task completion object. I actually tried introducing this in a previous attempt at these code changes but ended up not doing it because it complicated things even more - and I got hung up on the generic-less task completion source and where those were feasible to use versus not.

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft/Data/SqlClient/Utilities/AsyncHelperTest.cs
benrr101: While technically true, we're never using custom task objects in these tests.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Utilities/AsyncHelper.cs
benrr101: So the original implementation didn't observe the exception at all, except for in the WaitForCompletion. The one unit test for that method was brought over from the functional tests to make sure that the exception was observed. When I brought that test over, parallel execution of the tests (that all contained unobserved exceptions) caused the test to fail (because the default scheduler's unobserved exception handler was capturing the exceptions from the other tests). So that prompted fixing the issue by ensuring the exceptions are observed in the helper.

We even have an open issue related to unobserved exceptions (#3720), and this would potentially resolve that without breaking the behavior as it currently is. The issue with leaving the exceptions unobserved is that they also go silently unobserved unless a user adds an unobserved handler - it'll clog up their application, necessitate they report it to us, and likely won't capture the scenario that actually caused it.

I'm open to alternatives - I agree masking the issue isn't great... but outside of properly rewriting the code to observe exceptions properly (which would probably just cause us to give up and rewrite it as async/await properly), I'm not sure how to do this properly. I think the most feasible improvement would be to add logging to the observation code, but I'm not sure the _right_ way to do that.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Utilities/AsyncHelper.cs
benrr101: Fair enough - ok I think I've got it sorted now

benrr101:
@paulmedynski  @mdaigle I think I addressed the comments as best as I could without making major major changes. Let me know what you think

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Utilities/AsyncHelper.cs
mdaigle: At least right now there's some evidence of an exception, even if it's hard to do anything with it. I think swallowing it completely could be bad for trust because a user could experience unexpected behavior and see nothing reported. My preference is to leave it unhandled and leave that issue open until we can improve the async code. Users can and should report the issue to us, which will help us determine impact and prioritization.

I don't think there's any way to improve this code as is to handle the exceptions properly. Because we run a lot of stuff on continuations in the background, I can't think of a good way to pass an exception out to the original call site that originated the first async operation.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Utilities/AsyncHelper.cs
paulmedynski: Looks great!

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft/Data/SqlClient/Utilities/AsyncHelperTest.cs
paulmedynski: Okey dokey

paulmedynski (dismissed):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Utilities/AsyncHelper.cs
benrr101: I agree, getting some evidence of the exception is good. I also think it's bad practice to intentionally leave the exceptions unobserved as this can leak resources. The event isn't triggered until the GC runs, so it's also very unpredictable, meaning the exceptions will be completely context-less. I think a good middle ground would be to observe the exception but add logging to it. By logging it as it happens, we can capture more of the context, as well as give us an opportunity to debug. Users can still see the issue happen, not quite as loudly, but also not quite as invasively. 

One other point worth noting, if we left the exceptions unobserved and waited on users to report them ... how actionable would they be? We already know there's an issue with unobserved exceptions, and we know we can't fix how we handle it without changing large swaths of code. I think if we want to act on the exceptions that are not being handled, we want as much context to the exception as possible, so relying on the unobserved event to fire isn't as good as logging it when it happens.

@paulmedynski do you have any thoughts on this?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Utilities/AsyncHelper.cs
paulmedynski: This would make a great comment in the code :)

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Utilities/AsyncHelper.cs
paulmedynski: Ok - can you make that change throughout?

benrr101: Haha, whoops, I guess I thought I did

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Utilities/AsyncHelper.cs
benrr101: Sure thing

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Utilities/AsyncHelper.cs
```diff
@@ -297,7 +297,7 @@ internal static void ContinueTaskWithState<TState1, TState2>(
                         }
                         catch (Exception e)
                         {
-                            typedState2.TaskCompletionSource.SetException(e);
+                            typedState2.TaskCompletionSource.TrySetException(e);
```
paulmedynski, line 302: These new TrySetXyz() calls will change the exceptions we could throw (fewer now).  Does that change the behaviour of any of our callers?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Utilities/AsyncHelper.cs
paulmedynski: Logging with this thread's stack context, along with whatever the exception itself logs, seems like a good idea.  Then the app has a chance of figuring out what happened, when, where, and why.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Utilities/AsyncHelper.cs
benrr101: It'd take a while to run an analysis of all usages, but off the top of my head, I don't think so.
The exceptions would be thrown from trying to set* after it's already been set*. The only possible place it could be set would be in the callbacks. I can run through the usages and see if they're setting them in places where we were forcefully setting before. (Granted the usages of the passed in task completion objects open us up to thread un-safety, but those are high quality bugs and not features :) )

paulmedynski (dismissed):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Utilities/AsyncHelper.cs
mdaigle: I still don't love it and would prefer to just let the event bubble up, but I won't block if you're both ok with this approach. It's easy to undo if we change our minds later.

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft/Data/SqlClient/Utilities/AsyncHelperTest.cs
```diff
@@ -14,67 +15,78 @@ namespace Microsoft.Data.SqlClient.UnitTests.Microsoft.Data.SqlClient.Utilities
...
+            await RunWithTimeout(taskCompletionSource.Task, RunTimeout);
 
             // Assert
             Assert.Equal(TaskStatus.Faulted, taskCompletionSource.Task.Status);
-            onSuccess.Verify(action => action(), Times.Once);
-            onFailure.VerifyNeverCalled();
-            onCancellation.VerifyNeverCalled();
+            mockOnSuccess.Verify(action => action(), Times.Once);
+            mockOnFailure.VerifyNeverCalled();
+            mockOnFailure.VerifyNeverCalled();
```
mdaigle:
```suggestion
            mockOnCancellation.VerifyNeverCalled();
```

mdaigle (dismissed):

paulmedynski (dismissed):

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft/Data/SqlClient/Utilities/AsyncHelperTest.cs
benrr101: This would be worth requesting changes over!

paulmedynski (dismissed):

mdaigle (dismissed):

paulmedynski (dismissed):

paulmedynski (dismissed):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Utilities/AsyncHelper.cs
```diff
@@ -0,0 +1,796 @@
...
+        /// <param name="timeoutInSeconds">Number of seconds to wait until timing out the task</param>
+        /// <param name="onTimeout">
+        /// Callback to execute when the task does not complete within the allotted time. The
+        /// exception returned by the callback is set on the <paramref name="taskCompletionSource"/>.
+        /// </param>
+        /// <param name="cancellationToken">Cancellation token to prematurely cancel timeout</param>
+        internal static void SetTimeoutException(
+            TaskCompletionSource<object> taskCompletionSource,
+            int timeoutInSeconds,
+            Func<Exception> onTimeout,
```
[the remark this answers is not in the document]
benrr101: The callback is called when the initial task completion source fails to complete within the given timeout period. Thus, it's what happens "onTimeout". The callback returns an exception and that exception is set on the initial task completion source. The use cases for this helper are very slim hence the oddball behavior.

paulmedynski (dismissed):

benrr101:
@github-actions Hey! I'm not dead yet!

paulmedynski (dismissed):

mdaigle (dismissed):

paulmedynski (approved):

paulmedynski:
@benrr101 - These EventSource failures appear to be related to these changes:
```
    Failed Microsoft.Data.SqlClient.ManualTesting.Tests.EventSourceTest.EventSourceTestAll [37 ms]
##[error]EXEC(0,0): Error Message: 
EXEC : error Message:  [D:\a\_work\1\s\build2.proj]
     Assert.Contains() Failure: Item not found in set
  Set:       [0, 1, 2, 3, 4, ┬╖┬╖┬╖]
  Not found: 21
    Stack Trace:
       at Microsoft.Data.SqlClient.ManualTesting.Tests.EventSourceTest.EventSourceTestAll() in D:\a\_work\1\s\src\Microsoft.Data.SqlClient\tests\ManualTests\TracingTests\EventSourceTest.cs:line 36
     at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
     at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
```

apoorvdeshmukh (commented):
@benrr101 Can you check the mentioned test failures?

benrr101:
Closed and replaced by the much smaller #4334

PR #3709, opened 2025-10-21
Tests | Refactor and move CoreCryptoTests to UnitTests

edwardneal:
## Description

This picks up from a comment yesterday in #3660 from @mdaigle, moving CoreCryptoTests into the unit tests project.

The CEK-based tests (`TestRsaCryptoWithNativeBaseline`) had already removed the use of reflection, so I've ported those across. There were also AEAD-based tests which still used reflection, so I removed those too.

The old version of the tests contained the `TCECryptoNativeBaseline.txt` and `TCECryptoNativeBaselineRsa.txt` files, with various parsing logic to convert the hexadecimal strings into byte arrays. I've turned these byte arrays into embedded resources because several of them are about 4KB in size and I wanted to load them in a consistent way. I'm not completely happy with the approach though, it leads to a lot of small files. I did consider putting them in one or more JSON files (so we'd eliminate the use of the custom parsing logic without adding so many files - perhaps one file for each of the 32 AEAD-based test cases and the 3 CEK-based test cases) but it seemed like an unnecessary layer of indirection. I'm happy to go with whichever option you'd prefer.

It's also worth noting that this lifts a hardcoded certificate out of `TCECryptoNativeBaselineRsa.txt`. This was always there, but it's technically a hardcoded credential and might be noticed as such. We use this to decrypt a CEK and verify its contents against the SQL Server native code, so it needs to remain in situ.

## Issues

Follows up #3660 [comment](https://github.com/dotnet/SqlClient/pull/3660#discussion_r2437565253).

## Testing

New tests run successfully.

### DIFF
--- a/src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft/Data/SqlClient/AlwaysEncrypted/NativeColumnEncryptionKeyBaseline.cs
+++ b/src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft/Data/SqlClient/AlwaysEncrypted/NativeColumnEncryptionKeyBaseline.cs
@@ -11,6 +11,8 @@ namespace Microsoft.Data.SqlClient.UnitTests.AlwaysEncrypted;
 /// Provides unit tests which verify that a final cell produced by SQL Server's native Always Encrypted code
 /// can be decrypted to a known value.
 /// </summary>
+[SkipOnPlatform(TestPlatforms.OSX | TestPlatforms.MacCatalyst | TestPlatforms.iOS | TestPlatforms.tvOS,
+    "The supplied native column encryption key is a certificate which is incompatible with MacOS.")]
 public class NativeColumnEncryptionKeyBaseline : IClassFixture<NativeColumnEncryptionKeyCertificateBaselineFixture>
 {
     private readonly string _certificateThumbprint;


src/Microsoft.Data.SqlClient/tests/UnitTests/Resources.Designer.cs
paulmedynski, line 1: Why do we check in these files (`Resources.Designer.cs` and `Resources.resx`)?  Aren't they generated by the build?

paulmedynski:
@edwardneal - Build failures - see:  https://sqlclientdrivers.visualstudio.com/public/_build/results?buildId=127984&view=logs&j=4a2ec0b0-f4e2-5686-f3b1-9a2c684c31cc&t=047acb07-0880-5772-afc3-fa5d1648663c

edwardneal:
Thanks @paulmedynski - I'd missed the extracted CEK certificate. I've just added it and pushed.

edwardneal:
The NativeColumnEncryptionKeyBaseline class' test was failing to run on Mac because the baseline CEK certificate couldn't be loaded by the OS. I've skipped the class for now, could someone re-run CI please?

src/Microsoft.Data.SqlClient/tests/UnitTests/Resources.Designer.cs
edwardneal: `Resources.resx` can't be - it's the basis for generating `Resources.Designer.cs`. I've included `Resources.Designer.cs` because it's already [checked in](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/src/Resources/Strings.Designer.cs) by the main SqlClient project.

It doesn't look like the dotnet/runtime repo includes these any more, so the only reason I can see to keep the file is the localization job. Would removing the .designer.cs files from the repo have any impact on that?

paulmedynski: I suspect you're correct about the localization jobs!

PR #3713, opened 2025-10-22
Enable NuGet audit on all builds

paulmedynski:
## Description

We are now performing implicit NuGet audits on all builds (developer, PR, CI, official, etc).  Any vulnerabilities detected will fail the build.

I also removed our custom audit source, and we will now be using the CFS audit source from the same feed we get our dependencies from.

I left some comments in the main `Directory.Build.props` file about how to temporarily disable audit errors if they can't be addressed immediately.

### DIFF
--- a/eng/pipelines/common/templates/jobs/build-signed-package-job.yml
+++ b/eng/pipelines/common/templates/jobs/build-signed-package-job.yml
@@ -51,6 +51,7 @@ jobs:
   - template: ../steps/generate-nuget-package-step.yml@self
     parameters:
       OutputDirectory: $(artifactDirectory)
+      installNuget: false
   
   - template: ../steps/esrp-code-signing-step.yml@self
     parameters:
--- a/eng/pipelines/steps/compound-nuget-pack-step.yml
+++ b/eng/pipelines/steps/compound-nuget-pack-step.yml
@@ -27,10 +27,26 @@ parameters:
           - Project
 
 steps:
-    - task: NuGetToolInstaller@1
-      displayName: 'Install Latest Nuget'
-      inputs:
-          checkLatest: true
+    # This tool is failing on OneBranch pipelines, possibly due to new
+    # network isolation rules:
+    #
+    #   ERR:Client network socket disconnected before secure TLS connection was established
+    #
+    # Our AKV Official build uses this 1ES image:
+    #
+    #   Image: 1ES-OB-2022-D8-Netlock-V2_westus2_1_image
+    #
+    # An ICM for this issue exists:
+    #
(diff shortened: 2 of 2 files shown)


src/Directory.Build.props
```diff
@@ -62,27 +62,54 @@
...
   <PropertyGroup Condition="'$(BuildSimulator)' == 'true'">
     <DefineConstants>$(DefineConstants);ENCLAVE_SIMULATOR</DefineConstants>
   </PropertyGroup>
 
   <!-- NuGet Audit Settings -->
-  <PropertyGroup Condition="'$(TF_BUILD)' == 'true'">
+  <PropertyGroup>
+    <!--
+      See the NuGet Audit documentation here:
+
```
benrr101, line 75: ultra nit: extra space here :)

src/Directory.Build.props
```diff
@@ -62,27 +62,54 @@
...
+    <!-- We want all possible audit severity messages. -->
+    <NuGetAuditLevel>low</NuGetAuditLevel>
   </PropertyGroup>
+
+  <!-- Build Warning/Error Settings -->
   <PropertyGroup>
-    <NuGetAuditMode>all</NuGetAuditMode>
+    <!-- We treat all warnings as errors. -->
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
```
benrr101, line 95: surely this was being set somewhere else before?

benrr101 (approved):
Looks good to me. Only request is to make sure to run an (un)official mds and akv pipeline (especially akv) in ADO before merging. The official pipelines use the onebranch template that has the restrictive network rules that necessitated the original changes to disable auditing in official builds. The issues didn't come up until I rewrote the akv pipelines, since the mds pipeline is grandfathered into allowing blocked connections to, eg, nuget.org.

src/Directory.Build.props
paulmedynski: Yes, old file line 65.  I moved those properties here so I could talk about them in a sensible grouping.

src/Directory.Build.props
paulmedynski: It's an intentional blank line, like on line 109.

paulmedynski:
ADO.net project non-official builds are here (can't to official builds from a topic branch):

MDS: https://sqlclientdrivers.visualstudio.com/ADO.Net/_build/results?buildId=127989

AKV: https://sqlclientdrivers.visualstudio.com/ADO.Net/_build/results?buildId=127988

cheenamalhotra (approved):

PR #3715, opened 2025-10-23
[6.1] Refactor Test TDS Servers

mdaigle:
Backports #3488 to 6.1. We anticipate adding more simulated server tests as we address bugs, so this change will make the bugfix backports to 6.1 easier.


apoorvdeshmukh (approved):

mdaigle:
Bypassing checks. The one enclave job that's failing is due to agent provisioning issues in the 1ES hosted pool.

PR #3738, opened 2025-10-30
Merge | SqlCommand The Mop Up

benrr101:
## Description
In the final installment of merging the SqlCommand class, I've mopped up the remaining methods and whatnot into the main SqlCommand partial and a new partial for batch RPC methods (not convinced this is the right idea, but otherwise the main partial becomes way too big, imo). This PR completes the SqlCommand merge (and leaves one class left to merge). As with the other PRs as part of this series, each commit merges a small collection of members. It is recommended to step through the commits one at a time to review.

This PR merges the following members:
* _cachedAsyncState
* _cachedMetaData
* _currentlyExecutingBatch
* _internalEndExecuteInitiated
* _reconnectionCompletionSource
* _rpcArrayOf1
* _RPCList
* s_cancelIgnoreFailure
* AddBatchCommand
* BuildParamList (combine a bunch of StringBuilder.Append into AppendFormat)
* CachedAsyncState
* CachingQueryMetadataPostponed
* CancelIgnoreFailure
* CancelIgnoreFailureCallback
* CheckNotificationStateAndAutoEnlist
* CheckThrowSNIException (made private, rewritten as ?.)
* ClearBatchCommand (made private)
* CountSendableParameters (inverted the logic)
* CreateLocalCompletionTask
* DebugForceAsyncWriteDelay
* DeriveParameters
* GetBatchCommand
* GetCommandText
* GetCurrentBatchCommand
* GetCurrentBatchIndex
* GetCurrentParameterCollection
* GetErrors
* GetParameterCount
* GetParameterDirectionFromOleDbDirection
* GetParameterForOutputValueExtraction (made static, rewrote to remove boolean "found" flag and just return directly)
* GetRecordsAffected
* GetResetOptionsString (made static, renamed to GetOptionsResetString)
* GetRPCObject
* GetSetStringOptionsString (made static, renamed to GetOptionsSetString)
* GetStateObject
* IsProviderRetriable
* MaxRPCNameLength
* MetaData
* NotifyDependency (rewritten as ?.)
* OnConnectionClosed (rewritten as ?.)
* OnStatementCompleted
* ParseAndQuoteIdentifier
* PreSql2008ProcParamsNames
* ProcParamsColIndex
* PutStateObject
* QuoteIdentifier (using netcore version with Span<string>)
* RegisterForConnectionCloseNotification
* ReliablePutStateObject
* SetBatchRPCMode
* SetBatchRPCModeReadyToExecute
* SetUpRPCParameters
* ShouldSendParameter
* Sql2008ProcParamsNames
* SqlNotificationContext
* ThrowIfReconnectionHasBeenCanceled (inverted if statement, using conditional access)
* TriggerInternalEndAndRetryIfNecessary
* UnquoteProcedureName
* UnquoteProcedurePart
* ValidateCommand
* ValidateAsyncCommand
* VerifyEndExecuteState
* WaitForAsyncResults
* WriteBeginExecuteEvent
* WriteEndExecuteEvent (inverted if and renamed synchronous to isSynchronous)

## Issues
More work towards #1261 

## Testing
CI should validate 😬

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.cs
@@ -1866,7 +1866,7 @@ namespace Microsoft.Data.SqlClient
         private static void CancelIgnoreFailureCallback(object state) =>
             ((SqlCommand)state).CancelIgnoreFailure();
 
-        // @TODO: Assess if a parameterized version of this method is necessary or if a property can suffice.
+        // @TODO: Assess if a parameterized version of this method is necessary or if a property on SqlParameterCollection can suffice.
         private static int CountSendableParameters(SqlParameterCollection parameters)
         {
             if (parameters is null)
@@ -1896,18 +1896,21 @@ namespace Microsoft.Data.SqlClient
         private static string GetOptionsSetString(CommandBehavior behavior)
         {
             string s = null;
-            if (behavior is CommandBehavior.SchemaOnly or CommandBehavior.KeyInfo)
+
+            bool hasKeyInfo = (behavior & CommandBehavior.KeyInfo) == CommandBehavior.KeyInfo;
+            bool hasSchemaOnly = (behavior & CommandBehavior.SchemaOnly) == CommandBehavior.SchemaOnly;
+            if (hasKeyInfo || hasSchemaOnly)
             {
                 // SET FMTONLY ON will cause the server to ignore other SET OPTIONS, so turn if off
                 // before we ask for browse mode metadata
                 s = TdsEnums.FMTONLY_OFF;
 
-                if (behavior is CommandBehavior.KeyInfo)
+                if (hasKeyInfo)
                 {
                     s += TdsEnums.BROWSE_ON;
                 }
 
-                if (behavior is CommandBehavior.SchemaOnly)
+                if (hasSchemaOnly)
                 {
                     s += TdsEnums.FMTONLY_ON;
                 }
@@ -1920,12 +1923,12 @@ namespace Microsoft.Data.SqlClient
         {
             string s = null;
 
-            if (behavior is CommandBehavior.SchemaOnly)
+            if ((behavior & CommandBehavior.SchemaOnly) == CommandBehavior.SchemaOnly)
             {
                 s += TdsEnums.FMTONLY_OFF;
             }
 
-            if (behavior is CommandBehavior.KeyInfo)
+            if ((behavior & CommandBehavior.KeyInfo) == CommandBehavior.KeyInfo)
             {
                 s += TdsEnums.BROWSE_OFF;
             }
@@ -2125,7 +2128,7 @@ namespace Microsoft.Data.SqlClient
         private static string UnquoteProcedurePart(string part)
         {
             // @TODO: Combine if statements if this method should exist
-            if (part is not null && 2 <= part.Length)
+            if (part is not null && part.Length >= 2)
             {
                 if (part[0] == '[' && part[part.Length - 1] == ']')
                 {
@@ -2488,7 +2491,8 @@ namespace Microsoft.Data.SqlClient
 
                         if (!shouldRetry)
                         {
-                            // If we cannot retry, reset the async state to make sure we leave is clean.
+                            // If we cannot retry, reset the async state to make sure we leave a
+                            // clean state
                             CachedAsyncState?.ResetAsyncState();
 
                             try
@@ -2840,13 +2844,12 @@ namespace Microsoft.Data.SqlClient
                 return;
             }
 
-            if (_reconnectionCompletionSource?.Task?.IsCanceled ?? false)
+            if (_reconnectionCompletionSource?.Task.IsCanceled ?? false)
             {
                 throw SQL.CR_ReconnectionCancelled();
             }
         }
 
-        // @TODO: This method *needs* to be decomposed.
         private bool TriggerInternalEndAndRetryIfNecessary(
             CommandBehavior behavior,
             object stateObject,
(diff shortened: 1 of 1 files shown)


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.Reader.cs
```diff
@@ -503,14 +518,17 @@ private _SqlRPC BuildPrepExec(CommandBehavior behavior)
...
+                _parameters,
+                includeReturnValue: false);
             sqlParam = rpc.systemParams[1];
             // @TODO: This pattern is used quite a bit - it could be factored out
             sqlParam.SqlDbType = (paramList.Length << 1) <= TdsEnums.TYPE_SIZE_LIMIT
                 ? SqlDbType.NVarChar
                 : SqlDbType.NText;
-            sqlParam.Value = paramList;
             sqlParam.Size = paramList.Length;
+            sqlParam.Value = paramList;
```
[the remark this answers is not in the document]
benrr101: Theoretically should make no difference what order they go in, but ok

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.cs
```diff
@@ -2276,7 +2277,7 @@ and not SqlDbType.Udt
                         {
                             int actualBytes = parser.GetEncodingCharLength(
                                 value: s,
-                                numChars: sqlParam.GetActualScale(),
+                                numChars: sqlParam.GetActualSize(),
```
benrr101, line 2283: Bug was here - number of characters should be **Size** not Scale

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.cs
```diff
@@ -1896,18 +1896,21 @@ private static int CountSendableParameters(SqlParameterCollection parameters)
         private static string GetOptionsSetString(CommandBehavior behavior)
         {
             string s = null;
-            if (behavior is CommandBehavior.SchemaOnly or CommandBehavior.KeyInfo)
+
+            bool hasKeyInfo = (behavior & CommandBehavior.KeyInfo) == CommandBehavior.KeyInfo;
```
benrr101, line 1900: Wasn't confident in using `is` here since CommandBehavior is a `[Flags]` enum. And didn't like how the checks were so long and used >once, so I factored them out.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.cs
```diff
@@ -1920,12 +1923,12 @@ private static string GetOptionsResetString(CommandBehavior behavior)
         {
             string s = null;
 
-            if (behavior is CommandBehavior.SchemaOnly)
+            if ((behavior & CommandBehavior.SchemaOnly) == CommandBehavior.SchemaOnly)
```
benrr101, line 1926: Not confident on using `is` on a `[Flags]` enum.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.cs
```diff
@@ -2947,9 +2950,9 @@ private void ValidateCommand(bool isAsync, [CallerMemberName] string method = ""
                 throw ADP.ConnectionRequired(method);
             }
 
-            // Ensure that the connection is open and that the PArser is in the correct state
+            // Ensure that the connection is open and that the parser is in the correct state
             // @TODO: Remove cast when possible.
-            SqlInternalConnectionTds tdsConnection = (SqlInternalConnectionTds)_activeConnection.InnerConnection;
+            SqlInternalConnectionTds tdsConnection = _activeConnection.InnerConnection as SqlInternalConnectionTds;
```
benrr101, line 2955: Further down we check for `null`. `as` operator will give `null` if it's not the right type, but a straight cast will throw an invalid cast exception. Changed back to `as` to preserve original behavior.

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlCommand.netfx.cs
```diff
@@ -1072,7 +1072,7 @@ private void SetUpRPCParameters(_SqlRPC rpc, bool inSchema, SqlParameterCollecti
                                     parts[1] != null // server should not be null or we don't need to remove it
                                 )
                                 {
-                                    parameter.TypeName = QuoteIdentifier(parts, 2, 2);
+                                    parameter.TypeName = QuoteIdentifier(parts.AsSpan(2, 2));
```
paulmedynski: This adds a small (tiny?) performance hit.  Is this a hot path where that is important?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.cs
paulmedynski: `is` doesn't work for flag enums.  You can use `Enum.HasFlag(...)` or use bitwise operators.

priyankatiwari08 (approved):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlCommand.cs
edwardneal: I think the current use of bitwise operators is fine here. [`Enum.HasFlag` will box on netfx.](https://learn.microsoft.com/en-us/visualstudio/profiling/performance-insights-enum-hasflag?view=vs-2022)

benrr101: @paulmedynski I ran some tests a long time ago and found the perf hit for using `Enum.anything` is way worse than the perf hit to .AsSpan :)

which honestly sucks, since I'd much prefer to use a nice generic helper than manually do bitwise math

src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlCommand.netfx.cs
benrr101: Agreed it does add a small hit. But I think since we already adopted it in netcore that it's acceptable.

Issue #3725, opened 2025-10-28
Improve samples by building them


ErikEJ:
### Is your feature request related to a problem? Please describe.
None of the samples in the docs/samples folder are built or compiled, leading to potentially unusable samples

### Describe the solution you'd like
Add a project and solution to build these at least locally

One of the challenges is that most of the samples have this shape:

```c#
using System;
using System.Data;
// <Snippet1>
using Microsoft.Data.SqlClient;

class Program
{
    static void Main()
    {
        OpenSqlConnection();
        Console.ReadLine();
    }

    private static void OpenSqlConnection()
    {
        string connectionString = GetConnectionString();

        using (SqlConnection connection = new SqlConnection())
        {
            connection.ConnectionString = connectionString;

            connection.Open();

            Console.WriteLine("State: {0}", connection.State);
            Console.WriteLine("ConnectionString: {0}",
                connection.ConnectionString);
        }
    }

    static private string GetConnectionString()
    {
        // To avoid storing the connection string in your code, 
        // you can retrieve it from a configuration file.
        return "Data Source=MSSQL1;Initial Catalog=AdventureWorks;"
            + "Integrated Security=true;";
    }
}
// </Snippet1>
```

Which of course causes clashes as Program is in the Global namespace

A possible solution would be to add namespaces to sample:

```c#
using System;
using System.Data;
// <Snippet1>
using Microsoft.Data.SqlClient;

namespace SqlConnection_ConnectionString;

class Program
{
    static void Main()
    {
        OpenSqlConnection();
        Console.ReadLine();
    }

    private static void OpenSqlConnection()
    {
        string connectionString = GetConnectionString();

        using (SqlConnection connection = new SqlConnection())
        {
            connection.ConnectionString = connectionString;

            connection.Open();

            Console.WriteLine("State: {0}", connection.State);
            Console.WriteLine("ConnectionString: {0}",
                connection.ConnectionString);
        }
    }

    static private string GetConnectionString()
    {
        // To avoid storing the connection string in your code, 
        // you can retrieve it from a configuration file.
        return "Data Source=MSSQL1;Initial Catalog=AdventureWorks;"
            + "Integrated Security=true;";
    }
}
// </Snippet1>

```

### Additional context

let me know if you think this adds value and I can do it.

ErikEJ:
Further observations: 
 
- Samples must be built against .NET Framework with a newer language version in order to compile.
- Need to target WinForms as many samples use this.

[...]

PR #3740, opened 2025-10-31
New Samples project for building all samples

cheenamalhotra:
Addresses #3725 - this PR adds a new Sample project for build support.

There are issues with samples that need to be addressed gradually and can be done with community support, but this PR introduces a basic scaffolding to initiate build support.

### DIFF
--- a/doc/samples/Microsoft.Data.SqlClient.Samples.csproj
+++ b/doc/samples/Microsoft.Data.SqlClient.Samples.csproj
@@ -1,0 +1,28 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Library</OutputType>
+    <ImplicitUsings>disable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <LangVersion>Latest</LangVersion>
+    <TreatWarningsAsErrors>False</TreatWarningsAsErrors>
+  </PropertyGroup>
+  
+  <ItemGroup>
+    <PackageReference Include="Azure.Identity" />
+    <PackageReference Include="Azure.Core" />
+    <PackageReference Include="Microsoft.Identity.Client" />
+    <PackageReference Include="Microsoft.SqlServer.Server" />
+    <PackageReference Include="Microsoft.Data.SqlClient" />
+    <PackageReference Include="Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider" />
+    <PackageReference Include="Newtonsoft.Json" />
+  </ItemGroup>
+
+  <!-- .NET Framework references -->
+  <ItemGroup Condition="$(TargetGroup) == 'netfx'">
+    <PackageReference Include="System.Runtime.InteropServices.RuntimeInformation" />
+    <Reference Include="System.Transactions" />
+    <Reference Include="System.Windows.Forms" />
+  </ItemGroup>
+  
+</Project>
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -3,16 +3,14 @@
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
     <CentralPackageTransitivePinningEnabled>true</CentralPackageTransitivePinningEnabled>
   </PropertyGroup>
-  
   <!-- ===================================================================== -->
   <!-- Shared Dependencies -->
-
   <!-- Published -->
   <ItemGroup>
     <!-- MDS and AKV -->
     <PackageVersion Include="Azure.Core" Version="1.49.0" />
+    <PackageVersion Include="Microsoft.Identity.Client" Version="4.78.0" />
   </ItemGroup>
-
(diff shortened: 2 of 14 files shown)


doc/Directory.Package.props
```diff
@@ -0,0 +1,4 @@
+<Project>
```
ErikEJ: Why was this file added?

ErikEJ:
@cheenamalhotra Does the new project build? How do you solve the namespace challenges?

This seems like a lot of product changes to add a validation build of documentation samples.

in my POC, I had simply added a single .csproj in the samples folder (only challenge then was the namespaces)

```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFrameworks>net8.0:net9.0;net472</TargetFrameworks>
    <ImplicitUsings>disable</ImplicitUsings>
    <Nullable>disable</Nullable>
    <UseWindowsForms>true</UseWindowsForms>
    <LangVersion>10.0</LangVersion>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.Data.SqlClient" Version="7.0.0-preview2.25289.6" />
    <PackageReference Include="Microsoft.SqlServer.Server" Version="1.0.0" />
  </ItemGroup>

</Project>

```

doc/samples/Microsoft.Data.SqlClient.Samples.csproj
```diff
@@ -0,0 +1,43 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
```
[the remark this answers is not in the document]
paulmedynski: I agree here - let's target the same thing the driver does.

Also:
- Should the samples be executables rather than libraries?
- I think enabling ImplicitUsings will help de-clutter the samples.
- LangVersion should be 13 so we can use modern language features.

Our samples should also demonstrate best practices since they will likely be copy-pasted into apps as-is.

doc/Directory.Package.props
paulmedynski: It causes MSBuild to load our global NuGet package version file when building the samples project(s).

src/Directory.Build.props
paulmedynski, line 1: Oh boy this will conflict with my Azure split feature branch.  Who will win the conflict war?

You don't need this file at all - MSBuild will work up the directory tree on its own.  Same for the now-empty Directory.Packages.props file.

doc/Directory.Package.props
paulmedynski: Correct - this file isn't necessary.  MSBuild will find the root Directory.Packages.props file on its own.

doc/Directory.Build.props
paulmedynski, line 1: This file isn't necessary.

doc/samples/Microsoft.Data.SqlClient.Samples.csproj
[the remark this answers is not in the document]
cheenamalhotra: Usings should generally be included as in some samples, we do have types used from a specific namespace, while they exist in more than one namespace. e.g. Microsoft.Data.SqlClient.Server namespace is one of the contradictory ones. Maybe we can decide as we go, how much inclusion we need - on sample basis.

I will however, update the langversion and enable modern programming styles.

src/Directory.Build.props
cheenamalhotra: Not a priority, so you go first!

On second thought, I also think referencing a released version of MDS is a better idea as these docs are technically directly referenced in public MS documentation, so they should only be updated when a public release is out there with public APIs used. That should make the conflicts go away too.

cheenamalhotra:
> @cheenamalhotra Does the new project build? How do you solve the namespace challenges?
> 
> This seems like a lot of product changes to add a validation build of documentation samples.
> 
> in my POC, I had simply added a single .csproj in the samples folder (only challenge then was the namespaces)
> 
> ```
> <Project Sdk="Microsoft.NET.Sdk">
> 
>   <PropertyGroup>
>     <TargetFrameworks>net8.0:net9.0;net472</TargetFrameworks>
>     <ImplicitUsings>disable</ImplicitUsings>
>     <Nullable>disable</Nullable>
>     <UseWindowsForms>true</UseWindowsForms>
>     <LangVersion>10.0</LangVersion>
>   </PropertyGroup>
> 
>   <ItemGroup>
>     <PackageReference Include="Microsoft.Data.SqlClient" Version="7.0.0-preview2.25289.6" />
>     <PackageReference Include="Microsoft.SqlServer.Server" Version="1.0.0" />
>   </ItemGroup>
> 
> </Project>
> ```

Gonna revisit this, to make it simpler as well as do a public driver reference instead. This was a first hand experiment to understand impact of this work, which is indeed a lot (~700 errors). As of now the solution will not build the project so it will need to be worked upon to fix failures.

I mainly want to enable that and allow community to participate in helping us clear out compilation errors, maintaining sample's expected design and behavior.

ErikEJ:
@cheenamalhotra Sounds good, once the project is in I have some ideas how to make it build.

PR #3743, opened 2025-11-04
Cleanup | SqlInternalConnection properties

mdaigle:
## Description

Cleaned up properties in SqlInternalConnection to make them auto-properties, add summary doc comments, and remove consideration for old SQL Server versions (where applicable).

Should be reviewed commit by commit.

## Testing

No logic changes. Existing tests should cover everything.

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TransactionRequest.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TransactionRequest.cs
@@ -4,12 +4,18 @@
 
 namespace Microsoft.Data.SqlClient;
 
+/// <summary>
+/// Enum representing all of the possible transaction operations.
+/// Only used internally, mapped to <see cref="TdsEnums.TransactionManagerRequestType"/> for transmission
+/// over the wire.
+/// </summary>
 internal enum TransactionRequest
 {
     Begin,
     Promote,
     Commit,
     Rollback,
+    // TODO: since v2005, this is always mapped to the same behavior as Rollback. Remove in future.
     IfRollback,
     Save
 }


src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -670,14 +670,6 @@ internal override bool IsLockedForBulkCopy
             }
         }
 
-        internal protected override bool IsNonPoolableTransactionRoot
```
mdaigle, line 673: This property doesn't compute in my brain and makes it really hard to reason about conditionals where it's used (especially when it's negated). I removed this property in favor of doing the two component checks. I also dropped the !Is2008OrNewer. I don't know what was special about pre-2008, but we have no means to test that behavior and it's long out of support.

Removing it also highlights several critical spots that skip behavior on the assumption that it will be handled by the pool. This wasn't clear before when the check was hidden inside the property.

src/Microsoft.Data.SqlClient/tests/UnitTests/Microsoft/Data/SqlClient/ConnectionPool/ChannelDbConnectionPoolTest.cs
```diff
@@ -281,6 +281,7 @@ out DbConnectionInternal? recycledConnection
         }
 
         [Fact]
+        [ActiveIssue("https://github.com/dotnet/SqlClient/issues/3730")]
```
mdaigle, line 284: This test is flaky in the pipeline and its async counterpart has already been disabled. I'll address this in a future pooling related PR.

priyankatiwari08 (dismissed):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TransactionRequest.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.Data.SqlClient;
+
+internal enum TransactionRequest
```
paulmedynski, line 12: Some docs would be nice :)

src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlInternalConnectionTds.cs
```diff
@@ -949,7 +941,9 @@ protected override void InternalDeactivate()
             // cause our transaction to be rolled back and the connection
             // to be reset.  We'll get called again once the delegated
             // transaction is completed and we can do it all then.
-            if (!IsNonPoolableTransactionRoot)
+            // TODO: I think this logic cares about pooling because the pool
+            // will handle deactivation of pool-associated transaction roots?
+            if (!(IsTransactionRoot && Pool == null))
```
benrr101, line 946: I'm a biiiiiit confused on this one - let me work this one out real quick:
| Step | Action Performed |
| ----- | --------------- |
| `!IsNonPoolableTransactionRoot` | What we start with |
| `!(IsNonPoolableTransactionRoot)` | Add parentheses |
| `!(IsTransactionRoot && (!Is2008OrNewer \|\| Pool == null))` | Substitute in original definition |
| `!IsTransactionRoot \|\| !(!Is2008OrNewer \|\| Pool == null)` | Distribute `!`, which flips `&&` to `\|\|` |
| `!IsTransactionRoot \|\| (Is2008OrNewer && Pool != null)` | Distribute `!`, which flips `\|\| to `&&` |
| `!IsTransactionRoot \|\| (true && Pool != null)` | Substitute `true` for `Is2008OrNewer` |
| `!IsTransactionRoot \|\| Pool !=null` | Cancel out `true` |
| `!(IsTransactionRoot && Pool == null)` | Factor out `!` |

Ok, we should be good then. Just wanted to work it out.

I will say, tho ... the logic is a bit easier to digest if you distribute  the `!`. Just my opinion.

PR #3744, opened 2025-11-05
[DRAFT] Add tests for HasRows + Mutliple INFO tokens

### LINKED ISSUE #3018: HasRows fails to detect rows
### Describe the bug
For queries that do have records, SqlDataReader.HasRows method returns false when the DB returns more than 1 InfoMessageToken, before sending a RowToken, HasRows assumes no records are left, and returns false even though it should have been true. 

Using a db that returns two infomessages in the beginning is enough to reproduce the behavior.

```
string connectionString = "....";
var conn = new SqlConnection(connectionString);
conn.Open();

conn.InfoMessage += (sender, e) => Console.WriteLine(e.Message);

var cmd = conn.CreateCommand();
cmd.CommandText = "PRINT('1');SELECT * FROM t1";
using (var dr = cmd.ExecuteReader())
{
    Console.WriteLine($"HasRows: {dr.HasRows}"); // will return false if 2 infoMessages are sent by the DB.
    while (dr.Read())
    {
        Console.WriteLine(dr[0]);
    }

}
```

### Expected behavior
HasRows shoud return true.

**Additional context**
When trying to identify the issue we encounter that [this](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlDataReader.cs#L4888) should probably be a `while` instead of an `if` statement.

paulmedynski:
## Description

- Added tests for SqlDataReader.HasRows for responses with and without INFO tokens.
- Tweaked existing TDS Server bits to emit logs to xUnit output.

### DIFF
--- a/src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/HasRowsTests.cs
+++ b/src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/HasRowsTests.cs
@@ -108,18 +108,20 @@ public sealed class HasRowsTests : IDisposable
     }
 
     // Verify that HasRows is true when a variable number of INFO tokens is
-    // included in the response to a SQL batch that returns rows.
+    // included at the start of the response stream to a SQL batch that returns
+    // rows.
     [Theory]
     [InlineData(0)]
     [InlineData(1)]
     [InlineData(2)]
     [InlineData(3)]
     [InlineData(10)]
-    public void InfoAndRows(ushort infoCount)
+    public void InfoAndRows_Start(ushort infoCount)
     {
-        // Configure the engine to include the desired number of INFO tokens
-        // with its response.
+        // Configure the engine to specify the desired number and placement of
+        // INFO tokens with its response.
         _engine.InfoCount = infoCount;
+        _engine.InfoPlacement = InfoPlacement.Start;
 
         using SqlCommand command = new(
             // Use command text that is intercepted by our engine.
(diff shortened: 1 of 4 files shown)


src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/QueryEngine.cs
```diff
@@ -33,6 +33,7 @@ public class QueryEngine
         /// </summary>
         public QueryEngine(TdsServerArguments arguments)
         {
+            Log = arguments.Log;
```
paulmedynski, line 36: This ensures the engine has its Log set, regardless of what TdsServer may decide.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/QueryEngine.cs
```diff
@@ -327,8 +328,9 @@ protected virtual TDSMessageCollection CreateQueryResponse(ITDSServerSession ses
             }
             else
             {
-                // Create an info token that contains the query received
-                TDSInfoToken infoToken = new TDSInfoToken(2012, 2, 0, lowerBatchText);
+                // Create an info token that contains the verbatim query text we
+                // received (not the lower-cased version).
+                TDSInfoToken infoToken = new TDSInfoToken(2012, 2, 0, batchRequest.Text);
```
paulmedynski, line 333: Now the query text in the INFO matches what the test actually sent.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS.Servers/TdsServer.cs
```diff
@@ -20,6 +20,15 @@ public TdsServer(TdsServerArguments arguments) : base(arguments)
         {
         }
 
+        /// <summary>
+        /// Constructor with a query engine.  Uses the engine's servers arguments.
+        /// </summary>
+        /// <param name="queryEngine">Query engine</param>
+        public TdsServer(QueryEngine queryEngine)
```
paulmedynski, line 27: Convenience constructor to avoid having to specify ServerArguments twice in the test.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS/TDSUtilities.cs
```diff
@@ -345,109 +345,107 @@ public static void Log(TextWriter log, string prefix, object instance)
...
-                int index = 0;
-
-                // Log values
-                foreach (object o in (instance as System.Collections.IEnumerable))
+                // Check if null
+                if (instance == null)
                 {
-                    Log(log, string.Format("{0}[{1}]", prefix, index++), o);
+                    SerializedWriteLineToLog(log, string.Format("{0}: <null>", prefix));
+                    log.Flush();
```
paulmedynski, line 354: Added missing Flush() calls.

src/Microsoft.Data.SqlClient/tests/tools/TDS/TDS/TDSUtilities.cs
```diff
@@ -345,109 +345,107 @@ public static void Log(TextWriter log, string prefix, object instance)
...
         /// <param name="log">Destination</param>
         /// <param name="text">Text to log</param>        
         public static void SerializedWriteLineToLog(TextWriter log, string text)
         {
-            lock (s_logWriterLock)
-            {
-                log.WriteLine(string.Format("[{0}] {1}", DateTime.Now, text));
-            }
+            var now = DateTime.UtcNow;
+            log.WriteLine($"[{now:yyyy-MM-dd hh:mm:ss.fff}Z] {text}");
```
paulmedynski, line 448: Better timestamp formatting, and changed to UTC.

src/Microsoft.Data.SqlClient/tests/UnitTests/SimulatedServerTests/HasRowsTests.cs
```diff
@@ -0,0 +1,264 @@
...
+
+        using SqlCommand command = new(
+            // Use command text that is intercepted by our engine.
+            InfoQueryEngine.CommandText,
+            _connection);
+        using SqlDataReader reader = command.ExecuteReader();
+
+        // We should have read past the INFO tokens and determined that there
+        // are row results.
+        Assert.True(reader.HasRows);
```
paulmedynski, line 134: I was expecting this to fail for `infoCount > 1` based on #3018, but it passes.  I will investigate further.

paulmedynski (commented):
Commentary for reviewers.

PR #3746, opened 2025-11-06
Cleanup | Split out TransactedConnectionPool

mdaigle:
## Description

Moves the TransactedConnectionPool class out from WaitHandleDbConnectionPool. Modernizes the class, adds doc comments, and adds unit test coverage. These tests highlight some deficiencies with how transactions are currently managed in the WaitHandleDbConnectionPool. I comment on these issues, but do not address them in this PR.

**No logic changes are made to TransactedConnectionPool in this PR.**

Changes should be reviewed commit by commit.

## Testing

Adds unit testing.

### DIFF
--- a/src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/TransactedConnectionPoolTest.cs
+++ b/src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/TransactedConnectionPoolTest.cs
@@ -60,10 +60,11 @@ public class TransactedConnectionPoolTest
     {
         // Arrange
         var transactedPool = new TransactedConnectionPool(new MockDbConnectionPool());
-        using var transaction = new TransactionScope();
+        using var transactionScope = new TransactionScope();
+        var transaction = Transaction.Current!;
 
         // Act
-        var result = transactedPool.GetTransactedObject(Transaction.Current!);
+        var result = transactedPool.GetTransactedObject(transaction);
 
         // Assert
         Assert.Null(result);
@@ -669,7 +670,7 @@ public class TransactedConnectionPoolTest
         public DbConnectionPoolState State => throw new NotImplementedException();
         public bool UseLoadBalancing => throw new NotImplementedException();
 
-        public List<DbConnectionInternal> ReturnedConnections { get; } = new();
+        public ConcurrentBag<DbConnectionInternal> ReturnedConnections { get; } = new();
 
         public void Clear() => throw new NotImplementedException();
 


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/TransactedConnectionPool.cs
```diff
@@ -0,0 +1,348 @@
...
+        lock (_transactedCxns)
+        {
+            // Check if a transacted pool has been created for this transaction
+            if ((txnFound = _transactedCxns.TryGetValue(transaction, out connections)) 
+                && connections is not null)
+            {
+                // synchronize multi-threaded access with GetTransactedObject
+                lock (connections)
+                {
+                    SqlClientEventSource.Log.TryPoolerTraceEvent("<prov.DbConnectionPool.TransactedConnectionPool.PutTransactedObject|RES|CPOOL> {0}, Transaction {1}, Connection {2}, Pushing.", Id, transaction.GetHashCode(), transactedObject.ObjectID);
```
edwardneal, line 195: One missing debug assertion here - is that deliberate?

```cs
Debug.Assert(0 > connections.IndexOf(transactedObject), "adding to pool a second time?");
```

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/TransactedConnectionPool.cs
```diff
@@ -0,0 +1,348 @@
...
+    {
+        private Transaction _transaction;
+        
+        /// <summary>
+        /// Initializes a new instance of the TransactedConnectionList class with the specified
+        /// initial capacity and associated transaction.
+        /// </summary>
+        /// <param name="initialAllocation">The initial number of elements that the list can contain.</param>
+        /// <param name="tx">The transaction associated with the connections in this list.</param>
+        internal TransactedConnectionList(int initialAllocation, Transaction tx) : base(initialAllocation)
```
edwardneal, line 43: For later: do we need this class? We only ever dispose of it in `TransactionEnded`, and at that point we're inside a lock where we've successfully looked up the transaction in `_transactedCxns`. If we don't need to care about the disposal logic, we could either eliminate the now-unnecessary class or refactor it to contain the logic which maintains the `DbConnectionInternal` entries.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/TransactedConnectionPool.cs
```diff
@@ -0,0 +1,348 @@
...
+            if (_transaction != null)
+            {
+                _transaction.Dispose();
+            }
+        }
+    }
+
+    #region Fields
+
+    Dictionary<Transaction, TransactedConnectionList> _transactedCxns;
```
edwardneal: For later: if this is going to be part of the redesigned connection pool, we might be able to replace this with a `ConcurrentDictionary`. Doing so could eliminate the per-pool lock which is taken on this variable and simplify `PutTransactedObject`. It can also remove some of the not-null checks which had to be added for nullability analysis to work.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/TransactedConnectionPool.cs
```diff
@@ -0,0 +1,348 @@
...
+    /// Gets the unique identifier for this transacted connection pool instance.
+    /// </summary>
+    /// <value>A unique integer identifier used for logging and diagnostics.</value>
+    internal int Id => _objectID;
+
+    /// <summary>
+    /// Gets the main connection pool that this transacted pool is associated with.
+    /// </summary>
+    /// <value>The IDbConnectionPool instance that owns this transacted pool.</value>
+    internal IDbConnectionPool Pool { get; private init; }
```
edwardneal: We don't need `private init;` here.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/TransactedConnectionPool.cs
```diff
@@ -0,0 +1,348 @@
...
+    /// </remarks>
+    internal void TransactionEnded(Transaction transaction, DbConnectionInternal transactedObject)
+    {
+        SqlClientEventSource.Log.TryPoolerTraceEvent("<prov.DbConnectionPool.TransactedConnectionPool.TransactionEnded|RES|CPOOL> {0}, Transaction {1}, Connection {2}, Transaction Completed", Id, transaction.GetHashCode(), transactedObject.ObjectID);
+        TransactedConnectionList? connections;
+        int entry = -1;
+
+        // NOTE: because TransactionEnded is an asynchronous notification, there's no guarantee
+        //   around the order in which PutTransactionObject and TransactionEnded are called. As
+        //   such, it is possible that the transaction does not yet have a pool created.
```
edwardneal, line 293: Possibly for later: it might be worth clarifying whether this is still accurate, and which pool this comment refers to - the transacted pool, or its parent. This is only ever called if the connection's pool and the pool's `_transactedConnectionPool` variable is non-null.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/TransactedConnectionPool.cs
```diff
@@ -0,0 +1,348 @@
...
+
+                // Lock connections to avoid conflict with GetTransactionObject
+                lock (connections)
+                {
+                    entry = connections.IndexOf(transactedObject);
+
+                    if (entry >= 0)
+                    {
+                        connections.RemoveAt(entry);
+                    }
```
edwardneal, line 314: For later: this is an odd pattern, `List<T>.Remove` returns false if it failed. As mentioned above, perhaps this should be made the responsibility of the `TransactedConnectionList` class.

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/TransactedConnectionPoolTest.cs
```diff
@@ -0,0 +1,762 @@
...
+        var pool2 = new TransactedConnectionPool(new MockDbConnectionPool());
+
+        // Act & Assert
+        Assert.NotEqual(pool1.Id, pool2.Id);
+        Assert.True(pool1.Id > 0);
+        Assert.True(pool2.Id > 0);
+    }
+
+    [Fact]
+    public void Pool_ReturnsCorrectPool()
```
edwardneal: Possible duplicate of `Constructor_WithValidPool_SetsPoolProperty`?

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/TransactedConnectionPoolTest.cs
```diff
@@ -0,0 +1,762 @@
...
+                    retrievedConnections.Add(conn);
+                }
+            });
+        }
+
+        Task.WaitAll(tasks);
+
+        // Assert
+        Assert.Equal(connections.Length, retrievedConnections.Count);
+        Assert.True(retrievedConnections.All(connections.Contains));
```
edwardneal: Very tentatively: is this assertion relevant? We're running through every element in `retrievedConnections`, verifying whether or not `connections` contains it. If `GetTransactedObject` returned the same instance to two threads, it'd pass both assertions.

We essentially just want to make sure that both collections are equal. Maybe this means that `MockDbConnectionInternal` needs to implement IEquatable on the basis of its `MockId`.

edwardneal (commented):
No objections to the cleanup, (apart from one missing debug assertion) but I've got a few comments for the pile when you start looking at the overall design.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/TransactedConnectionPool.cs
```diff
@@ -0,0 +1,348 @@
...
+/// </remarks>
+internal class TransactedConnectionPool
+{
+    /// <summary>
+    /// A specialized list that holds database connections associated with a specific transaction.
+    /// Maintains a reference to the transaction for proper cleanup when the transaction completes.
+    /// </summary>
+    private sealed class TransactedConnectionList : List<DbConnectionInternal>
+    {
+        private Transaction _transaction;
```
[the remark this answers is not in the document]
paulmedynski: This change is a good idea for this PR.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/TransactedConnectionPool.cs
```diff
@@ -0,0 +1,348 @@
...
+/// maintaining transaction integrity and thread safety.
+/// </summary>
+/// <remarks>
+/// The TransactedConnectionPool works in conjunction with the main connection pool to handle 
+/// connections that are enlisted in System.Transactions. When a connection is enlisted in a 
+/// transaction, it cannot be returned to the general pool until the transaction completes.
+/// This class provides temporary storage for such connections, organized by transaction,
+/// allowing for efficient reuse within the same transaction scope.
+/// </remarks>
+internal class TransactedConnectionPool
```
paulmedynski, line 27: For future consideration:  Seems like this is `TransactingConnectionPool` (current tense) rather than past tense.  The transaction is ongoing, not completed.

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/TransactedConnectionPoolTest.cs
```diff
@@ -0,0 +1,762 @@
...
+
+    [Fact]
+    public void GetTransactedObject_WithNonExistentTransaction_ReturnsNull()
+    {
+        // Arrange
+        var transactedPool = new TransactedConnectionPool(new MockDbConnectionPool());
+        using var transaction = new TransactionScope();
+
+        // Act
+        var result = transactedPool.GetTransactedObject(Transaction.Current!);
```
paulmedynski: What happens if `Transaction.Current` is null?

Maybe we want to assert it isn't null before using it, which would eliminate the need for the `!`.

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/TransactedConnectionPoolTest.cs
```diff
@@ -0,0 +1,762 @@
...
+        var retrievedConnections = new ConcurrentBag<DbConnectionInternal>();
+        var tasks = new Task[connections.Length];
+
+        // Act - retrieve connections concurrently
+        for (int i = 0; i < tasks.Length; i++)
+        {
+            tasks[i] = Task.Run(() =>
+            {
+                var conn = transactedPool.GetTransactedObject(transaction);
+                if (conn != null)
```
paulmedynski: Should this be an `Assert.NotNull(conn);` ?

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/TransactedConnectionPoolTest.cs
```diff
@@ -0,0 +1,762 @@
...
+        using var transactionScope = new TransactionScope();
+        var transaction = Transaction.Current!;
+
+        // Act & Assert
+        Assert.Throws<NullReferenceException>(() => 
+            transactedPool.PutTransactedObject(transaction, null!));
+    }
+
+    [Fact]
+    public void PutTransactedObject_WithNewTransaction_CreatesNewPool()
```
paulmedynski: Creates new pool?  Not sure what you mean here.  Same with the next test.

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/TransactedConnectionPoolTest.cs
```diff
@@ -0,0 +1,762 @@
...
+        var transaction = Transaction.Current!;
+
+        // Add connection to transacted pool
+        transactedPool.PutTransactedObject(transaction, connection);
+
+        // Act
+        transactedPool.TransactionEnded(transaction, connection);
+
+        // Assert
+        Assert.Contains(connection, mockPool.ReturnedConnections);
```
paulmedynski, line 331: This is asymmetric behaviour:
- `PutTransactedObject` doesn't remove the connection from the main pool.
- `TransactionEnded` adds the connection to the main pool.

Not suggesting you change it here, but maybe add a comment that it's weird for the responsibility of moving connections in/out of the pools is split across 2 or more classes.

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/TransactedConnectionPoolTest.cs
```diff
@@ -0,0 +1,762 @@
...
+        transactedPool.PutTransactedObject(transaction, connection2);
+
+        // Act - end only one connection
+        transactedPool.TransactionEnded(transaction, connection1);
+
+        // Assert
+        Assert.Contains(connection1, mockPool.ReturnedConnections);
+        Assert.DoesNotContain(connection2, mockPool.ReturnedConnections);
+        
+        // Verify other connection is still in pool
```
paulmedynski, line 361: Why is connection2 still in the transaction pool when the transaction has ended?  Shouldn't all connections associated to the transaction be returned to the main pool?  How can a transaction end for a subset of the connections associated to it, but not the others?

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/TransactedConnectionPoolTest.cs
```diff
@@ -0,0 +1,762 @@
...
+
+        // Arrange
+        var mockPool = new MockDbConnectionPool();
+        var transactedPool = new TransactedConnectionPool(mockPool);
+        var connection = new MockDbConnectionInternal();
+
+        using var transactionScope = new TransactionScope();
+        var transaction = Transaction.Current!;
+
+        // Act - simulate race condition where TransactionEnded is called before PutTransactedObject
```
paulmedynski, line 463: This isn't really a race though - it's just two sequential operations on two separate transactions as far as the pool is concerned.

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/TransactedConnectionPoolTest.cs
```diff
@@ -0,0 +1,762 @@
...
+        transactedPool.TransactionEnded(transaction, connection);
+        transactedPool.PutTransactedObject(transaction, connection);
+
+        // Assert - connection should still be in the transacted pool
+        var retrievedConnection = transactedPool.GetTransactedObject(transaction);
+        Assert.Equal(connection, retrievedConnection);
```
paulmedynski: Reference equality using `Assert.Same()` is probably the correct choice here (and other places throughout the tests).

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/TransactedConnectionPoolTest.cs
```diff
@@ -0,0 +1,762 @@
...
+        
+        using var transactionScope = new TransactionScope();
+        var transaction = Transaction.Current!;
+
+        // Act & Assert
+        // 1. Put connection in transacted pool
+        transactedPool.PutTransactedObject(transaction, connection);
+        
+        // 2. Get connection from transacted pool
+        var retrievedConnection = transactedPool.GetTransactedObject(transaction);
```
paulmedynski, line 492: Does Get _remove_ the connection from the transaction pool?  If so, then Get is a bad choice of verb IMO.  Should you be confirming that the connection no longer exists in the pool after Get ?

paulmedynski (changes requested):

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/TransactedConnectionPoolTest.cs
edwardneal: `TransactedConnectionList` behaves like a stack, so Get and Put could easily be Pop and Push. The only difference between TransactedConnectionList and Stack is that the `TransactionEnded` method tries to remove the passed connection wherever it appears in the list rather than pushing and popping.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/TransactedConnectionPool.cs
```diff
@@ -0,0 +1,348 @@
...
+                    //   add a different connection to the transacted pool under the same 
+                    //   transaction. As a result, threadB may have completed creating the
+                    //   transacted pool while threadA was processing the above instructions.
+                    if ((txnFound = _transactedCxns.TryGetValue(transaction, out connections))
+                        && connections is not null)
+                    {
+                        // synchronize multi-threaded access with GetTransactedObject
+                        lock (connections)
+                        {
+                            Debug.Assert(0 > connections.IndexOf(transactedObject), "adding to pool a second time?");
```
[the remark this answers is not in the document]
mdaigle: Added a todo

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/TransactedConnectionPool.cs
```diff
@@ -0,0 +1,348 @@
...
+            }
+        }
+
+        // CONSIDER: the following code is more complicated than it needs to be to avoid cloning the 
+        //   transaction and allocating memory within a lock. Is that complexity really necessary?
+        if (!txnFound)
+        {
+            // create the transacted pool, making sure to clone the associated transaction
+            //   for use as a key in our internal dictionary of transactions and connections
+            Transaction? transactionClone = null;
```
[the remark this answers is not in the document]
mdaigle: Some change like this would be nice, but it is beyond the scope of this PR.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/TransactedConnectionPool.cs
mdaigle: I removed this because I feel it's better as something that's actually enforced by design. I added a test that shows that this isn't enforced and this line breaks that test. I've added a TODO here in its place.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/TransactedConnectionPool.cs
mdaigle: According to comments, this was done for performance to avoid iterating over the keys, but I don't think that's a real concern. Will take a look at this.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/TransactedConnectionPool.cs
mdaigle: That's my hope. I want to approach the locked sections carefully because there's some interplay with transaction ended callbacks.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/TransactedConnectionPool.cs
mdaigle: I'm very suspicious of a lot of the interactions here. I would like to clarify duties between the classes and get things safe by design. This will also benefit the existing connection pool.

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/TransactedConnectionPoolTest.cs
mdaigle: flipped the check to make sure everything from connections is in retrieved connections.

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/TransactedConnectionPoolTest.cs
mdaigle: Creating a transaction scope (above) sets the current transaction for the thread.
https://learn.microsoft.com/en-us/dotnet/framework/data/transactions/implementing-an-implicit-transaction-using-transaction-scope

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/TransactedConnectionPoolTest.cs
mdaigle: Clarified, should be connection list, not pool.

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/TransactedConnectionPoolTest.cs
mdaigle: Agreed, there are some wonky things in here. Added a comment.

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/TransactedConnectionPoolTest.cs
mdaigle: Removal from the pool is currently the duty of the TransactionEnded method. Each connection that participates in the transaction registers a callback to the TransactionEnded event that calls the TransactionEnded method. Removal from the pool happens connection by connection, as those callbacks are scheduled.

https://github.com/dotnet/SqlClient/blob/2fed073bfce7e4a096ce2b7ad32d9fb59ff20148/src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbConnectionInternal.cs#L381

A better way could be to register a single callback that clears the whole connection list for that transaction. Will address this in future PRs.

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/TransactedConnectionPoolTest.cs
paulmedynski: Well that's an "interesting" API.  Too bad the scope didn't have a Transaction getter.

I guess my only suggestion here would be to name the TransactionScope variable `scope` instead of `transaction`.

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/TransactedConnectionPoolTest.cs
mdaigle: It's a race in the sense that transaction ended is run as a callback. It's theoretically possible for the transaction to complete quickly and for that callback to be scheduled and run before the connection is put into the transacted pool. This test shows that we don't handle that case well and the connection is added to the pool even though the transaction has ended. This would strand the connection here as there's no other way to get it out if this occurs.

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/TransactedConnectionPoolTest.cs
paulmedynski: Indeed - there are some fragile design choices here that we could certainly improve.

paulmedynski (changes requested):
Resolved some comments, others remain.

mdaigle:
@paulmedynski just pushed changes addressing your comments

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/TransactedConnectionPoolTest.cs
paulmedynski: We should confirm that the `retrievedConnection` is no longer in the pool's list for this transaction.

paulmedynski: Ok you're already checking that on line 92.

Issue #3640, opened 2025-09-28
number-of-active-connections metric goes negative, "activated multiple times?" debug assertions failing


borland:
### Describe the bug

Since upgrading from SqlClient v6.0.2 to v6.1.1, we noticed in our metrics that the `number-of-active-connections` started to report negative values, sometimes quite large -- e.g. -357

Rolling back to v6.0.2 removed the issue.

I am not aware of anyone noticing any other problems with the 6.1.1 upgrade at this stage.

### To reproduce

I am not sure what exactly in our codebase is triggering the buggy code path; Our application is quite large and complex, it opens many SQL connections rapidly.

However, I checked out SqlClient locally, built it as part of our application's solution, and used `git bisect` to narrow down the commit which introduces the problem. 

```
0322d44c7cfec424cbe33c9f48430f6ad79a8a50 is the first bad commit
commit 0322d44c7cfec424cbe33c9f48430f6ad79a8a50
Author: Malcolm Daigle <mdaigle@microsoft.com>
Date:   Wed Jan 15 10:38:24 2025 -0800
```

When testing this, I found the [debug assertion in DeactivateConnection](https://github.com/dotnet/SqlClient/blob/eb7a9da3da429f3220a92a8a6b228a7471a884be/src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbConnectionInternal.cs#L524) would fire shortly after launching the application.

### Expected behavior

Metrics should not go negative, debug assertions should not fire.

### Further technical details
Microsoft.Data.SqlClient version: (found on the nuget or Microsoft.Data.SqlClient.dll)
.NET target: (e.g. .NET Framework 4.7.2, .NET 6.0)
SQL Server version: (e.g. SQL Server 2022)
Operating system: (e.g. Windows Server 2022, Ubuntu 24.04, macOS 14.7.1, Docker container)

**Additional context**

Our hosting platform runs Linux, and my local machine is a Mac with an M4 Max. I use JetBrains Rider to build and debug with .NET 8.0.413

borland:
Additional Info: In the process of writing this up, I noticed the PR [Fix Debug Assertion in Connection Pool Due to Double Deactivation](https://github.com/dotnet/SqlClient/pull/3587) which appears not to be available in a public release yet.

I checked out the code from the `main` branch, and indeed the debug assertion no longer fires. However, the metrics are still broken.

[...]

PR #3758, opened 2025-11-12
Fix 3640 | Remove extra connection deactivation.

mdaigle:
## Description

When I originally implemented #3019, due to a misunderstanding about when connections are activated/deactivated, I added an extraneous call to DbConnectionInternal.DeactivateConnection as part of the PutObjectFromTransactedPool. Connections are **always** deactivated at the time they are returned to the pool via the ReturnInternalConnection method and do not need to be activated/deactivated as they move between the main pool and the transacted pool. The activate and deactivate methods modify a counter value that tracks the number of connections in active use. This extra deactivate led to active connection counts going negative.

However, removing this DeactivateConnection call is also not entirely correct. Part of the deactivation process is setting the "reset" flags via a call to the ResetConnection method. These flags determine the [Status](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-tds/ce398f9a-7d47-4ede-8f36-9dd6fc21ca43) value sent to the server when a connection is recycled out of the pool. Connections that are placed in the transacted pool may use status RESETCONNECTIONSKIPTRAN if they are participating in a distributed transaction. This gives extra guarantees that the server will not reset the connection's transaction state. When moving a connection out of the transacted pool and back into the main pool, we no longer want the server to maintain the association to the transaction because the transaction has ended and we want to use the connection for general purpose commands. Therefore, we need to downgrade the connection's status value to RESETCONNECTION. This way, the next time the connection is used its association with the transaction will be removed.

To appropriately set the reset flags without impacting the active connection counter, I exposed the ResetConnection method as internal so that it can be called directly by the pool. All other actions taken in DeactivateConnection are safe to skip at this point.

I hope to clean up this flow as part of adding transaction support to the new pool implementation, but it's out of scope for this PR.

## Issues

#3640 

## Testing

None of our tests cover metrics. Working on adding that as part of the new connection pool work.
In the meantime, this change will be verified manually.

Existing distributed transaction tests show that the reset flags are still reevaluated correctly.

### DIFF
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/TracingTests/MetricsTest.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/TracingTests/MetricsTest.cs
@@ -156,7 +156,7 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests
             }
         }
 
-        [ConditionalFact(typeof(DataTestUtility), nameof(DataTestUtility.AreConnStringsSetup))]
+        [ConditionalFact(typeof(DataTestUtility), nameof(DataTestUtility.AreConnStringsSetup), nameof(DataTestUtility.IsNotAzureSynapse))]
         public void StasisCounters_Functional()
         {
             var stringBuilder = new SqlConnectionStringBuilder(DataTestUtility.TCPConnectionString) { Pooling = false, Enlist = false };


paulmedynski (approved):

src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbConnectionClosed.cs
```diff
@@ -55,6 +56,9 @@ internal override bool TryOpenConnection(
             TaskCompletionSource<DbConnectionInternal> retry,
             DbConnectionOptions userOptions) =>
             TryOpenConnectionInternal(outerConnection, connectionFactory, retry, userOptions);
+
+        /// <inheritdoc/>
+        internal override void ResetConnection() => throw ADP.ClosedConnectionError();
```
benrr101, line 61: I kinda think this violates SOLID principles 🤔 I'm kinda wondering if we need to make this method internal on the DbConnectionInternal (and as a result, require implementation on the "do nothing" DbConnection types). I'm wondering instead if we can/should only have the method exposed on SqlInternalConnection with a cast to SqlInternalConnection to check if it can be called (we do that in other places too even if its kinda weird)

benrr101 (changes requested):
Overall, probably pretty decent. Just want to see what's the best choice for whether to make it abstract or only defined on the SqlInternalConnectionTds

src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbConnectionClosed.cs
mdaigle: I'd like to collapse down the whole DbInternalConnection->SqlInternalConnection->SqlInternalConnectionTds hierarchy and get rid of DbConnectionClosed. I don't know what value they give us and, like you said, they lead to a bunch of weird casts. But for this PR my goal was to follow existing patterns.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbConnectionInternal.cs
```diff
@@ -767,6 +767,13 @@ internal void PrePush(DbConnection expectedOwner)
         internal void RemoveWeakReference(object value) =>
             ReferenceCollection?.Remove(value);
 
+        /// <summary>
+        /// Idempotently resets the connection so that it may be recycled without leaking state.
+        /// May preserve transaction state if the connection is enlisted in a distributed transaction.
+        /// Should be called before the first action is taken on a recycled connection.
```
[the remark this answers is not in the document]
mdaigle: This is an interesting idea. We could consider resetting at activation as well.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/ProviderBase/DbConnectionClosed.cs
benrr101: I'm about to submit a PR for flattening SqlInternalConnection and SqlInternalConnectionTds 😄

benrr101 (dismissed):

benrr101 (approved):
🚢

Issue #1687, opened 2022-07-28
Microsoft.Data.SqlClient is not compatible with AssemblyLoadContext unloading


agocke:
See https://github.com/dotnet/runtime/issues/31377 for context. The fundamental issue is that unloadability is based on GC references, so SqlClient containing a strong references via event handlers and similar. The suggested fix is that for callbacks like the TimerCallback in Microsoft.Data.SqlClient.SqlConnectionFactory, support for listening to the Unload event could be added and these roots could be cleared, so that the app could be unloaded.

lcheunglci:
Thanks for identifying the issue. The MDS team is working on a new release.  I believe someone will follow up with this issue as soon as it's out.  In the meantime, is there a repro sample that I could use to try?

Wraith2:
I have made some changes to help unloading in the past such as https://github.com/dotnet/SqlClient/pull/913 but it needs a specific case to track into and isolate the roots. Without specific guidance on patterns that need to be followed to allow unloadability there's a lot of library to cover.

agocke:
@janvorli might have some ideas since he investigated the original bug

JRahnama:
@agocke thank you for opening the issue. 
@Wraith2 thanks again for the support. Will you be looking into this issue?

janvorli:
The things to look for are:
* TimerCallback instances referencing methods from the Microsoft.Data.SqlClient assembly
* GC handles and static members holding references to objects of types from the Microsoft.Data.SqlClient assembly.
* Threads executing code from the Microsoft.Data.SqlClient assembly at the unload time. E.g. threads that never exit and just loop forever.
* Async methods from the Microsoft.Data.SqlClient assembly that can be waiting for something to complete at the unload time.

It is difficult to create an exhaustive list, but these are likely to be the most frequent offenders.

Wraith2:
@JRahnama yes I will look into this.
@janvorli that's the sort of information I was looking for. I appreciate that there can't be a definitive list.

PR #3768, opened 2025-11-14
Cleanup | Convert SniLoadHandle to static class

edwardneal:
## Description

This PR fixes a minor correctness issue and might also contribute to #1687 - this field was one of several lingering on the heap when trying to unload a collectible AssemblyLoadContext.

The managed SNI's `SniLoadHandle` currently has a per-thread `LastError` property which references a `ThreadLocal<SniError>`. ThreadLocal is disposable, but we have no clear opportunity to dispose of it when the AssemblyLoadContext is unloaded.

The entire class is a singleton which only exposes a single meaningful property, `LastError`. I've converted SniLoadHandle into a static class, then replaced the `ThreadLocal` with a simpler `[ThreadStatic]` variable.

## Issues

Related to #1687. It doesn't fix the issue, but it's one of the two fixes which are easiest to pick out from a broader set of changes.

## Testing

Automated tests continue to pass.


paulmedynski:
@edwardneal  This will need a rebase/merge from main to pickup the new PR pipeline configs.

edwardneal:
Thanks @paulmedynski I've just merged main.

PR #3772, opened 2025-11-14
Perf | Use source generation to serialize user agent JSON

edwardneal:
## Description

#3582 added functionality to send a small user agent JSON structure to SQL Server. While this works, it also drags in a reflection-based dependency, and the populated reflection-based information is held in memory, preventing the AssemblyLoadContext from being unloaded.

This PR swaps this over to using the built-in [source-generated JSON serializer](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/source-generation-modes). This should improve performance slightly, and it's the second of a set of changes needed to enable AssemblyLoadContext unloading.

While I don't think this'll be a problem, it's worth noting that this means we'll need to compile the netfx library using the net6.0+ SDK.

## Issues

Builds on #3582. Contributes to #1687.

## Testing

Automated unit tests continue to function.

### DIFF
--- a/src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
+++ b/src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
@@ -995,6 +995,9 @@
     <Compile Include="$(CommonSourceRoot)Microsoft\Data\SqlClient\UserAgent\UserAgentInfoDto.cs">
       <Link>Microsoft\Data\SqlClient\UserAgent\UserAgentInfoDto.cs</Link>
     </Compile>
+    <Compile Include="$(CommonSourceRoot)Microsoft\Data\SqlClient\UserAgent\UserAgentInfoDtoSerializerContext.cs">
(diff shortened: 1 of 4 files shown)


paulmedynski (dismissed):

paulmedynski:
Changes look good, and we should merge this PR even though the design has changed back to pipe-delimited and we won't be using JSON for the User Agent.  Still worthwhile to have this in the commit history so we can refer to it in the future if necessary.

paulmedynski:
@edwardneal This will need a rebase/merge from main to pickup the new PR pipeline configs.

edwardneal:
Thanks @paulmedynski, I've just merged and pushed.

apoorvdeshmukh:
@edwardneal Can you rebase with main? I'll kickstart a fresh pipeline run.

edwardneal:
Thanks @apoorvdeshmukh, I've just rebased and force-pushed. With that said, I can see #3826 now, and I've got no objection if we want to skip this PR in favour of that one.

cheenamalhotra (approved):

PR #3773, opened 2025-11-14
Flatten | SqlInternalConnectionTds and SqlInternalConnection

benrr101:
## Description
This PR is part of an ongoing (undocumented) effort to clean up, and modernize the codebase. In this installment, we have the SqlInternalConnection and SqlInternalConnectionTds class. These are a remnant of when context connections existed - two classes inherited from SqlInternalConnection: SqlInternalConnectionTds and SqlInternalConnectionSmi (I think). The SMI version was removed a while ago as part of removing functionally dead codepaths. As such, there's no need for all this inheritance. We can merge SqlInternalConnectionTds into SqlInternalConnection.

At the same time, we also move SqlInternalConnectionTds into the Connection namespace, and rename it to SqlConnectionInternal to alphabetically organize it with SqlConnection.

Comments have been added to the PR to highlight key changes in the code.

## Issues
N/A

## Testing
Code continues to compile, CI will validate.


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Connection/SqlConnectionInternal.cs
```diff
@@ -1746,66 +2042,76 @@ protected override void Activate(Transaction transaction)
...
-            // MDAC 73598 - add brackets around database
-            database = SqlConnection.FixupDatabaseTransactionName(database); // @TODO: Should go to a utility method
-            Task executeTask = _parser.TdsExecuteSQLBatch(
-                $@"USE {database}",
-                ConnectionOptions.ConnectTimeout,
-                notificationRequest: null,
-                _parser._physicalStateObj,
-                sync: true);
+        protected override void CleanupTransactionOnCompletion(Transaction transaction) =>
+            DelegatedTransaction?.TransactionEnded(transaction);
```
edwardneal, line 2048: This movement is missing a comment, and it highlights a slight change in approach: rather than getting the value of a property into a local variable and null-coalescing on the local variable, we null-coalesce on the property directly. If the value of the property changes halfway through the operation, we'll be operating on two different values (or might get a NullReferenceException.)

This might be fine - SqlConnection isn't supposed to be used in a non-thread-safe manner, and this is the first way I can see which would trigger that behaviour - but I want to make sure it's being done intentionally.

benrr101: As far as I understand, the compile will implicitly create a temporary variable for this. Thus, we get the same behavior as before, but in four less lines :)

edwardneal: I've just checked in SharpLab - you're right!

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnection.cs
```diff
@@ -2697,10 +2695,17 @@ private static void ChangePassword(string connectionString, SqlConnectionString
             // note: This is the only case where we directly construct the internal connection, passing in the new password.
             // Normally we would simply create a regular connection and open it, but there is no other way to pass the
             // new password down to the constructor. This would have an unwanted impact on the connection pool.
-            SqlInternalConnectionTds con = null;
+            SqlConnectionInternal con = null;
```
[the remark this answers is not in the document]
benrr101: You're not wrong, but because the internal connection is immediately disposed after construction, the cleanest way to write it is with the manual disposal. I'm too unsure what magic happens with the pools when the internal connection is constructed to want to mess with when the internal connection is disposed (ie, via a `using` declaration)

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlDelegatedTransaction.cs
```diff
@@ -420,7 +421,7 @@ private SqlInternalConnection GetValidConnection()
         // Dooms connection and throws and error if not a valid, active, delegated transaction for the given
         //  connection. Designed to be called AFTER a lock is placed on the connection, otherwise a normal return
         //  may not be trusted.
-        private void ValidateActiveOnConnection(SqlInternalConnection connection)
+        private void ValidateActiveOnConnection(SqlConnectionInternal connection)
         {
             bool valid = _active && (connection == _connection) && (connection.DelegatedTransaction == this);
```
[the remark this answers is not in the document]
benrr101: Technically true, though I'm not sure the implications of making this change.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlTransaction.cs
```diff
@@ -289,7 +294,7 @@ internal void Zombie()
             // For Yukon, we have to defer "zombification" until we get past the users' next
             // rollback, else we'll throw an exception there that is a breaking change. Of course,
             // if the connection is already closed, then we're free to zombify...
-            if (_connection.InnerConnection is SqlInternalConnection internalConnection && !_isFromApi)
+            if (_connection.InnerConnection is SqlConnectionInternal && !_isFromApi)
```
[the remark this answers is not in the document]
benrr101: Uhhh what?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlInternalConnection.cs
```diff
@@ -132,7 +132,7 @@ internal bool HasLocalTransactionFromAPI
         /// <summary>
         /// Indicates whether the connection is currently enlisted in a transaction.
         /// </summary>
-        internal bool IsEnlistedInTransaction { get; private set; }
+        internal bool IsEnlistedInTransaction { get; set; }
```
mdaigle: Change this back to private set

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Connection/SqlConnectionInternal.cs
```diff
@@ -914,7 +914,7 @@ public override void Dispose()
                 _fConnectionOpen = false;
             }
 
-            base.Dispose();
+            _whereAbouts = null;
```
mdaigle, line 1035: Missing a call to base.Dispose()

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlDataReader.cs
```diff
@@ -5794,7 +5796,7 @@ private void PrepareAsyncInvocation(bool useSnapshot)
                 {
                     if (_connection?.InnerConnection is SqlInternalConnectionTds sqlInternalConnection)
                     {
-                        _snapshot = Interlocked.Exchange(ref sqlInternalConnection.CachedDataReaderSnapshot, null) ?? new Snapshot();
+                        _snapshot = sqlInternalConnection.CachedContexts.ClearDataReaderSnapshot() ?? new Snapshot();
```
mdaigle: Lines like this make me think that "pop" might make more sense than "clear". The ultimate intent often isn't to wipe out the cache, it's to check out the cache value.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Connection/SqlConnectionInternal.cs
```diff
@@ -346,8 +346,12 @@ internal SqlInternalConnectionTds(
             string accessToken = null,
             IDbConnectionPool pool = null,
             Func<SqlAuthenticationParameters, CancellationToken, Task<SqlAuthenticationToken>> accessTokenCallback = null,
-            SspiContextProvider sspiContextProvider = null) : base(connectionOptions)
+            SspiContextProvider sspiContextProvider = null)
```
mdaigle, line 349: Keep this call to the base constructor, we still inherit from DbConnectionInternal

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnection.cs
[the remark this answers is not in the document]
mdaigle: It looks like this method is opening a new connection (including a round trip) just to seed a new connection pool. Then it's explicitly clearing the old pool so that it's no longer used...
I want to look into this more because it feels fishy.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionFactory.cs
```diff
@@ -645,15 +645,25 @@ protected virtual DbConnectionInternal CreateConnection(
 
                 if (pool == null || (pool != null && pool.Count <= 0))
                 { // Non-pooled or pooled and no connections in the pool.
-                    SqlInternalConnectionTds sseConnection = null;
+                    SqlConnectionInternal sseConnection = null;
```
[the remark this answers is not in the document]
mdaigle: This also looks odd to me. Will investigate...

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Connection/SqlConnectionInternal.cs
benrr101: While true, the call to SqlInternalConnection constructor was to the parameter-less constructor. So, when we flatten, this becomes a "redundant" call to the parameter-less base constructor, which in itself calls the parameter-full base constructor with hard-coded values.

Options for resolution:
1) Remove "redundant" call to parameter-less base constructor (ie, what is in the PR right now)
2) Put the parameter-less base constructor call back in
3) Remove parameter-less base constructor and just call the parameterized constructor.

Take your pick.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlDataReader.cs
benrr101: Tbh these names came from ai. I think the methods are really more of a "lease" than a "pop". Pop to me implies a stack that is being taken from. "Lease" isn't perfect either since it implies it'll be put back (and that there are mechanisms to enforce that). Maybe "Take" could be a compromise?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Connection/SqlConnectionInternal.cs
mdaigle: Makes sense. What we have now looks good.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlDataReader.cs
mdaigle: "Take" would make sense to me!

paulmedynski (approved):

mdaigle (approved):

PR #3774, opened 2025-11-17
Enable code cov to include net9.0 result files

cheenamalhotra:
Final Code coverage was not merging net9.0 test results, this PR enables the same.

### DIFF
--- a/eng/pipelines/dotnet-sqlclient-ci-core.yml
+++ b/eng/pipelines/dotnet-sqlclient-ci-core.yml
@@ -56,7 +56,7 @@ parameters:
 - name: codeCovTargetFrameworks
   displayName: 'Code Coverage Target Frameworks'
   type: object
-  default: [net462, net8.0]
+  default: [net462, net8.0, net9.0]
 
 - name: buildType
   displayName: 'Build Type'


eng/pipelines/dotnet-sqlclient-ci-core.yml
```diff
@@ -56,7 +56,7 @@ parameters:
 - name: codeCovTargetFrameworks
   displayName: 'Code Coverage Target Frameworks'
   type: object
-  default: [net462, net8.0]
+  default: [net462, net8.0, net9.0]
```
paulmedynski, line 59: @priyankatiwari08 - You will need to add `net10.0` here in your PR.

eng/pipelines/dotnet-sqlclient-ci-core.yml
```diff
@@ -56,7 +56,7 @@ parameters:
 - name: codeCovTargetFrameworks
```
mdaigle, line 56: We could also remove this parameter entirely and publish test results and coverage for all target frameworks. That way we wont have to remember to update it in the future.

cheenamalhotra: Let's plan it as we do more code coverage investigations.

PR #3776, opened 2025-11-17
[6.1] Fix 3640 | Remove extra connection deactivation.

mdaigle:
## Description

When I originally implemented #3019, due to a misunderstanding about when connections are activated/deactivated, I added an extraneous call to DbConnectionInternal.DeactivateConnection as part of the PutObjectFromTransactedPool. Connections are **always** deactivated at the time they are returned to the pool via the ReturnInternalConnection method and do not need to be activated/deactivated as they move between the main pool and the transacted pool. The activate and deactivate methods modify a counter value that tracks the number of connections in active use. This extra deactivate led to active connection counts going negative.

However, removing this DeactivateConnection call is also not entirely correct. Part of the deactivation process is setting the "reset" flags via a call to the ResetConnection method. These flags determine the [Status](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-tds/ce398f9a-7d47-4ede-8f36-9dd6fc21ca43) value sent to the server when a connection is recycled out of the pool. Connections that are placed in the transacted pool may use status RESETCONNECTIONSKIPTRAN if they are participating in a distributed transaction. This gives extra guarantees that the server will not reset the connection's transaction state. When moving a connection out of the transacted pool and back into the main pool, we no longer want the server to maintain the association to the transaction because the transaction has ended and we want to use the connection for general purpose commands. Therefore, we need to downgrade the connection's status value to RESETCONNECTION. This way, the next time the connection is used its association with the transaction will be removed.

To appropriately set the reset flags without impacting the active connection counter, I exposed the ResetConnection method as internal so that it can be called directly by the pool. All other actions taken in DeactivateConnection are safe to skip at this point.

I hope to clean up this flow as part of adding transaction support to the new pool implementation, but it's out of scope for this PR.

## Issues

#3640 

## Testing

None of our tests cover metrics. Working on adding that as part of the new connection pool work.
In the meantime, this change will be verified manually.

Existing distributed transaction tests show that the reset flags are still reevaluated correctly.

### DIFF
--- a/src/Microsoft.Data.SqlClient/tests/ManualTests/TracingTests/MetricsTest.cs
+++ b/src/Microsoft.Data.SqlClient/tests/ManualTests/TracingTests/MetricsTest.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Diagnostics;
 using System.Reflection;
+using System.Threading.Tasks;
 using System.Transactions;
 using Xunit;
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/WaitHandleDbConnectionPool.cs
```diff
@@ -1696,6 +1694,7 @@ public void PutObjectFromTransactedPool(DbConnectionInternal obj)
 
             if (State is Running && obj.CanBePooled)
             {
+                obj.ResetConnection();
```
cheenamalhotra, line 1697: Is it safe to Not reset it outside the condition block?

src/Microsoft.Data.SqlClient/tests/ManualTests/TracingTests/MetricsTest.cs
```diff
@@ -177,6 +177,73 @@ public void StasisCounters_Functional()
...
+        {
+            // This test verifies that the active connection count metric never goes negative
+            // when connections are returned to the pool while enlisted in a transaction.
+            // This is a regression test for issue #3640 where an extra DeactivateConnection
+            // call was causing the active connection count to go negative.
+
+            // Arrange
+            var stringBuilder = new SqlConnectionStringBuilder(DataTestUtility.TCPConnectionString)
+            {
+                Pooling = true,
```
cheenamalhotra, line 192: Can you toggle the property via parameters and test it for non-pooled connections as well?

src/Microsoft.Data.SqlClient/tests/ManualTests/TracingTests/MetricsTest.cs
```diff
@@ -177,6 +177,73 @@ public void StasisCounters_Functional()
...
+            long initialActiveHardConnections = SqlClientEventSourceProps.ActiveHardConnections;
+            long initialActiveConnections = SqlClientEventSourceProps.ActiveConnections;
+
+            // Act and Assert
+            // Verify counters at each step in the lifecycle of a transacted connection
+            using (var txScope = new TransactionScope())
+            {
+                using (var conn = new SqlConnection(stringBuilder.ToString()))
+                {
+                    conn.Open();
```
cheenamalhotra, line 211: Add test for async operations as well.

src/Microsoft.Data.SqlClient/tests/ManualTests/TracingTests/MetricsTest.cs
```diff
@@ -177,6 +177,73 @@ public void StasisCounters_Functional()
...
+                // Completing the transaction after the connection is closed ensures that the connection
+                // is in the transacted pool at the time the transaction ends. This verifies that the
+                // transition from the transacted pool back to the main pool properly updates the counters.
+                txScope.Complete();
+            }
+
+            if (SupportsActiveConnectionCounters)
+            {
+                Assert.Equal(initialActiveSoftConnections, SqlClientEventSourceProps.ActiveSoftConnections);
+                Assert.Equal(initialActiveHardConnections+1, SqlClientEventSourceProps.ActiveHardConnections);
```
cheenamalhotra: Nit: Fix Formatting

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/WaitHandleDbConnectionPool.cs
mdaigle: It would be safe. I placed it here because we _need_ the reset when we're putting a connection back in the main pool, this highlights that dependency.

src/Microsoft.Data.SqlClient/tests/ManualTests/TracingTests/MetricsTest.cs
mdaigle: This metric is only tracked for pooled connections.

cheenamalhotra (approved):

PR #3781, opened 2025-11-18
Merge | Remove Extraneous Files

benrr101:
## Description
This PR is the first in the next series towards a single MDS csproj. There are a handful of files that were left behind in the separate projects that could be easily swept up.

* netcore/MatchingRefApiCompatBaseline.txt - deleted, not referenced anywhere.
* netcore/Microsoft.Data.SqlClient.TypeForwards.cs - moved to TypeForwards.netcore.cs in common project
* netcore/.editorconfig - removed in favor of solution-wide .editorconfig
* netcore/PinvokeAnalyzerExceptionList.analyzerdata.netcoreapp - removed, (if this does what I think it does) replaced with APIScan
* netfx/.editorconfig - removed in favor of solution-wide .editorconfig
* netfx/GenerateThisAssemblyCs.targets - removed in favor of using the same one as netcore, only difference is it adds the NugetPackageVersion constant to the netfx assembly. This seems acceptable because it is internal only.

## Issues
N/A, but continues from #1261 which was sort of the "common project" issue.

## Testing
Projects still build, but CI (and likely an official build) will need to be used to verify.


src/Microsoft.Data.SqlClient/netfx/src/Microsoft.Data.SqlClient.csproj
```diff
@@ -1080,7 +1080,7 @@
     <PackageReference Include="System.ValueTuple" />
     <PackageReference Include="System.Threading.Channels" />
   </ItemGroup>
-  <Import Project="$(NetFxSource)tools\targets\GenerateThisAssemblyCs.targets" />
+  <Import Project="$(ToolsDir)targets\GenerateThisAssemblyCs.targets" />
   <Import Project="$(NetFxSource)tools\targets\GenerateAssemblyRef.targets" />
   <Import Project="$(NetFxSource)tools\targets\GenerateAssemblyInfo.targets" />
```
cheenamalhotra, line 1085: Aren't these moved yet?

PR #3791, opened 2025-11-20
Perf | Reuse XmlWriterSettings, eliminate MemoryCacheEntryOptions allocations

edwardneal:
## Description

This is a minor set of improvements in memory usage which I noticed while looking at the XML access pathway and thinking about some of the layers of caching within SqlClient.

The first commit caches and reuses `XmlWriterSettings` objects. As per the [remarks](https://learn.microsoft.com/en-us/dotnet/api/system.xml.xmlwriter.create?view=net-9.0#remarks) in the documentation, these can be reused and so don't need to be allocated every time. I've promoted these to a `static readonly` field in the `SqlStreamingXml` class because this is only used when reading XML data. The other places this is used are `TdsParser` and `ValueUtilsSmi`, and these don't necessarily need to use XML functionality so I lazily allocate them on first use.

In the second commit, we eliminate allocations of `MemoryCacheEntryOptions` objects. These are essentially DTOs which are used to modify the expiration times of entries in various MemoryCaches. Rather than allocating a new object, we can use a different overload which accepts the expiration timespan directly.

## Issues

None.

## Testing

Local tests of these pathways pass - could someone run CI please?

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/EnclaveProviderBase.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/EnclaveProviderBase.cs
@@ -68,7 +68,6 @@ namespace Microsoft.Data.SqlClient
     {
         #region Constants
         private const int NonceSize = 256;
-        private const int ThreadRetryCacheTimeoutInMinutes = 10;
         private const int LockTimeoutMaxInMilliseconds = 15 * 1000; // 15 seconds
         #endregion
 
@@ -85,6 +84,7 @@ namespace Microsoft.Data.SqlClient
 
         // It is used to save the attestation url and nonce value across API calls
         protected static readonly MemoryCache ThreadRetryCache = new MemoryCache(new MemoryCacheOptions());
+        private static readonly TimeSpan s_threadRetryCacheTimeout = TimeSpan.FromMinutes(10);
         #endregion
 
         #region protected methods
@@ -168,7 +168,7 @@ namespace Microsoft.Data.SqlClient
                     }
 
                     ThreadRetryCache.Set<string>(Thread.CurrentThread.ManagedThreadId.ToString(), retryThreadID,
-                        absoluteExpirationRelativeToNow: TimeSpan.FromMinutes(ThreadRetryCacheTimeoutInMinutes));
+                        absoluteExpirationRelativeToNow: s_threadRetryCacheTimeout);
                 }
             }
         }
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/EnclaveSessionCache.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/EnclaveSessionCache.cs
@@ -12,7 +12,7 @@ namespace Microsoft.Data.SqlClient
     internal class EnclaveSessionCache
     {
         // Cache timeout of 8 hours to be consistent with JWT validity.
-        private const int EnclaveCacheTimeOutInHours = 8;
+        private static readonly TimeSpan s_enclaveCacheTimeout = TimeSpan.FromHours(8);
 
         private readonly MemoryCache enclaveMemoryCache = new MemoryCache(new MemoryCacheOptions());
         private readonly object enclaveCacheLock = new object();
@@ -63,7 +63,7 @@ namespace Microsoft.Data.SqlClient
             {
                 enclaveSession = new SqlEnclaveSession(sharedSecret, sessionId);
                 enclaveMemoryCache.Set<SqlEnclaveSession>(cacheKey, enclaveSession,
-                    absoluteExpirationRelativeToNow: TimeSpan.FromHours(EnclaveCacheTimeOutInHours));
+                    absoluteExpirationRelativeToNow: s_enclaveCacheTimeout);
                 counter = Interlocked.Increment(ref _counter);
             }
 
--- a/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/ValueUtilsSmi.cs
+++ b/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/ValueUtilsSmi.cs
@@ -37,7 +37,18 @@ namespace Microsoft.Data.SqlClient.Server
         private const int DefaultBinaryBufferSize = 4096;  // Size of the buffer used to read input parameter of type Stream
         private const int DefaultTextBufferSize = 4096; // Size of the buffer (in chars) user to read input parameter of type TextReader
 
-        private static XmlWriterSettings s_writerSettings;
+        // @TODO: Replace field with the `field` keyword when LangVersion >= 14.
+        private static XmlWriterSettings s_xmlWriterSettings;
+
+        private static XmlWriterSettings XmlWriterSettings =>
+            s_xmlWriterSettings ??= new XmlWriterSettings()
+            {
+                // Don't close the memory stream
+                CloseOutput = false,
+                ConformanceLevel = ConformanceLevel.Fragment,
+                Encoding = System.Text.Encoding.Unicode,
+                OmitXmlDeclaration = true
+            };
 
         //
         //  User-visible semantics-laden Getter/Setter support methods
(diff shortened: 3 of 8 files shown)


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/EnclaveProviderBase.cs
```diff
@@ -167,11 +166,7 @@ protected void GetEnclaveSessionHelper(EnclaveSessionParameters enclaveSessionPa
                         retryThreadID = Thread.CurrentThread.ManagedThreadId.ToString();
                     }
 
-                    MemoryCacheEntryOptions options = new MemoryCacheEntryOptions
-                    {
-                        AbsoluteExpirationRelativeToNow = TimeSpan.FromMinutes(ThreadRetryCacheTimeoutInMinutes)
-                    };
-                    ThreadRetryCache.Set<string>(Thread.CurrentThread.ManagedThreadId.ToString(), retryThreadID, options);
+                    ThreadRetryCache.Set<string>(Thread.CurrentThread.ManagedThreadId.ToString(), retryThreadID, absoluteExpirationRelativeToNow: TimeSpan.FromMinutes(10));
```
cheenamalhotra: Continue to use constant for timeout

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/EnclaveSessionCache.cs
```diff
@@ -18,9 +18,6 @@ internal class EnclaveSessionCache
         // given that for Always Encrypted scenarios, the server is considered an "untrusted" man-in-the-middle.
         private long _counter;
 
-        // Cache timeout of 8 hours to be consistent with jwt validity.
-        private static int enclaveCacheTimeOutInHours = 8;
```
cheenamalhotra, line 22: Continue using constant (convert it to const instead)

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SignatureVerificationCache.cs
```diff
@@ -76,11 +76,7 @@ internal void AddSignatureVerificationResult(string keyStoreName, string masterK
             TrimCacheIfNeeded();
 
             // By default evict after 10 days.
-            MemoryCacheEntryOptions options = new MemoryCacheEntryOptions
-            {
-                AbsoluteExpirationRelativeToNow = TimeSpan.FromDays(10)
-            };
-            _cache.Set<bool>(cacheLookupKey, result, options);
+            _cache.Set<bool>(cacheLookupKey, result, absoluteExpirationRelativeToNow: TimeSpan.FromDays(10));
```
cheenamalhotra: Introduce const for magic number

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlQueryMetadataCache.cs
```diff
@@ -236,15 +236,13 @@ internal void AddQueryMetadata(SqlCommand sqlCommand, bool ignoreQueriesWithRetu
             }
 
             // By default evict after 10 hours.
-            MemoryCacheEntryOptions options = new MemoryCacheEntryOptions
-            {
-                AbsoluteExpirationRelativeToNow = TimeSpan.FromHours(10)
-            };
-            _cache.Set<Dictionary<string, SqlCipherMetadata>>(cacheLookupKey, cipherMetadataDictionary, options);
+            TimeSpan expirationPeriod = TimeSpan.FromHours(10);
```
cheenamalhotra: Introduce const for magic number

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/VirtualSecureModeEnclaveProviderBase.cs
```diff
@@ -211,11 +211,7 @@ private X509Certificate2Collection GetSigningCertificate(string attestationUrl,
                     throw SQL.AttestationFailed(string.Format(Strings.GetAttestationSigningCertificateFailedInvalidCertificate, attestationUrl), exception);
                 }
 
-                MemoryCacheEntryOptions options = new MemoryCacheEntryOptions
-                {
-                    AbsoluteExpirationRelativeToNow = TimeSpan.FromDays(1)
-                };
-                rootSigningCertificateCache.Set<X509Certificate2Collection>(attestationUrl, certificateCollection, options);
+                rootSigningCertificateCache.Set<X509Certificate2Collection>(attestationUrl, certificateCollection, absoluteExpirationRelativeToNow: TimeSpan.FromDays(1));
```
cheenamalhotra: Introduce const for magic number

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -12768,13 +12771,9 @@ private async Task WriteXmlFeed(XmlDataFeed feed, TdsParserStateObject stateObj,
...
-            XmlWriterSettings writerSettings = new XmlWriterSettings();
-            writerSettings.CloseOutput = false; // don't close the memory stream
-            writerSettings.ConformanceLevel = ConformanceLevel.Fragment;
-            if (_asyncWrite)
-            {
-                writerSettings.Async = true;
-            }
+            XmlWriterSettings writerSettings = _asyncWrite
+                ? (s_asyncWriterSettings ??= new() { CloseOutput = false, ConformanceLevel = ConformanceLevel.Fragment, Async = true })
+                : (s_syncWriterSettings ??= new() { CloseOutput = false, ConformanceLevel = ConformanceLevel.Fragment });
```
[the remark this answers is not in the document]
edwardneal: `SqlConnection` objects should not be shared across threads, so this scenario would only occur under invalid usage.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/EnclaveProviderBase.cs
edwardneal: Thanks - I've reintroduced the various constants in this comment and elsewhere.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/EnclaveSessionCache.cs
```diff
@@ -11,16 +11,16 @@ namespace Microsoft.Data.SqlClient
     // Maintains a cache of SqlEnclaveSession instances
     internal class EnclaveSessionCache
     {
+        // Cache timeout of 8 hours to be consistent with JWT validity.
+        private const int EnclaveCacheTimeOutInHours = 8;
```
paulmedynski: Would it be better to specify these constants as TimeSpans and avoid the conversion at each point of use?  These are only ever used in TimeSpan.FromXYZ() calls.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/ValueUtilsSmi.cs
```diff
@@ -35,7 +35,9 @@ internal static class ValueUtilsSmi
 
         // Constants
         private const int DefaultBinaryBufferSize = 4096;  // Size of the buffer used to read input parameter of type Stream
-        private const int DefaultTextBufferSize = 4096; // Size of the buffer (in chars) user to read input parameter of type TextReader       
+        private const int DefaultTextBufferSize = 4096; // Size of the buffer (in chars) user to read input parameter of type TextReader
+
+        private static XmlWriterSettings s_writerSettings;
```
paulmedynski: If this were a static readonly property, then the initialization code would live here with the definition, rather than 3400 lines away:
```
private static WriterSettings
{
    get
    {
        field ??= new XmlWriterSettings(...);
        return field;
    }
}
```
I guess you'll still need the explicit backing field, until we start using .NET 10 SDK and C# 14.  Maybe leave a TODO and we can optimize in a few weeks?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParser.cs
```diff
@@ -41,6 +41,9 @@ namespace Microsoft.Data.SqlClient
     internal sealed partial class TdsParser
     {
         private static int _objectTypeCount; // EventSource counter
+        private static XmlWriterSettings s_asyncWriterSettings;
```
paulmedynski: Same here - you can keep the definition and lazy initialization together with a property.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/ValueUtilsSmi.cs
edwardneal: After a little experimentation, I like the way that looks; the `field` keyword will make this even neater. TdsParser looks clumsier, but I think that's largely down to the size/style of the file.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/EnclaveSessionCache.cs
edwardneal: I think so, yes. It also gets rid of the InHours/InDays/etc. suffixes, which seem a little pointless in these contexts. I've just fixed this (along with the lazy field initialisation) and pushed the commit.

edwardneal:
This PR is not stale.

cheenamalhotra (approved):

PR #3794, opened 2025-11-21
Tests | Widen SqlVector test criteria, roundtrip additional values in tests

### LINKED ISSUE #3789: Extend tests for float.MinValue, 1.01f, float.NegativeZero

edwardneal:
## Description

This PR does two things:
1. Enables all `SqlVector` tests to be run against a database server which supports the `vector` type (not just Azure SQL instances.)
2. Adds three unusual values (`float.MinValue`, a value which can't be represented cleanly as a float, and negative zero) to demonstrate that they roundtrip, as per my comments on the original PR.

## Issues

Fixes #3789. Builds on my comments on the original PR #3433.

## Testing

`SqlVector` tests continue to pass against a local SQL 2025 instance and a SQL Azure instance.


src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
```diff
@@ -388,6 +388,29 @@ public static bool IsSupportedDataClassification()
...
         }
 
+        /// <summary>
+        /// Determines whether the SQL Server supports the 'vector' data type.
+        /// </summary>
+        /// <remarks>This method attempts to connect to the SQL Server and check for the existence of the
+        /// 'vector' data type. If a connection cannot be established or an error occurs during the query, the method
+        /// returns <see langword="false"/>.</remarks>
+        /// <returns><see langword="true"/> if the 'vector' data type is supported; otherwise, <see langword="false"/>.</returns>
+        public static bool IsSupportedSqlVector()
```
cheenamalhotra: Can you convert this into a property, and only query database on its initialization?

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
cheenamalhotra: I would also change this to a property and reduce server calls.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
edwardneal: Thanks. I've done this (along with `IsSupportedDataClassification` as you mentioned in the other comment, although I've changed this method's approach slightly to use `OBJECT_ID` rather than to wait for an error.)

I've also cleaned up the test conditions slightly, since both properties will return false if the TCP connection string is unspecified.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
```diff
@@ -351,6 +365,28 @@ public static bool IsDatabasePresent(string name)
...
 
+        public static bool IsObjectPresent(string objectName)
+        {
+            using SqlConnection connection = new(TCPConnectionString);
+            using SqlCommand command = new("SELECT OBJECT_ID(@name)", connection);
+
+            connection.Open();
+            command.Parameters.AddWithValue("@name", objectName);
+
+            return command.ExecuteScalar() is not DBNull;
```
paulmedynski, line 449: We're going to be moving away from dynamically discovering server capabilities at runtime, and more towards static configuration that simply knows what capabilities the server under test supports.  Something like "run these tests when using this server since we know it supports everything those tests are testing".  This will eliminate a ton of lag in the tests (once for test discovery, once for test execution), and allow us to validate that exactly the tests we expected to run have actually been run.  Right now we don't have any idea which tests are run, which are skipped, which tests _should_ be run, etc.  Just FYI - no need to make any changes to this PR.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
```diff
@@ -129,17 +132,28 @@ public static string SQLServerVersion
...
-                if (!string.IsNullOrEmpty(TCPConnectionString))
-                {
-                    s_isTDS8Supported = GetSQLServerStatusOnTDS8(TCPConnectionString);
-                }
-                return s_isTDS8Supported;
-            }
-        }
+        public static bool IsTDS8Supported =>
+            IsTCPConnStringSetup() &&
+                GetSQLServerStatusOnTDS8(TCPConnectionString);
```
paulmedynski, line 138: This is still querying the server every time this property is accessed.  Can you query the server once and store the result for all future access?

This would be a great use of the new `field` keyword if we were already using the .NET 10 SDK:

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/field

src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/VectorTest/NativeVectorFloat32Tests.cs
```diff
@@ -17,7 +17,7 @@ namespace Microsoft.Data.SqlClient.ManualTesting.Tests.SQL.VectorTest
     public static class VectorFloat32TestData
     {
         public const int VectorHeaderSize = 8;
-        public static float[] testData = new float[] { 1.1f, 2.2f, 3.3f };
+        public static float[] testData = new float[] { 1.1f, 2.2f, 3.3f, 1.01f, float.MinValue, -0.0f };
```
paulmedynski, line 20: Nice!  Negative zero is my favourite :)

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
edwardneal: I thought about doing that, but the behaviour wouldn't be correct unfortunately. This is only used in the [`CertificateTest`](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectionTestWithSSLCert/CertificateTest.cs) class; the constructor for this class runs the PowerShell script [`GenerateSelfSignedCertificate.ps1`](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/tests/ManualTests/SQL/ConnectionTestWithSSLCert/GenerateSelfSignedCertificate.ps1).

That PowerShell script changes whether TDS8 is supported; it generates a new certificate with the right ACL for the SQL Server service to use it, forces Windows to trust it, then configures SQL Server to use it. The value of the check would vary depending upon whether anything else has read the property before that caching occurred.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
edwardneal: That actually saves a fair amount of effort, so thanks. I've been a little unhappy with the way that `DataTestUtility` has evolved into a catch-all class, and was thinking about splitting it up: one to determine/cache server capabilities, another to centralise and store the test conditions, and others (maybe in the Common project) to store the general utilities. I'll leave that alone for now though.

src/Microsoft.Data.SqlClient/tests/ManualTests/DataCommon/DataTestUtility.cs
paulmedynski: I'm not sure I understand.  The TCPConnString is constant for all tests in the test run, so the server either supports TDS8 for all tests in the run, or it doesn't.  I'm not sure how that affects anything in the ps1 script - I don't see it performing any checks for TDS8.

edwardneal: TCPConnString is constant, but the tests in `CertificateTest` only run when that connection string points to localhost. The script is run before the tests, and it dynamically reconfigures the local SQL Server instance to meet all of the prerequisites for TDS8 - if we make `IsTDS8Supported` cache its result and a future test runs prior to a test from CertificateTest, we could return a stale value.

That can't happen right now (CertificateTest is the only test which uses this condition) but it'd cause a problem if we use it from any other test.

paulmedynski: Ok gotcha - thanks!  This is another good reason to avoid these runtime capability checks entirely.

paulmedynski:
@edwardneal - Looks like some conflicts need to be resolved here.

edwardneal:

cheenamalhotra (approved):

PR #3798, opened 2025-11-21
Fix code coverage job disk full

paulmedynski:
## Description

The code coverage jobs are running out of disk space.  They appear to consume upwards of 12GB of space to merge/convert  3GB of coverage logs from the various test jobs.  We can diagnose why that is later.  For now, I have:
- Switched the coverage job to use the ADO-MMS22-CodeCov 1ES image rather than a generic Azure Pipelines image.
  - The generic images have 14GB of disk space.
  - Our custom 1ES image has much more space.
- Removed unnecessary parameters/variables for code coverage job.
- Added debug output to help see disk usage throughout the job.

Successful coverage job run: https://sqlclientdrivers.visualstudio.com/public/SqlDevX%20(public)/_build/results?buildId=131331&view=logs&s=78db0df0-7afe-532d-b7fe-bdf808bbfe01

### DIFF
--- a/eng/pipelines/common/templates/jobs/ci-code-coverage-job.yml
+++ b/eng/pipelines/common/templates/jobs/ci-code-coverage-job.yml
@@ -15,6 +15,14 @@ parameters:
     type: boolean
     default: false
 
+  # The pool image to use.
+  - name: image
+    type: string
+
+  # The agent pool name.
+  - name: pool
+    type: string
+
   # Array of target frameworks to process code coverage for:
   #
   #   e.g. [net462, net8.0, net9.0]
@@ -31,9 +39,13 @@ jobs:
   - job: CodeCoverage
     displayName: Publish Code Coverage
 
-    pool: 
-      name: Azure Pipelines
-      vmImage: windows-2022
+    pool:
+      name: ${{ parameters.pool }}
+      ${{ if eq(parameters.pool, 'Azure Pipelines') }}:
+        vmImage: ${{ parameters.image }}
+      ${{ else }}:
+        demands:
+        - imageOverride -equals ${{ parameters.image }}
 
     variables:
       netFxDir: $(Build.SourcesDirectory)\coverageNetFx
--- a/eng/pipelines/dotnet-sqlclient-ci-core.yml
+++ b/eng/pipelines/dotnet-sqlclient-ci-core.yml
@@ -158,6 +158,8 @@ stages:
         - template: common/templates/jobs/ci-code-coverage-job.yml@self
           parameters:
             debug: ${{ parameters.debug }}
+            image: ADO-MMS22-CodeCov
+            pool: ${{ parameters.defaultPoolName }}
             targetFrameworks: ${{ parameters.codeCovTargetFrameworks }}
 
 # test stages configurations


eng/pipelines/common/templates/jobs/ci-code-coverage-job.yml
```diff
@@ -3,67 +3,82 @@
...
 parameters:
+  
+  # True to include debug steps.
   - name: debug
     type: boolean
     default: false
 
-  - name: upload
-    type: string
-    default: $(ci_var_uploadTestResult)
```
paulmedynski, line 13: This is a variable defined in the Classic UI pipeline settings.  We never use it, so I just made this a normal 'true' default here, and the new pipelines can decide whether or not that is a useful parameter.

We can't remove the Classic UI variable though, since the older branches still depend on it :(

eng/pipelines/common/templates/jobs/ci-code-coverage-job.yml
```diff
@@ -3,67 +3,82 @@
...
+      name: Azure Pipelines
+      vmImage: windows-2022
 
-    steps:
+    variables:
+      netFxDir: $(Build.SourcesDirectory)\coverageNetFx
+      netCoreDir: $(Build.SourcesDirectory)\coverageNetCore
 
-      # We don't need the source code for this job, so suppress checkout.
-      - checkout: none
```
paulmedynski, line 41: This wasn't true - we do use the source code below in the report generator command-line arguments.

eng/pipelines/common/templates/jobs/ci-code-coverage-job.yml
```diff
@@ -144,20 +155,28 @@ jobs:
...
-          #download Codecov CLI
-          $ProgressPreference = 'SilentlyContinue' 
-          Invoke-WebRequest -Uri https://cli.codecov.io/latest/windows/codecov.exe -Outfile codecov.exe
-          
-          ./codecov --verbose upload-process --fail-on-error -t $(CODECOV_TOKEN) -f "coveragereportNetFx\Cobertura.xml" -F netfx
-          ./codecov --verbose upload-process --fail-on-error -t $(CODECOV_TOKEN)  -f "coveragereportNetCore\Cobertura.xml" -F netcore
-          ./codecov --verbose upload-process --fail-on-error -t $(CODECOV_TOKEN)  -f "coveragereportAddOns\Cobertura.xml" -F addons
-        displayName: 'Upload to CodeCov'
-        condition: and(succeeded(), eq(variables['uploadTestResult'], 'true'))
+      - ${{if eq(parameters.upload, true)}}:
```
paulmedynski, line 185: Prefer compile-time conditionals to runtime `condition:` checks.  The former omits the step entirely when false.  The latter shows the step in the job list, but skips it, which is just clutter.

eng/pipelines/dotnet-sqlclient-ci-core.yml
```diff
@@ -161,16 +158,7 @@ stages:
         - template: common/templates/jobs/ci-code-coverage-job.yml@self
           parameters:
             debug: ${{ parameters.debug }}
-            downloadArtifactsSteps:
```
paulmedynski, line 164: Moved this into the ci-code-coverage-job template.  It doesn't need to be a parameter - we always download the same artifacts.

cheenamalhotra:
Do you have a successful run with this PR? Link in description if possible.

priyankatiwari08 (approved):

Issue #3356, opened 2025-05-19
Feature | Redesign the SqlClient Connection Pool to Improve Performance and Async Support


mdaigle:
### Discussed in https://github.com/dotnet/SqlClient/discussions/2612

<div type='discussions-op-text'>

<sup>Originally posted by **mdaigle** June 26, 2024</sup>

POC code: #3211 

<details open>
<summary> Work Plan</summary> 

1. [x] Scaffolding
    a. #3352 
    b. #3380 
    c. #3396 
1. [x] Get/Return connections
    a. #3404 
1. [x] Clear
    a. #4194 
1. [x] Context aware timeouts
    a. #4235
    b. #4237
    c. #4261 
    d. #4270
1. [x] Pool pruning (with clear counter)
    a. #1881 
    a. #4304 
1. [x] Configurable idle timeout
    a. #343
    a. #4295  
1. [x] Shutdown
    a. #4302 
1. [x] Rate limiting connection opens
    a. #4395 
    a. #4396 
1. [x] Warmup
    a. #4452 
1. [x] Connection resiliency (replace connection)
    a. #4415 
    b. #4429 
1. [x] Transaction support
    a. #3743 
    b. #3746 
    d. #3805 
    e. #4487 
1. [x] Reclaim leaked connections 
    a. #4490 / #4529
1. [x] Tracing and metrics
    a. #4504 
    b. #4505
        i. #3545 
1. [ ] Benchmarking and performance testing
    a. #3667 
    b. #4459 
</details>

<details open>
<summary>App context switches</summary>
- UseConnectionPoolV2
- Legacy
- 
</details>

<details open>
<summary>Design Document</summary>

# Design Document: `ChannelDbConnectionPool`
## Problem Statement
The current connection pool implementation is slow to open new connections and does not follow async best practices.

Connection opening is serialized, causing delays when multiple new connections are required simultaneously. This is done using a semaphore to rate limit connection creation. When multiple new connections are requested, they queue up waiting for the semaphore. Once acquired, the thread opens the connection and releases the semaphore, allowing the next thread to proceed. This approach was initially designed to prevent overwhelming the server but can lead to significant delays, especially in high-latency environments.

Async requests are also serialized through an additional queue. When an async request is made, it is added to a queue, and a reader thread processes these requests one by one. This method was chosen due to the lack of async APIs in native SNI, resulting in synchronous handling of async requests on a dedicated thread.

## Design Goals
- Enable parallel connection opening.
[...]

PR #3686, opened 2025-10-13
Adding support for .NET 10

priyankatiwari08:
## Description

As part of adding support for .NET 10 below changes has been done:
1. **Added conditional logic for PackageReference entries** for assemblies which are now included in the .NET 10 Base Class Library. Without this change, keeping them caused runtime redundancy and potential NU1510 errors during SDK pruning:

    - Sytem.Text.Json
    - System.Text.Encodings.Web
    - System.Buffers
    - System.Memory
    - System.Runtime.InteropServices.RuntimeInformation

3. **Enabled multi-targeting for .NET 10 in test projects** to ensure compatibility and proper coverage.

## Issues
[https://github.com/dotnet/SqlClient/issues/3576](https://github.com/dotnet/SqlClient/issues/3576)
Customer issue: [https://github.com/dotnet/SqlClient/discussions/3522](https://github.com/dotnet/SqlClient/discussions/3522)

PR #3805, opened 2025-11-24
Test | Connection pool transaction tests

mdaigle:
This pull request introduces several improvements and refactorings across prompt definitions and the SQL client connection pool implementation. The most notable changes include enhanced tool scoping for prompts, improved naming and file references, and a refactor of the transaction-aware connection pool for clarity and maintainability.

**Prompt and Documentation Improvements:**

* Added a `tools` field to prompt YAML frontmatter and updated documentation to recommend explicit tool scoping for each prompt, including a new section detailing tool selection best practices and syntax. This helps ensure prompts only access the tools they need, improving reliability and safety. [[1]](diffhunk://#diff-0e2dd07f5857e96daa1994c503f8942a65c9fabe67c69f70515e48001db42fa2R37) [[2]](diffhunk://#diff-0e2dd07f5857e96daa1994c503f8942a65c9fabe67c69f70515e48001db42fa2L53-R125) [[3]](diffhunk://#diff-0e2dd07f5857e96daa1994c503f8942a65c9fabe67c69f70515e48001db42fa2R152)
* Renamed prompts for clarity (`doc-comments` → `generate-doc-comments`, `test-minimize-overlap` → `refine-test-overlap`) and updated skill and script references to use consistent, absolute paths. [[1]](diffhunk://#diff-7795655bad2b6a0a0faedf121f72b2350696f94c0027db03fd65fadfd25f96d2L2-R2) [[2]](diffhunk://#diff-728c4b02780ec4d5f6779d9d911fbd887f01b49bd335c125c4a0ebf947347d53L2-R6) [[3]](diffhunk://#diff-728c4b02780ec4d5f6779d9d911fbd887f01b49bd335c125c4a0ebf947347d53L13-R27)

**Connection Pool Refactoring and Enhancements:**

* Exposed the `TransactedConnectionPool` through the `IDbConnectionPool` interface and relevant concrete classes, allowing better management of connections enlisted in transactions. [[1]](diffhunk://#diff-9062eeeba73117efeaf91e3e8d4af076600b9e1e175fb47c44a1e08742c23678R88-R92) [[2]](diffhunk://#diff-143f5828ac538b129f458e4224cede7e7134a0c59d2c26253a3ba2d92c898522R99) [[3]](diffhunk://#diff-143f5828ac538b129f458e4224cede7e7134a0c59d2c26253a3ba2d92c898522R152-R154) [[4]](diffhunk://#diff-923fa9d8547589a438d38063dca52e998bbb4120f907f1d1232b98fb3dfc5278R327-R328)
* Refactored `TransactedConnectionPool` to use `TransactedConnections` (a `Dictionary<Transaction, TransactedConnectionList>`) with improved naming, access modifiers, and thread safety in lock usage. The `TransactedConnectionList` is now `internal` for broader accessibility within the assembly. [[1]](diffhunk://#diff-f6b033d5b92465e738fe71f529b134d063ec646cf876f4359c803cdcd80d5ba0L34-R34) [[2]](diffhunk://#diff-f6b033d5b92465e738fe71f529b134d063ec646cf876f4359c803cdcd80d5ba0L63-L65) [[3]](diffhunk://#diff-f6b033d5b92465e738fe71f529b134d063ec646cf876f4359c803cdcd80d5ba0L83-R80) F2ff1f1cL93R93, F2ff1f1cL122R122, [[4]](diffhunk://#diff-f6b033d5b92465e738fe71f529b134d063ec646cf876f4359c803cdcd80d5ba0L126-R127) F2ff1f1cL181R181, F2ff1f1cL212R212, F2ff1f1cL237R237, F2ff1f1cL296R296, F2ff1f1cL318R318)
* Updated the `TryGetConnection` method signature in `IDbConnectionPool` to allow a nullable `TaskCompletionSource`, increasing flexibility in connection retrieval logic.

**Other Codebase Improvements:**

* Exposed `MaxPoolSize` and `MinPoolSize` as `internal` in `WaitHandleDbConnectionPool` for improved testability or cross-class usage.
* Added a summary comment to the `CreationTimeout` property in `DbConnectionPoolOptions` for clearer documentation.

## Issues

#3356 

## Testing

Added comprehensive tests for WaitHandleDbConnectionPool's transacted connection flows. Extended stress tests to include transaction scenarios.


src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.csproj
```diff
@@ -42,9 +42,6 @@
     <PackageReference Include="Microsoft.SqlServer.Server" />
     <PackageReference Include="System.Configuration.ConfigurationManager" />
     <PackageReference Include="System.Security.Cryptography.Pkcs" />
-
-    <!-- Transitive dependencies that would otherwise bring in older, vulnerable versions. -->
-    <PackageReference Include="System.Text.Json" />
```
benrr101, line 53: Don't we want our ref packages to have the same references as the implementation packages?

src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider.csproj
```diff
@@ -36,7 +36,6 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="Azure.Core" />
-    <PackageReference Include="System.Text.Encodings.Web" />
```
benrr101: How is this related to adding stress tests for the connection pool?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/TransactedConnectionPool.cs
```diff
@@ -97,6 +94,8 @@ internal TransactedConnectionPool(IDbConnectionPool pool)
     /// <value>The IDbConnectionPool instance that owns this transacted pool.</value>
     internal IDbConnectionPool Pool { get; }
 
+    internal Dictionary<Transaction, TransactedConnectionList> TransactedConnections { get; private set; }
```
paulmedynski: Don't need private set here.

src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider.csproj
paulmedynski: What's the scoop with these package removals?  Have the transitive problems been solved somehow?

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/WaitHandleDbConnectionPoolTransactionStressTest.cs
```diff
@@ -0,0 +1,1022 @@
...
+using Xunit;
+
+namespace Microsoft.Data.SqlClient.UnitTests.ConnectionPool;
+
+/// <summary>
+/// Stress tests for WaitHandleDbConnectionPool transaction functionality under high concurrency and load.
+/// These tests verify that pool metrics remain consistent when connections are rapidly opened and closed
+/// with intermingled transactions in a highly concurrent environment.
+/// </summary>
+public class WaitHandleDbConnectionPoolTransactionStressTest : IDisposable
```
benrr101: I'm really really really hesitant to approve putting stress tests into the unit test project. Our goal here should be to have unit tests be atomic tests of units of code that can be mocked and stubbed to coerce certain desired behaviors. Stress tests don't fit that bill, since they often are dependent on unexpected behaviors outside the scope of the code itself (system load, primarily). 

We've sort of boxed ourselves into a corner here. The test TDS server is way too large to really be considered "unit testing". The code itself isn't structured for proper unit testing. And the stress test project isn't really ready for things to be moved into it.

I'd really like it if we address this from a different angle. Either move it into the stress test project (and I can help with that since I've still got the stress test project plans in my head a bit), or come up with a way to get the fake TDS server tests/code out of (or isolated within) the unit test project.

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/WaitHandleDbConnectionPoolTransactionStressTest.cs
```diff
@@ -0,0 +1,1022 @@
...
+                    _pool.ReturnInternalConnection(connection, owner);
+                    scope.Complete();
+                }
+            });
+        }
+
+        Task.WaitAll(tasks);
+
+        // Assert
+        AssertPoolMetrics(_pool);
```
paulmedynski: Should you be checking that the min/max are never exceeded at each iteration?  This will just check that the pool _currently_ hasn't exceeded those thresholds, not that it never did.

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/WaitHandleDbConnectionPoolTransactionStressTest.cs
```diff
@@ -0,0 +1,1022 @@
...
+                using var scope = new TransactionScope();
+                var transaction = Transaction.Current;
+                Assert.NotNull(transaction);
+
+                // Get multiple connections within same transaction
+                for (int i = 0; i < iterationsPerThread; i++)
+                {
+                    var owner = new SqlConnection();
+                    _pool.TryGetConnection(owner, null, new DbConnectionOptions("", null), out var conn);
+                    Assert.NotNull(conn);
```
paulmedynski: Is there a way to confirm that this `conn` is enlisted with `transaction`?

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/WaitHandleDbConnectionPoolTransactionStressTest.cs
```diff
@@ -0,0 +1,1022 @@
...
+        using (var scope = new TransactionScope(TransactionScopeAsyncFlowOption.Enabled))
+        {
+            transaction = Transaction.Current;
+            Assert.NotNull(transaction);
+            // Act - Each transaction should be isolated
+            for (int t = 0; t < threadCount; t++)
+            {
+                tasks[t] = Task.Run(() =>
+                {
+                    using (var innerScope = new TransactionScope(transaction))
```
paulmedynski: What effect does `innerScope` have here?  Should `conn` be associated to the inner scope, the outer scope, something else?

src/Microsoft.Data.SqlClient/netcore/ref/Microsoft.Data.SqlClient.csproj
mdaigle: I'll clean these up as part of DRI this sprint. Had to include them here to get some build errors addressed.

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/WaitHandleDbConnectionPoolTransactionStressTest.cs
mdaigle: I agree, but these tests don't use the test TDS server, they use mocked connections.

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/WaitHandleDbConnectionPoolTransactionStressTest.cs
```diff
@@ -0,0 +1,1022 @@
...
+                        _pool.ReturnInternalConnection(conn, owner);
+                    }
+
+                    scope.Complete();
+                }
+            });
+        }
+
+        // Shutdown pool while operations are in progress
+        _pool.Shutdown();
```
paulmedynski: Are we certain there are operations in progress here?  It's possible all tasks have run to completion already.

src/Microsoft.Data.SqlClient/add-ons/AzureKeyVaultProvider/Microsoft.Data.SqlClient.AlwaysEncrypted.AzureKeyVaultProvider.csproj
mdaigle: I'll clean these up. Visual studio was complaining, so I did this to appease it. I think it was just holding on to stale partial build state.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
```diff
@@ -1102,9 +1102,6 @@
     <PackageReference Include="Microsoft.SqlServer.Server" />
     <PackageReference Include="System.Configuration.ConfigurationManager" />
     <PackageReference Include="System.Security.Cryptography.Pkcs" />
-
-    <!-- Transitive dependencies that would otherwise bring in older, vulnerable versions. -->
-    <PackageReference Include="System.Text.Json" />
```
priyankatiwari08: These packages were likely removed due to pruning errors in Visual Studio. This issue has been addressed in https://github.com/dotnet/SqlClient/pull/3686, so merging this branch with main should resolve it.

mdaigle:
Closing to reduce open PRs and will reopen when work resumes.

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/WaitHandleDbConnectionPoolTransactionStressTest.cs
mdaigle: Adjusted all of the tests to run deterministically. Stress scenarios were moved to the stress test suite.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/TransactedConnectionPool.cs
```diff
@@ -96,7 +93,9 @@ internal TransactedConnectionPool(IDbConnectionPool pool)
     /// Gets the main connection pool that this transacted pool is associated with.
     /// </summary>
     /// <value>The IDbConnectionPool instance that owns this transacted pool.</value>
-    internal IDbConnectionPool Pool { get; }
+    internal IDbConnectionPool Pool { get; init; }
```
paulmedynski: Is it even possible to init these properties?  There is no default constructor, so they're already being initialized by the (sole) constructor on line 77.  I don't think these need `init`.  I think just `get` is sufficient.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/TransactedConnectionPool.cs
```diff
@@ -123,9 +122,9 @@ internal TransactedConnectionPool(IDbConnectionPool pool)
         TransactedConnectionList? connections;
         bool txnFound = false;
 
-        lock (_transactedCxns)
+        lock (TransactedConnections)
```
paulmedynski, line 125: Locking on method calls is risky business in C#.  However, in this single case (get‑only auto‑property), it is safe.  Adding a comment explaining this coupling on line 98 is probably a good idea.

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/WaitHandleDbConnectionPoolTransactionTest.cs
```diff
@@ -0,0 +1,943 @@
...
+/// These tests exercise transacted connection pathways with controlled synchronization
+/// to verify correct behavior without relying on probabilistic concurrency.
+/// </summary>
+public class WaitHandleDbConnectionPoolTransactionTest : IDisposable
+{
+    private const int DefaultMaxPoolSize = 50;
+    private const int DefaultMinPoolSize = 0;
+    private const int DefaultCreationTimeoutInMilliseconds = 15000;
+
+    private IDbConnectionPool _pool = null!;
```
paulmedynski, line 29: It may be worth creating a few derived fixtures that supply the pool creation values to the base class constructror.  Then you could put the CreatePool() body in this base class constructor, and make `_pool` readonly and never null.  That would eliminate any possible misuse.
```
public class WaitHandleDbConnectionPoolTransactionTest : IDisposable
{
  protected readonly IDbConnectionPool _pool;

  public WaitHandleDbConnectionPoolTransactionTest( args with default values)
  {
    // Initialize the readonly _pool
  }
}

public class NoAffinityFixture : WaitHandleDbConnectionPoolTransactionTest
{
  public NoAffinityFixture() : base(hasTransactionAffinity: false)
  {
  }
}
```

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/WaitHandleDbConnectionPoolTransactionTest.cs
```diff
@@ -0,0 +1,943 @@
...
+    }
+
+    private DbConnectionInternal GetConnection(SqlConnection owner)
+    {
+        _pool.TryGetConnection(
+            owner,
+            taskCompletionSource: null,
+            new DbConnectionOptions("", null),
+            out DbConnectionInternal? connection);
+        return connection!;
```
paulmedynski, line 87: Should this throw if `connection` is null?  Or if `TryGetConnection()` returns false?

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/WaitHandleDbConnectionPoolTransactionTest.cs
```diff
@@ -0,0 +1,943 @@
...
+
+        var owner1 = new SqlConnection();
+        var conn1 = GetConnection(owner1);
+        Assert.NotNull(conn1);
+        ReturnConnection(conn1, owner1);
+
+        // Act - nested scope with Required shares the same transaction
+        using (var innerScope = new TransactionScope(TransactionScopeOption.Required))
+        {
+            Assert.Equal(outerTxn, Transaction.Current);
```
paulmedynski: Do we expect these to be `Same()` rather than `Equal()` ?  Here and wherever we're comparing transaction instances.

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/WaitHandleDbConnectionPoolTransactionTest.cs
```diff
@@ -0,0 +1,943 @@
...
+    private IDbConnectionPool _pool = null!;
+
+    public WaitHandleDbConnectionPoolTransactionTest()
+    {
+        _pool = CreatePool();
+    }
+
+    public void Dispose()
+    {
+        _pool?.Shutdown();
```
paulmedynski, line 41: Is there value in checking for any incomplete transactions here?  Just to keep the tests honest.

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/WaitHandleDbConnectionPoolTransactionTest.cs
```diff
@@ -0,0 +1,943 @@
...
+        Assert.Same(txn, Transaction.Current);
+        var owner3 = new SqlConnection();
+        var conn3 = GetConnection(owner3);
+        Assert.Same(conn1, conn3);
+        ReturnConnection(conn3, owner3);
+
+        // Assert - single transaction entry
+        Assert.Single(_pool.TransactedConnectionPool.TransactedConnections);
+
+        scope3.Complete();
```
paulmedynski, line 494: Does the order of completion matter here?  Same question above.

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/WaitHandleDbConnectionPoolTransactionTest.cs
```diff
@@ -0,0 +1,943 @@
...
+        });
+
+        acquiredSignal.Wait(); // Wait until thread 1 has the connection
+        Assert.Equal(1, _pool.Count);
+
+        // Release and let thread 1 return the connection
+        holdBarrier.Set();
+        holdTask.Wait();
+
+        // Assert - now we can get a connection again
```
paulmedynski: This test doesn't attempt to acquire a connection while the pool is saturated.  It's just a complicated serial get/return.

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/WaitHandleDbConnectionPoolTransactionTest.cs
```diff
@@ -0,0 +1,943 @@
...
+            innerScope.Complete();
+        });
+
+        var task2 = Task.Run(() =>
+        {
+            barrier.SignalAndWait(); // Sync: wait for thread 1 to get connection first
+
+            using var innerScope = new TransactionScope(transaction);
+            var owner = new SqlConnection();
+            var conn = GetConnection(owner);
```
paulmedynski: Should we `Assert.Same()` the two connection refs after both tasks complete?

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/WaitHandleDbConnectionPoolTransactionTest.cs
```diff
@@ -0,0 +1,943 @@
...
+        // Arrange
+        using var scope = new TransactionScope();
+        var owner = new SqlConnection();
+        var conn = GetConnection(owner);
+        Assert.NotNull(conn);
+
+        // Act - shutdown while connection is held (not yet returned)
+        _pool.Shutdown();
+
+        // Return after shutdown - should not throw
```
paulmedynski: Ok, but what do we expect returning to _do_ here?  What does it mean to return after shutdown?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/TransactedConnectionPool.cs
mdaigle: This is part of the behavior I plan to refactor. The tests in this PR plus the TransactedConnectionPoolTests are intended to give a baseline so that I can refactor more safely.

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/WaitHandleDbConnectionPoolTransactionTest.cs
mdaigle: Good suggestion. I kept it simple for now since only one test overrides the defaults (`GetConnection_WithTransactionAffinityDisabled_SkipsTransactedPool`), but I can refactor to the derived fixture pattern in a follow-up if the test count grows.

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/WaitHandleDbConnectionPoolTransactionTest.cs
mdaigle: The order is mandated by `TransactionScope` — inner scopes must be completed/disposed before outer ones. Completing out of order would cause the outer scope to abort. So the order here (`scope3`, `scope2`, `scope1`) follows the required nesting semantics.

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/WaitHandleDbConnectionPoolTransactionTest.cs
mdaigle: Good idea. Added a check in `Dispose()` that asserts `TransactedConnections` is empty, so any leaked transactions will fail the test immediately.

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/WaitHandleDbConnectionPoolTransactionTest.cs
mdaigle: Yes, added `Assert.Same(connFromTask1, connFromTask2)` after `Task.WaitAll` to verify both tasks got the same underlying connection via the transacted pool.

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/WaitHandleDbConnectionPoolTransactionTest.cs
mdaigle: Good catch. Reworked this into `PoolSaturation_BlocksUntilConnectionAvailable` — a second thread now attempts to acquire while the pool is full and we assert it blocks until the first thread releases.

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/WaitHandleDbConnectionPoolTransactionTest.cs
mdaigle: Good point — updated to `Assert.Same` for transaction instance comparisons. Within a `Required` nested scope, it's the same ambient `Transaction` object.

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/WaitHandleDbConnectionPoolTransactionTest.cs
mdaigle: The two are linked — `TryGetConnection` returns false if and only if the out parameter is null, so asserting not-null (which we do at each call site) is equivalent to checking the return value.

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/WaitHandleDbConnectionPoolTransactionTest.cs
mdaigle: The return after shutdown deactivates and disposes the connection rather than returning it to the pool. The test verifies this defensive path doesn't throw, since callers may not know the pool has been shut down. Added a comment clarifying the expected behavior.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/ConnectionPool/TransactedConnectionPool.cs
mdaigle: Agreed, updated to `{ get; }` only.

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/WaitHandleDbConnectionPoolTransactionTest.cs
```diff
@@ -566,72 +570,64 @@ public void SharedTransaction_DependentScopes_UseTransactedPool()
...
-        _pool.Shutdown();
-        _pool.Clear();
-        _pool = CreatePool(maxPoolSize: 3);
+        Assert.True(allAcquired.Wait(TimeSpan.FromSeconds(10)),
+            "Timed out waiting for connection to be acquired.");
+        Assert.Equal(1, _pool.Count);
 
-        // Act - acquire all 3 connections under separate transactions and return them
-        for (int i = 0; i < 3; i++)
+        // A fourth thread attempts to acquire while pool is full — it should block
```
paulmedynski: I only see 2 threads here.  This comment and the one on line 592 don't seem accurate.

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/WaitHandleDbConnectionPoolTransactionTest.cs
```diff
@@ -641,89 +637,101 @@ public void PoolSaturation_PostRelease_ConnectionObtainable()
     [Fact]
     public void TwoThreads_SeparateTransactions_IsolatedTransactedEntries()
     {
-        // Arrange
-        using var barrier = new Barrier(2);
+        // Arrange - sequence task1 before task2 so the connection is available
+        // in the pool when task2 requests. This verifies that separate transactions
+        // get can get the same physical connection after the first transaction is completed.
```
paulmedynski: get can get

Maybe:

This verifies that a second transaction can acquire the same underlying connection from the pool after the first transaction completes.

paulmedynski: Actually, I'm not sure what this test is trying to confirm.  Should the transactions get the same underlying connection or a different one?  What is the check on line 677 telling me?

mdaigle: This test really became "different transactions get different connections". I'll tweak it to make that the focus and remove the unnecessary threading.

src/Microsoft.Data.SqlClient/tests/UnitTests/ConnectionPool/WaitHandleDbConnectionPoolTransactionTest.cs
mdaigle: missed these. updated

cheenamalhotra (approved):

Issue #3372, opened 2025-05-22
Generate code from SqlMetaData.xml


paulmedynski:
Today we are reading in the `SqlMetaData.xml` file and parsing it into data structures at runtime.  See if we can do that work at compile time and generate some static code or data structures instead to avoid runtime overhead and possible errors.

edwardneal:
It might also be useful to account for where SqlClient stores its records of various types. At present, we've got a number of sources:
* Floating constants in [`TdsEnums`](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsEnums.cs#L412-L445)
* [`MetaType`](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlEnums.cs)
* [`SqlMetaDataPriv`](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/TdsParserHelperClasses.cs#L480)
* SqlMetaData.xml (returned via `SqlConnection.GetSchema("DataTypes")`.)

Having a single place to add new types and to "hydrate" them with a precision/length where necessary would be convenient. In the long term, it might also help to lift some of the parsing logic out of TdsParser (although hopefully SqlClientX will pre-empt that.)

tetolv:
Hello, 
I would like to try tackling this task using source generators, but while studying metadata handling in SqlClient I'm wondering why it is being kept in .xml file in the first place, instead of being hardcoded in C#?

edwardneal:
It's largely historical. This is used to support output for `DbConnection.GetSchema`, and in .NET Framework System.Data.SqlClient made heavy use of shared underlying ADO.NET components. In this situation, it made sense for multiple ADO.NET providers to share an implementation which returned a DataTable, to have that implementation clone a "reference DataSet" and to load that reference DataSet from an embedded resource.

When the library was split out into Microsoft.Data.SqlClient, the relevant shared ADO.NET components were split with it. Since then, we've had some movement:
* #3369 replaced the trim-unfriendly use of `DataSet.ReadXml` to load the reference DataSet with a trim-friendly XML deserializer
* #3818 removed the now-unnecessary inheritance hierarchy between the old ADO.NET DbMetadataFactory and the SqlClient-based SqlMetadataFactory
[...]

PR #3810, opened 2025-11-26
Merge | OS Flags

benrr101:
## Description
 Introduced _WINDOWS/_UNIX compile-time constants and TargetOs selection in src/Microsoft.Data.SqlClient/src/Microsoft.Data.SqlClient.csproj, plus mirrored updates to netcore/src/Microsoft.Data.SqlClient.csproj and netfx/src/Microsoft.Data.SqlClient.csproj so OS-gated code can be compiled from the common project and Windows-only ILLink substitutions are conditionally embedded (Resources/ILLink.Substitutions.xml).

Normalized OS-specific sources to .windows/.unix naming and wired them into the new flags (e.g., LocalDbApi.*, LocalDB.netcore.*, PacketHandle.*, SessionHandle.*, SqlFileStream.*, SqlColumnEncryption*, TdsParser*, SNI interop files), with new netfx-specific shims like PacketHandle.netfx.cs and SessionHandle.netfx.cs. Refactored shared helpers into OS- and target-aware partials: AdapterUtil.cs now delegates netfx- and netcore-specific pieces to AdapterUtil.netfx.cs/AdapterUtil.netcore.cs; DbConnectionPoolIdentity.cs and SqlDataSourceEnumerator.cs now pick managed vs native implementations based on _WINDOWS/_UNIX and LocalAppContextSwitches.UseManagedNetworking; added netfx-only System/Diagnostics/CodeAnalysis.netfx.cs.

Folded the netcore app–specific SSL over TDS stream logic into ManagedSni/SslOverTdsStream.netcore.cs, adding span-based sync/async read handling and packet encapsulation directly in the shared file.
  
Tidied LocalAppContextSwitches.cs for clearer NET vs NETFRAMEWORK behavior (globalization invariant handling and managed SNI defaults) under the new OS constants.

[The above is AI generated]

## Issues
Continuation of work from #1261 

## Testing
Projects still build, and the common project is building too 👀

### DIFF
--- a/src/Microsoft.Data.SqlClient/src/Interop/Windows/Kernel32/Kernel32Safe.netfx.cs
+++ b/src/Microsoft.Data.SqlClient/src/Interop/Windows/Kernel32/Kernel32Safe.netfx.cs
@@ -2,13 +2,12 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#if NETFRAMEWORK
+#if NETFRAMEWORK && _WINDOWS
 
 using System;
 using System.Runtime.InteropServices;
 using System.Runtime.Versioning;
 using System.Security;
-using System.Text;
 
 namespace Interop.Windows.Kernel32
 {
--- a/src/Microsoft.Data.SqlClient/netfx/src/Microsoft.Data.SqlClient.csproj
+++ b/src/Microsoft.Data.SqlClient/netfx/src/Microsoft.Data.SqlClient.csproj
@@ -82,9 +82,6 @@
     <DebugType>Pdbonly</DebugType>
     <Optimize>True</Optimize>
   </PropertyGroup>
-  <PropertyGroup>
-    <DefineConstants>$(DefineConstants);NETFRAMEWORK;</DefineConstants>
-  </PropertyGroup>
   <ItemGroup>
     <Reference Include="System" />
     <Reference Include="System.Configuration" />
@@ -99,7 +96,8 @@
 
   <!-- OS Constants ==================================================== -->
   <PropertyGroup>
-    <DefineConstants Condition="'$(OSGroup.ToUpper())'=='WINDOWS_NT'">$(DefineConstants),_WINDOWS</DefineConstants>
+    <!-- Netfx project only builds for Windows, so we always define _WINDOWS here -->
+    <DefineConstants>$(DefineConstants),_WINDOWS</DefineConstants>
   </PropertyGroup>
 
   <!-- Compiled Files ================================================== -->
--- a/src/Microsoft.Data.SqlClient/src/Interop/Windows/Sni/SniNativeMethodsArm64.netfx.cs
+++ b/src/Microsoft.Data.SqlClient/src/Interop/Windows/Sni/SniNativeMethodsArm64.netfx.cs
@@ -2,7 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#if NETFRAMEWORK
+#if NETFRAMEWORK && _WINDOWS
 
 using System;
 using System.Runtime.InteropServices;
--- a/src/Microsoft.Data.SqlClient/src/Interop/Windows/Sni/SniNativeMethodsNotSupported.netfx.cs
+++ b/src/Microsoft.Data.SqlClient/src/Interop/Windows/Sni/SniNativeMethodsNotSupported.netfx.cs
@@ -2,7 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#if NETFRAMEWORK
+#if NETFRAMEWORK && _WINDOWS
 
 using System;
 using System.Runtime.InteropServices;
--- a/src/Microsoft.Data.SqlClient/src/Interop/Windows/Sni/SniNativeMethodsX64.netfx.cs
+++ b/src/Microsoft.Data.SqlClient/src/Interop/Windows/Sni/SniNativeMethodsX64.netfx.cs
@@ -2,7 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#if NETFRAMEWORK
+#if NETFRAMEWORK && _WINDOWS
 
 using System;
 using System.Runtime.InteropServices;
--- a/src/Microsoft.Data.SqlClient/src/Interop/Windows/Sni/SniNativeMethodsX86.netfx.cs
+++ b/src/Microsoft.Data.SqlClient/src/Interop/Windows/Sni/SniNativeMethodsX86.netfx.cs
@@ -2,7 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#if NETFRAMEWORK
+#if NETFRAMEWORK && _WINDOWS
 
 using System;
 using System.Runtime.InteropServices;
--- a/src/Microsoft.Data.SqlClient/src/Interop/Windows/Sni/SqlDependencyProcessDispatcherStorage.netfx.cs
+++ b/src/Microsoft.Data.SqlClient/src/Interop/Windows/Sni/SqlDependencyProcessDispatcherStorage.netfx.cs
@@ -2,7 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#if NETFRAMEWORK
+#if NETFRAMEWORK && _WINDOWS
 
 using System;
 using System.Diagnostics;


src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
```diff
@@ -40,17 +38,129 @@
     <EmbeddedFiles Include="$(GeneratedAssemblyInfoFile)" />
     <EmbeddedFiles Include="$(TargetFrameworkMonikerAssemblyAttributesPath)" />
   </ItemGroup>
-  <!-- Contains common items shared between Unix and Windows -->
+
+  <!-- OS Constants ==================================================== -->
+  <PropertyGroup>
+    <DefineConstants Condition="'$(OSGroup.ToUpper())'=='UNIX'">_UNIX</DefineConstants>
+    <DefineConstants Condition="'$(OSGroup.ToUpper())'=='WINDOWS_NT'">_WINDOWS</DefineConstants>
```
edwardneal: There are likely to be other points to untangle, but netcore automatically defines the appropriate `WINDOWS` preprocessor symbol. It might be better to polyfill this in the netfx project rather than define a new one.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
```diff
@@ -834,265 +1007,77 @@
...
-    <Compile Include="$(CommonSourceRoot)Microsoft\Data\SqlTypes\SqlFileStream.netcore.Unix.cs">
-      <Link>Microsoft\Data\SqlTypes\SqlFileStream.netcore.Unix.cs</Link>
-    </Compile>
 
-    <EmbeddedResource Include="$(CommonSourceRoot)Resources\ILLink.Substitutions.Unix.xml">
-      <LogicalName>ILLink.Substitutions.xml</LogicalName>
-      <Link>Resources\ILLink.Substitutions.Unix.xml</Link>
+    <EmbeddedResource Include="$(CommonSourceRoot)Resources\Microsoft.Data.SqlClient.SqlMetaData.xml">
+      <Link>Resources\Microsoft.Data.SqlCLient.SqlMetaData.xml</Link>
+      <LogicalName>Microsoft.Data.SqlClient.SqlMetaData.xml</LogicalName> <!-- @TODO: How necessary is this? -->
```
edwardneal, line 1053: It's not necessary, the code to replace it is just bulky - #3372

src/Microsoft.Data.SqlClient/src/Interop/Windows/Kernel32/Kernel32Safe.netfx.cs
```diff
@@ -2,7 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#if NETFRAMEWORK
+#if NETFRAMEWORK && _WINDOWS
```
edwardneal: Doesn't `NETFRAMEWORK` imply Windows-only code?
(here and elsewhere)

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/PacketHandle.netcore.unix.cs
```diff
@@ -2,20 +2,23 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#if NET
+// @TODO: Introduce polymorphism to remove need for this level of indirection
+// @TODO: Also, why do we have any other type besides managed defined here?
```
edwardneal, line 6: PacketHandles are used to transport either native or managed packets around TdsParserStateObject.

src/Microsoft.Data.SqlClient/src/Resources/ILLink.Substitutions.Unix.xml
```diff
@@ -1,8 +0,0 @@
-<linker>
-  <assembly fullname="Microsoft.Data.SqlClient">
-    <type fullname="Microsoft.Data.SqlClient.LocalAppContextSwitches">
-      <!-- The Switch.Microsoft.Data.SqlClient.UseManagedNetworkingOnWindows switch must always be true on Unix platforms. -->
-      <method signature="System.Boolean get_UseManagedNetworking()" body="stub" value="true" />
```
edwardneal, line 5: This is different to the Windows ILLink.Substitutions.xml resource, and shouldn't be disregarded.

edwardneal (commented):
Very early review here, but the comment in `SqlColumnEncryptionCspProvider` has wider implications - how much of the work needs to be done at compile time? Could we replace the conditional compilations with runtime OS checks?

If that works then we could eliminate the need to build and package for multiple OSes completely.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/PacketHandle.netcore.unix.cs
benrr101: I know we discussed this before, but I'm still not convinced ... so I left these comments to consider it further.
But considering on unix, we never have native packets, I don't know why we need to have an object with both types defined.

src/Microsoft.Data.SqlClient/src/Resources/ILLink.Substitutions.Unix.xml
benrr101: Yes, but all it does is hard code UseManagedNetworking to `true` on unix - which at the time wasn't defined in the app context switches. We can do that with code that can be switched on/off via pre-compiler flags and eliminate a file that would confuse the heck out of someone unfamiliar with it.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
benrr101: Haha, I was specifically calling out the logical name tag, but yeah if we can junk the xml file altogether I'd approve of it.

src/Microsoft.Data.SqlClient/src/Interop/Windows/Kernel32/Kernel32Safe.netfx.cs
benrr101: yeah, probably ... I'm a bit torn on these. I think no matter which way it's done, it's a little confusing.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
benrr101: While true, as far as I know, there's no way to override those. We want to be able to build unix on windows and windows on unix (and to a lesser extent want to be able to get IDE highlighting for one on the other), so we need a flag that can be overridden. This seems like the best way to do it, as per the internet :)

benrr101:
@edwardneal As to your overall comment that it would be nice to eliminate the compile-time checks: yes, yes it would be nice :) @paulmedynski and I chatted about it and I think we came to the conclusion that this was a decent enough solution for now. A lot exists because we have native SNI - and it'd be nice if we can deprecate native SNI.

paulmedynski:
We will be auditing the compile-time checks (not as part of this PR) to see how many of them could become inexpensive runtime checks (i.e. checked once per runtime, or "not often").  If the runtime costs are very low, then we would consider converting them in order to gain project architecture efficiencies and eliminate a large chunk of the build setup that exists to support OS-specific artifacts.

edwardneal:
Thanks - that sounds good. The JIT will lift the OS-level checks via `OperatingSystem.IsWindows` in Release configuration, so they're very cheap and probably don't need to be cached (see SharpLab examples for [netcore](https://sharplab.io/#v2:C4LghgzgtgPgAgJgIwFgBQcDMACR2DC2A3utmbjgJYB2w2ASgKbACuATtQGpgA2LjACgCUxUuXGUAZtgEB5AA6M2YYDQDmcJADoAkhADqNACYB7AO4RhQseLIk0tx7gDs2AAwAPNwA4A3DdsAXwDxRh4IRhDyeydbOFdPJDd/Bydg1LJ08XQAYilsADkAUQAVADF6AEEAWSL9WXoAaQCsXCQANlwEbAUlFXVNAJinVs1OgCMTEx5sPUNqUwthbABeAD5sYDZ+FPF0nMYFqXRAoA=) and [netfx](https://sharplab.io/#v2:EYLgHgbALANALiAhgZwLYB8ACAmAjAWAChMBmAAhzIGEyBvIsxi8gSwDs4yAlAUzgFcATmwBqiADb8eACgCUdBkyUsAZmWkB5AA49BiOOwDmmXADoAksgDq7ACYB7AO7I5sxUsb1CHnxQDsZAAMYIEAHADc7h4AvlFKPOLIPHFMXr4emAHBuIGR3r6x+YyFSkQAxKpkAHIAogAqAGJcAIIAsjVWGlwA0lGkFLgQFNhk2rr6RiZRab79JkPA9vbiZJY2bA7OcmQAvAB8ZHCCUnlKhWU8G6pE0UA==).)

src/Microsoft.Data.SqlClient/src/Resources/ILLink.Substitutions.Unix.xml
edwardneal: The original reason for having two files was to block someone from switching off the managed SNI on Linux using the below RuntimeHostConfigurationOption in their client application's csproj when trying to publish in a trimmed environment:
```xml
<RuntimeHostConfigurationOption
    Include="Switch.Microsoft.Data.SqlClient.UseManagedNetworkingOnWindows"
    value="false"
/>
```

On second thoughts though, I don't think the use of the Windows ILLink.Substitutions.xml resource will cause a problem - it's just a guardrail on invalid usage of the library, trimming isn't technically supported yet, and this could be wrapped up in a runtime-level OS check.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
```diff
@@ -40,17 +38,129 @@
     <EmbeddedFiles Include="$(GeneratedAssemblyInfoFile)" />
     <EmbeddedFiles Include="$(TargetFrameworkMonikerAssemblyAttributesPath)" />
   </ItemGroup>
-  <!-- Contains common items shared between Unix and Windows -->
+
+  <!-- OS Constants ==================================================== -->
+  <PropertyGroup>
+    <DefineConstants Condition="'$(OSGroup.ToUpper())'=='UNIX'">_UNIX</DefineConstants>
```
paulmedynski: These will clobber any existing `DefineConstants` values.  Probably best to append:
```
<DefineConstants ...>$(DefineConstants),_UNIX</DefineConstants>
```

src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
```diff
@@ -834,265 +1007,77 @@
...
-    <EmbeddedResource Include="$(CommonSourceRoot)Resources\Microsoft.Data.SqlClient.SqlMetaData.xml">
-      <LogicalName>Microsoft.Data.SqlClient.SqlMetaData.xml</LogicalName>
-      <Link>Resources\Microsoft.Data.SqlClient.SqlMetaData.xml</Link>
+      <LogicalName>Microsoft.Data.SqlCLient.Resources.%(Filename).resources</LogicalName>
     </EmbeddedResource>
+
+    <!-- Strings designer that depends on the resx -->
+    <Compile Include="$(CommonSourceRoot)Resources\Strings.Designer.cs">
+      <Link>Resources\Strings.Designer.cs</Link>
+      <DesignTime>True</DesignTime>
```
paulmedynski, line 1074: I don't see `DesignTime` as a sub-element of `Compile`:

https://learn.microsoft.com/en-us/visualstudio/msbuild/common-msbuild-project-items?view=visualstudio#compile

What is the intention here?

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/LocalAppContextSwitches.cs
```diff
@@ -437,30 +463,34 @@ public static bool UseManagedNetworking
                 return s_useManagedNetworking == Tristate.True;
             }
         }
-#else
+        #else
         /// <summary>
-        /// .NET Framework does not support Globalization Invariant mode, so this will always be false.
+        /// .NET Core on Unix does not support the native SNI, so this will always be false.
```
paulmedynski: Shouldn't this always be true on Unix?

src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
benrr101: Good thought 👍

src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
benrr101: Welllllllllll - the link does say "common" items, not "all" items. And DesignTime is specified in the xsd for csproj.

I'm not 100% how this strings file works in our build, but my understanding is it's autogenerated, so we indicate that it it is written during design time. Check out this page for more details on some of that https://github.com/dotnet/project-system/blob/main/docs/design-time-builds.md

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/LocalAppContextSwitches.cs
benrr101: Yep, you are right, idk what I was thinking.

src/Microsoft.Data.SqlClient/netcore/src/Microsoft.Data.SqlClient.csproj
paulmedynski: I guess this is something used by Visual Studio, but has no effect in a real build.

paulmedynski (dismissed):

src/Microsoft.Data.SqlClient/tests/Common/LocalAppContextSwitchesHelper.cs
```diff
@@ -33,10 +33,16 @@ public sealed class LocalAppContextSwitchesHelper : IDisposable
     private readonly PropertyInfo _truncateScaledDecimalProperty;
     private readonly PropertyInfo _ignoreServerProvidedFailoverPartner;
     private readonly PropertyInfo _enableUserAgent;
-#if NET
+
+    #if NET
     private readonly PropertyInfo _globalizationInvariantModeProperty;
+    #endif
+    
+    #if NET && _WINDOWS
```
mdaigle, line 41: This definitely reads more cleanly than nested directives 👍

src/Microsoft.Data.SqlClient/src/Interop/Windows/Kernel32/Kernel32Safe.netfx.cs
paulmedynski: I think it's best to keep them separate.  Imagine if we decided to officially support Mono and we wanted NETFRAMEWORK but not _WINDOWS.

PR #3813, opened 2025-11-28
[6.1] Fix code coverage job disk full

paulmedynski:
## Description

**Backport of #3798 to 6.1**

The code coverage jobs are running out of disk space. They appear to consume upwards of 12GB of space to merge/convert 3GB of coverage logs from the various test jobs. We can diagnose why that is later. For now, I have:

- Switched the coverage job to use the ADO-MMS22-CodeCov 1ES image rather than a generic Azure Pipelines image.
  - The generic images have 14GB of disk space.
  - Our custom 1ES image has much more space.
- Removed unnecessary parameters/variables for code coverage job.
- Added debug output to help see disk usage throughout the job.

### DIFF
--- a/eng/pipelines/dotnet-sqlclient-ci-core.yml
+++ b/eng/pipelines/dotnet-sqlclient-ci-core.yml
@@ -92,9 +92,6 @@ variables:
   - name: artifactName
     value: Artifacts
 
-  - name: defaultHostedPoolName
-    value: 'Azure Pipelines'
-
 stages:
   - stage: build_nugets
     displayName: 'Build NuGet Packages'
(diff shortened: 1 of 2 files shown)


eng/pipelines/dotnet-sqlclient-ci-core.yml
```diff
@@ -152,16 +149,9 @@ stages:
...
-              - ${{ each targetFramework in parameters.codeCovTargetFrameworks }}:
-                - task: DownloadPipelineArtifact@2
-                  displayName: 'Download Coverage Reports [${{ targetFramework }}]'
-                  inputs:
-                    itemPattern: '**\${{ targetFramework }}*'
-                    ${{ if contains(targetFramework, 'net4') }}:
-                      targetPath: '$(Build.SourcesDirectory)\coverageNetFx'
-                    ${{ else }}:
-                      targetPath: '$(Build.SourcesDirectory)\coverageNetCore'
+            image: ADO-MMS22-CodeCov
```
mdaigle, line 152: I'm not really on board with this change. The MMS images are "kitchen sink" (including a ton of dependencies we don't need) and were created to facilitate migration from customer hosted pools to 1ES hosted pools. Using this image also requires us to hold space in our agent pools for these MMS images, duplicate them across regions for stability, etc. We should really try to use a vanilla image.

mdaigle: Will approve for now so that we can get disk usage details to fix the root cause.

paulmedynski: This is the image being used on the 5.1 branch.  The only thing it adds to the base image is a couple of .NET SDKs/Runtimes.

The vanilla images give you 14GB of disk space, we're down to about 12GB after agent prep.

I think our best bet will be to simply stop using integration tests for coverage, and greatly reduce the number of jobs producing coveage data to be merged.

mdaigle: It builds on top of the base MMS image. If you look at that base image's dependencies, you'll see how much is really in there: https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md
https://eng.ms/docs/coreai/devdiv/one-engineering-system-1es/1es-docs/1es-github-runners/createpoolportal

PR #3818, opened 2025-12-01
Flatten | DbMetaDataFactory -> SqlMetaDataFactory

edwardneal:
## Description

This continues some of the earlier work from #3435, which replaced a layer of inheritance between `DbConnectionFactory` and `SqlConnectionFactory`, and #3773, which does the same between `SqlInternalConnection` and `SqlInternalConnectionTds`.

Here, we remove `DbMetaDataFactory` and fold its responsibilities into `SqlMetaDataFactory`. These are comparatively large files, so it's simplest to move commit-by-commit with whitespace changes hidden. I proceed in stages:
1. Replace necessary references to DbMetaDataFactory with SqlMetaDataFactory.
2. Identify a dead code path: the ability to supply a normalized (i.e. zero-padded) server version string. SqlClient never used this capability.
3. Perform general code cleanup prior to the merge. This cleanup was a handful of fairly small points which are designed to bring the final file into alignment with the current code style.
4. Move methods and constants. At this stage, methods are being copied and pasted from source to destination (plus some touch-ups of their scope between `protected` and `private` as appropriate.)
5. Post-hoc touch-ups to remove the inheritance and unused methods.

## Issues

No associated issues.

## Testing

Automated tests continue to pass.


src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionFactory.cs
```diff
@@ -756,16 +743,13 @@ private static DbConnectionPoolGroupOptions CreateConnectionPoolGroupOptions(Sql
...
-            DbConnectionInternal internalConnection,
-            out bool cacheMetaDataFactory)
+        private static SqlMetaDataFactory CreateMetaDataFactory(DbConnectionInternal internalConnection)
         {
             Debug.Assert(internalConnection is not null, "internalConnection may not be null.");
 
             Stream xmlStream = Assembly.GetExecutingAssembly().GetManifestResourceStream("Microsoft.Data.SqlClient.SqlMetaData.xml");
             Debug.Assert(xmlStream is not null, $"{nameof(xmlStream)} may not be null.");
             
-            cacheMetaDataFactory = true;
```
mdaigle, line 768: fun stuff! 😆

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlMetaDataFactory.cs
```diff
@@ -41,229 +43,193 @@ private void addUDTsToDataTypesTable(DataTable dataTypesTable, SqlConnection con
...
-                {
-
-                    reader.GetValues(values);
-                    DataRow newRow = dataTypesTable.NewRow();
+                reader.GetValues(values);
+                DataRow newRow = dataTypesTable.NewRow();
 
-                    newRow[providerDbtype] = SqlDbType.Structured;
+                newRow[providerDbType] = SqlDbType.Structured;
+                newRow[columnSize] = values[ColumnSizeSqlIndex];
```
mdaigle, line 631: Can you explain why it's appropriate to remove the DBNull check here, but not for IsNullable?

edwardneal: This (and one or two other instances) is a result of the underlying database schema, removing a redundant DBNull check on a non-nullable field. In this case, we're looking at the `sys.types` table, where the `is_nullable` column is nullable but the `name` and `max_length` columns are non-nullable.

src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/SqlConnectionFactory.cs
edwardneal: Definitely! It was probably an artifact of netfx's S.D.S sharing an assembly with the base DbConnectionFactory type.

paulmedynski:
@edwardneal - Can you try merging main?  We have made changes to the dependencies that may fix the one build error.

edwardneal:

paulmedynski:
@edwardneal - Some conflicts to resolve here, and then I'll kick off a new CI run.

edwardneal:
Thanks @paulmedynski, I've just re-merged.

mdaigle (approved):

PR #3832, opened 2025-12-05
Test | Kill tests that run longer than 5 mins.

mdaigle:
## Description

Poor test design and resource contention sometimes lead to test cases that run longer than desired. In other cases, tests are genuinely stuck due to deadlocks. In these situations, it's best to kill the test case and rerun. Unfortunately, because we use xunit2 and vstest, we're stuck with rerunning the whole test suite as there's no option to rerun only failing tests.

All the same tests are run, but they should now "fail fast" instead of sitting and consuming resources.

I'll leave this PR open for a bit and rerun the pipeline on it to see how useful this behavior is in practice.

### DIFF
--- a/build.proj
+++ b/build.proj
@@ -226,6 +226,7 @@
         --collect "Code coverage"
         --results-directory $(ResultsDirectory)
         --filter "category!=failing%26category!=flaky"
+        --blame-hang-timeout 5m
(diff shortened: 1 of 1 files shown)


paulmedynski:
We can also consider an approach like this to get start/end logging, and per-test re-runs on hangs or transient failures:

https://andrewlock.net/tracking-down-a-hanging-xunit-test-in-ci-building-a-custom-test-framework/

PR #3836, opened 2025-12-10
Cleanup | Remove ResourceExposure and ResourceConsumption attributes

edwardneal:
## Description

This cleanup builds on the initial split of System.Data.SqlClient from the rest of netfx. Various methods within the codebase were decorated with [`[ResourceExposure]`](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.versioning.resourceexposureattribute?view=net-10.0) and [`[ResourceConsumption]`](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.versioning.resourceconsumptionattribute?view=net-10.0).

These attributes were in turn decorated with `[Conditional("RESOURCE_ANNOTATION_WORK")]`, which means that they were never compiled into the finished binary unless that conditional compilation symbol was set. It's only set for debug builds, so we don't need to include the attributes. Doing so also removes some unnecessary netfx-specific conditional compilation.

## Issues

None.

## Testing

Test cases continue to pass.


src/Microsoft.Data.SqlClient/src/Interop/Windows/Sni/SniNativeWrapper.cs
```diff
@@ -14,9 +14,6 @@
 using Microsoft.Data.SqlClient.Utilities;
 
 #if NETFRAMEWORK
-using System.Runtime.CompilerServices;
-using System.Runtime.Versioning;
-using System.Security;
```
paulmedynski, line 19: Can we remove some of the `<Reference>` elements from the csproj files?  For example in netfx:
```
<Reference Include="System.Security" />
```

edwardneal: I can compile if I remove references to System, System.Runtime.Serialization, System.Security, System.Data, System.Xml, System.Runtime.Caching. Of these, I think only System.Runtime.Caching can be removed due to not being used - the rest of them can be removed because they're already part of netfx.

edwardneal:
This PR is not stale.

edwardneal:
This PR is not stale.

paulmedynski:
@edwardneal - Needs some conflicts resolved.

edwardneal:

